PR #10, opened 2016-08-28
Added Dockerfiles

faddat:
Added dockerfiles because I can see how they might come in handy ;)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,3 @@
+FROM golang:onbuild
+COPY config-example.json /usr/bin/config.json
+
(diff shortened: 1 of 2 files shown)


thrasher-:
Thanks for that!!

faddat:
Any time :).

I very much like what you're doing, so "Thanks for that!" to you too!  I
may need to do some tweakage on the dockerfile for the app itself, but the
web app one is rolling along fine.

Jacob Gadikian
E-mail: faddat@gmail.com
SKYPE: faddat
Phone/SMS: +84 167 789 6421

PR #12, opened 2016-09-01
Update Dockerfile

faddat:
I think that the Dockerfile will work now.  Sorry about the previous lame attempt.


faddat:
You don't need/want this.

PR #28, opened 2017-04-09
GDAX support

140am:
This aims to address a few issues with GDAX:
- update websocket API endpoint to restore websocket feed data (old one no longer resolves)
- API auth based on unix time stamp and hashed path instead of URL (auth fails otherwise).
- support alternate currency notation on GDAX (AAA-BBB and AAABBB). the configuration currently uses AAABBB format on GDAX for calls to GetTickerPrice() but that method requires AAA-BBB format. this is used by example AllActiveExchangesAndCurrencies and addresses account retrieval via that.

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -11,6 +11,7 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"hash"
 	"io"
 	"io/ioutil"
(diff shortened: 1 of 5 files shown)


140am:
GDAX and Huobi were the only two exchanges which failed to retrieve product information via AllActiveExchangesAndCurrencies (/exchanges/enabled/latest/all) for me in current `master`. The "alternate currency format" patch here should address GDAX but there might be a better way to address the issue of configuration having one format and GDAX API requiring a different one.

PR #29, opened 2017-04-28
pls use api.btcchina.com to replace api.btcc.com

starit:
BTCC and BTTChina are two different sites now. The market of btccny is in BTCChina. So please use api.btcchina.com rather then api.btcc.com


thrasher-:
Thanks!

PR #31, opened 2017-06-24
Fix typo and type error in PoloniexDepositsWithdrawals

if1live:
## Golang cannot parse timestamp as time.Time

example code

https://play.golang.org/p/FLrGM2x4v8

```golang
package main

import (
	"fmt"
	"time"
	"encoding/json"
)

type Dummy struct {
	Val time.Time `json:"val"`
}

func main() {
	str := `{"val":1497314214}`
	data := []byte(str)

	var resp Dummy
	err := json.Unmarshal(data, &resp)
	if err != nil {
		panic(err)
	}
	fmt.Println(resp)
}
```

output
```
panic: parsing time "1497314214" as ""2006-01-02T15:04:05Z07:00"": cannot parse "1497314214" as """
```

so I modify type from `time.Time` to `int64`

## typo
string -> status
```
- Status string `json:"string"`
+Status string `json:"status"`
```

### DIFF
--- a/exchanges/poloniex/poloniex_types.go
+++ b/exchanges/poloniex/poloniex_types.go
@@ -1,9 +1,5 @@
 package poloniex
 
-import (
-	"time"
-)
-
 type PoloniexTicker struct {
 	Last          float64 `json:"last,string"`
 	LowestAsk     float64 `json:"lowestAsk,string"`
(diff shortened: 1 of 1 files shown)


thrasher-:
Thanks!

PR #50, opened 2017-09-09
Start with different nonce value for Liqui

kpashka:
After initial exchange setup, I was getting `Error encountered retrieving exchange account info for Liqui. Error invalid nonce`

After looking at the doc, I've found this https://liqui.io/api:

`Minimum nonce value - 1, maximum - 4294967294`

So setting elapsed unix nanoseconds was higher than maximum value.


thrasher-:
Thanks @kpashka! In line with other exchanges I've set it to time.Unix() which will be valid until Feb 7th, 2106. 

https://github.com/thrasher-/gocryptotrader/commit/89848eb6e9e9229096488a57037aeb7e4764ec51

PR #60, opened 2017-11-01
consistent access to sandbox environments on gemini and gdax

mattkanwisher:
Allow users to run the bot in sandbox mode, so we can test out new functionality


thrasher-:
ACK, thanks!

PR #65, opened 2017-11-24
Added database package with a ORM management system SQLBoiler.

shazbert:
New database system using an ORM management system SQLBoiler.
https://github.com/volatiletech/sqlboiler


thrasher-:
👍 nice work sir!

PR #71, opened 2017-12-28
initial support for all in orderbook for poloniex

gam-phon:
What do you think of the changes?
poloniex_wrapper.go is not working right now due to the changes in the struct. I will fix it later if you are willing to accept the feature.


thrasher-:
Thank you!

thrasher-:
As you stated, the last thing is to fix poloniex_wrapper.go, I can do this if you would like.

gam-phon:
@thrasher- Thanks, I would appreciate it

thrasher-:
Added in 5081a4b0c56354b1910d18c181cda66504bff2d7. Thanks!

PR #73, opened 2018-01-01
changes command to run application

mKurrels:
the readme was wrong when it said to run npm run web:start. It should be npm run start:web.


gloriousCode:
Good catch, a thanks!

PR #82, opened 2018-02-01
Added support for Binance exchange

shazbert:
Websocket not implemented yet. 
Authenticated requests will be added when needed.


thrasher-:
Fine work! 👍

PR #84, opened 2018-02-01
Add godep dependency manager and makefile

antonzhukov:
I propose to use go native dependency manager 'godep' plus using makefile to do all the job for you


thrasher-:
Great stuff, thanks!

PR #85, opened 2018-02-02
Added support for Yobit.net

askew-:


thrasher-:
Great stuff @askew-, thanks!

PR #87, opened 2018-02-08
Added support for Bitflyer exchange

shazbert:
Authenticated requests not done due to the fact I could not get an adequate account.


thrasher-:
Perfect, thanks! 👍

PR #88, opened 2018-02-13
Update Exchange Support Table.

daniel-cohen:
As discussed on slack, here is the updated exchange API info.

Supporting links
https://github.com/BTCMarkets/API/wiki/websocket
https://www.bitstamp.net/fix/

https://docs.gemini.com/websocket-api/
https://docs.gemini.com/fix-api/

https://hitbtc.com/fix

https://api.itbit.com/fix


thrasher-:
Great stuff, thanks!

PR #103, opened 2018-03-07
Type assertion when using gdax GetHistoricRates

cavapoo2:
Notice that sometimes gdax seems to return empty values when using GetHistoricRates. This caused 
```
s := History{
			Time:   int64(single[0].(float64)),
			Low:    single[1].(float64),
			High:   single[2].(float64),
			Open:   single[3].(float64),
			Close:  single[4].(float64),
			Volume: single[5].(float64),
		}*/
```
the type assertions to fail and then cause panic. 
the fix is not pretty here it is .
```
var s History
		a, _ := single[0].(float64)
		s.Time = int64(a)
		b, _ := single[1].(float64)
		s.Low = b
		c, _ := single[2].(float64)
		s.High = c
		d, _ := single[3].(float64)
		s.Open = d
		e, _ := single[4].(float64)
		s.Close = e
		f, _ := single[5].(float64)
		s.Volume = f
```
Now if the type assertion fails, then the value assigned will be zero of type float64 or int64. 
Let me know if you come up with better way than this.

### DIFF
--- a/exchanges/gdax/gdax.go
+++ b/exchanges/gdax/gdax.go
@@ -250,15 +250,6 @@ func (g *GDAX) GetHistoricRates(currencyPair string, start, end, granularity int
 		s.Close = e
 		f, _ := single[5].(float64)
 		s.Volume = f
-		/* type assertion poosible here, since sometimes gdax returns empty value
-		s := History{
-			Time:   int64(single[0].(float64)),
-			Low:    single[1].(float64),
-			High:   single[2].(float64),
-			Open:   single[3].(float64),
-			Close:  single[4].(float64),
-			Volume: single[5].(float64),
-		}*/
 		history = append(history, s)
 	}
 
--- a/exchanges/gdax/gdax.go
+++ b/exchanges/gdax/gdax.go
@@ -237,6 +237,20 @@ func (g *GDAX) GetHistoricRates(currencyPair string, start, end, granularity int
 	}
 
 	for _, single := range resp {
+		var s History
+		a, _ := single[0].(float64)
+		s.Time = int64(a)
+		b, _ := single[1].(float64)
+		s.Low = b
+		c, _ := single[2].(float64)
+		s.High = c
+		d, _ := single[3].(float64)
+		s.Open = d
+		e, _ := single[4].(float64)
+		s.Close = e
+		f, _ := single[5].(float64)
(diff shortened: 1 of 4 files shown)


cavapoo2:
By the way the other changes are related to the testing of exchange_test.go in root folder. You can reject these if you want, but should cause not problem. The idea with this was to remove the hard coded strings and instead use whatever exchanges are enabled in the testdata\configtest.json.

exchanges/gdax/gdax.go
```diff
@@ -237,14 +237,28 @@ func (g *GDAX) GetHistoricRates(currencyPair string, start, end, granularity int
...
+		c, _ := single[2].(float64)
+		s.High = c
+		d, _ := single[3].(float64)
+		s.Open = d
+		e, _ := single[4].(float64)
+		s.Close = e
+		f, _ := single[5].(float64)
+		s.Volume = f
+		/* type assertion poosible here, since sometimes gdax returns empty value
 		s := History{
```
shazbert: This code block can be removed before merge.

cavapoo2: shall i make another commit to remove the commented out code

cavapoo2:
just added the fix missing OkDefaultSet. It should compile now.

exchanges/gdax/gdax.go
shazbert: Yes please, thanks for all the good work!

cavapoo2:
don't like the look of these RACE Warnings. I'll check why this is happening. Maybe best to just get rid of these extra files that got submitted. I only wanted to submit the assertion bug, but since I'm new to this git, accidently added these test files.

cavapoo2:
The race occurs due to the the reads taking place in the Run() routine (running in its own go routine), for the vars Verbose and Websocket, and at same time writes to these same variables whenever ReloadExchange is called since in here it calls Setup which writes to Verbose and Websocket. One way to fix this would be to check in ReloadExchange to see if the exchange is already enabled. if it is just return without calling Setup again. ? unless you intended to have dynamic config ?

cavapoo2:
Can you delete this. The tests i added expose some data races. I'll create issue on that and maybe then i can add these tests at later date.

thrasher-:
Closed in favour of #104

PR #104, opened 2018-03-10
Fix for type assertion in Historic Rates in gdax.go

cavapoo2:
Hi, This time i just submit the fix (hopefully my git does not pull other files this time). Maybe you can delete my previous pull request since that seemed to open another can of worms with these data races. (I will create a issue on that)


thrasher-:
Thanks :)

PR #117, opened 2018-04-14
Fix test to use GDAX config from testdata

zeldrinn:


zeldrinn:
Odd, wasn't seeing this data race when running tests locally...

thrasher-:
Thanks! Those can be ignored as a fix will be published shortly.

PR #123, opened 2018-05-04
Added new package provider for foreign exchange rates.

shazbert:
- Contains segregated provider packages
- Waiting for review before full integration


config/config.go
```diff
@@ -99,6 +99,7 @@ type Config struct {
 	SMS                      SMSGlobalConfig  `json:"SMSGlobal"`
 	Webserver                WebserverConfig  `json:"Webserver"`
 	Exchanges                []ExchangeConfig `json:"Exchanges"`
+	Providers                []ProviderConfig `json:"Providers"`
```
thrasher-: Need to come up with a more specific name, perhaps ForexProviders or ForexConversionProviders.

config/config.go
```diff
@@ -230,6 +241,20 @@ func (c *Config) GetExchangeConfig(name string) (ExchangeConfig, error) {
 	return ExchangeConfig{}, fmt.Errorf(ErrExchangeNotFound, name)
 }
 
+// GetProviderConfig returns a provider configuration by its name
+func (c *Config) GetProviderConfig(name string) (ProviderConfig, error) {
+	c.m.Lock()
+	defer c.m.Unlock()
+	log.Println("What", len(c.Providers))
+	for i := range c.Providers {
+		log.Println(c.Providers[i].Name)
```
thrasher-: Just need to make sure this is removed upon merge.

config/config.go
```diff
@@ -341,6 +366,20 @@ func (c *Config) CheckWebserverConfigValues() error {
 	return nil
 }
 
+// CheckProviderConfigValues checks Provider configuration values
+func (c *Config) CheckProviderConfigValues() error {
```
thrasher-: A check here for default values would be good. Fatal error if the default values are set without the users ones unless the forex provider allows forex conversions without an API key.

config/config.go
shazbert: ForexRateProviders?

config/config.go
```diff
@@ -233,6 +245,20 @@ func (c *Config) GetExchangeConfig(name string) (ExchangeConfig, error) {
 	return ExchangeConfig{}, fmt.Errorf(ErrExchangeNotFound, name)
 }
 
+// GetProviderConfig returns a forex provider configuration by its name
+func (c *Config) GetProviderConfig(name string) (ForexProviderConfig, error) {
+	c.m.Lock()
+	defer c.m.Unlock()
+
+	for i := range c.ForexProviders {
+		log.Println(c.ForexProviders[i].Name)
```
thrasher-: Can remove this Println

config/config.go
```diff
@@ -97,10 +97,11 @@ type Config struct {
...
 	GlobalHTTPTimeout        time.Duration
-	Portfolio                portfolio.Base   `json:"PortfolioAddresses"`
-	SMS                      SMSGlobalConfig  `json:"SMSGlobal"`
-	Webserver                WebserverConfig  `json:"Webserver"`
-	Exchanges                []ExchangeConfig `json:"Exchanges"`
+	Portfolio                portfolio.Base        `json:"PortfolioAddresses"`
+	SMS                      SMSGlobalConfig       `json:"SMSGlobal"`
+	Webserver                WebserverConfig       `json:"Webserver"`
+	Exchanges                []ExchangeConfig      `json:"Exchanges"`
+	ForexProviders           []ForexProviderConfig `json:"Providers"`
```
thrasher-: Should make the Providers JSON ouput match the ForexProviders var

config/config.go
```diff
@@ -349,6 +375,23 @@ func (c *Config) CheckWebserverConfigValues() error {
...
+	for i := range c.ForexProviders {
+		if c.ForexProviders[i].Enabled == true {
+			if c.ForexProviders[i].APIKey == "Key" {
+				log.Fatal("provider api key not set - please set via your config.json file")
+			}
+			count++
+		}
+	}
+	if count < 1 {
+		log.Println("warning config.go CheckProviderConfigValues() no providers set")
```
thrasher-: Moving forward we will need to fatal error if there's no forex providers set (After the free unauthenticated fixer.io service is stopped).

shazbert: Added to trello.

thrasher-:
Squashed and placed in https://github.com/thrasher-/gocryptotrader/commit/58051b89c71737daa47fde09f968c65762977021

PR #126, opened 2018-05-21
Added new package communications

shazbert:
Initial commit for:
* Slack support
* SMSGlobal support
* SMTP support
* Telegram support


communications/telegram/telegram.go
```diff
@@ -0,0 +1,230 @@
...
+func (t *Telegram) HandleMessages(text string, chatID int64) error {
+	switch {
+	case common.StringContains(text, cmdHelp):
+		return t.SendMessage(fmt.Sprintf("%s: %s", talkRoot, cmdHelpReply), chatID)
+
+	case common.StringContains(text, cmdStart):
+		return t.SendMessage(fmt.Sprintf("%s: START COMMANDS HERE", talkRoot), chatID)
+
+	case common.StringContains(text, cmdOrders):
+		return t.SendMessage(fmt.Sprintf("%s: %s", talkRoot, t.GetOrderbook("ANX")), chatID)
```
thrasher-: Need to check to see if the supplied exchange is valid, and if so, get the orderbook for the desired exchange. Will also need to take in a param for a currency pair.

PR #128, opened 2018-05-30
Database

shazbert:


common/common.go
```diff
@@ -249,6 +249,15 @@ func RoundFloat(x float64, prec int) float64 {
 	return rounder / pow
 }
 
+// CountInt counts how many decimals in an integer and returns the count
```
gloriousCode: Do you mean to be counting digits? Or does golang int support decimal places?

main.go
```diff
@@ -26,29 +27,45 @@ type Bot struct {
...
 	configFile string
 }
 
-const banner = `
-   ______        ______                     __        ______                  __           
+const (
+	dbname     = "gocryptotrader"
+	dbhost     = "localhost"
+	dbuser     = "gocryptotrader"
+	dbpassword = "gocryptotrader"
```
gloriousCode: Password in readme says should be setup to be lol123

common/common.go
shazbert: Yes I do, please request change.

main.go
shazbert: Will change that, cheers.

ermalguni:
Any update on this?

alexmironof:
I advice to use embeded database like https://github.com/coreos/bbolt with wrapper like https://github.com/asdine/storm. Its battle tested, zero configuration and low overhead database

shazbert:
@ermalguni early next week;  shelling out a feature built on top of this and then I will push the new commits.
@alexmironof Hi, Thanks for the suggestion, that's a great idea for a stand alone product but the main idea was that we are going to be using the database as a microservice. If you would like to start a new branch based off master then we can give you a hand in implementing those features, and keep it updated as new features come available. If you are on our slack just pm me.

ermalguni:
@shazbert thanks for the reply. Is the DB feature supposed to store also the ticker/book/trade information retrieved from the respective exchanges?

shazbert:
@ermalguni 
The database will store exchange specific trade history; there will be a routine that might append to that list or the engine will periodically keep the trade history up to date. This will be used as a back testing feature. Ticker will not be put into the database that will be kept in memory and how much we will allocate as sufficient space will be determined. Orderbook information will not be stored in the database as this will only be updated for algorithmic orderbook processing when you are about to initiate an exchange order. These features will all be periodically updated when we work on the trading engine. The design specifics on the trade engine are not figured out yet. 

What else will be stored in the database for phase one, I regard phase one as a very loose term but there is a limited outline of what we want to accomplish ranging from, differing configurations, portfolios and order history; order history will be supplemented by taxable event management package so its easier for EOFY. 

I hope this gives you a better general outline of what to expect in the future, if you want to contribute, have any ideas, let us know.

ermalguni:
@shazbert thanks for the high level explanation. It was very helpful. Nonetheless, I think that storing ticker/book/trade data in the DB can also be helpful with the backtesting of different trading algos. I would be willing to help in that regard and will look into that after you have merged this PR.

shazbert:
@ermalguni That's a fantastic idea, your help would be appreciated!

thrasher-:
Needs rebase + don't run database tests if postgres isn't setup/configured.

mia0x75:
Why not mysql? Since it is more popular.

shazbert:
@mia0x75 Converting over the SQLite3 at the moment so there is an easier instance of a DB. If you need a specific DB we can always branch it and keep it updated.

shazbert:
Closed for rebase on new system

PR #133, opened 2018-06-09
bitfinex websocket trade fix

ermalguni:
According to: https://docs.bitfinex.com/v1/reference#ws-public-trades bitfinex sends a SEQ which is a string and is a trade id. Later it sends the real ID, so we can discard it the first time we receive it. 

Furthermore, bitfinex sends each trade in two messages: "te" and "tu". The "tu" messages contains the real trade id.


thrasher-:
Thanks!

PR #135, opened 2018-06-11
Fixed bug in Binance GetOrderbook.

shazbert:


shazbert:
@andreygrehov Thanks for spotting this!

PR #141, opened 2018-06-16
gofmt project

whilei:
Just ran

	gofmt -w .

on the project root. That's all.

https://blog.golang.org/go-fmt-your-code

---
> I made this PR with a project going on over at https://github.com/rotblauer/gofmt-att, and it's definitely a work in progress. So if I got something wrong, or this is annoying at all, please file an issue over there and we'll sort it out. On the other hand, if you think it's kind of funny like I do, please feel welcome to contribute or even fire up one of your own great automatic _gofmt_ers.

### DIFF
--- a/routines.go
+++ b/routines.go
@@ -6,8 +6,8 @@ import (
 	"sync"
 	"time"
 
-	"github.com/thrasher-/gocryptotrader/currency/pair"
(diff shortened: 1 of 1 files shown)


thrasher-:
Thanks!

PR #142, opened 2018-06-21
Trading functions

shazbert:


config/config.go
```diff
@@ -125,6 +125,19 @@ type ExchangeConfig struct {
...
+}
+
+// BankConfig is a single bank configuration
+type BankConfig struct {
+	Enabled       bool
+	BankName      string
+	AccountName   string
+	AccountNumber string
+	BSBNumber     string
+	SwiftCode     string
```
thrasher-: An IBAN number would be great to include as it is just one number used for cross boarder payments which includes the country, swift and account number.

exchanges/alphapoint/alphapoint_wrapper.go
```diff
@@ -98,24 +105,28 @@ func (a *Alphapoint) GetExchangeHistory(p pair.CurrencyPair, assetType string) (
 
 // SubmitExchangeOrder submits a new order and returns a true value when
 // successfully submitted
-func (a *Alphapoint) SubmitExchangeOrder(p pair.CurrencyPair, side string, orderType int, amount, price float64) (int64, error) {
-	return a.CreateOrder(p.Pair().String(), side, orderType, amount, price)
+func (a *Alphapoint) SubmitExchangeOrder(base, quote pair.CurrencyItem, side exchange.OrderSide, orderType exchange.OrderType, amount, price float64, clientID string) (int64, error) {
```
thrasher-: As discussed, it would be better to use a pair.CurrencyPair variable just like we do with the ticker and orderbook, that way we can use the exchange.FormatExchangeCurrency to line up with the exchanges expected request and would save time not having to adjust it to match the exchange.

config_example.json
```diff
@@ -147,6 +147,33 @@
    },
    "RequestCurrencyPairFormat": {
     "Uppercase": true
+   },
+   "InternationalBank": {
```
gloriousCode: Are banks supported by the exhcange's API? Or is this for GCT? If its for GCT, having an array of banks of BankConfig type would then allow for International, Local, Deposit to be properties and to have relationships to multiple exchanges. It would also cut down on config size, especially if a user wants to use one international bank for all exchanges for example.

shazbert: You make a good point, I wanted it like this because exchange configurations are going to be abstracted out to their respective exchange packages, and its easier for the user to enter it as is but can be changed @thrasher-  if your suggestion is a better approach. As for bank supported by exchange, they mostly are. So in the exchange wrapper functions you can pass in withdrawal details for either a local bank or an international bank, I would assume, not all cases, that the local bank will be done at a faster rate and an international bank will be slower but more accessible globally. The deposit bank would be the actual bank account associated with the exchange themselves for depositing purposes, so this will be used by a bank API to fund exchanges.

thrasher-: We'll need two properties:

An exchange will require a bank account associated with it in the following structure:

AccountName
AccountNumber
BankName
BankAddress
SWIFTCode 
IBAN
BSBNumber (omitempty for non Australian domestic transfers)

For SEPA tranfers, the IBAN code is used for the transfer. For International transfers either IBAN or SWIFT can be used in conjunction with the other information so there's no need for an entry for local, deposit or international bank entries.

We'll also need a config for the users own bank and stored as an array in case the user wishes to user multiple bank accounts.

thrasher-:
Superseded by https://github.com/thrasher-/gocryptotrader/commit/ed675bde3078c6c64f80b2044726c6461a1606f7

PR #144, opened 2018-06-27
Fixed issues with liqui GetTicker function. Added tests. Increased Un…

shazbert:
…auth request limit.

### DIFF
--- a/exchanges/liqui/liqui.go
+++ b/exchanges/liqui/liqui.go
@@ -34,7 +34,7 @@ const (
 	liquiWithdrawCoin      = "WithdrawCoin"
 
 	liquiAuthRate   = 0
-	liquiUnauthRate = 0
+	liquiUnauthRate = 1
 )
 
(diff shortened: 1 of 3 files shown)


exchanges/liqui/liqui.go
```diff
@@ -34,7 +34,7 @@ const (
 	liquiWithdrawCoin      = "WithdrawCoin"
 
 	liquiAuthRate   = 0
-	liquiUnauthRate = 0
+	liquiUnauthRate = 1
```
gloriousCode: Is liqui prone to authentication errors?

shazbert: Unauthenticated rate limits were being exceeded, I just bumped it up to one. When we have the trading engine online and processing we can implement an auto adjustment function for the rates.

gloriousCode: Thanks for the clarification!

PR #148, opened 2018-07-09
Fixed some typos in binance types and updated binance trading symbols in config example

marcofranssen:


CONTRIBUTORS
```diff
@@ -22,4 +22,4 @@ tongxiaofeng | https://github.com/tongxiaofeng
 idealhack | https://github.com/idealhack
 askew- | https://github.com/askew-
 snipesjr | https://github.com/snipesjr
-
+marcofranssen | https://github.com/marcofranssen
```
thrasher-: FYI, we have a tool to automatically generate this (can be located here: https://github.com/thrasher-/gocryptotrader/tree/master/tools/documentation), will update the contributors list in a separate commit.

PR #153, opened 2018-07-17
Removed package-lock.json from gitignore

marcofranssen:
Using node npm package-lock.json is recommended to commit in the repo. It ensures a specific package version is installed so an explicit update has to be done and committed.

package-lock.json was already committed though the file was still mentioned in .gitignore.

Furthermore removed an unneeded whiteline.


marcofranssen:
Will file another PR with more updates.

PR #163, opened 2018-08-02
Lowercase json to be more compliant with javascript coding standards

marcofranssen:
To have all the json start with a lowercase I added the annotations on the models. This way the json responses on rest api are more aligned with REST Standards + the fact someone can use the json objects in javascript in such way the json is still compliant with javascript coding standards.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -33,8 +33,8 @@ const (
 	configMaxAuthFailres                   = 3
 )
 
-// Variables here are mainly alerts and a configuration object
-var (
+// Constants here hold some messages
+const (
 	ErrExchangeNameEmpty                            = "Exchange #%d in config: Exchange name is empty."
 	ErrExchangeAvailablePairsEmpty                  = "Exchange %s: Available pairs is empty."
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -54,34 +54,38 @@ var (
...
 
 // Post holds the bot configuration data
 type Post struct {
-	Data Config `json:"Data"`
+	Data Config `json:"data"`
 }
 
 // CurrencyPairFormatConfig stores the users preferred currency pair display
 type CurrencyPairFormatConfig struct {
 	Uppercase bool
```
thrasher-: This can be made lowercase too

config/config.go
```diff
@@ -90,130 +94,130 @@ type CurrencyPairFormatConfig struct {
 type Config struct {
 	Name              string
```
thrasher-: These two settings (Name and EncryptConfig) can be made lowercase.

marcofranssen: good catch... missed these indeed

marcofranssen:
Updated config_example and resolved conflicts.

PR #172, opened 2018-08-10
Update README.md

basantashah:
Let people send btc to me :package:


shazbert:
@basantashah if only you left that PR open. ;)

PR #174, opened 2018-08-16
Fix build errors, warnings, vulnerabilities, linting for front-end

gloriousCode:
Updates things to later versions and fixes lingering issues that I've ignored for awhile

Vulnerabilities:
image

Build-Windows:
image

Lint:
image

Running application with no errors, console open:
image


web/src/polyfills.ts
```diff
@@ -41,8 +41,8 @@
 
 
 /** Evergreen browsers require these. **/
-import 'core-js/es6/reflect';
-import 'core-js/es7/reflect';
+// import 'core-js/es6/reflect';
```
thrasher-: Can these be removed?

gloriousCode: It would be hypocritical otherwise 😄 
I'll remove it now

PR #176, opened 2018-08-19
Decimal implementation`

kempeng:


shazbert:
@kempeng This is great, there are a couple of changes that I think need to be done before merge.

1. We can wrap over decimal dep with a new decimal package in common with helper functions and values; this will help readability.
2. Remove comments that are not needed 
3. Vendor in dependency
4. Add package to document generation tool

All of this I can do, thanks again for opening this PR

PR #192, opened 2018-10-12
Replace time.Sleep with time.Tick

mia0x75:
time.Tick is precise.

### DIFF
--- a/routines.go
+++ b/routines.go
@@ -179,7 +179,8 @@ func relayWebsocketEvent(result interface{}, event, assetType, exchangeName stri
 func TickerUpdaterRoutine() {
 	log.Println("Starting ticker updater routine.")
 	var wg sync.WaitGroup
-	for {
+	d := time.Duration(10) * time.Second
+	for range time.Tick(d) {
 		wg.Add(len(bot.exchanges))
 		for x := range bot.exchanges {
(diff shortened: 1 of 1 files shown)


shazbert:
LGTM,

@mia0x75  Thanks for opening this PR, it is very welcome.

If you checkout my PR in routines.go on the database branch: https://github.com/thrasher-/gocryptotrader/pull/128/commits/cb8439a340fe1ccee5ff145b5bbe0aae66ca8ab4 , Only difference is, I have had to monitor each individual exchange/currency pair/asset class because if there was request issue with one pair it would then slow down the update on other pairs. Let me know what you think. I am going to squash and rebase soon so we can get it merged after I finish off testing on the websocket PR. 

You also wanted MySQL on the database branch are you still wanting that as a feature request? And are you on Slack?

@thrasher- utACK

mia0x75:
I just checkout database branch, it looks good, hope it could be merged into master branch soon. And yes, I think it will a great feature if mysql is supported.

I don't have Slack right now, will get an account later ;-)

shazbert:
@mia0x75 It will be great to collaborate, once you have a slack account we can chat more there. Unfortunately we cannot merge this as I have been advised this will cause burst behaviour in fetching. We are going to leave as is for now. 
Still thanks for opening the PR. Chat soon.

PR #195, opened 2018-10-22
Telegram testing and race condition fix.

cranktakular:
Improving telegram Setup test
More coverage for telegram PushEvent test
Adding telegram HandleMessages test
Adding telegram GetUpdates test
Adding telegram TestConnection test
Adding telegram SendMessage test
(Hopefully) completely fixed race conditions with slack testing

### DIFF
--- a/communications/communications_test.go
+++ b/communications/communications_test.go
@@ -1,0 +1,28 @@
+package communications
+
+import (
+	"testing"
+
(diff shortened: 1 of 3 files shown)


thrasher-:
Thanks @cranktakular! A future PR can address the  successful HandleMessages test scenarios, will merge this as is so the race condition is fixed.

PR #199, opened 2018-10-31
Exchange withdraw capabilities

gloriousCode:
This PR adds a property to the base exchange struct type to determine what types of withdrawal methods an exchange supports; fiat and/or crypto, via api or website

Also adds account information API endpoint to ANX. Don't need it anymore, but can be used to determine API permissions and other things later anyway.

### DIFF
--- a/config_example.json
+++ b/config_example.json
@@ -135,6 +135,8 @@
    "httpTimeout": 15000000000,
    "httpUserAgent": "",
    "authenticatedApiSupport": false,
+   "apiWithdrawalSupport": true,
+   "automaticApiWithdrawlSupport": false,
    "apiKey": "Key",
    "apiSecret": "Secret",
    "apiUrl": "NON_DEFAULT_HTTP_LINK_TO_EXCHANGE_API",
--- a/config_example.json
+++ b/config_example.json
@@ -135,8 +135,6 @@
    "httpTimeout": 15000000000,
    "httpUserAgent": "",
    "authenticatedApiSupport": false,
-   "apiWithdrawalSupport": true,
-   "automaticApiWithdrawlSupport": false,
    "apiKey": "Key",
    "apiSecret": "Secret",
    "apiUrl": "NON_DEFAULT_HTTP_LINK_TO_EXCHANGE_API",
--- a/exchanges/alphapoint/alphapoint_wrapper.go
+++ b/exchanges/alphapoint/alphapoint_wrapper.go
@@ -181,3 +181,8 @@ func (a *Alphapoint) GetWebsocket() (*exchange.Websocket, error) {
 func (a *Alphapoint) GetFeeByType(feeBuilder exchange.FeeBuilder) (float64, error) {
 	return 0, errors.New("not yet implemented")
 }
+
+// GetWithdrawCapabilities returns the types of withdrawal methods permitted by the exchange
+func (a *Alphapoint) GetWithdrawCapabilities() string {
+	return a.GetWithdrawPermissions()
+}
--- a/exchanges/alphapoint/alphapoint_wrapper.go
+++ b/exchanges/alphapoint/alphapoint_wrapper.go
@@ -183,6 +183,6 @@ func (a *Alphapoint) GetFeeByType(feeBuilder exchange.FeeBuilder) (float64, erro
 }
 
 // GetWithdrawCapabilities returns the types of withdrawal methods permitted by the exchange
-func (a *Alphapoint) GetWithdrawCapabilities() string {
+func (a *Alphapoint) GetWithdrawCapabilities() uint32 {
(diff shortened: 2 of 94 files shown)


exchanges/anx/anx.go
```diff
@@ -443,3 +446,38 @@ func getInternationalBankWithdrawalFee(currency string, amount float64) float64
...
+// CheckAPIWithdrawPermission checks if the API key is allowed to withdraw
+func (a *ANX) CheckAPIWithdrawPermission() (bool, error) {
+	accountInfo, err := a.GetAccountInformation()
+	if err != nil {
+		return false, err
+	}
+	var apiAllowsWithdraw bool
+	for _, a := range accountInfo.Rights {
+		if a == "withdraw" {
+			log.Printf("API key is missing withdrawal permissions")
```
shazbert: Change: Wording does not reflect what the code shows.

exchanges/exchange.go
```diff
@@ -766,3 +812,76 @@ func (e *Base) GetAPIURLDefault() string {
...
+// GetWithdrawPermissions will return each of the exchange's compatible withdrawal methods
+func (e *Base) GetWithdrawPermissions() string {
+	services := []string{}
+	for i := 0; i < 32; i++ {
+		var check uint32 = 1 << uint32(i)
+		if e.APIWithdrawPermissions&check != 0 {
+			switch check {
+			case AutoWithdrawCrypto:
+				services = append(services, AutoWithdrawCryptoText)
+				break
```
shazbert: I think these breaks might be unnecessary. Test code functionality without them. They should just break out of the switch at end of case anyway. Or I need another coffee.

gloriousCode: I was intending to break the for loop. From reading more, break will _break_ the most immediate breakable thing, which in this case is the switch statement. I'll rewrite it to get out of the loop properly

config_example.json
```diff
@@ -135,6 +135,8 @@
    "httpTimeout": 15000000000,
    "httpUserAgent": "",
    "authenticatedApiSupport": false,
+   "apiWithdrawalSupport": true,
```
thrasher-: This seems non-needed as the withdraw permissions are set in the exchanges SetDefaults() without any config handling. In the future it would be nice to open the ability to adjust via config.

exchanges/exchange.go
```diff
@@ -773,3 +819,58 @@ func (e *Base) GetAPIURLDefault() string {
 func (e *Base) GetAPIURLSecondaryDefault() string {
 	return e.APIUrlSecondaryDefault
 }
+
+// GetWithdrawPermissions will return each of the exchange's compatible withdrawal methods
+func (e *Base) GetWithdrawPermissions() string {
```
thrasher-: The string return is useful for humans but for other logic we would require returning the bitmask directly without the need for string parsing. This means we can just check withdraw permissions like this: 

```
wdPerms := exch.GetWithdrawPermissions(); 
if (wdPerms & AutoWithdrawCrypto) { 
    // exchange supports automatic withdrawal of crypto
}
```

Another useful function would be a function which returns a bool based on the supplied bitmask:

```
wdPerms := AutoWithdrawFiat | AutoWithdrawCrypto
if exch.SupportsWithdrawPermissions(wdPerms) {
      // supports auto withdraw of fiat and crypto
}
```

Lastly we'd need an amendment to the exchange interface so we can access these functions from any application using it outside of individual package usage.

exchanges/alphapoint/alphapoint_wrapper.go
```diff
@@ -181,3 +181,8 @@ func (a *Alphapoint) GetWebsocket() (*exchange.Websocket, error) {
 func (a *Alphapoint) GetFeeByType(feeBuilder exchange.FeeBuilder) (float64, error) {
 	return 0, errors.New("not yet implemented")
 }
+
+// GetWithdrawCapabilities returns the types of withdrawal methods permitted by the exchange
+func (a *Alphapoint) GetWithdrawCapabilities() string {
```
thrasher-: GetWithdrawCapabilities is mentioned but needs to be added to the exchange interface

exchanges/exchange.go
```diff
@@ -83,6 +83,51 @@ type FeeBuilder struct {
 	Amount        float64
 }
 
+// Definitions for each type of withdrawal method for a given exchange
+const (
+	// No withdraw
+	NoAPIWithdrawalMethods                  uint32 = 0
```
thrasher-: Future refinement could use a type for these with String() or Bitmask() funcs

config_example.json
gloriousCode: Whoops, old code >_>

thrasher-:
👍 Nice work @gloriousCode!

PR #203, opened 2018-11-02
Requester update

shazbert:
## Type of change

# How Has This Been Tested?

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration

Unit test: "https://httpstat.us/200?sleep=20000" end point to sleep before reply

# Checklist:

### DIFF
--- a/exchanges/request/request.go
+++ b/exchanges/request/request.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"log"
+	"net"
 	"net/http"
 	"net/url"
 	"sync"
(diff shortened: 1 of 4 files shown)


exchanges/request/request.go
```diff
@@ -247,36 +248,50 @@ func (r *Requester) DoRequest(req *http.Request, method, path string, headers ma
 		log.Printf("%s exchange request path: %s requires rate limiter: %v", r.Name, path, r.RequiresRateLimiter())
 	}
 
-	resp, err := r.HTTPClient.Do(req)
+	var timeoutError error
+	for i := 0; i < 4; i++ { // 3 attempts for extra requests due to client timeout
```
thrasher-: Create a request var called DefaultTimeoutRetryAttempts and set that to 3, in the for loop change it to i < DefaultTimeoutRetryAttempts +1

Engine refactor will include code to make this overridable and optional based on user preferences.

shazbert: Okay will do

gloriousCode: Could we go with `i <= X` rather than `i < (X +1)` ?

thrasher-: Destinction of retry attempts val and primary request is better IMO.

shazbert: Yeah if we pass in 0 then it won't do a request.

PR #252, opened 2019-03-12
OKEX/OKCoin API combine via OKGroup

gloriousCode:
# Description
The OKCoin v3 API shares all endpoints with the v3 OKEX API. OKEX has additional endpoints pertaining to Swap, Futures and ETT. This PR is the implementation of both APIs sharing a common implementation in okgroup.

# Things of note:
- okgroup contains the shared endpoints between okcoin and okex. As a result, each file will look different to regular exchange files since they will be missing certain functions that are implemented in another area. Let me know if I need to comment this more for any future work readibility.

- I've made this api have request and response structs for each endpoint. The goal here was to minimise multiple arguments as well as any logic within the API funcs.

- I've added a new go mod. `github.com/google/go-querystring v1.0.0`. This works very similarly to how golang can handle structs with json tags. Take the following:
```
// GetSpotOrdersRequest request data for GetSpotOrders
type GetSpotOrdersRequest struct {
	Status string `url:"status"`// list the status of all orders (all, open, part_filled, canceling, filled, cancelled，ordering,failure)
	// （Multiple status separated by '|'，and '|' need encode to ' %7C'）
	InstrumentID string `url:"instrument_id"`          // trading pair ,information of all trading pair will be returned if the field is left blank
	From         int64  `url:"from,string,omitempty"`  // [optional] request page after this id (latest information) (eg. 1, 2, 3, 4, 5. There is only a 5 "from 4", while there are 1, 2, 3 "to 4")
	To           int64  `url:"to,string,omitempty"`    // [optional] request page after (older) this id.
	Limit        int64  `url:"limit,string,omitempty"` // [optional] number of results per request. Maximum 100. (default 100)
}
```

Once the request is sent to the function, query.Values(request) is declared and query.Encode() will HTML encode to URL parameters eg "?status=1&instrument_id=2&from=3&to=4&limit=100". No need to declare url.Values and then add each value individually, or convert from number to string yourself.

- Response variable declarations. Go supports declaration of response variables in the function declaration. I've utilised this to minimise lines.

- Function struct pointers. I've used pointers in some methods where structs are passed around to minimise large struct duplication (golang will duplicate a struct when sent to a function). I can remove these, or I can spread it to all the functions. Not sure if it impacts code readability too much.

- Really minor, but I've renamed the ws func `SetEnabled` to `SetWsStatusAndConnection` since SetEnabled can also connect to the endpoint

## Type of change

# How Has This Been Tested?

There are over 300 tests written. Every single REST endpoint has a test. Websocket tests are more difficult due to the nature of a message response not being something I can directly listen to and verify. If you have any suggestions, I'd be happy to implement them. I'd prefer websocket coverage. I have tested the endpoints though and verified via logs. Finally, there are wrapper tests and a few logic tests.

These have been verified with and without API credentials.

# Checklist:


exchanges/okgroup/okgroup_websocket.go
```diff
@@ -0,0 +1,715 @@
...
+		proxy, err := url.Parse(o.Websocket.GetProxyAddress())
+		if err != nil {
+			return err
+		}
+
+		dialer.Proxy = http.ProxyURL(proxy)
+	}
+
+	var err error
+	log.Printf("Attempting to connect to %v", o.Websocket.GetWebsocketURL())
```
shazbert: wrap this in a 
```go 
if o.Verbose {}
```
and change Printf -> Debugf

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -0,0 +1,715 @@
...
+	o.WebsocketConn, _, err = dialer.Dial(o.Websocket.GetWebsocketURL(),
+		http.Header{})
+	if err != nil {
+		return fmt.Errorf("%s Unable to connect to Websocket. Error: %s",
+			o.Name,
+			err)
+	}
+	log.Printf("Successful connection to %v", o.Websocket.GetWebsocketURL())
```
shazbert: Same as above.

exchanges/okex/okex.go
```diff
@@ -1,110 +1,59 @@
...
-
-	myWalletInfo = "wallet_info.do"
-
-	// just your average return type from okex
-	returnTypeOne = "map[string]interface {}"
-
-	okexAuthRate   = 0
-	okexUnauthRate = 0
+	okExAuthRate     = 0
+	okExUnauthRate   = 0
```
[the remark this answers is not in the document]
gloriousCode: Forgot to add them. Thanks for that

exchanges/okex/okex_test.go
```diff
@@ -1,300 +1,1587 @@
...
+
+func areTestAPIKeysSet() bool {
+	if o.APIKey != "" && o.APIKey != "Key" &&
+		o.APISecret != "" && o.APISecret != "Secret" {
+		return true
+	}
+	return false
+}
+
+func testStandardErrorHandling(t *testing.T, err error) {
```
xtda: I wonder if it would be better to have this return the error to allow for further checking 
That way we don't end up with examples like https://github.com/thrasher-/gocryptotrader/pull/252/files#diff-1405017cb73c41ce881d3e670589ee58R124 where areTestAPIKeysSet() is being double checked

gloriousCode: Yeah definitely do not want to double check things! I'll go over the cases. I think any test that requires more than whats in `testStandardErrorHandling` shouldn't use it and have its own error handling rather than returning the error. Thanks for picking that up

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -0,0 +1,719 @@
...
+			log.Debugf("%v Websocket orderbook data received", o.GetName())
+		}
+		// Locking, orderbooks cannot be processed out of order
+		orderbookMutex.Lock()
+		err := o.WsProcessOrderBook(response)
+		if err != nil {
+			log.Error(err)
+			subscriptionChannel := fmt.Sprintf("%v:%v", response.Table, response.Data[0].InstrumentID)
+			o.WsUnsubscribeToChannel(subscriptionChannel)
+			o.WsSubscribeToChannel(subscriptionChannel)
```
xtda: There may be times here where if WsProcessOrderbook returns an error and then WsUnsubscribeToChannel or WsSubscribeToChannel also error that the Websocket channel subscription is never re-established
Both of these do return errors that may be work handling and reporting on

gloriousCode: That's true. I'll handle the errors, but if it fails, even after x attempts to unsubscribe, resubscribe, what do you think should be done? Another log.Error highlighting the failure?

xtda: It comes down to:

Can this exchange still operate and trade without issues if this happens?
If not can the exchange be disabled without impacting anything further in the trade engine?

If it can't and will impact the trading these are times where even a fatal log might be needed

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -0,0 +1,759 @@
...
+			if err != nil {
+				log.Error(err)
+				time.Sleep(okGroupWsResubscribeDelayInSeconds * time.Second)
+				continue
+			}
+			successfulUnsubscribe = true
+			break
+		}
+		if !successfulUnsubscribe {
+			log.Fatalf("%v websocket channel %v failed to unsubscribe after %v attempts", o.GetName(), channel, okGroupWsResubscribeFailureLimit)
```
thrasher-: This will be improved in a seperate PR which introduces a websocket client which keeps track of all subscribed/unsubcribed channels as well as initialising an exchanges WS DataHandler but for now this will suffice.

PR #255, opened 2019-03-18
GateIO: Fixed OrderHistory unmarshalling

vadimzhukck:
# Description

Fixed OrderHistory unmarshalling on the GateIo.


exchanges/gateio/gateio_wrapper.go
```diff
@@ -375,7 +375,7 @@ func (g *Gateio) GetOrderHistory(getOrdersRequest exchange.GetOrdersRequest) ([]
 		side := exchange.OrderSide(strings.ToUpper(trade.Type))
 		orderDate := time.Unix(trade.TimeUnix, 0)
 		orders = append(orders, exchange.OrderDetail{
-			ID:           trade.OrderID,
+			ID:           string(trade.OrderID),
```
thrasher-: It's better to use strconv.FormatInt(trade.OrderID, 10) for type conversion due to errors doing it this way; can see an example here: https://play.golang.org/p/d10xE_Wxrlu

vadimzhukck: Fixed

PR #256, opened 2019-03-19
Added database support - postgres && sqlite3 as backup

shazbert:
# Description

Adds database support using an ORM SQLBoiler, SQLite 3 basic out of the box support and postgres for people who actually want to database. 

Fixes # NO DATABASE FOR BACK TESTING 

## Type of change

# How Has This Been Tested?

go test ./... and actually put data into database on local machine.

# Checklist:


routines.go
```diff
@@ -470,3 +499,165 @@ func WebsocketReconnect(ws *exchange.Websocket, verbose bool) {
...
+				}
+			}(x, &wg)
+		}
+		wg.Wait()
+		log.Debugln("All enabled currency platform trades fetched.")
+		time.Sleep(time.Minute * 5)
+	}
+}
+
+func fetchHistory(exch exchange.IBotExchange, p currency.Pair, assetType string, timeStart, timeEnd time.Time) {
```
xtda: As we discussed breifly yesteray the timeStart and timeEnd usage in fetchHistory is incorrect because GetPlatformHistory for most exchanges will only accept a start time and then each exchange wrapper sets up these times depending on the exchange limits (for example Binance is 3 months)

Also on lines 660-661 you are just printing time.Time{} which will cause the output to always be 1970-01-01 10:00:00

routines.go
```diff
@@ -470,3 +499,165 @@ func WebsocketReconnect(ws *exchange.Websocket, verbose bool) {
...
+				err)
+			return
+		}
+	}
+
+	printPlatformTradeSummary(&h,
+		exch.GetName(),
+		assetType,
+		p,
+		time.Time{},
```
xtda: This will create a new instance of time.Time which defaults to 01/01/1970 and not print the user selected time

routines.go
```diff
@@ -470,3 +499,165 @@ func WebsocketReconnect(ws *exchange.Websocket, verbose bool) {
...
+		time.Time{},
```
xtda: This will create a new instance of time.Time which defaults to 01/01/1970 and not print the user selected time

xtda:
Is it worth adding the option to disable the database completely? at the moment its either sqlite or postgres some users might not any

routines.go
shazbert: Will fix!

common/common.go
```diff
@@ -629,3 +647,87 @@ func CheckDir(dir string, create bool) error {
...
 	return nil
 }
+
+// PromptForUsername prompts for username stored in database and returns a
+// string from stdin
+func PromptForUsername() (string, error) {
+	log.Info("Please enter username to connect to the database:")
+	var username string
+	for {
+		i, err := fmt.Scanln(&username)
```
xtda: Scanln usage here will error if a user enters a blank username (as its considered a new line \n) or a username that enters with a space if not properly stripped

common/common.go
```diff
@@ -629,3 +647,87 @@ func CheckDir(dir string, create bool) error {
...
+
+// PromptForPassword prompts for password associated with username and returns a
+// byte array from stdin for comparison
+func PromptForPassword(newPassword bool) ([]byte, error) {
+	log.Info("Please enter password, then press enter to continue:")
+
+	var password, match []byte
+	var err error
+	for {
+		password, err = terminal.ReadPassword(int(syscall.Stdin)) // nolint:unconvert
```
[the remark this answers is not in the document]
shazbert: Good idea, thanks!

common/common.go
shazbert: What do you suggest?

shazbert:
closed due to rebasing on engine branch

PR #259, opened 2019-03-22
Binance: fix issue on order submit

vadimzhukck:
# Description

Binance exchange raw response: {"code":-1102,"msg":"Mandatory parameter 'timeInForce' was not sent, was empty/null, or malformed."}


shazbert:
@vadimzhukck Thanks for opening this PR. If you just push a new commit even a force update, travis will automatically build without the need to close and open.

PR #262, opened 2019-03-25
Optimisation: large structs/huge param fixes (part 2)

xtda:
# Description

Second round of pointer usage for large structs this should take care of all the simple changes that don't require code changes 

Also includes a case correction for newOrderBook to make it consistent across the code base

## Type of change

# How Has This Been Tested?

Automated testing using go test 
Manual building and running with all exchanges enabled as well

# Checklist:

### DIFF
--- a/currency/forexprovider/forexprovider.go
+++ b/currency/forexprovider/forexprovider.go
@@ -130,7 +130,7 @@ func StartFXService(fxProviders []base.Settings) (*ForexProviders, error) {
 	}
 
 	if handler.Primary.Provider == nil {
-		return nil, errors.New("no foreign exchange providers enabled")
+		return nil, errors.New("no Foreign primary provider enabled")
 	}
 
 	return handler, nil
--- a/currency/forexprovider/forexprovider.go
+++ b/currency/forexprovider/forexprovider.go
@@ -130,7 +130,7 @@ func StartFXService(fxProviders []base.Settings) (*ForexProviders, error) {
 	}
 
 	if handler.Primary.Provider == nil {
-		return nil, errors.New("no Foreign primary provider enabled")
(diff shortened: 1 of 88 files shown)


currency/storage.go
```diff
@@ -369,37 +369,42 @@ func (s *Storage) WriteCurrencyDataToFile(path string, mainUpdate bool) error {
 }
 
 // LoadFileCurrencyData loads currencies into the currency codes
-func (s *Storage) LoadFileCurrencyData(f File) error {
+func (s *Storage) LoadFileCurrencyData(f *File) error {
 	for _, contract := range f.Contracts {
```
shazbert: Might be able to dereference the pointer and get the mem address of the item so we don't double copy like so, might not work correctly?
```go
for i := range f.Contracts {
		err := s.currencyCodes.LoadItem(&(*f).Contracts[i])
		if err != nil {
			return err
		}
	}
```

currency/storage.go
```diff
@@ -369,37 +369,42 @@ func (s *Storage) WriteCurrencyDataToFile(path string, mainUpdate bool) error {
...
-		err := s.currencyCodes.LoadItem(contract)
+		contract := contract
```
thrasher-: Perhaps c := &contract and use it param for LoadItem would be better than this double assignment?

currency/forexprovider/forexprovider.go
```diff
@@ -130,7 +130,7 @@ func StartFXService(fxProviders []base.Settings) (*ForexProviders, error) {
 	}
 
 	if handler.Primary.Provider == nil {
-		return nil, errors.New("no foreign exchange providers enabled")
+		return nil, errors.New("no Foreign primary provider enabled")
```
thrasher-: "no primary forex provider enabled"

PR #266, opened 2019-04-03
Improved test coverage and functions in common.go

MadCozBadd:
## Type of change

# How Has This Been Tested?

Go Test ./...

# Checklist:


common/common_test.go
```diff
@@ -950,3 +953,71 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
 		t.Error("Test failed. Common TimeFromUnixTimestampFloat. Converted invalid syntax.")
 	}
 }
+
+func TestGetDefaultDataDir(t *testing.T) {
+	expectedOutput := os.Getenv("APPDATA") + GetOSPathSlash() + "GoCryptoTrader"
+	actualOutput := GetDefaultDataDir("windows")
+	t.Log(actualOutput)
```
thrasher-: Can remove this non-needed t.Log and need to add support for getting the default data dir on non-windows based environments.

common/common_test.go
```diff
@@ -950,3 +953,71 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+		dir = dir + GetOSPathSlash() + "GoCryptoTrader\\TestFileASDFG"
+		err = CreateDir(dir)
+		if err != nil {
+			t.Errorf("Err: %s", err)
+		}
+		os.Remove(dir)
+		dir, _ = os.LookupEnv("*")
+		err = CreateDir(dir)
+		if err == nil {
+			t.Errorf("Expected err due to invalid path, got err: %v", err)
```
thrasher-: Err will be nil in this case, so don't need to print its value. "Expected err for invalid path but got nil"

xtda:
Since we are now changing the default folder creation permissions from 777 to 770 do we need to add a start up check for existing permissions and update them to 770? @thrasher- thoughts?

This can be achieved with os.Stat & os.Chmod

common/common_test.go
```diff
@@ -950,3 +953,71 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+	if actualOutput != expectedOutput {
+		t.Errorf("Unexpected result. Got: %v Expected: %v", actualOutput, expectedOutput)
+	}
+}
+
+func TestCreateDir(t *testing.T) {
+	switch runtime.GOOS {
+	case "windows":
+		// test for a directory that exists
+		dir := "c:\\temp"
```
thrasher-: Would be better to use env TEMP as this current value is an old default used by Windows pre Windows XP but does not exist on Windows 10 (Instead it's located in %USER%\AppData\Local\Temp)

thrasher-:
> Since we are now changing the default folder creation permissions from 777 to 770 do we need to add a start up check for existing permissions and update them to 770? @thrasher- thoughts?
> 
> This can be achieved with os.Stat & os.Chmod

Definitely agree, we touched on this when changing the permissions originally but thought we could do it in a diff PR.

common/common_test.go
```diff
@@ -950,3 +953,71 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+		}
+		dir = path.Join(os.ExpandEnv("$HOME"), ".gocryptotrader", "TestFileASFG")
+		err = CreateDir(dir)
+		if err != nil {
+			t.Errorf("Err: %s", err)
+		}
+		os.Remove(dir)
+		err = CreateDir(":")
+		if err == nil {
+			t.Errorf("Expected err due to invalid path, got err: %v", err)
```
thrasher-: Same for here: Err will be nil in this case, so don't need to print its value. "Expected err for invalid path but got nil"

common/common_test.go
```diff
@@ -950,3 +953,70 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
 		t.Error("Test failed. Common TimeFromUnixTimestampFloat. Converted invalid syntax.")
 	}
 }
+
+func TestGetDefaultDataDir(t *testing.T) {
+	expectedOutput := os.Getenv("APPDATA") + GetOSPathSlash() + "GoCryptoTrader"
```
thrasher-: Last request here is that we'll need to add test support for darwin/linux, then this PR looks GTG!

common/common_test.go
```diff
@@ -950,3 +953,77 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+		if err == nil {
+			t.Errorf("Expected err due to invalid path, but got nil")
+		}
+	}
+}
+
+func TestListAllDir(t *testing.T) {
+	_, err := ListAllDir(GetDefaultDataDir("windows"))
+	if err != nil {
+		t.Error("Something bad happened", err)
```
shazbert: Please update wording here

common/common.go
```diff
@@ -610,22 +610,40 @@ func GetDefaultDataDir(env string) string {
...
-
 	log.Warnf("Directory %s does not exist.. creating.", dir)
-	err = os.MkdirAll(dir, 0777)
+	return os.MkdirAll(dir, 0770)
+}
+
+// ListAllDir lists all the directories and files in an array
+func ListAllDir(directory string) ([]string, error) {
+	var list []string
+	err := filepath.Walk(directory, func(path string, info os.FileInfo, err error) error {
```
shazbert: You need to check error within this anon func.

common/common.go
```diff
@@ -610,22 +610,40 @@ func GetDefaultDataDir(env string) string {
...
+	})
+	return list, err
+}
+
+// ChangePerm changes file permissions in a given directory to 0770
+func ChangePerm(env string) error {
+	a, err := ListAllDir(GetDefaultDataDir(env))
 	if err != nil {
-		return fmt.Errorf("failed to create dir. Err: %s", err)
+		return errors.New("Directory provided is invalid")
```
shazbert: all letters in error types need to be lower case

common/common.go
```diff
@@ -610,22 +610,40 @@ func GetDefaultDataDir(env string) string {
...
+	}
+	for i := range a {
+		b, _ := os.Stat(a[i])
```
shazbert: It might be better to at least handle this error for an edge case

common/common.go
```diff
@@ -610,22 +610,40 @@ func GetDefaultDataDir(env string) string {
...
-		return fmt.Errorf("directory %s does not exist. Err: %s", dir, err)
-	}
-
 	log.Warnf("Directory %s does not exist.. creating.", dir)
-	err = os.MkdirAll(dir, 0777)
+	return os.MkdirAll(dir, 0770)
+}
+
+// ListAllDir lists all the directories and files in an array
+func ListAllDir(directory string) ([]string, error) {
```
xtda: You have called this function ListAllDir's but due to your code on line 628 it will only append folders

not handling the error here will also cause the problem to segfault if the folder doesn't exist

common/common.go
xtda: The os.Fileinfo interface above also returns FileMode that we can use to obtain permissions instead of stating each one again

It might be worth merging both of these into a single function or holding the data and passing a pointer to it if you do want to leave them as seperate

common/common.go
```diff
@@ -610,22 +610,40 @@ func GetDefaultDataDir(env string) string {
...
+			list = append(list, path)
+		}
+		return nil
+	})
+	return list, err
+}
+
+// ChangePerm changes file permissions in a given directory to 0770
+func ChangePerm(env string) error {
+	a, err := ListAllDir(GetDefaultDataDir(env))
```
xtda: To make this a little more generic we could remove the usage of GetDefaultDataDir here that way it supports any folder and calls to it could pass the path which would also mean remaining the param from env to something more generic

common/common.go
```diff
@@ -610,22 +610,40 @@ func GetDefaultDataDir(env string) string {
...
 	if err != nil {
-		return fmt.Errorf("failed to create dir. Err: %s", err)
+		return errors.New("Directory provided is invalid")
+	}
+	for i := range a {
+		b, _ := os.Stat(a[i])
+		if b.Mode().Perm() != 0770 {
+			os.Chmod(a[i], 0770)
```
xtda: chmod does return an error that is worth checking to confirm there was no issues setting the permissions requested

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
 		t.Error("Test failed. Common TimeFromUnixTimestampFloat. Converted invalid syntax.")
 	}
 }
+
+func TestGetDefaultDataDir(t *testing.T) {
+	switch {
```
thrasher-: This can be optomised to switch (runtime.GOOS) and case "windows": etc

common/common.go
```diff
@@ -607,25 +607,37 @@ func GetDefaultDataDir(env string) string {
...
+}
+
+// ChangePerm lists all the directories and files in an array
+func ChangePerm(directory string) error {
+	return filepath.Walk(directory, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		if info.Mode().Perm() != 0770 {
+			chModErr := os.Chmod(path, 0770)
```
thrasher-: return os.Chmod(path, 0770)

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+			t.Fatalf("Unexpected result. Got: %v Expected: %v", actualOutput, dir)
+		}
+	}
+}
+
+func TestCreateDir(t *testing.T) {
+	switch runtime.GOOS {
+	case "windows":
+		// test for a directory that exists
+		dir, _ := os.LookupEnv("TEMP")
```
thrasher-: please add check for os.LookupEnv 2nd return here

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+		os.Remove(dir)
+		err = CreateDir(":")
+		if err == nil {
+			t.Fatal("expected err due to invalid path, but got nil")
+		}
+	}
+}
+
+func TestChangePerm(t *testing.T) {
+	switch {
```
thrasher-: can be changed to switch (runtime.GOOS) with case "linux", "darwin":

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+	switch runtime.GOOS {
+	case "windows":
+		// test for a directory that exists
+		dir, _ := os.LookupEnv("TEMP")
+		err := CreateDir(dir)
+		if err != nil {
+			t.Fatalf("CreateDir failed. Err: %v", err)
+		}
+		// test for creating a directory
+		dir, _ = os.LookupEnv("APPDATA")
```
shazbert: err check here

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+		dir = dir + GetOSPathSlash() + "GoCryptoTrader\\TestFileASDFG"
+		err = CreateDir(dir)
+		if err != nil {
+			t.Fatalf("CreateDir failed. Err: %s", err)
+		}
+		os.Remove(dir)
+		dir, _ = os.LookupEnv("*")
```
shazbert: err check here

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+			t.Fatalf("CreateDir failed. Err: %v", err)
+		}
+		// test for creating a directory
+		dir, _ = os.LookupEnv("APPDATA")
+		dir = dir + GetOSPathSlash() + "GoCryptoTrader\\TestFileASDFG"
+		err = CreateDir(dir)
+		if err != nil {
+			t.Fatalf("CreateDir failed. Err: %s", err)
+		}
+		os.Remove(dir)
```
shazbert: err check here

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+		dir, ok = os.LookupEnv("HOME")
+		if !ok {
+			t.Fatalf("LookupEnv of HOME failed")
+		}
+		dir = path.Join(dir, ".gocryptotrader", "TestFileASFG")
+		err = CreateDir(dir)
+		if err != nil {
+			t.Errorf("CreateDir failed. Err: %s", err)
+		}
+		os.Remove(dir)
```
shazbert: err check here

common/common_test.go
```diff
@@ -950,3 +953,156 @@ func TestTimeFromUnixTimestampFloat(t *testing.T) {
...
+		dir, ok = os.LookupEnv("HOME")
+		if !ok {
+			t.Fatalf("LookupEnv of HOME failed")
+		}
+		dir = path.Join(dir, ".gocryptotrader", "TestFileASFG")
+		err = CreateDir(dir)
+		if err != nil {
+			t.Fatalf("CreateDir failed. Err: %s", err)
+		}
+		os.Remove(dir)
```
shazbert: err check here

thrasher-:
tACK 👍

PR #274, opened 2019-04-09
Offline worst case trade fees

gloriousCode:
# Description

Adds offline trade fee calculation for each exchange using worst case values. On the wrapper, if the keys aren't set, then switch the fee type to offline to prevent API calls.

## Type of change

# How Has This Been Tested?

I've added a logic test `TestGetFeeByTypeOfflineTradeFee` to check if offline trade fee feetype gets hit

# Checklist:

### DIFF
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -31,42 +31,42 @@ const (
 )
 
 // FeeType custom type for calculating fees based on method
-type FeeType string
-
-// InternationalBankTransactionType custom type for calculating fees based on fiat transaction types
-type InternationalBankTransactionType string
+type FeeType int
 
 // Const declarations for fee types
 const (
-	BankFee                        FeeType = "bankFee"
-	InternationalBankDepositFee    FeeType = "internationalBankDepositFee"
-	InternationalBankWithdrawalFee FeeType = "internationalBankWithdrawalFee"
-	CryptocurrencyTradeFee         FeeType = "cryptocurrencyTradeFee"
-	CyptocurrencyDepositFee        FeeType = "cyptocurrencyDepositFee"
-	CryptocurrencyWithdrawalFee    FeeType = "cryptocurrencyWithdrawalFee"
-	SimulatedTransactionFee        FeeType = "simulatedTransactionFee"
+	BankFee                        FeeType = 0
+	InternationalBankDepositFee    FeeType = 1
+	InternationalBankWithdrawalFee FeeType = 2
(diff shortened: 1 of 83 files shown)


exchanges/exchange.go
```diff
@@ -31,41 +31,42 @@ const (
...
 
 // Const declarations for fee types
 const (
-	BankFee                        FeeType = "bankFee"
-	InternationalBankDepositFee    FeeType = "internationalBankDepositFee"
-	InternationalBankWithdrawalFee FeeType = "internationalBankWithdrawalFee"
-	CryptocurrencyTradeFee         FeeType = "cryptocurrencyTradeFee"
-	CyptocurrencyDepositFee        FeeType = "cyptocurrencyDepositFee"
-	CryptocurrencyWithdrawalFee    FeeType = "cryptocurrencyWithdrawalFee"
+	BankFee                        FeeType = 0
```
thrasher-: This can be simplified to iota and the values below will automatically be incremented

exchanges/exchange.go
```diff
@@ -31,41 +31,42 @@ const (
...
-	RapidTransfer   InternationalBankTransactionType = "rapidTransfer"
-	MisterTangoSEPA InternationalBankTransactionType = "misterTangoSepa"
-	Qiwi            InternationalBankTransactionType = "qiwi"
-	VisaMastercard  InternationalBankTransactionType = "visaMastercard"
-	WebMoney        InternationalBankTransactionType = "webMoney"
-	Capitalist      InternationalBankTransactionType = "capitalist"
-	WesternUnion    InternationalBankTransactionType = "westernUnion"
-	MoneyGram       InternationalBankTransactionType = "moneyGram"
-	Contact         InternationalBankTransactionType = "contact"
+	WireTransfer    InternationalBankTransactionType = 0
```
thrasher-: Same here

exchanges/okex/okex_test.go
```diff
@@ -1551,26 +1551,6 @@ func TestGetETTOrderDetails(t *testing.T) {
 	testStandardErrorHandling(t, err)
 }
 
-// TestGetETTConstituents API endpoint test
```
thrasher-: Can t.Skip these, if they don't bring them back online then we can remove the functions and associated tests.

gloriousCode: Far out that wasn't thought through 😄 "didn't work, better remove the tests!" Will add back. Can remove ETT if it doesn't return in another PR later 😄

PR #306, opened 2019-05-21
Config: Fix AuthenticatedAPISupport default values check

gloriousCode:
# Description
Coinut only uses one key (`APIKey`). If you try and set `AuthenticatedApiSupport: true` in config and only set the `APIKey`, `AuthenticatedApiSupport` will always be set to false due to an if statement that will always match.

I've swapped the if statements around so the more specific scenario is hit first and correctly sets `authenticatedApiSupport`

## Type of change

# How Has This Been Tested?
I was wondering why authenticated WS testing for COINUT wasn't working for way too long before realising that `APIKey`, `ClientId` and  `AuthenticatedApiSupport` were not being properly set due to the above. 

Switching the if around fixed it. All tests pass

Checklist:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -799,17 +799,17 @@ func (c *Config) CheckExchangeConfigValues() error {
 				return fmt.Errorf(ErrExchangeBaseCurrenciesEmpty, c.Exchanges[i].Name)
 			}
 			if c.Exchanges[i].AuthenticatedAPISupport { // non-fatal error
-				if c.Exchanges[i].APIKey == "" || c.Exchanges[i].APISecret == "" ||
-					c.Exchanges[i].APIKey == DefaultUnsetAPIKey ||
-					c.Exchanges[i].APISecret == DefaultUnsetAPISecret {
-					c.Exchanges[i].AuthenticatedAPISupport = false
-					log.Warnf(WarningExchangeAuthAPIDefaultOrEmptyValues, c.Exchanges[i].Name)
-				} else if c.Exchanges[i].Name == "ITBIT" || c.Exchanges[i].Name == "Bitstamp" || c.Exchanges[i].Name == "COINUT" || c.Exchanges[i].Name == "CoinbasePro" {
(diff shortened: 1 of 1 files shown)


config/config.go
```diff
@@ -799,16 +799,21 @@ func (c *Config) CheckExchangeConfigValues() error {
...
-				if c.Exchanges[i].APIKey == "" || c.Exchanges[i].APISecret == "" ||
-					c.Exchanges[i].APIKey == DefaultUnsetAPIKey ||
-					c.Exchanges[i].APISecret == DefaultUnsetAPISecret {
+				if c.Exchanges[i].APIKey == "" || c.Exchanges[i].APIKey == DefaultUnsetAPIKey {
+					c.Exchanges[i].AuthenticatedAPISupport = false
+					log.Warnf(WarningExchangeAuthAPIDefaultOrEmptyValues, c.Exchanges[i].Name)
+				}
+
+				if (c.Exchanges[i].APISecret == "" || c.Exchanges[i].APISecret == DefaultUnsetAPISecret) &&
+					c.Exchanges[i].AuthenticatedAPISupport && c.Exchanges[i].Name != "COINUT" {
```
thrasher-: You can drop the additional check for c.Exchanges[i].AuthenticatedAPISupport here as it's being done up above

config/config.go
```diff
@@ -799,16 +799,21 @@ func (c *Config) CheckExchangeConfigValues() error {
...
+					c.Exchanges[i].AuthenticatedAPISupport = false
+					log.Warnf(WarningExchangeAuthAPIDefaultOrEmptyValues, c.Exchanges[i].Name)
+				}
+
+				if (c.Exchanges[i].ClientID == "ClientID" || c.Exchanges[i].ClientID == "") && c.Exchanges[i].AuthenticatedAPISupport &&
```
thrasher-: Same for here, then this looks GTG!

gloriousCode: Main reason was to prevent multiple log.Warns at the end. However, I'll remove those extra checks and just add a 
```
if !c.Exchanges[i].AuthenticatedAPISupport {
					log.Warnf(WarningExchangeAuthAPIDefaultOrEmptyValues, c.Exchanges[i].Name)
				}
```
at the end

thrasher-: Sounds good 👍

PR #311, opened 2019-05-25
Bump golangci-lint version to 1.16.0

shazbert:
## Type of change

# How Has This Been Tested?

# Checklist:

### DIFF
--- a/exchanges/btcmarkets/btcmarkets_wrapper.go
+++ b/exchanges/btcmarkets/btcmarkets_wrapper.go
@@ -232,6 +232,7 @@ func (b *BTCMarkets) CancelAllOrders(_ *exchange.OrderCancellation) (exchange.Ca
 		orderIDInt, err := strconv.ParseInt(openOrders[i].ID, 10, 64)
 		if err != nil {
 			cancelAllOrdersResponse.OrderStatus[openOrders[i].ID] = err.Error()
+			continue
 		}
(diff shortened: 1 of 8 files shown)


exchanges/btcmarkets/btcmarkets_wrapper.go
```diff
@@ -231,7 +231,7 @@ func (b *BTCMarkets) CancelAllOrders(_ *exchange.OrderCancellation) (exchange.Ca
 	for i := range openOrders {
 		orderIDInt, err := strconv.ParseInt(openOrders[i].ID, 10, 64)
 		if err != nil {
-			cancelAllOrdersResponse.OrderStatus[openOrders[i].ID] = err.Error()
+			return cancelAllOrdersResponse, err
```
thrasher-: Since we're cancelling a bunch of orders, we want to report that this specific conversion of the order ID failed, rather then return an err this early on and ignore cancelling a bunch of other orders.

shazbert: Yeah will revert, my reasoning was its a massive issue which will occur across all ids if ID structure was changed and cant be parsed by str conv.

thrasher-: Yeah I understand, this occurence would be very, very rare (breaking change for all API users) but at least we still check for it.

PR #319, opened 2019-06-24
New logging system

xtda:
# Description

This PR replaces the current logging system with an extended log system that supports sublogging  
Any sub logger can have its own level and output (multi output is supported and can be sent to anything that implements io.Writer interface)

This changes the syntax to Level("subsystem", "log message")

Example:
```sh
log.Info("log", "this is a log message)
```

It also extends the configuration options for the spacer timestamp and headers to be user configurable 

### Example Config:
```js
 "logging": {
  "enabled": true,
  "advancedSettings": {
   "spacer": " | ",
   "timeStampFormat": " 02/01/2006 15:04:05 ",
   "headers": {
    "info": "[INFO] ",
    "warn": "[WARN] ",
    "debug": "[DEBUG]",
    "error": "[ERROR]"
   }
  },
  "subloggers": [
   {
    "name": "log",
    "level": "INFO|DEBUG|WARN|ERROR",
    "output": "stdout"
   }
  ]
 },
```

## Type of change

# How Has This Been Tested?

Automated testing & Benchmarks

# Checklist:


logger/logger.go
```diff
@@ -1,120 +1,53 @@
...
+	}
+	e.data = append(e.data, l.Spacer...)
+	e.data = append(e.data, []byte(data)...)
 
-	debugLogger.SetOutput(ioutil.Discard)
-	infoLogger.SetOutput(ioutil.Discard)
-	warnLogger.SetOutput(ioutil.Discard)
-	errorLogger.SetOutput(ioutil.Discard)
-	fatalLogger.SetOutput(ioutil.Discard)
+	e.output.Write(e.data)
```
gloriousCode: I've noticed you append "\n" on almost every log. Is there a reason for it being sometimes added? Rather than put the onus on the individual log to have a new line (and ruin the output if they forget), I think it should be added once here (or somewhere where its only appended once). You could always check here if its already on the end before appending to prevent an additional "\n". I think the benefits of adding it here it are better than the chance that someone adds an extra empty line

logger/logger_types.go
```diff
@@ -2,34 +2,90 @@ package logger
...
+			return &LogEvent{
+				data: make([]byte, 0, 80),
+			}
+		},
+	}
 	LogPath string
+)
+
+const (
+	LogGlobal = "log"
```
gloriousCode: I feel like this can be renamed to "Global" to minimise duplication re log.LogGlobal

logger/logger_types.go
```diff
@@ -2,34 +2,90 @@ package logger
...
+
+	SubSystemConnMgr = "connection"
```
gloriousCode: This is the minor of the minoriest nits. I feel that these don't need the "SubSystem" prefix. Any confusion could be prevented by comments. Since a global/subsystem is required for each log, then the user will understand that the specific log type is a subsytem on account of not being called Global. This isn't enough for me to request changes though and don't mind if you prefer the prefix and it stays because I also like things being explicit 😆

logger/logger_types.go
```diff
@@ -2,34 +2,90 @@ package logger
...
+	SubSystemCommMgr = "communications"
+	SubSystemConfMgr = "config"
+	SubSystemOrdrMgr = "order"
+	SubSystemPortMgr = "portfolio"
+	SubSystemSyncMgr = "syncer"
+	SubSystemTimeMgr = "timekeeper"
+	SubSystemWsocMgr = "websocket"
+	SubSystemEvntMgr = "event"
```
gloriousCode: Another minor thought is that this could be a custom type. You already have some code to protect from custom subsystems from being passed in, but if someone is looking to log and you see a `string` is passed in, one could easily conclude that they could input anything. The lack of any comments on the[ `func Info`](https://github.com/thrasher-/gocryptotrader/pull/319/files#diff-5c91e5a763e453613ac4f04d193662fcR5) etc allow for this too. Only at compile time will they discover that their log is not allowed and instead just says `log`. 

By having a type, a user can know that there is a strict definition of the kinds of subsystems that can be passed to it and then append any when new subsystems are developed.

coverage.txt
```diff
@@ -0,0 +1,3325 @@
+mode: atomic
```
gloriousCode: Is this required to be added?

logger/logger.go
xtda: This was something we discussed on when a new line should be added

In the std logger (which is what we currently use) a new line is appended if one is not found 
```go
	if len(s) == 0 || s[len(s)-1] != '\n' {
		l.buf = append(l.buf, '\n')
	}
```
It was decided that Info/Info (and all other versions of it Debug/Debugf etc) will not append a new line and output exactly as the user inputs it while Infoln would append a new line

I went through and updated all usage of Info/Infof to append a new line as required
I am happy to revert this change if we would rather just append a new line 

@thrasher-

coverage.txt
xtda: No this was committed by mistake will removed

logger/logger_types.go
xtda: The subsystem naming is something i am fixing today i am not happy with it overall the way that they are setup and added  

As you mentioned below the first part just accepting a string can cause confusion on what should be passed 

I am going to rework the way that subsystems are registered and remove the "global" as being log subsystem 

You will end up with a default level/output as part of the main config and then any optional subsystems override these settings

something like
```js
 "logging": {
  "enabled": true,
  "level": "INFO|DEBUG|WARN|ERROR",
  "output": "stdout",
  "advancedSettings": {
   "spacer": " | ",
   "timeStampFormat": " 02/01/2006 15:04:05 ",
   "headers": {
    "info": "[INFO] ",
    "warn": "[WARN] ",
    "debug": "[DEBUG]",
    "error": "[ERROR]"
   }
  },
  "subloggers": [
   {
    "name": "ticker",
    "level": "INFO",
    "output": "file"
   },
   {
    "name": "portfolio",
    "level": "INFO|DEBUG",
    "output": "file"
   }
  ]
 }
```

logger/logger.go
gloriousCode: Thanks for the explanation. I appreciate that its being explicit there with `log` vs` logln`. Though I also feel that the use-case for wanting to have extra data/logs on the same line from multiple `log.Debug` could easily be handled via `log.Debugf("%v%v%v")`. The downside though is if a user forgets the  "\n" in any future PR we will be asking  "are you intending to have multiple logs on the same line?". But if its ever a mistake, its something that a user will only forget once (hopefully).
So I'm okay if it stays, but think that it will be something we might run into a few times mistakenly

go.mod
```diff
@@ -5,6 +5,7 @@ go 1.12
 require (
 	github.com/gogo/protobuf v1.2.1 // indirect
 	github.com/golang/protobuf v1.3.1
+	github.com/golangci/golangci-lint v1.16.0 // indirect
```
thrasher-: It looks like a go get github.com/golangci/golangci-lint was done with GO111MODULE=on in the root dir, it's better to revert this to master as it's an optional dependancy

go.sum
```diff
@@ -1,20 +1,87 @@
 cloud.google.com/go v0.26.0 h1:e0WKqKTd5BnrG8aKH3J3h+QvEIQtSUcf2n5UZ5ZgLtQ=
 cloud.google.com/go v0.26.0/go.mod h1:aQUYkXzVsufM+DwF1aE+0xfcU+56JwCaLick0ClmMTw=
+github.com/BurntSushi/toml v0.3.1 h1:WXkYYl6Yr3qBf1K79EBnL4mak0OimBfB0XUf9Vl28OQ=
```
thrasher-: Same as the previous go mod comment above.

logger/logger_test.go
```diff
@@ -11,77 +11,92 @@ var (
 	falseptr = func(b bool) *bool { return &b }(false)
```
thrasher-: falseptr is unused

testdata/configtest.json
```diff
@@ -4,10 +4,31 @@
  "globalHTTPTimeout": 15000000000,
  "logging": {
   "enabled": true,
-  "file": "debug.txt",
-  "colour": false,
-  "level": "DEBUG|WARN|INFO|ERROR|FATAL",
-  "rotate": true
+  "advancedSettings": {
```
thrasher-: Please revert this to master, but with the logger changes as this config is kept in sync with master to test backwards compatibility between engine and master

logger/logger_file.go
```diff
@@ -0,0 +1,40 @@
...
+	"path/filepath"
+)
+
+type fileWriters struct {
+	fileHandle []os.File
+}
+
+var f = fileWriters{}
+
+func createFileHandle(subsystem string) (io.Writer, error) {
```
thrasher-: Perhaps make this ability to write to custom subsystems an advanced setting as I tend to like all logs in one single file, even though others might prefer a log file per subsystem

xtda: Made the change to support both

there is "file" and "ownfile" output options now
with file being a global file handle that writes to GCTPATH/logs/log.txt and ownfile being its own that writes to GCTPATH/logs/<loggername.txt> (you can also write to both at the same time if you wish)

engine/comms_relayer.go
```diff
@@ -64,7 +64,7 @@ func (c *commsManager) Stop() error {
 	}
 
 	close(c.shutdown)
-	log.Debugln("Communications manager shutting down...")
+	log.Debugln(log.CommunicationMgr, "Communications manager Inside front door to the right behind utting down...")
```
gloriousCode: 👀

xtda: The communications manager is inside the front door right behind utting :laughing:

gloriousCode:
Running this off of a `config_example.json` prior to your `config_example.json` changes results in the following logging details for `config.json`:
```
 "logging": {
  "enabled": true,
  "level": "DEBUG|WARN|INFO|ERROR|FATAL",
  "output": "",
  "advancedSettings": {
   "spacer": "",
   "timeStampFormat": "",
   "headers": {
    "info": "",
    "warn": "",
    "debug": "",
    "error": ""
   }
  }
 },
```
which will not output anything. So anyone with a config prior to your changes would not upgrade their config to the defaults provided in your config_example.

xtda:
> Running this off of a `config_example.json` prior to your `config_example.json` changes results in the following logging details for `config.json`:
> 
> ```
>  "logging": {
>   "enabled": true,
>   "level": "DEBUG|WARN|INFO|ERROR|FATAL",
>   "output": "",
>   "advancedSettings": {
>    "spacer": "",
>    "timeStampFormat": "",
>    "headers": {
>     "info": "",
>     "warn": "",
>     "debug": "",
>     "error": ""
>    }
>   }
>  },
> ```
> 
> which will not output anything. So anyone with a config prior to your changes would not upgrade their config to the defaults provided in your config_example.

Good point I will update CheckLoggerConfig() to also allow for a migration from old config and apply some defaults to these if not set

logger/loggers.go
```diff
@@ -2,79 +2,160 @@ package logger
...
+		return
+	}
+
+	Debug(sl, fmt.Sprintf(data, v...))
 }
 
-// Warn handler takes any input returns unformatted output to warnLogger writer
-func Warn(v ...interface{}) {
-	warnLogger.Print(v...)
+// Debugf takes a pointer subLogger struct & string  and sends to newLogEvent()
```
shazbert: Debugf -> Warn and there is an extra space between *string* and *and*

cmd/portfolio/portfolio.go
```diff
@@ -64,7 +65,8 @@ func main() {
 
 	defaultCfg, err := config.GetFilePath("")
 	if err != nil {
-		log.Fatal(err)
+		log.Println(err)
```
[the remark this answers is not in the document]
xtda: portfolio is a standalone command and does not import the logger instead uses the standard go logger which only has Println()

logger/logger.go
```diff
@@ -1,120 +1,79 @@
...
-func clearAllLoggers() {
-	debugLogger.SetFlags(0)
-	infoLogger.SetFlags(0)
-	warnLogger.SetFlags(0)
-	errorLogger.SetFlags(0)
-	fatalLogger.SetFlags(0)
+func Level(s string) (*Levels, error) {
+	found, logger := validSubLogger(s)
+	if !found {
+		return nil, fmt.Errorf("logger %v not found", logger)
```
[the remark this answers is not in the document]
xtda: Whoops was suppose to the input not the logger and pass it down

Fixed!

logger/logger.go
```diff
@@ -1,120 +1,79 @@
...
-		if err != nil {
-			return
-		}
-		logOutput = io.MultiWriter(os.Stdout, logFileHandle)
-	} else {
-		logOutput = os.Stdout
+func SetLevel(s, level string) (*Levels, error) {
+	found, logger := validSubLogger(s)
+	if !found {
+		return nil, fmt.Errorf("logger %v not found", logger)
```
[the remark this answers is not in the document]
xtda: Since the pointer is passed back from validSubLoggers there is almost no performance benefit for or against having it return a bool and comes down to code complexity/readability

Is it easier to return a bool, pointer and check if the bool is true/false or return a pointer and check if its nil?

BenchmarkBothValidSubLoggers/NoBool-8         	  500000	      3389 ns/op	    1243 B/op	      16 allocs/op
BenchmarkBothValidSubLoggers/Bool-8           	  500000	      3386 ns/op	    1304 B/op	      16 allocs/op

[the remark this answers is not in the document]
shazbert: Up to you.

logger/logger_multiwriter.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	results := make(chan data)
+
+	for _, wr := range mw.writers {
+		go func(w io.Writer, p []byte, ch chan data) {
+
+			n, err = w.Write(p)
+			if err != nil {
+				ch <- data{n, err}
```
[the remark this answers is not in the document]
xtda: Errors being out of order don't matter and it wont block

It does cause a goroutine leak but so good pick up :D

```bash
goroutine profile: total 9506
9322 @ 0x430d1f 0x406c2b 0x406c01 0x4069e5 0x745f71 0x45e1b1
#	0x745f70	github.com/thrasher-/gocryptotrader/logger.(*multiWriter).Write.func1+0x130	/home/andrew/dev/projects/thrashercorp/gocryptotrader-xtda/logger/logger_multiwriter.go:54
```

image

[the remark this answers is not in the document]
shazbert: :+1: spawn 9322 routines "This is fine" :dancer: Haha

shazbert:
LGTM!!!

xtda:
I renamed the file json entry to fileSettings to maintain support for a pre-engine config migration as old logger also included a filed "file" that caused go to try unmarshall it as a string

shazbert:
@thrasher- There's a linter issue, do you want this done or in another PR?

config_example.json
```diff
@@ -4,10 +4,24 @@
  "globalHTTPTimeout": 15000000000,
  "logging": {
   "enabled": true,
-  "file": "debug.txt",
-  "colour": false,
-  "level": "DEBUG|WARN|INFO|ERROR|FATAL",
-  "rotate": false
+  "level": "INFO|WARN|DEBUG|ERROR",
+  "output": "console",
+  "file": {
```
thrasher-: fileSettings for the JSON key name

cmd/exchange_template/wrapper_file.tmpl
```diff
@@ -24,9 +24,9 @@ func ({{.Variable}} *{{.CapitalName}}) Start(wg *sync.WaitGroup) {
 // Run implements the {{.CapitalName}} wrapper
 func ({{.Variable}} *{{.CapitalName}}) Run() {
 	if {{.Variable}}.Verbose {
-{{if .WS}} log.Debugf("%s Websocket: %s. (url: %s).\n", {{.Variable}}.GetName(), common.IsEnabled({{.Variable}}.Websocket.IsEnabled()), {{.Variable}}.Websocket.GetWebsocketURL()) {{end}}
-		log.Debugf("%s polling delay: %ds.\n", {{.Variable}}.GetName(), {{.Variable}}.RESTPollingDelay)
-		log.Debugf("%s %d currencies enabled: %s.\n", {{.Variable}}.GetName(), len({{.Variable}}.EnabledPairs), {{.Variable}}.EnabledPairs)
+{{if .WS}} log.Debugf("", %s Websocket: %s. (url: %s).\n", {{.Variable}}.GetName(), common.IsEnabled({{.Variable}}.Websocket.IsEnabled()), {{.Variable}}.Websocket.GetWebsocketURL()) {{end}}
```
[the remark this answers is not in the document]
xtda: done

config_example.json
xtda: whoops missed that when i made the change last night! updated

xtda:
> Epic stuff @xtda! Just 2 last nits and please go fmt addr_helpers_test.go before this is merged in.

Done local linter also picked up some additional files in exchange/orderbook that i ran gofmt on as well

shazbert:
@thrasher- @xtda  Bug in engine test which is on my local as well. Builds fine and runs well.

thrasher-:
> @thrasher- @xtda Bug in engine test which is on my local as well. Builds fine and runs well.

Bug was because of a new line added in this PR https://github.com/thrasher-/gocryptotrader/pull/319/files#diff-3895c8f98f6613a5e46ea0f26ecfe893R146, will merge master and fix it in that

PR #338, opened 2019-08-08
Update URLs for transfer into org repo

thrasher-:
# Description

Updates all project URLs to the new organisation repo (thrasher-corp/gocryptotrader) instead of on my personal account. All GCT related repos will live here from now on and we can now utilise the team feature to manage specific repos. This requires ACKs from the team before it can be merged in for security reasons.


README.md
```diff
@@ -1,10 +1,10 @@
-<img src="https://github.com/thrasher-/gocryptotrader/blob/master/web/src/assets/page-logo.png?raw=true" width="350px" height="350px" hspace="70">
+<img src="https://github.com/thrasher-corp/gocryptotrader/blob/master/web/src/assets/page-logo.png?raw=true" width="350px" height="350px" hspace="70">
 
 [![Build Status](https://travis-ci.com/thrasher-/gocryptotrader.svg?branch=master)](https://travis-ci.com/thrasher-/gocryptotrader)
```
xtda: This and most of the other readme's are still pointing to the old link [Travis](https://travis-ci.com/thrasher-corp/gocryptotrader) has updated

codecov is also he same in a few of the sub package readmes
Don't want to tag them all since there is 59

Unsure if intended or not

thrasher-: Thanks! Have just pushed an update to correct those

PR #339, opened 2019-08-20
Bithumb fixes due to API change

shazbert:
# Description

API change occurred and broke type assertion. So fixed that and ticker type.

## Type of change

- [x]  Bug fix (non-breaking change which fixes an issue)

# How Has This Been Tested?

go test package

# Checklist:

- [n/a] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool 
- [n/a] Any dependent changes have been merged and published in downstream modules

### DIFF
--- a/exchanges/bithumb/bithumb.go
+++ b/exchanges/bithumb/bithumb.go
@@ -178,36 +178,15 @@ func (b *Bithumb) GetAllTickers() (map[string]Ticker, error) {
 
 	result := make(map[string]Ticker)
 	for k, v := range response.Data {
-		data, ok := v.(map[string]interface{})
-		if !ok {
+		if k == "date" {
 			continue
 		}
-
-		openingPrice, _ := data["opening_price"].(string)
-		closingPrice, _ := data["closing_price"].(string)
-		minPrice, _ := data["min_price"].(string)
-		maxPrice, _ := data["max_price"].(string)
-		unitsTraded, _ := data["units_traded"].(string)
-		accTradeValue, _ := data["acc_trade_value"].(string)
-		prevClosingPrice, _ := data["prev_closing_price"].(string)
-		unitsTraded24hr, _ := data["units_traded_24H"].(string)
-		accTradeValue24hr, _ := data["acc_trade_value_24H"].(string)
-		fluctateRate24hr, _ := data["fluctate_rate_24H"].(string)
-
-		var t Ticker
-		t.OpeningPrice, _ = strconv.ParseFloat(openingPrice, 64)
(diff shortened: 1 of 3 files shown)


exchanges/bithumb/bithumb.go
```diff
@@ -173,32 +182,41 @@ func (b *Bithumb) GetAllTickers() (map[string]Ticker, error) {
...
 
-	if response.Status != noError {
+	if response.Status != "0000" {
 		return nil, errors.New(response.Message)
 	}
 
 	result := make(map[string]Ticker)
 	for k, v := range response.Data {
-		if k == "date" {
+		data, ok := v.(map[string]interface{})
```
gloriousCode: This can be rewritten much more simply:
```
type Response struct {
		ActionStatus
		Data map[string]json.RawMessage
	}

	response := Response{}
	path := fmt.Sprintf("%s%s%s", b.APIUrl, publicTicker, "all")

	err := b.SendHTTPRequest(path, &response)
	if err != nil {
		return nil, err
	}

	if response.Status != "0000" {
		return nil, errors.New(response.Message)
	}

	result := make(map[string]Ticker)
	for k, v := range response.Data {
		var helloMoto Ticker
		err := common.JSONDecode(v, &helloMoto)
		if err != nil {
			log.Error(err)
		}
		result[k] = helloMoto
	}
	return result, nil
```

exchanges/bithumb/bithumb.go
```diff
@@ -143,19 +144,27 @@ func (b *Bithumb) GetTradablePairs() ([]string, error) {
...
 
 	err := b.SendHTTPRequest(path, &response)
 	if err != nil {
-		return response, err
+		return response.Data, err
 	}
 
-	if response.Status != noError {
-		return response, errors.New(response.Message)
+	if response.Status != "0000" {
```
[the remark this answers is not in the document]
shazbert: The API was returning ints instead of strings for a passing moment last night that's why I changed, will change back because strings are being returned now.

exchanges/bithumb/bithumb.go
gloriousCode: Perhaps having an if for the date parsing, this way still reduces the complexity and doesn't ignore error checking
```if k == "date" {
// continue because its not used
} else {
 // ticker parse
}
```

shazbert: That's already handled in this because v.(type) == string
```go
data, ok := v.(map[string]interface{})
		if !ok {
			continue
		}
```

shazbert: Okay I updated to what you suggested

PR #343, opened 2019-08-26
Fix unnecessary log.Fatals and os.Exits

thrasher-:
# Description

Cleans up unnecessary log.Fatals around the codebase where neccessary. There still exists some for exchange loading via Setup but these have all been fixed in engine branch.

Also moves the os.Exit(1) if number of loaded exchanges == 0 from SetupExchanges to main, so developers don't have to look through the code too far if they find out their programs exits on them unexpectedly. 

## Type of change

# How Has This Been Tested?

go test ./... -race

# Checklist:

### DIFF
--- a/communications/telegram/telegram.go
+++ b/communications/telegram/telegram.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"net/http"
+	"time"
 
 	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/communications/base"
@@ -43,9 +44,16 @@ const (
 	talkRoot = "GoCryptoTrader bot"
 )
 
+var (
+	// ErrWaiter is the default timer to wait if an err occurs
+	// before retrying after successfully connecting
+	ErrWaiter = time.Second * 30
+)
+
 // Telegram is the overarching type across this package
 type Telegram struct {
 	base.Base
+	initConnected     bool
 	Token             string
 	Offset            int64
 	AuthorisedClients []int64
@@ -64,6 +72,8 @@ func (t *Telegram) Connect() error {
 	if err := t.TestConnection(); err != nil {
 		return err
 	}
+
+	log.Debugln("Telegram: Connected successfully!")
 	t.Connected = true
 	go t.PollerStart()
 	return nil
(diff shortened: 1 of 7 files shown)


communications/telegram/telegram.go
```diff
@@ -43,9 +44,16 @@ const (
 	talkRoot = "GoCryptoTrader bot"
 )
 
+var (
+	// ErrWaiter is the default timer to wait if an err occurs
+	// before retrying after successfully connecting
+	ErrWaiter = time.Second * 30
```
gloriousCode: I don't think this needs to be exported

currency/storage.go
```diff
@@ -116,15 +116,20 @@ func (s *Storage) RunUpdater(overrides BotOverrides, settings *MainConfiguration
...
 			Name:        settings.CryptocurrencyProvider.Name,
 			Enabled:     true,
 			AccountPlan: settings.CryptocurrencyProvider.AccountPlan,
 			APIkey:      settings.CryptocurrencyProvider.APIkey,
 			Verbose:     settings.CryptocurrencyProvider.Verbose,
 		})
-
-		s.currencyAnalysis = c
+		if err != nil {
+			log.Errorf("Unable to setup CoinMarketCap analysis. Error: %s", err)
```
gloriousCode: Golang loves lowercase starting error messages when it isn't a acronym or pronoun

main.go
```diff
@@ -124,6 +124,9 @@ func main() {
 	log.Debugf("Global HTTP request timeout: %v.\n", common.HTTPClient.Timeout)
 
 	SetupExchanges()
+	if len(bot.exchanges) == 0 {
+		log.Fatal("No exchanges were able to be loaded. Exiting")
```
gloriousCode: Same here, but its a fatal, and its more go having the issue, not me :D

currency/storage.go
shazbert: https://github.com/golang/go/wiki/Errors; for ref, it does say that, but seeing as this is used as a prefix for the err, all that is needed is the underlying exported error type should be lower case?

xtda: To add a little more

The reason behind this comes from https://github.com/golang/go/wiki/CodeReviewComments#error-strings

Normally you would pass the error down the chain until its finally displayed

so you may end up with something 
```go
return fmt.Errorf("Something bad")
```

then later down the track
```go 
log.Errorf("Some error: %v", err)
```

That will cause it to display 
```
Some error: Something bad
```

if you are just logging the output it's fine but Goland code detection does not pick up on this and just warns either way (same with some other linters)

communications/telegram/telegram.go
thrasher-: This is intentionally done for customisability is engine and for standalone apps

currency/storage.go
thrasher-: This is just the prefix of displaying an error. All throughout our code base we use capital letters for the initial prefix of things and errors being lower case as per go spec, that's why our linters don't mind.

gloriousCode: An issue for the engine branch then 🦀 😄

currency/coinmarketcap/coinmarketcap_test.go
```diff
@@ -41,7 +41,14 @@ func TestSetup(t *testing.T) {
 	cfg.Enabled = true
 	cfg.AccountPlan = "basic"
 
-	c.Setup(cfg)
+	if err := c.Setup(cfg); err != nil {
+		t.Error(err)
+	}
+
+	cfg.AccountPlan = "meow"
```
shazbert: :100:

PR #347, opened 2019-08-30
(Engine) Logger - fixes issues with multiwriter go routine 

xtda:
# Description

This PR Fixes an issue with unbuffered channels and multiwriter

Also some general improvements and added some comments

_please ignore the nolint in sublogger_types.go I didn't want to comment 16 times_
```go
        // ConnectionMgr sublogger does stuff and things
	ConnectionMgr    *subLogger
        // CommunicationMgr sublogger does stuff and things
	CommunicationMgr *subLogger
```
## Type of change

# How Has This Been Tested?

Automated testing using go test ./..
Manual testing of application running
pprof monitoring

# Checklist:

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -34,11 +34,15 @@ func (l *Logger) newLogEvent(data, header string, w io.Writer) error {
 	if data == "" || data[len(data)-1] != '\n' {
 		e.data = append(e.data, '\n')
 	}
-	e.output.Write(e.data)
+	_, err := e.output.Write(e.data)
+	if err != nil {
+		fmt.Printf("Log Write failed: %v\n", err)
+	}
+
 	e.data = e.data[:0]
 	eventPool.Put(e)
 
-	return nil
+	return err
 }
 
 // CloseLogger is called on shutdown of application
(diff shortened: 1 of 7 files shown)


logger/logger.go
```diff
@@ -34,11 +34,15 @@ func (l *Logger) newLogEvent(data, header string, w io.Writer) error {
 	if data == "" || data[len(data)-1] != '\n' {
 		e.data = append(e.data, '\n')
 	}
-	e.output.Write(e.data)
+	_, err := e.output.Write(e.data)
+	if err != nil {
+		fmt.Printf("Log Write failed: %v\n", err)
```
thrasher-: We return an err if io.Writer is nil but we currently don't act on it in any of the calling funcs, but report on this one (as well as returning it). Could simplify this by reporting both issues in the calling func if err != nil

xtda: Okay made the change

I pass the error back down from newLogEvent to each sub logger and handle the display in there
there is not much that can be done so at the moment all it does is display the error message

I was unsure on fmt v log to display the message so i went with standard logger to at least include timestamp etc on the chance that the gct logger fails to write

PR #371, opened 2019-10-25
Bugfix: Unlocking an unlocked mutex PANIC + Increase dispatcher job capacity via commandline

gloriousCode:
image
> Meditation can help prevent panics

# Description
## Unlock of an unlocked mutex
A bug in `request.go` where there is a chance where a timeout  can occur which calls an unlock of a mutex, followed by a sending a message to an empty channel, then calling unlock again, which causes a panic.

The purpose of this request timer is to ensure that requests are always done one at a time and in order. Poloniex is very strict about their nonce values, so ensuring that everything is sent correctly is very important. 

There is an underlying assumption that something can go wrong that the requester does not know about and so implements a timer for the event where between calling `GetNonce()` and `SendPayload()` there is an error. If that timer runs out, it will unlock the mutex to allow the next nonce to be generated and sent away. For 99.999% of cases, the timeout is never hit, so there is rarely an issue, but when it does, it causes a panic.

Upon inspection, there is little chance for errors to occur between these two points. ~~So I've removed the timer and just call `x.Requester.FifoUnlock()` for each instance of an error between the two points to ensure the next request can continue. This does come with an additional burden on exchange implementations though, so I've added comments to `GetNonce` and `FifoUnlock` to explain. I understand though that increasing the burden may get this PR rejected though.~~

I have updated this to have a concept of a timed mutex. This maintains the current mindset of errors potentially happening outside of the current context, but prevents panics from happening via unlocking an unlocked mutex.

I also move request types to `request_types.go` 🦀 

## Dispatcher capacity
The second issue is when you enable many many currencies, the `dispatch.go` dispatcher will start disposing of jobs since it has a capacity of 100. For example, enabling 150 currencies for EXMO will reach the limit of the channel before jobs can effectively be dismissed, causing jobs to be taking out the back to the garbage collector. 

~~It's a bit of a temporary fix though, if we enable all currencies we might get a bunch of updates thrown out. We'll need to look into speeding up dispatcher checks or moving checks earlier to prevent them from being thrown out too soon.~~

I've created a new command line function to increase the job buffer limit. Its a much less resource intensive way of handling more jobs than max workers can handle in time.

## Type of change

# How Has This Been Tested?
Running the application with varying timeouts and channel capacities to trigger each bug.
Then doing the same thing after applying my fixes to observe that they do not get hit anymore.
`go test ./... -race` and `golangci-lint run`

# Checklist:

### DIFF
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -257,7 +257,7 @@ func (d *Dispatcher) publish(id uuid.UUID, data interface{}) error {
 	select {
 	case d.jobs <- newJob:
 	default:
-		return fmt.Errorf("dispatcher buffer at max capacity [%d] current worker count [%d], spawn more workers via --dispatchworkers=x"+
+		return fmt.Errorf("dispatcher jobs at limit [%d] current worker count [%d]. Spawn more workers via --dispatchworkers=x"+
 			", or increase the jobs limit via --dispatchjobslimit=x",
 			len(d.jobs),
 			atomic.LoadInt32(&d.count))
--- a/dispatch/dispatch.go
+++ b/dispatch/dispatch.go
@@ -14,7 +14,6 @@ import (
 func init() {
 	dispatcher = &Dispatcher{
 		routes: make(map[uuid.UUID][]chan interface{}),
-		jobs:   make(chan *job, DefaultJobBuffer),
 		outbound: sync.Pool{
 			New: func() interface{} {
 				// Create unbuffered channel for data pass
@@ -25,14 +24,14 @@ func init() {
 }
 
 // Start starts the dispatch system by spawning workers and allocating memory
-func Start(workers int) error {
+func Start(workers int, jobBuffer int) error {
 	if dispatcher == nil {
 		return errors.New(errNotInitialised)
 	}
 
 	mtx.Lock()
 	defer mtx.Unlock()
-	return dispatcher.start(workers)
+	return dispatcher.start(workers, jobBuffer)
 }
 
 // Stop attempts to stop the dispatch service, this will close all pipe channels
@@ -78,7 +77,7 @@ func SpawnWorker() error {
 
 // start compares atomic running value, sets defaults, overides with
 // configuration, then spawns workers
-func (d *Dispatcher) start(workers int) error {
+func (d *Dispatcher) start(workers int, channelCapacity int) error {
 	if atomic.LoadUint32(&d.running) == 1 {
 		return errors.New(errAlreadyStarted)
 	}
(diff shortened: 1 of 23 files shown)


thrasher-:
The other thing I'd suggest for the dispatcher default limit is that 150 currencies is quite a lot (most people may enable 4 or so per exchange) and in that case the user can adjust the cmd line param -dispatchworkers (the intended purpose of why its there) to increase it without adjusting the default limit impacting every exchange. 1000 goroutines is quite excessive as a default setting which would lead to thread contention on some systems.

dispatch/dispatch_types.go
```diff
@@ -9,7 +9,7 @@ import (
 
 const (
 	// DefaultJobBuffer defines a maxiumum amount of jobs allowed in channel
-	DefaultJobBuffer = 100
+	DefaultJobBuffer = 1000
```
shazbert: I am having no issues with reverting this one back on linux.

shazbert:
Elegant. That's a tACK for me, only issue is the job channel buffer. Keeping it tighter is better to expose issues. But not a merge issue.

dispatch/dispatch_types.go
gloriousCode: Yeah I'll be making this a config var so people can change it permanently if they enable many many currencies

dispatch/dispatch.go
```diff
@@ -88,7 +87,12 @@ func (d *Dispatcher) start(workers int) error {
 			"Dispatcher: workers cannot be zero using default values")
 		workers = DefaultMaxWorkers
 	}
-
+	if channelCapacity < 1 {
+		log.Warn(log.DispatchMgr,
+			"Dispatcher: jobs limit cannot be zero using default values")
```
thrasher-: Please add a comma after zero

exchanges/orderbook/orderbook_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	err := dispatch.Start(1)
+	err := dispatch.Start(1, 0)
```
thrasher-: Can set this to dispatch.DefaultJobsLimit, so the check < 0 won't trigger a warning and that the value is set to defaults anyway

exchanges/ticker/ticker_test.go
```diff
@@ -15,7 +15,7 @@ import (
 )
 
 func TestMain(m *testing.M) {
-	err := dispatch.Start(1)
+	err := dispatch.Start(1, 0)
```
thrasher-: Can set this to dispatch.DefaultJobsLimit, so the check < 0 won't trigger a warning and that the value is set to defaults anyway

PR #386, opened 2019-11-27
Fix Docker os.Rename invalid cross-device link issue

thrasher-:
# Description

Adds a new file.Move func for moving files as recent testing has shown that docker volumes and Golang trigger a "Invalid cross-device link" error when using os.Rename as found here: https://github.com/thrasher-corp/gocryptotrader/pull/382#discussion_r351078142

## Type of change

# How Has This Been Tested?

go test ./... -race and golangci-lint

# Checklist:

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -5,7 +5,7 @@ import (
 	"io/ioutil"
 	"log"
 
-	gctfile "github.com/thrasher-corp/gocryptotrader/common/file"
+	"github.com/thrasher-corp/gocryptotrader/common/file"
 	"github.com/thrasher-corp/gocryptotrader/config"
 )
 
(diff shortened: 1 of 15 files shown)


thrasher-:
> tACK with docker and it works! 🎉 Thanks for writing those tests 🎉 As discussed, this fix does highlight that there is a nonce generation issue down the line, but it is out of scope here.

Correct, will be introducing a fix for that in a separate PR

thrasher-:
> tACK - Awesome! Only thing I would suggest is standardising gctfile/file package name; have one or the other.

Done! Originally did this because it would clash with local var names, but have pushed a commit to make it consistent

common/file/file.go
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	outputFile, err := os.Create(destPath)
+	if err != nil {
+		inputFile.Close()
+		return err
+	}
+	defer outputFile.Close()
+
+	_, err = io.Copy(outputFile, inputFile)
```
xtda: Is it worth checking the length written here to make sure it matches up with input file? avoid possible corruption

common/file/file.go
```diff
@@ -0,0 +1,39 @@
...
+	inputFile.Close()
+	if err != nil {
```
xtda: If for whatever reason the copy fails should we also remove the file it created above?

common/file/file.go
thrasher-: io.Copy does that internally. (nr != nw) check will throw an ErrShortWrite (plus additional checks along the way like err when writing to the dest and also when reading the src)

common/file/file.go
thrasher-: Yeah we should, good idea

cmd/config/config.go
```diff
@@ -5,7 +5,7 @@ import (
 	"io/ioutil"
```
nitsuaaisblink, line 5: cmd/config/config.go

PR #394, opened 2019-12-10
(BTC Markets): Wrapper SubmitOrder parameter order fix & IsOrderPlaced condition correction

xtda:
# PR Description

Corrects parameter order for BTCMarkets SubmitOrder() wrapper to NewOrder() Method
(OrderType & OrderSide were incorrect)

Also fixes the condition check on IsOrderPlaced in order manager Submit() Method that will cause it to return order cannot be placed error if IsOrderPlaced is true

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


thrasher-:
Just realised that there's another bug: https://github.com/thrasher-corp/gocryptotrader/blob/master/exchanges/btcmarkets/btcmarkets_wrapper.go#L547

This will return all orders (inactive/cancelled etc) so we can specify "open" as the last param to retrieve only open orders

xtda:
> Just realised that there's another bug: https://github.com/thrasher-corp/gocryptotrader/blob/master/exchanges/btcmarkets/btcmarkets_wrapper.go#L547
> 
> This will return all orders (inactive/cancelled etc) so we can specify "open" as the last param to retrieve only open orders

Addressed this and some other issues with GetActiveOrders() 

At quick glance GetOrderHistory() also needs to be reworked will push through another update soon and get you to rereview it

common/common.go
```diff
@@ -340,3 +340,17 @@ func ChangePermission(directory string) error {
 		return nil
 	})
 }
+
+// SplitStringSliceByLimit splits a slice of strings into slices by input limit and returns a slice of slice of strings ¯\_(ツ)_/¯
+func SplitStringSliceByLimit(in []string, limit int) [][]string {
```
thrasher-: Please add a test for this

exchanges/btcmarkets/btcmarkets_wrapper.go
```diff
@@ -532,30 +536,45 @@ func (b *BTCMarkets) GetFeeByType(feeBuilder *exchange.FeeBuilder) (float64, err
...
 			tempResp.CurrencyPair = req.Currencies[x]
+			tempResp.ID = tempData[y].OrderID
 			tempResp.OrderSide = order.Bid
 			if tempData[y].Side == ask {
 				tempResp.OrderSide = order.Ask
 			}
 			tempResp.OrderDate = tempData[y].CreationTime
+
+			switch tempData[y].Type {
+			case "Limit":
```
thrasher-: Can use consts: https://github.com/thrasher-corp/gocryptotrader/blob/86c04d96be6b5da7baaa91b31c89ada92d902846/exchanges/btcmarkets/btcmarkets.go#L64

common/common.go
xtda: Test coverage has been added for function

exchanges/btcmarkets/btcmarkets_wrapper.go
xtda: Done

ntpclient/ntpclient.go
```diff
@@ -38,7 +37,10 @@ func NTPClient(pool []string) (time.Time, error) {
 
 		defer con.Close()
 
-		con.SetDeadline(time.Now().Add(5 * time.Second))
+		err = con.SetDeadline(time.Now().Add(5 * time.Second))
```
thrasher-: Please change to 
```
if err := con.SetDeadline(time.Now().Add(5 * time.Second)); err != nil {
	log.Warnf(log.TimeMgr, "Unable to SetDeadline. Error: %s\n", err)
        con.Close()
	continue
}
```

The other thing I noticed with this is we are deferring con.Close() but on each for loop iteration, we are reassigning con. The defer will only execute upon func exit, so it's possible for lingering cons

ntpclient/ntpclient.go
```diff
@@ -55,5 +57,6 @@ func NTPClient(pool []string) (time.Time, error) {
 
 		return time.Unix(int64(secs), nanos), nil
 	}
-	return time.Unix(0, 0), errors.New("no valid time servers")
+	log.Warnln(log.Global, "No valid ntp server found returning current system time")
```
thrasher-: Please use log.TimeMgr here, and just a basic grammar fix:

`No valid NTP servers found, using current system time`

ntpclient/ntpclient.go
```diff
@@ -30,30 +29,36 @@ type ntppacket struct {
 // NTPClient create's a new NTPClient and returns local based on ntp servers provided timestamp
 func NTPClient(pool []string) (time.Time, error) {
```
thrasher-: Since we don't use err, we could do a check at the top to check if len(pool) == 0 and complain if it's empty

exchanges/btcmarkets/btcmarkets_wrapper.go
```diff
@@ -532,30 +536,45 @@ func (b *BTCMarkets) GetFeeByType(feeBuilder *exchange.FeeBuilder) (float64, err
 
 // GetActiveOrders retrieves any orders that are active/open
 func (b *BTCMarkets) GetActiveOrders(req *order.GetOrdersRequest) ([]order.Detail, error) {
-	var resp []order.Detail
-	var tempResp order.Detail
-	var tempData []OrderData
 	if len(req.Currencies) == 0 {
-		allPairs := b.GetEnabledPairs(asset.Spot)
+		allPairs := b.GetAvailablePairs(asset.Spot)
```
thrasher-: We should keep this as GetEnabledPairs, so that we're only interested in the ones the user has enabled (otherwise they can set it in req.Currencies). Our order manager will use enabled pairs in a future update.

ntpclient/ntpclient.go
xtda: I will remove the error return value 
a pool should never be empty because of https://github.com/thrasher-corp/gocryptotrader/blob/master/config/config.go#L1250

ntpclient/ntpclient_test.go
```diff
@@ -1,25 +1,16 @@
...
-	if err == nil {
-		t.Errorf("Expected error")
-	}
+	v := NTPClient(pool)
 
-	firstInvalid := []string{"pool.thisisinvalid.org", "pool.ntp.org:123", "0.pool.ntp.org:123"}
-	_, err = NTPClient(firstInvalid)
-	if err != nil {
-		t.Errorf("failed to get time %v", err)
+	if reflect.TypeOf(v) != reflect.TypeOf(time.Time{}) {
```
shazbert: This will always be a time.Time so can RM this check. Or could force a time.Now().UTC() check?

xtda: Honestly this was just a lazy way to maintain test coverage to at least know its returning a time.Now()

I can do a time.Now.UTC() check but i would have to also parse the time into a format that doen't include anything above milliseconds or return value of NTPClient and time.Now.UTC() will be out and cause a failure

common/common_test.go
```diff
@@ -499,3 +500,30 @@ func TestChangePermission(t *testing.T) {
 		}
 	}
 }
+
+func initStringSlice(t *testing.T, size int) (out []string) {
+	t.Helper()
```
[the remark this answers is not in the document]
xtda: Helper() is called here to cause this function to be ignored when printing out warnings/errors/line information

ntpclient/ntpclient_test.go
thrasher-: Another one is just comparing the year/day/month is the same, that's a good enough check

xtda: I kept the type check because you can never have to much test coverage right :D?
Also added a check for IsZero() and using 2006-01-02T15:04:05 format

common/common_test.go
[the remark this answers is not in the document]
shazbert: It actually shifts the error to the calling function line if it has a t.Error or t.Fatal, it doesn't add anything to the output. 

for example:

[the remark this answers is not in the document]
shazbert:
```
func initStringSlice(t *testing.T, size int) (out []string) {
	t.Helper()
	for x := 0; x < size; x++ {
		out = append(out, "gct-"+strconv.Itoa(x))
	}
	t.Error("Cats")
	return
}
```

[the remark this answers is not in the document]
shazbert: Will indicate error at ln 514 instead of ln 509

PR #399, opened 2019-12-16
(BTC Markets) Ticker batching support

xtda:
# PR Description

Enables ticker batching support for BTC Markets using the [Get Multiple Tickers](https://api.btcmarkets.net/doc/v3#tag/Market-data-APIs/paths/~1v3~1markets~1tickers/get) endpoint

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/exchanges/btcmarkets/btcmarkets_test.go
+++ b/exchanges/btcmarkets/btcmarkets_test.go
@@ -17,7 +17,7 @@ var b BTCMarkets
 const (
 	apiKey                  = ""
 	apiSecret               = ""
-	canManipulateRealOrders = true
+	canManipulateRealOrders = false
 	BTCAUD                  = "BTC-AUD"
 	LTCAUD                  = "LTC-AUD"
 	ETHAUD                  = "ETH-AUD"
--- a/exchanges/btcmarkets/btcmarkets_test.go
+++ b/exchanges/btcmarkets/btcmarkets_test.go
@@ -94,7 +94,6 @@ func TestGetMarketCandles(t *testing.T) {
 
 func TestGetTickers(t *testing.T) {
 	t.Parallel()
-	b.Verbose = true
 	temp := currency.NewPairsFromStrings([]string{LTCAUD, BTCAUD})
 	_, err := b.GetTickers(temp)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


exchanges/btcmarkets/btcmarkets_test.go
```diff
@@ -94,7 +94,8 @@ func TestGetMarketCandles(t *testing.T) {
 
 func TestGetTickers(t *testing.T) {
 	t.Parallel()
-	temp := []string{BTCAUD, LTCAUD, ETHAUD}
+	b.Verbose = true
+	temp := currency.NewPairsFromStrings([]string{"BTC-AUD", "LTC-AUD"})
```
MadCozBadd: const em. pls dont give ryan the pleasure

exchanges/btcmarkets/btcmarkets_test.go
```diff
@@ -94,7 +94,8 @@ func TestGetMarketCandles(t *testing.T) {
 
 func TestGetTickers(t *testing.T) {
 	t.Parallel()
-	temp := []string{BTCAUD, LTCAUD, ETHAUD}
+	b.Verbose = true
```
shazbert: rm

exchanges/btcmarkets/btcmarkets_wrapper.go
```diff
@@ -546,7 +547,7 @@ func (b *BTCMarkets) GetActiveOrders(req *order.GetOrdersRequest) ([]order.Detai
 	}
 	var err error
 	for x := range req.Currencies {
-		tempData, err = b.GetOrders(b.FormatExchangeCurrency(req.Currencies[x], asset.Spot).String(), -1, -1, -1, "")
+		tempData, err = b.GetOrders(b.FormatExchangeCurrency(req.Currencies[x], asset.Spot).String(), -1, -1, -1, "open")
```
shazbert: const "open" and change string in get all orders

xtda: consting provides no real benefit (folding, propagation etc, sure its immutable but you aren't changing the value of open at run time and since its a const and passed to a method its always passed by value anyway)

Since GetOrders can only ever be open or all (or blank that defaults to all) it might even be worth while having this as a bool for openOnly over a string then GetOrders can just be 
```go
	if openOnly {
		params.Set("status", "open")
	}
```

what are your thoughts on this? @shazbert @thrasher- @MadCozBadd

shazbert: I am happy with a bool, I also enjoy a good consting.

xtda: I am happy either way both have benefits

Keeping it a string keeps it inline with what their API expects (and if they ever add other options in future its easier to add)

Having a bool is less error prone and less bytes passed across

MadCozBadd: bool seems good imo

xtda: pushed up a change to move this over to a bool
By default status is not sent all records are returned so take advantage of that

PR #409, opened 2020-01-06
Withdraw additional functionality (validation/submission/tracking)

xtda:
# PR Description

This PR adds additional features to the withdraw functionality to GCT

Current working and planned features:

- [x] Validation for Crypto & Fiat Requests
- [x] New withdraw request submission manager in engine
~~- [ ] monitor and track withdrawal request life cycle and update correct portfolio managers~~
**_API limitations on exchanges make this currently not feasible_**
- [x] Database tables and tracking for created withdrawal requests
- [x] Ability to track & search previous requests via CLI
- [x] submit withdrawal request via CLI (for both crypto and fiat)

## Type of change

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally and on Travis with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #383, opened 2019-11-24
Engine: Scripting support

xtda:
# Description

This PR adds in scripting functionality into GCT 

Current working features:

- Execute scripts
- Terminate scripts
- Autoload scripts on bot startup
- Current Exchange features supported:
- Enabled Exchanges
- Enabled currency pairs
- Account information
- Query Order
- Submit Order
- Cancel Order
- Ticker
- Orderbook

A [README](https://github.com/xtda/gocryptotrader/blob/gctscript/gctscript/README.md) has been created to go over usage

***PR is not completely finished yet but wanted to gauge early feedback as change count is getting high***

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

# How Has This Been Tested?

Automated tests 
Manual testing

Test scripts have also been provided in ./testdata/gctscript/

# Checklist:

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool 
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally and on Travis with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #420, opened 2020-01-22
(COMMON): Basic (Least Recently Used) LRU caching system

xtda:
# PR Description

This PR adds a basic least recently used cache system that will be used as part of #409 (and others parts if people find it useful)

But as I learned my lesson with #383 and its 15,000 changes I decided to break it up into smaller PR's for additional non-directly related functionality 

[LRU](https://en.wikipedia.org/wiki/Cache_replacement_policies#Least_recently_used_(LRU)) is a type of caching style that stores elements and moves the most recently accessed to the front and disgards the least used if it would grow beyond capacity

There is both a concurrent safe( with mutex)
```
func New(capacity uint64) *LRUCache {
	return &LRUCache{
		lru: NewLRUCache(capacity),
	}
}
```
or called directly for non-concurrent safe version:
```
func NewLRUCache(capacity uint64) *LRU {
	return &LRU{
		Cap:   capacity,
		List:  list.New(),
		Items: make(map[interface{}]*list.Element),
	}
}
```

Obviously up to user to decide on requirements as if they need it to be goroutine safe or not and performance of mutexs

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


xtda:
Also I have not added the ability to "Peek" (get a keys value without moving it to the front of the cache) or List all entires as I personally feel this goes against the point of of a LRU cache as you are accessing it and it was recently used 

Happy to add if people want this functionality

common/cache/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+
+[![Build Status](https://travis-ci.org/thrasher-corp/gocryptotrader.svg?branch=master)](https://travis-ci.org/thrasher-corp/gocryptotrader)
+[![Software License](https://img.shields.io/badge/License-MIT-orange.svg?style=flat-square)](https://github.com/thrasher-corp/gocryptotrader/blob/master/LICENSE)
+[![GoDoc](https://godoc.org/github.com/thrasher-corp/gocryptotrader?status.svg)](https://godoc.org/github.com/thrasher-corp/gocryptotrader/portfolio)
+[![Coverage Status](http://codecov.io/github/thrasher-corp/gocryptotrader/coverage.svg?branch=master)](http://codecov.io/github/thrasher-corp/gocryptotrader?branch=master)
+[![Go Report Card](https://goreportcard.com/badge/github.com/thrasher-corp/gocryptotrader)](https://goreportcard.com/report/github.com/thrasher-corp/gocryptotrader)
+
+
+This database package is part of the GoCryptoTrader codebase.
```
thrasher-: cache package

common/cache/README.md
```diff
@@ -0,0 +1,67 @@
...
+
+##### Basic Usage:
+
+```go
+package main
+
+import ("github.com/thrasher-corp/gocryptotrader/common/cache")
+
+func main() {
+	lruCache := New(5)
```
thrasher-: cache.New

common/cache/cache_types.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"container/list"
+	"sync"
+)
+
+// LRUCache thread safe fixed size LRU cache
+type LRUCache struct {
+	lru *LRU
+	m   sync.RWMutex
```
thrasher-: We don't use RLock/RUnlock so a sync.Mutex could be substituted here

common/cache/README.md
xtda: done :D

common/cache/README.md
xtda: done :D

common/cache/cache_types.go
xtda: Ah yeah was originally looking at doing read locks for some of the calls (Get & Contains) but turns out didn't really fit in the end swapped back to normal mutex

common/cache/cache_test.go
```diff
@@ -0,0 +1,133 @@
...
+func TestNewLRUCache_ContainsOrAdd(t *testing.T) {
+	lruCache := New(5)
+
+	f := lruCache.ContainsOrAdd("hello", "world")
+	if f {
+		t.Fatal("expected ContainsOrAdd() to add new key when not found")
+	}
+
+	f = lruCache.ContainsOrAdd("hello", "world")
+	if !f {
```
gloriousCode: The minorinoist of minor nits
If `f` which equals `true` is `false`, then `fatal` :)
You could remove the funnyness with just `if lruCache.ContainsOrAdd("hello", "world") {` and `if !lruCache.ContainsOrAdd("hello", "world") {`

common/cache/cache_test.go
```diff
@@ -0,0 +1,133 @@
...
+		t.Fatal("expected \"hello\" key to be removed from cache")
+	}
+
+	v = lruCache.Get("hello")
+	if v != nil {
+		t.Fatal("expected cache to not contain \"hello\" key")
+	}
+}
+
+func TestNewLRUCache_ContainsOrAdd(t *testing.T) {
```
gloriousCode: A similarly minor one `TestContainsOrAdd` would be enough so we don't need underscores

common/cache/cache_test.go
```diff
@@ -0,0 +1,133 @@
...
+	lruCache.Add(3, 3)
+	if lruCache.Len() != 2 {
+		t.Fatal("expected cache to have 2 entries")
+	}
+
+	v := lruCache.Get(1)
+	if v != nil {
+		t.Fatal("expected cache to no longer contain \"1\" key")
+	}
+	v = lruCache.Get(2)
```
gloriousCode: Could you add a check here that `Get(2)` brings 2 to the newest?

common/cache/lru.go
```diff
@@ -0,0 +1,97 @@
...
+func (l *LRU) Get(key interface{}) interface{} {
+	if i, f := l.items[key]; f {
+		l.l.MoveToFront(i)
+		return i.Value.(*item).value
+	}
+	return nil
+}
+
+// GetOldest returns the oldest entry
+func (l *LRU) GetOldest() (key, value interface{}) {
```
gloriousCode: Should GetOldest and Newest be exported? Not a nit, but I can't think of a use case that isn't for testing/internal use. Especailly since GetOldest doesn't bring it to the front which kinda circumvents LRU

common/cache/cache_test.go
xtda: Default auto complete names of goland like to do this was lazy

common/cache/cache_test.go
xtda: Done

common/cache/lru.go
xtda: unexported both as they are not required and really just there for testing

common/cache/cache_test.go
xtda: what are you not happy with
f being true & t being false :D

common/cache/cache.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Contain checks if cache contains key
+func (l *LRUCache) Contains(key interface{}) bool {
+	l.m.Lock()
+	defer l.m.Unlock()
+	return l.lru.Contains(key)
+}
+
+func (l *LRUCache) Remove(key interface{}) bool {
```
thrasher-: Please add a comment for exported func

common/cache/cache.go
```diff
@@ -0,0 +1,74 @@
...
+	l.m.Lock()
+	defer l.m.Unlock()
+	if l.lru.Contains(key) {
+		return true
+	}
+	l.lru.Add(key, value)
+	return false
+}
+
+// Contain checks if cache contains key
```
thrasher-: // Contains

common/cache/cache.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Get looks up a key's value from the cache.
+func (l *LRUCache) getNewest() (key, value interface{}) {
+	l.m.Lock()
+	defer l.m.Unlock()
+	return l.lru.getNewest()
+}
+
+// Contain checks if cache contains key if not adds to cache
```
thrasher-: // ContainsOrAdd checks if cache contains key and if not, adds it to the cache

common/cache/cache.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// GetOldest looks up old key's value from the cache.
+func (l *LRUCache) getOldest() (key, value interface{}) {
+	l.m.Lock()
+	defer l.m.Unlock()
+	return l.lru.getOldest()
+}
+
+// Get looks up a key's value from the cache.
```
thrasher-: This returns the newest k,v from the LRU

PR #437, opened 2020-02-05
General engine improvements

thrasher-:
# PR Description

Generic engine improvements:

- Exchange loading/unloading is protected by sync.Mutex, originally this wasn't and would crash the bot if you tried to enable/disable an exchange when it's iterating over Bot.Exchanges in a separate go routine
- Removed the ability to reload an already loaded exchange as configs are pointers, so it's not required once you adjust a config setting
- Adds gRPC TLS cert regeneration upon expiration (plus checking for it)
- Uses Go logger until the GCT logger is initialised, otherwise you miss out on config/engine loading errors
- Use file.Exists method to reduce the need to os.Stat all the things

Generic non-engine improvements:
- Makes the donation address a var
- Update documentation tool with missing items/contributors

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


engine/engine.go
```diff
@@ -297,63 +306,63 @@ func (e *Engine) Start() error {
...
 	if *e.Config.Logging.Enabled && strings.Contains(e.Config.Logging.Output, "file") {
-		log.Debugf(log.Global, "Using log file: %s\n",
-			filepath.Join(log.LogPath, e.Config.Logging.LoggerFileConfig.FileName))
+		gctlog.Debugf(gctlog.Global, "Using gctlog file: %s\n",
+			filepath.Join(gctlog.LogPath, e.Config.Logging.LoggerFileConfig.FileName))
 	}
-	log.Debugf(log.Global,
-		"Using %d out of %d logical processors for runtime performance\n",
+	gctlog.Debugf(gctlog.Global,
+		"Using %d out of %d gctlogical processors for runtime performance\n",
```
shazbert: gctlogical  -> logical

common/file/file.go
```diff
@@ -45,3 +52,9 @@ func Move(sourcePath, destPath string) error {
 
 	return os.Remove(sourcePath)
 }
+
+// Exists returns whether or not a file or path exists
+func Exists(name string) bool {
```
gloriousCode: Thanks for changing this to the positive 👍

engine/exchange.go
```diff
@@ -59,78 +64,100 @@ func dryrunParamInteraction(param string) {
...
+	err := e.removeExchange(exchangeName)
 	if err != nil {
 		return err
 	}
 
-	exchCfg.Enabled = false
-	err = Bot.Config.UpdateExchangeConfig(exchCfg)
+	exchCfg, err := Bot.Config.GetExchangeConfig(exchangeName)
 	if err != nil {
 		return err
```
gloriousCode: Should the exchange be re-added here if this fails? A failed unloadExchange puts gct in a limbo where the exchange is not loaded, but the config says its enabled. Not that I'd expect this to fail, but more a precaution. Otherwise, this puts the onus on the calling function to determine the type of error and adjust accordingly

engine/exchange_test.go
```diff
@@ -42,6 +39,51 @@ func CleanupTest(t *testing.T) {
...
+	t.Parallel()
+	var e exchangeManager
+	if err := e.removeExchange("Bitfinex"); err != ErrNoExchangesLoaded {
+		t.Error("no exchanges should be loaded")
+	}
+	bitfinex := new(bitfinex.Bitfinex)
+	bitfinex.SetDefaults()
+	e.add(bitfinex)
+	if err := e.removeExchange(testExchange); err != ErrExchangeNotFound {
+		t.Error("Bitstamp exchange should return an error")
```
gloriousCode: Bitfinex?

engine/exchange_test.go
```diff
@@ -42,6 +39,51 @@ func CleanupTest(t *testing.T) {
...
+	}
+	if err := e.removeExchange("BiTFiNeX"); err != nil {
+		t.Error("exchange should of been removed")
```
gloriousCode: should have

engine/exchange_test.go
thrasher-: Nope, we loaded Bitfinex and Bitstamp shouldn't be loaded

engine/exchange.go
thrasher-: That's a good point, I've changed the order of operations around so that both the config and removal have to pass before exchCfg.Enabled is set to false. Theoretically this should never be hit unless memory corruption occurs, since LoadExchange has to pass config checks also and add isn't exported

PR #438, opened 2020-02-05
Engine: SyncTimeout can't be set

vazha:
# PR Description

Program always uses the DefaultSyncerTimeout because c.SyncTimeout in engine/syncer.go newer set in code

Fixes # (issue)

## Type of change

## How has this been tested
Tested by build and run


shazbert:
need to run gofmt -w on engine.go linters broken because there is a tab between syncTimeout and settings var.

PR #444, opened 2020-02-12
(Logger) Rename package to log

xtda:
# PR Description

Just a quick small change that renames the logger package to "log" to stop the requirement of adding logger infront of the import that has annoyed me for awhile now :D

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -18,7 +18,7 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/dispatch"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/request"
 	gctscript "github.com/thrasher-corp/gocryptotrader/gctscript/vm"
-	gctlog "github.com/thrasher-corp/gocryptotrader/logger"
+	gctlog "github.com/thrasher-corp/gocryptotrader/log"
 	"github.com/thrasher-corp/gocryptotrader/portfolio"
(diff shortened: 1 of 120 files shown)


engine/engine.go
```diff
@@ -525,6 +525,6 @@ func (e *Engine) Stop() {
 	e.ServicesWG.Wait()
 	err := gctlog.CloseLogger()
 	if err != nil {
-		log.Printf("Failed to close logger. Error: %v\n", err)
+		log.Printf("Failed to close log. Error: %v\n", err)
```
thrasher-: logger

log/logger_setup.go
```diff
@@ -63,7 +63,7 @@ func GenDefaultSettings() (log Config) {
 func configureSubLogger(logger, levels string, output io.Writer) error {
 	found, logPtr := validSubLogger(logger)
 	if !found {
-		return fmt.Errorf("logger %v not found", logger)
+		return fmt.Errorf("log %v not found", logger)
```
thrasher-: sublogger (or logger) %v not found

engine/engine.go
xtda: :D vscode doesn't have as smart refactoring as Goland thought i picked up all these changes !

reverted this and the other instance

thrasher-: No problemo, that's why we review things 😉

PR #446, opened 2020-02-13
Feature: Websocket order handling

gloriousCode:
# Websocket order handling
ORDER
_Dramatic reenactment of websocket datahandler_

This is an update focused on websocket order update data handling. 
* It adds or updates exchange support for order submission and modification responses
* It synchronises those updates with the order manager
* It seperates websocket message handling from the `wsReadData` function to allow for easy websocket testing
* It adds testing for the majority websocket responses
* It removes `order.Detail` stuttering with things like `order.OrderStatus` being changed to `order.Status`
* It adds testing to a few `engine` packages such as `routines` and `orders`
* It maintains compatibility with websocket order submission via the wrapper
* Creates a fake exchange in the engine test package

## Things to consider
* I've not live tested each exchange with order updates. This testing is relying on exchange API documentation being accurate.... Which means there's a high chance that some response handling could fail. We'll need to test this in the future with real keys and real orders to ensure that it works 

## Type of change

## How has this been tested
* I've written over 400 unit tests. These tests cover websocket updates, websocket data handling, order manager and other parts of the `engine` package
* I've run the application to ensure no errors or panics
* `go test ./... -race`
* ` golangci-lint`

## Checklist


exchanges/gemini/gemini_websocket.go
```diff
@@ -165,96 +163,169 @@ func (g *Gemini) WsHandleData() {
...
+	return nil
+}
+
+func responseToStatus(response string) order.Status {
+	switch response {
+	case "accepted":
+		return order.New
+	case "booked":
+		return order.Active
+	case "filled":
```
thrasher-: https://docs.gemini.com/websocket-api/#filled

Please rename to "fill", we'll need to add a note here that an order is considered "filled" when the remaining_amount is 0

exchanges/hitbtc/hitbtc_websocket.go
```diff
@@ -72,50 +72,83 @@ func (h *HitBTC) WsHandleData() {
...
+	}
+}
+
+func (h *HitBTC) wsGetTableName(respRaw []byte) (string, error) {
+	var init capture
+	err := json.Unmarshal(respRaw, &init)
+	if err != nil {
+		return "", err
+	}
+	if init.Error.Code == 1002 {
```
thrasher-: Please make a const type for this (AuthFailure) to prevent the use of magic numbers

exchanges/hitbtc/hitbtc_websocket.go
```diff
@@ -72,50 +72,83 @@ func (h *HitBTC) WsHandleData() {
...
 
-func (h *HitBTC) handleSubscriptionUpdates(resp wshandler.WebsocketResponse, init capture) {
-	switch init.Method {
+func (h *HitBTC) wsHandleData(respRaw []byte) error {
+	name, err := h.wsGetTableName(respRaw)
+	if err != nil {
+		return err
+	}
+	switch name {
+	case "":
```
thrasher-: Good ol exchange API's 😆

exchanges/hitbtc/hitbtc_websocket.go
```diff
@@ -72,50 +72,83 @@ func (h *HitBTC) WsHandleData() {
...
+		return nil
 	case "ticker":
```
thrasher-: Please add TickerFetching to true for WS supported features

exchanges/hitbtc/hitbtc_websocket.go
```diff
@@ -72,50 +72,83 @@ func (h *HitBTC) WsHandleData() {
...
 		var wsTicker WsTicker
-		err := json.Unmarshal(resp.Raw, &wsTicker)
+		err := json.Unmarshal(respRaw, &wsTicker)
 		if err != nil {
-			h.Websocket.DataHandler <- err
-			return
+			return err
 		}
 		ts, err := time.Parse(time.RFC3339, wsTicker.Params.Timestamp)
```
thrasher-: Out of scope for this PR, but we can use time.Time for the wsTicker.Params.Timestamp for an efficiency gain

exchanges/kraken/kraken_test.go
```diff
@@ -737,3 +738,624 @@ func TestWsCancelOrder(t *testing.T) {
...
+}`)
+	err := k.wsHandleData(pressXToJSON)
+	if err != nil {
+		t.Error(err)
+	}
+}
+
+func TestWsSystemStatus(t *testing.T) {
+	pressXToJSON := []byte(`{
+  "connectionID": 1333337,
```
thrasher-: The API puts this at '8628615390848610000', can make the ConnectionID var use an int64

exchanges/kraken/kraken_websocket.go
```diff
@@ -164,278 +232,167 @@ func (k *Kraken) wsPingHandler() error {
...
 		}
-	} else {
-		k.Websocket.DataHandler <- errors.New(k.Name + " - Invalid own trades data")
+		return nil
 	}
+	return errors.New(k.Name + " - Invalid own trades data")
 }
 
-func (k *Kraken) wsProcessOpenOrders(ownOrders interface{}) {
+func (k *Kraken) wsProcessOpenOrders(ownOrders interface{}) error {
```
thrasher-: Please add GetOrders to the list of supported protocol features for WS

exchanges/okgroup/okgroup_types.go
```diff
@@ -1306,93 +1306,73 @@ type WebsocketEventResponse struct {
...
+		BestBid        float64 `json:"best_bid,string"`
+		BestBidSize    float64 `json:"best_bid_size,string"`
+		High24h        float64 `json:"high_24h,string"`
+		InstrumentID   string  `json:"instrument_id"`
+		Last           float64 `json:"last,string"`
+		LastQty        float64 `json:"last_qty,string"`
+		Low24h         float64 `json:"low_24h,string"`
+		Open24h        float64 `json:"open_24h,string"`
+		QuoteVolume24h float64 `json:"quote_volume_24h,string"`
+		Timestamp      string  `json:"timestamp"`
```
thrasher-: Judging by the tests, you can use time.Time here

exchanges/okgroup/okgroup_types.go
```diff
@@ -1306,93 +1306,73 @@ type WebsocketEventResponse struct {
...
-	Side    string  `json:"side,omitempty"`
-	Qty     float64 `json:"qty,string,omitempty"`
-	TradeID string  `json:"trade_id,omitempty"`
+	Table string `json:"table"`
+	Data  []struct {
+		Price        float64 `json:"price,string"`
+		Size         float64 `json:"size,string"`
+		InstrumentID string  `json:"instrument_id"`
+		Side         string  `json:"side"`
+		Timestamp    string  `json:"timestamp"`
```
thrasher-: Same with this

exchanges/okgroup/okgroup_types.go
```diff
@@ -1409,110 +1389,30 @@ type WebsocketUserSwapPositionHoldingData struct {
...
+		MarginTrading  string `json:"margin_trading"`
+		Notional       string `json:"notional"`
+		OrderID        string `json:"order_id"`
+		OrderType      string `json:"order_type"`
+		Price          string `json:"price"`
+		Side           string `json:"side"`
+		Size           string `json:"size"`
+		State          string `json:"state"`
+		Status         string `json:"status"`
+		Timestamp      string `json:"timestamp"`
```
thrasher-: Same for here

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -292,113 +233,174 @@ func (o *OKGroup) WsLogin() error {
...
+	if err == nil && eventResponse.Event != "" {
+		if eventResponse.Event == "login" {
+			o.Websocket.SetCanUseAuthenticatedEndpoints(eventResponse.Success)
+		}
+		if o.Verbose {
+			log.Debugf(log.ExchangeSys,
+				"WS Event: %v on Channel: %v for %s",
+				eventResponse.Event,
+				eventResponse.Channel,
+				o.Name)
```
thrasher-: To keep things consistent, it's good to put the exchange name first

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -292,113 +233,174 @@ func (o *OKGroup) WsLogin() error {
...
-				}
-				o.Websocket.ResubscribeToChannel(channelToResubscribe)
-			}
-		}
-		orderbookMutex.Unlock()
-	case okGroupWsTicker:
-		o.wsProcessTickers(response)
-	case okGroupWsTrade:
-		o.wsProcessTrades(response)
+func stringToOrderStatus(num string) order.Status {
```
thrasher-: After reviewing a lot of these in this PR and considering the nature of this, it would be good to add test coverage for these conversion funcs

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -292,113 +233,174 @@ func (o *OKGroup) WsLogin() error {
...
+		o.Websocket.DataHandler <- wshandler.UnhandledMessageWarning{Message: o.Name + wshandler.UnhandledMessage + string(respRaw)}
+		return nil
+	}
+
+	var errorResponse WebsocketErrorResponse
+	err = json.Unmarshal(respRaw, &errorResponse)
+	if err == nil && errorResponse.ErrorCode > 0 {
+		if o.Verbose {
+			log.Debugf(log.ExchangeSys,
+				"WS Error Event: %v Message: %v for %s",
```
thrasher-: To keep things consistent, please place the exchange name first like the following fmt.Errorf

exchanges/poloniex/poloniex_websocket.go
```diff
@@ -87,196 +94,318 @@ func (p *Poloniex) WsHandleData() {
...
+
+func (p *Poloniex) wsHandleData(respRaw []byte) error {
+	var result interface{}
+	err := json.Unmarshal(respRaw, &result)
+	if err != nil {
+		return err
+	}
+	if data, ok := result.([]interface{}); ok {
+		if len(data) == 0 {
+			return errors.New("c'mon man, no data")
```
thrasher-: 😆 in this case, we can just return nil as the exchange didn't send us any meaningful data

exchanges/poloniex/poloniex_websocket.go
```diff
@@ -87,196 +94,318 @@ func (p *Poloniex) WsHandleData() {
...
+										Exchange:        p.Name,
+										ID:              strconv.FormatFloat(notification[1].(float64), 'f', -1, 64),
+										Type:            order.Limit,
+										Status:          oStatus,
+										AssetType:       asset.Spot,
+									}
+									p.Websocket.DataHandler <- response
+								case "n":
+									var timeParse time.Time
+									timeParse, err = time.Parse("2006-01-02 15:04:05", notification[6].(string))
```
thrasher-: Since this layout is used quite a lot, can make this a const similar to how we've done it in other changes for easy usage by other devs

testdata/http_mock/binance/binance.json
```diff
@@ -45420,6 +45420,34 @@
     }
    ]
   },
+  "/api/v3/userDataStream": {
+   "POST": [
+    {
+     "data": {
+      "listenKey": "LOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLOLO"
```
thrasher-: https://makeagif.com/i/ZENCib

engine/orders.go
```diff
@@ -92,39 +161,8 @@ func (o *orderManager) Stop() error {
...
-				msg := fmt.Sprintf("Order manager: Exchange %s order ID=%v cancelled.",
-					k, v[y].ID)
-				log.Debugln(log.OrderBook, msg)
-				Bot.CommsManager.PushEvent(base.Event{
-					Type:    "order",
-					Message: msg,
-				})
-			}
-		}
+		o.CancelAllOrders(nil)
```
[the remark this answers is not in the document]
gloriousCode: It's definitely a lazy thing by me. Sometimes I don't like adding parameters so I'll give arguments a different use, like nil here. I agree that cancelling all orders for every exchange should probably come with an explicit instruction to do so. Or we just always provide a list of all enabled exchanges so the onus is always on the caller and not an accidental nil slice

exchanges/binance/binance_types.go
```diff
@@ -612,3 +604,95 @@ type WithdrawResponse struct {
 	Msg     string `json:"msg"`
 	ID      string `json:"id"`
 }
+
+type UserAccountStream struct {
```
xtda: Needs a comment

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -123,201 +124,322 @@ func (b *Bitmex) wsHandleIncomingData() {
...
-
-				case bitmexWSTrade:
-					var trades TradeData
-					err = json.Unmarshal(resp.Raw, &trades)
+					var oSide order.Side
+					oSide, err = order.StringToOrderSide(response.Data[x].Side)
 					if err != nil {
-						b.Websocket.DataHandler <- err
-						continue
+						b.Websocket.DataHandler <- errors.New(b.Name + " Unable to convert orderside: " + response.Data[x].Side)
```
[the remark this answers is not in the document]
gloriousCode: I would like you're feedback here. The reason I did this is because we're receiving an order which  is important. If for some reason, we haven't coded/the API changes for some status, not storing an order for management because of that seems drastic. I figure logging that we're not up to date with order status/type/side and setting them to their unknown types allows us to keep tracking the order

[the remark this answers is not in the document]
shazbert: Yeah you make a good point, some do return an error though so its not consistent. @thrasher- @xtda What shall we do here?

engine/routines.go
```diff
@@ -252,121 +251,124 @@ func WebsocketRoutine() {
...
 			}
+		} else {
+			od, err := Bot.OrderManager.orderStore.GetByExchangeAndID(d.Exchange, d.ID)
+			if err != nil {
+				return err
+			}
+			od.UpdateOrderFromDetail(d)
+		}
+	case *order.Cancel:
+		err := Bot.OrderManager.Cancel(d)
```
xtda: as nothing further is being done here can just return 

This can be done for a few of the returns in this function if you prefere return early style since after the switch return is always nil

engine/routines.go
```diff
@@ -252,121 +251,124 @@ func WebsocketRoutine() {
...
-						d.Asset)
-				}
-			default:
-				if Bot.Settings.Verbose {
-					log.Warnf(log.WebsocketMgr,
-						"%s websocket Unknown type: %+v\n",
-						ws.GetName(),
-						d)
-				}
+func WebsocketDataHandler(exchName string, data interface{}) error {
```
xtda: Needs a comment

xtda:
Also I really like FakePassingExchange can be helpful in testing other parts but not a fan of it sitting in engine/ over exhanges/

gloriousCode:
> Also I really like FakePassingExchange can be helpful in testing other parts but not a fan of it sitting in engine/ over exhanges/

Yeah, its a bit messy, but didn't want it to be included in the main application. Engine package made sense to me since at that area, we don't really care for exchange responses. Happy to hear feedback on another place. I was thinking to have it in its own test file in the exchange package (and then we could add a fakefailingexchange too)

exchanges/bitmex/bitmex_websocket.go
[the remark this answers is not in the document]
thrasher-: Order should still be stored, but we need a way to standardise the error so that the calling application can know that a potentially serious field is missing so we can address it ASAP

exchanges/binance/binance_test.go
```diff
@@ -516,3 +516,256 @@ func TestGetDepositAddress(t *testing.T) {
 		t.Error("Mock GetDepositAddress() error", err)
 	}
 }
+
+func TestWSSubscriptionHandling(t *testing.T) {
+	t.Parallel()
+	pressXToJSON := []byte(`{
```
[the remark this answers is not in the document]
gloriousCode: :D I can change it if you want something a bit more serious

exchanges/bitmex/bitmex_test.go
```diff
@@ -50,6 +51,16 @@ func TestMain(m *testing.M) {
 	if err != nil {
 		log.Fatal("Bitmex setup error", err)
 	}
+	b.CurrencyPairs.AssetTypes = asset.Items{asset.Spot}
```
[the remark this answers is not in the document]
gloriousCode: This was due to old test code which forced asset types into spot. I've removed that too.

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -33,91 +33,97 @@ func (b *Bitstamp) WsConnect() error {
...
+		}
+	case "trade":
+		wsTradeTemp := websocketTradeResponse{}
+		err := json.Unmarshal(respRaw, &wsTradeTemp)
+		if err != nil {
+			return err
+		}
+		currencyPair := strings.Split(wsResponse.Channel, "_")
+		p := currency.NewPairFromString(strings.ToUpper(currencyPair[2]))
+		b.Websocket.DataHandler <- wshandler.TradeData{
```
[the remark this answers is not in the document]
gloriousCode: This one doesn't have any tests, but I'll add some

exchanges/kraken/kraken_test.go
gloriousCode: That breaks 32 bit testing

engine/orders.go
```diff
@@ -204,18 +313,14 @@ func (o *orderManager) Submit(exchName string, newOrder *order.Submit) (*orderSu
...
-	if err != nil {
-		log.Warnf(log.OrderMgr,
-			"Order manager: Unable to generate UUID. Err: %s\n",
-			err)
-	}
-
+	// Lock at this point to ensure wrapper + websocket order submissions return
+	// before any routines.go order submission competes and this returns an error
+	o.orderStore.m.Lock()
+	defer o.orderStore.m.Unlock()
```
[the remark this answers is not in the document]
gloriousCode: So this was added to address a data race where a wrapper websocket request and the datahandler would want to add an order to the orderstore at the same time. There was potential for the wrapper caller to get an "order already exists" error because of that. To address this, I've added the following function:
```
func (w *WebsocketConnection) IsIDWaitingForResponse(id int64) bool {
	w.Lock()
	for k := range w.IDResponses {
		if k == id && w.IDResponses[k] == nil {
			w.Unlock()
			return true
		}
	}
	w.Unlock()
	return false
}
```
Now, when we are doing a wrapper websocket request, we add an ID to the `IDResponses` map with a nil value which says we are awaiting a response.
When a response (or a timeout) occurs, that id will be removed from the map. So now there are no locks used here because we now know when to pass data through the data handler and when not to.

**tl;dr**, removed those locks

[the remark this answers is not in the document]
shazbert: Awesome work mate!

exchanges/bitmex/bitmex_websocket.go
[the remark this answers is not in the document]
gloriousCode: I could create a type which is handled by the routines.go datahandler. Like an `OrderClassificationError` or something to encompass Type, Side and Status. Then a future PR can be around storing the error in the database.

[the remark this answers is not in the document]
gloriousCode: I've created `order.ClassificationError` which contains `Exchange`, `OrderID`, and the `Error` from `order.StringToStatus/Type/Side`. It is now handled separately in `routines.go` so if we decide to store these issues in the db, its very simple to do in a future PR 🦀 
There is no more inconsistency with return errs either. It all goes through the datahandler and continues to add the order to the order manager

exchanges/binance/binance_websocket.go
```diff
@@ -94,142 +130,248 @@ func (b *Binance) WsHandleData() {
...
+			oType, err := order.StringToOrderType(data.OrderType)
+			if err != nil {
+				b.Websocket.DataHandler <- errors.New(b.Name + " Unable to convert order type: " + data.OrderType)
+			}
+			var oSide order.Side
+			oSide, err = order.StringToOrderSide(data.Side)
+			if err != nil {
+				return err
+			}
+			ts := time.Unix(data.OrderCreationTime, 0)
```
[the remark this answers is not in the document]
gloriousCode: Thanks for that, I converted that wrong

engine/routines.go
```diff
@@ -252,121 +251,124 @@ func WebsocketRoutine() {
...
+	if data == nil {
+		return fmt.Errorf("routines.go - exchange %s nil data sent to websocket",
+			exchName)
+	}
+	switch d := data.(type) {
+	case string:
+		switch d {
+		case wshandler.WebsocketNotEnabled:
+			return fmt.Errorf("routines.go - exchange %s websocket not enabled",
```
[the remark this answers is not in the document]
gloriousCode: d here is `"exchange_websocket_not_enabled"`, I don't think this scenario can ever exist. I think I designed this under the impression someone could disable the websocket via the CLI. I'm going to remove it

exchanges/btse/btse_websocket.go
```diff
@@ -37,14 +38,58 @@ func (b *BTSE) WsConnect() error {
...
+		return order.New
+	case "ORDER_CANCELLED":
+		return order.Cancelled
+	case "ORDER_FULL_TRANSACTED":
+		return order.Filled
+	case "ORDER_PARTIALLY_TRANSACTED":
+		return order.PartiallyFilled
+	case "TRIGGER_ACTIVATED":
+		return order.Active
+	case "INSUFFICIENT_BALANCE", "MARKET_UNAVAILABLE":
```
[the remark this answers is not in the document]
gloriousCode: I've added them as order status 👍

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -406,7 +552,7 @@ func (b *Bitmex) GenerateDefaultSubscriptions() {
 		}
 	}
 
-	channels := []string{bitmexWSOrderbookL2, bitmexWSTrade}
+	channels := []string{ /*bitmexWSOrderbookL2, */ bitmexWSTrade}
```
shazbert: Just check this is not an oversight

exchanges/huobi/huobi_websocket.go
```diff
@@ -119,129 +121,210 @@ func (h *HUOBI) wsMultiConnectionFunnel(ws *wshandler.WebsocketConnection, url s
...
+		var oStatus order.Status
+		oStatus, err = stringToOrderStatus(response.Data.OrderState)
+		if err != nil {
+			h.Websocket.DataHandler <- order.ClassificationError{
+				Exchange: h.Name,
+				OrderID:  orderID,
+				Err:      err,
+			}
+		}
+		p := currency.NewPairFromString(data[1])
```
thrasher-: Please use currency.NewPairFromFormattedPairs here otherwise this will cause pair mismatches

exchanges/huobi/huobi_wrapper.go
```diff
@@ -616,28 +616,55 @@ func (h *HUOBI) GetOrderInfo(orderID string) (order.Detail, error) {
...
 	orderDetail = order.Detail{
 		Exchange:       h.Name,
-		ID:             strconv.FormatInt(respData.ID, 10),
+		ID:             orderID,
 		AccountID:      strconv.FormatInt(respData.AccountID, 10),
-		CurrencyPair:   currency.NewPairFromString(respData.Symbol),
-		OrderType:      orderType,
-		OrderSide:      orderSide,
-		OrderDate:      time.Unix(0, respData.CreatedAt*int64(time.Millisecond)),
+		Pair:           currency.NewPairFromString(respData.Symbol),
```
thrasher-: Out of scope, but please use currency.NewPairFromFormattedPairs here otherwise this will cause pair mismatches

exchanges/huobi/huobi_types.go
```diff
@@ -526,14 +520,21 @@ type WsAuthenticatedOrdersUpdateResponseData struct {
...
 }
 
 // WsAuthenticatedOrdersResponse response from Orders authenticated subscription
 type WsAuthenticatedOrdersResponse struct {
-	WsAuthenticatedDataResponse
+	WsResponse
 	Data []WsAuthenticatedOrdersResponseData `json:"data"`
 }
 
+// WsAuthenticatedOrdersResponse response from Orders authenticated subscription
```
thrasher-: Please change the comment to match WsOldOrderUpdate

exchanges/okgroup/okgroup_types.go
```diff
@@ -1409,110 +1389,30 @@ type WebsocketUserSwapPositionHoldingData struct {
...
-	MarginTrading  int64   `json:"margin_trading,omitempty"`
-	Type           string  `json:"type"`
-	// Price        A member, but part already exists as part of WebsocketDataResponse
-	// InstrumentID A member, but part already exists as part of WebsocketDataResponse
-	// Timestamp    A member, but part already exists as part of WebsocketDataResponse
-	// OrderID      A member, but part already exists as part of WebsocketDataResponse
+	Table string `json:"table"`
+	Data  []struct {
+		ClientOid      string    `json:"client_oid"`
+		CreatedAt      string    `json:"created_at"`
```
thrasher-: Can use time.Time here

exchanges/okgroup/okgroup_types.go
```diff
@@ -1409,110 +1389,30 @@ type WebsocketUserSwapPositionHoldingData struct {
...
+		FilledNotional string    `json:"filled_notional"`
+		FilledSize     string    `json:"filled_size"`
+		InstrumentID   string    `json:"instrument_id"`
+		LastFillPx     string    `json:"last_fill_px"`
+		LastFillQty    string    `json:"last_fill_qty"`
```
thrasher-: Many of these should be converted to more useful types instead of autoconverted to strings

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -494,36 +520,38 @@ func (o *OKGroup) wsProcessCandles(response *WebsocketDataResponse) {
...
 
 // WsProcessOrderBook Validates the checksum and updates internal orderbook values
-func (o *OKGroup) WsProcessOrderBook(response *WebsocketDataResponse) (err error) {
+func (o *OKGroup) WsProcessOrderBook(respRaw []byte) error {
+	var response WebsocketOrderBooksData
+	err := json.Unmarshal(respRaw, &response)
+	if err != nil {
+		return err
+	}
+	orderbookMutex.Lock()
```
thrasher-: should use defer orderbookMutex.Unlock() here, many code paths here which could potentially lock this for eternity

exchanges/order/order_test.go
```diff
@@ -534,3 +542,375 @@ func TestStringToOrderStatus(t *testing.T) {
 		})
 	}
 }
+
+func TestUpdateOrderFromModify(t *testing.T) {
+	var leet = "1337"
```
thrasher-: m3g4 l33t

exchanges/kraken/kraken_types.go
```diff
@@ -475,19 +468,69 @@ type WsTokenResponse struct {
...
+	RequestID    int64  `json:"reqid"`
+	Status       string `json:"status"`
+	Subscription struct {
+		Depth    int    `json:"depth"`
+		Interval int    `json:"interval"`
+		Name     string `json:"name"`
+	} `json:"subscription"`
+}
+
+type WsOpenOrder struct {
```
thrasher-: Needs commenterino

exchanges/okgroup/okgroup_types.go
```diff
@@ -1306,93 +1306,73 @@ type WebsocketEventResponse struct {
...
 type WebsocketOrderBooksData struct {
-	Asks     [][]interface{} `json:"asks,omitempty"` // [0] Price, [1] Size, [2] Number of orders
-	Bids     [][]interface{} `json:"bids,omitempty"` // [0] Price, [1] Size, [2] Number of orders
-	Checksum int32           `json:"checksum,omitempty"`
+	Table  string               `json:"table"`
+	Action string               `json:"action"`
+	Data   []WebsocketOrderBook `json:"data"`
+}
+
+type WebsocketOrderBook struct {
```
thrasher-: Please add commenterino

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -292,113 +233,182 @@ func (o *OKGroup) WsLogin() error {
...
-// WsHandleErrorResponse sends an error message to ws handler
-func (o *OKGroup) WsHandleErrorResponse(event WebsocketErrorResponse) {
-	errorMessage := fmt.Sprintf("%v error - %v message: %s ",
-		o.Name,
-		event.ErrorCode,
-		event.Message)
-	if o.Verbose {
-		log.Error(log.ExchangeSys, errorMessage)
+// wsReadData receives and passes on websocket messages for processing
+func (o *OKGroup) WsReadData(wg *sync.WaitGroup) {
```
thrasher-: WsReadData differs to wsReadData and will throw a linter err

exchanges/okgroup/okgroup_websocket.go
```diff
@@ -292,113 +233,182 @@ func (o *OKGroup) WsLogin() error {
...
-	switch table[:assetIndex] {
-	case asset.Futures.String():
-		return asset.Futures
-	case asset.Spot.String():
-		return asset.Spot
-	case "swap":
-		return asset.PerpetualSwap
-	case asset.Index.String():
-		return asset.Index
+func StringToOrderStatus(num string) (order.Status, error) {
```
thrasher-: Please add comment

exchanges/websocket/wshandler/wshandler.go
```diff
@@ -620,14 +620,14 @@ func (w *Websocket) CanUseAuthenticatedEndpoints() bool {
 	return w.canUseAuthenticatedEndpoints
 }
 
-// AddResponseWithID adds data to IDResponses with locks and a nil check
-func (w *WebsocketConnection) AddResponseWithID(id int64, data []byte) {
+// SetResponseIDAndData adds data to IDResponses with locks and a nil check
+func (w *WebsocketConnection) SetResponseIDAndData(id int64, data []byte) {
 	w.Lock()
-	defer w.Unlock()
```
thrasher-: Achieves the same result, it's easier to know your lock will get unlocked if you put a defer straight after, otherwise you run into scenarios like this: https://github.com/thrasher-corp/gocryptotrader/pull/446/files#diff-c756b319860bc00c5b63bb5399dbb40bR554

engine/orders.go
```diff
@@ -17,47 +17,121 @@ import (
...
+		return errors.New("unable to get exchange by name")
+	}
 	if o.exists(order) {
 		return ErrOrdersAlreadyExists
 	}
-
+	// Untracked websocket orders will not have internalIDs yet
+	if order.InternalOrderID == "" {
+		id, err := uuid.NewV4()
+		if err != nil {
```
xtda: At this point order.InternalOrderID wont be set to anything is there ever a case where we would need to set these to some generic known id so they are not blank later?

gctscript/wrappers/gct/exchange/exchange.go
```diff
@@ -89,7 +89,7 @@ func (e Exchange) QueryOrder(exch, orderID string) (*order.Detail, error) {
 
 // SubmitOrder submit new order on exchange
 func (e Exchange) SubmitOrder(exch string, submit *order.Submit) (*order.SubmitResponse, error) {
```
xtda: exch is unused here now and can be removed from the interface & implementations

exchanges/btse/btse_test.go
```diff
@@ -358,3 +360,54 @@ func TestCancelAllExchangeOrders(t *testing.T) {
...
+func TestWsOrderNotification(t *testing.T) {
+	status := []string{"ORDER_INSERTED", "ORDER_CANCELLED", "TRIGGER_INSERTED", "ORDER_FULL_TRANSACTED", "ORDER_PARTIALLY_TRANSACTED", "INSUFFICIENT_BALANCE", "TRIGGER_ACTIVATED", "MARKET_UNAVAILABLE"}
+	for i := range status {
+		pressXToJSON := []byte(`{"topic": "notificationApi","data": [{"symbol": "BTC-USD","orderID": "1234","orderMode": "MODE_BUY","orderType": "TYPE_LIMIT","price": "1","size": "1","status": "` + status[i] + `","timestamp": "1580349090693","type": "STOP","triggerPrice": "1"}]}`)
+		err := b.wsHandleData(pressXToJSON)
+		if err != nil {
+			t.Error(err)
+		}
+	}
+}
```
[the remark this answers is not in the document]
gloriousCode: I'm not sure what you mean here. Could you point me to the notification so I can add a test scenario? I see no fundNotification or orderChange in btse websocket notifications

exchanges/coinbene/coinbene_websocket.go
```diff
@@ -78,264 +82,296 @@ func (c *Coinbene) GenerateAuthSubs() {
...
+				c.Websocket.DataHandler <- order.ClassificationError{
+					Exchange: c.Name,
+					OrderID:  orders.Data[i].OrderID,
+					Err:      err,
 				}
-				c.Websocket.DataHandler <- orders
-			default:
-				c.Websocket.DataHandler <- fmt.Errorf("%s - unhandled response '%s'", c.Name, stream.Raw)
+			}
+			p := currency.NewPairFromString(orders.Data[i].Symbol)
```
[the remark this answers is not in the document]
gloriousCode: Should I use this across all websocket instances instead of `NewPairFromString`?

exchanges/btse/btse_test.go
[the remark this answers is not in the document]
thrasher-: This ended up in the wrong place, this was meant for BTC Markets: https://github.com/thrasher-corp/gocryptotrader/pull/446/files#diff-7a6e4ee38b8c801f85798221829fbe89R180

exchanges/btcmarkets/btcmarkets_websocket.go
```diff
@@ -68,147 +65,208 @@ func (b *BTCMarkets) WsHandleData() {
...
+			High:         tick.High24,
+			Low:          tick.Low24h,
+			Bid:          tick.Bid,
+			Ask:          tick.Ask,
+			Last:         tick.Last,
+			LastUpdated:  tick.Timestamp,
+			AssetType:    asset.Spot,
+			Pair:         p,
+		}
+	case fundChange:
```
thrasher-: It would be nice to add tests for fundChange and orderChange

exchanges/coinbene/coinbene_websocket.go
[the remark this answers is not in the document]
thrasher-: For this exchange, yep. Others are context dependent, based on if the exchange doesn't send us pairs with delimiters AND if pairs are > 6 characters in length (btcusd would be fine, but dashusdt wouldn't (would become das-husdt)

exchanges/btcmarkets/btcmarkets_test.go
```diff
@@ -500,3 +503,14 @@ func TestGetActiveOrders(t *testing.T) {
 		t.Fatal(err)
 	}
 }
+
+func TestWsOrderNotification(t *testing.T) {
```
thrasher-: Please add ws test coverage for heatrbeat, ticker, orderbook, trade plus the already existing orderChange / fundChange request

thrasher-:
Please merge/rebase

engine/orders.go
gloriousCode: This one is weird because its dependent on an external library. It failing will impact many things, but I don't think it should stop the logging of an order. If it ends up erroring, later down the line, I'd want it to still have a GUID there for consistency, but if its GUID gen is erroring, I can't really put the same code in at a later stage. I'm not sure what to do here

thrasher-: We can add our own internal GUID generator in the case that this happens, even though it should theoretically be close to nil

xtda: Yeah not sure on a clean fix

I was more just thinking if an order is added with no internal id due to uuid in the very rare case erroring GetByInternalOrderID() wont work 

So in that case is it work not adding the order? how important is it that intenralorderid is filled out correctly?

gloriousCode: Right now, its actually not very important. The order manager is not storing orders in the database, so when you reload the application, you would end up with the same order having a different GUID. So I think our own GUID handling as well as order storage will be in a future PR

exchanges/coinbene/coinbene_websocket.go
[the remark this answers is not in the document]
gloriousCode: I went through websocket orders recently and tried to remove any hard coded asset types. Sadly doing this will reintroduce asset types. I'll try and work out a way of achieving future proofing and currency formatting

exchanges/binance/binance_websocket.go
```diff
@@ -94,142 +133,270 @@ func (b *Binance) WsHandleData() {
...
+					Err:      err,
 				}
-				continue
-			case "ticker":
-				t := TickerStream{}
-				err := json.Unmarshal(multiStreamData.Data, &t)
+			}
+			var p currency.Pair
+			var a asset.Item
+			p, a, err = b.GetRequestFormattedPairAndAssetType(data.Symbol)
```
thrasher-: We'll need to address this in another PR as this logic will break down when an exchange has spot and perpetual swaps (since they can both be BTC-USD, but would be fine for indexes and futures (but dependent on each exchange).) currency.NewPairFromFormattedPairs is less error prone since you have to specify the asset type

engine/orders.go
```diff
@@ -17,47 +17,120 @@ import (
...
+		if r[x].ID == id {
+			return r[x], nil
+		}
+	}
+	return nil, ErrOrderNotFound
+}
+
+// GetByExchange returns orders by exchange
+func (o *orderStore) GetByExchange(exchange string) ([]*order.Detail, error) {
+	o.m.Lock()
```
xtda: Is it worth having these are rwmutex that way multiple goroutines can get read locks?
nothing is updated within GetByExchange() so it should be safe for multiple reads

engine/orders.go
```diff
@@ -17,47 +17,120 @@ import (
...
+	if !ok {
+		return nil, ErrExchangeNotFound
+	}
+	return r, nil
+}
+
+// GetByInternalOrderID will search all orders for our internal orderID
+// and return the order
+func (o *orderStore) GetByInternalOrderID(internalOrderID string) (*order.Detail, error) {
+	o.m.Lock()
```
xtda: same here if its swapped over to rwmutex you can lock for reading in the get and lock for read/write in updates

engine/orders.go
xtda: Happy with that good to resolve

exchanges/coinbasepro/coinbasepro_types.go
```diff
@@ -475,19 +439,39 @@ type WebsocketL2Update struct {
...
+	Sequence  int64  `json:"sequence"`
+	ProductID string `json:"product_id"`
+}
+
+type wsStatus struct {
+	Currencies []struct {
+		ConvertibleTo []string    `json:"convertible_to"`
+		Details       struct{}    `json:"details"`
+		ID            string      `json:"id"`
+		MaxPrecision  string      `json:"max_precision"`
```
thrasher-: These and MinSize can be float64's

exchanges/coinbasepro/coinbasepro_types.go
```diff
@@ -475,19 +439,39 @@ type WebsocketL2Update struct {
...
+		MinSize       string      `json:"min_size"`
+		Name          string      `json:"name"`
+		Status        string      `json:"status"`
+		StatusMessage interface{} `json:"status_message"`
+	} `json:"currencies"`
+	Products []struct {
+		BaseCurrency   string      `json:"base_currency"`
+		BaseIncrement  string      `json:"base_increment"`
```
thrasher-: BaseIncrement, BaseMaxSize, BaseMinSize can be float64s

exchanges/coinbasepro/coinbasepro_types.go
```diff
@@ -475,19 +439,39 @@ type WebsocketL2Update struct {
...
+		BaseMaxSize    string      `json:"base_max_size"`
+		BaseMinSize    string      `json:"base_min_size"`
+		CancelOnly     bool        `json:"cancel_only"`
+		DisplayName    string      `json:"display_name"`
+		ID             string      `json:"id"`
+		LimitOnly      bool        `json:"limit_only"`
+		MaxMarketFunds string      `json:"max_market_funds"`
```
thrasher-: This and MinMarketFunds, QuoteIncrement can be float64's

engine/fake_exchange_test.go
```diff
@@ -0,0 +1,192 @@
...
+}
+func (h *FakePassingExchange) GetDefaultConfig() (*config.ExchangeConfig, error) { return nil, nil }
+func (h *FakePassingExchange) GetBase() *exchange.Base                           { return nil }
+func (h *FakePassingExchange) SupportsAsset(_ asset.Item) bool                   { return true }
+func (h *FakePassingExchange) GetHistoricCandles(_ currency.Pair, _, _ int64) ([]exchange.Candle, error) {
+	return []exchange.Candle{}, nil
+}
+func (h *FakePassingExchange) DisableRateLimiter() error { return nil }
+func (h *FakePassingExchange) EnableRateLimiter() error  { return nil }
+func (h *FakePassingExchange) WithdrawCryptocurrencyFunds(withdrawRequest *withdraw.Request) (*withdraw.ExchangeResponse, error) {
```
xtda: unused param if you want to _ it

engine/fake_exchange_test.go
```diff
@@ -0,0 +1,192 @@
...
+	return nil, nil
+}
+func (h *FakePassingExchange) WithdrawFiatFunds(withdrawRequest *withdraw.Request) (*withdraw.ExchangeResponse, error) {
```
xtda: unused param if you want to _ it

engine/fake_exchange_test.go
```diff
@@ -0,0 +1,192 @@
...
+	return nil, nil
+}
+func (h *FakePassingExchange) WithdrawFiatFundsToInternationalBank(withdrawRequest *withdraw.Request) (*withdraw.ExchangeResponse, error) {
```
xtda: unused param if you want to _ it

exchanges/coinut/coinut_wrapper.go
```diff
@@ -721,22 +721,22 @@ func (c *COINUT) GetActiveOrders(req *order.GetOrdersRequest) ([]order.Detail, e
...
+					RemainingAmount: openOrders.Orders[i].OpenQuantity,
 				})
 			}
 		}
 	} else {
 		var instrumentsToUse []int64
-		if len(req.Currencies) > 0 {
-			for x := range req.Currencies {
-				curr := c.FormatExchangeCurrency(req.Currencies[x],
+		if len(req.Pairs) > 0 {
```
thrasher-: Can drop this if check, since a range will effectively do the same thing and instantly continue if it's 0

exchanges/coinut/coinut_wrapper.go
```diff
@@ -816,9 +816,9 @@ func (c *COINUT) GetOrderHistory(req *order.GetOrdersRequest) ([]order.Detail, e
...
```
thrasher-: Same here

exchanges/huobi/huobi_websocket.go
```diff
@@ -271,14 +353,20 @@ func (h *HUOBI) wsHandleMarketData(resp WsMessage) {
 				h.GetEnabledPairs(asset.Spot), h.GetPairFormat(asset.Spot, true)),
 			Timestamp: time.Unix(0, trade.Tick.Timestamp*int64(time.Millisecond)),
 		}
-	case strings.Contains(init.Channel, "detail"):
+	case strings.Contains(init.Channel, "detail"),
+		strings.Contains(init.Rep, "detail"):
 		var wsTicker WsTick
-		err := json.Unmarshal(resp.Raw, &wsTicker)
+		err := json.Unmarshal(respRaw, &wsTicker)
```
thrasher-: Please add ticker fetching to the WS protocol supported features

exchanges/huobi/huobi_wrapper.go
thrasher-: Bump, this would be good to do in this PR

exchanges/okex/okex_wrapper.go
```diff
@@ -137,6 +137,8 @@ func (o *OKEX) SetDefaults() {
 				Unsubscribe:            true,
 				AuthenticatedEndpoints: true,
 				MessageCorrelation:     true,
+				GetOrders:              true,
```
thrasher-: Please add AccountBalance to WS supported protocol features

engine/orders.go
gloriousCode: That's a good point. I'll change it

exchanges/binance/binance_websocket.go
thrasher-: Should also mention that a better solution for this would be to detect the context in the exchange wrapper or func, then use the specific asset type, but this outside the scope of this PR

exchanges/exchange.go
```diff
@@ -340,6 +341,24 @@ func (e *Base) GetEnabledPairs(assetType asset.Item) currency.Pairs {
 	return pairs.Format(format.Delimiter, format.Index, format.Uppercase)
 }
 
+// GetRequestFormattedPairAndAssetType is a method that returns the enabled currency pair of
+// along with its asset type
+func (e *Base) GetRequestFormattedPairAndAssetType(p string) (currency.Pair, asset.Item, error) {
```
shazbert: Only issue with this and the one above is if we have a duplicate in assets i.e. BTC-USD for spot, index like in OKEX and it might come back with a false positive.

exchanges/exchange.go
```diff
@@ -340,6 +341,24 @@ func (e *Base) GetEnabledPairs(assetType asset.Item) currency.Pairs {
...
+		format := e.GetPairFormat(assetTypes[i], true)
+		pairs := e.CurrencyPairs.GetPairs(assetTypes[i], true)
+		for j := range pairs {
+			formattedPair := pairs[j].Format(format.Delimiter, format.Uppercase)
+			if strings.EqualFold(formattedPair.String(), p) {
+				return formattedPair, assetTypes[i], nil
+			}
+		}
+	}
+	return response, "", errors.New("Pair not found: " + p)
```
shazbert: might be a linter issue better to lower case

exchanges/coinut/coinut_websocket.go
```diff
@@ -116,31 +121,116 @@ func (c *COINUT) WsHandleData() {
...
+		err := json.Unmarshal(respRaw, &orders)
+		if err != nil {
+			return err
+		}
+		for i := range orders {
+			o, err2 := c.parseOrderContainer(&orders[i])
+			if err2 != nil {
+				return err2
+			}
+			c.Websocket.DataHandler <- o
```
shazbert: out of scope: could send as a slice.

engine/routines.go
```diff
@@ -252,121 +252,126 @@ func WebsocketRoutine() {
...
+		return fmt.Errorf("routines.go exchange %s websocket error - %s", exchName, data)
+	case wshandler.TradeData:
+		if Bot.Settings.Verbose {
+			log.Infof(log.WebsocketMgr, "%s websocket %s %s trade updated %+v",
+				exchName,
+				FormatCurrency(d.CurrencyPair),
+				d.AssetType,
+				d)
+		}
+		return nil
```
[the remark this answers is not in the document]
gloriousCode: I added them to address Andrew's nit about return earlier :D I can remove them again

exchanges/binance/binance_wrapper.go
```diff
@@ -96,10 +96,14 @@ func (b *Binance) SetDefaults() {
 				CryptoWithdrawalFee: true,
 			},
 			WebsocketCapabilities: protocol.Features{
-				TradeFetching:     true,
-				TickerFetching:    true,
-				KlineFetching:     true,
-				OrderbookFetching: true,
+				TradeFetching:          true,
```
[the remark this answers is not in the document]
gloriousCode: It does, but our current implementation doesn't. We use a big URL to do the subscriptions. I'll still add it

exchanges/binance/binance_websocket.go
```diff
@@ -94,142 +133,270 @@ func (b *Binance) WsHandleData() {
...
+					wsKline.AssetType = asset.Spot
+					wsKline.Exchange = b.Name
+					wsKline.StartTime = time.Unix(0, kline.Kline.StartTime*int64(time.Millisecond))
+					wsKline.CloseTime = time.Unix(0, kline.Kline.CloseTime*int64(time.Millisecond))
+					wsKline.Interval = kline.Kline.Interval
+					wsKline.OpenPrice, _ = strconv.ParseFloat(kline.Kline.OpenPrice, 64)
+					wsKline.ClosePrice, _ = strconv.ParseFloat(kline.Kline.ClosePrice, 64)
+					wsKline.HighPrice, _ = strconv.ParseFloat(kline.Kline.HighPrice, 64)
+					wsKline.LowPrice, _ = strconv.ParseFloat(kline.Kline.LowPrice, 64)
+					wsKline.Volume, _ = strconv.ParseFloat(kline.Kline.Volume, 64)
```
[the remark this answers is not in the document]
gloriousCode: Definitely in scope :D

exchanges/exchange.go
gloriousCode: I've added additional commentary to suggest not using it when there are duplicates. OKEX has not implemented this. One which did have an issue (coinbene) only has SWAP support for websocket right now, so I've used `NewPairFromFormattedPairs` there

thrasher-: Awesome, thanks!

PR #474, opened 2020-03-26
(FIX) Update ZB API URLs

gloriousCode:
# PR Description
ZB has changed its websocket URL, I've changed it. See: https://www.zb.com/api#WebSocket%20API
Its also changed its Trading and Data API urls, however, they appear to be redirecting. I've updated those now too

Fixes # (issue)

## How has this been tested
- Comparing results of old URL to current URL. The new one can connect, the old one can't
- go test ./... race
- Testing with API credentials

## Checklist

### DIFF
--- a/exchanges/zb/zb_test.go
+++ b/exchanges/zb/zb_test.go
@@ -511,20 +511,6 @@ func TestWsTransferFunds(t *testing.T) {
 	}
 }
 
-// TestWsCreateSuUserKey ws test
-func TestWsCreateSuUserKey(t *testing.T) {
-	setupWsAuth(t)
-	subUsers, err := z.wsGetSubUserList()
-	if err != nil {
-		t.Fatal(err)
-	}
-	userID := subUsers.Message[0].UserID
-	_, err = z.wsCreateSubUserKey(true, true, true, true, "subu", strconv.FormatInt(userID, 10))
-	if err != nil {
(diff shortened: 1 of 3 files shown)


gloriousCode:
Actually, looks like ZB has changed more than just the URL, I'll investigate further and update the PR later

exchanges/zb/zb_test.go
```diff
@@ -183,7 +183,7 @@ func setFeeBuilder() *exchange.FeeBuilder {
 func TestGetFeeByTypeOfflineTradeFee(t *testing.T) {
 	var feeBuilder = setFeeBuilder()
 	z.GetFeeByType(feeBuilder)
-	if !areTestAPIKeysSet() {
+	if !z.ValidateAPICredentials() {
```
shazbert: When we call this, we are calling UpdateAccountInfo so it might be better to reside in testmain and only call it once to validate the keys. It's not a huge issue but we are doubling our outbound auth requests in testing.

gloriousCode: Could you point me to the code you're referring to here? `ValidateAPICredentials` is a base exchange function which looks at config values

shazbert: Doh! That's my wrong, you are correct. I was looking at the Wrapper func ValidateCredentials. My mistake!

PR #491, opened 2020-04-23
Implement Request Retry and Backoff

dackroyd:
Refactoring the timeout retries into a more general 'retry policy' with support for retrying on HTTP 429 (Too Many Requests) and other responses with a `Retry-After` header

The delay between requests is controlled by a combination of a 'backoff' (currently only a simple linear backoff), and honouring the `Retry-After` value (longest delay wins)

This makes the 'rate limiter' an optional argument as well, removing the use of `nil` when one isn't supplied

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test Binance with client rate limiter disabled, and a large number of pairs: this still fails with rate limit errors, but it does retry before doing so. Having the client side limits enabled as well should avoid the bulk of the cases, and any 429s which do occur will be retried

I have not run through checking against the real exchanges for each case. The change which affects them all is the limiter moving to be an option, which should be easy to verify.

I still need to run through a few other checks, such as for the main application flags

## Checklist


engine/engine_types.go
```diff
@@ -59,6 +59,7 @@ type Settings struct {
 	EnableExchangeRESTSupport      bool
 	EnableExchangeWebsocketSupport bool
 	MaxHTTPRequestJobsLimit        int
+	RequestMaxRetryAttempts        int
```
dackroyd, line 62: This is supposed to replace the `RequestTimeoutRetryAttempts` but I've kept that around for the moment given it is a flag supported by the application

I need to verify that this (and the old flag) are working as expected

exchanges/request/request.go
```diff
@@ -19,14 +19,25 @@ import (
 )
 
 // New returns a new Requester
-func New(name string, httpRequester *http.Client, l Limiter) *Requester {
-	return &Requester{
-		HTTPClient:           httpRequester,
-		Limiter:              l,
```
dackroyd, line 25: Stopped exporting the `Limiter` since it is supplied as an argument, unsure if there is really a need to keep it exported (hopefully not?)

exchanges/request/request_test.go
```diff
@@ -174,25 +195,35 @@ func TestDoRequest(t *testing.T) {
 	t.Parallel()
 	r := New("test",
 		new(http.Client),
-		&globalshell)
+		WithLimiter(&globalshell))
 
 	err := r.SendPayload(&Item{})
```
dackroyd, line 200: I'm not clear what a lot of these cases were actually _trying_ to verify, since they expect an error (or not), but were not very prescriptive about _which_ error. Without the 'endpoint limit' applied to the `Item`, many of these were failing, but for the **wrong** reason, and still passing the test

exchanges/request/request.go
```diff
@@ -19,14 +19,25 @@ import (
...
-		Name:                 name,
-		timeoutRetryAttempts: TimeoutRetryAttempts,
-		timedLock:            timedmutex.NewTimedMutex(DefaultMutexLockTimeout),
+func New(name string, httpRequester *http.Client, opts ...RequesterOption) *Requester {
+	maxRetries := MaxRetryAttempts
+	if maxRetries == 0 && TimeoutRetryAttempts > 0 {
```
dackroyd, line 24: 'Max retries' replaces 'timeout retry attempts', but aiming to keep some backward compatibility to avoid breaking existing configs. 'Timeout retries' isn't accurate anymore, as we retry on more than just timeouts now. As it stands, this doesn't work as `MaxRetryAttempts` is always non-zero, so `TimeoutRetryAttempts` will never be considered.

Need feedback regarding what the preference is for how this behaves / whether the existing 'timeout retry attempts' flag should be dropped entirely

exchanges/request/request.go
```diff
@@ -122,27 +133,46 @@ func (r *Requester) doRequest(req *http.Request, p *Item) error {
...
+				r.drainBody(resp.Body)
+			}
+
+			if attempt > r.maxRetries {
+				if err != nil {
+					return fmt.Errorf("request.go error - failed to retry request, err: %v", err)
 				}
-				timeoutError = err
-				continue
+				return fmt.Errorf("request.go error - failed to retry request, status: %s", resp.Status)
```
dackroyd, line 156: Example error from here:

```
[ERROR] | 23/04/2020 22:14:36 | Binance failed to update tradable pairs. Err: request.go error - failed to retry request, status: 418 I'm a teapot
```

i.e. Binance response status for continuing to exceed the rate limit.

exchanges/request/retry_test.go
```diff
@@ -0,0 +1,124 @@
...
+		Response *http.Response
+	}
+	type want struct {
+		Delay time.Duration
+	}
+	testTable := map[string]struct {
+		Args args
+		Want want
+	}{
+		"No Response": {},
```
dackroyd, line 86: I've omitted values here where the _zero value_ is being used, since they are populated by default.

If it is preferred, I can declare them explicitly, e.g:

```
"No Response": {
	Args:  args{Response: nil},
	Want: want{Delay: 0},
},
```

Setting `Now` is optional as it is only used in the case where the `Retry-After` header is present AND uses a date.

thrasher-: Am happy to keep them omitted! This is an aim also for the project as a whole and makes the code neater

exchanges/request/request.go
thrasher-: Originally we wanted to differentiate between network transient errors and outright errors, say a bad request to an invalid endpoint or an internal server error. Because we have a connection monitor now, I think it's fine to ditch the timeout retry attempts var (as this is a cmdline param, it's OK to get rid of. Anything config related however, requires an upgrade path)

engine/engine_types.go
dackroyd: Great! I'll go about removing the old config later today 👍

dackroyd: Old setting/flag has now been removed

exchanges/request/request.go
```diff
@@ -237,3 +261,13 @@ func (r *Requester) SetProxy(p *url.URL) error {
 	}
 	return nil
 }
+
+func (r *Requester) drainBody(body io.ReadCloser) {
+	defer body.Close()
+	if _, err := io.Copy(ioutil.Discard, io.LimitReader(body, drainBodyLimit)); err != nil {
```
shazbert, line 267: If all we are doing is draining, probably just read 0 or 1 byte instead of 100,000? Just say the body is bigger than 100k does this not drain? 
If we fail to drain as well can we send/re-use another request, might want to return that as an error and then stop a new potential request that will fail?

xtda: io.LimitReader will read to N bytes then return EOF so it will still drain but since we are not doing anything still most likely not worth draining up to 100,000 bytes

exchanges/request/request.go
```diff
@@ -237,3 +265,13 @@ func (r *Requester) SetProxy(p *url.URL) error {
 	}
 	return nil
 }
+
+func (r *Requester) drainBody(body io.ReadCloser) {
+	defer body.Close()
```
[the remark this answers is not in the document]
dackroyd: Is there a particular concern here with the use of `defer`? In Go 1.14 this performance difference is negligible for defer or not: https://golang.org/doc/go1.14#runtime

> This release improves the performance of most uses of defer to incur almost zero overhead compared to calling the deferred function directly. As a result, defer can now be used in performance-critical code without overhead concerns.

Go 1.13 made a lot of headway towards this, with Go 1.14 there is supposed to be no need to avoid `defer`

exchanges/request/backoff.go
```diff
@@ -0,0 +1,22 @@
...
+func DefaultBackoff() Backoff {
+	return LinearBackoff(100*time.Millisecond, time.Second)
+}
+
+// LinearBackoff applies a backoff increasing by a base amount with each retry capped at a maximum duration.
+func LinearBackoff(base, max time.Duration) Backoff {
+	return func(n int) time.Duration {
+		d := base * time.Duration(n)
+		if d > max {
+			d = max
```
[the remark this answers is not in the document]
dackroyd: That would be better, will update!

exchanges/request/request.go
dackroyd: 100,000 bytes was chosen arbitrarily, probably needs to be adjusted based on what the typical response sizes are for cases that may need to be retried.

> Just say the body is bigger than 100k does this not drain?

Correct: a body larger than 100k will not fully drain, so a 'keep alive' connection won't be able to be reused.

In regards to the draining, my understanding is that unless the bytes are _actually_ read from the HTTP response body, the HTTP connection cannot be reused. Having a smaller limit reader means we read less of the HTTP body, and thus are more likely to not read the entire response (thus not be able to reuse the connection). This is true when we haven't read the entire response body, even if we close it.

```
// [...] It is the caller's responsibility to
// close Body. The default HTTP client's Transport may not
// reuse HTTP/1.x "keep-alive" TCP connections if the Body is
// not read to completion and closed.
```

The limit reader returning EOF doesn't mean that the entire HTTP response body has been read, as the limit reader just wraps the original reader, and returns EOF if at/beyond the limit: https://github.com/golang/go/blob/master/src/io/io.go#L450

dackroyd:
> This is a really needed update if we fall beyond any of our rate limits. Only issue I have, is with sending a request from different routines. There is no routine communication so we utilise our synchronisation for orderbook/ticker data etc, when we finally receive 429 errors from Binance, we have multiple routines sleeping in parallel but are not being concurrent which results in achieving a 418 state. This unfortunately needs to be addressed. I will add to this PR if I find a fix. Thank you again for opening this PR!

I don't think I've fully understood the case here where the limits become a problem, but I do want to understand it better. Certainly I've seen 418 errors for Binance in my own testing (but only with client side rate limits disabled). On their own, the retries and backoff are only trying to address the _current_ request (as was the existing timeout handling), and were intended to be combined with the client side rate limits to avoid any 'thundering herd' kinds of issues (which would result in further 429, and 418 failures).

Whether the retries / backoff can (or should?) try to coordinate across all concurrent requests I'm not sure: that may be unnecessary complexity if client side rate limits can take care of the rest?

exchanges/request/backoff.go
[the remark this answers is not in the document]
dackroyd: Updated 👍

exchanges/request/request.go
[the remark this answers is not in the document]
shazbert: Thanks for the link, I will resolve this!

exchanges/request/request.go
shazbert: > 100,000 bytes was chosen arbitrarily, probably needs to be adjusted based on what the typical response sizes are for cases that may need to be retried.

Thanks for clarifying, to be honest the error body going above and beyond 100k bytes to me looks like a low probability/issue anyway, but sending a request if it fails we can handle so if you can return an error here that would be better. 

@thrasher- for more input.

dackroyd: > [...] but sending a request if it fails we can handle so if you can return an error here that would be better.

I'm not sure that returning an error here is the desired behaviour, but I assume that isn't what is intended. If I'm understanding what the intention is, you would prefer _not_ draining the body at all and just closing it, having a new HTTP connection created for the next request?

Would a smaller limit be ok as an alternative? If errors mostly aren't larger than 5k (or maybe even smaller?), maybe that is a better limit to apply and have _most_ cases drain successfully, while anything larger than that will not drain fully, and force a new connection to be established in those cases.

main.go
```diff
@@ -80,7 +80,7 @@ func main() {
 	flag.BoolVar(&settings.ExchangePurgeCredentials, "exchangepurgecredentials", false, "purges the stored exchange API credentials")
 	flag.BoolVar(&settings.EnableExchangeHTTPRateLimiter, "ratelimiter", true, "enables the rate limiter for HTTP requests")
 	flag.IntVar(&settings.MaxHTTPRequestJobsLimit, "requestjobslimit", int(request.DefaultMaxRequestJobs), "sets the max amount of jobs the HTTP request package stores")
-	flag.IntVar(&settings.RequestTimeoutRetryAttempts, "exchangehttptimeoutretryattempts", request.DefaultTimeoutRetryAttempts, "sets the amount of retry attempts after a HTTP request times out")
+	flag.IntVar(&settings.RequestMaxRetryAttempts, "exchangehttpmaxretryattempts", request.DefaultMaxRetryAttempts, "sets the amount of retry attempts after a retryable HTTP failure")
```
xtda, line 83: as the requestor package and these new rate limit changes are used outside of exchanges the exchange prefix on the command (and the 3 below) can be dropped now 

What are everyone else's thoughts on this?>

exchanges/request/request.go
gloriousCode: I think the limit being 100k is fine. It will ensure that most of response bodies will be drained and a connection is allowed to be reused. The reader will stop anyway if an EOF is hit for anything smaller. I was going to request you change it to just being `io.Copy(ioutil.Discard, body)` but there's a chance of a massive body that causes delays. 100k bytes isn't very much

Also, I don't think an error needs to be returned. If draining the body fails because the response body had an unexpected EOF or that the body is too large, there's not much we can action other than increasing the limit, log that it occurred or do nothing. It just means that we'll have to make a new connection

main.go
dackroyd: Can do 👍 I'd guess that we want to remove the mention of 'exchange' from the `settings` field, the flag name and the flag description?

dackroyd:
> Code wise everything looks to be fine for what is implemented
> 
> I just have a few bits of feedback for how it works and possible changes
> 
> First one is if you are rate limited and try load another instance of GCT you will get held up waiting here:
> 
> https://github.com/dackroyd/gocryptotrader/blob/retry-and-backoff/engine/exchange.go#L371
> due to the use of waitgroup and fetching pairs on Start()
> 
> https://github.com/dackroyd/gocryptotrader/blob/retry-and-backoff/engine/exchange.go#L339-L346
> 
> Second one is a discussion that a few of us had regarding nonce generation on exchanges that use timestamp based nonces with very tight tolerances if the timeout period exceed the retry wait period as they are generated at request creation not sending
> 
> Happy to continue this over slack if needed instead of back and forth in PR comments

Yep, Slack might be easiest. I'll start a thread for this PR

main.go
dackroyd: Noticed that there seems to be some overlap between these and 'global' HTTP Timeout / 
 User Agent / Proxy?

```
gctlog.Debugf(gctlog.Global, "- COMMON SETTINGS:")
	gctlog.Debugf(gctlog.Global, "\t Global HTTP timeout: %v", s.GlobalHTTPTimeout)
	gctlog.Debugf(gctlog.Global, "\t Global HTTP user agent: %v", s.GlobalHTTPUserAgent)
gctlog.Debugf(gctlog.Global, "\t Global HTTP proxy: %v", s.ExchangeHTTPProxy)
```

dackroyd:
> This is a really needed update if we fall beyond any of our rate limits. Only issue I have, is with sending a request from different routines. There is no routine communication so we utilise our synchronisation for orderbook/ticker data etc, when we finally receive 429 errors from Binance, we have multiple routines sleeping in parallel but are not being concurrent which results in achieving a 418 state. This unfortunately needs to be addressed. I will add to this PR if I find a fix. Thank you again for opening this PR!

Do you have instructions or a test case that I can use to reproduce this issue?

xtda:
> Do you have instructions or a test case that I can use to reproduce this issue?


Are you testing with multiple exchanges enabled or just binance? 
best way to see this is enable multiple exchanges get yourself rate limited you will see orderbook/ticker updates will halt due to the sync routines behind held up by the rate limit waiting

thrasher-:
> I've reviewed your latest changes and I really like them. This PR has highlighted that there are bigger issues for us regarding things like Signature generation on retry, what requests should we actually should bother retrying and how we can lock the setup process if we mess up our rate limits.
> 
> Those things are not a result of your PR though. I have not been able to be rate limited while testing the application as a result of the client side limits put in place. The retry and backoffs do work when triggered.
> 
> Very happy with the features, the test coverage as well as the new deadline support. Thank you for your contribution!
> 
> Please rebase master and then I'll change my review to approved

I am in agreeance with this summary, will tACK once the rebase has been done. Thanks David!

go.sum
```diff
@@ -78,6 +116,36 @@ github.com/golang/protobuf v1.4.0-rc.2/go.mod h1:LlEzMj4AhA7rCAGe4KMBDvJI+AwstrU
...
+github.com/golangci/gofmt v0.0.0-20190930125516-244bba706f1a h1:iR3fYXUjHCR97qWS8ch1y9zPNsgXThGwjKPrYfqMPks=
+github.com/golangci/gofmt v0.0.0-20190930125516-244bba706f1a/go.mod h1:9qCChq59u/eW8im404Q2WWTrnBUQKjpNYKMbU4M7EFU=
+github.com/golangci/golangci-lint v1.24.0 h1:OcmSTTMPqI/VT4GvN1fKuE9NX15dDXIwolO0l08334U=
+github.com/golangci/golangci-lint v1.24.0/go.mod h1:yIqiAZ2SSQqg+1JeFlAdvEWjGVz4uu5jr4lrciqA1gE=
+github.com/golangci/ineffassign v0.0.0-20190609212857-42439a7714cc h1:gLLhTLMk2/SutryVJ6D4VZCU3CUqr8YloG7FPIBWFpI=
+github.com/golangci/ineffassign v0.0.0-20190609212857-42439a7714cc/go.mod h1:e5tpTHCfVze+7EpLEozzMB3eafxo2KT5veNg1k6byQU=
+github.com/golangci/lint-1 v0.0.0-20191013205115-297bf364a8e0 h1:MfyDlzVjl1hoaPzPD4Gpb/QgoRfSBR0jdhwGyAWwMSA=
+github.com/golangci/lint-1 v0.0.0-20191013205115-297bf364a8e0/go.mod h1:66R6K6P6VWk9I95jvqGxkqJxVWGFy9XlDwLwVz1RCFg=
+github.com/golangci/maligned v0.0.0-20180506175553-b1d89398deca h1:kNY3/svz5T29MYHubXix4aDDuE3RWHkPvopM/EDv/MA=
+github.com/golangci/maligned v0.0.0-20180506175553-b1d89398deca/go.mod h1:tvlJhZqDe4LMs4ZHD0oMUlt9G2LWuDGoisJTBzLMV9o=
```
thrasher-, line 140: Just one nit: This is an optional thing (golangci-lint), so we like to exclude it from the go depends. Please set this back to master (aside from any go mod tidy changes)

main.go
thrasher-: The difference between them is that a) the exchange related ones aim to adjust the exchange requests user agent / proxy settings / timeout as these are all exchange config adjustable and the global settings are for all other subsystems (like the forex handler, SMS gateway etc).  In a different PR, it would be good to have one settings which adjusts both, so everything is unified but out of scope for this PR IMO

go.sum
dackroyd: These slipped through in one of the updates, have removed now 😃

PR #496, opened 2020-05-03
Update to GCT script library

shazbert:
* Adds script link to GCT logger package.
* Adds ability to save data as CSV via script for easy exportation.

~~**Need help on where to save and how to save CSV if this is going to be implemented.**~~

## Type of change

## How has this been tested

- [x] error driven dev 💯 
- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


gctscript/modules/gct/common.go
```diff
@@ -0,0 +1,385 @@
...
+
+	candles, ok := obj.Value["candles"]
+	if !ok {
+		return nil, errors.New("candles not found in object map")
+	}
+
+	data := candles.Iterate()
+
+	for data.Next() {
+		var date, open, high, low, close, volume string
```
xtda, line 361: while Go doesn't restrict it "close" is a a built in name and its use for variables might cause confusion if close is ever used later

shazbert:
Just pretty printing really. Also need your thoughts on where to save, if this is something that wants to be merged?

xtda:
> Just pretty printing really. Also need your thoughts on where to save, if this is something that wants to be merged?

the lack of timestamp (and pretty printing output) was something that came up when i originally implemented scripting but ultimately it came down to am i ever really going to need to print/log the output of a script to console/file outside of testing/example scripts? which is why i decided against it

The helpers for names are a great thing to add and would like to see those added either way

We already have a CSV output method 
https://github.com/thrasher-corp/gocryptotrader/blob/master/common/common.go#L265-L280

That can be removed as the newer method has additional checks

As for the location to save maybe default to gocryptotrader datadir but allow overrides?

To me its most common usage seems like you would send a request to engine via gctcli to save some data to csv and then it would be sent via gctcli to a local client to save there which means on the engine side it would just be a tempfile anyway

shazbert:
Yeah I will fix everything up if we get consensus. I can add a gctcli command flag for it but it seemed easier as a script. I can also segregate the log calls to its out sublogger so we can just not log that output completely so its always pretty?

gctscript/modules/gct/common.go
```diff
@@ -0,0 +1,385 @@
...
+		},
+	}
+
+	obj, ok := objects.ToInterface(a).(*indicators.BBands)
+	if !ok {
+		return nil, errors.New("casting failure")
+	}
+
+	for x := range obj.Value {
+		bro := obj.Value[x].Iterate()
```
gloriousCode, line 123: 👀

gctscript/modules/gct/common.go
```diff
@@ -0,0 +1,385 @@
...
+		},
+	}
+
+	obj, ok := objects.ToInterface(a).(*indicators.MACD)
+	if !ok {
+		return nil, errors.New("casting failure")
+	}
+
+	for x := range obj.Value {
+		bro := obj.Value[x].Iterate()
```
gloriousCode, line 191: 👀

gctscript/modules/gct/logger.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	switch options {
+	case info:
+		log.Infoln(log.GCTScriptMgr, printArgs)
+	case debug:
+		log.Debugln(log.GCTScriptMgr, printArgs)
+	case warn:
+		log.Warnln(log.GCTScriptMgr, printArgs)
+	case errorbro:
```
gloriousCode, line 75: 👀

gctscript/modules/gct/common_test.go
```diff
@@ -0,0 +1,66 @@
...
+			Value: []objects.Object{
+				&objects.String{Value: "3"},
+				&objects.String{Value: "4"},
+				&objects.String{Value: "LOLOLOLOLOL"},
+			},
+		},
+	},
+}
+
+func TestCommonWriteToCSV(t *testing.T) {
```
gloriousCode, line 56: This test fails for me

thrasher-:
As discussed, I think the CSV output file should go to: gctdir/scripts/output/name of script file-output.csv

gctscript/examples/csv.gct
```diff
@@ -0,0 +1,28 @@
...
+    // construct ta values 
+    avgtr := atr.calculate(ohlcvData.candles, 14)
+    simma := sma.calculate(ohlcvData.candles, 9)
+    expma := ema.calculate(ohlcvData.candles, 9)
+    
+    // 'output' is already defined when we construct our bytecode from file. 
+    // It points to our default output file in the scripts folder.
+    // Input the rest of the fields, any order, doesn't matter. Once the script
+    // is executed it will save as csv and you can open in excel.
+    common.writeascsv(output, ohlcvData, avgtr, simma, expma)
```
xtda, line 25: common module needs to be imported

gctscript/modules/gct/common.go
```diff
@@ -0,0 +1,408 @@
...
+
+		if len(bucket) == 0 {
+			bucket = temp
+		} else {
+			for i := range temp {
+				bucket[i] = append(bucket[i], temp[i]...)
+			}
+		}
+	}
+	return nil, file.WriteAsCSV(target+".csv", bucket)
```
xtda, line 84: Should we be prepending the "output" path here ?

For example if we call this from a script:
```go
    common.writeascsv("/Users/andrew/HELLOWORLD", ohlcvData, avgtr, simma, expma)
```

It will write to /Users/andrew/HELLOWORLD.csv

and if you leave it as output it will output to current execution path/output

shazbert: How would you like it?

shazbert:
With the headers, I don't mind, I just wanted to quickly raster what params were used for different indicators. If people want them up higher I am happy with that. Let me know.

gctscript/modules/gct/common.go
xtda: Unless i am mistaken based on the response from @thrasher- above and

https://github.com/thrasher-corp/gocryptotrader/pull/496/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR1215-R1220

I thought the intent was to only allow scripts to output into ~/.gocryptotrader/scripts/output

I don't really mind either way but if this is no longer the case we can remove the creation of the output folder in checkGCTScriptConfig()

gctscript/modules/gct/common.go
```diff
@@ -0,0 +1,408 @@
...
+		case indicators.OnBalanceVolume:
+			temp, err = convertOBV(args[i])
+		case indicators.RelativeStrengthIndex:
+			temp, err = convertRSI(args[i])
+		case indicators.SimpleMovingAverage:
+			temp, err = convertSMA(args[i])
+		case indicators.OHLCV:
+			temp, err = convertOHLCV(args[i])
+			front = true
+		case "string":
```
gloriousCode, line 51: By removing the output/string case you can also not worry about weird validations like 
`common.writeascsv(output,ret,ret2,"C:\\lol\\*&#$^@$%@$#%@$#")` where the second string overrides the first, and plus with the above, it prevents the rest of the script from running

gctscript/examples/csv.gct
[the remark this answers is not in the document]
shazbert: I have added in a a condition if the target is "" it will return an error. Don't really have a good idea of handling the multiple strings, I might just set and check again if its already set then error.

gctscript/modules/gct/common.go
shazbert: Ok so I changed the way I handled it. It will always save in our scripts/output file. I also fixed a possible file transversal bug that I need you to check over. So now you can just pass in a ctx string which is defined when the script compiles or you can define the file yourself. You can actually change the file extension, so that file transversal might be bad, so let me know if you want this to always be a csv.

gctscript/examples/csv.gct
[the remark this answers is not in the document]
gloriousCode: Yeah, I think only allowing one string to set the name sounds good

shazbert:
Up to you guys if you want that to error. If the default script context is used, this shouldn't occur. But if you want to name it like kekeke.lol its going to overwrite. Which for me in testing or quickly updating that was good enough for me. Happy to change though.

gctscript/modules/gct/common.go
```diff
@@ -0,0 +1,457 @@
...
+			}
+
+			// Removes file transversal
+			target = filepath.Base(target)
+
+			// checks to see if file is context defined, if not it will allow
+			// a client defined filename and append a date, forces the use of
+			// .csv file extension
+			switch {
+			case !strings.HasSuffix(target, ".csv") && strings.Contains(target, ".gct"):
```
xtda, line 80: you could also use filepath.Ext() here to get the flle extension as an alternative

common/file/file.go
```diff
@@ -58,3 +61,33 @@ func Exists(name string) bool {
 	_, err := os.Stat(name)
 	return !os.IsNotExist(err)
 }
+
+// WriteAsCSV dumps file as CSV
```
gloriousCode, line 65: Totally minor, but I think this could be changed to `WriteAsCSV takes a table of records and writes it as CSV`/something more descriptive since its not taking in a file to dump

shazbert: Yeah good suggestion just pushed change.

gctscript/modules/gct/common_test.go
```diff
@@ -0,0 +1,174 @@
...
+		t.Fatal(err)
+	}
+
+	_, err = WriteAsCSV(&objects.String{Value: "test.gct-script-temp2"},
+		atrPayload)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = os.RemoveAll(OutputDir)
```
thrasher-, line 170: Can add this as a defer cleaup() func as the other writetocsv funcs could potentially fail without cleaning up after itself.

gctscript/modules/gct/common.go
thrasher-: Bumperino, is neater and already using the package for other helper methods

gctscript/modules/gct/common_test.go
shazbert: 🦅

shazbert:
I guess I could add in a unix timestamp to everything for this edge case, If nobody cares about filename length? @xtda

thrasher-:
> I guess I could add in a unix timestamp to everything for this edge case, If nobody cares about filename length? @xtda

This will spam output files in the gctscript output dir and then you'd have to worry about the directory size requiring to be cleaned up if you have hundreds or thousands of scripts being executed. Overwriting the existing one is OK for now and the option to skip headers + appendage would be a better solution in the future.

PR #506, opened 2020-05-16
Binance: Fix Orderbook Management

dackroyd:
With the incorrect implementation currently, orderbooks have incorrect entries in them which never get removed. This results in an unsynchronised state, having issues such as bid prices greater than the lowest ask prices

The websocket depth updates must be subscribed to before getting a snapshot. Any updates from the websocket that have earlier update ids must be discarded.

https://binance-docs.github.io/apidocs/spot/en/#how-to-manage-a-local-order-book-correctly

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Custom code monitoring orderbooks for bid/ask price overlap

## Checklist

### DIFF
--- a/exchanges/binance/binance_websocket.go
+++ b/exchanges/binance/binance_websocket.go
@@ -71,14 +71,6 @@ func (b *Binance) WsConnect() error {
 			listenKey
 	}
 
-	enabledPairs := b.GetEnabledPairs(asset.Spot)
-	for i := range enabledPairs {
-		err = b.SeedLocalCache(enabledPairs[i])
-		if err != nil {
-			return err
-		}
-	}
-
 	b.WebsocketConn.URL = wsurl
 	b.WebsocketConn.Verbose = b.Verbose
 
@@ -93,6 +85,15 @@ func (b *Binance) WsConnect() error {
 		MessageType:       websocket.PongMessage,
 		Delay:             pingDelay,
 	})
+
+	enabledPairs := b.GetEnabledPairs(asset.Spot)
+	for i := range enabledPairs {
+		err = b.SeedLocalCache(enabledPairs[i])
+		if err != nil {
+			return err
+		}
+	}
+
 	go b.wsReadData()
 	go b.KeepAuthKeyAlive()
 	return nil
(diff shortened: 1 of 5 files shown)


exchanges/binance/binance_websocket.go
```diff
@@ -428,12 +433,24 @@ func (b *Binance) SeedLocalCache(p currency.Pair) error {
...
 }
 
 // UpdateLocalCache updates and returns the most recent iteration of the orderbook
 func (b *Binance) UpdateLocalCache(wsdp *WebsocketDepthStream) error {
+	currencyPair := currency.NewPairFromFormattedPairs(wsdp.Pair, b.GetEnabledPairs(asset.Spot),
+		b.GetPairFormat(asset.Spot, true))
+	currentBook := b.Websocket.Orderbook.GetOrderbook(currencyPair, asset.Spot)
+
+	// Drop any event where u is <= lastUpdateId in the snapshot.
+	// The first processed event should have U <= lastUpdateId+1 AND u >= lastUpdateId+1.
```
dackroyd, line 448: The second and third parts here are _implied_, but not enforced.

To detect whether the 'first' event actually being processed met this condition would likely involve yet another field on the common orderbook type, which would be specific for Binance handling, hence I've avoided it.

Ensuring that the new event's `U` value is equal to the previous event's `u+1` can be done by comparing to the last update id stored against the orderbook. It's unclear what action to take if this fails however (though the Futures API equivalent would suggest re-seeding and resyncing the WS updates?).

In testing I believe I did see some cases where the `new U != old u+1`, though I've not reproduced that with the current code. Could be a timing/load factor with the servers.

dackroyd:
I think we can still get in a situation where the Orderbook is out of sync through the 'fallback to REST'. This is because websocket updates may come in and be processed against the _current_ copy of the orderbook while the fallback REST process is happening, leaving some updates missed from the snapshot

gloriousCode:
> I think we can still get in a situation where the Orderbook is out of sync through the 'fallback to REST'. This is because websocket updates may come in and be processed against the current copy of the orderbook while the fallback REST process is happening, leaving some updates missed from the snapshot

Since you're loading the orderbook via: ` b.Websocket.Orderbook.GetOrderbook(currencyPair, asset.Spot)` this should not be occurring. If we were to load the websocket via `orderbook.Get(...)` we'd run into this as `orderbook.Get(..)` retreives the latest orderbook regardless of source. But we store the latest orderbook as sourced from websocket in ` b.Websocket.Orderbook.GetOrderbook(currencyPair, asset.Spot)` for this reason

PR #507, opened 2020-05-17
Binance: Update NewOrder API

dackroyd:
Including new fields for request/response and allowing test orders.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/exchanges/binance/binance.go
+++ b/exchanges/binance/binance.go
@@ -321,14 +321,35 @@ func (b *Binance) GetBestPrice(symbol string) (BestPrice, error) {
 // NewOrder sends a new order to Binance
 func (b *Binance) NewOrder(o *NewOrderRequest) (NewOrderResponse, error) {
 	var resp NewOrderResponse
+	if err := b.newOrder(newOrder, o, &resp); err != nil {
+		return resp, err
+	}
 
-	path := b.API.Endpoints.URL + newOrder
+	if resp.Code != 0 {
+		return resp, errors.New(resp.Msg)
+	}
+
+	return resp, nil
+}
+
+// NewOrderTest sends a new test order to Binance
+func (b *Binance) NewOrderTest(o *NewOrderRequest) error {
+	var resp NewOrderResponse
+	return b.newOrder(newOrderTest, o, &resp)
+}
+
+func (b *Binance) newOrder(api string, o *NewOrderRequest, resp *NewOrderResponse) error {
+	path := b.API.Endpoints.URL + api
 
 	params := url.Values{}
 	params.Set("symbol", o.Symbol)
 	params.Set("side", o.Side)
 	params.Set("type", string(o.TradeType))
-	params.Set("quantity", strconv.FormatFloat(o.Quantity, 'f', -1, 64))
+	if o.QuoteOrderQty > 0 {
+		params.Set("quoteOrderQty", strconv.FormatFloat(o.QuoteOrderQty, 'f', -1, 64))
+	} else {
+		params.Set("quantity", strconv.FormatFloat(o.Quantity, 'f', -1, 64))
+	}
 	if o.TradeType == BinanceRequestParamsOrderLimit {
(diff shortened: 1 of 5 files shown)


exchanges/binance/binance.go
```diff
@@ -321,14 +321,35 @@ func (b *Binance) GetBestPrice(symbol string) (BestPrice, error) {
...
+
+func (b *Binance) newOrder(api string, o *NewOrderRequest, resp *NewOrderResponse) error {
+	path := b.API.Endpoints.URL + api
 
 	params := url.Values{}
 	params.Set("symbol", o.Symbol)
 	params.Set("side", o.Side)
 	params.Set("type", string(o.TradeType))
-	params.Set("quantity", strconv.FormatFloat(o.Quantity, 'f', -1, 64))
+	if o.QuoteOrderQty > 0 {
```
gloriousCode, line 348: Could you add testing coverage to this new fields in your test `TestNewOrderTest`?

exchanges/binance/binance.go
```diff
@@ -321,14 +321,35 @@ func (b *Binance) GetBestPrice(symbol string) (BestPrice, error) {
...
+		params.Set("quoteOrderQty", strconv.FormatFloat(o.Quantity, 'f', -1, 64))
```
gloriousCode, line 349: You are not converting the correct field. Should be 
`params.Set("quoteOrderQty", strconv.FormatFloat(o.QuoteOrderQty, 'f', -1, 64))`

dackroyd: 🤦 must've made this error between stashing/re-editing. Will fix it up

exchanges/binance/binance.go
```diff
@@ -321,14 +321,35 @@ func (b *Binance) GetBestPrice(symbol string) (BestPrice, error) {
...
+	} else {
+		params.Set("quantity", strconv.FormatFloat(o.Quantity, 'f', -1, 64))
+	}
```
MadCozBadd, line 352: could u also add recvWindow n timestamp to the struct plz - https://binance-docs.github.io/apidocs/spot/en/#new-order-trade

exchanges/binance/binance.go
dackroyd: Done

exchanges/binance/binance.go
dackroyd: Fixed

exchanges/binance/binance.go
```diff
@@ -321,14 +321,35 @@ func (b *Binance) GetBestPrice(symbol string) (BestPrice, error) {
...
+
+	if resp.Code != 0 {
+		return resp, errors.New(resp.Msg)
+	}
 
-	path := b.API.Endpoints.URL + newOrder
+	return resp, nil
+}
+
+// NewOrder sends a new test order to Binance
```
MadCozBadd, line 335: can u change the comment -> // NewOrderTest sends a new test order to Binance

exchanges/binance/binance.go
dackroyd: Those get added already inside `SendAuthHTTPRequest`:

```
if params == nil {
	params = url.Values{}
}
recvWindow := 5 * time.Second
params.Set("recvWindow", strconv.FormatInt(convert.RecvWindow(recvWindow), 10))
params.Set("timestamp", strconv.FormatInt(time.Now().Unix()*1000, 10))
```

I can add them to the struct if that is still desired, though I'd imagine it's better to leave the `timestamp` to be auto set, and the `recvWindow` would need to use the current value as a default unless set explicitly

MadCozBadd: ooooooooo didnt see! ur right sry bout that

dackroyd: It might be nice to be able to give a different `recvWindow` if operating under stricter (or looser?) constraints around the order timing, so it may still make sense. Might leave that for a separate discussion though 😄

PR #537, opened 2020-08-12
Exchange: Poloniex fixes

shazbert:
# PR Description

Fixes panic in websocket as the endpoint sends different types.
Fixes nonce generation issue as it requires unixnano timestamp. 

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/poloniex/poloniex.go
+++ b/exchanges/poloniex/poloniex.go
@@ -785,7 +785,7 @@ func (p *Poloniex) SendAuthenticatedHTTPRequest(method, endpoint string, values 
 	headers := make(map[string]string)
 	headers["Content-Type"] = "application/x-www-form-urlencoded"
 	headers["Key"] = p.API.Credentials.Key
-	values.Set("nonce", p.Requester.GetNonce(true).String())
+	values.Set("nonce", strconv.FormatInt(time.Now().UnixNano(), 10))
 	values.Set("command", endpoint)
 
 	hmac := crypto.GetHMAC(crypto.HashSHA512,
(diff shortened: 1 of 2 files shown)


exchanges/poloniex/poloniex.go
```diff
@@ -785,7 +785,7 @@ func (p *Poloniex) SendAuthenticatedHTTPRequest(method, endpoint string, values
 	headers := make(map[string]string)
 	headers["Content-Type"] = "application/x-www-form-urlencoded"
 	headers["Key"] = p.API.Credentials.Key
-	values.Set("nonce", p.Requester.GetNonce(true).String())
+	values.Set("nonce", strconv.FormatInt(time.Now().UnixNano(), 10))
```
gloriousCode, line 788: Would you be able to elaborate on this `Fixes nonce generation issue as it requires unixnano timestamp.`?
`p.Requester.GetNonce(true)` is a unixnano timestamp.

shazbert: Doesn't like the timestamp to be incremented needs to be within a certain threshold per call. Don't have funds on the exchange, need to do a quick limit order submissions to test if we need to add in specific padding between calls now that I think about it.

gloriousCode: This does not seem to match their documentation:
>Additionally, all queries must include a "nonce" POST parameter. The nonce parameter is an integer which must always be greater than the previous nonce used and does not need to increase by one. Using the epoch in milliseconds is an easy choice here but be careful about time synchronization if using the same API key across multiple servers.
Did you have multiple instances running by any chance?

shazbert: image

Tested with for loop deploying limit order. The update seems fine. Just need somebody else to confirm.

xtda: > Would you be able to elaborate on this `Fixes nonce generation issue as it requires unixnano timestamp.`?
> `p.Requester.GetNonce(true)` is a unixnano timestamp.

GetNonce() is technically only the current time in UnixNano the first request additional requests by the same requestor instance just increment by one this has thrown exchanges off in the past as well

PR #549, opened 2020-09-08
Config/Engine: Add data directory to config.json

Rots:
# PR Description

The provided data directory is not honored by the engine in some cases, instead a default path is used.
This change makes the data directory configuration part of config.json and enables the override using command line flag. 

## Type of change

## How has this been tested

- [x] go test ./config ./engine -race
- [x] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2037,7 +2037,7 @@ func TestRemoveExchange(t *testing.T) {
 	}
 }
 
-func TestConfig_GetDataPath(t *testing.T) {
+func TestGetDataPath(t *testing.T) {
 	tests := []struct {
 		name string
 		dir  string
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2064,6 +2064,7 @@ func TestConfig_GetDataPath(t *testing.T) {
 		},
 	}
 	for _, tt := range tests {
+		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			c := &Config{
 				DataDir: tt.dir,
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -1,6 +1,8 @@
 package config
 
 import (
+	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
@@ -2032,5 +2034,43 @@ func TestRemoveExchange(t *testing.T) {
 	}
 	if success := c.RemoveExchange("1D10TH0RS3"); success {
 		t.Fatal("exchange shouldn't exist")
+	}
+}
+
+func TestConfig_GetDataPath(t *testing.T) {
+	tests := []struct {
+		name string
+		dir  string
+		elem []string
+		want string
+	}{
+		{
+			name: "empty",
+			dir:  "",
+			elem: []string{},
+			want: common.GetDefaultDataDir(runtime.GOOS),
+		},
+		{
+			name: "empty a b",
+			dir:  "",
+			elem: []string{"a", "b"},
+			want: filepath.Join(common.GetDefaultDataDir(runtime.GOOS), "a", "b"),
+		},
+		{
+			name: "target",
+			dir:  "target",
+			elem: []string{"a", "b"},
+			want: filepath.Join("target", "a", "b"),
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := &Config{
+				DataDir: tt.dir,
+			}
+			if got := c.GetDataPath(tt.elem...); got != tt.want {
+				t.Errorf("Config.GetDataPath() = %v, want %v", got, tt.want)
+			}
+		})
 	}
 }
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2067,7 +2067,7 @@ func TestConfig_GetDataPath(t *testing.T) {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
 			c := &Config{
-				DataDir: tt.dir,
+				DataDirectory: tt.dir,
 			}
 			if got := c.GetDataPath(tt.elem...); got != tt.want {
 				t.Errorf("Config.GetDataPath() = %v, want %v", got, tt.want)
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2066,6 +2066,7 @@ func TestConfig_GetDataPath(t *testing.T) {
 	for _, tt := range tests {
 		tt := tt
 		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
 			c := &Config{
 				DataDirectory: tt.dir,
 			}
--- a/config/config_types.go
+++ b/config/config_types.go
@@ -78,7 +78,7 @@ var (
 // Exchanges
 type Config struct {
 	Name              string                  `json:"name"`
-	DataDir           string                  `json:"datadir,omitempty"`
+	DataDirectory     string                  `json:"dataDirectory"`
 	EncryptConfig     int                     `json:"encryptConfig"`
 	GlobalHTTPTimeout time.Duration           `json:"globalHTTPTimeout"`
 	Database          database.Config         `json:"database"`
--- a/config/config_types.go
+++ b/config/config_types.go
@@ -78,6 +78,7 @@ var (
 // Exchanges
 type Config struct {
 	Name              string                  `json:"name"`
+	DataDir           string                  `json:"datadir,omitempty"`
 	EncryptConfig     int                     `json:"encryptConfig"`
 	GlobalHTTPTimeout time.Duration           `json:"globalHTTPTimeout"`
 	Database          database.Config         `json:"database"`
--- a/config_example.json
+++ b/config_example.json
@@ -1,6 +1,6 @@
 {
  "name": "Skynet",
- "dataDirectory":"data",
+ "dataDirectory": "",
  "encryptConfig": 0,
  "globalHTTPTimeout": 15000000000,
  "database": {
--- a/config_example.json
+++ b/config_example.json
@@ -1,5 +1,6 @@
 {
  "name": "Skynet",
+ "dataDirectory":"data",
  "encryptConfig": 0,
  "globalHTTPTimeout": 15000000000,
  "database": {
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -120,6 +120,11 @@ func loadConfigWithSettings(settings *Settings) (*config.Config, error) {
 	}
 	// Apply overrides from settings
 	if flagSet["datadir"] {
+		// warn if dryrun isn't  enabled
+		if !settings.EnableDryRun {
+			log.Println("Command line argument '-datadir' induces dry run mode.")
+		}
+		settings.EnableDryRun = true
 		conf.DataDirectory = settings.DataDir
 	}
 
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -72,14 +72,9 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
 	}
 
 	var b Engine
-	b.Config = &config.Cfg
-	filePath, err := config.GetFilePath(settings.ConfigFile)
-	if err != nil {
-		return nil, err
-	}
+	var err error
 
-	log.Printf("Loading config file %s..\n", filePath)
-	err = b.Config.LoadConfig(filePath, settings.EnableDryRun)
+	b.Config, err = loadConfigWithSettings(settings)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load config. Err: %s", err)
 	}
@@ -95,7 +90,7 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
 		gctlog.Infoln(gctlog.Global, "Logger initialised.")
 	}
 
-	b.Settings.ConfigFile = filePath
+	b.Settings.ConfigFile = settings.ConfigFile
 	b.Settings.DataDir = settings.DataDir
 	b.Settings.CheckParamInteraction = settings.CheckParamInteraction
 
@@ -106,6 +101,25 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
 
 	ValidateSettings(&b, settings)
 	return &b, nil
+}
+
+// loadConfigWithSettings creates configuration based on the provided settings
+func loadConfigWithSettings(settings *Settings) (*config.Config, error) {
+	filePath, err := config.GetFilePath(settings.ConfigFile)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("Loading config file %s..\n", filePath)
+
+	conf := &config.Cfg
+	err = conf.ReadConfig(filePath, settings.EnableDryRun)
+	if err != nil {
+		return nil, fmt.Errorf(config.ErrFailureOpeningConfig, filePath, err)
+	}
+	//Apply overrides from settings
+	conf.DataDir = settings.DataDir
+
+	return conf, conf.CheckConfig()
 }
 
 // ValidateSettings validates and sets all bot settings
@@ -129,6 +143,12 @@ func ValidateSettings(b *Engine, s *Settings) {
 		} else {
 			b.Settings.PortfolioManagerDelay = PortfolioSleepDelay
 		}
+	}
+
+	if flagSet["datadir"] {
+		b.Settings.DataDir = s.DataDir
+	} else {
+		b.Settings.DataDir = b.Config.DataDir
 	}
 
 	if flagSet["grpc"] {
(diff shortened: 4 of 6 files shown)


engine/engine.go
```diff
@@ -95,23 +92,42 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
...
+	b.Settings.DataDir = b.Config.GetDataPath()
 	b.Settings.CheckParamInteraction = settings.CheckParamInteraction
 
 	err = utils.AdjustGoMaxProcs(settings.GoMaxProcs)
 	if err != nil {
 		return nil, fmt.Errorf("unable to adjust runtime GOMAXPROCS value. Err: %s", err)
 	}
 
-	ValidateSettings(&b, settings)
+	validateSettings(&b, settings)
```
Rots, line 104: is it OK to unexport this? It's not used in other packages in this codebase and it doesn't look like it's designed for external use

config/config_types.go
```diff
@@ -78,6 +78,7 @@ var (
 // Exchanges
 type Config struct {
 	Name              string                  `json:"name"`
+	DataDir           string                  `json:"datadir,omitempty"`
```
shazbert, line 81: DataDir -> DataDirectory; 
`json:"datadir -> `json:"dataDirectory`
I think it might be better to remove omitempty, just so it's more obvious for people to set. What do you think?

engine/engine.go
shazbert: Yeah, good spotting!

engine/engine.go
```diff
@@ -95,23 +92,42 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
...
+
+	conf := &config.Cfg
+	err = conf.ReadConfig(filePath, settings.EnableDryRun)
+	if err != nil {
+		return nil, fmt.Errorf(config.ErrFailureOpeningConfig, filePath, err)
+	}
+	// Apply overrides from settings
+	if flagSet["datadir"] {
+		conf.DataDir = settings.DataDir
+	}
```
shazbert, line 124: I don't think we should be overwriting our configuration file. This needs a bit of discussion.

shazbert (changes requested):
Thanks for opening this PR this is a great addition! I tested the config data directory and flag -datadir, just a few things flagged for discussion before I can approve.

config/config_types.go
shazbert: Also, config_example.json needs to be updated with this field as well.

engine/engine.go
thrasher-: I like the following logic:
a) If the datadir flag isn't set and the config contains no datadir value, use the default data directory and update the config to use it.
b) If the datadir flag is set, skip the config setting and use the flag value, but don't save this updated datadir in the config. The user should update the datadir setting in their config if they want a persistent datadir location

Rots: a) is already the case
b) yep, this was my initial idea as well, but... 
* without this change the -datadir isn't properly applied at all - (logs and scripts are using default regardless of the flag), so for that at least the current change is an improvement I think
* the initialisation of the directories is done in the 'config' package without any dependency on engine and Settings, this would have to be turned around and the `(c *Config) Check*()` functions would have to be in the engine instead. What is your policy on breaking changes? As this would change a lot.

config/config_types.go
Rots: 1. Renames, done.
2. remove omitempty, done.
3. For config_example.json, Not sure what to put there, the default (in user home) would be used if not set. If I set a relative directory, then it depends on the working directory of the process. Another (more intuitive?) option would be to change behaviour to resolve it to a relative directory of the config file?

engine/engine.go
```diff
@@ -95,23 +92,42 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
...
+	validateSettings(&b, settings)
 	return &b, nil
 }
 
-// ValidateSettings validates and sets all bot settings
-func ValidateSettings(b *Engine, s *Settings) {
-	flag.Visit(func(f *flag.Flag) { flagSet[f.Name] = true })
+// loadConfigWithSettings creates configuration based on the provided settings
+func loadConfigWithSettings(settings *Settings) (*config.Config, error) {
+	filePath, err := config.GetFilePath(settings.ConfigFile)
```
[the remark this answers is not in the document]
Rots: Emm... it is checked on line 111... False positive? Or am I missing something...

engine/engine.go
gloriousCode: For b) You could add `settings.EnableDryRun = true` inside the above if statement and a log warn that dryrun is enabled for when its applied. We do have `func dryrunParamInteraction("datadir")`, but Bot is not setup yet at this time to use it. Using  `settings.EnableDryRun = true` will preserve the user's settings while allowing you to have a custom datadir for that run, which properly creates directories, logs, script outputs etc.

config_example.json
```diff
@@ -1,5 +1,6 @@
 {
  "name": "Skynet",
+ "dataDirectory":"data",
```
shazbert, line 3: If someone does: `.\gocryptotrader.exe -config="config_example.json"`
It will populate a local data folder which I see your point on keeping it relative with the configuration but will leave this open for more discussion, I think for now it's probably better to just change to `"dataDirectory":"",` as this will just use default folder settings. I personally like to have multiple configurations and a centralised data directory as default if not specified.

shazbert (changes requested):
Thank you for changing the requested!

config/config_test.go
```diff
@@ -2034,3 +2036,42 @@ func TestRemoveExchange(t *testing.T) {
...
+		},
+		{
+			name: "target",
+			dir:  "target",
+			elem: []string{"a", "b"},
+			want: filepath.Join("target", "a", "b"),
+		},
+	}
+	for _, tt := range tests {
+		tt := tt
```
MadCozBadd, line 2067: need to add t.Parallel() otherwise this line doesnt really help. without t.Parallel() normal iteration should suffice

engine/engine_test.go
```diff
@@ -0,0 +1,76 @@
...
+				DataDir:      somePath,
+				EnableDryRun: true,
+			},
+			want:    &somePath,
+			wantErr: false,
+		},
+	}
+	config.TestBypass = true
+	for _, tt := range tests {
+		tt := tt
```
MadCozBadd, line 54: can change it here too

MadCozBadd (commented):
will leave the rest for the grownups. thx for the PR!

config/config_test.go
MadCozBadd, line 2067: can we add t.Parallel() here

engine/engine_test.go
MadCozBadd, line 54: this only helps when t.Parallel() is used so can we add t.Parallel() here pls

engine/engine.go
Rots: Done.

config_example.json
Rots: Done.

config/config_test.go
Rots: Done

engine/engine_test.go
Rots: It's the other way round :) if Parallel() is used, then this definitely needs to be assigned to a local variable, otherwise it also helps to:
a) prevent other potentional issues even without Parallel
b) avoid linter errors

Anyway, these are really small tests so the gain from Parallel is almost non-existant.
Added anyway.

Rots: Or apparently not...
The global variable config.Cfg is always used when reading files (not sure why or if we could change this), it's not safe to do this in parallel.

engine/engine.go
```diff
@@ -95,23 +92,47 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
...
+	log.Printf("Loading config file %s..\n", filePath)
+
+	conf := &config.Cfg
+	err = conf.ReadConfig(filePath, settings.EnableDryRun)
+	if err != nil {
+		return nil, fmt.Errorf(config.ErrFailureOpeningConfig, filePath, err)
+	}
+	// Apply overrides from settings
+	if flagSet["datadir"] {
+		// warn if dryrun isn't  enabled
```
gloriousCode, line 123: absolutely critical double space before "enabled" 😄

shazbert (approved):
Thank you very much for the work you did! Tested approved!

engine/engine_test.go
```diff
@@ -0,0 +1,76 @@
...
+	somePath := "somePath"
+	tests := []struct {
+		name     string
+		flags    []string
+		settings *Settings
+		want     *string
+		wantErr  bool
+	}{
+		{
+			name:     "empty",
```
gloriousCode, line 20: This test relies on a user's real config. As that is what gets loaded. If said user has an encrypted config or a custom datadir set, this test will fail. I don't think this scenario is needed as a result
```
2020/09/11 10:15:14 Loading config file C:\Users\gloriousCode\AppData\Roaming\GoCryptoTrader\config.json..
    engine_test.go:64: loadConfigWithSettings() error = <nil>, wantErr true
```

engine/engine_test.go
```diff
@@ -0,0 +1,76 @@
...
+		},
+		{
+			name: "invalid file",
+			settings: &Settings{
+				ConfigFile: "nonExistent.json",
+			},
+			wantErr: true,
+		},
+		{
+			name: "test file",
```
gloriousCode, line 32: I am also getting it here, which I don't understand, because the test output says it loads the testconfig, but the next line will attempt to load my real config's datadir:
```
== RUN   Test_loadConfigWithSettings/test_file
2020/09/11 10:07:58 Loading config file ../testdata/configtest.json..
Directory asdf\database does not exist.. creating.
...
    engine_test.go:71: loadConfigWithSettings() = asdf, want 
    --- FAIL: Test_loadConfigWithSettings/test_file (0.04s)
```

> configtest.json:
```
{
 "name": "Skynet",
 "encryptConfig": -1,
 "globalHTTPTimeout": 15000000000,
 },
```

>Realconfig:
```
{
 "name": "Skynet",
 "dataDirectory": "asdf",
 "encryptConfig": -1,
 "globalHTTPTimeout": 15000000000,
 },
```

To fix this, you'll need to update `testconfig.json` to include the `dataDirectory` property like you have to `config_example.json`

gloriousCode (changes requested):
Thanks for making those changes! Its looking good now, however I came across an issue with your tests.

Another thing that is actually outside of the scope of this PR, if you want to look into it, is that we have the function `common.CreateDir`, when testing this with 
```
 "dataDirectory": "lks`.?.\u003c';:-=~`adgalkjfgslkgf",
```
No error occurs inside `common.CreateDir`. So only If you feel like changing that, it would be greatly appreciated! Otherwise, I'll look into it after this, so there's no pressure at all.

Thanks again for making those changes and thanks for contributing!

engine/engine_test.go
MadCozBadd: SORRY bout that! i tested it for the comment above and when i came across another tt := tt i blindly assumed it was a similar context! but yeah t.parallel wont hurt anyways i guess so dont need to revert back or anything

engine/engine.go
Rots: :) Done.

engine/engine_test.go
Rots: Ahh... true. I run my stuff in a clean environment so the user directory doesn't exist, the same is for for travis, but I can see that if the environment is not controlled, an existing config may exist in the default location.
Removed the case.

engine/engine_test.go
Rots: Hmm... I can't 100% put a finger on it as I'm unable to reproduce, but I suspect it's to do with the shared package variables `config.Cfg`, `engine.flagSet` and (parallel?) tests. Then, same as above, this is interacting with your real config because the default config (loaded in `config.GetFilePath`) from some of the other tests leaks through to `config.Cfg` and wrong configuration is used as a result instead.
What is your exact command you are executing tests with? `make test`?

Some more thoughts/issues:
1. The current tests do already (before this change) create the default data dir paths if they don't exist (and don't clean them up). I'm not convinced setting the `dataDirectory` to a relative directory in the test config helps. Anyway, that'll create a lot of directories (in go packages) that don't get cleaned up after tests (versus just the one in the home dir at the moment).
2. why are `flagSet` and `config.Cfg` package level variables instead of proper function parameters? I've changed `engine.NewFromSettings()` to create a new config struct instead. Does that fix your test failure?
3. since some tests always fail on my local environment, it's difficult to see regressions unless I push an update, so forgive my small (not always completely tested) commits

Rots:
> No error occurs inside `common.CreateDir`
I think depending on your file system, a file/directory name is just a sequence of bytes, so it is not necessarily incorrect to have non-human-readable directory path

engine/engine_test.go
```diff
@@ -0,0 +1,71 @@
+package engine
+
+import (
+	"testing"
+
+	"github.com/thrasher-corp/gocryptotrader/config"
+)
+
+func Test_loadConfigWithSettings(t *testing.T) {
```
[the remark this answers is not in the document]
Rots: Done.

config/config_test.go
```diff
@@ -2034,3 +2036,43 @@ func TestRemoveExchange(t *testing.T) {
 		t.Fatal("exchange shouldn't exist")
 	}
 }
+
+func TestConfig_GetDataPath(t *testing.T) {
```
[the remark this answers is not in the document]
Rots: Done.

engine/engine.go
```diff
@@ -95,23 +92,47 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
...
-	flag.Visit(func(f *flag.Flag) { flagSet[f.Name] = true })
+// loadConfigWithSettings creates configuration based on the provided settings
+func loadConfigWithSettings(settings *Settings) (*config.Config, error) {
+	filePath, err := config.GetFilePath(settings.ConfigFile)
+	if err != nil {
+		return nil, err
+	}
+	log.Printf("Loading config file %s..\n", filePath)
+
+	conf := &config.Config{}
```
xtda, line 116: while I know its not deal to have globals we currently have config.Cfg which is read by GetConfig()

https://github.com/thrasher-corp/gocryptotrader/blob/091ffae8df1c858abe374bbc8119165e187f1942/config/config.go#L1818-L1820

While its mostly used in test coverage that the old LoadConfig() method there is a few cases in non-test code that use GetConfig() and return it 

https://github.com/thrasher-corp/gocryptotrader/blob/091ffae8df1c858abe374bbc8119165e187f1942/exchanges/exchange.go#L232-L244

https://github.com/thrasher-corp/gocryptotrader/blob/091ffae8df1c858abe374bbc8119165e187f1942/engine/events.go#L309-L318

Because its no longer set as part of loadConfigWithSettings it will cause some nice _unintended consequences_

engine/engine.go
```diff
@@ -95,23 +92,47 @@ func NewFromSettings(settings *Settings) (*Engine, error) {
...
+		// warn if dryrun isn't enabled
+		if !settings.EnableDryRun {
+			log.Println("Command line argument '-datadir' induces dry run mode.")
+		}
+		settings.EnableDryRun = true
+		conf.DataDirectory = settings.DataDir
+	}
+
+	return conf, conf.CheckConfig()
+}
```
xtda, line 132: This one is more of a note for myself to add dryrun checks around logging and script output because technically at the moment its not a dry run with these files being touched :D

xtda (changes requested):

engine/engine.go
Rots: OK, I switched it because of https://github.com/thrasher-corp/gocryptotrader/pull/549#discussion_r487303039
If it's just for test coverage then I would actually prefer to keep the new struct creation, as the old tests are not changed and new tests cover this function.
Switched back anyway.

engine/engine.go
Rots: Maybe have two separate flags and call the other one `--disable-config-update` or `--config-read-only` or something?

engine/engine.go
xtda: I am happy for it to be changed just the things mentioned above would also need to be updated because they would cause panics when trying to access

Rots: Yes, these functions would explicitly need access to the config from the call stack, I suspect the changes are not going to be trivial and will need some conceptual decisions.
Anyway, I think it's best to do the refactoring in a separate change, so we can leave this as is...

xtda (approved):
tested approved

shazbert (approved):
Changes tested and approved

engine/engine.go
thrasher-: Yeah I agree, dryrun is quite ambiguous under specific use cases but this can be implemented in a different PR

gloriousCode (approved):
Thanks for making those changes! Tested and its all good. Thank you for your contribution @Rots 
# 🎉

thrasher- (approved):
tACK, thanks for your 1st contribution @Rots 💯

PR #563, opened 2020-09-19
Engine: Refactor use of global Bot

Rots:
* make the dependency on having an existing running Engine more obvious
* use explicit dependency of engine in RPCServer
* reduce static dependencies in rpcserver
* improve helpers

# PR Description

This change should have no functional impact on the code. Just a Refactoring to use less of `engine.Bot` global variable.

## Type of change

It will break external code that uses static methods, but it's only a small syntactic change, replacing
`engine.GetXXX()` with `engine.Bot.GetXXX()` is the fix (e.g. `engine.Bot.GetExchangeNames(true)`)
but ideally `engine.Bot` should not be used at all, instead your local intance of the bot should be preferred.

## How has this been tested

Ran all the existing tests
- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

Is there a list somewhere?

### DIFF
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -72,7 +72,7 @@ func TestLoadConfigWithSettings(t *testing.T) {
 	}
 }
 func TestStartStopDoesNotCausePanic(t *testing.T) {
-	t.Skip("Due to race condition with global config.Cfg being used from multiple tests and RPC server and REST proxy never propoerly stopped when engine.Stop() is called")
+	t.Skip("Due to race condition with global config.Cfg being used from multiple tests and RPC server and REST proxy never properly stopped when engine.Stop() is called")
 	botOne, err := NewFromSettings(&Settings{
 		ConfigFile:   config.TestFile,
 		EnableDryRun: true,
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -72,6 +72,7 @@ func TestLoadConfigWithSettings(t *testing.T) {
 	}
 }
 func TestStartStopDoesNotCausePanic(t *testing.T) {
+	t.Skip("Due to race condition with global config.Cfg being used from multiple tests and RPC server and REST proxy never propoerly stopped when engine.Stop() is called")
 	botOne, err := NewFromSettings(&Settings{
 		ConfigFile:   config.TestFile,
 		EnableDryRun: true,
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -71,3 +71,18 @@ func TestLoadConfigWithSettings(t *testing.T) {
 		})
 	}
 }
+func TestStartStopDoesNotCausePanic(t *testing.T) {
+	botOne, err := NewFromSettings(&Settings{
+		ConfigFile:   config.TestFile,
+		EnableDryRun: true,
+	})
+	if err != nil {
+		t.Error(err)
+	}
+
+	if err = botOne.Start(); err != nil {
+		t.Error(err)
+	}
+
+	botOne.Stop()
+}
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -71,6 +71,7 @@ func TestLoadConfigWithSettings(t *testing.T) {
 		})
 	}
 }
+
 func TestStartStopDoesNotCausePanic(t *testing.T) {
 	t.Skip("Due to race condition with global config.Cfg being used from multiple tests and RPC server and REST proxy never properly stopped when engine.Stop() is called")
 	botOne, err := NewFromSettings(&Settings{
--- a/engine/exchange.go
+++ b/engine/exchange.go
@@ -355,7 +355,7 @@ func (bot *Engine) LoadExchange(name string, useWG bool, wg *sync.WaitGroup) err
 	return nil
 }
 
-// SetupExchanges sets up the exchanges used by the e
+// SetupExchanges sets up the exchanges used by the Bot
 func (bot *Engine) SetupExchanges() {
 	var wg sync.WaitGroup
 	configs := bot.Config.GetAllExchangeConfigs()
--- a/engine/exchange.go
+++ b/engine/exchange.go
@@ -139,26 +139,26 @@ func (e *exchangeManager) unloadExchange(exchangeName string) error {
 }
 
 // GetExchangeByName returns an exchange given an exchange name
-func GetExchangeByName(exchName string) exchange.IBotExchange {
-	return Bot.exchangeManager.getExchangeByName(exchName)
+func (bot *Engine) GetExchangeByName(exchName string) exchange.IBotExchange {
+	return bot.exchangeManager.getExchangeByName(exchName)
 }
 
 // UnloadExchange unloads an exchange by name
-func UnloadExchange(exchName string) error {
-	return Bot.exchangeManager.unloadExchange(exchName)
+func (bot *Engine) UnloadExchange(exchName string) error {
+	return bot.exchangeManager.unloadExchange(exchName)
 }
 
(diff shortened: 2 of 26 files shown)


engine/exchange_test.go
```diff
@@ -113,27 +113,27 @@ func TestGetExchangeByName(t *testing.T) {
 }
 
 func TestUnloadExchange(t *testing.T) {
-	SetupTestHelpers(t)
+	e := SetupTestHelpers(t)
 
-	err := UnloadExchange("asdf")
+	err := e.UnloadExchange("asdf")
```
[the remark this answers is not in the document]
Rots: it was already like this before, didn't change this

engine/exchange.go
```diff
@@ -355,16 +355,16 @@ func LoadExchange(name string, useWG bool, wg *sync.WaitGroup) error {
 	return nil
 }
 
-// SetupExchanges sets up the exchanges used by the Bot
-func SetupExchanges() {
+// SetupExchanges sets up the exchanges used by the e
```
shazbert, line 358: revert

engine/exchange_test.go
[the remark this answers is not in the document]
shazbert: It would be helpful if you did, but not stopping anything.

shazbert (changes requested):
Thanks again for the continued updates. There are some issues that I found while testing but they are not related to the change therefore out of scope: 
1) There's a panic caused by BTCMarkets wrapper in exchange_wrapper_coverage:
```go
[signal 0xc0000005 code=0x0 addr=0x0 pc=0xdd92f9]

goroutine 263 [running]:
gocryptotrader/exchanges/btcmarkets.(*BTCMarkets).WithdrawCryptocurrencyFunds(0xc000002780, 0xc000208150, 0xc000108b90, 0x0, 0x0)
        gocryptotrader/exchanges/btcmarkets/btcmarkets_wrapper.go:573 +0x79
main.testWrappers(0x1814820, 0xc000002780, 0xb, 0x715f7373616c6361, 0x63223a2265746f75)
       gocryptotrader/cmd/exchange_wrapper_coverage/main.go:178 +0xb4a
main.main.func1(0xc0001de200, 0x1c, 0x20, 0xc0006903c0, 0xc000030300, 0x1a)
        gocryptotrader/cmd/exchange_wrapper_coverage/main.go:54 +0x97
created by main.main
        gocryptotrader/cmd/exchange_wrapper_coverage/main.go:52 +0x42a
exit status 2
```

exchange wrapper issues:
```2020/09/21 10:43:04 Error loading config: 'json: cannot unmarshal number into Go struct field Bank.bankAccount.bankAccountNumber of type string', generating empty config```

Config_builder not working as intended: 
```
HitBTC cannot set pairs manager, pairs already set
2020/09/21 11:35:14 Failed to get exchanges default config. Err: exchange HitBTC: SetAPIURL error. URL vals are empty etc...
```

The only critique I have is the usage of name bot with the receiver type engine. In the future, we will most likely be spawning multiple bots via the engine. Not requesting a change.

**Review by MadCozBadd — APPROVED** [review]
not much i could find! thx for the PR m8888

File: engine/exchange.go (hunk 2 above)
Comment (Rots) [reply engine/exchange.go]: Oops, done.

**Comment by Rots** [comment]
The `UnloadEnchage()` - if err was nil, the test would panic, so the check is not really necessary. Added an explicit check to the if statement anyway.

The name bot - yes I had a though about this one as well... I decided to use it, 
1. mostly because this is how it's already used and referred to - `engine.Bot`, so we know we're dealing with the same thing, 
1.  to use`e` instead as in 
https://github.com/thrasher-corp/gocryptotrader/pull/563/files#diff-9312417691ccc0ddaba643d50097ba0cL322
and friends, would loose some readability
1. In several places the change from `Bot` to `bot` felt more natural

File: engine/exchange_test.go (hunk 1 above)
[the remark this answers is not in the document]
Comment (Rots) [reply engine/exchange_test.go:118]: The UnloadEnchage() - if err was nil, the test would panic, so the check is not really necessary. Added an explicit check to the if statement anyway.

**Review by xtda — COMMENTED** [review]
Thanks for the changes made they are helpful with moving towards future goal of independent control of bot instances 

There are a few cases of global engine.Bot usage that can be updated if you wish for example

https://github.com/thrasher-corp/gocryptotrader/blob/master/engine/restful_router.go#L35

Another example is engine subsystem's currently use engine.Bot 

https://github.com/thrasher-corp/gocryptotrader/blob/master/engine/orders.go#L174
https://github.com/thrasher-corp/gocryptotrader/blob/master/engine/orders.go#L167
https://github.com/thrasher-corp/gocryptotrader/blob/master/engine/orders.go#L220
(along with database/gctscript/portfolio)

none of which will be set if you do something like and will cause some happy little panics

```go
func main() {
	settingsOne := &engine.Settings{
		ConfigFile: "one.json",
	}
	botOne, err := engine.NewFromSettings(settingsOne)
	if err != nil {
		log.Fatal(err)
	}

	if err = botOne.Start(); err != nil {
		log.Fatal(err)
	}

	interrupt := signaler.WaitForInterrupt()
	log.Printf("Captured %v, shutdown requested.\n", interrupt)
	botOne.Stop()
}
```

I am happy to approve still if these changes go beyond what you wanted to do as there is still quiet a few instances of its usage and will get a ticket created to do another pass as this PR does not break out of box experience due to engine.Bot being set https://github.com/thrasher-corp/gocryptotrader/blob/master/main.go#L112

Rots:
> happy little panics

Added a simple test case for that.

I'm sure there's plenty of more cases where engine.Bot is used, for now I settled with this.

Rots:
> Added a simple test case for that.

Apparently there's plenty of issues to be solved before that test would work as expected (with no side effects). Skipped the test for now.

shazbert (approved):
Tested approved

engine/engine_test.go
```diff
@@ -72,6 +72,7 @@ func TestLoadConfigWithSettings(t *testing.T) {
 	}
 }
 func TestStartStopDoesNotCausePanic(t *testing.T) {
+	t.Skip("Due to race condition with global config.Cfg being used from multiple tests and RPC server and REST proxy never propoerly stopped when engine.Stop() is called")
```
shazbert, line 75: propoerly -> properly

shazbert (changes requested):
Found with recent commit.

engine/engine_test.go
Rots: Done

xtda (approved):
Going to go ahead and approve this 
In its current state it does not break current out of box experience that is supported and moves further towards separation out into engine instance

engine/engine_test.go
```diff
@@ -71,3 +71,19 @@ func TestLoadConfigWithSettings(t *testing.T) {
 		})
 	}
 }
+func TestStartStopDoesNotCausePanic(t *testing.T) {
```
gloriousCode, line 74: Could you please add one new line above here?

gloriousCode (approved):
I've tested this with gct cli, go test and running the application. Everything is operating properly and settings are being set. This is a nice update which will allow for some much easier times testing in the engine package with some further updates there. Thank you for your contribution @Rots!
# 🎉

thrasher- (approved):
Tested and working fine. This PR is a good step at refining things left over from the engine branch merge. Great stuff @Rots

engine/engine_test.go
Rots: Done. Should add a rule to linter?

thrasher-: This is covered by gofumpt which can be enabled on golangci-lint, but out of scope for this PR

PR #570, opened 2020-09-29
websocket: configuration and connection update

shazbert:
# PR Description

~~Prior implementation fetched order books when connecting this would increase the time until channel subscription by n~~ ~~HTTP calls. Missing check when listening to stream after initial sync (initial ID should be equal to the previous event's last ID +1)~~

~~- Move initial synchronisation downstream only when a WS order book update occurs.~~
~~- Buffer updates as they come in while the initial sync is occurring.~~
~~- Bleed buffer when initial sync has occurred.~~
~~- BAU when everything is a go~~

~~**Just need peeps to do a quick check, then I can refactor and optimise**~~

~~- [ ] Check reconnection, unsibscribe, resubscribe functionality via change in pairs GRPC~~
~~- [ ] Check reconnection when disconnection occurs~~

~~Might need a new connection if we exceed a check in subscribe ie all pairs enabled (Out of Scope)~~

Above was moved into its own PR for review. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [x] Ran it

## Checklist


exchanges/binance/binance_websocket.go
```diff
@@ -454,66 +455,158 @@ func (b *Binance) UpdateLocalBuffer(wsdp *WebsocketDepthStream) error {
...
-	currencyPair, err := currency.NewPairFromFormattedPairs(wsdp.Pair,
+	currencyPair, err := currency.NewPairFromFormattedPairs(u.Pair,
 		enabledPairs,
 		format)
 	if err != nil {
 		return err
 	}
 
-	currentBook := b.Websocket.Orderbook.GetOrderbook(currencyPair, asset.Spot)
+	return b.CheckAndLoadBook(currencyPair, asset.Spot, u)
```
gloriousCode, line 465: I understand that spot and margin orderbooks are the same according to Binance, but there are complications that arise from these assumptions.
- If you have both spot and margin enabled and you wish to stream the orderbook via the following command `.\gctcli.exe getorderbookstream --exchange=binance --pair=ltc-usdt --asset=spot` , only the spot version gets websocket updates
- If you only have margin enabled and run `.\gctcli.exe getorderbookstream --exchange=binance --pair=ltc-usdt --asset=margin`, you will receive `orderbook item not found for binance ltcusdt margin`, even though the orderbook exists and is getting updates.
  - Further to this, you also cannot retrieve via spot, since the asset is not enabled

exchanges/binance/binance_websocket.go
```diff
@@ -454,66 +455,158 @@ func (b *Binance) UpdateLocalBuffer(wsdp *WebsocketDepthStream) error {
...
+}
+
+// CheckAndLoadBook buffers updates, spawns workers for http book request and
+// lets slip routine, validates buffer contents and returns required work when
+// needed
+func (b *Binance) CheckAndLoadBook(c currency.Pair, ass asset.Item, u *WebsocketDepthStream) error {
```
gloriousCode, line 471: None of these functions have tests. I understand that websocket tests can become more difficult but there are a lot of things going on that we shouldn't rely on eyeballing results since we are much more likely to miss things. This PR has also caused tests to fail. The best case would be to have a simulated test, but we also have the capability to do live test if needed

exchanges/binance/binance_websocket.go
```diff
@@ -454,66 +455,158 @@ func (b *Binance) UpdateLocalBuffer(wsdp *WebsocketDepthStream) error {
...
+
+	if b.fetchingbook[c.String()] {
+		return nil
+	}
+
+	return b.BleedPipe(c, ass)
+}
+
+// BleedPipe applies the buffer to the orderbook
+func (b *Binance) BleedPipe(cp currency.Pair, a asset.Item) error {
```
gloriousCode, line 494: I think this would be more accurate to call it something like `applyBufferedUpdated`, but its not very high on the list of things :D

exchanges/binance/binance_websocket.go
```diff
@@ -454,66 +455,158 @@ func (b *Binance) UpdateLocalBuffer(wsdp *WebsocketDepthStream) error {
...
+				// u >= lastUpdateId+1.
+				if d.FirstUpdateID > id && d.LastUpdateID < id {
+					return errors.New("initial sync failure")
+				}
+				b.initialSync[cp.String()] = false
+			} else {
+				// While listening to the stream, each new event's U should be
+				// equal to the previous event's u+1.
+				if d.FirstUpdateID != id {
+					return errors.New("synchronisation failure")
```
gloriousCode, line 527: This definitely did no enjoy being disconnected and reconnected. It makes no attempt at resubscribing and will constantly pass this error to the datahandler.

It also couldn't handle disabling a pair and re-enabling. I think its not clearing the orderbook when that occurs

gloriousCode (changes requested):
This is looking good! I appreciate the effort to get Binance websocket orderbook working properly. From eyeballing, things seem to be matching. However there are a few complications and little nits. Adding new pairs via CLI worked well

xtda (commented):
I don't have anything extra to add that @gloriousCode hasn't mentioned 
My only bit of feedback is regarding the error return of subscription stream length is over 1024 

while a user _should_ know the limitations of an API these messages also fly past so quickly during the setup that its easily missed can't think of the best way to handle it from a user experience point of view at the moment

shazbert:
Components stale will cherry-pick and open another PR when I can get to it.

PR #571, opened 2020-09-30
Config: do to not store decryption variables globally

Rots:
* save session variables as unexported fields in config
* unexport internal encryption helper functions
* add encryption/decryption tests

## Type of change

- [x] Cleanup, internal refactoring

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [X] Test added

## Checklist

Still don't know what these are

### DIFF
--- a/config/config_encryption.go
+++ b/config/config_encryption.go
@@ -192,12 +192,12 @@ func getScryptDK(key, salt []byte) ([]byte, error) {
 func makeNewSessionDK(key []byte) (dk, storedSalt []byte, err error) {
 	storedSalt, err = crypto.GetRandomSalt([]byte(SaltPrefix), SaltRandomLength)
 	if err != nil {
-		return nil, storedSalt, err
+		return nil, nil, err
 	}
 
 	dk, err = getScryptDK(key, storedSalt)
 	if err != nil {
-		return nil, storedSalt, err
+		return nil, nil, err
 	}
 
 	return dk, storedSalt, nil
--- a/config/config_encryption.go
+++ b/config/config_encryption.go
@@ -27,33 +27,22 @@ const (
 	errAESBlockSize = "config file data is too small for the AES required block size"
 )
 
-var (
-	storedSalt []byte
-	sessionDK  []byte
-)
-
-// PromptForConfigEncryption asks for encryption key
-func (c *Config) PromptForConfigEncryption(configPath string, dryrun bool) bool {
+// promptForConfigEncryption asks for encryption confirmation
+// returns true if encryption was desired, false otherwise
+func promptForConfigEncryption() (bool, error) {
 	log.Println("Would you like to encrypt your config file (y/n)?")
 
 	input := ""
 	_, err := fmt.Scanln(&input)
 	if err != nil {
-		return false
+		return false, err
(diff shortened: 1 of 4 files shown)


gloriousCode (changes requested):
This change has broken the config encryption prompt. I also tested on master to ensure it wasn't an existing bug and it works there.
See below:
https://i.imgur.com/VjgMcLH.gif

Rots:
Oops, yes and no were switched around, added a test as well.

config/config_encryption_test.go
```diff
@@ -106,8 +118,182 @@ func TestRemoveECS(t *testing.T) {
...
+}
+
+// setAnswersFile sets the given file as the current stdin
+// returns the close function to defer for reverting the stdin
+func setAnswersFile(t *testing.T, answerFile string) func() {
+	oldIn := os.Stdin
+
+	inputFile, err := os.Open(answerFile)
+	if err != nil {
+		t.Fatalf("Problem opening temp file at %s: %s\n", inputFile.Name(), err)
```
xtda, line 243: There is a chance that inputFile might be nil here should this be answerFile instead?

xtda (changes requested):
prompt works correctly now just one small test related comment

config/config_encryption_test.go
Rots: Done.

config/config_encryption_test.go
```diff
@@ -106,8 +118,182 @@ func TestRemoveECS(t *testing.T) {
...
+		t.Error("Loaded conf not the same as original")
+	}
+}
+
+// setAnswersFile sets the given file as the current stdin
+// returns the close function to defer for reverting the stdin
+func setAnswersFile(t *testing.T, answerFile string) func() {
+	oldIn := os.Stdin
+
+	inputFile, err := os.Open(answerFile)
```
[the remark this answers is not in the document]
Rots: Since the file is read by other methods we can only close it after they don't expect any more input.
This method only sets it up, the close should be done by calling the returned cleanup function.

config/config_encryption.go
```diff
@@ -188,17 +189,16 @@ func getScryptDK(key, salt []byte) ([]byte, error) {
...
 	if err != nil {
-		return nil, err
+		return nil, storedSalt, err
 	}
 
-	dk, err := getScryptDK(key, storedSalt)
+	dk, err = getScryptDK(key, storedSalt)
 	if err != nil {
-		return nil, err
+		return nil, storedSalt, err
```
gloriousCode, line 200: This is really minor, but is there any reason to return the salt here? If there's an error on the other side, this storedSalt does not get used

gloriousCode (approved):
Thanks for fixing that up! Thank you for your contribution! I very much appreciate the extra tests you wrote! tACK!

xtda (approved):
👍 approved

thrasher- (approved):
tACK, tested config encryption/decryption successfully! Thanks for the additional test coverage too 💯

config/config_encryption.go
Rots: Done.

PR #575, opened 2020-10-08
Config: fix don't create empty dir when resolving path

Rots:
# PR Description

Resolving paths shouldn't really edit file system

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test added

## Checklist

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1534,6 +1534,9 @@ func migrateConfigToDefaultPath(configfile string) (string, error) {
 	} else {
 		target = filepath.Join(newDir, File)
 	}
+	if configfile == target {
+		return configfile, nil
+	}
 	if file.Exists(target) {
 		log.Warnf(log.ConfigMgr, "config file already found in '%s'; not overwriting, defaulting to %s", target, configfile)
 		return configfile, nil
--- a/config/config.go
+++ b/config/config.go
@@ -1516,12 +1516,12 @@ func GetFilePath(configfile string) (configPath string, isImplicitDefaultPath bo
 // migrateConfigToDefaultPath will move the config file to the default
 // config directory as `File` or `EncryptedFile` depending on whether the config
 // is encrypted
-func migrateConfigToDefaultPath(configfile string) (string, error) {
-	if !file.Exists(configfile) {
-		return "", fmt.Errorf("file %s does not exist", configfile)
+func migrateConfigToDefaultPath(configFile string) (string, error) {
+	if !file.Exists(configFile) {
+		return "", fmt.Errorf("file %s does not exist", configFile)
 	}
 
-	data, err := ioutil.ReadFile(configfile)
+	data, err := ioutil.ReadFile(configFile)
 	if err != nil {
 		return "", err
 	}
@@ -1534,15 +1534,15 @@ func migrateConfigToDefaultPath(configfile string) (string, error) {
 	} else {
 		target = filepath.Join(newDir, File)
 	}
-	if configfile == target {
-		return configfile, nil
+	if configFile == target {
+		return configFile, nil
 	}
 	if file.Exists(target) {
-		log.Warnf(log.ConfigMgr, "config file already found in '%s'; not overwriting, defaulting to %s", target, configfile)
-		return configfile, nil
+		log.Warnf(log.ConfigMgr, "config file already found in '%s'; not overwriting, defaulting to %s", target, configFile)
+		return configFile, nil
 	}
 
-	err = file.Move(configfile, target)
+	err = file.Move(configFile, target)
 	if err != nil {
 		return "", err
 	}
--- a/config/config.go
+++ b/config/config.go
@@ -1467,115 +1467,93 @@ func DefaultFilePath() string {
 	return f
 }
 
+// GetAndMigrateDefaultPath returns the target config file
+// migrating it from the old default location to new one,
+// if it was implicitly loaded from a default location and
+// wasn't already in the correct 'new' default location
+func GetAndMigrateDefaultPath(configFile string) (string, error) {
+	filePath, wasDefault, err := GetFilePath(configFile)
+	if err != nil {
+		return "", err
+	}
+	if wasDefault {
+		return migrateConfigToDefaultPath(filePath)
+	}
+	return filePath, nil
+}
+
 // GetFilePath returns the desired config file or the default config file name
-// based on if the application is being run under test or normal mode. It will
-// also move/rename the config file under the following conditions:
-// 1) If a config file is found in the executable path directory and no explicit
-//    config path is set, plus no config is found in the GCT data dir, it will
-//    move it to the GCT data dir. If a config already exists in the GCT data
-//    dir, it will warn the user and load the config found in the GCT data dir
-// 2) If a config file in the GCT data dir has the file extension .dat but
-//    contains json data, it will rename to the file to config.json
-// 3) If a config file in the GCT data dir has the file extension .json but
-//    contains encrypted data, it will rename the file to config.dat
-func GetFilePath(configfile string) (string, error) {
+// and whether it was loaded from a default location (rather than explicitly specified)
+func GetFilePath(configfile string) (configPath string, isImplicitDefaultPath bool, err error) {
 	if configfile != "" {
-		return configfile, nil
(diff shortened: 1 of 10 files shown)


MadCozBadd (approved):
nothing much from me! Thanks for the PR

xtda (commented):
I don't think this should be removed (or GetFilePath() refactored further)

In all current execution paths this call wont get hit because a path is passed into all calls to GetFilePath()

But this directory creation was part of config upgrade handling (along with renaming encrypted config) 

```go
	newDir := common.GetDefaultDataDir(runtime.GOOS)
```

there is a chance that GetDefaultDataDir() will return a folder that does not exist and break the rest of execution and existing functionality

Rots:
> I don't think this should be removed (or GetFilePath() refactored further

I disagree, getters with side effects is bad design, especially when that's an exported method

> there is a chance that GetDefaultDataDir() will return a folder that does not exist and break the rest of execution and existing functionality

* `GetFilePath("")` returns an error if the default config file does not exist, the existence of the parent directory doesn't make a difference
* the `file.Write()` - which is used to store the config file -  creates the parent directories, if necessary
* if a config file exists in the `oldDirs` https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1486-L1494 then the new directory is created, but not even used
* the `(*Config)check*` methods create the subdirectories if necessary, i.e. https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1305-L1306
https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1327-L1328
https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1364-L1365

Is there any other aspects?

xtda:
> > I don't think this should be removed (or GetFilePath() refactored further
> 
> I disagree, getters with side effects is bad design, especially when that's an exported method
> 
> > there is a chance that GetDefaultDataDir() will return a folder that does not exist and break the rest of execution and existing functionality
> 
>     * `GetFilePath("")` returns an error if the default config file does not exist, the existence of the parent directory doesn't make a difference
> 
>     * the `file.Write()` - which is used to store the config file -  creates the parent directories, if necessary
> 
>     * if a config file exists in the `oldDirs` https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1486-L1494
>        then the new directory is created, but not even used
> 
>     * the `(*Config)check*` methods create the subdirectories if necessary, i.e. https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1305-L1306
>       
>       https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1327-L1328
>       
>       https://github.com/thrasher-corp/gocryptotrader/blob/fcf01cc1367c9150bcf9e17c8d43a67a3e86d425/config/config.go#L1364-L1365
> 
> 
> Is there any other aspects?

This was more I feel that GetFilePath() should be refactored further to remove these additional checks out into their own method not that it shouldn't be refactored at all wasn't entirely clear on that

its intended functionality has been broken for a bit and it no longer handles the config move and rename cleanly

Rots:
I see... the `file.Move` method doesn't create directories. I'll add that and refactor the migration functionality.

thrasher- (commented):
Very much appreciate the work being done here on the config loading! I noticed that if GCT is ran as default, the application will set the config flag var to config.DefaultFilePath(). The issue is that when run under this context, `GetAndMigrateDefaultPath` will always interpret `wasDefault` as false since https://github.com/Rots/gocryptotrader/blob/config_read/config/config.go#L1488 will detect that the configFile != "" and thus return `isImplicitDefaultPath` as false even though it actually is a default path so the rest of the migration logic will be bypassed (for example, if you have a `config.json` file which is encrypted in the default data directory, it won't be moved to `config.dat` and vice versa. The other changes look good!

Rots:
Not entirely sure what the desired behaviour is (as the current behaviour was already like this), but I unified the default config file lookup, and changed so that the migration is applied when no -config flag is passed

shazbert (approved):
Tested; changes are good. 

Tested master as well and you are correct we are not changing our config file if it contains encrypted data -> config.dat and back to config.json if it's unencrypted. 

Functionality seems to have been broken. It would be nice if you could update, but not holding me up from approving this.

config/config.go
```diff
@@ -1457,129 +1457,104 @@ func (c *Config) CheckConnectionMonitorConfig() {
...
+}
 
-		if ConfirmECS(data) {
-			if filepath.Ext(newDirs[x]) == ".dat" {
-				return newDirs[x], nil
-			}
+// migrateConfigToDefaultPath will move the config file to the default
+// config directory as `File` or `EncryptedFile` depending on whether the config
+// is encrypted
+func migrateConfigToDefaultPath(configfile string) (string, error) {
```
gloriousCode, line 1519: I think it would be good to add coverage to this function

gloriousCode (changes requested):
Thanks for making all those changes! The old functionality where GCT will create the settings folder and move the config.json from the execution path has been restored! 🎉

thrasher-:
> Not entirely sure what the desired behaviour is (as the current behaviour was already like this), but I unified the default config file lookup, and changed so that the migration is applied when no -config flag is passed

This was the original intention: https://github.com/thrasher-corp/gocryptotrader/blob/master/config/config.go#L1473 but some of these things were broken in master like you identified. Thanks for extending the migration when no -config flag is passed!

config/config.go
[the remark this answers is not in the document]
Rots: Done

Rots: Yes, I also thought so, except there are still plenty of (other) tests that don't clean up after themselves (leave things in the default data directory), and we'd have to skip the migration tests to not risk people losing their actual config

Rots:
> Functionality seems to have been broken. It would be nice if you could update

@shazbert is the "migrate when `-config` flag is not set" enough or do you have something else in mind?

config/config.go
gloriousCode: Fair enough. What are your thoughts for changing the function signature to `migrateConfig(configFile, destinationDir string)`? Which will allow for this to become more testable and not interfere with other tests

shazbert:
> > Functionality seems to have been broken. It would be nice if you could update
> 
> @shazbert is the "migrate when `-config` flag is not set" enough or do you have something else in mind?
> 
> as a user I wouldn't expect my config file to be moved if I start the app with specifying my current config (i.e. I expect that running the same command again would produce the same result), which would not be the case if we migrate "the default config file" even when setting the `-config` flag

What you have done is perfect and that functionality is logical, I don't have anything else in mind. I was more meaning changing the file extension from .json to .dat due to it being encrypted and back again. But that is out of scope, it was broken prior.  
For reference: https://github.com/thrasher-corp/gocryptotrader/blob/master/config/config.go#L1477

So still tACK from me. Thanks again for the continuous awesome updates!

config/config.go
Rots: Done

Rots:
> changing the file extension from .json to .dat due to it being encrypted and back again

yes, that happens now after these changes if you don't specify `-config`

thrasher- (approved):
You're the man @Rots ! Tested existing config.json in executable dir with no config specified: 

- it moves it successfully to the default data directory
- if the config file is encrypted and file extension is .json, it will move it to .dat
- if the config file is non-encrypted and ends in .dat, moves it to .json

Thanks for restoring functionality of this behavior!

shazbert (approved):
All new changes tested approved!

xtda (approved):

gloriousCode (approved):
Thanks for making those changes with the testing! tACK! Thank you for your contribution Rots!

PR #577, opened 2020-10-08
Config: refactor config file loaders

Rots:
# PR Description
* add a layer of abstraction so that config can be loaded from non-files
* use io.Reader / io.Writer abstraction to separate data operations from
file operations
* remove dryrun option from SaveConfig - now it always saves

## Type of change

- [x] Refactoring

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Added tests for reader

## Checklist


config/config.go
```diff
@@ -1545,94 +1547,125 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
 		return err
 	}
-
-	fileData, err := ioutil.ReadFile(defaultPath)
+	confFile, err := os.Open(defaultPath)
 	if err != nil {
 		return err
 	}
+	defer confFile.Close()
+	result, wasEncrypted, err := ReadConfig(confFile, func() ([]byte, error) { return PromptForConfigKey(false) })
```
gloriousCode, line 1564: I think it would be better to rename one of these functions. Having two `ReadConfig`'s in the one package that do different things will make it confusing, even if one is belonging to `Config`

config/config_encryption.go
```diff
@@ -41,6 +42,11 @@ func promptForConfigEncryption() (bool, error) {
 	return common.YesOrNo(input), nil
 }
 
+// Unencrypted provides the default key provider implementation for unencrypted files
+func Unencrypted() ([]byte, error) {
```
gloriousCode, line 46: The only place this is used is in tests and seems like it could just be made inline rather than added the GCT binary on compilation

config/config.go
```diff
@@ -1545,94 +1547,125 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
-	errCounter := 0
-	for {
-		if errCounter >= maxAuthFailures {
-			return errors.New("failed to decrypt config after 3 attempts")
-		}
-		key, err := PromptForConfigKey(false)
+// ReadConfig verifies and checks for encryption and loads the config.
+// Prompts for decryption key, if target data is encrypted.
+// Returns the loaded configuration and whether it was encrypted.
+func ReadConfig(configReader io.Reader, keyProvider func() ([]byte, error)) (*Config, bool, error) {
```
gloriousCode, line 1596: Does this need to be exported?

config/config.go
```diff
@@ -1545,94 +1547,125 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
+	if err != nil {
+		return err
+	}
+	defer writer.Close()
+	return c.Save(writer, func() ([]byte, error) { return PromptForConfigKey(true) })
+}
 
+// Save saves your configuration to the writer
+// with encryption, if configured
+func (c *Config) Save(configWriter io.Writer, keyProvider func() ([]byte, error)) error {
```
gloriousCode, line 1668: This seems like it should be unexported for the same reason of `ReadConfig` and `ReadConfig`, causing confusion

gloriousCode (changes requested):
Interesting PR! I have a few comments on it

config/config.go
Rots: The new ReadConfig is a thread-safe constructor (alternative to the weird singleton approach `c:=&config.Cfg; c.ReadConfig(...)`) for creating config objects. What do you propose? ReadJson? NewFromJson?

config/config_encryption.go
Rots: Yep, I thought about that as well, but decided to include it in the config package (rather than just test file) for convenience for the "other consumers" of the `ReadConfig` method. If you disagree, I'm OK to remove, but I do prefer to keep it. See next comment.

config/config.go
Rots: The whole point of this change was to enable loading configuration from non-files and this would be the entry point for these use cases, so this is actually intentionally exported.
Imagine being able to load config from (configuration) database, some custom secure storage, cloud service or a unit test - the ability to not having to create temporary files (and messing with the Stdin for encryption key) just for loading config is useful.

config/config.go
Rots: again exported intentionally, see previous comment

config/config.go
gloriousCode: With the context of your other comments, I think perhaps `ReadConfigFromFile`?

config/config.go
gloriousCode: Then I think `SaveConfig` needs to become `SaveConfigToFile`

config/config.go
Rots: renamed the old ReadConfig

config/config.go
Rots: Done

common/file/file.go
```diff
@@ -24,6 +24,19 @@ func Write(file string, data []byte) error {
 	return ioutil.WriteFile(file, data, 0770)
 }
 
+// Writer creates a writer to a file or returns an error if it fails. This
+// func also ensures that all files are set to this permission (only rw access
+// for the running user and the group the user is a member of)
+func Writer(file string) (*os.File, error) {
```
shazbert, line 30: can we add a test for this please?

config/config.go
```diff
@@ -1545,94 +1547,125 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
-	if !ConfirmECS(fileData) {
-		err = json.Unmarshal(fileData, c)
-		if err != nil {
-			return err
-		}
-
-		if c.EncryptConfig == fileEncryptionDisabled {
-			return nil
-		}
+	if dryrun || wasEncrypted || (!wasEncrypted && c.EncryptConfig == fileEncryptionDisabled) {
```
shazbert, line 1571: remove `!wasEncrypted &&`

config/config.go
```diff
@@ -1545,94 +1547,125 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
 
-		if c.EncryptConfig == fileEncryptionPrompt {
-			confirm, err := promptForConfigEncryption()
-			if err == nil {
-				if confirm {
-					c.EncryptConfig = fileEncryptionEnabled
-					return c.SaveConfig(defaultPath, dryrun)
-				}
+	if c.EncryptConfig == fileEncryptionPrompt {
+		confirm, err := promptForConfigEncryption()
```
shazbert, line 1576: Can we please check this error and return? Or maybe drop the error return altogether.

config/config.go
shazbert: Yeah, I don't mind it being exported, it might be nice but out of scope for this; add a quick GRPC endpoint to upload a configuration file and flush all settings to utilise this for future implementations and or maybe even a live reload situation if enabled. This is pretty cool, endless config possibilities.

shazbert (changes requested):
This is great! just some minor nits. Thanks for adding in all those tests.

common/file/file.go
Rots: Done

config/config.go
Rots: Done

config/config.go
Rots: This was the behaviour before - it (or the lack of error) is already checked, I guess the returned error is ignored (not propagated), because even if the user prompt for encryption fails (e.g. running in some non-interactive mode), we've already loaded the configuration and it can be used, but yes I agree the user is entitled to at least log message if they've requested for encryption prompt (and at this stage have not confirmed plain text or encryption option), but it failed. 
A similar case is a few lines below, after they've confirmed plaintext option, but the choice cannot be saved (e.g. file permissions problems)

"Dropping the error" option I don't like as we'd want to repeat the prompt next time instead of defaulting to not encrypted.

shazbert: What you have done is reasonable. Thank you very much!

shazbert (approved):
tACK

gloriousCode (approved):
tACK! Thank you for your contribution. I appreciate the tests, the abstraction and the potential for more capabilities.

shazbert (changes requested):
Could you please merge master changes.

thrasher- (changes requested):
Refactoring looks good! Just found the following edge cases:

image


If I press y to prompt for config encryption and press enter on Windows, an unexpected newline error will be thrown and the config will be purged entirely:

image

The same behavior happens if I press ctrl+C to terminate the program. On master, these things should generate an error also, but the config would be left preserved.

config/config.go
```diff
@@ -1545,94 +1547,133 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
+			writer.Close()
+		}
+	}()
+	return c.Save(provider, func() ([]byte, error) { return PromptForConfigKey(true) })
+}
 
+// Save saves your configuration to the writer as a JSON object
+// with encryption, if configured
+// If there is an error when preparing the data to store, the writer is never requested
+func (c *Config) Save(writerProvider func() (io.Writer, error), keyProvider func() ([]byte, error)) error {
```
Rots, line 1676: not sure what's the "nicest" approach.
Currently just deferred creating the writer.
Alternatively could juggle with .tmp files and leave the Save() method signature as previously

config/config.go
```diff
@@ -1657,17 +1657,23 @@ func (c *Config) SaveConfigToFile(configPath string) error {
 	if err != nil {
 		return err
 	}
-	writer, err := file.Writer(defaultPath)
-	if err != nil {
-		return err
+	var writer *os.File
+	provider := func() (io.Writer, error) {
+		writer, err = file.Writer(defaultPath)
```
shazbert, line 1662: Just a suggestion but we could return an io.WriteCloser and call defer close after the error check from the calling fuction.

shazbert (changes requested):
This is pretty much a tACK from me.

config/config.go
Rots: mm... I'm not convinced... `bytes.Buffer` for instance doesn't have the `Close()` implemented. I think leaving it as is is more generic.

shazbert: That's a really good point! My mistake. Thanks.

shazbert (approved):
Tested approved. 👍

gloriousCode (approved):
My config didn't get wiped when closing the application on entering encryption keys. tACK!

config/config.go
thrasher-: I think this is good as is

thrasher- (approved):
tACK, thanks for addressing the previous concerns. Config doesn't get purged anymore 👍

PR #580, opened 2020-10-14
Engine/GCTScript: Refactor script manager

Rots:
* remove singleton GCTScriptConfig
* create constant for ".gct" extension
* move GctScriptManager into vm package
* reduce script manager global dependencies
* use manager struct to store runtime override values
* enable/disable scripting subsystem now doesn't store the setting in
config (aligned with other subsystems)
* setting max VMs via start option doesn't change config

# PR Description

Reducing singletons, cleaning up dependencies. 
Script manager does not need live in the bloated engine package.

Also, changes to runtime configuration were automatically applied to configureation:
`go run main.go -config config_example.json -gctscriptmanager=false`
and 
`go run main.go -config config_example.json -maxvirtualmachines 3`
updates the configuration file when shut down, but 
`go run main.go -config config_example.json -grpc=false`
doesn't.

With this change, the updates to gct script manager status and maxVMs are also runtime-only

## Type of change

- [x] code clean up

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


gctscript/vm/manager.go
```diff
@@ -0,0 +1,124 @@
...
+func (g *GctScriptManager) Started() bool {
+	return atomic.LoadInt32(&g.started) == 1
+}
+
+// Start starts gctscript subsystem and creates shutdown channel
+func (g *GctScriptManager) Start(config *Config, wg *sync.WaitGroup) (err error) {
+	g.config = config
+	if !g.IsEnabled() {
+		return errors.New("manager has been disabled")
+	}
```
gloriousCode, line 38: Overwriting the config before performing any checks seems dangerous. It could already be running, this gets called and now the existing script manager has had all its settings changed and likely to have created a data race or a panic because nil was sent in.

gctscript/vm/manager.go
```diff
@@ -0,0 +1,124 @@
...
+		if err != nil {
+			log.Errorf(log.GCTScriptMgr, "%v failed to load: %v", filepath.Base(scriptPath), err)
+			continue
+		}
+		go temp.CompileAndRun()
+	}
+}
+
+// IsEnabled returns whether the manager is enabled
+func (g *GctScriptManager) IsEnabled() bool {
```
gloriousCode, line 114: This function needs data race protection, we have GRPC CLI functions that can enable and disable the use of scripts and a data race can occur. Having this dedicated function to check is good, but the value can be changed while reading.

gloriousCode (changes requested):
I like these changes. There are two issues here that stand out to me

gctscript/vm/manager.go
Rots: created a constructor/factory method for script manager

gctscript/vm/manager.go
Rots: I'm now thinking we don't actually need the extra enable flag and just use `Started()` to check this...
It's simpler if the `Enabled` flag is only considered on the Engine level to decide whether to start the script manager and other decisions are based on the run status, not the enablement.

Rots:
Tests need a review

gloriousCode (changes requested):
This is looking good, thanks for making those changes! I did come across a panic though in testing:
```
.\gocryptotrader.exe -maxvirtualmachines=1

2020/10/22 11:46:47 Loading config file C:\Users\gloriousCode\AppData\Roaming\GoCryptoTrader\config.json..
[INFO] |  22/10/2020 11:46:47  | Logger initialised.
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x18f6dd2]

goroutine 1 [running]:
github.com/thrasher-corp/gocryptotrader/engine.validateSettings(0xc000378500, 0xc000346500, 0xc00047beb0)
        C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/engine.go:184 +0x6d2
github.com/thrasher-corp/gocryptotrader/engine.NewFromSettings(0xc000346500, 0xc00047beb0, 0x1d3d8e4, 0x6, 0x2550b40)
        C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/engine.go:99 +0x265
main.main()
        C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/main.go:122 +0x17ee
```

shazbert (changes requested):
Just a couple of things I found that could be addressed.  **Not a change request** but I don't exactly agree with removing gctscript.go from the engine because start-stop services were coupled with the engine for readability but it's not entirely important, as long as moving forward we can decouple everything sufficiently to this standard, then no issues. 

Thanks for the PR, what you're doing is awesome.

gctscript/vm/manager.go
```diff
@@ -0,0 +1,119 @@
...
+	started  int32
+	stopped  int32
+	shutdown chan struct{}
+	// Optional values to override stored config ('nil' if not overridden)
+	MaxVirtualMachines *uint8
+}
+
+// NewManager creates a new instance of script manager
+func NewManager(config *Config) (*GctScriptManager, error) {
+	return &GctScriptManager{
```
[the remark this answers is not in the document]
Rots: Ouch, the nil pointer was actually because of the order of creation of the manager vs updating the maxVMs field. Added a check here as well anyway.

engine/engine.go
```diff
@@ -57,6 +57,12 @@ func New() (*Engine, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to load config. Err: %s", err)
 	}
+	var scriptManager *gctscript.GctScriptManager
+	scriptManager, err = gctscript.NewManager(&b.Config.GCTScript)
```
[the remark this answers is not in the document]
Rots: Yes, not necessary.

engine/engine.go
```diff
@@ -91,6 +97,14 @@ func NewFromSettings(settings *Settings, flagSet map[string]bool) (*Engine, erro
 	}
 
 	validateSettings(&b, settings, flagSet)
+
+	var scriptManager *gctscript.GctScriptManager
+	scriptManager, err = gctscript.NewManager(&b.Config.GCTScript)
```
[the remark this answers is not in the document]
Rots: Done.

shazbert (approved):
tACK - Had to merge master, ran a script, deployed correctly. 
Thank you!

gctscript/vm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+	config   *Config
+	started  int32
+	stopped  int32
+	shutdown chan struct{}
+	// Optional values to override stored config ('nil' if not overridden)
+	MaxVirtualMachines *uint8
+}
+
+// NewManager creates a new instance of script manager
+func NewManager(config *Config) (*GctScriptManager, error) {
```
gloriousCode, line 28: I think the last thing I'd like is that this file gets test coverage

gctscript/vm/manager.go
```diff
@@ -0,0 +1,123 @@
...
+			continue
+		}
+		var name = g.config.AutoLoad[x]
+		if filepath.Ext(name) != common.GctExt {
+			name += common.GctExt
+		}
+		scriptPath := filepath.Join(ScriptPath, name)
+		err := temp.Load(scriptPath)
+		if err != nil {
+			log.Errorf(log.GCTScriptMgr, "%v failed to load: %v", filepath.Base(scriptPath), err)
```
gloriousCode, line 110: So this issue is outside of the scope of your PR, but it can be a "while you're here" 😄 completely optional and you don't have to fix this as it exists on `master` branch. 

An issue I've found is if there is an issue loading a script, for example that doesn't exist, it will still be loaded.
My config contains three scripts that don't exist:
```
"auto_load": [
   "hello",
   "mr",
   "rots"
  ],
```
Will keep 3 scripts loaded, even if they aren't found. If I then execute the following:
`gocryptotrader\cmd\gctcli> .\gctcli.exe script status`
I will get:
```
{
 "status": "3 of 10 virtual machines running",
 "scripts": [
  {
   "UUID": "fdc8f1f7-ffe3-4041-be0f-e6c997d2b6f9",
   "name": ".",
   "next_run": "0001-01-01 00:00:00 +0000 UTC"
  },
  {
   "UUID": "f4822f78-1cd7-4ae9-9995-e80f655df5b6",
   "name": ".",
   "next_run": "0001-01-01 00:00:00 +0000 UTC"
  },
  {
   "UUID": "ace9d470-3536-4b0d-a751-ac83741cffb7",
   "name": ".",
   "next_run": "0001-01-01 00:00:00 +0000 UTC"
  }
 ]
}
```

So to fix this one, what I'm suggesting is that when this error is hit on this line, that you also add
`err = temp.unregister()` so we can free up the virtual machines from error'd scripts that won't do anything

gloriousCode (changes requested):
Thanks for making those change rots! I have one last request and one optional request and I think this is good to go!

gctscript/vm/manager.go
Rots: Done

gctscript/vm/manager.go
Rots: Done

gloriousCode (approved):
Thank you for making those last changes. I've retested scripting functionality and its all good. Thank you very much for your contribution @Rots! 
# 🎉

shazbert (approved):
re-tACK. Fine work!

MadCozBadd (approved):
tACK! ran script successfully

thrasher- (approved):
Generally I'd like all subsystem start/stop/running etc. to be maintained by engine for consistency but I can see the point of abstracting it out for other use cases so am ultimately okay with this PR. tACK, thanks for the PR!

PR #588, opened 2020-11-07
Binance: implement get historic trades

Rots:
* get binance trade data based on aggregate trade list
* fix small issue in rpc server: gctcli stops retrieving when there's a gap in data

## Type of change

## How has this been tested
- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test adjusted

## Checklist


gloriousCode (changes requested):
Thanks Rots! Thanks for patching up the rpcserver stream. I don't know why I didn't add `GetHistoricTrades` support to Binance since this demonstrates you can get trade information from years back.  
# 🤷
Anyway, just a few comments.

shazbert (commented):
LGTM!

exchanges/binance/binance.go
```diff
@@ -161,27 +162,89 @@ func (b *Binance) GetHistoricalTrades(symbol string, limit int, fromID int64) ([
...
+		params.Set("fromId", strconv.FormatInt(arg.FromID, 10))
+	}
+	if !arg.StartTime.IsZero() {
+		params.Set("startTime", strconv.FormatInt(convert.UnixMillis(arg.StartTime), 10))
+	}
+	if !arg.EndTime.IsZero() {
+		params.Set("endTime", strconv.FormatInt(convert.UnixMillis(arg.EndTime), 10))
+	}
+	// If startTime and endTime are sent, time between startTime and endTime must be less than 1 hour.
+	if arg.Limit == 0 && !arg.StartTime.IsZero() && !arg.EndTime.IsZero() && arg.EndTime.Sub(arg.StartTime) > time.Hour {
```
MadCozBadd, line 189: theoretically if someone wanted to get a maximum of 500 trades OR all trades within past 2 hours, then this if condition becomes an issue since you are restricting people to take all the trades within a given timeframe or not setting a timeframe.

exchanges/binance/binance.go
```diff
@@ -161,27 +162,89 @@ func (b *Binance) GetHistoricalTrades(symbol string, limit int, fromID int64) ([
...
+			return resp, err
+		}
+	}
+
+	// other requests follow from the last aggregate trade id and have no time window
+	params.Del("startTime")
+	params.Del("endTime")
+	for {
+		// Keep requesting new data after last retrieved trade
+		params.Add("fromId", strconv.FormatInt(resp[len(resp)-1].ATradeID, 10))
```
MadCozBadd, line 227: params.Add adds a new fromId to a map causing it to set 2 fromIds. either delete the previous id or params.Set() which i think overwrites it (atleast i think so):

 https://api.binance.com/api/v3/aggTrades?fromId=412630625&fromId=412631624&limit=1000&symbol=BTCUSDT
Binance request type: GET
    c:\dev\gocryptotrader\exchanges\binance\binance_test.go:104: Binance GetAggregatedTrades() error Binance unsuccessful HTTP status code: 500 raw response: {"code":-1000,"msg":"An unknown error occured while processing the request."}

MadCozBadd (commented):
Thanks a lot for the PR m88888888888888888!

exchanges/binance/binance_types.go
```diff
@@ -200,6 +200,18 @@ type HistoricalTrade struct {
 	IsBestMatch  bool    `json:"isBestMatch"`
 }
 
+// AggregatedTradeRequestParams holds request params
+type AggregatedTradeRequestParams struct {
+	Symbol currency.Pair // Required field; example LTCBTC, BTCUSDT
```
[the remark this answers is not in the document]
Rots: hmm.. really? I think using the currency.Pair keeps the "type safety", the same reason there's `time.Time` in the start and end times, not just `int64` or a `string` (everything gets converted to string for the url parameter in the end, but as the caller I shouldn't care if the timestamp is encoded in milliseconds or seconds; or if the currency pair is encoded with '-' as the separator or empty string) - i.e. normally you want to push the "safe types" all the way down so the API would be easier to understand and you can do the conversion from the "stronger type" to "weaker one" in a single place only when you need it. 
I do recognise that might be an extra dependency the callers might need, but the dependency is already there for the package and I believe it is small price to pay for extra safety.
If you still disagree and think that the exchange API should be as generic as possible, I'm OK to change it

exchanges/binance/binance.go
```diff
@@ -161,27 +162,89 @@ func (b *Binance) GetHistoricalTrades(symbol string, limit int, fromID int64) ([
...
 
+	var resp []AggregatedTrade
 	path := b.API.Endpoints.URL + aggregatedTrades + "?" + params.Encode()
 	return resp, b.SendHTTPRequest(path, limitDefault, &resp)
 }
 
+// batchAggregateTrades fetches trades in multiple requests
+// first phase, hourly requests until the first trade (or end time) is reached
+// second phase, limit requests from previous trade until end time is reached
+func batchAggregateTrades(b *Binance, arg *AggregatedTradeRequestParams, params url.Values) ([]AggregatedTrade, error) {
```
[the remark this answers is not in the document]
Rots: Sorry, my bad. That's why it didn't work.

exchanges/trade/README.md
```diff
@@ -60,7 +60,7 @@ _b in this context is an `IBotExchange` implemented struct_
 | Exchange | Recent Trades via REST | Live trade updates via Websocket | Trade history via REST |
 |----------|------|-----------|-----|
 | Alphapoint | No  | No        | No  |
-| Binance| Yes  | Yes        | No  |
+| Binance| Yes  | Yes        | Yes  |
```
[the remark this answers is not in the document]
Rots: That's what I was looking for, thanks.

exchanges/binance/binance.go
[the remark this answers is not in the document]
Rots: No, not a good one... the reason was that I refactored it from the `GetAggregatedTrades` and this is (kind of) what the editor/function extraction gave me for the function signature.

exchanges/binance/binance.go
Rots: Good catch, originally I was thinking keeping the limit to mean limit per request, but it makes sense to keep the semantics of "maximum number of trades to return" for the batch request

exchanges/binance/binance.go
Rots: Yes, sorry... fixed...

MadCozBadd (commented):
ye m88888888 looks good. but 1 more thing is can u add support for limit > 1000 to the batching aspect pls. rest works perfect

exchanges/binance/binance_types.go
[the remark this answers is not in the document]
thrasher-: Across the codebase we've made the exchange API as generic/native as possible (for people who don't care about the GCT built-ins) and have instead made the wrappers utilise the GCT built-ins and convert them to the generic types. That way, a person can choose between whether they want the GCT wrapped functionality or just the generic API implementation. I do see your point though in pushing the safe types all the way to the API level and it could be addressed in a separate PR as it would be a large undertaking.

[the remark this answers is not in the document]
Rots: > for people who don't care about the GCT built-ins
I still fail to understand the use case... if I just wanted a way to send requests to exchange API and handle the serialisation on my own, I would use curl or dynamically typed JavaScript or Python, not a golang library - this is exactly the added value that I expect from the library - to make using the exchange API easier. This includes not being forced to think about low level serialisation issues, but in core concepts.

exchanges/binance/binance_wrapper.go
```diff
@@ -534,8 +534,36 @@ func (b *Binance) GetRecentTrades(p currency.Pair, assetType asset.Item) ([]trad
 }
 
 // GetHistoricTrades returns historic trade data within the timeframe provided
-func (b *Binance) GetHistoricTrades(_ currency.Pair, _ asset.Item, _, _ time.Time) ([]trade.Data, error) {
-	return nil, common.ErrFunctionNotSupported
+func (b *Binance) GetHistoricTrades(p currency.Pair, a asset.Item, from, to time.Time) ([]trade.Data, error) {
+	req := AggregatedTradeRequestParams{
+		Symbol:    p,
```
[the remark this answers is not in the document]
Rots: This is exactly what I'm talking about with the type safety - as the caller of the `GetAggregatedTrades` I don't want to care about this.
In my opinion the "clean" solution would be separation of the core concept - currency pair - and it's representation... again, think of the `time.Time` analogy. It represents the concept of an instance of time, but how it is represented in different contexts (on the wire, configuration) is different. This approach would work for the currency pairs as well. The separator character is only relevant in the context of specific serialisation format, but not as part of the concept of a currency pair.

Rots:
I added the arbitrary limit support, but the more I think about the correct abstraction level, the worse I feel about this change... If this API is intended to be kept as close to original API as possible, then this change - batched requests and arbitrary limit - is directly violating this. The user cannot use the native API anymore.

exchanges/binance/binance_wrapper.go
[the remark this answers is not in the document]
thrasher-: The clean solution would be to move the formatting/serialisation of pairs and built-in types to the API funcs instead of the wrappers. At the moment, the wrappers act as the "unified API" for all exchanges and add value like you say. They primarily serialise the supplied currency pair ("BTC-USDT", "btc_usdt", "BTC~~~YOLO420~~~USDT"; whether from gctcli or an external app) to the correct format the exchange requires ("BTCUSDT"). This is what engine uses to perform actions for all exchanges and why it's done in them. However, I do agree that we can instead move the type safety out of the wrappers and into the actual exchange API funcs, but this is out of the scope for this PR and will require a large overhaul. Am happy to leave your change but I can see that you've added a commit to keep it consistent with everything else.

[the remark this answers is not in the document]
Rots: OK, let's keep it now...

MadCozBadd (approved):
Looks perfect to me! Thanks for all the comments in the code too! makes it super ez for a noob like me to understand lol

gloriousCode (approved):
Tested and everything is working. Thank you for another contribution Rots! Thank you for also engaging with us to talk about the design.
I have found one thing, but its out of scope, so its more a comment to highlight what I found

I don't mind whether the logic of `batchAggregateTrades` exists inside `GetHistoricTrades` or acts as a helper function in the base `binance.go`

engine/rpcserver.go
```diff
@@ -2792,13 +2792,16 @@ func (s *RPCServer) GetHistoricTrades(r *gctrpc.GetSavedTradesRequest, stream gc
 		Asset:        r.AssetType,
 		Pair:         r.Pair,
 	}
-	iterateStartTime := UTCStartTime
-	iterateEndTime := iterateStartTime.Add(time.Hour)
-	for iterateStartTime.Before(UTCEndTime) {
+
+	for iterateStartTime := UTCStartTime; iterateStartTime.Before(UTCEndTime); iterateStartTime = iterateStartTime.Add(time.Hour) {
+		iterateEndTime := iterateStartTime.Add(time.Hour)
 		trades, err = exch.GetHistoricTrades(cp, asset.Item(r.AssetType), iterateStartTime, iterateEndTime)
```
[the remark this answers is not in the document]
Rots: or
4. add a limit argument to `GetHistoricTrades`

Rots:
Is there still something pending that I should do?

exchanges/binance/binance.go
```diff
@@ -185,10 +186,12 @@ func (b *Binance) GetAggregatedTrades(arg *AggregatedTradeRequestParams) ([]Aggr
...
 	}
-	// If startTime and endTime are sent, time between startTime and endTime must be less than 1 hour.
-	if arg.Limit == 0 && !arg.StartTime.IsZero() && !arg.EndTime.IsZero() && arg.EndTime.Sub(arg.StartTime) > time.Hour {
+	// Fall back to batch requests
+	// if the requested limit is not supported by exchange or
+	// if startTime and endTime are set and time between startTime and endTime is more than 1 hour
+	if !limitSupported || (!arg.StartTime.IsZero() && !arg.EndTime.IsZero() && arg.EndTime.Sub(arg.StartTime) > time.Hour) {
 		// Split the request into multiple
-		return batchAggregateTrades(b, arg, params)
+		return b.batchAggregateTrades(arg, params)
```
shazbert, line 194: If I change my limit up to 1001 I get no error and no data, I know this is somewhat intentional but it breaks from what I am expecting. I know its a pain but we may need a check for unset startTime as well as an incorrect limit to return an error? 
```go
_, err := b.GetAggregatedTrades(&AggregatedTradeRequestParams{
		Symbol: currency.NewPair(currency.BTC, currency.USDT).String(),
		Limit:  1001,
	})
	if err != nil {
		t.Error("Binance GetAggregatedTrades() error", err)
	}
```
Maybe scoped here we can do a start time check and return the error: https://github.com/thrasher-corp/gocryptotrader/pull/588/files/a4341f513e6fceb0fb991c5e1194708c0328b2e8..b7214a1d514795754d543472e254e8fccae3c003#diff-8ff10965769d44e4c1fffd1a0d55ccf0bc16b1bdcce3b2cb7bea45fc4548a27fR174

Let me know what you think. Maybe I am misinterpreting things a little bit.

shazbert (changes requested):
There's only one issue that may be impactful, all other changes are looking really good. Thank you very much.

exchanges/binance/binance.go
MadCozBadd: start time is being set to 0 since arg.StartTime is empty n then u add 1 hr to that for endTime and no results coming back during that period

Rots: Added some more tests as well

exchanges/binance/binance.go
```diff
@@ -169,10 +169,17 @@ func (b *Binance) GetHistoricalTrades(symbol string, limit int, fromID int64) ([
 func (b *Binance) GetAggregatedTrades(arg *AggregatedTradeRequestParams) ([]AggregatedTrade, error) {
 	params := url.Values{}
 	params.Set("symbol", arg.Symbol)
-	limitSupported := true
+	// fromId xor start time must be set
+	canBatch := arg.FromID == 0 != arg.StartTime.IsZero()
```
MadCozBadd, line 173: image
had to take a moment to appreciate that

exchanges/binance/binance_test.go
```diff
@@ -462,6 +462,45 @@ func TestGetAggregatedTradesBatched(t *testing.T) {
...
+		{
+			name: "start time and fromId cannot be both set",
+			args: &AggregatedTradeRequestParams{
+				Symbol:    currency.NewPair(currency.BTC, currency.USDT).String(),
+				StartTime: start,
+				EndTime:   start.Add(75 * time.Minute),
+				FromID:    2,
+			},
+		},
+	}
```
shazbert, line 491: Please add this case:
```go 
		{
			name: "custom limit with start time set, no end time",
			args: &AggregatedTradeRequestParams{
				Symbol:    currency.NewPair(currency.BTC, currency.USDT).String(),
		                StartTime: time.Date(2020, 11, 18, 12, 0, 0, 0, time.UTC),
		                Limit:     1001,
			},
		},
```

shazbert (changes requested):
Just found an issue when no end time is set on a custom limit. 
Tested with end time set and get full payload as expected so that is perfect!

MadCozBadd (commented):
everything works perfectly m8888 except for the case which shazbert mentioned where it doesnt error and just returns nil. However SetValues() function in binance.go which sets validLimits includes 5000 as a valid limit too, so either a new function needs to be created to set limit values or need to add an edgecase or something so that 5000 value doesnt get thro

Rots:
Also, thanks for being so patient and thorough with the reviews.

Rots:
Oops, accidentally closed this :fearful: some keyboard combo.

exchanges/binance/binance_test.go
Rots: Fixed.

exchanges/binance/binance.go
```diff
@@ -161,27 +162,130 @@ func (b *Binance) GetHistoricalTrades(symbol string, limit int, fromID int64) ([
...
+// then the trades are collected with multiple backend requests.
+// https://binance-docs.github.io/apidocs/spot/en/#compressed-aggregate-trades-list
+func (b *Binance) GetAggregatedTrades(arg *AggregatedTradeRequestParams) ([]AggregatedTrade, error) {
+	params := url.Values{}
+	params.Set("symbol", arg.Symbol)
+	// fromId xor start time must be set
+	canBatch := arg.FromID == 0 != arg.StartTime.IsZero()
+	needBatch := false
+	if arg.Limit > 0 {
+		err := b.CheckLimit(arg.Limit)
```
MadCozBadd, line 176: Last last last last change from me promise lol. but if valid limits dont really apply to this function. so can we just remove the arg limit check n instead just add one for if > 1000. coz the endpoint doesnt restrict requests to 5, 10, 15, ..., 500, 1000. we can put in 69 or 111 or any random number below 1000 and theyll return the exact amt

MadCozBadd (commented):
last one promise

exchanges/binance/binance.go
Rots: aha, great catch! The CheckLimit was there before so I never questioned whether this was necessary. Removed.

MadCozBadd (approved):
looks good to me :D (hopefully to others too)

gloriousCode (approved):
tACK! Thanks for your contribution Rots!
> thanks everyone for being so patient and thorough with the reviews

Thanks for sticking around through all our comments 😄

thrasher- (approved):
tACK, thanks for addressing the feedback!

shazbert (approved):
tACK! Looking good!

PR #607, opened 2020-12-04
Poloniex: Implementation of GetOrderInfo method

vazha:
# PR Description

Poloniex GetOrderInfo now return trades list on the closed orders and status information on active orders.

Fixes # (issue)

Also in GetOrder method (engine/rpcservet.go) handled creationTime and updateTime results to prvent negative timestamps returning if no date received

## Type of change

## How has this been tested

- [x] go test ./... -race

## Checklist


vazha:
I see linter messages  "should rewrite switch statement to if statement", but I can't use if statement to check type of var. Should I use reflection to avoid this? But, in my opinion, this would be an unnecessary complication of the code.
What do you think about it?

exchanges/poloniex/poloniex_types.go
```diff
@@ -66,6 +66,27 @@ type TradeHistory struct {
 	Total         float64 `json:"total,string"`
 }
 
+// OrderStatus holds order status data
+type OrderStatus struct {
+	Result  map[string]interface{} `json:"result"`
```
shazbert, line 71: `Result  map[string]interface{} `json:"result"` can be changed to:
```go
Result  map[string]wsStatus `json:"result"`

// include new type
// wsStatus defines order status details
type wsStatus struct {
	Pair           string  `json:"currencyPair"`
	Rate           float64 `json:"rate,string"`
	Amount         float64 `json:"amount,string"`
	Total          float64 `json:"total,string"`
	StartingAmount float64 `json:"startingAmount,string"`
	Type           string  `json:"type"`
	Status         string  `json:"status"`
	Date           string  `json:"date"`
	Fee            float64 `json:"fee,string"`
}
```

This should cut down on type handling.

exchanges/poloniex/poloniex.go
```diff
@@ -92,9 +94,7 @@ func (p *Poloniex) GetOrderbook(currencyPair string, depth int) (OrderbookAll, e
 		if err != nil {
 			return oba, err
 		}
-		if resp.Error != "" {
-			return oba, fmt.Errorf("%s GetOrderbook() error: %s", p.Name, resp.Error)
-		}
```
shazbert, line 97: Please revert as this stops errors from coming back from endpoint.

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -589,8 +589,147 @@ func (p *Poloniex) CancelAllOrders(_ *order.Cancel) (order.CancelAllResponse, er
 
 // GetOrderInfo returns order information based on order ID
 func (p *Poloniex) GetOrderInfo(orderID string, pair currency.Pair, assetType asset.Item) (order.Detail, error) {
-	var orderDetail order.Detail
-	return orderDetail, common.ErrNotYetImplemented
+	var orderInfo order.Detail
+	trades, err := p.GetAuthenticatedOrderTrades(orderID)
```
shazbert, line 593: I think it's better to return an error.  We can always range over trade returns that could be nil. This reduces another potential outbound call from p.GetAuthenticatedOrderStatus.

engine/rpcserver.go
```diff
@@ -861,15 +871,15 @@ func (s *RPCServer) GetOrder(_ context.Context, r *gctrpc.GetOrderRequest) (*gct
 		AssetType:     result.AssetType.String(),
 		OrderSide:     result.Side.String(),
 		OrderType:     result.Type.String(),
-		CreationTime:  result.Date.Unix(),
+		CreationTime:  creationTime,
```
shazbert, line 874: Is there a specific reason you have done the checks above this? I would just revert back to result.Date.Unix() and closetime.

shazbert (changes requested):
Thank you Vazha! Came across some issues that need to be addressed before the merge.

gloriousCode (commented):
Thanks for another contribution Vazha! I can confirm that it works, however I agree with the points Shazbert has already raised
```
{
 "exchange": "Poloniex",
 "base_currency": "USDT",
 "quote_currency": "BTC",
 "order_side": "BUY",
 "creation_time": 1607299645,
 "status": "OPEN",
 "price": 0.5,
 "amount": 10,
 "fee": 0.00125,
 "cost": 5
}
```

exchanges/poloniex/poloniex_types.go
```diff
@@ -66,6 +66,27 @@ type TradeHistory struct {
...
+	Success int64                  `json:"success"`
+}
+
+// OrderTrade holds order trade data
+type OrderTrade struct {
```
MadCozBadd, line 76: https://docs.poloniex.com/#returntradehistory-private
startingAmount isnt a part of dat

MadCozBadd (commented):
could only find 1 small issue

exchanges/poloniex/poloniex_types.go
```diff
@@ -35,7 +35,7 @@ type OrderbookResponse struct {
 	Asks     [][]interface{} `json:"asks"`
 	Bids     [][]interface{} `json:"bids"`
 	IsFrozen string          `json:"isFrozen"`
-	Error    string          `json:"error"`
```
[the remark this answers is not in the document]
vazha: I deleted it because there is no such field in Poloniex doc: https://docs.poloniex.com/#returnorderbook
But since you see fit, I revert it back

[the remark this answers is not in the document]
shazbert: Thankyou, the reasoning being the JSON that is being returned when an error occurs is different to the actual struct they require. To limit doubling up on unmarshal and the use of an intermediary error struct we can include the field directly into the orderbookresponse. 

from https://poloniex.com/public?command=returnOrderBook&currencyPair=BTC_ETH&depth=10:
image

With incorrect pair https://poloniex.com/public?command=returnOrderBook&currencyPair=BADBADBAD&depth=10:
image

exchanges/poloniex/poloniex.go
```diff
@@ -418,6 +418,44 @@ func (p *Poloniex) GetAuthenticatedTradeHistory(start, end, limit int64) (Authen
...
+func (p *Poloniex) GetAuthenticatedOrderTrades(orderID string) (o []OrderTrade, err error) {
+	values := url.Values{}
+
+	if orderID == "" {
+		return o, fmt.Errorf("no orderId passed")
+	}
+
+	values.Set("orderNumber", orderID)
+	var result json.RawMessage
+	err = p.SendAuthenticatedHTTPRequest(http.MethodPost, poloniexOrderTrades, values, &result)
```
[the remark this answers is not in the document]
vazha: Here we can't use direct unmarshalling because in error case we should unmarshal struct, not slice.

exchanges/poloniex/poloniex_types.go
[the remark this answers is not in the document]
vazha: Agree, thanks :)

exchanges/poloniex/poloniex.go
```diff
@@ -418,6 +418,44 @@ func (p *Poloniex) GetAuthenticatedTradeHistory(start, end, limit int64) (Authen
...
+// GetAuthenticatedOrderStatus returns the status of a given orderId.
+func (p *Poloniex) GetAuthenticatedOrderStatus(orderID string) (o OrderStatus, err error) {
+	values := url.Values{}
+
+	if orderID == "" {
+		return o, fmt.Errorf("no orderId passed")
+	}
+
+	values.Set("orderNumber", orderID)
+	var result json.RawMessage
```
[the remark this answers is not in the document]
vazha: Nice!

exchanges/poloniex/poloniex.go
```diff
@@ -418,6 +418,44 @@ func (p *Poloniex) GetAuthenticatedTradeHistory(start, end, limit int64) (Authen
 	return mainResult, json.Unmarshal(result, &mainResult.Data)
 }
 
+// GetAuthenticatedOrderStatus returns the status of a given orderId.
+func (p *Poloniex) GetAuthenticatedOrderStatus(orderID string) (o OrderStatus, err error) {
```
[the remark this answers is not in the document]
vazha: done

exchanges/poloniex/poloniex.go
```diff
@@ -418,6 +418,44 @@ func (p *Poloniex) GetAuthenticatedTradeHistory(start, end, limit int64) (Authen
...
+	if err != nil {
+		return o, err
+	}
+
+	err = json.Unmarshal(result, &o)
+	return
+}
+
+// GetAuthenticatedOrderTrades returns all trades involving a given orderId.
+func (p *Poloniex) GetAuthenticatedOrderTrades(orderID string) (o []OrderTrade, err error) {
```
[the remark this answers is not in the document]
vazha: done

exchanges/poloniex/poloniex_wrapper.go
vazha: Sorry, I don't quite understand your comment. I am trying to get complete information about the order, including trades (if any). On the closed orders we will get info from the GetAuthenticatedOrderTrades method, and from the GetAuthenticatedOrderStatus for open orders. And I think on partially filled orders we will get info from both methods.

Concerning "Is there a specific reason you have done the checks above this? I would just revert back to result.Date.Unix() and closetime." - reason why I make the checks that if we get no Date from API responses, we will get negative Timestamp after .Unix() conversion

shazbert: maybe something like this:

```go
trades, err := p.GetAuthenticatedOrderTrades(orderID)
	if err != nil && err.Error() != "Order not found, or you are not the person who placed it." {
		return order.Detail{}, err
	}
```

So if there is an api key issue or malformed orderNumber we don't send another request. Let me know what you think.

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -589,8 +589,147 @@ func (p *Poloniex) CancelAllOrders(_ *order.Cancel) (order.CancelAllResponse, er
...
+	if err == nil {
+		orderInfo.Exchange = p.Name
+		orderInfo.Pair = pair
```
shazbert, line 596: Can we move the assignments outside of this scope?

exchanges/poloniex/poloniex_wrapper.go
shazbert: Thanks for clarifying on the Unix output, I will resolve.

exchanges/poloniex/poloniex.go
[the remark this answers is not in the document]
shazbert: okay, thanks.

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -589,8 +589,147 @@ func (p *Poloniex) CancelAllOrders(_ *order.Cancel) (order.CancelAllResponse, er
...
+			orderInfo.Trades = append(orderInfo.Trades, tradeHistory)
+		}
+	}
+
+	resp, err := p.GetAuthenticatedOrderStatus(orderID)
+	if err != nil {
+		return orderInfo, err
+	}
+
+	for _, v := range resp.Result {
```
shazbert, line 623: Just a suggestion and for readability: 
```go
// OrderStatus holds order status data
type OrderStatus struct {
	Result  json.RawMessage `json:"result"`
	Success int64                  `json:"success"`
}
```

Then unmarshal into correct types based on success.

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -589,8 +589,147 @@ func (p *Poloniex) CancelAllOrders(_ *order.Cancel) (order.CancelAllResponse, er
...
+			}
+			return orderInfo, fmt.Errorf(resultType)
+		case map[string]interface{}: // case of data message
+			orderStatus := v.(map[string]interface{})
+			orderInfo.Exchange = p.Name
+
+			if statusInterface, ok := orderStatus["status"]; ok {
+				switch varType := statusInterface.(type) {
+				case string:
+					orderInfo.Status, _ = order.StringToOrderStatus(varType)
```
shazbert, line 637: can we check all error returns

shazbert (changes requested):
Thanks for making those changes and clarifying certain areas. I have some more suggestions.

exchanges/poloniex/poloniex_wrapper.go
vazha: GetAuthenticatedOrderTrades returns "Order not found" on open orders, so we will lose info from GetAuthenticatedOrderStatus  and vice versa - GetAuthenticatedOrderStatus returns  "Order not found"  on closed orders

vazha: "Can we move the assignments outside of this scope?" - done

exchanges/poloniex/poloniex_wrapper.go
vazha: On error case StringToOrderStatus return an order.UnknownStatus, so in my opinion we no need to handle it

exchanges/poloniex/poloniex_wrapper.go
vazha: Nice, the code significantly reduced :)

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -589,8 +590,78 @@ func (p *Poloniex) CancelAllOrders(_ *order.Cancel) (order.CancelAllResponse, er
...
+			if err != nil {
+				return orderInfo, err
+			}
+
+			orderInfo.Date, err = time.Parse(common.SimpleTimeFormat, s.Date)
+			if err != nil {
+				return orderInfo, err
+			}
+		}
+	}
```
shazbert, line 662: Can we please move this into GetAuthenticatedOrderStatus? This would be easier for people who might use this package as a library.

go.sum
```diff
@@ -42,9 +42,17 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
```
shazbert, line 42: can you please run a `go mod tidy`

exchanges/poloniex/poloniex_wrapper.go
shazbert: That's a good point. Will resolve.

exchanges/poloniex/poloniex_types.go
shazbert: Poloniex request path: https://poloniex.com/tradingApi
Poloniex request header [Sign]: []
Poloniex request header [Content-Type]: [application/x-www-form-urlencoded]
Poloniex request header [Key]: []
Poloniex request type: POST
Poloniex request body: command=returnOrderTrades&nonce=1607558318602531900&orderNumber=18365912899
Poloniex raw response: [{"globalTradeID":326981405,"tradeID":660369,"currencyPair":"BTC_OMG","type":"sell","rate":"0.00173628","amount":"49.40355195","total":"0.08577839","fee":"0.00150000","date":"2018-01-14 23:34:50"}
**There doesn't seem to be a field startingAmount in the returned payload from this endpoint.**

shazbert (changes requested):

exchanges/poloniex/poloniex_types.go
vazha: Yes, thanks!

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -589,8 +589,59 @@ func (p *Poloniex) CancelAllOrders(_ *order.Cancel) (order.CancelAllResponse, er
...
 func (p *Poloniex) GetOrderInfo(orderID string, pair currency.Pair, assetType asset.Item) (order.Detail, error) {
-	var orderDetail order.Detail
-	return orderDetail, common.ErrNotYetImplemented
+	orderInfo := order.Detail{
+		Exchange: p.Name,
+		Pair:     pair,
+	}
+
+	trades, err := p.GetAuthenticatedOrderTrades(orderID)
+	if err == nil {
```
gloriousCode, line 598: While I understand processing only on err nil (since Poloniex has a weird API), I think it would be nicer if we at least log the error. eg
```
trades, err := p.GetAuthenticatedOrderTrades(orderID)
err != nil && err.Error() != "Order not found, or you are not the person who placed it." {
   log.Error(log.Exchange, err)
}
```
as the type of error received could be of interest to us (eg the API has changed, the request is denied)
The for loop will be skipped on error anyway as there will be no trades in the slice to iterate over. We don't need to return on the error in case there are no trade details to return, but logging it will be helpful

Speaking of slices! Could you also change the for loop to be like `for i := range trades { ... ` as it is better for memory usage

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -589,8 +589,59 @@ func (p *Poloniex) CancelAllOrders(_ *order.Cancel) (order.CancelAllResponse, er
...
+			tradeHistory.Amount = o.Amount
+			tradeHistory.Total = o.Total
+			tradeHistory.Fee = o.Fee
+			orderInfo.Trades = append(orderInfo.Trades, tradeHistory)
+		}
+	}
+
+	resp, err := p.GetAuthenticatedOrderStatus(orderID)
+	if err != nil {
+		if len(orderInfo.Trades) > 0 { // on closed orders return trades only
```
gloriousCode, line 621: Poloniex has a really strange API for this https://docs.poloniex.com/#returnorderstatus so unfortunately this logic cannot be changed.
Does the error contain the status of the order? Unfortunately I don't have anything that can get to this scenario. It would be helpful to set the status of the returned orderInfo so at least the user can be made aware of why they are not receiving their requested order information.

exchanges/poloniex/poloniex.go
```diff
@@ -418,6 +420,82 @@ func (p *Poloniex) GetAuthenticatedTradeHistory(start, end, limit int64) (Authen
...
+		if err != nil {
+			return
+		}
+		if resp.Error != "" {
+			err = fmt.Errorf(resp.Error)
+		}
+	case '[': // data received
+		err = json.Unmarshal(result, &o)
+	default:
+		return o, fmt.Errorf("server wrong response")
```
gloriousCode, line 493: Would you mind updating this to  `return nil, fmt.Errorf("received unexpected response")` ?

MadCozBadd (approved):
Dont have a poloniex account to test but everything else looks good! utACK

gloriousCode (changes requested):
Thanks for making those changes Vazha! I have a couple of questions/comments for you

exchanges/poloniex/poloniex_wrapper.go
vazha: Yes, of course it's much better :)

exchanges/poloniex/poloniex_wrapper.go
vazha: OK, we can set manually order status as CLOSED if GetAuthenticatedOrderStatus return  "Order not found" while we have trades for this order.

exchanges/poloniex/poloniex_types.go
```diff
@@ -66,6 +68,39 @@ type TradeHistory struct {
 	Total         float64 `json:"total,string"`
 }
 
+// OrderStatus holds order status data
+type OrderStatus struct {
+	Result  json.RawMessage `json:"result"`
+	Success int64           `json:"success"`
+}
+
+// wsStatus defines order status details
```
gloriousCode, line 77: Please update this to `OrderStatusData`

gloriousCode (commented):
Thanks for making those latest changes Vazha! I'm happy with your implementation, just one last comment

shazbert (approved):
tACK! Thanks for making those changes, errors returned adequately and trade data returned.

gloriousCode (approved):
tACK! Thanks for that last update Vazha 👍

MadCozBadd (approved):
ack new changes

thrasher- (approved):
utACK, thanks @vazha !

PR #530, opened 2020-07-30
Feature: Implement funding rates, futures and coin margin (exchange API coverage)

ghost:
# PR Description

Addition of some api helper functions

Fixes # (issue)

## Type of change

- [] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [] This change requires a documentation update

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Tests added

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally and on Travis with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #608, opened 2020-12-09
binance: use currency.Pair for request arguments

Rots:
* request calls use type safe arguments
* binance API handles the correct conversion now on a lower level
* consolidate formatting to a single place
* remove unused Pair.ID

This change adds type safety to binance requests that are using currency pairs as input.

## Type of change

- [x] Refactoring - no change in functionality, but low level exchange API usage needs to be adjusted

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/currency/pair.go
+++ b/currency/pair.go
@@ -100,8 +100,8 @@ func NewPairFromString(currencyPair string) (Pair, error) {
 // apply the same format
 func NewPairFromFormattedPairs(currencyPair string, pairs Pairs, pairFmt PairFormat) (Pair, error) {
 	for x := range pairs {
-		fPair := pairs[x].Format(pairFmt.Delimiter, pairFmt.Uppercase)
-		if strings.EqualFold(fPair.String(), currencyPair) {
+		fPair := pairFmt.Format(pairs[x])
+		if strings.EqualFold(fPair, currencyPair) {
 			return pairs[x], nil
 		}
 	}
--- a/currency/pair.go
+++ b/currency/pair.go
@@ -108,6 +108,11 @@ func NewPairFromFormattedPairs(currencyPair string, pairs Pairs, pairFmt PairFor
 	return NewPairFromString(currencyPair)
 }
 
+// Format formats the given pair as a string
+func (f *PairFormat) Format(pair Pair) string {
+	return pair.Format(f.Delimiter, f.Uppercase).String()
+}
+
 // MatchPairsWithNoDelimiter will move along a predictable index on the provided currencyPair
 // it will then split on that index and verify whether that currencypair exists in the
 // supplied pairs
--- a/exchanges/binance/binance.go
+++ b/exchanges/binance/binance.go
@@ -10,7 +10,6 @@ import (
 	"net/url"
 	"sort"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/thrasher-corp/gocryptotrader/common"
@@ -89,7 +88,11 @@ func (b *Binance) GetOrderBook(obd OrderBookDataRequestParams) (OrderBook, error
 	}
 
 	params := url.Values{}
-	params.Set("symbol", strings.ToUpper(obd.Symbol))
+	symbol, err := b.formatSymbol(obd.Symbol)
+	if err != nil {
+		return orderbook, err
+	}
+	params.Set("symbol", symbol)
 	params.Set("limit", fmt.Sprintf("%d", obd.Limit))
 
 	var resp OrderBookData
@@ -142,7 +145,11 @@ func (b *Binance) GetMostRecentTrades(rtr RecentTradeRequestParams) ([]RecentTra
 	var resp []RecentTrade
 
 	params := url.Values{}
-	params.Set("symbol", strings.ToUpper(rtr.Symbol))
+	symbol, err := b.formatSymbol(rtr.Symbol)
+	if err != nil {
+		return nil, err
+	}
+	params.Set("symbol", symbol)
 	params.Set("limit", fmt.Sprintf("%d", rtr.Limit))
 
 	path := fmt.Sprintf("%s%s?%s", b.API.Endpoints.URL, recentTrades, params.Encode())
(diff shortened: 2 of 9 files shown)


exchanges/binance/type_convert.go
```diff
@@ -344,3 +346,12 @@ func (a *wsListStatus) UnmarshalJSON(data []byte) error {
 	a.Data.TransactionTime = aux.Data.TransactionTime.Time()
 	return nil
 }
+
+// formatSymbol formats the given pair to a string suitable for exchange API requests
+func (b *Binance) formatSymbol(pair currency.Pair) (string, error) {
```
shazbert, line 351: This is just a comment: We have an open PR: #530 which will introduce USDT Margined Futures and Coin Margined Futures; coin margined futures requires a different request format compared to spot and USDT margined futures. This might need to be upgraded in the future when this gets merged and asset.Item type passed in with the currency pair from wrapper to these functions that require formatting. This is not exactly ideal for what you are doing now.

shazbert (commented):
Thanks for opening this PR @Rots! I am a fan of the string -> pair type but what you have done will add more complexity down the track and if we are going to upgrade across all exchanges to this standard. I think it might be better to call formatting before these non interface functions are called. Or maybe add a type to the currency package e.g. `type FormattedPair string` and force the usage of formatting on these then we can replicate across all the exchange packages account for different asset types? Let me know what you think.

MadCozBadd (approved):
No errors that I can spot!

currency/pair.go
```diff
@@ -108,6 +108,11 @@ func NewPairFromFormattedPairs(currencyPair string, pairs Pairs, pairFmt PairFor
 	return NewPairFromString(currencyPair)
 }
 
+// Format formats the given pair as a string
+func (f *PairFormat) Format(pair Pair) string {
+	return pair.Format(f.Delimiter, f.Uppercase).String()
```
thrasher-, line 113: Please add test coverage for this

exchanges/binance/type_convert.go
thrasher-: For now since we only support spot and margin assets which use the same request format, this is fine as is. The support for different asset request formatting can implemented in #530

thrasher- (changes requested):
Only 1 minor nit, the rest looks good. Thanks @Rots !

exchanges/binance/type_convert.go
Rots: With each (non-interface) API call you implicitly have the context whether the call is for spot or margin or other asset types - thus I'm not worried about losing information between the wrapper/API boundary.
That's also why I kept it non-exported - it should only be used in the context that is relevant - currently for Spot (and Margin) only.
I added a comment for now.
If you wish, I'd be happy to generalise it to:
`func (b *Binance) formatSymbol(pair currency.Pair, a asset.Type) (string, error)`
to make it useful for future calls?

currency/pair.go
Rots: Done.

Rots:
I'm happy to explain the reasoning...
> will add more complexity down the track

I don't agree. The complexity is already there and needs to be there (because of the wrapper interface) and will be there. With this change I'm just shifting it from a higher level (wrapper) to lower level - to the exchange API calls, ensuring safety for callers of the library (see changes in `binance_websocket.go` and `binance_wrapper.go`, if anything, I would say this change simplifies usage of the API calls). The serialisation (also 'formatting' in this context) has to be done anyway, the best place to do it is only when it's needed to ensure that the correct format is used. Relying on callers to do something that is mandatory, is not a good approach for encapsulation and for the correct level of abstraction.

Also see the discussion earlier [here](https://github.com/thrasher-corp/gocryptotrader/pull/588#discussion_r520851830)

> I think it might be better to call formatting before these non interface functions are called.

Let me give you a use case...
As a user of the go library I want to use the Binance API (non-interface functions) directly to interact with exchange as part of my software. I don't need the wrapper, as information (because it's generalised) coming from wrapper is 'repackaged' and some details are lost. For me it would be safer if I didn't have to care - in addition to providing the currency pairs - about low level serialisation and formatting.
If I forget to call the `ex.FormatExchangeCurrency` and/or accidentally mix the pair coming from another exchange in a different format, that can cause issues.

If the non-interface functions are not meant to be public, then they should be unexported and then I wouldn't care that much at which stage the formatting is done (I would still have the level of abstraction argument though)

As a side note, the current formatting was already inconsistently implemented in the non-interface functions, example the symbol is not always converted to uppercase see https://github.com/thrasher-corp/gocryptotrader/pull/608/files#diff-8ff10965769d44e4c1fffd1a0d55ccf0bc16b1bdcce3b2cb7bea45fc4548a27fL145 and https://github.com/thrasher-corp/gocryptotrader/pull/608/files#diff-8ff10965769d44e4c1fffd1a0d55ccf0bc16b1bdcce3b2cb7bea45fc4548a27fL171
This change consolidates and unifies approach.

> add a type to the currency package e.g. `type FormattedPair string`

That wouldn't help - each API call for different exchange would have to use the correct format for that exchange. We can't be sure that the "user" (i.e. caller of the exchange API) didn't pass in a formatted string that is valid for one exchange to the API call of another exchange - we would be in exactly the same position as now with the `currency.Pair` that carries the format along.
To make this approach work, we'd have to have a type per exchange. e.g. `binance.FormattedPair`, `kraken.FormattedPair`etc. and again the callers would have make sure to always construct and provide the correct FormattedPair - it would add a bit of clarity compared to the current `string` but would still mandate the callers to construct these 



My vision for a completely nice solution would be to strip down the current `currency.Pair` to a bare minimum - not carrying the formatting information, only the pair of currencies. That would also imply changes to `currency.Code` or to create a new simple parallel type for that.
For parsing and formatting there would be other means, like standard library provides for `time.Time`.

gloriousCode (approved):
tACK! I appreciate the direction this is heading towards. Your arguments around calling at the lowest level and not relying on earlier functions are good and this ensures that someone can't really screw up a request by putting in "btc_usdt" when Binance wants a different format. I do have one comment, but its not mandatory and consider it only if there are other comments from the others

exchanges/binance/type_convert.go
shazbert: I have no issue with keeping it as it is. We can do that later. Thanks for the continued clarification.

shazbert (approved):
Tested Approved. Thanks for taking the time to make the changes and clarifying everything. Much appreciated.

thrasher- (approved):
Thanks for adding in the test and this PR!

exchanges/binance/binance.go
```diff
@@ -491,15 +530,23 @@ func (b *Binance) CancelExistingOrder(symbol string, orderID int64, origClientOr
...
+		var err error
+		symbol, err = b.formatSymbol(*pair)
+		if err != nil {
+			return resp, err
+		}
+	}
 	if symbol != "" {
-		params.Set("symbol", strings.ToUpper(symbol))
+		params.Set("symbol", symbol)
 	}
```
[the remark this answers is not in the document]
gloriousCode: Nah, that's all good 👍 Thanks for the clarification!

MadCozBadd (approved):
tACK new changes

PR #610, opened 2020-12-21
Engine/websocket: subscribe to default channels only when actually needed

Rots:
* subscribe if this is required by ws routines or sync manager
* restore previous subscriptions on reconnect (also don't restore anything if there were no subscriptions)

## Type of change
Mostly refactoring

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

start engine with WS disabled
enable WS via gctcli

## Checklist

### DIFF
--- a/engine/syncer.go
+++ b/engine/syncer.go
@@ -529,9 +529,12 @@ func (e *ExchangeCurrencyPairSyncer) Start() {
 			}
 
 			if !ws.IsConnected() && !ws.IsConnecting() {
-				go WebsocketDataReceiver(ws)
+				go Bot.WebsocketDataReceiver(ws)
 
 				err = ws.Connect()
+				if err == nil {
+					err = ws.FlushChannels()
+				}
 				if err != nil {
 					log.Errorf(log.SyncMgr,
 						"%s websocket failed to connect. Err: %s\n",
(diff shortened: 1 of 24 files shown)


shazbert (approved):
tACK, no issues from me.

engine/syncer.go
```diff
@@ -529,9 +529,12 @@ func (e *ExchangeCurrencyPairSyncer) Start() {
 			}
 
 			if !ws.IsConnected() && !ws.IsConnecting() {
-				go WebsocketDataReceiver(ws)
+				go Bot.WebsocketDataReceiver(ws)
 
 				err = ws.Connect()
+				if err == nil {
+					err = ws.FlushChannels()
```
gloriousCode, line 536: Given we have an `else` after `if err != nil {` would you be able to combine the two? Either put both in `if err == nil` or in the `else`

gloriousCode (changes requested):
This is a nice update! Thanks Rots! Just one mini comment

engine/syncer.go
Rots: not quite... I'd still need to add another `err!=nil` check for after the FlushChannels and then duplicate the lines 538-543, I prefer to keep it shorter like this

gloriousCode: Ahh I missed that. Yeah fair enough.

gloriousCode (approved):
tACK! Thanks for clarifying your reasoning and thank you for another update rots!

MadCozBadd (approved):
tACK

thrasher- (approved):

PR #616, opened 2021-01-07
Config: Race fix (draft)

shazbert:
# PR Description

This includes a hotfix for a potential race condition when the engine starts and stops a global service and multiple services are interacting with the configuration system. 

Moves config fields to a subtype to be protected by a mutex and not accessible out of the package. Not idiomatic but added getter and setter methods ad hoc. 

Open as a draft to gauge feedback and to get a better direction. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -58,7 +58,8 @@ func New() (*Engine, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to load config. Err: %s", err)
 	}
-	b.GctScriptManager, err = gctscript.NewManager(&b.Config.GCTScript)
+	confScript := b.Config.GetGCTScript()
+	b.GctScriptManager, err = gctscript.NewManager(&confScript)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create script manager. Err: %s", err)
 	}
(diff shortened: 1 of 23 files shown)


engine/events.go
```diff
@@ -308,10 +308,10 @@ func EventManger() {
 
 // IsValidExchange validates the exchange
 func IsValidExchange(exchangeName string) bool {
-	cfg := config.GetConfig()
-	for x := range cfg.Exchanges {
-		if strings.EqualFold(cfg.Exchanges[x].Name, exchangeName) && cfg.Exchanges[x].Enabled {
-			return true
+	exchanges := config.GetConfig().GetAllExchangeConfigs()
```
Rots, line 311: a possibly better approach (to take care of the race problem) is to eliminate all the uses of `config.GetConfig()` and always get it from the *Engine object instead

engine/engine.go
```diff
@@ -52,13 +52,14 @@ var (
 // New starts a new engine
 func New() (*Engine, error) {
 	var b Engine
-	b.Config = &config.Cfg
+	b.Config = config.GetConfig()
```
Rots, line 55: I think the purpose here is to always have a clean config to start with... the GetConfig() is risky as the returned config may already be used. So I think the previous was correct.
If you want to make the construction explicit, you can add a `config.New()` function and use that instead.

Rots (commented):
Were you able to reproduce the race condition with a test?
I'm not convinced this improves the code...

thrasher-:
Closed due to inactivity and fixes implemented in the backtester PR

PR #599, opened 2020-11-25
exchange: binance orderbook fix 

shazbert:
# PR Description

Ported from #570 draft, that PR is getting big and complex. This should fix an issue where we do not apply the correct buffered updates correctly on startup. 

Main differences to approach:
1) Shifted the acquisition of books to different worker routines util. buffered job channel. This is for processing a wide array of enabled currencies.
2) Only acquire books that actually receive an update from the websocket, this allows for higher volume books to be fetched first.
3) When books are in fetching mode we will buffer *all* updates and apply them ad hoc when we receive the payload. 

**Designed this way for future competing asset types incl. margin** 

Found some happy little accidents when I found that there were indeed some artifacts in the order books across multiple exchanges these continued updates are trying to reduce the chance of this occurring and adding in little optimisations:

- [x] Updated verbosity in printOrderbookSummary
- [x] Added websocket orderbook buffer enabled to exchanges that are missing. **Default buffer to off as this impedes orderbook synchronisation.** Linked up buffer to configuration. 
- [x] Updated orderbook.go processing to not repair but return an error of any issues found, expanding validation resulted in multiple issues across REST and Websocket handling.
- [x] Added in Not Aggregated field to orderbook.Base to differentiate between exchanges that return the full orderbook with multiple same price levels but different amounts. 
- [x] Added exported sorting functions for orderbook
- [x] Updated buffer handling hash tables
- [x] Added reversing of orderbook Items which is faster than sorting
- [x] Removed reliance of using sort algorithm, added in code that inserts when the correct price level has been obtained thus reducing the iteration across the entire book.
- [x] Bitfinex fix for funding rate book.
- [x] Nil pointer panic fix for orderbook summary.

~~**BUFFER TESTS ARE BROKEN** I will fix this soon.~~ Done

**Need to test REST and Websocket on all exchanges to see if there are any more updates I need to do.**

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [x] I ran it
[...]

PR #619, opened 2021-01-11
Binance websocket orderbook sync fix

vazha:
# PR Description

As I wrote in Slack - PR #599  does not fixed issues with Binance websocket orderbook syncronisattion.
I came to the conclusion that the reason is that the subscription to the order book by websockets occurs after the getting snapshot of the orderbook received by the REST API. I mean The LastUpdateID (of smapshot) is less than the FirstUpdateID of the websocket updates. Thus, some of the updates are lost, which leads to freezing of prices in the order book.
I have moved part of code to different goroutine, to let websocket subscripton starts earlyer. I have tested this solution many times and it works fine.

Fixes # (issue)

## Type of change

## Checklist


gloriousCode (approved):
Thanks for this PR @vazha, This is an improvement given that we could only subscribe to websocket orderbook updates once all REST requests were done, which could be awhile one has many currencies enabled. I went back and tested and was able to get out of sync orderbooks and I have not since using your fix.

exchanges/binance/binance_websocket.go
```diff
@@ -84,21 +84,28 @@ func (b *Binance) WsConnect() error {
...
-		return err
+		log.Errorf(log.ExchangeSys, "orderBookProcess, GetEnabledPairs error: %s", err)
+		return
 	}
 
 	for i := range enabledPairs {
 		err = b.SeedLocalCache(enabledPairs[i])
 		if err != nil {
-			return err
+			log.Errorf(log.ExchangeSys, "orderBookProcess, SeedLocalCache error: %s", err)
```
thrasher-, line 102:
```suggestion
		log.Errorf(log.ExchangeSys, "%s orderBookProcess, SeedLocalCache error: %s", b.Name, err)
```

thrasher- (changes requested):
Thanks for this PR! Only have a basic nit due to the ambiguity of having no exchange name associated with the error message when multiple exchanges are enabled

gloriousCode (approved):

MadCozBadd (approved):
tACK!!!!!

exchanges/binance/binance_websocket.go
```diff
@@ -84,21 +84,28 @@ func (b *Binance) WsConnect() error {
...
+	go b.orderBookProcess()
+	return nil
+}
+
+// orderBookProcess prepare orderbook handling
+func (b *Binance) orderBookProcess() {
 	enabledPairs, err := b.GetEnabledPairs(asset.Spot)
 	if err != nil {
-		return err
+		log.Errorf(log.ExchangeSys, "orderBookProcess, GetEnabledPairs error: %s", err)
```
thrasher-, line 95: Just missing one more b.Name here and then this is good to merge!

thrasher- (changes requested):

MadCozBadd (approved):

gloriousCode (approved):

thrasher- (approved):
Thanks @vazha !

PR #630, opened 2021-02-11
Binance websocket orderbook reconnection fix, some bug fix

vazha:
# PR Description

I noticed that the Binance websocket is still very unstable
I found several errors that affected the work of the Binance websocket order book, this led to problems with synchronization after the websocket reconnection.
There was also a problem with the program freezing at startup, if the exchange refused to connect (for banning reason for example). This happened because an Background context was passed to the SendPayload function instead of WithTimeout context.

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race

## Checklist


exchanges/binance/binance_websocket.go
```diff
@@ -400,8 +382,11 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 							err)
 					}
 
-					err = b.UpdateLocalBuffer(&depth)
+					init, err := b.UpdateLocalBuffer(&depth)
 					if err != nil {
+						if init {
```
gloriousCode, line 387: There is no scenario where `init` can return `true` and `err != nil`. 
Perhaps you meant to store the error from `err = b.obm.stageWsUpdate(wsdp, currencyPair, asset.Spot)` and use that? Because reading the code for `b.obm.checkIsInitialSync(currencyPair)`, the only scenario where `init` returns `true`, `err` is `nil`

gloriousCode (changes requested):
Thanks for continuously testing and helping us solve Binance orderbook issues @vazha! It's really appreciated. I'll do a more thorough round of testing soon, but I just wanted to highlight the first two things I came across

exchanges/request/request.go
```diff
@@ -174,7 +174,8 @@ func (r *Requester) doRequest(req *http.Request, p *Item) error {
 				delay = after
 			}
 
-			if d, ok := req.Context().Deadline(); ok && d.After(time.Now().Add(delay)) {
+			// what to do if req.Context().Deadline() is 0 ? Because it never passed here...
```
shazbert, line 177: We have a default retry policy that should stop this from continuously retrying. Are you having issues here? There are 10 exchanges that utilise context.WithDeadline() functionality.

shazbert (commented):
Thanks @vazha, changes seem good, reconnection works, just need some clarification on the context.Deadline issue.

exchanges/request/request.go
vazha: Yes, on Binance I ran into a problem when a lot of requests were sent to the server, the server banned me and returned a funny message "I'm a teapot". And, due to the fact that in this place we do not receive context.WithDeadline () - the gocryptortader was frozen for 20 minutes, and it was not possible to stop it using CTRL+C, only via KILL -9.  I have not yet figured out the issue, why the retry policy did not work. But on Binance this happened often, during orderbooks synchronization for websockets (via rest) on a large number of pairs

gloriousCode: Are you by any chance running multiple instances of GoCryptoTrader? Our rate limits should prevent the teapot error from ever occurring along with the banning from Binance's side. If the rate limits are inaccurate, we could re-assess them and potentially add a context.Deadline for Binance in another PR.

gloriousCode (commented):
I have not been able to replicate the orderbook checksum issues that you have been receiving. I can verify that the orderbook is accurate on this branch before & after re-connection and that the added protections would be helpful for addressing issues such as false errors with initial sync, lastUpdateID and processing when asks and bids are nil.

So while I do think this is a good PR, would you be able to provide any additional details for me to try and experience the issues?

vazha:
> I have not been able to replicate the orderbook checksum issues that you have been receiving. I can verify that the orderbook is accurate on this branch before & after re-connection and that the added protections would be helpful for addressing issues such as false errors with initial sync, lastUpdateID and processing when asks and bids are nil.
> 
> So while I do think this is a good PR, would you be able to provide any additional details for me to try and experience the issues?

I didn't mean the checksum problems, but the initial synchronization process itself. Perhaps even this is not a big problem, but implementation costs. In order to hide these errors, I use the additional return parameter init

exchanges/request/request.go
vazha: No, I do not run a multiple instances, And I still don't understand why the requests limiter did not work.

gloriousCode:
My apologies! I approached reviewing this incorrectly thinking it was a checksum issue. I also mistakenly pushed a `git merge master` to your branch sorry, feel free to revert it, but has no impact on your changes. I'll retest

vazha:
> My apologies! I approached reviewing this incorrectly thinking it was a checksum issue. I also mistakenly pushed a `git merge master` to your branch sorry, feel free to revert it, but has no impact on your changes. I'll retest

@gloriousCode no problem, I just waiting a rewue and commiting if it's OK

> My apologies! I approached reviewing this incorrectly thinking it was a checksum issue. I also mistakenly pushed a `git merge master` to your branch sorry 😅 , feel free to revert it, but has no impact on your changes. I'll retest

No problem, I'm just waiting for reviewing and approving of my PR, if everything is OK, to continue with the Kraken websocket. I just don't like doing tasks at the same time)

gloriousCode (approved):
tACK! Thanks for your PR vahza!

MadCozBadd (approved):
tACK, thanks for the contribution vazha!!!!!!!!!

exchanges/request/request.go
shazbert: Please remove comment

vazha: done

exchanges/binance/binance_websocket.go
```diff
@@ -472,51 +457,40 @@ func (b *Binance) SeedLocalCacheWithBook(p currency.Pair, orderbookNew *OrderBoo
...
-		if cleanupErr != nil {
-			log.Errorf(log.WebsocketMgr,
-				"%s cleanup websocket orderbook error: %v",
-				b.Name,
-				cleanupErr)
-		}
-
-		return err
+		b.flushAndCleanup(currencyPair)
+		return false, err
```
shazbert, line 490: remove or change LN 493 to `return false, nil`

exchanges/binance/binance_websocket.go
```diff
@@ -743,6 +722,8 @@ func (o *orderbookManager) stageWsUpdate(u *WebsocketDepthStream, pair currency.
 	case state.buffer <- u:
 		return nil
 	default:
+		<-state.buffer    // pop one element
+		state.buffer <- u // to shift buffer on fail
```
shazbert, line 726: I am not entirely sure about this change, as it means the initial book has not been fetched yet. It might be better to extend the buffer for reasons being; if you popped off an element that does fit within the update of the fetched order book snapshot, maybe there is latency on the return leg of your request exceeding 1 second (very unlikely, so I see why you have done this.) There is a chance that it could skew liquidity. So if this is occuring quite regarly it might mean a different approach.

shazbert (changes requested):
Thank you for this PR! I have one little nit and something open for discussion, I am sorry for the slow down due to it being such a small possibility. @thrasher- @gloriousCode for extra input.

exchanges/binance/binance_websocket.go
vazha: Yes of course.

exchanges/binance/binance_websocket.go
vazha: Thank you, you understand very precisely why I applied this solution. The reason why I decided to cut the buffer is - sometimes (perhaps it depended on the number of pairs used) the buffer overflowed and in this case the program went into an infinite loop when new updates could not be written to the buffer. But that was before I discovered the bugs that were fixed in this update, possible it may be not necessary now. So in theory, the exchange server can return a snapshot of the order book with a delay, because it's a REST and it's a internet , then a buffer overflow will occur. And will it cure itself without buffer pop or not - need additional tests.
Regarding the loss of liquidity, can this happen if recent.LastUpdateID of snapshot is between updt.FirstUpdateID and updt.LastUpdateID of update, as we chеck in func (u *update) validate ...  ?

shazbert: Thanks for your reply!

Maybe it's outside probability like you suggest. I am more thinking of the off chance we get an update it instantly sends a REST request within that 100ms window but the return leg exceeds 10 seconds or 100 additional updates for that currency pair. So in that buffer, we could reach 100 updates applicable to the receiving order book snapshot and then another one comes after and pops the very initial update out. 

So here's my suggestion which may be better for both of us. Keep what you have done because if we have competing currency updates (What you have done should still occur), but extend the channel buffer to the maximum default REST handshake (exchange.DefaultHTTPTimeout). Say change **maxWSUpdateBuffer** from 100 to 15 seconds * 10 (Potential updates per second)  = 150 updates and might ensure that REST will disengage and not apply the snapshot and hopefully retry? If that makes sense. 

TLDR;
**maxWSUpdateBuffer = 150**

vazha: OK, I think this is the right decision. In any case, if the problem appears again, we will think further :)

shazbert (approved):
Thank you! tACK!

MadCozBadd (approved):
Tested and Approved! Thanks for your ongoing contributions @vazha

gloriousCode (approved):
uTACK latest changes

thrasher- (approved):
Thanks @vazha + @shazbert

PR #633, opened 2021-02-21
fix buffer bug

woshidama323:
# PR Description

As local buffer could not be updated successfully, the gocryptrader would print errors like :

`[ERROR] | WEBSOCKET | 16/02/2021 18:29:27 | routines.go exchange Binance websocket error - Binance - UpdateLocalCache error: websocket orderbook synchronisation failure for pair BTC-USDT and asset spot
[ERROR] | WEBSOCKET | 16/02/2021 18:29:27 | routines.go exchange Binance websocket error - Binance - UpdateLocalCache error: websocket orderbook synchronisation failure for pair BTC-USDT and asset spot`

and orderbook could not be updated like binanace (i only test it on binance exchange maybe need more clarification)

Fixes # (issue)
The root cause i think is that a bug in buffer.go, you can find  it in the review part

## Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)

## How has this been tested

When running the gocryptotrader , you will find such errors in the log file or console.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ x ] I have performed a self-review of my own code

### DIFF
--- a/exchanges/stream/buffer/buffer.go
+++ b/exchanges/stream/buffer/buffer.go
@@ -109,7 +109,7 @@ func (w *Orderbook) Update(u *Update) error {
 // defined by w.obBufferLimit it well then sort and apply updates.
 func (w *Orderbook) processBufferUpdate(o *orderbookHolder, u *Update) (bool, error) {
 	*o.buffer = append(*o.buffer, *u)
-	if len(*o.buffer) < w.obBufferLimit {
+	if len(*o.buffer) > w.obBufferLimit {
 		return false, nil
 	}
 


gloriousCode (commented):
After doing some investigation, the reason this is coming up is because your config has a setting which should not be enabled for Binance. 
If your config is not encrypted, please update the following area under the `binance` exchange: 
```
 "orderbook": {
    "verificationBypass": true,
    "websocketBufferLimit": 5,
    "websocketBufferEnabled": false
   }
```
The key perpetrator is `websocketBufferEnabled` and Binance does not need the buffer. I can replicate your issue when `websocketBufferEnabled` is enabled.

The reason the buffer cannot be enabled for Binance is because we have code which attempts to synchronise the orderbook based on IDs. When we have the buffer enabled, it will group many updates and attempt to apply them, which it cannot because the ids are incorrect now. We should probably force it to always be disabled, but that's another issue

The change you are requesting will break the buffer for other exchanges if enabled there.

Are you able to set/check that `websocketBufferEnabled` for Binance and test again?

woshidama323:
Yes when revert to origin code and set 
`"orderbook": {
    "verificationBypass": true,
    "websocketBufferLimit": 5,
    "websocketBufferEnabled": false
   }
 `
it did work I will close this PR ?

PR #643, opened 2021-03-07
orderbook: Implement initial linked list

shazbert:
# PR Description

Add in linked-list for orderbook depth. Improved updating throughput and load on CPU. 

image

Changes:
- [x] Have single order book linked-list depth pointer shared to selective subsystems and potential future strategy systems.
	- Streaming buffer system sets up a local map for exchange which includes a pointer to the orderbook.Depth. The prior system would setup orderbook.Base copy,  would then update this local copy, lockdown the entire order book map, then flush orderbook.Base with new contents. The new system acts directly on depth. 
- [x] Add alert system for when the depth gets updated, hypothetically routine(s) can call wait() and be instantly alerted of a change.
- [ ] Shared book invalidation - Because depths are shared between REST and Streaming endpoints, a REST update will invalidate a streaming update and will require a resubscribe (This has not been implemented yet as it will have to be done exchange by exchange).
- [x] Added sync time.Duration for a streaming endpoint default being time.Minute.
- [x] Node slice/stack implementation:
	- A shared stack between the bid and ask linked lists to leverage hot memory speed (FIFO)
	- A service routine on the stack to clean up nodes that haven't been used within 30 seconds every minute

Current master:
image
Usually spikes up to ~50% CPU usage with all exchanges running but not all pairs enabled. 

This Branch:
image
Usually spikes up to ~20% CPU usage with all exchanges running but not all pairs enabled. 

Noteworthy:
- Reduced the interaction with the data handler to alert the sync manager (and subsequent display to stdout) via a streaming update as this caused significant overhead.
- The dispatch system seems to not be working as efficiently as will be needed in the future and a thorough benchmark and update need to be done. 
- GCTCLI streaming needs to be improved as this adds a lot of load on the system as well. Maybe adjust updating to 100ms instead of direct memory change. 

**All exchanges rest and websocket need to be tested again. (Sorry in advance)**

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


gloriousCode (changes requested):
Cool and exciting stuff 😎 I still have lots of reviewing to go at a later point, but here's a few things I came across initially 🦇

MadCozBadd (commented):
Some really lame nits while I was trying to get the hang of things

exchanges/coinbasepro/coinbasepro_websocket.go
```diff
@@ -331,7 +331,7 @@ func (c *CoinbasePro) ProcessUpdate(update WebsocketL2Update) error {
 			return err
 		}
 
-		if update.Changes[i][0].(string) == order.Buy.Lower() {
+		if update.Changes[i][0].(string) == "buy" {
```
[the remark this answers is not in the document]
shazbert: Yeah good point, maybe in a future update instead of calling strings.Lower() we can have the lower case already in memory for each order side.

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,675 @@
...
+			}
+			prev = *tip
+		}
+	}
+	return nil
+}
+
+// asks imbed a linked list to attach methods for ask depth specific
+// functionality
+type asks struct {
```
gloriousCode, line 423: Why is this like this instead of on the `Depth` struct ie
```
type Depth struct {
	asks linkedList
        bids linkedList
        ...
}
```

It looks as though you're trying to prevent a `bids` from ever calling an `asks` implementation while also re-using function names between them. I'm not saying to revert this, but I do find it strange.

There is a huge amount of reused code between the asks and bids functions and it would be good to extract that out into singular functions either way

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,675 @@
...
+				break // Continue updates
+			}
+
+			if (*tip).value.Price == updts[x].Price { // Price already found
+				return fmt.Errorf("%w for price %f",
+					errCollisionDetected,
+					updts[x].Price)
+			}
+
+			// Correct position/allignment found for price level
```
gloriousCode, line 617: alignment

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,675 @@
...
+				ll.length++
+				if (*tip).prev == nil {
+					// Place new node in front of current node
+					(*tip).prev = n
+					n.next = *tip
+					// Replace head entry
+					*tip = n
+				} else {
+					old := *tip
+					prev.next = n
```
gloriousCode, line 631: prev hasn't been checked here and could cause a panic

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,675 @@
...
+					bookmark.prev = tip
+					bookmark.next = nil
+					continue updates
+				}
+			}
+		}
+		n := stack.Pop()
+		n.value = updts[x]
+		switch {
+		case bookmark.prev == nil:
```
gloriousCode, line 575: bookmark needs a nil check

exchanges/orderbook/depth.go
```diff
@@ -0,0 +1,325 @@
...
+func (d *Depth) InsertBidAskByID(bidUpdts, askUpdts Items) {
+	if len(bidUpdts) == 0 && len(askUpdts) == 0 {
+		return
+	}
+	d.Lock()
+	if len(bidUpdts) != 0 {
+		d.bids.insertUpdates(bidUpdts, d.stack)
+	}
+	if len(askUpdts) != 0 {
+		d.asks.insertUpdates(askUpdts, d.stack)
```
gloriousCode, line 196: unhandled errors

exchanges/orderbook/node.go
```diff
@@ -0,0 +1,110 @@
...
+)
+
+// node defines a linked list node for an orderbook item
+type node struct {
+	value Item
+	next  *node
+	prev  *node
+	// Denotes time pushed to stack, this will influence cleanup routine when
+	// there is a pause or minimal actions during period
+	shelfed time.Time
```
gloriousCode, line 20: shelved?

exchanges/orderbook/node.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// stack defines a FIFO list of reusable nodes
+type stack struct {
+	nodes []*node
+	s     uint32
+	count int32
+}
+
+// newstack returns a ptr to a new stack instance, also starts the cleaning
```
gloriousCode, line 30: newStack

exchanges/orderbook/node.go
```diff
@@ -0,0 +1,110 @@
...
+// serbvice
```
gloriousCode, line 31: service

exchanges/orderbook/orderbook.go
```diff
@@ -198,58 +230,68 @@ func (b *Base) Verify() error {
...
-			}
+// checker defines specific functionality to determine ascending/descending
+// validation
+type checker func(current Item, previous Item) error
 
-			if b.Bids[i].ID != 0 && b.Bids[i].ID == b.Bids[i-1].ID {
-				return fmt.Errorf(bidLoadBookFailure, b.ExchangeName, b.Pair, b.AssetType, errIDDuplication)
-			}
-		}
+// asc specfically defines ascending price check
```
gloriousCode, line 254: specifically

exchanges/orderbook/orderbook.go
```diff
@@ -198,58 +230,68 @@ func (b *Base) Verify() error {
...
+var asc = func(current Item, previous Item) error {
+	if current.Price < previous.Price {
+		return errPriceOutOfOrder
+	}
+	return nil
+}
+
+// dsc specfically defines descending price check
```
gloriousCode, line 262: specifically

exchanges/orderbook/orderbook.go
```diff
@@ -259,29 +301,35 @@ func (b *Base) Verify() error {
...
 
-	if !b.VerificationBypass && !b.HasChecksumValidation {
+	if b.CanVerify() {
 		err := b.Verify()
 		if err != nil {
 			return err
 		}
 	}
-	return service.Update(b)
+	service.Update(b)
```
gloriousCode, line 326: `return service.Update(b)`

exchanges/orderbook/depth_test.go
```diff
@@ -0,0 +1,331 @@
...
+	// If we add anymore options to the options struct later this will complain
+	// generally want to return a full carbon copy
+	mirrored := reflect.Indirect(reflect.ValueOf(d.options))
+	for n := 0; n < mirrored.NumField(); n++ {
+		structVal := mirrored.Field(n)
+		if structVal.IsZero() {
+			t.Fatalf("struct value options not set for field %v",
+				mirrored.Type().Field(n).Name)
+		}
+	}
```
gloriousCode, line 69: Could you clarify this? reflect indirect checks a pointer's value, if its 0 it has no value, but why would adding more fields affect this? d.options also isn't a pointer Is this test just a barrier test to prevent someone from adding too many fields to the struct?

gloriousCode (changes requested):
Thanks for making those changes! Still just looking at code right now.

Something that I have noticed is that initial sync takes significantly longer than `master` branch (though it could be nothing and its just how youve lowered the amount of cmd output). The Initial sync is very CPU intensive. I will look to discover how/why/where its happening on my next review.

exchanges/orderbook/node.go
```diff
@@ -48,62 +48,63 @@ func (s *stack) Push(n *node) {
...
+		// We are going to iterate through slice running man styles
+		// As the old nodes are going to be left justified on this slice we
+		// should just be able to shift the nodes that are still within time
+		// allowance all the way to the left. Not going to resize capacity
+		// because if it can get this big, it might as well stay this big.
+		// TODO: Test and rethink if sizing is an issue
+		nodesLen := atomic.LoadInt32(&s.count)
+		for x := int32(0); x < nodesLen; x++ {
+			// find the first good one, everything to the left can be
+			// reassigned
```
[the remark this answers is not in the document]
shazbert: I guess so, but in that scoped functionality its an old node that was found and we are continuing against that until we find a node that is still under the time limit. I will update the comment to be more specific.

exchanges/orderbook/depth.go
```diff
@@ -0,0 +1,325 @@
...
+	if !atomic.CompareAndSwapUint32(&d.waiting, 1, 0) {
+		// return if no waiting routines
+		return
+	}
+	d.wMtx.Lock()
+	go func() {
+		close(d.wait)
+		d.wg.Wait()
+		d.wait = make(chan struct{})
+		d.wMtx.Unlock()
```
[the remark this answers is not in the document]
shazbert: What your suggesting only alerts one waiting routine, when I close there could be n number of routines that are needed to check the depth sides. I use the mutex locks as a rolling gate so multiple routines can sit and wait and when an alert comes in they can process when they are woken up by the go scheduler downside being there's no ticket system so it can be completely random when routines wake up, so this might be a contention point but we can consider this in later improvements.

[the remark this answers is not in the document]
shazbert: There is a perfomance reason on the depth side if there are no waiting routines then return straight away and only interacts with the initial sync lock and pushes the rest of the work in a routine. Which can let slip more processes until that routine finishes and actuates that unlock. So it gives it a window of breathing room. Which can still stack up potential routines in between.

exchanges/orderbook/depth.go
```diff
@@ -0,0 +1,325 @@
...
+
+// POC: timeInForce allows a kick
+func timeInForce(t time.Duration) kicker {
+	ch := make(chan struct{})
+	go func(ch chan<- struct{}) {
+		time.Sleep(t)
+		close(ch)
+	}(ch)
+	return ch
+}
```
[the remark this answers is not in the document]
shazbert: It's a POC demonstrating that it can cancel waiting for a change on the orderbook. You could use Ticker to actuate every supplied duration but then that would happen every time.Duration until you stopped it. You could just pass in the read only channel, I guess, either way if you want me to delete this functionality I can?

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,675 @@
...
```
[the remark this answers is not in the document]
shazbert: good spotting 👍

exchanges/orderbook/linked_list.go
shazbert: There are subtle differences in the functions and I wanted to make sure there was segregation between the ask and bid methods because of these differences. There is a lot of re-used code but we should go through everything with a fine tooth comb to figure out whats going to be inlined correctly. I prefer (which you can disagree) this expanded form for now to make sure everything is working as intended. Then we can look at breaking everything up. @thrasher-

exchanges/orderbook/linked_list.go
shazbert: current (*tip).prev == prev; prev.next should == *tip. 
I think construction wise there has been a serious fault and it should panic because all subsequant nodes are off. Have you been able to replicate a panic?

exchanges/orderbook/linked_list.go
shazbert: epic find

exchanges/orderbook/depth_test.go
shazbert: No it was just for me really when I was adding in new items options so I remembered to add the assignments over as well, I can remove it if you'd like as its not needed: 
If I add in field gloriousField to the options struct and then run the tests it will complain:
image

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,678 @@
...
+func (ll *linkedList) cleanup(maxChainLength int, stack *stack) {
+	// Reduces the max length of total linked list chain, occurs after updates
+	// have been implemented as updates can push length out of bounds, if
+	// cleaved after that update, new update might not applied correctly.
+	if maxChainLength == 0 || ll.length <= maxChainLength {
+		return
+	}
+
+	n := ll.head
+	for i := 0; i < maxChainLength; i++ {
```
MadCozBadd, line 141: I'm still trying to get my head around this stuff so I could be wrong but it seems like here we are going forward 1 too many times and then we do n.prev.next = nil. Instead we could just check if i == 0 { continue } after if n.next == nil check

MadCozBadd (commented):
Really cool pr!!!

exchanges/orderbook/depth.go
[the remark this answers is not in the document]
gloriousCode: I misspoke :) I meant NewTimer

exchanges/orderbook/depth.go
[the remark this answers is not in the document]
gloriousCode: Yeah I see what you mean with regards to the _n_ number and affecting many channel listeners at once and that's cool. I don't have a solution yet, but I feel like there's probably a better purpose way of doing this and I'll have a think about it.

MadCozBadd (commented):
Two things to consider:
Maybe we just need a singly linked list? (could be done in the future)
Do we need a UpdateById function since within that function we check the price sequence anyways

shazbert:
Good point about the singly linked list we can consider that after. As for the **updateInsertByID** function which seems like it's solely used by Bitfinex, unfortunately, the price levels aren't aggregated together so we will have different ID's at the same price. If we do use another functionality our checksum will fail. Will look into it more when I do the ID alignment pass for Bitfinex.

exchanges/orderbook/depth_test.go
gloriousCode: That's cool. It can always become its own dedicated test

exchanges/orderbook/linked_list.go
gloriousCode: I don't understand the way you wrote your example code sorry, I'll speak to you soon to get a better understanding. prev doesn't get assigned to until the end of the loop. My IDE is flagging it, it will take me longer to understand everything here, but I haven't yet seen a reason not to protect it. It seems like a scenario can exist where this can be hit and prev has either been assigned nil or it hasn't been set yet.

Leaning on a panic isn't something we should be doing. Better to be able to identify the issue and handle it instead.

gloriousCode (changes requested):
Thanks for making all of those changes! End of day review results - Better us keep talking about it while I try to understand what's going on. I think you fixed the slow initial sync potentially with [this](https://github.com/thrasher-corp/gocryptotrader/pull/643/commits/5f2723befb9d4bba4299aac9e5573eea50e44c16) commit! I don't seem to be experiencing it now 🤷

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,618 @@
...
+		// bypassing current node
+		old.prev.next = old.next
+		old.next.prev = old.prev
+	}
+	ll.length--
+	return old
+}
+
+// insert inserts at a tip target (can inline)
+func insert(ll *linkedList, tip **node, updt Item, stack *stack) {
```
MadCozBadd, line 538: A little hard to explain this one so we will definitely talk about this tomorrow but I'll have a go here anyways. Currently insert has no way of inserting a node at the end since if I'm understanding correctly, we insert a node just before where the tip is pointing. I think insert function should be really generic and should be able to insert a node wherever it needs to be inserted whether it be in the end or the beginning. This will help prevent the user from having to consider which type of insert needs to be used in which scenario.

MadCozBadd (commented):
Great changes!!! Few more to go

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,618 @@
...
+	for n != nil {
+		pruned++
+		pending := n.next
+		stack.Push(n)
+		n = pending
+	}
+	ll.length -= pruned
+}
+
+// Liquidity returns total depth liquidity
```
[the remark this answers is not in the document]
MadCozBadd: this one is up for discussion @thrasher- @gloriousCode

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,618 @@
...
+
+// insert inserts at a tip target (can inline)
+func insert(ll *linkedList, tip **node, updt Item, stack *stack) {
+	n := stack.Pop()
+	n.value = updt
+	n.next = *tip
+	n.prev = (*tip).prev
+	if (*tip).prev == nil { // Tip is at head
+		// Replace head which will push everything to the right
+		// when this node will reference new node below
```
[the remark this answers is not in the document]
shazbert: if (*tip).prev == nil then the tip is at the head position. *tip == ll.head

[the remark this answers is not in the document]
MadCozBadd: Yes but we are adding a new node there just before it which now isn't representing head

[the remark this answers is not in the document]
shazbert: We do that by dereferencing the tip **node and reassigning its underlying node address with the new address n
[here](https://github.com/thrasher-corp/gocryptotrader/pull/643/files/0b822444295b5bdaa93486dad8d116704a4cd70b#diff-fad5980df4c3abb89d89d4eb4dad8ecff32c84217d66c25617aa641417e260adR546)

exchanges/orderbook/linked_list.go
shazbert: No user should be calling this function at all. It's only used and inlined into the function that requires it. The alignment is handled by the calling function. And it's stripped bare to do only one thing in context with that calling function.

gloriousCode (changes requested):
End of day review comments 🎉

main.go
```diff
@@ -63,8 +63,10 @@ func main() {
 	flag.BoolVar(&settings.EnableTradeSyncing, "tradesync", false, "enables trade syncing for all enabled exchanges")
 	flag.IntVar(&settings.SyncWorkers, "syncworkers", engine.DefaultSyncerWorkers, "the amount of workers (goroutines) to use for syncing exchange data")
 	flag.BoolVar(&settings.SyncContinuously, "synccontinuously", true, "whether to sync exchange data continuously (ticker, orderbook and trade history info")
-	flag.DurationVar(&settings.SyncTimeout, "synctimeout", engine.DefaultSyncerTimeout,
-		"the amount of time before the syncer will switch from one protocol to the other (e.g. from REST to websocket)")
+	flag.DurationVar(&settings.SyncTimeoutREST, "synctimeoutrest", engine.DefaultSyncerTimeoutREST,
+		"the amount of time before the syncer will switch from rest protocol to the streaming protocol (e.g. from REST to websocket)")
+	flag.DurationVar(&settings.SyncTimeoutStream, "synctimeoutstream", engine.DefaultSyncerTimeoutStream,
```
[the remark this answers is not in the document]
shazbert: Yeah, I just wanted to call it 'stream' so if in the future FIX implementations are added it can use the same settings variable. But I will change.

exchanges/orderbook/depth.go
```diff
@@ -0,0 +1,331 @@
...
+	waiting uint32
+	wg      sync.WaitGroup
+	wMtx    sync.Mutex
+	// -----
+
+	mux *dispatch.Mux
+	id  uuid.UUID
+
+	options
+	sync.Mutex
```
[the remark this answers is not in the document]
shazbert: Segregation mainly so routine(s) waiting for a change does not interfere with the main lock until it gets an alert. As for the `wMtx` that refers to waiting or queue not necessarily a 'write'. I think your suggestion about using a RWMutex might cause overhead and slow execution, because there might be a lot of write locking if there are many routines waiting or about to wait.

exchanges/orderbook/depth.go
```diff
@@ -0,0 +1,331 @@
...
+	d.flush()
+	d.Unlock()
+}
+
+// flush will pop entire bid and ask node chain onto stack when invalidated or
+// required for full flush when resubscribing
+func (d *Depth) flush() {
+	d.bids.load(nil, d.stack)
+	d.asks.load(nil, d.stack)
+}
```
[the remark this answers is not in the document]
shazbert: Yeah, that's a good point, was going to use this to invalidate a depth if an error occurs so there's no way a strategy can traverse through liquidity. Not being used for now so I will remove it. Will think of another way to do this in another PR.

exchanges/orderbook/depth.go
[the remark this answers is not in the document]
gloriousCode: Thanks for clarifying!

exchanges/orderbook/linked_list_test.go
```diff
@@ -0,0 +1,1170 @@
...
+		{Price: 3, Amount: 1, ID: 3},
+		{Price: 5, Amount: 1, ID: 5},
+		{Price: 7, Amount: 1, ID: 7},
+		{Price: 9, Amount: 1, ID: 9},
+		{Price: 11, Amount: 1, ID: 11},
+	}
+	a.load(asksSnapshot, s)
+
+	// Update one instance with matching ID
+	a.updateInsertByID(Items{
```
gloriousCode, line 537: unhandled errs

exchanges/orderbook/orderbook.go
```diff
@@ -18,144 +18,183 @@ func Get(exchange string, p currency.Pair, a asset.Item) (*Base, error) {
...
 func (s *Service) Update(b *Base) error {
-	name := strings.ToLower(b.ExchangeName)
+	name := strings.ToLower(b.Exchange)
 	s.Lock()
-	m1, ok := s.Books[name]
+	m1, ok := s.books[name]
 	if !ok {
-		m1 = make(map[asset.Item]map[*currency.Item]map[*currency.Item]*Book)
-		s.Books[name] = m1
+		uuid, err := s.Mux.GetID()
```
gloriousCode, line 54: rename var since it matches a package name

exchanges/orderbook/depth.go
```diff
@@ -0,0 +1,325 @@
...
+	}
+}
+
+// Publish alerts any subscribed routines using a dispatch mux
+func (d *Depth) Publish() {
+	err := d.mux.Publish([]uuid.UUID{d.id}, d.Retrieve())
+	if err != nil {
+		log.Errorf(log.ExchangeSys, "Cannot publish orderbook update to mux %v", err)
+	}
+}
```
gloriousCode, line 50: Please add test

exchanges/stream/buffer/buffer_types.go
```diff
@@ -23,9 +27,16 @@ type Orderbook struct {
...
+// orderbook depth
 type orderbookHolder struct {
-	ob     *orderbook.Base
+	ob     *orderbook.Depth
 	buffer *[]Update
+	// Reduces the amount of outbound alerts to the data handler for example
+	// coinbasepro can have up too 100 updates per second introducing overhead.
+	// The sync agent only requires an alert every 15 seconds for a specific
+	// currency.
+	ticker *time.Ticker
```
gloriousCode, line 39: Can this functionality be overridable for people who want all the datas displayed?

exchanges/orderbook/depth.go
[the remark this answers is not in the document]
gloriousCode: The close function seems like the best way to go about this, resolving 🦀

gloriousCode (changes requested):
Mini review before going in`*Depth` for the linkedlist functionality

exchanges/stream/buffer/buffer_types.go
shazbert: I guess it could link up to config; but why you would want to spam your file or stdout? Is it checking throughput on implementation?

gloriousCode: For the same reason we have verbose and debug logs

shazbert: Added but not linked up to config, for now, I have just linked it up to the verbose bool.

gloriousCode: Thanks! I think that's good enough 🎉

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,608 @@
...
+					insertAtTip(&ll.linkedList, tip, updts[x], stack)
+				}
+				break // Continue updates
+			}
+
+			if (*tip).next == nil { // Tip is at tail
+				if updts[x].Amount > 0 {
+					insertAtTail(&ll.linkedList, tip, updts[x], stack)
+				}
+				break
```
MadCozBadd, line 198: since tip is at tail maybe we don't need the break 👀

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,608 @@
...
+
+			if (*tip).value.Price < updts[x].Price { // Insert
+				if updts[x].Amount > 0 { // Filter delete, should already be
+					insertAtTip(&ll.linkedList, tip, updts[x], stack)
+				}
+				break // Continue updates
+			}
+
+			if (*tip).next == nil { // Tip is at tail
+				if updts[x].Amount > 0 {
```
MadCozBadd, line 195: can pull out this if statement, revert it and if amount is <= 0 break right there. this should ideally save some iterations

exchanges/orderbook/linked_list.go
shazbert: We don't check for nil in the for loop anymore so this will add another node.

exchanges/orderbook/linked_list.go
shazbert: Good spotting!

shazbert: Actually, from re-looking at code, this was an optimisation when an exchange sends a delete update for a non-existing price that occurred on some exchanges. I will add comments to reflect that justification.

MadCozBadd: :

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,619 @@
...
+	}
+
+	// Push has references to dangling nodes that need to be removed and pushed
+	// back onto stack for re-use
+	var push *node
+	// Cleave unused reference chain from main chain
+	if prev == nil {
+		// The entire chain will need to be pushed back on to stack
+		push = *tip
+		ll.head = nil
```
gloriousCode, line 57: This scenario looks like it can only exist if `items` is nil/len(0) because prev is set to tip after ensuring tip is not nil. So in this scenario, wouldn't that also make `tip` = nil too? If this then makes `push` = nil, then should this kind of item length check be at the start and return earlier?

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,619 @@
...
+	}
+	(*bookmark).value = updt
+	(*bookmark).next.prev = (*bookmark).prev
+	if (*bookmark).prev == nil { // Bookmark is at head
+		*head = (*bookmark).next
+	} else {
+		(*bookmark).prev.next = (*bookmark).next
+	}
+	tip.next = *bookmark
+	(*bookmark).prev = tip
```
gloriousCode, line 616: These lines overwrite what you've just done in 613.
bookmark.prev.next = bookmark.next
tip.next = bookmark
bookmark.prev = tip
.*. bookmark.prev.next  = bookmark
then finally you set bookmark.next = nil

This makes it seem like a different scenario. I may need a summary of what's going on here

exchanges/orderbook/orderbook_types.go
```diff
@@ -70,31 +66,54 @@ type Item struct {
...
 	// user configuration
 	VerificationBypass bool `json:"-"`
 	// HasChecksumValidation defines an allowance to bypass internal
 	// verification if the book has been verified by checksum.
 	HasChecksumValidation bool `json:"-"`
+	// RestSnapshot defines if the depth was applied via the REST protocol thus
+	// an update cannot be applied via websocket mechanics and a resubscription
+	// would need to take place to maintain book integrity
+	RestSnapshot bool
+	// Checks if the orderbook needs ID alignment as well as price alignmentment
```
gloriousCode, line 97: alignment

gloriousCode (changes requested):
There's still more for me to understand what's going on, but I thought I should provide reviews more frequently here to help/gain understanding.

exchanges/orderbook/node.go
```diff
@@ -0,0 +1,111 @@
...
+func newStack() *stack {
+	s := &stack{}
+	go s.cleaner()
+	return s
+}
+
+// Push pushes a node pointer into the stack to be reused the time is passed in
+// to allow for inlining which sets the time at which the node is theoretically
+// pushed to a stack.
+func (s *stack) Push(n *node, tn time.Time) {
```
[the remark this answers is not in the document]
shazbert: time.Now() doesn't inline so I would rather inline the push function and do one time.Now() function call, instead of function call push and then function call time.Now(). If they are nanoseconds apart or microseconds apart, it doesn't really make a big difference as the cleanup is 30-second intervals anyway, if it's sitting on the stack for that long it's going to be cleaned up eventually. I see your point though but in this instance, I don't see an actual gain for being precise.

exchanges/orderbook/node.go
```diff
@@ -0,0 +1,111 @@
...
+	for range tt.C {
+		atomic.StoreUint32(&s.sema, cleanerActive)
+		// As the old nodes are going to be left justified on this slice we
+		// should just be able to shift the nodes that are still within time
+		// allowance all the way to the left. Not going to resize capacity
+		// because if it can get this big, it might as well stay this big.
+		// TODO: Test and rethink if sizing is an issue
+		nodesLen := atomic.LoadInt32(&s.count)
+		for x := int32(0); x < nodesLen; x++ {
+			if time.Since(s.nodes[x].shelved) > defaultAllowance {
```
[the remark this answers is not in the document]
shazbert: I don't think so. If they are located on the stack, then they aren't being utilized (not on the linked list chain). I would be expecting that all nodes that aren't being used, pushed to the stack. When updates stop coming through the ones on the stack will be 'cleaned' (reassigned and the node count reduced, but not garbage collected because they are still referenced on the slice beyond the node count up to the capacity of the slice). The other ones in the linked list shouldn't be affected.

exchanges/orderbook/linked_list.go
shazbert: No, if there is already data on the linked list and we want to purge the reference and still push the entire chain on the stack *tip will still be populated. I guess I can still add a *tip == nil check and return on ln: 55

exchanges/orderbook/orderbook_types.go
shazbert: Its doubly aligned 😆

exchanges/orderbook/linked_list.go
shazbert: `ln: 613` removes the reference of the current bookmarked node in the chain.
`tip.next = *bookmark` inserts the bookmark after the tip node
`bookmark.prev = tip` reassigns the previous relationship with the current tip 
 `bookmark.next = nil` because we are at the tail end of the chain we nill this reference out
`.*. bookmark.prev.next = bookmark`  != `bookmark.prev.next = bookmark.next`

Have you come across some problems in testing?

exchanges/orderbook/node.go
```diff
@@ -0,0 +1,111 @@
...
+type node struct {
+	value Item
+	next  *node
+	prev  *node
+	// Denotes time pushed to stack, this will influence cleanup routine when
+	// there is a pause or minimal actions during period
+	shelved time.Time
+}
+
+// stack defines a FIFO list of reusable nodes
```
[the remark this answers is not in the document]
shazbert: Yeah, I think so!

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,619 @@
...
+// 1) node ID found amount amended (best case)
+// 2) node ID found amount and price amended and node moved to correct position
+// (medium case)
+// 3) Update price exceeds traversal node price before ID found, save node
+// address for either; node ID matches then re-address node or end of depth pop
+// a node from the stack (worst case)
+func (ll *bids) updateInsertByID(updts Items, stack *stack) error {
+updates:
+	for x := range updts {
+		if updts[x].Amount == 0 {
```
MadCozBadd, line 224: we could change that to <= 0 and rename the error accordingly

MadCozBadd (commented):
Have tried faulting the logic a during the last round of testing as well as this round but no issues! tACK from me soon just 1 change

exchanges/orderbook/node.go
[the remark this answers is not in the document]
gloriousCode: Yeah in the end, since its inlined, the time.Now() delay isnt going to be much. The main thing is still around taking a variable time.Time, opening the possibility of variance. Its not happening now, but the only value you want is `time.Now()` but this opens up the possibility of other values being passed in which gives potential longer term debug fun. I understand you are keen on inlining these functions, so would you be more open to handling the `n`'s properties before pushing and removing the time variable?
This still allows for inlining
```
func (n *node) prepareForShelving() {
	n.next = nil
	n.prev = nil
	n.value = Item{}
	n.shelved = time.Now()
}
```
and then its 
```
n.prepareForShelving()
stack.Push(n)
```
Or even something that returns `n` so that you can call `stack.Push(prepareForShelving(n))`

exchanges/orderbook/linked_list.go
gloriousCode: Yeah, this function doesn't have a test, so I had to write one:
```

func TestShiftBookmark(t *testing.T) {
	prev := &node{
		value: Item{
			ID: 1336,
		},
		next:    nil,
		prev:    nil,
		shelved: time.Time{},
	}
	next := &node{
		value: Item{
			ID: 1338,
		},
		next:    nil,
		prev:    prev,
		shelved: time.Time{},
	}
	bookmark := &node{
		value: Item{
			ID: 1337,
		},
		next:    next,
		prev:    prev,
		shelved: time.Time{},
	}
	tip := &node{
		value: Item{
			ID: 69420,
		},
		next:    next,
		prev:    prev,
		shelved: time.Time{},
	}
	updt := Item{
		ID: 1339,
	}
	prev.next = bookmark
	next.prev = bookmark
	shiftBookmark(tip, &bookmark, nil, updt)
}

```

The line `	(*bookmark).prev = tip` does still overwrite what happens in 
```
 else {
		(*bookmark).prev.next = (*bookmark).next
	}
```
But maybe you can help with clarifying how I've structured my test (and then please add one :D)

exchanges/orderbook/depth.go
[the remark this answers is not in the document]
gloriousCode: re-unexport

exchanges/orderbook/linked_list.go
shazbert: Added tests for this.

exchanges/orderbook/node.go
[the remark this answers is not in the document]
shazbert: Added type for time and constructor so as to not change much.

exchanges/orderbook/depth.go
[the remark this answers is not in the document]
shazbert: This should be done

MadCozBadd (commented):
Didn't get a chance to review today but just saw this when I glanced over the recent commits

exchanges/orderbook/node.go
```diff
@@ -40,17 +40,25 @@ func newStack() *stack {
...
+	return now(time.Now())
+}
+
 // Push pushes a node pointer into the stack to be reused the time is passed in
 // to allow for inlining which sets the time at which the node is theoretically
 // pushed to a stack.
-func (s *stack) Push(n *node, tn time.Time) {
+func (s *stack) Push(n *node, tn now) {
 	if atomic.LoadUint32(&s.sema) == cleanerActive {
 		// Cleaner is activated, for now we can derefence pointer
```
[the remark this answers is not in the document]
shazbert: That is exactly how that word should be; whatever that word is? 🤷 😆 Will change.

MadCozBadd (approved):
tACK!!! tested with all sorts of illiquid currencies on various exchanges. No issues at all!!! Awesome stuff. Just need to make appveyor happy by the way.

exchanges/stream/buffer/buffer.go
```diff
@@ -57,27 +64,40 @@ func (w *Orderbook) validate(u *Update) error {
...
 			u.Pair,
 			u.Asset)
 	}
 
+	// Checks for when the rest protocol overwrites a streaming dominated book
+	// will stop updating book via incremental updates. This occurs because our
+	// sync manager (engine/sync.go) timer has elapsed for streaming. Usually
+	// because the book is highly illiquid.
+	if book.ob.IsRestSnapshot() {
+		return fmt.Errorf("%w for Exchange %s CurrencyPair: %s AssetType: %s",
```
gloriousCode, line 89: I'm not expecting this in this PR, but this change seems like there is no recovery for websocket. If this ever gets hit, any websocket orderbook update that comes through will see this error as there is nothing to reset the orderbook to not have this be true. Perhaps add a TODO to signify that such an update will happen?
For a simple proof of concept, I basically mean reading the error on the {{exchange}}_websocket file to check that its `ErrRestOverwrite` and un/resubscribe. 
Or maybe just a first time `warn` the user to consider extending `synctimeoutwebsocket`

gloriousCode (changes requested):
Thanks for making those changes! Most of my nits are small now 🤙 
This isn't a full request, but a good test you could add to all exchanges (which don't have it yet) is a test for  `UpdateOrderbook` wrapper function. The test proves that your orderbook changes work across all exchanges, which calls the API endpoint, translates to structs, verifies its all in order, converts to linked list and returns an orderbook.Base. So its a bit of a full process cycle test.

exchanges/orderbook/depth.go
```diff
@@ -320,3 +274,79 @@ func (d *Depth) IsFundingRate() bool {
...
+	a.wg.Add(1)
+	if atomic.CompareAndSwapUint32(&a.sema, 0, 1) {
+		a.forAlert = make(chan struct{})
+	}
+	go a.hold(reply, kick)
+	a.m.Unlock()
+	return reply
+}
+
+// hold waits on both channels in the event that the routine has finished or an
```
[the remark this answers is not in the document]
shazbert: What's the difference; I thought they were 'both' synonyms?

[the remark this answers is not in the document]
gloriousCode: Yeah I misread your comment 👀 Thought you were suggesting it waits until both channels are hit which isn't the case

exchanges/orderbook/depth.go
```diff
@@ -320,3 +274,79 @@ func (d *Depth) IsFundingRate() bool {
...
+	defer a.wg.Done()
+	select {
+	// In a select statement, if by chance there is no receiever or its late,
+	// we can still close and return, limiting dead-lock potential.
+	case <-a.forAlert: // Main waiting channel from alert
+		select {
+		case ch <- false:
+		default:
+		}
+	case <-kick: // This can be nil.
```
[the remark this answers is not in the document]
shazbert: It's for whatever you want, timeout, shutdown, all of the above, it's very generic. You can define your timer/ticker whatever you want to be scoped to the caller e.g. 

```go 
var kick = make(chan struct{})
go func () {
timerSlimer := time.NewTimer(time.Second)
select {
case <- timerSlimer.C:
case <-shutdown: // This is the case I will probably use.
}
close(kick)

// then call wait
for {
select {
    case invalid := <-someDepth.Wait(kick):
    // or  case invalid := <-someDepth.Wait(nil):
        if invalid {
        continue
        }
        // do work
    case <-shutdown: // This also links up with the go func above and then within wait will also allow the goroutine to exit cleanly even though it's not really needed only if liquidity is zero then we might have a dangly routine. 
        return
}
}
}()

```

Also, that kicker type I removed was the basis for what was going to be represented above.

exchanges/orderbook/linked_list_test.go
```diff
@@ -0,0 +1,1365 @@
...
+			ID: 69420,
+		},
+		next:    nil, // In this case tip will be at tail
+		prev:    nil,
+		shelved: time.Time{},
+	}
+
+	tipprev := &node{
+		value: Item{
+			ID: 69420,
```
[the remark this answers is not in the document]
shazbert: Yeah, it should, will change, but what really matters is the pointer address.

exchanges/orderbook/depth.go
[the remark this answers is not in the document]
gloriousCode: Fair enough to be generic, thanks for the explanation. Also, `chan interface{}` does not allow for receiving any type over a chan like how an `interface{}` param would, so I'll resolve this 🤙

exchanges/orderbook/linked_list.go
```diff
@@ -0,0 +1,618 @@
...
+// Amount returns total depth liquidity and value
+func (ll *linkedList) amount() (liquidity, value float64) {
+	for tip := ll.head; tip != nil; tip = tip.next {
+		liquidity += tip.value.Amount
+		value += tip.value.Amount * tip.value.Price
+	}
+	return
+}
+
+// Retrieve returns a full slice of contents from the linked list
```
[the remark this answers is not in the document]
gloriousCode: amount too!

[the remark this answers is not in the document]
shazbert: That's done

exchanges/orderbook/linked_list_test.go
```diff
@@ -0,0 +1,1377 @@
...
+		{Price: 5, Amount: 1, ID: 5},
+		{Price: 7, Amount: 1, ID: 7},
+		{Price: 9, Amount: 1, ID: 9},
+		{Price: 11, Amount: 1, ID: 11},
+	}
+	a.load(askSnapshot, s)
+
+	liquidity, value := a.amount()
+	if liquidity != 6 {
+		t.Fatalf("incorrect liquidity calculation expected 6 but receieved %f", liquidity)
```
MadCozBadd, line 1261: 👀 and a couple lines below too

MadCozBadd (approved):
tick tack!

engine/syncer.go
```diff
@@ -18,8 +18,9 @@ const (
 	SyncItemOrderbook
 	SyncItemTrade
 
-	DefaultSyncerWorkers = 15
-	DefaultSyncerTimeout = time.Second * 15
+	DefaultSyncerWorkers          = 15
+	DefaultSyncerTimeoutREST      = time.Second * 15
+	DefaultSyncerTimeoutWebsocket = time.Minute
```
gloriousCode, line 23: As discussed, the issue I've experienced with this is with an Exchange like Binance. Where an asset type which is unsupported by the websocket is delayed by a minute in order to switch to REST. The initial sync cannot complete until thie minute expires and sync continues. My short term solution is to reduce this timer or have an ability to identify if an asset type cannot be synced via websocket even if websocket is true. Anything larger requires its own PR

gloriousCode (changes requested):
Thanks for making those changes! I'll be comparing orderbook streams against exchange data soon. Here are some near-final comments on things

engine/syncer.go
shazbert: Have added a quick fix for this but I do get this warning when spot is not enabled, which we can probably address later: [WARN] |WEBSOCKET| 09/04/2021 10:26:11 |Binance websocket has been disconnected. Reason: read tcp 192.168.1.109:51983->52.68.174.232:9443: use of closed network connection
I am assuming there is nothing to subscribe so it opens a connection and then returns.

exchanges/exchange.go
```diff
@@ -1266,3 +1267,38 @@ func (u URL) String() string {
 func (e *Base) UpdateOrderExecutionLimits(a asset.Item) error {
 	return common.ErrNotYetImplemented
 }
+
+// AssetWebsocketFunctionality defines the availability of websocket
+// functionality to the specific asset type. TODO: Deprecate as this is a temp
+// item to address certain limitations quickly.
+type AssetWebsocketFunctionality struct {
```
gloriousCode, line 1277: WebsocketAssetSupport

exchanges/exchange.go
```diff
@@ -1266,3 +1267,38 @@ func (u URL) String() string {
...
+	NoFunctionality map[asset.Item]bool
```
gloriousCode, line 1278: Unsupported

exchanges/exchange.go
```diff
@@ -1266,3 +1267,38 @@ func (u URL) String() string {
...
+	m               sync.RWMutex
+}
+
+// SetAssetWebsocketFunctionalityOff forces websocket functionality for the
+// supplied asset item to off. In the case that websocket functionality has not
+// yet been implemented for that specific asset type.This is a base method to
+// check availabilty of asset type
+func (b *Base) SetAssetWebsocketFunctionalityOff(aType asset.Item) error {
```
gloriousCode, line 1286: DisableWebsocketAssetSupport

exchanges/exchange.go
```diff
@@ -1266,3 +1267,38 @@ func (u URL) String() string {
...
+	if !b.SupportsAsset(aType) {
+		return fmt.Errorf("%s %w this should be called after asset type initialisation",
+			aType,
+			asset.ErrNotSupported)
```
gloriousCode, line 1290: I don't think it needs the flavour text since it may just be unsupported. Just `("%w %s", asset.ErrNotSupported, aType)`

exchanges/exchange.go
```diff
@@ -1266,3 +1267,38 @@ func (u URL) String() string {
...
+		b.AssetWebsocketFunctionality.NoFunctionality = make(map[asset.Item]bool)
+	}
+	b.AssetWebsocketFunctionality.NoFunctionality[aType] = true
+	b.AssetWebsocketFunctionality.m.Unlock()
+	return nil
+}
+
+// IsAssetWebsocketFunctional checks to see if the supplied asset type has a
+// form of websocket functionality.
+func (a *AssetWebsocketFunctionality) IsAssetWebsocketFunctional(aType asset.Item) bool {
```
gloriousCode, line 1303: IsWebsocketAssetSupported

exchanges/exchange.go
gloriousCode, line 1277: Add to `exchange_types.go`

engine/syncer.go
```diff
@@ -560,6 +546,13 @@ func (e *ExchangeCurrencyPairSyncer) Start() {
 				continue
 			}
 
+			wsAssetFunctional := exchanges[x].IsAssetWebsocketFunctional(assetTypes[y])
+			if !wsAssetFunctional {
+				log.Warnf(log.SyncMgr,
+					"%s asset type %s websocket functionality has been turned off, REST fetching only.",
```
gloriousCode, line 552: `"%s asset type %s websocket functionality is unsupported, REST fetching only",`

engine/syncer.go
gloriousCode: I can confirm that its the reason for the error. I don't mind this so much, the user has been warned that the asset is unsupported. I don't think we need to forcefully disable the websocket on them, the user can still make modifications via GRPC or they can choose to disable the websocket themselves/enable a supported asset.

gloriousCode (changes requested):
I like your quick fix. It addresses the issue without sweeping changes. Thank you for that! These nits are just semantics on the naming conventions

exchanges/orderbook/node_test.go
```diff
@@ -0,0 +1,101 @@
...
+		t.Fatalf("incorrect stack count expected %v but received %v", 0, s.getCount())
+	}
+}
+
+// Display nodes for testing purposes
+func (s *stack) Display() {
+	for i := int32(0); i < s.getCount(); i++ {
+		fmt.Printf("NODE IN STACK: %+v %p \n", s.nodes[i], s.nodes[i])
+	}
+	fmt.Println("Tatal Count:", s.getCount())
```
MadCozBadd, line 62: 👀

MadCozBadd (approved):
reTACK

gloriousCode (changes requested):
Tested each exchange for verification problems. Within the context of this PR, only came across the Bitmex issue 🎉

exchanges/exchange_test.go
```diff
@@ -2353,7 +2353,7 @@ func TestAssetWebsocketFunctionality(t *testing.T) {
 	}
 
 	if b.IsAssetWebsocketSupported(asset.Spot) {
-		t.Fatal("error asset is not turned off, so this should be functional")
+		t.Fatal("error asset is not turned off, unexpected responsel")
```
gloriousCode, line 2356: image

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -508,11 +508,11 @@ func (b *Bitmex) processOrderbook(data []OrderBookL2, action string, p currency.
 					data[i].Side)
```
gloriousCode, line 508: Please disable asset weboskcet for `downsideprofitcontract` and `index`. They don't support orderbooks, but also don't receive updates via websocket

gloriousCode (changes requested):
Thank you very much for investigating and solving the orderbook verification issues. Adding in a verify again helps with confidence. I've tested each exchange and each asset type and these are my findings

gloriousCode:
Please fix the orderbook_test.go tests and then I think its good 🤞

exchanges/orderbook/orderbook_test.go
```diff
@@ -588,37 +564,34 @@ func deployUnorderedSlice() []Item {
...
-	SortBids(b.Bids)
+	b.Bids.SortBids()
 	err = b.Verify()
 	if err != nil {
 		t.Fatal(err)
 	}
 }
 
-func deploySliceOrdered() []Item {
+func deploySliceOrdered() Items {
```
MadCozBadd, line 594: This is a really useful function and theres a good chance this, along with deploySliceUnordered, might get used by someone else in the future. However the next person is restricted to a 1000 length slice. If the length is modified by someone in the future, your current tests may start failing. So for that reason could we add in a length param for these two functions

MadCozBadd (commented):
Everything else looks good!!

exchanges/orderbook/depth_test.go
```diff
@@ -326,40 +326,81 @@ func TestWait(t *testing.T) {
 	wg.Add(100)
 	for x := 0; x < 100; x++ {
 		go func() {
-			<-wait.Wait(nil)
+			w := wait.Wait(nil)
+			wg.Done()
+			<-w
 			wg.Done()
 		}()
 	}
```
gloriousCode: While it took me awhile to try to understand how these tests work, I think it would be good to verify the wait values in all instances to verify that they were closed in the fashion you expected

exchanges/orderbook/depth_test.go
```diff
@@ -326,40 +326,81 @@ func TestWait(t *testing.T) {
...
-	time.Sleep(time.Millisecond)
 	go wait.alert()
-	go close(ch)
+	close(ch)
 	wg.Wait()
+	select {
+	case <-wait.Wait(ch):
+		t.Fatal("leaky waiter")
+	default:
+	}
```
gloriousCode, line 405: I'm not sure this part is needed as you close the channel `ch` above and `wait.alert()` closes the `forAlert` channel

exchanges/orderbook/orderbook_test.go
```diff
@@ -716,17 +689,17 @@ func BenchmarkSortAsksAscending(b *testing.B) {
 	s := deploySliceOrdered()
 	for i := 0; i < b.N; i++ {
 		ts := append(s[:0:0], s...)
-		SortAsks(ts)
+		ts.SortAsks()
 	}
 }
 
 // 12565	     97257 ns/op	   49208 B/op	       3 allocs/op
```
gloriousCode, line 696: Unfortunately, all of these benchmarks are now an order of magnitude greater than they used to be (at least on my machine™) . Please update so people can get a more accurate indication of what they are for the any future changes to compare to.

gloriousCode (changes requested):
Just some minor extra comments 📖

exchanges/orderbook/depth_test.go
shazbert: It's to ensure we do not let slip routine calls and check the 'depth' after the 'gate' has closed. It should only theoretically actuate when there is an alert/change in the book. But you bring up a good point about the channel `ch` it should actuate because it's closed. Will check that.

exchanges/orderbook/depth_test.go
shazbert: Will do

exchanges/orderbook/orderbook_test.go
shazbert: Will redo these all

shazbert: On my machine, they are about the same 🤷 CPU: Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz. Changed the ones that needed the change.

shazbert: Im actually getting faster on that actual benchmark: `15340	     75205 ns/op	   49198 B/op	       3 allocs/op`

exchanges/orderbook/orderbook_test.go
shazbert: When somebody needs to change their slice length they can update everything to accommodate.

exchanges/orderbook/orderbook_test.go
gloriousCode: To be clearer, here's my comparison:
| Benchmark                   | Commented benchmark                                          | Resulting benchmark                                          | Difference in ns/op |
| --------------------------- | ------------------------------------------------------------ | ------------------------------------------------------------ | ------------------- |
| BenchmarkReverse            | `705985	      1856 ns/op	       0 B/op	       0 allocs/op` | `521190	      2185 ns/op	       0 B/op	       0 allocs/op` | +329                |
| BenchmarkSortAsksDecending  | `20209	     56385 ns/op	   49189 B/op	       2 allocs/op` | `20868	     57500 ns/op	   49180 B/op	       2 allocs/op` | +1115               |
| BenchmarkSortBidsAscending  | `14924	     79199 ns/op	   49206 B/op	       3 allocs/op` | `14114	    104291 ns/op	   49199 B/op	       3 allocs/op` | +25092              |
| BenchmarkSortAsksStandard   | `9842        133761 ns/op      49194 B/op         2 allocs/op` | `9694	    120073 ns/op	   49186 B/op	       2 allocs/op` | -13588              |
| BenchmarkSortBidsStandard   | `7058        155057 ns/op      49214 B/op         3 allocs/op` | `8169	    160649 ns/op	   49204 B/op	       3 allocs/op` | +5592               |
| BenchmarkSortAsksAscending  | `20565        57001 ns/op      49188 B/op         2 allocs/op` | `18172	     65505 ns/op	   49181 B/op	       2 allocs/op` | +8504               |
| BenchmarkSortBidsDescending | `12565        97257 ns/op      49208 B/op        3 allocs/op` | `13639	     88840 ns/op	   49199 B/op	      3 allocs/op` | -8417               |
| BenchmarkDuplicatingSlice   | `923154        1169 ns/op       4096 B/op         1 allocs/op` | `83696	     12014 ns/op	   49153 B/op	      1 allocs/op` | +10845              |
| BenchmarkCopySlice          | `705922       1546 ns/op       4096 B/op         1 allocs/op` | `93628	     14365 ns/op	   49153 B/op	      1 allocs/op` | +12819              |

shazbert: There is always variance on the run, it's a benchmark and could swing wildly between go versions and hardware variability. I can delete the comments on the benchmarks and remove BenchmarkDuplicatingSlice with BenchmarkCopySlice as there isn't much difference anymore and we aren't exactly using them. The others can stay for reference i.e. why we would use reverse instead of sort functionality, but if you want all these deleted I can. 🤷

gloriousCode: Nah,you're right, its not really something worth focusing on. I came at it from the perspective of if someone were to make changes and then compare it to the comments, one could lead to think they've impacted things more than reality

exchanges/orderbook/depth.go
```diff
@@ -0,0 +1,352 @@
...
+		select {
+		case ch <- true:
+		default:
+		}
+	}
+	a.wg.Done()
```
gloriousCode, line 350: Last question I think! Should this be below close?

exchanges/orderbook/orderbook.go
```diff
@@ -18,144 +18,186 @@ func Get(exchange string, p currency.Pair, a asset.Item) (*Base, error) {
...
-			exchange)
+		return dispatch.Pipe{}, fmt.Errorf("%w for %s exchange",
+			errCannotFindOrderbook, exchange)
 	}
-	return service.mux.Subscribe(id)
+	return service.Mux.Subscribe(exch.ID)
+}
+
+// GetDispatch returns the service mux for subsystem alerting
+func GetDispatch() *dispatch.Mux {
```
gloriousCode, line 47: This is unused now

exchanges/orderbook/depth.go
shazbert: I put this before close because this directly relates to the alert routine's a.wg.Wait(). It has a higher priority because it wants to swap back to a neutral state and unlock, if it stays locked the biggest impedance on the alert system is the Wait() call because there can be updates occurring while the alert system is in this locked state (potentially) and at this point, there could be routines blocked by the mutex slip ring functionality and not actually get alerted (small chance). Also, an alert has happened either by forAlert or kick, the close channel just makes sure if the calling routine is late it will still fire in this context until a new Wait() call.
That was my intention anyway, if logic is incorrect let me know.

gloriousCode: Yeah it doesn't look like it causes much impact, so its all g 👍

gloriousCode (approved):

MadCozBadd (approved):
tACK!!!!

gloriousCode (changes requested):
Nice consolidation in the latest updates! Tested and its still all working. Just one critical bug I came across

gloriousCode (approved):
retACK! Thanks for fixing that up

thrasher- (approved):
tACK, epic PR @shazbert !!

vazha:
Gentlemen, I must point out that this does not work at all for the Kraken. I mean reconnecting a websocket after a disconnection.

PR #644, opened 2021-03-08
Gemini: Fix pair handling and upgrade websocket connection for public market data

thrasher-:
# PR Description

User whiskeywhiskey reported a pair handling and websocket connection cycle issue impacting Gemini. Originally Gemini had pairs with lengths of 6 characters, however they now support a wide range of new pairs. This performs a config upgrade to fix this issue. Note: the API does not provide a separator for pairs, so a fix similar to Bitfinex has been implemented until we can accurately parse a pair to its individual components.

This also switches from a multiconnection per asset for public market data (trades, orderbook, candles and auction events) to a single connection supporting multiple pairs via their market data v2 websocket API. This also implements subscribing/unsubscribing of pairs through `gctcli`. Ticker streaming updates are still not supported via their websocket API, so that will still have to be fetched via REST.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

- [N/A] Any dependent changes have been merged and published in downstream modules


shazbert (changes requested):
Basic nits! thanks for fixing!

exchanges/gemini/gemini_websocket.go
```diff
@@ -411,60 +545,51 @@ func (g *Gemini) wsProcessUpdate(result WsMarketUpdateResponse, pair currency.Pa
...
 		}
+		trades = append(trades, trade.Data{
+			Timestamp:    time.Unix(result.Trades[x].Timestamp/1000, 0),
+			CurrencyPair: pair,
+			AssetType:    asset.Spot,
+			Exchange:     g.Name,
+			Price:        result.Trades[x].Price,
+			Amount:       result.Trades[x].Quantity,
+			Side:         tSide,
+			TID:          strconv.FormatInt(result.Trades[x].EventID, 10),
```
gloriousCode, line 588: Please add a `!= "" && != "0"` check before setting the `TID` here. Not all trades seem to come with eventIDs and Gemini sends "0". The trade table has a unique check for TID + exchangename, so if multiple have "0", then it will be ignored/overwritten. For example, LINK-USD.
If the property `TID` is not set, then you can still proceed as there is another unique constrain that checks everything else but the ID for uniqueness:
```
 CONSTRAINT uniquetrade
        unique(exchange_name_id, base, quote, asset, price, amount, side, timestamp)
```

exchanges/gemini/gemini_websocket.go
```diff
@@ -261,15 +322,62 @@ func (g *Gemini) wsHandleData(respRaw []byte, curr currency.Pair) error {
...
+
+			tradeEvent := trade.Data{
+				Timestamp:    time.Unix(result.Timestamp/1000, 0),
+				CurrencyPair: pair,
+				AssetType:    asset.Spot,
+				Exchange:     g.Name,
+				Price:        result.Price,
+				Amount:       result.Quantity,
+				Side:         tSide,
+				TID:          strconv.FormatInt(result.EventID, 10),
```
gloriousCode, line 370: Same as previous nit regarding trades and TID check

exchanges/gemini/gemini_websocket.go
```diff
@@ -261,15 +322,62 @@ func (g *Gemini) wsHandleData(respRaw []byte, curr currency.Pair) error {
...
+			}
+			if g.IsSaveTradeDataEnabled() {
```
gloriousCode, line 372: This check may as well be at the start of this case, we don't do anything with trades if its not enabled

gloriousCode (changes requested):
Its nice to be able to use the Gemini websocket! 👀 I appreciate the changes, config migration works and its good to not need so many ws connections. Just a few things I've noticed:

exchanges/gemini/gemini_websocket.go
thrasher-: Nice catch, this was caused by an invalid tag! Should be fixed now

exchanges/gemini/gemini_websocket.go
```diff
@@ -261,15 +317,61 @@ func (g *Gemini) wsHandleData(respRaw []byte, curr currency.Pair) error {
...
+			var l2MarketData *wsL2MarketData
+			err := json.Unmarshal(respRaw, &l2MarketData)
+			if err != nil {
+				return err
+			}
+			return g.wsProcessUpdate(l2MarketData)
+		case "trade":
+			if !g.IsSaveTradeDataEnabled() {
+				return nil
+			}
```
shazbert, line 330: Suggestion: I would put this after json.Unmarshal just to alert us if there is a data type change from the exchange side that we don't notice.

shazbert (changes requested):
Pretty much tACK only one suggestion.

exchanges/gemini/gemini_websocket.go
thrasher-: This is a good point for validation but we already process websocket trades in wsProcessUpdate (they share the same struct) so this is already covered and thus don't have to waste CPU cycles for a second round of validation

MadCozBadd (approved):
no issues found! tACK!

shazbert (approved):
changes ACK!

PR #647, opened 2021-03-12
Exchanges: Add in exchange defined tolerance settings 

shazbert:
# PR Description

Add exchange defined validation level to order manager onto all outbound calls for minimum, maximum, increment, and notional checks for price and amount. 

Exchanges affected:
[x] - Binance

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [n/a] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool


exchanges/tolerance_test.go
```diff
@@ -0,0 +1,260 @@
...
+	if !errors.Is(err, ErrAmountExceedsStep) {
+		t.Fatalf("expected error %v but received %v", ErrAmountExceedsStep, err)
+	}
+	err = tt.Conforms(200000, .003)
+	if !errors.Is(err, nil) {
+		t.Fatalf("expected error %v but received %v", nil, err)
+	}
+}
+
+func TestConformToAmount(t *testing.T) {
```
MadCozBadd, line 228: Maybe we should add in a test case where amount is smaller than stepsize just to see if it returns 0

MadCozBadd (commented):
Cool PR! Just a few changes required

exchanges/tolerance.go
```diff
@@ -0,0 +1,287 @@
...
+// ConformToAmount (POC) conforms amount to its amount interval
+func (t *Tolerance) ConformToAmount(amount float64) float64 {
+	t.Lock()
+	defer t.Unlock()
+	if t.stepSizeAmount == 0 {
+		return 0
+	}
+	increase := 1 / t.stepSizeAmount
+	// math floor used because we don't want to ever increase the amount
+	return math.Floor(amount*increase) / increase
```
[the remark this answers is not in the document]
shazbert: I just left it for readability and keeping it in line with examples.

exchanges/binance/binance.go
```diff
@@ -940,3 +941,139 @@ func (b *Binance) MaintainWsAuthStreamKey() error {
...
+
+			limits = append(limits, exchange.MinMaxLevel{
+				Pair:             cp,
+				Asset:            assets[z],
+				MinPrice:         spot.Symbols[x].Filters[0].MinPrice,
+				MaxPrice:         spot.Symbols[x].Filters[0].MaxPrice,
+				StepPrice:        spot.Symbols[x].Filters[0].TickSize,
+				MultiplierUp:     spot.Symbols[x].Filters[1].MultiplierUp,
+				MultiplierDown:   spot.Symbols[x].Filters[1].MultiplierDown,
+				AveragePriceMins: spot.Symbols[x].Filters[1].StepSize,
```
MadCozBadd, line 988: 👀 StepSize -> AvgPriceMins

exchanges/binance/binance.go
```diff
@@ -940,3 +941,139 @@ func (b *Binance) MaintainWsAuthStreamKey() error {
...
+				MaxAmount:        spot.Symbols[x].Filters[2].MaxQty,
+				MinAmount:        spot.Symbols[x].Filters[2].MinQty,
+				StepAmount:       spot.Symbols[x].Filters[2].StepSize,
+				MinNotional:      spot.Symbols[x].Filters[3].MinNotional,
+				MaxIcebergParts:  spot.Symbols[x].Filters[4].MaxNumIcebergOrders,
```
MadCozBadd, line 993: Maybe I'm misunderstanding how this is being done but I can't seem to find MaxNumIcebergOrders in spot exchange info data

exchanges/binance/binance_test.go
```diff
@@ -2475,3 +2475,23 @@ func TestUFuturesHistoricalTrades(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestSetTolerances(t *testing.T) {
```
MadCozBadd, line 2479: t.Parallel please

exchanges/binance/binance.go
MadCozBadd: Maybe that is supposed to be spot.Symbols[x].Filters[4].Limit?

MadCozBadd (commented):
Just a quick review from me

exchanges/binance/binance.go
shazbert: Good spotting!

MadCozBadd (approved):
tACK!!!!!!

exchanges/binance/binance.go
```diff
@@ -940,3 +941,138 @@ func (b *Binance) MaintainWsAuthStreamKey() error {
...
+				StepPrice:        spot.Symbols[x].Filters[0].TickSize,
+				MultiplierUp:     spot.Symbols[x].Filters[1].MultiplierUp,
+				MultiplierDown:   spot.Symbols[x].Filters[1].MultiplierDown,
+				AveragePriceMins: spot.Symbols[x].Filters[1].AvgPriceMins,
+				MaxAmount:        spot.Symbols[x].Filters[2].MaxQty,
+				MinAmount:        spot.Symbols[x].Filters[2].MinQty,
+				StepAmount:       spot.Symbols[x].Filters[2].StepSize,
+				MinNotional:      spot.Symbols[x].Filters[3].MinNotional,
+				MaxIcebergParts:  spot.Symbols[x].Filters[4].Limit,
+				MarketMinimumQty: spot.Symbols[x].Filters[5].MinQty,
```
MadCozBadd, line 994: Really a lame request but since everything else is min and max you could change this to MarketMinQty

exchanges/tolerance.go
```diff
@@ -0,0 +1,364 @@
...
+	marketMinQty     float64
+	marketMaxQty     float64
+	marketStepSize   float64
+	maxTotalOrders   int64
+	maxAlgoOrders    int64
+	sync.Mutex
+}
+
+// Conforms checks outbound parameters
+func (t *Tolerance) Conforms(price, amount float64, marketOrder bool) error {
```
gloriousCode, line 241: I think it would be more future-proof if instead of `marketOrder bool` the ordertype is just passed in

gloriousCode (changes requested):
This is some nice clean code Shazbert 🤙 🎉  I appreciate all the comments, errors and tests! I look forward to the day that I get an error for trying to place an order on Binance saying I'm exceeding the max order amount allowed

exchanges/tolerance.go
```diff
@@ -0,0 +1,293 @@
+package exchange
```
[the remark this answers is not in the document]
shazbert: It was more that I couldn't think of a name so it's the first thing I had in my head. Tolerance, like in a permissible limit or limits of variation in numbers, it's not an investopedia term 😆 . I see your point though. I will change.

exchanges/binance/binance.go
```diff
@@ -940,3 +941,139 @@ func (b *Binance) MaintainWsAuthStreamKey() error {
 		HTTPRecording: b.HTTPRecording,
 	})
 }
+
+// SetExchangeTolerances sets exchange tolerances for all assets
+func (b *Binance) SetExchangeTolerances() error {
+	var limits []exchange.MinMaxLevel
+
+	spot, err := b.GetExchangeInfo()
```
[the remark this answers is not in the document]
shazbert: I hope you counted them by hand. Will add that extra functionality.

engine/fake_exchange_test.go
```diff
@@ -240,3 +240,11 @@ func (h *FakePassingExchange) WithdrawFiatFunds(_ *withdraw.Request) (*withdraw.
...
 	return nil, nil
 }
+
+func (h *FakePassingExchange) GetTolerance(a asset.Item, cp currency.Pair) (*exchange.Tolerance, error) {
+	return nil, nil
+}
+
+func (h *FakePassingExchange) CheckTolerance(a asset.Item, cp currency.Pair, price, amount float64, marketOrder bool) error {
+	return nil
+}
```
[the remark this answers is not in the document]
shazbert: Yeah, I embedded the struct in the exchange base so this isn't needed. 🦅

exchanges/binance/binance.go
```diff
@@ -940,3 +941,138 @@ func (b *Binance) MaintainWsAuthStreamKey() error {
...
+				MarketMinQty:     spot.Symbols[x].Filters[5].MinQty,
+				MarketMaxQty:     spot.Symbols[x].Filters[5].MaxQty,
+				MarketStepSize:   spot.Symbols[x].Filters[5].StepSize,
+				MaxTotalOrders:   spot.Symbols[x].Filters[6].MaxNumOrders,
+				MaxAlgoOrders:    spot.Symbols[x].Filters[7].MaxNumAlgoOrders,
+			})
+		}
+	}
+
+	usdtFutures, err := b.UExchangeInfo()
```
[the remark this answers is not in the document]
shazbert: This isn't a shadow declaration.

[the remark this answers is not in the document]
gloriousCode: Stop gaslighting me 😆 line 949

exchanges/binance/binance.go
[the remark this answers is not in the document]
shazbert: In the future, I want to get this via sync manager that will fetch every hour (?). I will add them based on asset and not the individual currency, we pretty much have that in memory anyway, might as well set it. If we enable and disable individual pairs I don't want to retrigger a re-fetch every time.

exchanges/interfaces.go
```diff
@@ -87,4 +86,8 @@ type IBotExchange interface {
 	// pair,asset, url/proxy or subscription change
 	FlushWebsocketChannels() error
 	AuthenticateWebsocket() error
+	// Exchange order related execution limits
+	GetLimit(a asset.Item, cp currency.Pair) (*order.Limits, error)
```
thrasher-, line 90: I think GetLimit is too ambiguous as it could mean many different things. Would be nice to reference order execution limit in its name. `GetOrderExecutionLimits`.

exchanges/order/limits.go
```diff
@@ -0,0 +1,364 @@
...
+	"sync"
+
+	"github.com/thrasher-corp/gocryptotrader/currency"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+)
+
+var (
+	// ErrExchangeLimitNotLoaded defines if an exchange does not have minmax
+	// values
+	ErrExchangeLimitNotLoaded = errors.New("exchange Limits not loaded")
```
thrasher-, line 16: Limits -> limits to keep case sensitivity with other words in the error message

exchanges/order/limits.go
```diff
@@ -0,0 +1,364 @@
...
+	// amount Limit accepted by the exchange for a market order
+	ErrMarketAmountExceedsMin = errors.New("market order amount exceeds minimum Limit")
+	// ErrMarketAmountExceedsMax is when the amount is higher than the maximum
+	// amount Limit accepted by the exchange for a market order
+	ErrMarketAmountExceedsMax = errors.New("market order amount exceeds maximum Limit")
+	// ErrMarketAmountExceedsStep is when the amount is not divisible by its
+	// step for a market order
+	ErrMarketAmountExceedsStep = errors.New("market order amount exceeds step Limit")
+
+	errCannotValidateAsset         = errors.New("cannot check Limit asset not loaded")
```
thrasher-, line 46: Grammatical issue for this one and a few others below

```suggestion
	errCannotValidateAsset         = errors.New("cannot check limit, asset not loaded")
```

exchanges/order/limits.go
```diff
@@ -0,0 +1,364 @@
...
+	errInvalidPriceLevels          = errors.New("invalid price levels cannot load Limits")
+	errInvalidAmountLevels         = errors.New("invalid amount levels cannot load Limits")
+)
+
+// ExecutionLimits defines minimum and maximum values in relation to
+// order size, order pricing, total notional values, total maximum orders etc
+// for execution on an exchange.
+type ExecutionLimits struct {
+	m map[asset.Item]map[currency.Code]map[currency.Code]*Limits
+	sync.Mutex
```
thrasher-, line 62: I think a RWMutex would be better here for now up until a routine is added for updating them every X ticker.  Once these are loaded, we'll mainly be performing reads.

thrasher- (changes requested):
Very nice! Only a few nits.

exchanges/order/limits.go
```diff
@@ -0,0 +1,365 @@
...
+			t.maxAmount > t.marketMaxQty &&
+			amount > t.marketMaxQty {
+			return fmt.Errorf("%w max: %f suppplied %f",
+				ErrMarketAmountExceedsMax,
+				t.marketMaxQty,
+				amount)
+		}
+		if t.marketStepIncrementSize != 0 && t.stepIncrementSizeAmount != t.marketStepIncrementSize {
+			increase := 1 / t.marketStepIncrementSize
+			if math.Mod(amount*increase, t.marketStepIncrementSize*increase) != 0 {
```
MadCozBadd, line 334: We can remove the creation of increase and just divide by marketStepIncrementSize when needed. Here and elsewhere please

MadCozBadd (commented):
Great changes!!!! Just another small change will go through the rest tomorrow.

exchanges/order/limits.go
shazbert: That's done, if I do it in the conform function the rounding issues get worse, can you check that out when you have time?

exchanges/order/limits.go
```diff
@@ -0,0 +1,362 @@
...
+	if !ok {
+		return nil, errExchangeLimitBase
+	}
+
+	limit, ok := m2[cp.Quote]
+	if !ok {
+		return nil, errExchangeLimitQuote
+	}
+
+	return limit, nil
```
gloriousCode, line 182: While a lock helps with the read, it doesn't necessarily fix a potential data race because you're returning the pointer, which could make modifications while its being read elsewhere. Yes its not happening right now, its much more likely with your syncer plans you've mentioned. The limit struct is pretty basic, it may serve better to return a copy

engine/orders.go
```diff
@@ -351,6 +351,18 @@ func (o *orderManager) Submit(newOrder *order.Submit) (*orderSubmitResponse, err
...
+
+	// Checks for exchange min max limits for order amounts before order
+	// execution can occur
+	err := exch.CheckOrderExecutionLimits(newOrder.AssetType,
+		newOrder.Pair,
+		newOrder.Price,
+		newOrder.Amount,
+		newOrder.Type)
+	if err != nil {
+		return nil, err
```
gloriousCode, line 363: It would be good to decorate this error with the exchange name since this function does not.

exchanges/order/limits.go
```diff
@@ -0,0 +1,362 @@
...
+	marketMinQty            float64
+	marketMaxQty            float64
+	marketStepIncrementSize float64
+	maxTotalOrders          int64
+	maxAlgoOrders           int64
+	sync.RWMutex
+}
+
+// Conforms checks outbound parameters
+func (t *Limits) Conforms(price, amount float64, orderType Type) error {
```
gloriousCode, line 242: l for limit? MINUSCULE NIT

exchanges/order/limits.go
```diff
@@ -0,0 +1,362 @@
...
+	"fmt"
+	"math"
+	"sync"
+
+	"github.com/thrasher-corp/gocryptotrader/currency"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+)
+
+var (
+	// ErrExchangelimitNotLoaded defines if an exchange does not have minmax
```
gloriousCode, line 14: `ErrExchangeLimitNotLoaded`

exchanges/order/limits.go
MadCozBadd: Seems like the way you are doing it fixes it 🤷‍♂️ the float rounding issues result in 1.00099999999999988987 1.00100000000000011191 which don't end up matching. But the approach you take where you create increase works and the result matches the value 1.1 as a float holds

gloriousCode (changes requested):
Thanks for making those changes! 🎉 Here's a fresh batch of more requests

exchanges/binance/binance_test.go
```diff
@@ -2475,3 +2475,39 @@ func TestUFuturesHistoricalTrades(t *testing.T) {
...
```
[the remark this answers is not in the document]
shazbert: No, I forgot to put it there.

exchanges/binance/binance_wrapper.go
```diff
@@ -302,6 +302,19 @@ func (b *Binance) Run() {
 		}
 	}
 
+	a := b.GetAssetTypes()
+	for x := range a {
+		if err = b.CurrencyPairs.IsAssetEnabled(a[x]); err != nil {
```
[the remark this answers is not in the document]
shazbert: 🦅

exchanges/order/limits.go
shazbert: All fields are unexported; there's a lock on the limit and a separate lock in the execution limits manager so the underlying can be read and updated. I don't see the race issue. Can I get an example?

gloriousCode: Your reply has highlighted more issues to me. By just giving `Limits` the property of `sync.rwmutex`, it means it can be locked from anywhere. 
An easy example is to nil the limit. When you have the `limit` pointer struct, I can modify it with no regard for the lock, the mutex lock even being exported means anything can lock it and not unlock it at any time too.

I would actually argue for the removal of the function `GetOrderExecutionLimits` and just keep `CheckOrderExecutionLimits` since it doesn't allow for any of the things I mentioned.

shazbert: Ahh yes, I will unexport the mutex that's a good point,  Locking it from anywhere is not a good thing, but if you are nil-ing the limit means your just nilling the scoped pointer value in a calling routine? The underlying struct is never changed by this? 

I disagree with the removal as having a pointer to the underlying `Limits` struct is helpful for strategies that are specifically targetting a currency pair and it needs to conform to an amount before issuing an order `ConformToAmount` POC func for example. When we are doing a large number of orders and checking these values via `CheckOrderExecutionLimits` before submission occurs, I didn't want to add in the overhead of also a strategy locking down and inhibiting an outbound check as well.

MadCozBadd (commented):
image
Last changes from me! Just need to change things up with a market order. It shouldn't matter what price gets specified when doing a market order. (Price and Notional checks should be by passed if market order)

exchanges/order/limits.go
```diff
@@ -0,0 +1,376 @@
...
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+)
+
+var (
+	// ErrExchangeLimitNotLoaded defines if an exchange does not have minmax
+	// values
+	ErrExchangeLimitNotLoaded = errors.New("exchange limits not loaded")
+	// ErrPriceExceedsMin is when the price is lower than the minimum price
+	// limit accepted by the exchange
+	ErrPriceExceedsMin = errors.New("price exceeds minimum limit")
```
MadCozBadd, line 19: The errors that have minimum requirements should be worded slightly different. In the case of this error the price is below the minimum limit yet we claim it exceeds it

MadCozBadd (approved):
🥳 tACK

exchanges/order/limits.go
gloriousCode: You're right re nil. Learned that only the localised variable gets nulled. While you can actually still manipulate the base in other ways, Golang makes it difficult/immediately noticeable and so I'm not concerned about that anymore.

Given that the function is unused (except for some tests), I still don't think this function is needed, but since it is unused, I'll re-asses things when it comes into play

MadCozBadd (approved):
changes look good! re tACK!

MadCozBadd (approved):
Tested backtester during this round of testing! and ack new changes too!

exchanges/order/limits.go
thrasher-: Pedantic nit but +1 on this:

```
submitorder --exchange=binance --pair=ETH-BTC --asset=spot --side=SELL --type=LIMIT --amount=0.0000001 --price=5000
2021/03/23 10:47:46 rpc error: code = Unknown desc = order manager: exchange binance unable to place order: amount exceeds minimum limit min: 0.001000 supplied 0.000000 for spot ETHBTC
```
"amount below minimum limit" rather than "amount exceeds minimum limit"

thrasher- (changes requested):

gloriousCode (changes requested):
Things are looking good! I have some minor nits and paternal instincts to protect the backtester 😝

backtester/eventhandlers/exchange/exchange.go
```diff
@@ -90,8 +90,15 @@ func (e *Exchange) ExecuteOrder(o order.Event, data data.Handler, bot *engine.En
...
+	// Conforms the amount to the exchange order defined step amount reducing it
+	// when needed
+	finalAmount := cs.Limits.ConformToAmount(reducedAmount)
+	if finalAmount != reducedAmount {
+		f.AppendReason(fmt.Sprintf("Order size shrunk from %v to %v to remain within exchange step amount limits", reducedAmount, finalAmount))
+	}
+
 	var orderID string
-	orderID, err = e.placeOrder(adjustedPrice, reducedAmount, cs.UseRealOrders, f, bot)
+	orderID, err = e.placeOrder(adjustedPrice, finalAmount, cs.UseRealOrders, f, bot)
```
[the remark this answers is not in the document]
gloriousCode: Could you please add a check for if amount == 0 then set it the status to `common.CouldNotBuy`/`common.CouldNotSell`? That should handle things appropriately

backtester/backtest/backtest.go
```diff
@@ -280,6 +279,12 @@ func (bt *BackTest) setupExchangeSettings(cfg *config.Config) (exchange.Exchange
 			MaximumTotal: cfg.CurrencySettings[i].SellSide.MaximumTotal,
 		}
 		sellRule.Validate()
+
+		limits, err := exch.GetOrderExecutionLimits(a, pair)
+		if err != nil {
+			return resp, err
+		}
```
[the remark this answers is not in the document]
shazbert: Yeah, I actually agree with what you are saying, the majority of 'in sample' amounts are not going to execute on exchanges at all. I am wondering if I should still set it in exchange settings and have a check before an order occurs to reduce the amount. I think that might be better because when we switch over to out-of-sample testing we can immediately conform to exchange settings.

exchanges/binance/binance.go
```diff
@@ -940,3 +941,62 @@ func (b *Binance) MaintainWsAuthStreamKey() error {
...
+		var assets []asset.Item
+		for y := range spot.Symbols[x].Permissions {
+			switch spot.Symbols[x].Permissions[y] {
+			case "SPOT":
+				assets = append(assets, asset.Spot)
+			case "MARGIN":
+				assets = append(assets, asset.Margin)
+			case "LEVERAGED": // leveraged tokens not available for spot trading
+			default:
+				return nil, fmt.Errorf("unhandled asset type for tolerance loading %s",
```
gloriousCode, line 969: exchange limits to prevent confusion

exchanges/binance/binance_wrapper.go
```diff
@@ -302,6 +302,19 @@ func (b *Binance) Run() {
 		}
 	}
 
+	a := b.GetAssetTypes()
+	for x := range a {
+		if err = b.CurrencyPairs.IsAssetEnabled(a[x]); err == nil {
+			err = b.UpdateOrderExecutionLimits(a[x])
+			if err != nil {
+				log.Errorf(log.ExchangeSys,
+					"Could not set %s exchange tolerances: %v",
```
gloriousCode, line 311: exchange limits

exchanges/order/limits.go
```diff
@@ -0,0 +1,378 @@
...
+			m2[levels[x].Pair.Quote] = limit
+		}
+
+		if levels[x].MinPrice >= levels[x].MaxPrice {
+			return fmt.Errorf("%w for %s %s supplied min: %f max: %f",
+				errInvalidPriceLevels,
+				levels[x].Asset,
+				levels[x].Pair,
+				levels[x].MinPrice,
+				levels[x].MaxPrice)
```
gloriousCode, line 125: Is this a valid check when min/max price is 0?
> Any of the above variables can be set to 0, which disables that rule in the price filter. In order to pass the price filter, the following must be true for price/stopPrice of the enabled rules:

[here](https://binance-docs.github.io/apidocs/delivery/en/#filters)

It would seem that the rest of the checks are still valid if min max price are 0

gloriousCode (changes requested):
Just a few minor things left I think 🌯 🌯 🌯 Thanks for updating those other things!

exchanges/order/limits.go
shazbert: Yeah, good spotting, can just check if they are set around the wrong way.

gloriousCode (changes requested):
Thanks for those changes! I've found some things I'm not sure of

exchanges/order/limits.go
```diff
@@ -0,0 +1,378 @@
...
+	if l.maxAmount != 0 && amount > l.maxAmount {
+		return fmt.Errorf("%w min: %.8f supplied %.8f",
+			ErrAmountExceedsMax,
+			l.maxAmount,
+			amount)
+	}
+	if l.stepIncrementSizeAmount != 0 {
+		dAmount := decimal.NewFromFloat(amount)
+		dStep := decimal.NewFromFloat(l.stepIncrementSizeAmount)
+		if !dAmount.Mod(dStep).IsZero() {
```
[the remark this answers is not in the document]
shazbert: Yeah, this was a weird one, I'm pretty sure we can cancel out that because we check for if the quantity is below the minQty by the time it hits this check. If you minus the minQty from the equation it just reduces your floor it doesn't look like it affects the mod equation at all. Will run another check to make sure.

gloriousCode (changes requested):
Thanks for clarifying that point, I agree. Last nit (🤞) from me

exchanges/order/limits_test.go
```diff
@@ -0,0 +1,310 @@
...
+	err = tt.Conforms(200000, 10, Market)
+	if !errors.Is(err, ErrMarketAmountExceedsMax) {
+		t.Fatalf("expected error %v but received: %v", ErrMarketAmountExceedsMax, err)
+	}
+
+	tt.marketStepIncrementSize = 10
+	err = tt.Conforms(200000, 9.1, Market)
+	if !errors.Is(err, ErrMarketAmountExceedsStep) {
+		t.Fatalf("expected error %v but received: %v", ErrMarketAmountExceedsStep, err)
+	}
```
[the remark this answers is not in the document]
shazbert: I will actually add that in as it makes no difference, there might be an unknown-unknown I'm not considering. Will do!

gloriousCode:
One more request 😄 Please add new interface functions support to exchange_wrapper_issues and exchange_wrapper_coverage cmd applications

cmd/exchange_wrapper_issues/main.go
```diff
@@ -475,6 +475,20 @@ func testWrappers(e exchange.IBotExchange, base *exchange.Base, config *Config)
 				Response:   getHisotirCandlesExtendedResponse,
 				SentParams: jsonifyInterface([]interface{}{p, assetTypes[i], startTime, endTime, kline.OneDay}),
 			})
+
+			err = e.UpdateOrderExecutionLimits(asset.Spot)
```
gloriousCode, line 479: `assetTypes[i]`

cmd/exchange_wrapper_issues/main.go
```diff
@@ -475,6 +475,20 @@ func testWrappers(e exchange.IBotExchange, base *exchange.Base, config *Config)
...
+			msg = ""
+			if err != nil {
+				msg = err.Error()
+				responseContainer.ErrorCount++
+			}
+
+			responseContainer.EndpointResponses = append(responseContainer.EndpointResponses, EndpointResponse{
+				SentParams: jsonifyInterface([]interface{}{asset.Spot}),
```
gloriousCode, line 487: `assetTypes[i]`

gloriousCode (commented):
last last last last last last last last nits

gloriousCode (approved):
tACK! Thank you for addressing and discussing all the nits
# 🌧️🌧️🌧️🌧️🌧️🌧️🌧️ 🌞

MadCozBadd (approved):
image

backtester/config/examples/dca-api-candles-multiple-currencies.strat
```diff
@@ -32,7 +32,8 @@
    "max-slippage-percent": 0,
    "maker-fee-override": 0.001,
    "taker-fee-override": 0.002,
-   "maximum-holdings-ratio": 0
+   "maximum-holdings-ratio": 0,
+   "use-exchange-order-limits": false
```
thrasher-, line 36: Tested and working well! 

test

Two changes before merging this in would be:

- Outputting this setting to the backtester console and HTML report, as you've already done in the config builder.
- A warning for exchanges which support order limits, but have it disabled should warn the user that the results are prone to errors (order execution won't work when in production).

thrasher- (changes requested):

backtester/report/tpl.gohtml
```diff
@@ -254,6 +254,33 @@
 					</tbody>
 				</table>
 			</div>
+			<div class="view view-cascade bg-info">
```
thrasher-, line 257: I think this looks better being `bg-warning`

backtester/config/examples/dca-api-candles-multiple-currencies.strat
gloriousCode: Could the message become "may not work when in production"? A user can already set their own min/max buying rules and there may have been no issue at all even if they were turned on

shazbert: Yes!

gloriousCode (approved):
tACK!

thrasher- (approved):

MadCozBadd (approved):
new changes tACK!!!!!!!!!!!!!!!!!!!!!!

PR #648, opened 2021-03-17
Feature: Faster start & stop times

gloriousCode:
# PR Description
Anytime I'm testing with many exchanges and currencies enabled (eg `-enableallpairs -enableallexchanges`), GoCryptoTrader can be slow to start and difficult to stop. At times I have to cancel the process due to it being unable to properly shutdown when so much is going on. Investigating this lead me to find that mainly the portfolio manager updating config address balances was to blame. When all currencies are enabled, it will take awhile and will not listen to a shutdown until the process is completed.

This is made worse when rate limits are involved. If you get banned on Binance and the portfolio manager is still attempting to load balances, you cannot shutdown until you are unbanned AND have finished the portfolio manager. 

### Changes
- Updated exchange loading during startup to be more asynchronous as there is no need for exchange loading to be done one at a time
- Portfolio manager calls was the main culprit in slow shutdowns. Its updated to not allow simultaneous processing
- Engine subsystems now call their ticker timeout tasks as a go routine as there is no need to prevent a shutdown from occurring as the subsystem tasks do not cause a negative impact if not completed. If you are wary of this though, we can discuss on a case-by-case basis
  - There is no need for the `for` loop to wait for the task to be completed.
  - Long running portfolio manager task won't double up due to the addition of the `processing` atomic
- The `stopped` property was not really utilised or necessary, so its been removed 
- `atomic.AddInt32(&c.started, 1)` use is bad because calling something twice can lead to a value of 2, which none of the subsystems could actually handle. Updated to `!atomic.CompareAndSwapInt32(&c.started, 0, 1)`

## Type of change
- [?] Bug fix (non-breaking change which fixes an issue)
- [?] New feature (non-breaking change which adds functionality)

## How has this been tested
- [x] Starting and stopping the application
- [x] `go run . -enableallexchanges -enableallpairs` 
- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


engine/timekeeper.go
```diff
@@ -116,15 +108,20 @@ func (n *ntpManager) FetchNTPTime() time.Time {
...
-	NTPcurrentTimeDifference := NTPTime.Sub(currentTime)
+	diff := NTPTime.Sub(currentTime)
 	configNTPTime := *Bot.Config.NTPClient.AllowedDifference
-	configNTPNegativeTime := (*Bot.Config.NTPClient.AllowedNegativeDifference - (*Bot.Config.NTPClient.AllowedNegativeDifference * 2))
-	if NTPcurrentTimeDifference > configNTPTime || NTPcurrentTimeDifference < configNTPNegativeTime {
-		log.Warnf(log.TimeMgr, "NTP manager: Time out of sync (NTP): %v | (time.Now()): %v | (Difference): %v | (Allowed): +%v / %v\n", NTPTime, currentTime, NTPcurrentTimeDifference, configNTPTime, configNTPNegativeTime)
-		if n.inititalCheck {
-			n.inititalCheck = false
+	negDiff := *Bot.Config.NTPClient.AllowedNegativeDifference
+	configNTPNegativeTime := negDiff - (negDiff * 2)
```
MadCozBadd, line 115: instead of negDiff - (negDiff*2) you could just use -negDiff

MadCozBadd (approved):
Great PR! Can confirm that it reduces the exit time significantly! Just a thing to consider though is the LakeBTC issue we discussed. Not sure if it is to be handled in this PR so for the rest, its a tACK!

gloriousCode:
> 
> 
> Great PR! Can confirm that it reduces the exit time significantly! Just a thing to consider though is the LakeBTC issue we discussed. Not sure if it is to be handled in this PR so for the rest, its a tACK!

I've just added a fun new func [here](https://github.com/thrasher-corp/gocryptotrader/pull/648/commits/b3490b00911253027c8b648254aad21f7f2d4603#diff-7ec5c2935d290e7ab0c356b218a0edfad2615e695d2fa7cfe193b06af37f28c6R413) which will allow the exit of the application even if LakeBTC is down/a user wishes to close the application while an exchange is loading. It works, but I can understand if say @thrasher- doesn't want os signal interpretation to be at the `engine` level. I can expand `signaler` package to allow a `chan os.Signal` to be passed in to allow for [this](https://github.com/thrasher-corp/gocryptotrader/pull/648/commits/b3490b00911253027c8b648254aad21f7f2d4603#diff-7cf1b16a9bacb975383d3822122ba21edbd7154f72bd84cad97b5018e01bebc9R210) test to still exist

engine/comms_relayer.go
```diff
@@ -82,15 +81,13 @@ func (c *commsManager) PushEvent(evt base.Event) {
...
-		atomic.CompareAndSwapInt32(&c.stopped, 1, 0)
-		atomic.CompareAndSwapInt32(&c.started, 1, 0)
 		log.Debugln(log.CommunicationMgr, "Communications manager shutdown.")
 	}()
 
 	for {
 		select {
 		case msg := <-c.relayMsg:
-			c.comms.PushEvent(msg)
+			go c.comms.PushEvent(msg)
```
shazbert, line 90: This routine doesn't seem interact with a wait group, so this doesn't add much. If this is a standardisation just disregard it.

engine/exchange.go
```diff
@@ -375,17 +384,54 @@ func (bot *Engine) SetupExchanges() {
...
-			configs[x].Name,
-			common.IsEnabled(configs[x].API.AuthenticatedSupport),
-			common.IsEnabled(configs[x].Verbose),
-		)
+		wg.Add(1)
+		go func(iterator int) {
+			err := bot.LoadExchange(configs[iterator].Name, true, &wg)
+			if err != nil {
+				log.Errorf(log.ExchangeSys, "LoadExchange %s failed: %s\n", configs[iterator].Name, err)
+				wg.Done()
```
shazbert, line 392: can call `defer wg.Done()` in the go func

engine/exchange.go
```diff
@@ -375,17 +384,54 @@ func (bot *Engine) SetupExchanges() {
...
-		}
-		log.Debugf(log.ExchangeSys,
-			"%s: Exchange support: Enabled (Authenticated API support: %s - Verbose mode: %s).\n",
-			configs[x].Name,
-			common.IsEnabled(configs[x].API.AuthenticatedSupport),
-			common.IsEnabled(configs[x].Verbose),
-		)
+		wg.Add(1)
+		go func(iterator int) {
+			err := bot.LoadExchange(configs[iterator].Name, true, &wg)
```
shazbert, line 389: would make a config struct copy `cpy := configs[iterator]` and pass by value in the `go func( cfg confStruct) {....} (cpy)` because when your accessing configs via your iterator in different routines it becomes a race issue due to the reads.

engine/exchange.go
```diff
@@ -375,17 +384,54 @@ func (bot *Engine) SetupExchanges() {
...
+				common.IsEnabled(configs[iterator].Verbose),
+			)
+			wg.Done()
+		}(x)
+	}
+	interruptChan := make(chan os.Signal)
+	waitForCompletionOrInterrupt(&wg, interruptChan)
+}
+
+// waitForCompletionOrInterrupt When starting exchanges, "UpdateTradablePairs" is called which is a REST request
```
shazbert, line 408: When -> when

engine/subsystem/subsystem.go
```diff
@@ -1,13 +1,11 @@
...
-	// ErrSubSystemAlreadyStopped message to return when a subsystem is already stopped
-	ErrSubSystemAlreadyStopped = "already stopped"
-	// ErrSubSystemNotStarted message to return when subsystem not started
-	ErrSubSystemNotStarted = "not started"
+import "errors"
 
+const (
+	errSubSystemAlreadyStarted = "manager already started"
+	errSubSystemAlreadyStopped = "already stopped"
+	errSubSystemNotStarted     = "not started"
```
shazbert, line 8: probably pop these strings into the err vars instead

engine/timekeeper.go
```diff
@@ -20,50 +21,40 @@ var (
...
-
+		n.initialCheck = true
 		// Sometimes the NTP client can have transient issues due to UDP, try
 		// the default retry limits before giving up
 		for i := 0; i < NTPRetryLimit; i++ {
-			err = n.processTime()
+			err := n.processTime()
 			switch err {
 			case nil:
 				break
```
shazbert, line 53: This looks like it breaks the switch case might want to break the for loop?

engine/timekeeper.go
```diff
@@ -105,7 +92,12 @@ func (n *ntpManager) run() {
 		case <-n.shutdown:
 			return
 		case <-t.C:
-			n.processTime()
+			go func() {
```
shazbert, line 95: here as well but if standard 🤷

shazbert (changes requested):
This is a good change! 👍 So speedy! Much ctrl+C!
For some reason we are displaying multiple same warnings 🤷 :
image

engine/timekeeper.go
gloriousCode: haha sweet pickup of some old codes

engine/comms_relayer.go
gloriousCode: Nah you're right 🎉 🚗

shazbert (approved):
tACK! Thanks for making those changes, this is snappy 💃

MadCozBadd (approved):
Looks good to me!!!!!

thrasher-:
> > Great PR! Can confirm that it reduces the exit time significantly! Just a thing to consider though is the LakeBTC issue we discussed. Not sure if it is to be handled in this PR so for the rest, its a tACK!
> 
> I've just added a fun new func [here](https://github.com/thrasher-corp/gocryptotrader/pull/648/commits/b3490b00911253027c8b648254aad21f7f2d4603#diff-7ec5c2935d290e7ab0c356b218a0edfad2615e695d2fa7cfe193b06af37f28c6R413) which will allow the exit of the application even if LakeBTC is down/a user wishes to close the application while an exchange is loading. It works, but I can understand if say @thrasher- doesn't want os signal interpretation to be at the `engine` level. I can expand `signaler` package to allow a `chan os.Signal` to be passed in to allow for [this](https://github.com/thrasher-corp/gocryptotrader/pull/648/commits/b3490b00911253027c8b648254aad21f7f2d4603#diff-7cf1b16a9bacb975383d3822122ba21edbd7154f72bd84cad97b5018e01bebc9R210) test to still exist

Yeah I'm not a fan of os signals being handled by engine, it will create conflicts with other packages/calling applications which require signal processing.

gloriousCode:
> Yeah I'm not a fan of os signals being handled by engine, it will create conflicts with other packages/calling applications which require signal processing.

Fair enough! I've removed it. Due to the current way we load exchanges, there is no way interrupting `engine.NewFromSettings(&settings, flagSet)` (where exchanges are loaded). So this PR still does address other slow start stop issues, it won't be able to cancel in the event an exchange is down and the request has been made. I'll consider alternatives in a future PR

shazbert (approved):
Still snappy 💃 ! changes ACK

MadCozBadd (approved):
LakeBTC seems to have figured out that you weren't going to stop the bot for them so instead they fixed their issues 🤷‍♂️. New Changes tACK!!!

thrasher- (approved):
tACK, thanks @gloriousCode !

PR #652, opened 2021-03-25
rpcserver/exchanges: Add additional param checks plus other minor bugfixes/improvements

ghost:
# PR Description

A test for backtester keeps creating files every time its run. This PR is supposed to delete the files right after they are created.

Also while testing other PRs I found it difficult to navigate through my silly mistakes based on the errors provided: 
image

This PR also tries to make the errors more verbose:
image

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [ ] Test X

## Checklist


engine/orders_types.go
```diff
@@ -9,7 +9,7 @@ import (
 )
 
 var (
-	errOrderCannotBeEmpty = errors.New("order cannot be empty")
+	errOrderCannotBeEmpty = errors.New("orderID cannot be empty")
```
shazbert, line 12: errOrderCannotBeEmpty -> errOrderIDCannotBeEmpty ?

engine/rpcserver.go
```diff
@@ -3077,3 +3044,36 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
 
 	return resp, nil
 }
+
+func (s *RPCServer) checkParams(exch string, a asset.Item, p currency.Pair) (exchange.IBotExchange, error) {
```
shazbert, line 3048: I would have this function take in an IbotExchange instead of returning one.  And not make this a method of RPCServer.

shazbert (changes requested):
Nice, just two things I noticed.

backtester/report/report_test.go
```diff
@@ -290,7 +292,21 @@ func TestGenerateReport(t *testing.T) {
...
+	}
+	var nickName string
+	if d.Config.Nickname != "" {
+		nickName = d.Config.Nickname + "-"
+	}
+	fileName := fmt.Sprintf(
+		"%v%v-%v.html",
+		nickName,
+		d.Statistics.StrategyName,
+		timeNow.Format("2006-01-02-15-04-05"))
```
gloriousCode, line 308: This can be done without needing to change the function signature to have a `time.Time`. `d` has a property `OutputPath` which we can use to set to a temporary directory. Then we just delete everything inside that directory
```
tempDir, err := ioutil.TempDir("", "")
	if err != nil {
		t.Fatalf("Problem creating temp dir at %s: %s\n", tempDir, err)
	}
	defer os.RemoveAll(tempDir)
d.OutputPath = tempDir
```

engine/rpcserver.go
```diff
@@ -3072,3 +3044,39 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
...
+	e := s.GetExchangeByName(exch)
+	if e == nil {
+		return nil, fmt.Errorf("exchange %s is not supported", exch)
```
gloriousCode, line 3051: `fmt.Erroff("%w: %s", errExchangeNotLoaded, exch)`

engine/rpcserver.go
```diff
@@ -3072,3 +3044,39 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
...
+	}
+	if !e.IsEnabled() {
+		return nil, fmt.Errorf("exchange %s is not enabled", exch)
```
gloriousCode, line 3054: Please make all of these errors actual errors and then wrap them. eg
`errExchangeNotEnabled = errors.New("exchange is not enabled")`
`fmt.Errorf("%w %s", errExchangeNotEnabled, exch)`

This allows you to do better error checks in tests or inspect errors at runtime

engine/rpcserver.go
gloriousCode: I agree. With a name `checkParams`, it should only be checking, not creating new params to return

engine/rpcserver.go
```diff
@@ -3072,3 +3044,39 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
...
+	}
+	if a.IsValid() {
```
gloriousCode, line 3056: What if its invalid?

gloriousCode (changes requested):
Thank you for making our errors a bit more friendly!

Also, as discussed, in this PR could you fix the 
```
RealOrders          bool   `json:"fake-orders"`
```
to `json:"real-orders"` that I foolishly coded, too?

engine/rpcserver.go
MadCozBadd: All the functions initially have a which is an assetType string. We then convert a to asset.Item using asset.New which returns an error. so that makes sure that the asset type is valid for the functions that require it to be. And then we input the asset.Item that we created into checkParams so in the case assetType is invalid that means that assetType must've been empty in the function that calls this checkParams

gloriousCode: Those checks occur outside this function and relies on each implementation being coded correctly. For a checking function, it would be better for the checking to occur within this to ensure that everything is set appropriately.
```
_, err := s.checkParams("Bitstamp", "", currency.Pair{})
```
returns no errors.

Also, please add test coverage

gloriousCode: I think this nit is overly pedantic and I don't think its worth worrying about the invalid scenario so much as its concerned with theoretical future bad implementations.

engine/rpcserver.go
```diff
@@ -1874,14 +1905,29 @@ func (s *RPCServer) GetAuditEvent(_ context.Context, r *gctrpc.GetAuditEventRequ
 
 // GetHistoricCandles returns historical candles for a given exchange
 func (s *RPCServer) GetHistoricCandles(_ context.Context, r *gctrpc.GetHistoricCandlesRequest) (*gctrpc.GetHistoricCandlesResponse, error) {
-	if r.Exchange == "" {
-		return nil, errExchangeNameUnset
+	if r.Start >= r.End {
```
gloriousCode, line 1908: `r.Start` and `r.End` are `string` type.  I don't think this should be the basis of comparing times.

engine/rpcserver.go
```diff
@@ -3072,3 +3072,39 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
...
+		b := e.GetBase()
+		if b == nil {
+			return errExchangeBaseNotFound
+		}
+		err := b.CurrencyPairs.IsAssetEnabled(a)
+		if err != nil {
+			return fmt.Errorf("%v %w", a, errAssetTypeDisabled)
+		}
+	}
+	if !p.IsEmpty() {
```
gloriousCode, line 3093: Please invert this if and return if it is empty

exchanges/order/limits.go
```diff
@@ -94,26 +94,26 @@ func (e *ExecutionLimits) LoadLimits(levels []MinMaxLevel) error {
 	e.mtx.Lock()
 	defer e.mtx.Unlock()
 	if e.m == nil {
-		e.m = make(map[asset.Item]map[currency.Code]map[currency.Code]*Limits)
+		e.m = make(map[asset.Item]map[*currency.Item]map[*currency.Item]*Limits)
```
gloriousCode, line 97: How come you made this change?

gloriousCode (changes requested):
Thanks for making those changes! Just a few things

exchanges/order/limits.go
MadCozBadd: So the CLI could accept lowercase and uppercase currency pairs

engine/rpcserver.go
```diff
@@ -1905,7 +1905,7 @@ func (s *RPCServer) GetAuditEvent(_ context.Context, r *gctrpc.GetAuditEventRequ
 
 // GetHistoricCandles returns historical candles for a given exchange
 func (s *RPCServer) GetHistoricCandles(_ context.Context, r *gctrpc.GetHistoricCandlesRequest) (*gctrpc.GetHistoricCandlesResponse, error) {
-	if r.Start >= r.End {
+	if r.Start == r.End {
```
gloriousCode, line 1908: You are still comparing strings here and its not a good way of checking if the time is invalid. Please use the  `UTCStartTime` and `UTCEndTime` to validate the times with .Equal and .After

engine/rpcserver.go
```diff
@@ -50,15 +51,18 @@ import (
 )
 
 var (
+	errExchangeNotEnabled   = errors.New("exchange is not enabled")
```
gloriousCode, line 54: This is unused, looks like youre using `errExchangeDisabled` instead, so this can be removed

engine/rpcserver.go
```diff
@@ -50,15 +51,18 @@ import (
...
 	errInvalidArguments     = errors.New("invalid arguments received")
 	errExchangeNameUnset    = errors.New("exchange name unset")
 	errCurrencyPairUnset    = errors.New("currency pair unset")
-	errStartEndTimesUnset   = errors.New("invalid start and end times")
+	errInvalidTimes         = errors.New("invalid start and end times")
+	errAssetTypeDisabled    = errors.New("asset type is disabled")
 	errAssetTypeUnset       = errors.New("asset type unset")
 	errDispatchSystem       = errors.New("dispatch system offline")
 	errCurrencyNotEnabled   = errors.New("currency not enabled")
+	errCurrencyPairInvalid  = errors.New("errCurrencyPairInvalid")
```
gloriousCode, line 65: Please explain the error

exchanges/bithumb/bithumb_test.go
```diff
@@ -232,11 +232,17 @@ func TestUpdateTicker(t *testing.T) {
...
 	_, err = b.UpdateTicker(cp, asset.Spot)
 	if err != nil {
 		t.Fatal(err)
 	}
+
+	cp = currency.NewPair(currency.DASH, currency.KRW)
+	_, err = b.UpdateTicker(cp, asset.Spot)
+	if err == nil {
+		t.Error("expecting DASH to return an error")
+	}
```
gloriousCode, line 245: I don't think we need this in the event they add it back. If you're wanting to test for errors, perhaps consider `DOGE ALICE`

gloriousCode (changes requested):

exchanges/bithumb/bithumb_test.go
MadCozBadd: good point!

MadCozBadd: In that case I've chosen to remove the error check since they could add any pair 🤷‍♂️

gloriousCode (approved):

engine/rpcserver.go
```diff
@@ -335,6 +338,13 @@ func (s *RPCServer) GetTicker(_ context.Context, r *gctrpc.GetTickerRequest) (*g
 		return nil, err
 	}
 
+	e := s.GetExchangeByName(r.Exchange)
+	checkParams(r.Exchange, e, a, currency.Pair{
```
shazbert, line 342: check error

engine/rpcserver.go
```diff
@@ -1874,26 +1904,41 @@ func (s *RPCServer) GetAuditEvent(_ context.Context, r *gctrpc.GetAuditEventRequ
...
-	var UTCEndTime time.Time
-	UTCEndTime, err = time.Parse(common.SimpleTimeFormat, r.End)
+
+	exch := s.GetExchangeByName(r.Exchange)
+	err = checkParams(r.Exchange, exch, a, pair)
 	if err != nil {
 		return nil, err
 	}
+
+	var klineItem kline.Item
```
shazbert, line 1941: this can be moved to ln 1950

engine/rpcserver.go
```diff
@@ -3072,3 +3071,40 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
 
 	return resp, nil
 }
+
+func checkParams(exchName string, e exchange.IBotExchange, a asset.Item, p currency.Pair) error {
```
shazbert, line 3075: remove exchName, rename e -> exch and use e.GetName()

engine/rpcserver.go
```diff
@@ -3072,3 +3071,40 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
...
+		}
+	}
+	if p.IsEmpty() {
+		return nil
+	}
+	enabledPairs, err := e.GetEnabledPairs(a)
+	if err != nil {
+		return err
+	}
+	if !enabledPairs.Contains(p, true) {
```
shazbert, line 3099:
```go
if enabledPairs.Contains(p, true) {
return nil
}
```

engine/rpcserver_test.go
```diff
@@ -811,12 +824,13 @@ func TestUpdateAccountInfo(t *testing.T) {
...
 	updateResponse, err := s.UpdateAccountInfo(context.Background(), &gctrpc.GetAccountInfoRequest{Exchange: fakePassExchange, AssetType: asset.Futures.String()})
-	if err != nil {
-		t.Fatalf("TestGetAccountInfo: Failed to update account info: %s", err)
+	if !errors.Is(err, errAssetTypeDisabled) {
+		t.Errorf("expected %v, received %v", errAssetTypeUnset, err)
 	}
-
-	if getResponse.Accounts[0].Currencies[0].TotalValue == updateResponse.Accounts[0].Currencies[0].TotalValue {
-		t.Fatalf("TestGetAccountInfo: Unexpected value of the 'TotalValue'")
+	if err == nil {
```
shazbert, line 830: This is strange. because we are going to be erroring above on ln 826, it should be its own case.

shazbert (changes requested):
Looking good; just some minor nits.

engine/rpcserver.go
MadCozBadd: panics if e is empty

shazbert (approved):
tACK, thanks for making those changes!

thrasher- (approved):
tACK!

image

engine/rpcserver_test.go
```diff
@@ -810,12 +823,18 @@ func TestUpdateAccountInfo(t *testing.T) {
 		t.Fatalf("TestGetAccountInfo: Failed to get account info: %s", err)
 	}
 
-	updateResponse, err := s.UpdateAccountInfo(context.Background(), &gctrpc.GetAccountInfoRequest{Exchange: fakePassExchange, AssetType: asset.Futures.String()})
-	if err != nil {
-		t.Fatalf("TestGetAccountInfo: Failed to update account info: %s", err)
+	_, err = s.UpdateAccountInfo(context.Background(), &gctrpc.GetAccountInfoRequest{Exchange: fakePassExchange, AssetType: asset.Futures.String()})
+	if !errors.Is(err, errAssetTypeDisabled) {
+		t.Errorf("expected %v, received %v", errAssetTypeUnset, err)
```
gloriousCode, line 828: these do not match

engine/rpcserver.go
```diff
@@ -3072,3 +3074,40 @@ func (s *RPCServer) GetRecentTrades(_ context.Context, r *gctrpc.GetSavedTradesR
...
+	if e == nil {
+		return fmt.Errorf("%s %w", exchName, errExchangeNotLoaded)
+	}
+	if !e.IsEnabled() {
+		return fmt.Errorf("%s %w", exchName, errExchangeDisabled)
+	}
+	if a.IsValid() {
+		b := e.GetBase()
+		if b == nil {
+			return errExchangeBaseNotFound
```
gloriousCode, line 3088: To be in line with your other more helpful errors, how about `return fmt.Errorf("%s %w", exchName, errExchangeBaseNotFound)` ?

gloriousCode (changes requested):
Two minor nits and then all G!

gloriousCode (approved):

PR #687, opened 2021-05-12
add irix

al3rez:


thrasher- (changes requested):
Thanks for your PR @al3rez however the updated imports are too intrusive. Could you please refactor your code to be in line with our other exchange implementations. Furthermore, there doesn't seem to be any changes apart from updated imports?

PR #699, opened 2021-06-22
GCT: general updates across codebase

shazbert:
# PR Description

Allows GCT to be used more as a library in a strategy context. Opened as a draft as I will need to integrate with an external strategy and iron out any bugs. 

Engine:
* LoadExchange method remove boolean param
* Add method WaitForInitialSync so external calls can wait for initial sync via sync manager

SyncManager: 
* AddWaitForSync method
* Add service wait group for that external call

Orderbook:
* export linked list nodes and add in Unsafe type for external strategy inspection
* add a method to Unsafe type to lock down diametrically opposed books

Ticker:
* Reduce map lookup contention
* Change lock to standard mutex as we are writing constantly which are being impeded by multiple reads
* Add findLast function to package to find a valid last ticker disregarding exchange

Logger:
* Export SubLogger to tie in an external package to the GCT library logger system. 

General:
* remove true verbose assignments in tests

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


thrasher- (changes requested):
Only one basic nit

thrasher-:
NOTE: https://github.com/thrasher-corp/gocryptotrader/runs/3157065555#step:7:215 these tests are currently failing too

exchanges/orderbook/unsafe.go
```diff
@@ -0,0 +1,62 @@
...
+func (src *Unsafe) LockWith(dst sync.Locker) {
+	src.m.Lock()
+	dst.Lock()
+}
+
+// UnlockWith unlocks both books for the context of cross orderbook inspection
+func (src *Unsafe) UnlockWith(dst sync.Locker) {
+	dst.Unlock() // Unlock in reverse order
+	src.m.Unlock()
+}
```
gloriousCode, line 50: Just a thought on this dual lock system. You may wish to somehow protect dual locks from individual locks. As in, if you `LockWith`, you can also just call `Unlock` on each individually. You may be able to do it atomically to state that there is a dualLock in place, so cannot just call Unlock.
Since its for external use I'm less concerned, but I thought I'd just say the availability of both does present an opportunity for panics.

gloriousCode (changes requested):
🎉

engine/engine.go
```diff
@@ -705,8 +705,9 @@ func (bot *Engine) GetExchanges() []exchange.IBotExchange {
 	return bot.ExchangeManager.GetExchanges()
 }
 
-// LoadExchange loads an exchange by name
-func (bot *Engine) LoadExchange(name string, useWG bool, wg *sync.WaitGroup) error {
+// LoadExchange loads an exchange by name. Optional wait group can be added for
+// external synchronization.
+func (bot *Engine) LoadExchange(name string, optional *sync.WaitGroup) error {
```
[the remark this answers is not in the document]
shazbert: How about `optionalWaitGroupThisCanBeNil`? How descriptive do you want it?

engine/sync_manager.go
```diff
@@ -98,6 +99,8 @@ func (m *syncManager) Start() error {
 	if !atomic.CompareAndSwapInt32(&m.started, 0, 1) {
 		return ErrSubSystemAlreadyStarted
 	}
+	m.initSyncWG.Add(1)
+	m.inService.Done()
```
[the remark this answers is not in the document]
shazbert: That's not ideal. 🦅

exchanges/ticker/ticker.go
```diff
@@ -51,108 +57,137 @@ func SubscribeToExchangeTickers(exchange string) (dispatch.Pipe, error) {
...
 	s.Unlock()
-	return nil
+	return s.mux.Publish(ids, p)
 }
 
 // SetItemID retrieves and sets dispatch mux publish IDs
-func (s *Service) SetItemID(p *Price, fmtName string) error {
-	if p == nil {
-		return errors.New(errTickerPriceIsNil)
-	}
```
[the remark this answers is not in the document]
shazbert: I moved the check to process ticker function. If it was nil before this call then it would have been panic at the disco.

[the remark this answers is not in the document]
gloriousCode: Since its an exported function its still important. Though you could also just internalise the function since its only used there and resolve this.

exchanges/orderbook/unsafe.go
shazbert: Individual locks should not be an issue as you call lock first and then unlock. The problem occurs when you are locking multiple books for inspection. e.g. BTC-USD exch1, BTC-USDT exch2  in two routines we could check this set at the same time locking book1 and then book2, this is fine but if one routine locks it out of order say, locks book2 and then book1 then at the same time another routine locks book1 and then book2 that is where the dead lock occurs.

exchanges/ticker/ticker.go
```diff
@@ -51,108 +57,137 @@ func SubscribeToExchangeTickers(exchange string) (dispatch.Pipe, error) {
...
+			a)
 	}
 
-	return &service.Tickers[exchange][p.Base.Item][p.Quote.Item][tickerType].Price, nil
+	cpy := t.Price // Don't let external functions have access to underlying
+	return &cpy, nil
+}
+
+// FindLast searches for a currency pair and returns the first available
+func FindLast(p currency.Pair, a asset.Item) (float64, error) {
```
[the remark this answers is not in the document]
shazbert: Yeah things may need to be updated. It's not for speed , think of it as a relational standpoint if I was making calculation on an exchange but didn't want to convert using foreign exchange system e.g BTC-USD but I wanted it to be quoted it in LTC to get a ratio for an approximation. I don't need to know which exchange actually is synchronizing that BTC-LTC pair.

I will opt to remove until a solid use case can be implemented? @gloriousCode

engine/engine.go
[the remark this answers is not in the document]
gloriousCode: Shazbert, the comment already states it is optional. We don't name variables solely based on whether they are required or optional. Your suggestion at least describes that it is a waitgroup, but reverting its name would be an improvement too as it is quite standard in our codebase.

exchanges/orderbook/unsafe.go
gloriousCode: Is there value in preventing a lock if either one is currently locked? By having an atomic switch first, whether dual or single lock, at least when you attempt to lock either of those pairs again, you can see the atomic status and return with the suggestion that the opportunity has been lost as a currency is already under a lock. Like, if you attempt to lock a lock, its blocking until it unlocks, but the atomic will allow you to check before you attempt.

shazbert: Sure but it might not needed on the GCT library you can easily add that externally for whatever you need:

```go
// Depth type to expand specific functionality for the orderbook.Unsafe type
type Depth struct {
	orderbook.Unsafe
        isLocked uint32 
}

func (d *Depth) IsLockedBeforeLock() error {
      // atomic fun 
}
```

shazbert: In that case above you would need to share a semaphore between relationships and those relationships would need to be tracked at a higher level as well. Structuring locking sequence might more efficient than introducing atomic checks.

gloriousCode: Yeah fair enough. Can be left for any theoretical external packages to manage 🎉

engine/engine.go
```diff
@@ -886,25 +887,30 @@ func (bot *Engine) SetupExchanges() error {
...
 	wg.Wait()
 	if len(bot.ExchangeManager.GetExchanges()) == 0 {
 		return ErrNoExchangesLoaded
 	}
 	return nil
 }
+
+// WaitForInitialSync allows for the initial sync wait for external strategy
+// utilisation
+func (bot *Engine) WaitForInitialSync() error {
```
gloriousCode, line 914: Maybe `WaitForInitialCurrencySync` as we're on the Engine level?

gloriousCode (changes requested):
Thanks for making those changes! Just one minor thing

engine/engine.go
shazbert: Good point will change!

gloriousCode (approved):
tACK! Thanks for making those changes! 🎉

thrasher-:
Note: Test failures are not related to this PR

thrasher- (approved):

PR #712, opened 2021-07-13
gctrpc/ordermanager/binance: Add new getManagedOrders command and various improvements

MarkDzulko:
# PR Description

**Add**: GCTRPC command `getManagedOrders` to retrieve the orders from the OrderManager.
**Add**: `GetOrdersFiltered` for OrderManager
**Add**: `MatchFilter` method for Order.Detail
**Fix**: Added missing `ClientOrderId` in `gctrpc.GetOrder`
**Fix**: Added missing `ClientOrderID` to `gctrpc.submitOrder`
**Fix**: Binance `newClientOrderId` was spelled incorrectly and was not applied
**Changed**: `grpc.CancelOrder` nopw calls the orderManager instead of directly canceling through the exchange
**Changed**: Binance: Canceling an order keeps the old ClientOrderId instead of overwriting it (**needs discussion**)
**Changed**: Simplified `FilterOrdersByCurrencies`
**Changed**: Used RWMutex for order store (maybe in a different PR there should be a mutex per exchange?)

## Type of change

- [x] New feature (potentially breaking change which adds functionality)

## How has this been tested

Created test cases for new GetManagedOrders and MatchFilter functionality.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


engine/order_manager.go
```diff
@@ -388,6 +388,34 @@ func (m *OrderManager) GetOrdersSnapshot(s order.Status) ([]order.Detail, time.T
 	return os, latestUpdate
 }
 
+// GetOrdersFiltered returns a snapshot of all orders in the order store.
+// Filtering is applied based on the order.Filter unless entries are empty
+func (m *OrderManager) GetOrdersFiltered(f *order.Filter) ([]order.Detail, time.Time, error) {
```
shazbert, line 393: for now, it might be an idea to drop the time.Time return, it doesn't seem like it's being used.

engine/order_manager.go
```diff
@@ -388,6 +388,34 @@ func (m *OrderManager) GetOrdersSnapshot(s order.Status) ([]order.Detail, time.T
...
+	if m == nil {
+		return nil, time.Time{}, fmt.Errorf("order manager %w", ErrNilSubsystem)
+	}
+
+	if m == nil || atomic.LoadInt32(&m.started) == 0 {
+		return nil, time.Time{}, fmt.Errorf("order manager %w", ErrSubSystemNotStarted)
+	}
+	var os []order.Detail
+	var latestUpdate time.Time
+	for _, v := range m.orderStore.Orders {
```
shazbert, line 403: out of scope: might be better fixing up (make a new map, don't ref any orders to the underlying map) and using m.orderStore.get() method. Let me know what you think.

engine/order_manager.go
```diff
@@ -388,6 +388,34 @@ func (m *OrderManager) GetOrdersSnapshot(s order.Status) ([]order.Detail, time.T
...
+		for i := range v {
+			if !v[i].MatchFilter(f) {
+				continue
+			}
+			if v[i].LastUpdated.After(latestUpdate) {
+				latestUpdate = v[i].LastUpdated
+			}
+
+			cpy := *v[i]
+			os = append(os, cpy)
```
shazbert, line 413: `os = append(os, *v[i])`

shazbert (changes requested):
Changes look really good! You have added in some really great code improvements and tests.

Out of scope: For `GCTCLI` we could consolidate all order management functionality into one e.g. pair_management.go but that can be done in the future.

shazbert:
Also regarding: 

> Changed: Binance: Canceling an order keeps the old ClientOrderId instead of overwriting it (needs discussion)

I am for not overwriting any initial data. Maybe we can add an extra field for `CancelledClientOrderID` just to handle it for now.

gloriousCode (commented):
I appreciate this and like having the ability to view order manager orders. The main thing I thought I'd give input on is the cancelledOrderId part. As its a draft, I didn't go in depth, but happy to provide more feedback if requested.

engine/rpcserver.go
```diff
@@ -1043,14 +1128,15 @@ func (s *RPCServer) SubmitOrder(_ context.Context, r *gctrpc.SubmitOrderRequest)
...
-		ClientID:  r.ClientId,
-		Exchange:  r.Exchange,
-		AssetType: a,
+		Pair:          p,
+		Side:          order.Side(r.Side),
+		Type:          order.Type(r.OrderType),
+		Amount:        r.Amount,
+		Price:         r.Price,
+		ClientID:      r.ClientId,
+		ClientOrderID: r.ClientId,
```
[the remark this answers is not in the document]
MarkDzulko: I am actually unclear when ClientID is used versus ClientOrderID. I noticed that GRPC is mixing this even further because it just uses ClientID then, even if ClientOrderID is previously used.

exchanges/binance/binance_websocket.go
```diff
@@ -240,6 +240,10 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 				if err != nil {
 					return err
 				}
+				clientOrderID := data.Data.ClientOrderID
+				if oStatus == order.Cancelled {
+					clientOrderID = data.Data.CancelledClientOrderID
```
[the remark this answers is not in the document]
MarkDzulko: From a user perspective and especially when interacting via GRPC, it would be nice to keep the ClientOrderID teh same even when the order is cancelled to allow for clear lookup. With Binance being different to most of the other exchanges, we could maybe think about a "NewCancellationId" or something?
At least for the typical user, they can go through the active and cancelled orders the same way then and don't have to be made aware that Binance is different.

engine/order_manager.go
MarkDzulko: I moved the filtering to the order store instead.

engine/order_manager.go
MarkDzulko: implemented

engine/order_manager.go
shazbert: Changes look awesome! Will mark this for review.

exchanges/binance/binance_websocket.go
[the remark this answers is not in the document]
thrasher-: I can see the simplicity in overwriting the clientOrderID to match whatever the latest is in the short term however I like the idea of an additional "NewCancellationID / CancelledClientOrderID" field for exchanges like Binance. Preserving all information helps with auditing and debugging. Can leave this as-is though and have it addressed in a separate PR.

engine/rpcserver.go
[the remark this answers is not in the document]
thrasher-: Yeah this looks like a duplicated field, we use ClientID throughout the framework for client submitted order ID's. The only other time we use ClientID's for exchanges are ones which require a client ID as part of their API key credentials but that isn't related to orders

shazbert (changes requested):
Thank you very much for addressing my issues. Executing and cancelling via front end looks good!
```json
gocryptotrader\cmd\gctcli> go run ./... getmanagedorders --exchange="binance" --asset="spot" --pair="btc-usdt"
{
 "orders": [
  {
   "exchange": "binance",
   "id": "6842269113",
   "client_order_id": "web_4a57ac4d98344e848a90659e02b56ec1",
   "base_currency": "BTC",
   "quote_currency": "USDT",
   "asset_type": "spot",
   "order_side": "SELL",
   "order_type": "LIMIT",
   "creation_time": 1626389336,
   "status": "NEW",
   "price": 70000,
   "amount": 0.000153,
   "open_volume": 0.000153
  }
 ]
}
gocryptotrader\cmd\gctcli> go run ./... getmanagedorders --exchange="binance" --asset="spot" --pair="btc-usdt"
{}
```
testing order submit via grpc doesn't seem to register, which might be out of scope, but if you can quickly take a look at whats happening that would be awesome:
```json
gocryptotrader\cmd\gctcli> go run ./... submitorder --exchange="binance" --pair="btc-usdt" --side="SELL" --type="LIMIT" --amount=0.000153 --price=70000 --client_id="smellysocks" --asset="spot"
{
 "order_placed": true,
 "order_id": "6842307731"
}
gocryptotrader\cmd\gctcli> go run ./... getmanagedorders --exchange="binance" --asset="spot" --pair="btc-usdt"
{}
```
Front end:
image

exchanges/order/orders.go
```diff
@@ -359,6 +359,58 @@ func (d *Detail) UpdateOrderFromModify(m *Modify) {
...
+	}
+	if f.Status != "" && f.Status != AnyStatus && d.Status != f.Status {
+		return false
+	}
+	if f.AssetType != "" && d.AssetType != f.AssetType {
+		return false
+	}
+	if !f.Pair.IsEmpty() && !d.Pair.Equal(f.Pair) {
+		return false
+	}
```
gloriousCode, line 400: I think these two should be moved up to below the `f.Exchange` check as this should filter the majority of orders very quickly compared to the other checks

engine/order_manager.go
```diff
@@ -736,3 +747,32 @@ func (s *store) add(det *order.Detail) error {
 
 	return nil
 }
+
+// getFilteredOrders returns a filtered copy of the orders
+func (s *store) getFilteredOrders(f *order.Filter) []order.Detail {
+	s.m.RLock()
```
gloriousCode, line 753: Could you please add a test for this function?

exchanges/order/orders.go
```diff
@@ -359,6 +359,58 @@ func (d *Detail) UpdateOrderFromModify(m *Modify) {
...
+	if f.InternalOrderID != "" && d.InternalOrderID != f.InternalOrderID {
+		return false
+	}
+	if f.ID != "" && d.ID != f.ID {
+		return false
+	}
+	if f.ClientOrderID != "" && d.ClientOrderID != f.ClientOrderID {
+		return false
+	}
+	if f.AccountID != "" && d.AccountID != f.AccountID {
```
gloriousCode, line 377: Please add a test scenario for this one too

engine/order_manager.go
```diff
@@ -379,15 +379,26 @@ func (m *OrderManager) GetOrdersSnapshot(s order.Status) ([]order.Detail, time.T
...
+			os = append(os, *v[i])
 		}
 	}
 
 	return os, latestUpdate
 }
 
+// GetOrdersFiltered returns a snapshot of all orders in the order store.
+// Filtering is applied based on the order.Filter unless entries are empty
+func (m *OrderManager) GetOrdersFiltered(f *order.Filter) ([]order.Detail, error) {
```
gloriousCode, line 391: Could you please add a test for this function?

gloriousCode (changes requested):
Thank you for making those changes! Its looking good. Just some minor comments

MarkDzulko:
@shazbert I tried to reproduce your issue, but `SubmitOrder` is working fine for me and the order is correctly recorded in the managed orders after using your example code.
However I fixed that the `clientOrderId` was not recorded in the Order Manager after Submit.

shazbert (approved):
tested approved 👍

MadCozBadd (commented):
Thanks a lot for this comprehensive PR. Due to such good code quality and understanding I couldn't find any flaws. Just have one minor change to ask of you. Thanks again!

engine/order_manager_test.go
```diff
@@ -559,3 +559,44 @@ func TestProcessOrders(t *testing.T) {
...
 	m.processOrders()
 }
+
+func TestGetOrdersFiltered(t *testing.T) {
+	m := OrdersSetup(t)
+	_, err := m.GetOrdersFiltered(nil)
+	if err == nil {
+		t.Error("Expected error from nil filter")
+	}
+	// main functionality testing is done in Test_getFilteredOrders
```
gloriousCode, line 569: I understand that the testing of `getFilteredOrders` is done in more detail with the test `Test_getFilteredOrders`, but it would be good to test each type of scenario for `GetOrdersFiltered` as this test is only checking for one type of failure. We do not care for the result of `getFilteredOrders` here, but knowing via testing that you can actually call the function successfully would improve this test a lot

exchanges/binance/binance_wrapper.go
```diff
@@ -1191,17 +1193,18 @@ func (b *Binance) GetActiveOrders(req *order.GetOrdersRequest) ([]order.Detail,
...
+					Date:          resp[x].Time,
+					Exchange:      b.Name,
+					ID:            strconv.FormatInt(resp[x].OrderID, 10),
+					ClientOrderID: resp[x].ClientOrderID,
+					Side:          orderSide,
+					Type:          orderType,
+					Price:         resp[x].Price,
+					Status:        order.Status(resp[x].Status),
+					Pair:          req.Pairs[i],
+					AssetType:     asset.Spot,
```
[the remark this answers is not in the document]
MarkDzulko: Sure, wasn't my code, but I am happy to change it.

shazbert (approved):
Changes ACK! 🦅 awsome work!

MadCozBadd (approved):
changes ACK! 🎉

thrasher- (approved):
ACK, thank you for the PR @MarkDzulko !

gloriousCode (approved):
tACK! Thank you for making those changes @MarkDzulko 🎉

PR #724, opened 2021-07-27
gctrpc/order manager: Add ModifyOrder endpoint

ydm:
# PR Description

This PR introduces a new gRPC call that modifies an existing order.

This PR is not in a final stage, it needs attention from somebody with deeper knowledge of GCT.

My questions are:
1. How to manage the internal state of the modified order without introducing race conditions? In this case, after successful request to the exchange, what's the proper way of updating locally managed orders? What if a cancel request is issued concurrently? What if the order gets (partially) filled in the meantime?
2. Should I call `OrderManager.orderStore.commsManager.PushEvent()`?
3. Should I use `UpdateOrderFromModify`()? What if the `Modify` struct has unsupported-by-the-exchange fields populated that get copied over the original order `Detail`?

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


engine/order_manager.go
```diff
@@ -284,6 +284,49 @@ func (m *OrderManager) validate(newOrder *order.Submit) error {
...
+	id, err := exch.ModifyOrder(mod)
+	if err != nil {
+		return nil, err
+	}
+
+	// We now have to reflect this change in GCT's internal state of managed orders.
+
+	// TODO: Current implementation is here for illustrative purposes.  It introduces
+	// race conditions.
+	det, err := m.orderStore.getByExchangeAndID(mod.Exchange, mod.ID)
```
shazbert, line 311: We don't have an update procedure just yet that doesn't cause a race condition with order.Modify properties. 
You can also see in [websocketroutine_manager.go](https://github.com/thrasher-corp/gocryptotrader/blob/master/engine/websocketroutine_manager.go#L234) the POC implementation, that this is not ideal and this needs to change.

We can continue implementing features like this but a broad update on the order management system will need to be implemented and will be one of our priorities moving forward.

engine/order_manager.go
```diff
@@ -284,6 +284,49 @@ func (m *OrderManager) validate(newOrder *order.Submit) error {
...
+	if err != nil {
+		// TODO: How to handle that?  The order is updated on the
```
shazbert, line 313: Would need to fetch or build an order.Detail from order.Modify.

engine/order_manager.go
```diff
@@ -284,6 +284,49 @@ func (m *OrderManager) validate(newOrder *order.Submit) error {
...
+		// exchange, but not found locally.
+		return nil, err
+	}
+
+	// TODO: Is using det.UpdateOrderFromModify() better in this case, since
```
shazbert, line 318: Maybe for a quick implementation we could, instead of returning an order.Detail we can have that as a field and return pointer to something like order.Managed struct which includes a mutex which will have the methods to mutate changes.

shazbert (changes requested):
Just had a quick look. We can chat more via Slack.

engine/order_manager.go
ydm: Could you please elaborate a bit further on that? The `Modify struct` has just partial information at this point, I don't see how we can build a `Detail` reliably out of that.. Plus there *has to be* a managed Detail already, right?. I was thinking of panic()ing here as it's a case we should never hit. How bad would that be? :)

shazbert: You are correct, it can't reliable generate an order detail. 

> Plus, there has to be a managed Detail already, right?

There definitely should be but if we don't sync orders fast enough this might be a very unlikely scenario. I would even opt to 1) Add in a error log output and then retrieve the order from the exchange (albeit this is slow and adds http.Request friction) to then build out the order detail. The probability of this actually hitting and executing is low. 

> I was thinking of panic()ing here as it's a case we should never hit. How bad would that be? 

We try not to introduce panics in this codebase, I do see why you would put one there though. Its been a convention for a while to keep the program running and logging.

gloriousCode: In such a scenario, it may be best to still create the order in the order manager when it receives an order.Modify for an order it does not yet have knowledge about. - it having knowledge that an order exists is still very important information, even if its just an orderID and its asset type. Then as Shazbert said, fetch order details afterwards separately in a go routine to update the sparse order

engine/order_manager.go
```diff
@@ -284,6 +284,49 @@ func (m *OrderManager) validate(newOrder *order.Submit) error {
...
+	if atomic.LoadInt32(&m.started) == 0 {
+		return nil, fmt.Errorf("order manager %w", ErrSubSystemNotStarted)
+	}
+
+	exch := m.orderStore.exchangeManager.GetExchangeByName(mod.Exchange)
+	if exch == nil {
+		return nil, ErrExchangeNotFound
+	}
+
+	id, err := exch.ModifyOrder(mod)
```
gloriousCode, line 302: I think you should be updating the ModifyOrder function to return an `order.Modify` that is populated by supported fields. eg in ftx_wrapper, we know that the price and size is the only fields that can be modifed for an order. If the submitted Modify request is successful, we only pass back that amount and price. This will satisfy your issue in regards to unsupported data being populated. It needs to be at the exchange/exchange wrapper level to know what fields it can actually pass back.

Furthermore, you can validate requests and the exchange wrapper level. Again, knowing that FTX can only modify price and amount, if other fields are populated in the request, return an error. We don't want the situation where someone submits a modify request to make the asset type "Futures", the response is successful, but the asset type is unchanged because its unsupported.

gloriousCode (changes requested):
Hello. I've just done a basic lookover. I appreciate the attempt, though as you've realised, there is more work to be done to have a robust order manager for GCT.

> How to manage the internal state of the modified order without introducing race conditions? In this case, after successful request to the exchange, what's the proper way of updating locally managed orders? What if a cancel request is issued concurrently? What if the order gets (partially) filled in the meantime?

Many exchanges provide live order updates via a websocket subscription which we then try to funnel through the websocket datahandler as an `order.Modify` which does help with updating orders, but it is lacking. Binance for example does not track the updates properly right now. 

Plus as you said, we still have raciness.  The changes required here are likely large. I would think a timestamp, event-based order builder would be on the cards, so even if we receive an update out of step, we would rebuild the order based on all events received sorted by timestamp and then we would always have the latest picture of the order. 

I do not expect this issue to be fixed or added in this PR

> OrderManager.orderStore.commsManager.PushEvent()?

Yes, the commsManager is useful for pushing updates out to many communication endpoints, if a user has them enabled it then they wish to receive order updates.

> Should I use UpdateOrderFromModify()? What if the Modify struct has unsupported-by-the-exchange fields populated that get copied over the original order Detail?

Yes, I think you should. With the comment I made about returning an `order.Modify` for modify order requests, you can have the confidence that only supported changes have been modified.

ydm:
@gloriousCode @shazbert , can you please take a look at latest changes. I think I addressed all of your comments and suggestions.

cmd/gctcli/commands.go
```diff
@@ -2236,6 +2236,39 @@ var cancelAllOrdersCommand = &cli.Command{
...
+			Usage: "exchange this order is submitted to",
+		},
+		&cli.StringFlag{
+			Name:  "asset",
+			Usage: "required asset type",
+		},
+		&cli.StringFlag{
+			Name:  "pair",
+			Usage: "required trading pair",
+		},
```
shazbert, line 2256: You might be able to add:
```go
Required: true,
```

engine/order_manager.go
```diff
@@ -284,6 +284,45 @@ func (m *OrderManager) validate(newOrder *order.Submit) error {
 	return nil
 }
 
+// Modify depends on the order.Modify.ID and order.Modify.Exchange fields to uniquely
+// identify an order to modify.
+func (m *OrderManager) Modify(mod *order.Modify) (*order.ModifyResponse, error) {
```
shazbert, line 289: Could we please add a test for this?

exchanges/bitfinex/bitfinex_wrapper.go
```diff
@@ -647,27 +647,36 @@ func (b *Bitfinex) SubmitOrder(o *order.Submit) (order.SubmitResponse, error) {
...
-		})
-		return action.ID, err
-	}
-	return "", common.ErrNotYetImplemented
+		}
+		if action.Side == order.Sell && action.Amount > 0 {
+			request.Amount *= -1
+		}
+		err = b.WsModifyOrder(&request)
+		return order.Modify{
```
shazbert, line 669: Can we dereference action pointer instead, then assign as needed?

exchanges/bithumb/bithumb.go
```diff
@@ -329,6 +329,8 @@ func (b *Bithumb) ModifyTrade(orderID, orderCurrency, transactionType string, un
 
 	params := url.Values{}
 	params.Set("order_currency", strings.ToUpper(orderCurrency))
+	// XXX: Why "payment_currency" starts with an uppercase letter?  On
```
shazbert, line 332: I can confirm this is an invalid parameter, please change to lower case. Awesome spotting!

shazbert (changes requested):
Changes look great, thank you! Would it be possible to also change [websocketroutine_manager.go](https://github.com/thrasher-corp/gocryptotrader/blob/master/engine/websocketroutine_manager.go#L234) or is that out of scope? Just some minor nits.

engine/order_manager.go
```diff
@@ -653,6 +692,24 @@ func (s *store) updateExisting(od *order.Detail) error {
 	return ErrOrderNotFound
 }
 
+func (s *store) modifyExisting(id string, mod *order.Modify) error {
```
gloriousCode, line 695: Also please add a test for this function

exchanges/bithumb/bithumb_wrapper.go
```diff
@@ -438,22 +438,31 @@ func (b *Bithumb) SubmitOrder(s *order.Submit) (order.SubmitResponse, error) {
...
 // market conversion
-func (b *Bithumb) ModifyOrder(action *order.Modify) (string, error) {
+func (b *Bithumb) ModifyOrder(action *order.Modify) (order.Modify, error) {
 	if err := action.Validate(); err != nil {
-		return "", err
+		return order.Modify{}, err
 	}
 
-	order, err := b.ModifyTrade(action.ID,
+	o, err := b.ModifyTrade(action.ID,
```
gloriousCode, line 446: You'll need to do a check here for the `side` parameter. Since your RPCServer function only modifies price and amount, it means the side will be unset if performed via GRPC. You could otherwise populate some fields inside `(m *OrderManager) Modify` when you retrieve the existing order from the order manager to ensure fields such as this are still populated.

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -557,10 +557,20 @@ func (p *Poloniex) ModifyOrder(action *order.Modify) (string, error) {
 		action.PostOnly,
 		action.ImmediateOrCancel)
```
gloriousCode, line 558: Similar to the bithumb request, you may want to have checks on these as you  may unintentionally modify an order

gloriousCode (changes requested):
Nice work actioning all the things I mentioned yesterday! I have only a few concerns on top of Shazbert's from looking over things

cmd/gctcli/commands.go
ydm: Truth is user can modify just one of them. If required==true, both will have to be supplied, which would be an artificial requirement.

exchanges/bitfinex/bitfinex_wrapper.go
ydm: I'm not sure I understand you correctly. You want me to assign the change on the input Modify struct?

shazbert: What I mean was when you return we could just pass on a copy, so we could reduce some lines (up to you if you want to do it, resolve if not) e.g:
```go
return *action, nil 
// or 
re := *action
re.Price = 123
re.ID = 1234
return re, nil
```

shazbert (changes requested):
Thanks for making those changes, just some slight nits then it looks good to go. 👍

engine/order_manager.go
```diff
@@ -284,6 +284,61 @@ func (m *OrderManager) validate(newOrder *order.Submit) error {
...
+	if det == nil || err != nil {
+		return nil, fmt.Errorf("order does not exist: %w", err)
+	}
+
+	// Populate additional Modify fields as some of them are required by various
+	// exchange implementations.
+	mod.Pair = det.Pair                           // Used by Bithumb.
+	mod.Side = det.Side                           // Used by Bithumb.
+	mod.PostOnly = det.PostOnly                   // Used by Poloniex.
+	mod.ImmediateOrCancel = det.ImmediateOrCancel // Used by Poloniex.
```
gloriousCode, line 308: You may wish to set the price and amount here if the passed in `price` or `amount` is zero. The reason being is that they are optional parameters, and our exchange API implementations are not checking > 0 before sending to the API and one may modify an order unintentionally due to default values being `0`. You could place checks across all the exchange API level implementations of `ModifyOrder()`, but a check here also fixes the issue.

engine/order_manager_test.go
```diff
@@ -555,6 +597,58 @@ func TestSubmit(t *testing.T) {
...
+
+	mod := order.Modify{
+		// These fields identify the order.
+		Exchange:  testExchange,
+		AssetType: asset.Spot,
+		Pair:      pair,
+		ID:        "nonexistent_order_id",
+
+		// These fields modify the order.
+		Price:  11,
```
gloriousCode, line 627: With the `!= 0` check in mind, can you add a scenario where price is 0 and another where amount is 0?

engine/order_manager_test.go
```diff
@@ -333,6 +339,42 @@ func TestExists(t *testing.T) {
 	}
 }
 
+func Teststore_modifyOrder(t *testing.T) {
```
gloriousCode, line 342: Please update test name to make it a valid test

gloriousCode (changes requested):
Thanks for making those changes! I picked up on something though.

engine/order_manager.go
```diff
@@ -326,7 +344,17 @@ func (m *OrderManager) Modify(mod *order.Modify) (*order.ModifyResponse, error)
 	// Notify observers.
 	if err != nil {
```
shazbert, line 345: I think we need to invert this comparison.

shazbert (changes requested):
Just one thing from new changes and everything else looks good. Thanks heaps. 🚀

engine/order_manager.go
ydm: Thanks for noticing! :)

shazbert (approved):
Some things I have noticed while testing this is pretty much an ACK from me:

```
❯ go run ./... modifyorder --exchange="ftx" --asset="spot" --pair="btc-usd" --order_id="68281343789"
2021/08/04 08:15:42 rpc error: code = Unknown desc = FTX unsuccessful HTTP status code: 400 raw response: {"success":false,"error":"Must modify either price or size"}
exit status 1
```

I think a check for zero values before sending an HTTP request on the wrapper side would be good. But this is somewhat out of scope. 

```json
❯ go run ./... getmanagedorders --exchange="ftx" --asset="spot" --pair="btc-usd"
{
 "orders": [
  {
   "exchange": "ftx",
   "id": "68284655030",
   "client_order_id": "multipart-926377b3-149e-46a8-87cf-eaa4781ee5cc",
   "base_currency": "BTC",
   "quote_currency": "USD",
   "asset_type": "spot",
   "order_side": "BUY",
   "order_type": "LIMIT",
   "creation_time": 1628027840,
   "status": "NEW",
   "price": 1250,
   "amount": 0.001,
   "open_volume": 0.001
  },
  {
   "exchange": "ftx",
   "id": "68281343789",
   "client_order_id": "multipart-926377b3-149e-46a8-87cf-eaa4781ee5cc",
   "base_currency": "BTC",
   "quote_currency": "USD",
   "asset_type": "spot",
   "order_side": "BUY",
   "order_type": "LIMIT",
   "status": "PARTIALLY_CANCELLED",
   "amount": 0.001,
   "open_volume": 0.001
  },
  {
   "exchange": "ftx",
   "id": "68283692574",
   "base_currency": "BTC",
   "quote_currency": "USD",
   "asset_type": "spot",
   "order_side": "BUY",
   "order_type": "LIMIT",
   "status": "PARTIALLY_CANCELLED",
   "amount": 0.001,
   "open_volume": 0.001
  },
  {
   "exchange": "ftx",
   "id": "68284451196",
   "base_currency": "BTC",
   "quote_currency": "USD",
   "asset_type": "spot",
   "order_side": "BUY",
   "order_type": "LIMIT",
   "status": "PARTIALLY_CANCELLED",
   "amount": 0.002,
   "open_volume": 0.002
  }
 ]
}
```

It seems like we are partially cancelling the modified order and creating a new one (for FTX) price does not seem to be reflected and client_order_id is dropped from some managed  orders as well. This is also out of scope.

cmd/gctcli/commands.go
```diff
@@ -2269,6 +2302,95 @@ func cancelAllOrders(c *cli.Context) error {
...
+	// Parse optional flags.
+	var price float64
+	var amount float64
+
+	if c.IsSet("price") {
+		price = c.Float64("price")
+	}
+	if c.IsSet("amount") {
+		amount = c.Float64("amount")
+	}
```
gloriousCode, line 2364: I agree with Shazbert. Please do a check where if price and amount is 0, then don't proceed

gloriousCode (changes requested):
Thanks for making those changes! Its a little change and then I'm good 🎉

cmd/gctcli/commands.go
ydm: What if the client wants to modify just one of them? This is already supported in OrderManager.Modify.

ydm:
> Some things I have noticed while testing this is pretty much an ACK from me:
> 
> ```
> ❯ go run ./... modifyorder --exchange="ftx" --asset="spot" --pair="btc-usd" --order_id="68281343789"
> 2021/08/04 08:15:42 rpc error: code = Unknown desc = FTX unsuccessful HTTP status code: 400 raw response: {"success":false,"error":"Must modify either price or size"}
> exit status 1
> ```
> 
> I think a check for zero values before sending an HTTP request on the wrapper side would be good. But this is somewhat out of scope.
> 
> ```json
> ❯ go run ./... getmanagedorders --exchange="ftx" --asset="spot" --pair="btc-usd"
> {
>  "orders": [
>   {
>    "exchange": "ftx",
>    "id": "68284655030",
>    "client_order_id": "multipart-926377b3-149e-46a8-87cf-eaa4781ee5cc",
>    "base_currency": "BTC",
>    "quote_currency": "USD",
>    "asset_type": "spot",
>    "order_side": "BUY",
>    "order_type": "LIMIT",
>    "creation_time": 1628027840,
>    "status": "NEW",
>    "price": 1250,
>    "amount": 0.001,
>    "open_volume": 0.001
>   },
>   {
>    "exchange": "ftx",
>    "id": "68281343789",
>    "client_order_id": "multipart-926377b3-149e-46a8-87cf-eaa4781ee5cc",
>    "base_currency": "BTC",
>    "quote_currency": "USD",
>    "asset_type": "spot",
>    "order_side": "BUY",
>    "order_type": "LIMIT",
>    "status": "PARTIALLY_CANCELLED",
>    "amount": 0.001,
>    "open_volume": 0.001
>   },
>   {
>    "exchange": "ftx",
>    "id": "68283692574",
>    "base_currency": "BTC",
>    "quote_currency": "USD",
>    "asset_type": "spot",
>    "order_side": "BUY",
>    "order_type": "LIMIT",
>    "status": "PARTIALLY_CANCELLED",
>    "amount": 0.001,
>    "open_volume": 0.001
>   },
>   {
>    "exchange": "ftx",
>    "id": "68284451196",
>    "base_currency": "BTC",
>    "quote_currency": "USD",
>    "asset_type": "spot",
>    "order_side": "BUY",
>    "order_type": "LIMIT",
>    "status": "PARTIALLY_CANCELLED",
>    "amount": 0.002,
>    "open_volume": 0.002
>   }
>  ]
> }
> ```
> 
> It seems like we are partially cancelling the modified order and creating a new one (for FTX) price does not seem to be reflected and client_order_id is dropped from some managed orders as well. This is also out of scope.

That is another issue which is fixed by @MarkDzulko and I guess his change is already merged into the master.

cmd/gctcli/commands.go
gloriousCode: I mean an 
`If amount == 0 && price == 0 { return err } ` 
This is just a simple client side check to ensure that the request is valid and saving on a call to the application when invalid

shazbert (approved):
Changes ACK - I merged master and those changes work perfectly after multiple modifications. Thank you!

```json
❯ go run ./... modifyorder --exchange="ftx" --asset="spot" --pair="btc-usd" --order_id="68499009227"
2021/08/05 08:40:09 either --price or --amount should be present
exit status 1
                                                                                                                                                                                 [08:40]
❯ go run ./... modifyorder --exchange="ftx" --asset="spot" --pair="btc-usd" --order_id="68499009227" --price="1500"
{
 "modified_order_id": "68499363259"
}

❯ go run ./... modifyorder --exchange="ftx" --asset="spot" --pair="btc-usd" --order_id="68499363259" --price="1250" --amount=0.002
{
 "modified_order_id": "68499522519"
}
❯ go run ./... getmanagedorders --exchange="ftx" --asset="spot" --pair="btc-usd"
{
 "orders": [
  {
   "exchange": "ftx",
   "id": "68499522519",
   "client_order_id": "multipart-df4f0241-ce96-4052-9557-55e60425bf02",
   "base_currency": "BTC",
   "quote_currency": "USD",
   "asset_type": "spot",
   "order_side": "BUY",
   "order_type": "LIMIT",
   "status": "NEW",
   "price": 1250,
   "amount": 0.002,
   "open_volume": 0.002
  }
 ]
}

```

cmd/gctcli/commands.go
```diff
@@ -2236,6 +2236,39 @@ var cancelAllOrdersCommand = &cli.Command{
 	},
 }
 
+var modifyOrderCommand = &cli.Command{
+	Name:      "modifyorder",
+	Usage:     "modify price and/or amount of a previously submitted order",
+	ArgsUsage: "<exchange> <asset> <pair> <order_id>",
```
thrasher-, line 2242: missing price and amount args

thrasher- (changes requested):
Only have one minor nit and then this looks good to go. Awesome work @ydm !

cmd/gctcli/commands.go
ydm: Yes, because these two are conditional and are expected to be passed as flags, not as positional arguments.

thrasher- (approved):

PR #732, opened 2021-07-30
Multi instance exchange

lrascao:
# PR Description

This allows us to have multiple instances of the same exchange that
might be configured differently. (eg. two instances of Binance with
different credentials). This functionality is also a preparation and
paves the way for multiple instances of in-house custom exchanges
that might be used for load/correctness/consistency/validation testing.

## Type of change

- [X] This change requires a documentation update

## How has this been tested

- [X] go test ./... -race
- [X] golangci-lint run
- [X] Test TestInstanceOfExchange

## Checklist

- [?] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool


shazbert (changes requested):
Thanks for opening this PR! So there are a few things that I think should be included for this to be merged. 
* Requester and rate limiting shared between all instances because multiple instances on Binance were achieving  `418 I'm a teapot` errors. 
* The syncer acts on all instances when it should collectively only sync the currency pairs for the base exchange so that we don't do multiple fetching of each pair. 
* On some exchanges there can only be one websocket connection per instance or it will be forcefully closed. 

An idea might be to abstract an instance type out of our exchange base type with:
*  Independent credentials 
*  Requester/ratelimer pointer from 'instanceOf' base exchange
*  Instance name
*  Enabled pairs list which it can only operate on. 

So there might need to be some change with our exchange interface types as well. 
And maybe we need to segregate that in the config to keep instances nice and small and relative. 

Let me know what you think. 🚀

lrascao:
closing in favour of #756 and #757

PR #743, opened 2021-08-10
Bugfix: Syncer & websocket circular dependency

gloriousCode:
# PR Description

One of our wonderful Slack members: Przemysław Skiba, highlighted an issue:
> 1. SyncManager is created and needs WebsocketRoutineManager as dependency. In my case, during SyncManager startup, WebsocketRoutineManager is nil.
> 2. SyncManager tries set WebsocketDataReceiver to nil object and transforms WS to "connecting" state
> 3. During WebsocketRoutineManager startup, it checks WS state. State is "connecting", so WebsocketRoutineManager returns from start function and not set WebsocketDataReceiver.
> 
> The result is no updates from WS routine

We have the `CurrencyPairSyncer` depending on the `WebsocketRoutineManager` and the `WebsocketRoutineManager` wants the  `CurrencyPairSyncer`. The `CurrencyPairSyncer` was given the ability to control the websocket connections, but its not necessary. It is not the responsibility of the `CurrencyPairSyncer` to manage websocket connections. Without this ability, the `CurrencyPairSyncer` is fine as the  `WebsocketRoutineManager` maintains connections and if there were an issue, the `CurrencyPairSyncer` reverts to REST.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Running the application to ensure that the currency pair syncer utilises websocket updates without controlling the websocket connection.

## Checklist

### DIFF
--- a/engine/sync_manager.go
+++ b/engine/sync_manager.go
@@ -41,7 +41,7 @@ var (
 )
 
 // setupSyncManager starts a new CurrencyPairSyncer
-func setupSyncManager(c *Config, exchangeManager iExchangeManager, remoteConfig *config.RemoteControlConfig) (*syncManager, error) {
+func setupSyncManager(c *Config, exchangeManager iExchangeManager, remoteConfig *config.RemoteControlConfig, websocketRoutineManagerEnabled bool) (*syncManager, error) {
 	if !c.SyncOrderbook && !c.SyncTicker && !c.SyncTrades {
 		return nil, errNoSyncItemsEnabled
 	}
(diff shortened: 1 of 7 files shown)


shazbert (approved):
enabled/disabled sync subsystem, changes look good! tested approved 🦅

engine/sync_manager.go
```diff
@@ -110,44 +110,15 @@ func (m *syncManager) Start() error {
...
-						exchangeName,
-						err)
-					usingREST = true
-				} else {
-					usingWebsocket = true
-				}
-			} else {
-				usingWebsocket = true
-			}
+			usingWebsocket = true
```
thrasher-, line 121: I'd like to see an additional check here to see if the websocket manager is enabled, otherwise we're wasting CPU cycles until this times out per asset which is quite costly if we have hundreds of items

thrasher- (changes requested):

thrasher- (approved):
thanks for making that change!

shazbert (approved):
Changes tested ACK!

PR #749, opened 2021-08-13
engine/exchange manager: Add support for custom exchanges

lrascao:
This PR allows applications that import GCT as a lib to declare custom exchanges that do not require full GCT integration.

- [X] go test ./... -race
- [X] golangci-lint run

## Checklist

- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -105,6 +105,8 @@ func NewFromSettings(settings *Settings, flagSet map[string]bool) (*Engine, erro
 		return nil, fmt.Errorf("failed to create script manager. Err: %s", err)
 	}
 
+	b.ExchangeManager = SetupExchangeManager()
+
 	validateSettings(&b, settings, flagSet)
 
 	return &b, nil
@@ -363,7 +365,6 @@ func (bot *Engine) Start() error {
 		bot.Config.PurgeExchangeAPICredentials()
 	}
 
-	bot.ExchangeManager = SetupExchangeManager()
 	gctlog.Debugln(gctlog.Global, "Setting up exchanges..")
 	err = bot.SetupExchanges()
 	if err != nil {
--- a/engine/exchange_manager.go
+++ b/engine/exchange_manager.go
@@ -45,10 +45,17 @@ var (
 	ErrExchangeFailedToLoad  = errors.New("exchange failed to load")
 )
 
+// CustomExchangeBuilder interface allows external applications to create
+// custom/unsupported exchanges that satisfy the IBotExchange interface.
+type CustomExchangeBuilder interface {
+	NewExchangeByName(name string) (exchange.IBotExchange, error)
+}
+
 // ExchangeManager manages what exchanges are loaded
 type ExchangeManager struct {
 	m         sync.Mutex
 	exchanges map[string]exchange.IBotExchange
+	Builder   CustomExchangeBuilder
 }
 
 // SetupExchangeManager creates a new exchange manager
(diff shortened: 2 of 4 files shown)


engine/engine.go
```diff
@@ -105,6 +105,8 @@ func NewFromSettings(settings *Settings, flagSet map[string]bool) (*Engine, erro
 		return nil, fmt.Errorf("failed to create script manager. Err: %s", err)
 	}
 
+	b.ExchangeManager = SetupExchangeManager()
+
```
gloriousCode, line 109: This is already set in `Start()`. As it seems like you're using a different workflow to setup/load unsupported exchanges, the `ExchangeManager` is already exported and you should set it up from that context.

engine/exchange_manager.go
```diff
@@ -45,10 +45,15 @@ var (
 	ErrExchangeFailedToLoad  = errors.New("exchange failed to load")
 )
 
+type CustomExchangeBuilder interface {
```
gloriousCode, line 48: Please add a comment

engine/exchange_manager.go
```diff
@@ -183,6 +188,9 @@ func (m *ExchangeManager) NewExchangeByName(name string) (exchange.IBotExchange,
 	case "zb":
 		exch = new(zb.ZB)
 	default:
+		if m.Builder != nil {
+			return m.Builder.NewExchangeByName(nameLower)
+		}
```
gloriousCode, line 193: Please add test coverage

gloriousCode (changes requested):
Hi lrascao, just have some comments on the implementation

shazbert (commented):
No issues from me (other than whats already been said), thanks for the PR.

engine/engine.go
lrascao: That's the tricky bit, on my service (using this patch) i go:
```
engine.Bot, err = engine.NewFromSettings(&settings, flagSet)
...
engine.Bot.ExchangeManager.Builder = ExchangeBuilder{}
...
if err = engine.Bot.Start(); err != nil {
```
Main idea here is to have a functional `ExchangeManager` that can be tweaked around before `Start`.
Currently, (pre-patch)  `Start` creates and overwrites any existing `ExchangeManager` and copying the entire contents of `NewFromSettings` and `Start` seems a bit wasteful.

shazbert (approved):
Changes ACK. Also, this allows for the loading of the same base exchange multiple times which can make things quite unstable. Might be prudent doing a reflection check when adding the exchange but I will leave that up to you.

engine/engine.go
gloriousCode: When reviewing yesterday it looked like you had left the original `bot.ExchangeManager = SetupExchangeManager()` in `Start()` and I didn't want the two of them there. So long as there is only one, I'm fine with it 👍

gloriousCode (approved):
tACK! Thanks for making those changes lrascao. As Shazbert pointed out, this does allow for loading multiple instances of the same exchange, do you still need https://github.com/thrasher-corp/gocryptotrader/pull/732 ?

thrasher- (approved):
tACK! Thanks @lrascao

lrascao:
> tACK! Thanks for making those changes lrascao. As Shazbert pointed out, this does allow for loading multiple instances of the same exchange, do you still need #732 ?

👍🏻  @gloriousCode , you're right that the entire content of #732 is no longer needed, i'm splitting it into #756 and #757 to further facilitate the custom exchange use-case

PR #755, opened 2021-08-18
docs: Establish a standard for trade/order-update websocket events

ydm:
This PR adds a description of what's expected from exchange implementations when handling trades and generally order update events.


docs/WS_ORDER_EVENTS.md
```diff
@@ -0,0 +1,31 @@
+# Websocket trade events
+
+GoCryptoTrader unifies trades and order update events by composing an
+order.Detail object.  This is the full list of order.Detail fields
+that exchange implementations should populate on streamed
+trade/order-update events.  As exchanges provide different APIs, not
+all fields are mandatory.
+
+```
+| order.Detail field   | Description                                                       | Condition                                               | Presence  |
```
shazbert, line 10: One pedantic nit: We could drop the code block and have it as a table instead so it can resize a bit better.

Like here: 
image

Instead of: 
image

shazbert (changes requested):
This is an awesome update. Documentation improvements are very welcome.

docs/WS_ORDER_EVENTS.md
```diff
@@ -0,0 +1,31 @@
...
+|----------------------+-------------------------------------------------------------------+---------------------------------------------------------+-----------|
+| Price                | Original price assigned to order                                  | Depends on order type (e.g. limit orders have prices)   | Mandatory |
+| Amount               | Original quantity assigned to order                               |                                                         | Mandatory |
+| AverageExecutedPrice | Average price of what's traded thus far                           | Order is filled, partially filled or partially canceled | Mandatory |
```
gloriousCode, line 14: I think this one should be optional. It is not included in all exchange API responses and potentially can be calculated on our side if we gather the trades

gloriousCode (changes requested):
Thanks for writing some documentation! Just one comment

docs/WS_ORDER_EVENTS.md
ydm: Sounds painful to exclude it. :(

shazbert (approved):
Thanks for doing those changes, looking good. Ack

gloriousCode (approved):
Thanks for making that change! Thank you for adding to our documentation! 🎉

thrasher- (approved):
Appreciate the documentation improvement! Thanks @ydm!

PR #757, opened 2021-08-18
engine: allow exchange name to be set

lrascao:
Add a `SetName` to base exchange that allows custom exchange builders to set dynamic names.

- [X] go test ./... -race
- [X] golangci-lint run

- [] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -340,6 +340,11 @@ func (b *Base) GetName() string {
 	return b.Name
 }
 
+// SetName is a method that overrides the name of the exchange base
+func (b *Base) SetName(name string) {
+	b.Name = name
+}
+
 // GetEnabledFeatures returns the exchanges enabled features
 func (b *Base) GetEnabledFeatures() FeaturesEnabled {
 	return b.Features.Enabled
--- a/exchanges/exchange_test.go
+++ b/exchanges/exchange_test.go
@@ -739,6 +739,20 @@ func TestGetName(t *testing.T) {
 	}
 }
 
+func TestSetName(t *testing.T) {
+	t.Parallel()
+
+	b := Base{
+		Name: "TESTNAM3",
+	}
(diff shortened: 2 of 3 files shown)


shazbert (changes requested):
Linter issues need to be fixed in test file then its an ACK from me.

exchanges/exchange.go
```diff
@@ -340,6 +340,11 @@ func (b *Base) GetName() string {
 	return b.Name
 }
 
+// SetName is a method that overrides the name of the exchange base
+func (b *Base) SetName(name string) {
+	b.Name = name
```
gloriousCode, line 345: Given GRPC's asynchronous nature and how much we reference exchange's by their names, having the ability to change an exchange's name at any time, this function introduces data races. 

There are a couple of options:
- One can already modify exchange names via `b := e.GetBase()`, `b.Name = "wow!"`. It doesn't solve the data race, but `SetName` isn't really an option I want to export to the interface without protection.
- You may instead wish to modify setup of exchanges to allow the name to be changed at that stage and never again. 
  - eg in your interface `CustomExchangeBuilder`, modify `NewExchangeByName(name string)` to have `NewExchangeByName(referenceName, modifiedExchangeName string)`.  
- Otherwise, introducing locks around getting setting the exchange name sounds far more arduous but offers race protection

gloriousCode (changes requested):

exchanges/exchange.go
lrascao: Seeing that this is a very specific use case maybe it's best to keep things simple and instead only offer the `b := e.GetBase(), b.Name = "wow!"`option for now and revisit if similar scenario comes along.

shazbert: @lrascao I agree with that option for now, as well.

lrascao:
closing

PR #758, opened 2021-08-18
gctcli: remove all exchange name client-side validation

lrascao:
Since now exchange names can be user-assigned we can no longer have
client-side validation, all exchange name validation must now occur
on the server (it was already doing that).

- [X] go test ./... -race
- [X] golangci-lint run

## Checklist

### DIFF
--- a/cmd/gctcli/commands.go
+++ b/cmd/gctcli/commands.go
@@ -291,10 +291,6 @@ func enableExchange(c *cli.Context) error {
 		exchangeName = c.Args().First()
 	}
 
-	if !validExchange(exchangeName) {
-		return errInvalidExchange
-	}
-
 	conn, err := setupClient()
 	if err != nil {
 		return err
@@ -341,10 +337,6 @@ func disableExchange(c *cli.Context) error {
 		exchangeName = c.Args().First()
 	}
 
-	if !validExchange(exchangeName) {
-		return errInvalidExchange
-	}
-
 	conn, err := setupClient()
 	if err != nil {
 		return err
@@ -389,10 +381,6 @@ func getExchangeOTPCode(c *cli.Context) error {
 		exchangeName = c.String("exchange")
 	} else {
 		exchangeName = c.Args().First()
-	}
-
-	if !validExchange(exchangeName) {
-		return errInvalidExchange
 	}
 
 	conn, err := setupClient()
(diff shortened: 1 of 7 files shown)


shazbert (approved):
tACK

thrasher- (changes requested):
Can also rid the errInvalidExchange var as it is no longer needed, then this looks good to go

cmd/gctcli/commands.go
```diff
@@ -3589,10 +3483,6 @@ func getTickerStream(c *cli.Context) error {
 		exchangeName = c.Args().First()
 	}
 
-	if !validExchange(exchangeName) {
```
gloriousCode, line 3592: The only concern here is that rpc function `GetTickerStream` will return the error `fmt.Errorf("ticker item not found for %s %s %s",
			exchange,
			p,
			a)`
which is less clear than exchange not found. Would you be willing to add
 ```
exch := bot.GetExchangeByName(r.Exchange)
	if exch == nil {
		return nil, ErrExchangeNotFound
	}
``` 
to `GetExchangeTickerStream`?

File: cmd/gctcli/commands.go [hunk 2]
```diff
@@ -391,10 +383,6 @@ func getExchangeOTPCode(c *cli.Context) error {
 		exchangeName = c.Args().First()
 	}
```
gloriousCode, line 385: Same with here. You'll receive a generic error `"exchange does not have a OTP secret stored"` which doesn't say its not loaded

gloriousCode (changes requested):
Thanks for this PR! Two minor comments

shazbert (approved):
Changes ACK! Thank you. 

Only slight issues I found, I don't know if you want to improve these outputs?
```
❯ go run ./... getexchangeorderbookstream -exchange="xxx"
2021/08/23 08:27:46 rpc error: code = Unknown desc = cannot find orderbook(s) for xxx exchange

❯ go run ./... gettickerstream -exchange="xxx" -pair="btc-usdt" -asset=spot
2021/08/23 08:29:08 rpc error: code = Unknown desc = ticker item not found for xxx btcusdt spot
```

NOTE: Also maybe in future we can return an error in engine method getExchangeByName(r.Exchange). I may open a separate PR for that.

gloriousCode (approved):
tACK! Thanks for addressing that server side validation 🎉

gloriousCode (approved):
ACK changes!

thrasher- (approved):
Thanks for adding in the additional gRPC checks, tACK

shazbert (approved):
Changes ACK!

PR #774, opened 2021-09-02
engine/exchanges: Add exchange currency state subsystem

shazbert:
# PR Description

Add asset state system (POC) for correctly identifying operational currencies for trading, withdrawal, depositing. Adds in waiting functionality so we can pend operations based on change of state. 

**Opened as draft for feedback and additional functionality request that is in scope.**

Things needed to be done: 

- [x] Add to withdraw and deposit manager for checking
- [x] Go across and add functionality to all exchanges (this can be done in a separate PRs)
- [x] Add gRPC fetching of currency functionality
- [ ] Add more here ....

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


engine/state_manager.go
```diff
@@ -0,0 +1,136 @@
...
+
+const (
+	StateManagement          = "StateManagement"
+	DefaultStateManagerDelay = time.Minute
+)
+
+var (
+	errSubsystemNotSetup = errors.New("subsystem not set up")
+	errNilInterface      = errors.New("interface is nil")
+)
```
gloriousCode, line 26: The first error already exist as `ErrNilSubsystem` and the second one can be moved to `subsystem_types`

engine/state_manager.go
```diff
@@ -0,0 +1,136 @@
...
+	if atomic.LoadInt32(&s.started) == 0 {
+		return fmt.Errorf("%s %w", StateManagement, ErrSubSystemNotStarted)
+	}
+
+	log.Debugf(log.ExchangeSys, "%s %s", StateManagement, MsgSubSystemShuttingDown)
+	close(s.shutdown)
+	s.wg.Wait()
+	s.shutdown = make(chan struct{})
+	log.Debugf(log.ExchangeSys, "%s %s", StateManagement, MsgSubSystemShutdown)
+	atomic.CompareAndSwapInt32(&s.started, 1, 0)
```
gloriousCode, line 89: Can be changed to `StoreInt32` for performance as you've already done a value check at start. Its 2 whole nanoseconds in savings! :D Unless its because you're worried that it could fail from another status change elsewhere?

engine/state_manager.go
```diff
@@ -0,0 +1,136 @@
...
+			}
+			wg.Wait() // This causes some variability in the timer due to
+			// longest length of request time. Can do time.Ticker but don't
+			// want routines to stack behind, this is more uniform.
+			timer.Reset(s.sleep)
+		}
+	}
+}
+
+func update(exch exchange.IBotExchange, wg *sync.WaitGroup, enabledAssets asset.Items) {
```
gloriousCode, line 124: Please add this to `(s *stateManager)` as its related to it. Normally it wouldn't matter so much, but its best to avoid it when in the large `engine` package

engine/withdraw_manager.go
```diff
@@ -45,6 +47,12 @@ func (m *WithdrawManager) SubmitWithdrawal(ctx context.Context, req *withdraw.Re
 		RequestDetails: *req,
 	}
 
+	// Determines if the currency can be withdrawn from the exchange
+	errF := exch.CanWithdraw(req.Currency, asset.Spot)
```
gloriousCode, line 51: Any reason why you're making `errF` instead of using `err`?

exchanges/state/state.go
```diff
@@ -0,0 +1,392 @@
...
+	"sync"
+
+	"github.com/thrasher-corp/gocryptotrader/common/convert"
+	"github.com/thrasher-corp/gocryptotrader/currency"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/alert"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+)
+
+var (
+	manager Manager
```
gloriousCode, line 15: Is there a way we can do this without globals? I would really prefer it. It looks like the manager aspect of this package could really all be moved to the engine level manager and this package still works. You're mostly using this manager as a pointer at the engine level anyway

exchanges/state/state.go
```diff
@@ -0,0 +1,392 @@
...
+		c.trading = true
+		c.tradingAlerts.Alert()
+	} else if c.trading != *o.Trade {
+		c.trading = *o.Trade
+		c.tradingAlerts.Alert()
+	}
+	c.mtx.Unlock()
+}
+
+// CanTrade returns if the currency is currently tradeable
```
gloriousCode, line 325: tradable, even though it looks wrong

exchanges/state/state.go
```diff
@@ -0,0 +1,392 @@
...
+	return e, nil
+}
+
+// States defines all currency states for an exchange
+type States struct {
+	m   map[asset.Item]map[*currency.Item]*Currency
+	mtx sync.RWMutex
+}
+
+// GetSnapshot returns the current exchange snapshot of all currency operation
```
gloriousCode, line 119: GetStateSnapshot

gloriousCode (changes requested):
Good stuff! This is a cool feature. The main things I'd like to see changed is the removal of the global, more specific naming than "state" as state can have many usages and the usage here is specific to exchange currency status and also the tests are currently broken.

exchanges/state/state.go
shazbert: Yeah the reason for this is to keep in line with the orderbook/ticker packages so if somebody uses exchange package(s) decoupled from the engine you can still update, retrieve and implement easily.

engine/withdraw_manager.go
shazbert: I did it this way because I check for an error ErrCurrencyStateNotFound which is suppressed, further functionality will continue and be error free but it will return this error. So I specifically quarantine this error that has been suppressed so I don't need to nil it afterwards.

exchanges/state/state.go
gloriousCode: I don't think the manager here needs to have knowledge of all exchange currency states in this case. I think an exchange can keep track of its own `*States` property, in the same way each exchange retrieves and updates what currencies are available to itself. 

Someone using package level exchanges would still have access to the `States` property of the `exchange.Base` and call `UpdateCurrencyStates`. Then the engine level `CurrencyStateManager` is responsible for asking each exchange to update their states on a timely basis.

If you remove the `manager` global var and the `Manager` type from `/exchanges/currencystate/currency_state.go`, you can do the following:
```
func SetupCurrencyState() *States {
	return &States{m: make(map[asset.Item]map[*currency.Item]*Currency)}
}
```
Then in `/exchanges/exchange.go` function `SetupDefaults(...)` becomes
```
b.States = currencystate.SetupCurrencyState()
```
Finally, in `/engine/currency_state_manager.go` you can remove the `*currencystate.Manager` property and everything is still tracked and working because the `order_manager` for example checks if the currency can be traded via an exchange interface call. If things were checked against the `/engine/currency_state_manager.go`, then that could be different, but at this current point it is not.

I think it can simplify things a bit while still addressing your concerns

shazbert: Thanks for explaining that, yeah I see what your saying, I will follow your example.

engine/currency_state_manager.go
```diff
@@ -0,0 +1,129 @@
...
+
+	"github.com/thrasher-corp/gocryptotrader/common"
+	exchange "github.com/thrasher-corp/gocryptotrader/exchanges"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+	"github.com/thrasher-corp/gocryptotrader/log"
+)
+
+const (
+	CurrencyStateManagementName = "CurrencyStateManagement"
+	DefaultStateManagerDelay    = time.Minute
```
gloriousCode, line 19: Should these have comments?

engine/currency_state_manager.go
```diff
@@ -0,0 +1,129 @@
...
+	var c CurrencyStateManager
+	if interval <= 0 {
+		log.Warnf(log.ExchangeSys,
+			"%s interval is invalid, defaulting to: %s",
+			CurrencyStateManagementName,
+			DefaultStateManagerDelay)
+		c.sleep = DefaultStateManagerDelay
+	} else {
+		c.sleep = interval
+	}
```
gloriousCode, line 42:
```suggestion
if interval <= 0 {
	log.Warnf(log.ExchangeSys,
			"%s interval is invalid, defaulting to: %s",
			CurrencyStateManagementName,
			DefaultStateManagerDelay)
	interval = DefaultStateManagerDelay
}
c.sleep = interval
```

engine/currency_state_manager.go
```diff
@@ -0,0 +1,129 @@
...
+// CurrencyStateManager manages currency states
+type CurrencyStateManager struct {
+	started  int32
+	shutdown chan struct{}
+	wg       sync.WaitGroup
+	iExchangeManager
+	sleep time.Duration
+}
+
+// Setup applies configuration parameters before running
```
gloriousCode, line 31:
```suggestion
// SetupCurrencyStateManager applies configuration parameters before running
```

engine/currency_state_manager.go
```diff
@@ -0,0 +1,129 @@
+package engine
```
gloriousCode, line 1: This whole file needs tests

engine/currency_state_manager.go
```diff
@@ -0,0 +1,129 @@
...
+					CurrencyStateManagementName,
+					err)
+			}
+			for x := range exchs {
+				wg.Add(1)
+				go c.update(exchs[x], &wg, exchs[x].GetAssetTypes(true))
+			}
+			wg.Wait() // This causes some variability in the timer due to
+			// longest length of request time. Can do time.Ticker but don't
+			// want routines to stack behind, this is more uniform.
```
gloriousCode, line 111: Yeah this is better than the `processing` flags we've used to prevent doubling up of processing in other subsystems. Good stuff 🎉

engine/rpcserver.go
```diff
@@ -3889,3 +3889,98 @@ func (s *RPCServer) UpdateDataHistoryJobPrerequisite(_ context.Context, r *gctrp
 	}
 	return &gctrpc.GenericResponse{Status: status, Data: fmt.Sprintf("Set job '%v' prerequisite job to '%v' and set status to paused", r.Nickname, r.PrerequisiteJobNickname)}, nil
 }
+
+// StateGetAll returns a full snapshot of currency states, whether they are able
+// to be withdrawn, deposited or traded on an exchange.
+func (s *RPCServer) StateGetAll(_ context.Context, r *gctrpc.StateGetAllRequest) (*gctrpc.StateResponse, error) {
```
gloriousCode, line 3895: Let's rename these functions too with CurrencyState

engine/rpcserver.go
```diff
@@ -3889,3 +3889,98 @@ func (s *RPCServer) UpdateDataHistoryJobPrerequisite(_ context.Context, r *gctrp
...
+			Asset:           sh[x].Asset.String(),
+			WithdrawEnabled: sh[x].Withdraw == nil || *sh[x].Withdraw,
+			DepositEnabled:  sh[x].Deposit == nil || *sh[x].Deposit,
+			TradingEnabled:  sh[x].Trade == nil || *sh[x].Trade,
+		})
+	}
+	return resp, nil
+}
+
+// StateDeposit determines via RPC if the currency code is operational for
```
gloriousCode, line 3919: CurrencyStateWithdraw

engine/rpcserver.go
```diff
@@ -3889,3 +3889,98 @@ func (s *RPCServer) UpdateDataHistoryJobPrerequisite(_ context.Context, r *gctrp
...
+		return nil, err
+	}
+
+	sh, err := exch.GetCurrencyStateSnapshot()
+	if err != nil {
+		return nil, err
+	}
+
+	var resp = &gctrpc.StateResponse{}
+	for x := range sh {
```
gloriousCode, line 3907: Quite minor, but if you also override the function `GetCurrencyStateSnapshot` to return a singular snappy, then this for loop will also get coverage

exchanges/currencystate/currency_state.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	p, err := s.Get(c, a)
+	if err != nil {
+		return err
+	}
+	if !p.CanWithdraw() {
+		return errWithdrawalsNotAllowed
+	}
+	return nil
```
gloriousCode, line 108: It would be good to get test coverage of these two paths

exchanges/currencystate/currency_state.go
```diff
@@ -0,0 +1,312 @@
...
+	}
+
+	p, err := s.Get(c, a)
+	if err != nil {
+		return err
+	}
+	if !p.CanDeposit() {
+		return errDepositNotAllowed
+	}
+	return nil
```
gloriousCode, line 124: It would be good to get test coverage of these two paths

exchanges/currencystate/currency_state.go
```diff
@@ -0,0 +1,312 @@
...
+// Options defines the current allowable options for a currency, using a bool
+// pointer for optional setting for incomplete data, so we can default to true
+// on nil values.
+type Options struct {
+	Withdraw *bool
+	Deposit  *bool
+	Trade    *bool
+}
+
+// Snapshot defines a snpashot of the internal asset for exportation
```
gloriousCode, line 307:
```suggestion
// Snapshot defines a snapshot of the internal asset for exportation
```

exchanges/alert/alert.go
```diff
@@ -0,0 +1,83 @@
...
+	sema uint32
+	// After closing the forAlert channel this will notify when all the routines
+	// that have waited, have either checked the orderbook depth or finished.
+	wg sync.WaitGroup
+	// Segregated lock only for waiting routines, so as this does not interfere
+	// with the main depth lock, acts as a rolling gate.
+	m sync.Mutex
+}
+
+// alert establishes a state change on the orderbook depth.
```
gloriousCode, line 24:
```suggestion
// Alert establishes a state change on the orderbook depth.
```

exchanges/alert/alert.go
```diff
@@ -0,0 +1,83 @@
...
+	close(n.forAlert)
+	// Wait for waiting routines to receive alert and return.
+	n.wg.Wait()
+	atomic.SwapUint32(&n.sema, 0) // Swap back to neutral state.
+	n.m.Unlock()
+}
+
+// Wait pauses calling routine until depth change has been established via depth
+// method alert. Kick allows for cancellation of waiting or when the caller has
+// has been shut down, if this is not needed it can be set to nil. This
```
gloriousCode, line 50:
```suggestion
// method alert. Kick allows for cancellation of waiting or when the caller
// has been shut down, if this is not needed it can be set to nil. This
```

gloriousCode (changes requested):
Thanks for making those changes! Mostly just semantics and test coverage now 🎉

gloriousCode:
Oh one other thing I forgot to mention. Is this something you would want to add to the config? I imagine if someone wants to have a custom polling time or having to enable/disable the currency state manager every time may be cumbersome

gloriousCode (changes requested):
Thanks for making those changes! The last concern I have is getting fuller picture over why there isn't a currency state returned. Tested all CLI endpoints and tested against supported and unsupported exchanges 🎉

engine/currency_state_manager.go
```diff
@@ -0,0 +1,279 @@
...
+				exch.GetName(),
+				enabledAssets[y],
+				err)
+		}
+	}
+}
+
+// CurrencyStateGetAllRPC returns a full snapshot of currency states, whether they
+// are able to be withdrawn, deposited or traded on an exchange for RPC.
+func (c *CurrencyStateManager) CurrencyStateGetAllRPC(exchName string) (*gctrpc.CurrencyStateResponse, error) {
```
gloriousCode, line 181: Given that we are in the CurrencyStateManager, I do not think you need to specify `CurrencyState` on these function names. Also update comments across each of these

engine/currency_state_manager.go
```diff
@@ -0,0 +1,279 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return enabled, nil
+}
+
+// CurrencyStateDeposit determines via RPC if the currency code is operational for
+// depositing to an exchange for RPC
+func (c *CurrencyStateManager) CurrencyStateDepositRPC(exchName string, cc currency.Code, a asset.Item) (*gctrpc.GenericResponse, error) {
+	if !c.IsRunning() {
```
gloriousCode, line 230: I think the test coverage for these functions should be within currency_state_manager_tests.go. The rpcserver_test.go coverage of their functions simply need to call their function and not care about the results.

exchanges/bithumb/bithumb_wrapper.go
```diff
@@ -891,3 +893,20 @@ func (b *Bithumb) UpdateOrderExecutionLimits(ctx context.Context, _ asset.Item)
 	}
 	return b.LoadLimits(limits)
 }
+
+// UpdateCurrencyStates updates currency states for exchange
+func (b *Bithumb) UpdateCurrencyStates(ctx context.Context, a asset.Item) error {
```
gloriousCode, line 898: Please add coverage

exchanges/bithumb/bithumb.go
```diff
@@ -150,6 +150,21 @@ func (b *Bithumb) GetAssetStatus(ctx context.Context, symbol string) (*Status, e
 	return &response, nil
 }
 
+// GetAssetStatus returns the withdrawal and deposit status for the symbol
```
gloriousCode, line 153:
```suggestion
// GetAssetStatusAll returns the withdrawal and deposit status for all symbols
```

gloriousCode (changes requested):
Thanks for making those changes! Few more minor niterinos

exchanges/alert/alert.go
```diff
@@ -61,8 +61,8 @@ func (n *Notice) Wait(kick <-chan struct{}) <-chan bool {
 	return reply
 }
 
-// hold waits on either channel in the event that the routine has finished or an
-// alert from a depth update has occurred.
+// hold waits on either channel in the event that the routine has
+// finished/canceled or an alert from a depth update has occurred.
```
gloriousCode, line 65: This is incredibly minor, but we have 248 usages of `cancelled` in our codebase and only 53 of `canceled` additionaly, the `canceled` usage is interpreting foreign APIs. Can we use `cancelled`?

engine/currency_state_manager_test.go
```diff
@@ -63,6 +73,43 @@ func (f *fakerino) GetName() string {
...
+}
+
+func (f *fakerino) CanTradePair(p currency.Pair, a asset.Item) error {
+	if f.errorMe {
+		return errExchange
+	}
+	return nil
+}
+
 func TestCurrencyStateManagerCoolRunnings(t *testing.T) {
```
gloriousCode, line 113: 👀 I missed this one earlier 👀

engine/currency_state_manager_test.go
```diff
@@ -50,7 +60,7 @@ type fakerino struct {
 
 func (f *fakerino) UpdateCurrencyStates(_ context.Context, _ asset.Item) error {
 	if f.errorMe {
-		return errors.New("norty")
+		return errExchange
```
gloriousCode, line 63: Can this one become `common.ErrNotYetImplemented` to utilise the error path in the function `update`? Also I really like the `errorMe` bool use for function overrides 🤙

gloriousCode (changes requested):
A few more really minor things :D Thanks for making those changes!

thrasher- (commented):
A nice PR! Tested all the currency state gctcli commands for bithumb and unimplemented exchanges, plus an empty currency code and invalid asset to ensure it's handled. Futher improvements around this would be:

1) Adding functionality to whether or not a user can trade certain types of assets. For example, Binance no longer supports specific assets for some regions of the world (leveraged tokens, futures)
2) Knowing whether an exchange has a global shared asset state for all funds (like FTX) compared to ones which have internal mechanisms for transferring funds from your spot to futures/options/margin accounts.

shazbert:
> A nice PR! Tested all the currency state gctcli commands for bithumb and unimplemented exchanges, plus an empty currency code and invalid asset to ensure it's handled. Futher improvements around this would be:
> 
> 1. Adding functionality to whether or not a user can trade certain types of assets. For example, Binance no longer supports specific assets for some regions of the world (leveraged tokens, futures)
> 2. Knowing whether an exchange has a global shared asset state for all funds (like FTX) compared to ones which have internal mechanisms for transferring funds from your spot to futures/options/margin accounts.

Public trello item has been created for these pending updates: [TRELLO](https://trello.com/c/sCCyYnUz/134-currency-state-management)

exchanges/bitfinex/bitfinex.go
```diff
@@ -493,8 +493,17 @@ func (b *Bitfinex) GetDerivativeStatusInfo(ctx context.Context, keys, startTime,
 			return finalResp, fmt.Errorf("%v GetDerivativeStatusInfo: %w for MarkPrice", b.Name, errTypeAssert)
 		}
 
-		if response.OpenInterest, ok = result[z][18].(float64); !ok {
-			return finalResp, fmt.Errorf("%v GetDerivativeStatusInfo: %w for OpenInterest", b.Name, errTypeAssert)
+		switch t := result[z][18].(type) {
```
gloriousCode, line 496: Any chance you could swoop in for a quick typo fix for the property `NextFundingAccured` -> `NextFundingAccrued`? 🥲

gloriousCode (approved):
tACK! Nice work @shazbert I have a mico-semi-related nit that caught my eye since you fixed a test 👀 It would be rather nice to not have the typo, but I'm not blocking this from being merged because its already existing.

exchanges/bitfinex/bitfinex.go
shazbert: 🦅 done.

gloriousCode (approved):
uTACK speedy speedy changes!
# 🏎️💨💨💨

thrasher- (approved):
tACK 🚀

PR #775, opened 2021-09-02
engine/order manager: Initial REST managed order updating (resolves #772)

### LINKED ISSUE #772: Managed orders aren't properly updated when in REST mode
# New Issue

## Expected Behavior

When a placed order is filled or partially filled, the orderStore is expected to update the order status accordingly, both in WS and in REST mode.

## Current Behavior

For exchanges where no websocket stream is available/active, active orders will never get updated: they'll never turn into filled or partially filled orders.

## Failure Information (for bugs)

When websockets are enabled, WebsocketDataHandler() handles *order.Detail data: it either adds or updates it in the orderStore.

When websockets are not enabled, the orderStore items are regularly updated in order_manager.go:processOrders(). Here, ActiveOrders are fetched using REST, and active orders are added to the orderStore.
(Orders are also added when using the orderManager's Submit() function.)
However, if this order already was present in the orderStore, it's not updated (eg to reflect a new partially filled status).
Also, if the order was taken, it's not returned by the REST call, and subsequently it's not removed from the orderStore.

### Steps to Reproduce

1. Disable websockets
2. Place an order
3. Cancel the order
4. Query or debug the active order list

TaltaM:
# PR Description

When a placed order is filled or partially filled, the orderStore is expected to update the order status accordingly, both in WS and in REST mode. Currently, for exchanges where no websocket stream is available/active, active orders will never get updated: they'll never turn into filled or partially filled orders.
This PR addresses that issue by tracking missing orders and performing REST calls to verify and update their status.

Fixes #772 

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] I tested this by placing an order on an exchange and cancelling it (after adding some additional debugging output).

## Checklist

### DIFF
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -641,16 +641,15 @@ func (m *OrderManager) processOrders() {
 			}
 
 			for z := range result {
-				err = m.UpsertOrder(&result[z])
+				upsertResponse, err := m.UpsertOrder(&result[z])
 				if err != nil {
 					log.Error(log.OrderMgr, err)
 				}
-				requiresProcessing[result[z].InternalOrderID] = false
+				requiresProcessing[upsertResponse.OrderDetails.InternalOrderID] = false
 			}
 			if !exchanges[i].GetBase().GetSupportedFeatures().RESTCapabilities.GetOrder {
 				continue
 			}
-
 			wg.Add(1)
 			go m.processMatchingOrders(exchanges[i], orders, requiresProcessing, &wg)
 		}
@@ -684,7 +683,8 @@ func (m *OrderManager) FetchAndUpdateExchangeOrder(exch exchange.IBotExchange, o
 		return err
 	}
 	fetchedOrder.LastUpdated = time.Now()
-	return m.UpsertOrder(&fetchedOrder)
+	_, err = m.UpsertOrder(&fetchedOrder)
+	return err
 }
 
 // Exists checks whether an order exists in the order store
@@ -738,15 +738,15 @@ func (m *OrderManager) UpdateExistingOrder(od *order.Detail) error {
 }
 
 // UpsertOrder updates an existing order or adds a new one to the orderstore
-func (m *OrderManager) UpsertOrder(od *order.Detail) error {
+func (m *OrderManager) UpsertOrder(od *order.Detail) (resp *OrderUpsertResponse, err error) {
 	if m == nil {
-		return fmt.Errorf("order manager %w", ErrNilSubsystem)
+		return nil, fmt.Errorf("order manager %w", ErrNilSubsystem)
 	}
 	if atomic.LoadInt32(&m.started) == 0 {
-		return fmt.Errorf("order manager %w", ErrSubSystemNotStarted)
+		return nil, fmt.Errorf("order manager %w", ErrSubSystemNotStarted)
 	}
 	if od == nil {
-		return errNilOrder
+		return nil, errNilOrder
 	}
 	var msg string
 	defer func(message *string) {
@@ -763,25 +763,25 @@ func (m *OrderManager) UpsertOrder(od *order.Detail) error {
 	upsertResponse, err := m.orderStore.upsert(od)
 	if err != nil {
 		msg = fmt.Sprintf(
-			"Order manager: Exchange %s unable to upsert order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v: %s",
-			od.Exchange, od.ID, od.InternalOrderID, od.Pair, od.Price, od.Amount, od.Side, od.Type, err)
-		return err
+			"Order manager: Exchange %s unable to upsert order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v status=%v: %s",
+			od.Exchange, od.ID, od.InternalOrderID, od.Pair, od.Price, od.Amount, od.Side, od.Type, od.Status, err)
+		return nil, err
 	}
 
 	status := "updated"
 	if upsertResponse.IsNewOrder {
 		status = "added"
 	}
-	msg = fmt.Sprintf("Order manager: Exchange %s %s order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v.",
+	msg = fmt.Sprintf("Order manager: Exchange %s %s order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v status=%v.",
 		upsertResponse.OrderDetails.Exchange, status, upsertResponse.OrderDetails.ID, upsertResponse.OrderDetails.InternalOrderID,
 		upsertResponse.OrderDetails.Pair, upsertResponse.OrderDetails.Price, upsertResponse.OrderDetails.Amount,
-		upsertResponse.OrderDetails.Side, upsertResponse.OrderDetails.Type)
+		upsertResponse.OrderDetails.Side, upsertResponse.OrderDetails.Type, upsertResponse.OrderDetails.Status)
 	if upsertResponse.IsNewOrder {
 		log.Info(log.OrderMgr, msg)
-		return nil
+		return upsertResponse, nil
 	}
 	log.Debug(log.OrderMgr, msg)
-	return nil
+	return upsertResponse, nil
 }
 
 // get returns all orders for all exchanges
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -570,10 +570,6 @@ func (m *OrderManager) processSubmittedOrder(newOrder *order.Submit, result orde
 // processOrders iterates over all exchange orders via API
 // and adds them to the internal order store
 func (m *OrderManager) processOrders() {
-	if m.orderStore.exchangeManager == nil {
-		log.Errorf(log.OrderMgr, "order manager: %v", ErrNilSubsystem)
-	}
-	log.Warnf(log.OrderMgr, "order manager: getting exchanges")
 	exchanges, err := m.orderStore.exchangeManager.GetExchanges()
 	if err != nil {
 		log.Errorf(log.OrderMgr, "Order manager cannot get exchanges: %v", err)
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -658,7 +658,11 @@ func (m *OrderManager) processOrders() {
 }
 
 func (m *OrderManager) processMatchingOrders(exch exchange.IBotExchange, orders []order.Detail, requiresProcessing map[string]bool, wg *sync.WaitGroup) {
-	defer wg.Done()
+	defer func() {
+		if wg != nil {
+			wg.Done()
+		}
+	}()
 	for x := range orders {
 		if time.Since(orders[x].LastUpdated) < time.Minute {
 			continue
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -601,6 +601,14 @@ func (m *OrderManager) processOrders() {
 				Exchange: exchanges[i].GetName(),
 			}
 			orders, err := m.orderStore.getActiveOrders(filter)
+			if err != nil {
+				log.Errorf(log.OrderMgr,
+					"Order manager: Unable to get active orders for %s and asset type %s: %s",
+					exchanges[i].GetName(),
+					supportedAssets[y],
+					err)
+				continue
+			}
 			order.FilterOrdersByCurrencies(&orders, pairs)
 			requiresProcessing := make(map[string]bool, len(orders))
 			for x := range orders {
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -567,6 +567,10 @@ func (m *OrderManager) processSubmittedOrder(newOrder *order.Submit, result orde
 // processOrders iterates over all exchange orders via API
 // and adds them to the internal order store
 func (m *OrderManager) processOrders() {
+	if m.orderStore.exchangeManager == nil {
+		log.Errorf(log.OrderMgr, "order manager: %v", ErrNilSubsystem)
+	}
+	log.Warnf(log.OrderMgr, "order manager: getting exchanges")
 	exchanges := m.orderStore.exchangeManager.GetExchanges()
 	for i := range exchanges {
 		if !exchanges[i].GetAuthenticatedAPISupport(exchange.RestAuthentication) {
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -570,10 +570,17 @@ func (m *OrderManager) processSubmittedOrder(newOrder *order.Submit, result orde
 // processOrders iterates over all exchange orders via API
 // and adds them to the internal order store
 func (m *OrderManager) processOrders() {
+	if !atomic.CompareAndSwapInt32(&m.processingOrders, 0, 1) {
+		return
+	}
+	defer func() {
+		atomic.StoreInt32(&m.processingOrders, 0)
+	}()
 	exchanges, err := m.orderStore.exchangeManager.GetExchanges()
 	if err != nil {
 		log.Errorf(log.OrderMgr, "Order manager cannot get exchanges: %v", err)
 	}
+	var wg sync.WaitGroup
 	for i := range exchanges {
 		if !exchanges[i].GetAuthenticatedAPISupport(exchange.RestAuthentication) {
 			continue
@@ -652,12 +659,15 @@ func (m *OrderManager) processOrders() {
 				continue
 			}
 
-			go m.processMatchingOrders(exchanges[i], orders, requiresProcessing)
+			wg.Add(1)
+			go m.processMatchingOrders(exchanges[i], orders, requiresProcessing, &wg)
 		}
 	}
+	wg.Wait()
 }
 
-func (m *OrderManager) processMatchingOrders(exch exchange.IBotExchange, orders []order.Detail, requiresProcessing map[string]bool) {
+func (m *OrderManager) processMatchingOrders(exch exchange.IBotExchange, orders []order.Detail, requiresProcessing map[string]bool, wg *sync.WaitGroup) {
+	defer wg.Done()
 	for x := range orders {
 		if time.Since(orders[x].LastUpdated) < time.Minute {
 			continue
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -87,8 +87,11 @@ func (m *OrderManager) Stop() error {
 func (m *OrderManager) gracefulShutdown() {
 	if m.cfg.CancelOrdersOnShutdown {
 		log.Debugln(log.OrderMgr, "Order manager: Cancelling any open orders...")
-		m.CancelAllOrders(context.TODO(),
-			m.orderStore.exchangeManager.GetExchanges())
+		exchanges, err := m.orderStore.exchangeManager.GetExchanges()
+		if err != nil {
+			log.Errorf(log.OrderMgr, "Order manager cannot get exchanges: %v", err)
+		}
+		m.CancelAllOrders(context.TODO(), exchanges)
 	}
 }
 
@@ -571,7 +574,10 @@ func (m *OrderManager) processOrders() {
 		log.Errorf(log.OrderMgr, "order manager: %v", ErrNilSubsystem)
 	}
 	log.Warnf(log.OrderMgr, "order manager: getting exchanges")
-	exchanges := m.orderStore.exchangeManager.GetExchanges()
+	exchanges, err := m.orderStore.exchangeManager.GetExchanges()
+	if err != nil {
+		log.Errorf(log.OrderMgr, "Order manager cannot get exchanges: %v", err)
+	}
 	for i := range exchanges {
 		if !exchanges[i].GetAuthenticatedAPISupport(exchange.RestAuthentication) {
 			continue
@@ -674,7 +680,7 @@ func (m *OrderManager) FetchAndUpdateExchangeOrder(exch exchange.IBotExchange, o
 	if ord == nil {
 		return errors.New("order manager: Order is nil")
 	}
-	fetchedOrder, err := exch.GetOrderInfo(ord.ID, ord.Pair, assetType)
+	fetchedOrder, err := exch.GetOrderInfo(context.TODO(), ord.ID, ord.Pair, assetType)
 	if err != nil {
 		ord.Status = order.UnknownStatus
 		return err
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -242,12 +242,12 @@ func (m *OrderManager) GetOrderInfo(ctx context.Context, exchangeName, orderID s
 		return order.Detail{}, err
 	}
 
-	_, err = m.orderStore.upsert(&result)
+	upsertResponse, err := m.orderStore.upsert(&result)
 	if err != nil {
 		return order.Detail{}, err
 	}
 
-	return result, nil
+	return upsertResponse.OrderDetails, nil
 }
 
 // validate ensures a submitted order is valid before adding to the manager
@@ -748,7 +748,7 @@ func (m *OrderManager) UpsertOrder(od *order.Detail) error {
 		})
 	}(&msg)
 
-	updatedOrder, err := m.orderStore.upsert(od)
+	upsertResponse, err := m.orderStore.upsert(od)
 	if err != nil {
 		msg = fmt.Sprintf(
 			"Order manager: Exchange %s unable to upsert order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v: %s",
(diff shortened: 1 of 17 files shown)


engine/order_manager.go
```diff
@@ -601,22 +623,72 @@ func (m *OrderManager) processOrders() {
...
+			for x := range ordersToCheck {
+				curTime := time.Now()
+				// Only check if at least 1 minute not updated
+				if curTime.Sub(ordersToCheck[x].LastUpdated) < time.Minute {
+					checkedOrderIDs[ordersToCheck[x].InternalOrderID] = true
+				} else if checkedOrderIDs[ordersToCheck[x].InternalOrderID] == false {
+					log.Debugf(log.OrderMgr, "Order manager: Exchange %s Unchecked order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v",
+						ordersToCheck[x].Exchange, ordersToCheck[x].ID, ordersToCheck[x].InternalOrderID, ordersToCheck[x].Pair, ordersToCheck[x].Price,
+						ordersToCheck[x].Amount, ordersToCheck[x].Side, ordersToCheck[x].Type)
+					go m.FetchAndUpdateExchangeOrder(exchanges[i], &ordersToCheck[x], supportedAssets[y], curTime)
```
shazbert, line 668: This may cause some issues when we have ^1000 limit orders. This *might* bunch everything up at the rate limiter. I think we could employ a more linear approach: 

```go
			go func() {
				for x := range ordersToCheck {
					curTime := time.Now()
					// Only check if at least 1 minute not updated
					if curTime.Sub(ordersToCheck[x].LastUpdated) < time.Minute {
						continue
					}

					if checkedOrderIDs[ordersToCheck[x].InternalOrderID] {
						continue
					}

					log.Debugf(log.OrderMgr, "Order manager: Exchange %s Unchecked order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v",
						ordersToCheck[x].Exchange, ordersToCheck[x].ID, ordersToCheck[x].InternalOrderID, ordersToCheck[x].Pair, ordersToCheck[x].Price,
						ordersToCheck[x].Amount, ordersToCheck[x].Side, ordersToCheck[x].Type)
					m.FetchAndUpdateExchangeOrder(exchanges[i], &ordersToCheck[x], supportedAssets[y], curTime)
				}
			}()
```

and then maybe add a waitgroup to processOrders() just so we don't exceed order manager delay with our HTTP requests. Let me know what you think.

engine/order_manager.go
```diff
@@ -875,3 +947,44 @@ func (s *store) getFilteredOrders(f *order.Filter) ([]order.Detail, error) {
...
+
+// getActiveOrders returns copy of the orders that are active
+func (s *store) getActiveOrders(f *order.Filter) ([]order.Detail, error) {
+	if f == nil {
+		return nil, errors.New("filter is nil")
+	}
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	var os []order.Detail
```
shazbert, line 959: can we change os to another variable name as it conflicts with a package.

engine/order_manager.go
```diff
@@ -601,22 +623,72 @@ func (m *OrderManager) processOrders() {
...
 }
 
+func (m *OrderManager) FetchAndUpdateExchangeOrder(exch exchange.IBotExchange, ord *order.Detail, assetType asset.Item, curTime time.Time) {
+	fetchedOrder, err := exch.GetOrderInfo(ord.ID, ord.Pair, assetType)
+	if err != nil {
+		log.Errorf(log.OrderMgr, "Unable to get additional order info: %s", err.Error())
+		ord.Status = order.UnknownStatus
+		return
+	}
+	fetchedOrder.LastUpdated = curTime
```
shazbert, line 682: fetch request above might take some time due to rate limit etc. MIght be better to do time.Now() here.

shazbert (changes requested):
Had a quick look over the changes. Test coverage needs to be added and there are a few linter issues, apart from that looks good. Thanks for opening this PR. 🚀

gloriousCode (changes requested):
Thank you TaltaM for your contribution! Ive had a lookover and appreciate the feature. While testing I had made changes and so rather than leave individual comments, I've made a git patch here: https://gist.github.com/gloriousCode/e209074b514dfffc8b3f14d325e8527e 
Check it out and let me know what you think for changes. Otherwise, I can leave comments.

TaltaM:
@gloriousCode: Thank you! I love it.

One thing. Internal order ID's weren't generated by upsert when creating new entries. Mind if the function gets changed as per commit https://github.com/thrasher-corp/gocryptotrader/pull/775/commits/ffeef198c6eda82102f1d5878d2152ebf5bfa2eb ?

TaltaM:
> Had a quick look over the changes. Test coverage needs to be added and there are a few linter issues, apart from that looks good. Thanks for opening this PR. rocket

Cheers!
Can you point me to what needs test coverage and how?

engine/order_manager.go
TaltaM: Works for me!
This could create an issue if say 1000 orders are cancelled; if you want I can set up a waitgroup for the current solution (thanks @gloriousCode !)

engine/order_manager.go
```diff
@@ -675,7 +731,37 @@ func (m *OrderManager) UpsertOrder(od *order.Detail) error {
 	if atomic.LoadInt32(&m.started) == 0 {
 		return fmt.Errorf("order manager %w", ErrSubSystemNotStarted)
 	}
-	return m.orderStore.upsert(od)
+	if od == nil {
+		return errNilOrder
+	}
+	var msg string
+	defer func(message string) {
```
shazbert, line 738: can we change the string to a *string

engine/order_manager.go
```diff
@@ -743,27 +829,34 @@ func (s *store) modifyExisting(id string, mod *order.Modify) error {
...
+		return nil, err
 	}
 	s.m.Lock()
 	defer s.m.Unlock()
 	r, ok := s.Orders[lName]
 	if !ok {
+		od.GenerateInternalOrderID()
 		s.Orders[lName] = []*order.Detail{od}
-		return nil
+		return nil, nil
```
shazbert, line 847: might be better to return the pointer here as if its a nil error there might be a chance for panic to a calling function.

engine/order_manager.go
```diff
@@ -743,27 +829,34 @@ func (s *store) modifyExisting(id string, mod *order.Modify) error {
...
-			return nil
+			orderCopy := r[x].Copy()
+			return &orderCopy, nil
 		}
 	}
+	// Untracked websocket orders will not have internalIDs yet
+	od.GenerateInternalOrderID()
 	s.Orders[lName] = append(s.Orders[lName], od)
-	return nil
+	return nil, nil
```
shazbert, line 859: here as well.

exchanges/order/orders.go
```diff
@@ -208,6 +209,9 @@ func (d *Detail) UpdateOrderFromDetail(m *Detail) {
 	if d.ID == "" {
 		d.ID = m.ID
 	}
+	if d.InternalOrderID == "" {
```
shazbert, line 212: This is failing a test: ```gocryptotrader\exchanges\order\order_test.go:870: Should not be able to update the internal order ID```

exchanges/order/orders.go
```diff
@@ -405,6 +409,41 @@ func (d *Detail) MatchFilter(f *Filter) bool {
...
+	if d.Amount > 0 && d.Amount == d.ExecutedAmount {
+		return true
+	}
+	return d.Status == Filled || d.Status == Cancelled || d.Status == InsufficientBalance || d.Status == MarketUnavailable ||
+		d.Status == Rejected || d.Status == PartiallyCancelled || d.Status == Expired || d.Status == Closed
+}
+
+// GenerateInternalOrderID sets a new V4 order ID or a V5 order ID if
+// the V4 function returns an error
+func (d *Detail) GenerateInternalOrderID() {
```
shazbert, line 435: Please add tests for these functions in order_test.go

exchanges/order/orders.go
```diff
@@ -405,6 +409,41 @@ func (d *Detail) MatchFilter(f *Filter) bool {
...
+	if d.InternalOrderID == "" {
+		var id uuid.UUID
+		id, err := uuid.NewV4()
+		if err != nil {
+			id = uuid.NewV5(uuid.UUID{}, d.ID)
+		}
+		d.InternalOrderID = id.String()
+	}
+
```
shazbert, line 444: linter issues need to be rectified.

engine/order_manager.go
```diff
@@ -875,3 +959,46 @@ func (s *store) getFilteredOrders(f *order.Filter) ([]order.Detail, error) {
 	}
 	return os, nil
 }
+
+// getActiveOrders returns copy of the orders that are active
```
shazbert, line 963: If you have time could you please add a test for this and for the functions you have added in ```order_manager.go in``` ```order_manager_test.go```. Let me know if you can, if you can't, I can add a public trello Item.

shazbert (changes requested):
Changes looking great, just some minor critiques.

engine/order_manager.go
shazbert: also add in a nil check after pointer implementation.

exchanges/order/orders.go
TaltaM: Indeed: this PR changes how the internal order ID works, where I assume the scope of the ID to be the order manager.
- UpdateOrderFromDetail updates unset order ID's - and ignores new order ID's after the first update of the order ID.
- Order IDs are always generated: when a seeding error occurs for generating a V4 UUID, a V5 ID is deterministically derived from the exchange's order ID.

I'll be so free to adapt the test, not the code.

engine/order_manager.go
```diff
@@ -743,27 +832,45 @@ func (s *store) modifyExisting(id string, mod *order.Modify) error {
 
 // upsert (1) checks if such an exchange exists in the exchangeManager, (2) checks if
 // order exists and updates/creates it.
-func (s *store) upsert(od *order.Detail) error {
+func (s *store) upsert(od *order.Detail) (resp *OrderUpsertResponse, err error) {
+	if od == nil {
+		return &OrderUpsertResponse{}, errNilOrder
```
gloriousCode, line 837: `return nil, errNilOrder` I've read https://github.com/thrasher-corp/gocryptotrader/pull/775#discussion_r702890047 your comment on this, but when you're also returning an error, you only ever check that and return, so no need to instantiate a new struct+pointer

engine/order_manager.go
```diff
@@ -743,27 +832,45 @@ func (s *store) modifyExisting(id string, mod *order.Modify) error {
...
+	}
 	lName := strings.ToLower(od.Exchange)
-	_, err := s.exchangeManager.GetExchangeByName(lName)
+	_, err = s.exchangeManager.GetExchangeByName(lName)
 	if err != nil {
-		return err
+		return &OrderUpsertResponse{}, err
```
gloriousCode, line 842: Same with this one: `return nil, err`

engine/order_manager.go
```diff
@@ -875,3 +973,46 @@ func (s *store) getFilteredOrders(f *order.Filter) ([]order.Detail, error) {
...
+func (s *store) getActiveOrders(f *order.Filter) ([]order.Detail, error) {
+	if f == nil {
+		return nil, errors.New("filter is nil")
+	}
+	s.m.RLock()
+	defer s.m.RUnlock()
+
+	var orders []order.Detail
+	switch {
+	case f == nil:
```
gloriousCode, line 987: The first line in this function blocks this first `switch` `case` from being hit. Please either remove the `f==nil` check of remove this case

gloriousCode (changes requested):
Thanks for making those changes! I've got a couple more comments, thank you for taking the time to do these things

engine/order_manager.go
TaltaM: Ah yes I was confused

engine/order_manager.go
TaltaM: OK I added them, let's see if it's sufficient :)
I made 2 code changes based on the test cases, good!

TaltaM:
Thanks! Tests added, comments addressed, let's see :)

shazbert (changes requested):
Changes look good, however there seems be an issue of initially fetching of orders on startup, these may never be managed internally if only REST protocol is enabled now.

We may need to add an initial sync on startup just so any old orders from last instance can be included in managed orders. We might also add (this is open for discussion) a full sync every 10 minutes if REST is only enabled just so we can include front end orders if they occur while the bot is running.

TaltaM:
> Changes look good, however there seems be an issue of initially fetching of orders on startup, these may never be managed internally if only REST protocol is enabled now.

Orders are managed fine in REST-only mode for me, at least on Bittrex. I did have to add this line to bittrex_wrapper.go: https://github.com/thrasher-corp/gocryptotrader/blob/33b00cc8ce7be49690fa425247d0809fe2711e19/exchanges/bittrex/bittrex_wrapper.go#L86

Perhaps the exchange you're using for testing does not have support for GetOrderInfo() or does not have it set as enabled?

Currently orders start to be managed after the first timer tick (so after 10s). I'll add a commit that fetches all open orders before the timer is started (https://github.com/thrasher-corp/gocryptotrader/pull/775/commits/1df432351cf31c3f26b1795368caee8357229f21)

> 
> We may need to add an initial sync on startup just so any old orders from last instance can be included in managed orders. We might also add (this is open for discussion) a full sync every 10 minutes if REST is only enabled just so we can include front end orders if they occur while the bot is running.

I might misunderstand what you're saying, but this is what happens according to me in REST-only mode:
- The first time processOrders() runs it:
  - fetches all current open orders
  - adds all of them to the orderStore
- When processOrders() runs after an order has been closed/filled/..:
  - GetActiveOrders returns a list where that order is missing
  - GetOrderInfo fetches the missing order
  - As the order is closed, the order in the orderStore is marked as closed
  - The order won't show up next run
- If an order is submitted through the OrderManager (Submit()), it is added to the orderStore
- If an order is submitted outside of the OrderManager, it is added to the orderStore as long as the order is open when the 10s timer fires

So there is a possibility that orders are created and closed without ending up in the orderStore, namely, if opening an closing happens within this 10s window.

Is that the type of order you are targeting in this 10 minute full sync?

TaltaM:
Hm the test runs fine locally (`go test ./... -race`), but not here on github.. Did I do something silly?

gloriousCode:
> 
> 
> Hm the test runs fine locally (`go test ./... -race`), but not here on github.. Did I do something silly?

They do for me too. It may be more a result of other tests as the stacktrace does not mention which test caused it. However, looking at the panic, for the function ` (m *ExchangeManager) GetExchanges()` referenced, you may wish to have a check
```
	if m == nil {
		return nil, fmt.Errorf("exchange manager: %w", ErrNilSubsystem)
	}
```
This could help you narrow down how this is happening. The other thing which may be happening is that in the test `OrderSetup(t)` has the line `err = m.Start()` which creates the go routine loop of `processOrders`, which isn't necessary for the tests. You may wish to replace it with just  `m.started = 1`

I'll have another look later and edit this if I find anything else

TaltaM:
Thanks @gloriousCode 

I added nil checks. In the exchange manager, a nil item generates an error, while in the bot it is interpreted as 'there are no exchanges'.
Also, I disabled the runner thread, because indeed all it does is continuously run processOrders, which won't work anyway.

It is a bit dis-satisfactory, but after getting the processOrders function to work (setting up an exchanges with GetOrder REST capabilities, enabling currencies, creating a mock GetActiveOrders function, ...) I can only conclude that it actually does work as expected in the test runner without hitting a nil object..

engine/order_manager_test.go
```diff
@@ -156,7 +156,7 @@ func OrdersSetup(t *testing.T) *OrderManager {
 	if !errors.Is(err, nil) {
 		t.Errorf("error '%v', expected '%v'", err, nil)
 	}
-	err = m.Start()
+	m.started = 1
 	if !errors.Is(err, nil) {
```
gloriousCode, line 160: Can remove this `errors.Is` now

engine/order_manager.go
```diff
@@ -675,7 +742,43 @@ func (m *OrderManager) UpsertOrder(od *order.Detail) error {
...
+	}
+	msg = fmt.Sprintf("Order manager: Exchange %s %s order ID=%v internal ID=%v pair=%v price=%.8f amount=%.8f side=%v type=%v.",
+		upsertResponse.OrderDetails.Exchange, status, upsertResponse.OrderDetails.ID, upsertResponse.OrderDetails.InternalOrderID,
+		upsertResponse.OrderDetails.Pair, upsertResponse.OrderDetails.Price, upsertResponse.OrderDetails.Amount,
+		upsertResponse.OrderDetails.Side, upsertResponse.OrderDetails.Type)
+	if upsertResponse.IsNewOrder {
+		log.Infof(log.OrderMgr, "%s", msg)
+		return nil
+	}
+	log.Debugf(log.OrderMgr, "%s", msg)
```
gloriousCode, line 780: Think this shoud become `.Infof` as well

gloriousCode (changes requested):
Nice work on fixing the tests! Just two small things from me and I think this is good to go 🎉

engine/order_manager.go
```diff
@@ -552,7 +569,10 @@ func (m *OrderManager) processSubmittedOrder(newOrder *order.Submit, result orde
 // processOrders iterates over all exchange orders via API
 // and adds them to the internal order store
 func (m *OrderManager) processOrders() {
-	exchanges := m.orderStore.exchangeManager.GetExchanges()
+	exchanges, err := m.orderStore.exchangeManager.GetExchanges()
```
gloriousCode, line 572: Apologies for not picking up on this earlier, I think the only extra thing this feature needs is an atomic blocker here. By that I mean, if we are already running processOrders, another one won't run. The main concern I have is that if I was an aspiring trader with hundreds of active trades (😎 📈) and it takes longer than 10 seconds to process them all, we will have another `processOrders` running due to this function being called via `case <-tick.C: go m.processOrders()` and this could build up in the background as the application is running.

In order to prevent this, we can add a property to `OrderManager`: `processingOrders int32`
At the start of this function, we would have:
```
if !atomic.CompareAndSwapInt32(&m.processingOrders, 0, 1) {
		return
	}
	defer atomic.StoreInt32(&m.processingOrders, 0)
```

this then prevents any doubling up of processing orders. If you want to see what I mean, change `orderManagerDelay` to `time.Millisecond` and see how many processes are spawned

engine/order_manager.go
```diff
@@ -591,32 +629,61 @@ func (m *OrderManager) processOrders() {
...
+				if err != nil {
+					log.Error(log.OrderMgr, err)
 				}
+				requiresProcessing[result[z].InternalOrderID] = false
 			}
+			if !exchanges[i].GetBase().GetSupportedFeatures().RESTCapabilities.GetOrder {
+				continue
+			}
+
+			go m.processMatchingOrders(exchanges[i], orders, requiresProcessing)
```
gloriousCode, line 654: As a result of the above `defer` statement, I think this would be improved by removing the `go` call here, so that way once this function is done, we know that all possible orders have been processed and stored

gloriousCode (changes requested):
Thank you again for this. I have _one last request_, I know I just said the last one was, but this one is a precaution. If you don't have time to do this one, let me know and I can make a PR after this to add it.

gloriousCode:
I'm also noting some data races locally, surrounding the is authenticated checks via `if !exchanges[i].GetAuthenticatedAPISupport(exchange.RestAuthentication)`:
```
WARNING: DATA RACE
Write at 0x00c00013fb18 by goroutine 205:
  github.com/thrasher-corp/gocryptotrader/engine.TestGetOrders()
      C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/rpcserver_test.go:1079 +0x21d2
  testing.tRunner()
      C:/Program Files/Go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      C:/Program Files/Go/src/testing/testing.go:1306 +0x47

Previous read at 0x00c00013fb18 by goroutine 11:
  github.com/thrasher-corp/gocryptotrader/exchanges.(*Base).GetAuthenticatedAPISupport()
      C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/exchange.go:342 +0x8a
  github.com/thrasher-corp/gocryptotrader/exchanges/binance.(*Binance).GetAuthenticatedAPISupport()
      <autogenerated>:1 +0x44
  github.com/thrasher-corp/gocryptotrader/engine.(*OrderManager).processOrders()
      C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/order_manager.go:577 +0x195
  github.com/thrasher-corp/gocryptotrader/engine.(*OrderManager).run()
      C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/order_manager.go:100 +0x99
  github.com/thrasher-corp/gocryptotrader/engine.(*OrderManager).Start·dwrap·27()
      C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/order_manager.go:63 +0x39

Goroutine 205 (running) created at:
  testing.(*T).Run()
      C:/Program Files/Go/src/testing/testing.go:1306 +0x726
  testing.runTests.func1()
      C:/Program Files/Go/src/testing/testing.go:1598 +0x99
  testing.tRunner()
      C:/Program Files/Go/src/testing/testing.go:1259 +0x22f
  testing.runTests()
      C:/Program Files/Go/src/testing/testing.go:1596 +0x7ca
  testing.(*M).Run()
      C:/Program Files/Go/src/testing/testing.go:1504 +0x9d1
  main.main()
      _testmain.go:393 +0x22b

Goroutine 11 (running) created at:
  github.com/thrasher-corp/gocryptotrader/engine.(*OrderManager).Start()
      C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/order_manager.go:63 +0x164
  github.com/thrasher-corp/gocryptotrader/engine.TestGetOrders()
      C:/Users/gloriousCode/go/src/github.com/thrasher-corp/gocryptotrader/engine/rpcserver_test.go:1005 +0xb46
  testing.tRunner()
      C:/Program Files/Go/src/testing/testing.go:1259 +0x22f
  testing.(*T).Run·dwrap·21()
      C:/Program Files/Go/src/testing/testing.go:1306 +0x47
==================
--- FAIL: TestGetOrders (0.00s)
    testing.go:1152: race detected during execution of test
```
I'm thinking order manager is starting and since things are done asynchrounously, its updating them in tests at the same time processOrders is occurring

engine/order_manager.go
TaltaM: OK I see what you mean. But because that same aspiring trader would most likely be active on several exchanges, I'm sure he'd like things to run in parallel. I'll try with a WaitGroup.

TaltaM:
OK, good that we continued with the testing.. processOrders() used REST response data to mark which orders did not need to be checked again, but the REST responses do obviously not have the internal order ID's. Should be fixed now!

engine/order_manager_test.go
```diff
@@ -702,8 +744,156 @@ func TestOrderManager_Modify(t *testing.T) {
 }
 
 func TestProcessOrders(t *testing.T) {
-	m := OrdersSetup(t)
+	log.Warnf(log.OrderMgr, "TestProcessOrders - set up orders starting")
```
gloriousCode, line 747: please remove

engine/order_manager.go
```diff
@@ -641,22 +640,24 @@ func (m *OrderManager) processOrders() {
...
+			wg.Add(1)
+			go m.processMatchingOrders(exchanges[i], orders, requiresProcessing, &wg)
 		}
 	}
+	wg.Wait()
 }
 
-func (m *OrderManager) processMatchingOrders(exch exchange.IBotExchange, orders []order.Detail, requiresProcessing map[string]bool) {
+func (m *OrderManager) processMatchingOrders(exch exchange.IBotExchange, orders []order.Detail, requiresProcessing map[string]bool, wg *sync.WaitGroup) {
+	defer wg.Done()
```
gloriousCode, line 660: A nil check on the `wg` will ensure no panics

gloriousCode (approved):
tACK! Thanks for those updates! 
The two extra comments are optional fixes, though if another reviewer has issue with these/something else comes up, please address them.

gloriousCode (changes requested):
Seeing you have merge conflicts to resolve, please address the two comments above as well, but otherwise its good to me 🎉

TaltaM:
Comments addressed, and rebased!

```
[INFO]  | 13/09/2021 12:32:19 | Order manager: Exchange Bittrex added order ID=724ced58-efdc-40c4-bb64-134cf418d775 internal ID=06662e91-1fbf-4580-b0a1-2e60e6d2779c pair=VITE-BTC price=0.00000161 amount=300.00000000 side=BUY type=LIMIT status=ACTIVE.
[DEBUG]  | 13/09/2021 12:32:29 | Order manager: Exchange Bittrex updated order ID=724ced58-efdc-40c4-bb64-134cf418d775 internal ID=06662e91-1fbf-4580-b0a1-2e60e6d2779c pair=VITE-BTC price=0.00000161 amount=300.00000000 side=BUY type=LIMIT status=ACTIVE.
[DEBUG]  | 13/09/2021 12:32:39 | Order manager: Exchange Bittrex updated order ID=724ced58-efdc-40c4-bb64-134cf418d775 internal ID=06662e91-1fbf-4580-b0a1-2e60e6d2779c pair=VITE-BTC price=0.00000161 amount=300.00000000 side=BUY type=LIMIT status=ACTIVE.
[DEBUG]  | 13/09/2021 12:32:49 | Order manager: Exchange Bittrex updated order ID=724ced58-efdc-40c4-bb64-134cf418d775 internal ID=06662e91-1fbf-4580-b0a1-2e60e6d2779c pair=VITE-BTC price=0.00000161 amount=300.00000000 side=BUY type=LIMIT status=ACTIVE.
[DEBUG]  | 13/09/2021 12:33:00 | Order manager: Exchange Bittrex updated order ID=724ced58-efdc-40c4-bb64-134cf418d775 internal ID=06662e91-1fbf-4580-b0a1-2e60e6d2779c pair=VITE-BTC price=0.00000161 amount=300.00000000 side=BUY type=LIMIT status=CANCELLED.
```

gloriousCode (approved):
tACK! Thanks for making those changes and thanks for your contribution!

shazbert:
@TaltaM [For this comment](https://github.com/thrasher-corp/gocryptotrader/pull/775#issuecomment-915012232) My mistake, this was a an issue with my configuration on my part. Thank you very much for the explanation. I am sorry I wasted a bit of your time.

shazbert (approved):
Tested approved! Thank you for putting a lot of work into this.

thrasher- (changes requested):
Just some minor nits and then this looks good to go. Thanks for the PR @TaltaM and for addressing the feedback 👍

TaltaM:
> Just some minor nits and then this looks good to go. Thanks for the PR @TaltaM and for addressing the feedback +1

Done!
Thanks for being thorough :)

thrasher- (approved):
Thanks for making those changes @TaltaM !

PR #802, opened 2021-10-09
binance: order history use request pair

khcchiu:
# PR Description

When retrieving orders from Binance for pair AVAX/USDT (or like RAMPUSDT), the response is returning AVA/XUSDT (or RAM/PUSDT) instead. Changed to use request pair to make order history work as expected first.

But the problem lies deeper in:

```
// NewPairFromString converts currency string into a new CurrencyPair
// with or without delimeter
func NewPairFromString(currencyPair string) (Pair, error) {
	for x := range delimiters {
		if strings.Contains(currencyPair, delimiters[x]) {
			return NewPairDelimiter(currencyPair, delimiters[x])
		}
	}
	if len(currencyPair) < 3 {
		return Pair{},
			fmt.Errorf("cannot create pair from %s string",
				currencyPair)
	}
	return NewPairFromStrings(currencyPair[0:3], currencyPair[3:])
}
```

Wondering if there's any reason to split the pair string at index 3? Is it simply because back in 2019 coins are usually with 3 characters? This could be a wider problem as e.g. AVA and AVAX are both trading in large exchanges, (another problem is there's also coins like RAM,RAMP, PUSD, USD, thus RAMPUSD could actually represent two pairs depending splitting at which index, a more sophisticated NewPairFromString() is needed).

Fixes # Pair being incorrectly parsed in Binance spot order history

## Type of change

## How has this been tested

- [x] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/binance/binance_wrapper.go
+++ b/exchanges/binance/binance_wrapper.go
@@ -1403,10 +1403,6 @@ func (b *Binance) GetOrderHistory(ctx context.Context, req *order.GetOrdersReque
 					continue
 				}
 
-				pair, err := currency.NewPairFromString(resp[i].Symbol)
-				if err != nil {
-					return nil, err
-				}
 				orders = append(orders, order.Detail{
 					Amount:   resp[i].OrigQty,
 					Date:     resp[i].Time,
@@ -1415,7 +1411,7 @@ func (b *Binance) GetOrderHistory(ctx context.Context, req *order.GetOrdersReque
 					Side:     orderSide,
 					Type:     orderType,
 					Price:    resp[i].Price,
-					Pair:     pair,
+					Pair:     req.Pairs[x],
 					Status:   order.Status(resp[i].Status),
 				})
 			}


thrasher-:
Thanks for the PR @khcchiu ! Yeah, originally a lot of exchanges initially were a length of 6, so `NewPairFromString` would suffice. However, as time went on pairs had varying lengths so we now opt by default to store currency pairs with a delimiter where possible (since config loading has to know how to load a pair correctly) and then that also has to conform on the exchange side in terms of requests an responses. We have methods to help assist with matching up pairs provided by the exchange which don't have a delimiter - https://github.com/thrasher-corp/gocryptotrader/blob/master/currency/pair.go#L101 . Sadly in this case, `NewPairFromString` shouldn't of been used here and instead `currency.NewPairFromFormattedPairs` should have been used. You can see an example in this PR: https://github.com/thrasher-corp/gocryptotrader/pull/762/files of the correct usage.

Furthermore, I'll be introducing a PR which performs a config upgrade for Binance to store pairs with a delimiter for the ones which haven't. This PR: https://github.com/thrasher-corp/gocryptotrader/pull/794 has a fix for correct ticker handling on Binance with the style of pairs you've mentioned

PR #821, opened 2021-11-05
binance: don't enable BTC/USDT instrument without user's knowledge

lrascao:
If not instruments were enabled refrain from trying to be clever and
auto-enabling BTC/USDT.

- [X] go test ./... -race
- [X] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/binance/binance_wrapper.go
+++ b/exchanges/binance/binance_wrapper.go
@@ -264,60 +264,9 @@ func (b *Binance) Run() {
 		b.PrintEnabledPairs()
 	}
 
-	forceUpdate := false
-	format, err := b.GetPairFormat(asset.Spot, false)
-	if err != nil {
-		log.Errorf(log.ExchangeSys, "%s failed to get enabled currencies. Err %s\n",
-			b.Name,
-			err)
-		return
-	}
-	pairs, err := b.GetEnabledPairs(asset.Spot)
-	if err != nil {
-		log.Errorf(log.ExchangeSys, "%s failed to get enabled currencies. Err %s\n",
-			b.Name,
-			err)
-		return
-	}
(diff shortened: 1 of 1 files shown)


exchanges/binance/binance_wrapper.go
```diff
@@ -264,59 +264,10 @@ func (b *Binance) Run() {
 	}
 
 	forceUpdate := false
```
shazbert, line 266: you can remove this as well.

shazbert (changes requested):
Just one thing. I think time has sufficiently passed to warrant this removal.

thrasher-:
NOTE: This is currently the way we perform config pair format upgrades whenever a delimiter is required to parse pairs properly. See https://github.com/thrasher-corp/gocryptotrader/pull/762 as an example. By default, this behaviour is enabled and provides warnings to the user whenever such a change to the config takes place. Although this one was a fix introduced years ago, there will be future ones made. A config setting moving forward would be better for users who don't require this functionality, but for an out of box experience we want it to be as pain free for users as possible when it comes to things like this enabled by default. Another thing to consider is the config loader will assign a random pair if a user has enabled an asset, but hasn't filled in any enabled pairs on GCT start-up.

shazbert (approved):
Changes ACK.

gloriousCode (approved):
This scenario seemingly can only come up if you're purposely storing pairs in a different format. Using `config_example.json`, (our basic backwards-compatible config) doesn't trigger this piece of code. So in that sense its fine to remove this specific instance of config pair storage fixing as it doesn't impact config backwards compatibility. tACK.

thrasher- (approved):

PR #836, opened 2021-11-11
Binance: Fixes expiring contract formatting

gloriousCode:
# PR Description
I noticed this when trying to do a futures test run for the GoCryptoTrader Backtester that GoCryptoTrader could not format requests for Binance expiring USDTMarginedFutures  contracts. We don't store the currency with an underscore because our config & request format do not have delimiters. Even if we do have underscores stored, when we make a request, formatting the pair removes it. 

For example, the pair `BTCUSDT_211231` requests as `BTCUSDT211231` due to all other perpetual contract formats not having an underscore.

This issue was fun because so many other Binance functions format pairs, so I overrode the default `Base` functions to allow the delimiter to remain

## Type of change

## How has this been tested
- [x] `go test ./... -race`
- [x] `golangci-lint run`
- [x] `TestGetEnabledPairs`
- [x] `TestFormatExchangeCurrency`
- [x] `TestFormatSymbol`
- [x] Running the application and noting a lack of "invalid symbol" errors
- [x] ` .\gctcli.exe getticker binance btcusdt-211231 usdtmarginedfutures`
- [x]  `.\gctcli.exe getticker binance btc-usdt usdtmarginedfutures`
- [x]  `.\gctcli.exe getticker binance lto-usdt margin`
- [x]  ` .\gctcli.exe getticker binance BTC-usdt spot`
- [x]  ` .\gctcli.exe getticker binance BTCUSD-PERP coinmarginedfutures`
- [x] `.\gctcli.exe getticker binance BTCUSD-211231 coinmarginedfutures`

## Checklist


exchanges/binance/binance_wrapper.go
```diff
@@ -387,11 +387,21 @@ func (b *Binance) FetchTradablePairs(ctx context.Context, a asset.Item) ([]strin
 		}
 		for u := range uInfo.Symbols {
 			if uInfo.Symbols[u].Status == "TRADING" {
-				curr, err := currency.NewPairFromString(uInfo.Symbols[u].Symbol)
-				if err != nil {
-					return nil, err
+				var curr currency.Pair
+				if strings.Contains(uInfo.Symbols[u].Symbol, currency.UnderscoreDelimiter) {
+					index := strings.Index(uInfo.Symbols[u].Symbol, currency.UnderscoreDelimiter)
```
shazbert, line 392: Might be able to strings.Split?

exchanges/binance/binance_wrapper.go
```diff
@@ -387,11 +387,21 @@ func (b *Binance) FetchTradablePairs(ctx context.Context, a asset.Item) ([]strin
...
+					curr, err = currency.NewPairFromStrings(uInfo.Symbols[u].Symbol[0:index], uInfo.Symbols[u].Symbol[index:])
```
shazbert, line 393: if not `uInfo.Symbols[u].Symbol[0:index]` -> `uInfo.Symbols[u].Symbol[:index]`

exchanges/binance/binance_wrapper.go
```diff
@@ -387,11 +387,21 @@ func (b *Binance) FetchTradablePairs(ctx context.Context, a asset.Item) ([]strin
...
+					if err != nil {
+						return nil, err
+					}
+					pairs = append(pairs, format.Format(curr))
```
shazbert, line 397: can move this append outside of the if else scope and remove the other one.

exchanges/binance/binance_wrapper.go
```diff
@@ -1793,3 +1803,68 @@ func (b *Binance) GetAvailableTransferChains(ctx context.Context, cryptocurrency
...
+			}
+			resp = append(resp, pair)
+		}
+		return resp, nil
+	}
+
+	return enabledPairs.Format(format.Delimiter,
+			format.Index,
+			format.Uppercase),
+		nil
```
shazbert, line 1842:
```go
	enabledPairs, err := b.CurrencyPairs.GetPairs(a, true)
	if err != nil {
		return nil, err
	}
	format, err := b.GetPairFormat(a, false)
	if err != nil {
		return nil, err
	}

	fmted := enabledPairs.Format(format.Delimiter, format.Index, format.Uppercase)
	if a == asset.USDTMarginedFutures {
		for i := range fmted {
			if fmted[i].Delimiter == currency.UnderscoreDelimiter {
				// we cannot remove expiring contract delimiters from requests
				fmted[i].Delimiter = currency.UnderscoreDelimiter
			}
		}
	}
	return fmted, nil
```

Format selected and then add the delimiter.

shazbert (changes requested):
Looks good! Just some nits.

exchanges/binance/binance_wrapper.go
gloriousCode: By doing a split, we lose the delimiter. I'll go with your other option

exchanges/binance/binance_wrapper.go
gloriousCode: That would remove the delimiter, there'd be no way of knowing if there was an underscore delimiter to keep it. The issue is that for USDTMarginedFutures, we need to have an optional delimiter for the asset type. For all other USDTMarginedFutures pairs, we don't have a delimiter and Binance would reject it. For the specific currencies that do, it requires the underscore.

It does highlight that my test `TestGetEnabledPairs` wasn't adequate enough as it passes despite this change, but the currency has no delimiter

shazbert: 👍

exchanges/binance/binance_wrapper.go
```diff
@@ -387,9 +387,18 @@ func (b *Binance) FetchTradablePairs(ctx context.Context, a asset.Item) ([]strin
 		}
 		for u := range uInfo.Symbols {
 			if uInfo.Symbols[u].Status == "TRADING" {
```
shazbert, line 389: OOS: invert comparison and continue; also ones above.

exchanges/binance/binance_wrapper.go
```diff
@@ -387,9 +387,18 @@ func (b *Binance) FetchTradablePairs(ctx context.Context, a asset.Item) ([]strin
...
-				curr, err := currency.NewPairFromString(uInfo.Symbols[u].Symbol)
-				if err != nil {
-					return nil, err
+				var curr currency.Pair
+				if strings.Contains(uInfo.Symbols[u].Symbol, currency.UnderscoreDelimiter) {
+					index := strings.Index(uInfo.Symbols[u].Symbol, currency.UnderscoreDelimiter)
+					curr, err = currency.NewPairFromStrings(uInfo.Symbols[u].Symbol[:index], uInfo.Symbols[u].Symbol[index:])
+					if err != nil {
+						return nil, err
+					}
```
shazbert, line 396: can move this and the one below out of the if else scope.

shazbert (changes requested):
tested looks good, pedantic nits.

shazbert (approved):
Changes ACK.

thrasher- (changes requested):
Thanks for this PR! This fixes one issue with the long dated USDT coin margin futures but still highlights another problem with it introduced in https://github.com/thrasher-corp/gocryptotrader/pull/530 where there's no delimiter to separate the pairs. This means that when GCT loads a Binance USDT margin future perp pair which has > 3 characters for the base, e.g DOGE USDT, it will still be handled internally as DOG EUSDT. You can see this behaviour with this PR by printing a pairs base and quote when calling `UpdateTicker`

Since this PR aims to address a pair formatting issue, I think a better solution would be to:

1) Check all USDT margin futures pairs in Run to see if they contain a "_". If not, force a config upgrade with a delimiter like how we do it with other exchanges.
2) You can easily identify which pairs are a long dated futures contract (currently only 2 of them by the quote being all digits, or use the FAPI UExchangeInfo endpoint and match up the pair with the contract being != "perpetual". Aside from the perps, there's only quarterly contracts available currently. So if the pair is a long dated futures contact, use the "_" in the request. Otherwise omit it. 

The above will fix both existing issues

gloriousCode:
- I've updated the config format to use an underscore delimiter, I only add the underscore to the request if the `Quote` is all number. 
  - `ETHUSDT_211231`: `[INFO]  | TICKER |  16/11/2021 08:51:13  | Binance REST ethusdt211231 USDTMARGINEDFUTURES: TICKER: Last 4685.44000000 Ask 0.00000000 Bid 0.00000000 High 4891.06000000 Low 4653.18000000 Volume 21212.35500000`
  - `DOGE_USDT`: `[INFO]  | TICKER |  16/11/2021 08:52:16  | Binance REST dogeusdt USDTMARGINEDFUTURES: TICKER: Last 0.25840000 Ask 0.00000000 Bid 0.00000000 High 0.26541000 Low 0.25531000 Volume 1829996253.00000000`
- I've added the full list of fields for the USDTMFutures ExchangeInfo struct.
- I've added a fancy new `bypassConfigFormatUpgrades` field to the `currencyPairs` config struct and to the `exchange.go Base` which allows a user to bypass config format upgrades to cater to https://github.com/thrasher-corp/gocryptotrader/pull/821

shazbert (changes requested):
Just a couple of nits below from testing, Which can be replicated across the other exchanges.

shazbert (approved):
[Changes ACK](https://youtu.be/6qJ1E8EljVM) 👨‍🎤 
```
 Binance Updating available pairs [SPOT] - Added: [ETH-BTC LTC-BTC BNB-BTC NEO-BTC QTUM-ETH EOS-ETH SNT-ETH BNT-ETH GAS-BTC BNB-ETH BTC-USDT ETH-USD.....
```

thrasher- (approved):
tACK! Thanks for making those changes:

Config upgrade from master to this PR
image

Correct pair handling:
image

PR #845, opened 2021-11-23
log: expose output writer interface

lrascao:
This allows applications that import GCT to define their own
writer on their own terms.

## Type of change

## How has this been tested

- [X] go test ./... -race
- [X] golangci-lint run

## Checklist

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -19,10 +19,18 @@ func NewSubLogger(name string) (*SubLogger, error) {
 		return nil, errEmptyLoggerName
 	}
 	name = strings.ToUpper(name)
-	if _, ok := subLoggers[name]; ok {
+	if _, ok := SubLoggers[name]; ok {
 		return nil, errSubLoggerAlreadyregistered
 	}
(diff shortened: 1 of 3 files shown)


log/logger_types.go
```diff
@@ -90,7 +90,7 @@ type Levels struct {
 type SubLogger struct {
 	name string
 	Levels
-	output io.Writer
+	Output io.Writer
```
shazbert, line 93: Would it be possible instead to add a method to set a new io.Writer? Just so we can use the `RWM = &sync.RWMutex{}` in `logger_types.go` and make a bit more protected?

shazbert (changes requested):
Just one suggestion, let me know what you think.

log/logger_types.go
lrascao: sure, sounds good

shazbert (approved):
Thanks for making that change! ACK.

thrasher- (approved):

thrasher-:
Thanks @lrascao

PR #848, opened 2021-11-24
request: provide observability over HTTP request latency

lrascao:
# PR Description

This PR allows applications to set an optional global reporter that will provide observability over latency of all HTTP requests (eg. through Prometheus). It also allows a per-exchange setting that is overridden by a global one.

## Type of change

## How has this been tested

- [X] go test ./... -race
- [X] golangci-lint run

## Checklist

### DIFF
--- a/exchanges/request/report.go
+++ b/exchanges/request/report.go
@@ -1,0 +1,17 @@
+package request
+
+import (
+	"time"
+)
+
+// Reporter interface groups observability functionality over
+// HTTP request latency.
+type Reporter interface {
+	Latency(name, method, path string, t time.Duration)
+}
+
+// SetupGlobalReporter sets a reporter interface to be used
+// for all exchange requests
+func SetupGlobalReporter(r Reporter) {
+	globalReporter = r
+}
--- a/exchanges/request/request.go
+++ b/exchanges/request/request.go
@@ -39,6 +39,7 @@ func New(name string, httpRequester *http.Client, opts ...RequesterOption) *Requ
 		retryPolicy: DefaultRetryPolicy,
 		maxRetries:  MaxRetryAttempts,
 		timedLock:   timedmutex.NewTimedMutex(DefaultMutexLockTimeout),
+		reporter:    globalReporter,
 	}
 
 	for _, o := range opts {
@@ -150,7 +151,14 @@ func (r *Requester) doRequest(ctx context.Context, endpoint EndpointLimit, newRe
 			}
 		}
 
+		start := time.Now()
+
 		resp, err := r.HTTPClient.Do(req)
+
+		if r.reporter != nil {
+			r.reporter.Latency(r.Name, p.Method, p.Path, time.Since(start))
+		}
+
 		if retry, checkErr := r.retryPolicy(resp, err); checkErr != nil {
 			return checkErr
 		} else if retry {
--- a/exchanges/request/request_types.go
+++ b/exchanges/request/request_types.go
@@ -23,12 +23,14 @@ const (
 var (
 	MaxRequestJobs   = DefaultMaxRequestJobs
 	MaxRetryAttempts = DefaultMaxRetryAttempts
+	globalReporter   Reporter
 )
 
 // Requester struct for the request client
 type Requester struct {
 	HTTPClient         *http.Client
 	limiter            Limiter
+	reporter           Reporter
 	Name               string
 	UserAgent          string
 	maxRetries         int
--- a/exchanges/request/options.go
+++ b/exchanges/request/options.go
@@ -20,3 +20,10 @@ func WithRetryPolicy(p RetryPolicy) RequesterOption {
 		r.retryPolicy = p
 	}
 }
+
+// WithReporter configures the reporter for a Requester.
+func WithReporter(rep Reporter) RequesterOption {
+	return func(r *Requester) {
+		r.reporter = rep
+	}
+}


exchanges/request/report.go
```diff
@@ -0,0 +1,27 @@
+package request
+
+import (
+	"time"
+
+	"github.com/thrasher-corp/gocryptotrader/log"
+)
+
+type Reporter interface {
+	Latency(name, method, path string, t time.Duration)
```
shazbert, line 10: Can we add in a verbose param? This is so we can omit the basic reporter output and it won't really affect any custom reporters or the global if not defined that way.

shazbert (changes requested):
This is a great start to see request times.

exchanges/request/report.go
lrascao: sounds good, i ended up doing the same thing but slightly different, looked a bit weird to pass a verbose flag to the interface method just for the sake of `DefaultReporter`

exchanges/request/report.go
```diff
@@ -0,0 +1,32 @@
+package request
+
+import (
+	"time"
+
+	"github.com/thrasher-corp/gocryptotrader/log"
+)
+
+type Reporter interface {
```
shazbert, line 9: Please add comment

shazbert (approved):
The changes you made are much better than my suggestion. 👍 Just one pedantic nit and it looks good to go from me.

shazbert (approved):
Thanks, tACK!

lrascao:
rebased onto latest master

exchanges/request/request.go
```diff
@@ -150,7 +150,22 @@ func (r *Requester) doRequest(ctx context.Context, endpoint EndpointLimit, newRe
...
+		if reporter == nil {
+			// no global reporter defined, fallback to the request's
+			reporter = r.reporter
+			if reporter == nil {
+				// no custom reporter as well, fallback to the default
+				// that will look at the Verbose flag to decide wether
+				// to log or not
+				reporter = DefaultReporter(p.Verbose)
+			}
+		}
```
gloriousCode, line 163: This seems like it would be better to do this at a setup/requester setup. When no reporters are defined, we are doing multiple checks and creating structs on every request.

exchanges/request/request_types.go
```diff
@@ -23,12 +23,14 @@ const (
 var (
 	MaxRequestJobs   = DefaultMaxRequestJobs
 	MaxRetryAttempts = DefaultMaxRetryAttempts
+	GlobalReporter   Reporter
```
gloriousCode, line 26: I would prefer this to be unexported and you instead have a setup function

exchanges/request/request.go
```diff
@@ -150,7 +150,22 @@ func (r *Requester) doRequest(ctx context.Context, endpoint EndpointLimit, newRe
...
+
+		start := time.Now()
 		resp, err := r.HTTPClient.Do(req)
+		reporter.Latency(r.Name, p.Method, p.Path, time.Since(start))
```
gloriousCode, line 167: This reporting is optional behavior and I would prefer that things only are run if it is enabled

gloriousCode (changes requested):
Just a few things from me.

exchanges/request/request.go
lrascao: i ended up dropping the DefaultReporter to simplify this logic, if an exchange defines a reporter then that takes precedence over a globally defined one as well.

exchanges/request/request_types.go
lrascao: `request.SetupGlobalReporter` as been added to this end

exchanges/request/request.go
lrascao: `Latency` now will be done only when there's a defined reporter (either global or exchange scoped)

exchanges/request/report.go
```diff
@@ -0,0 +1,15 @@
...
+	"time"
+)
+
+// Reporter interface groups observability functionality over
+// HTTP request latency.
+type Reporter interface {
+	Latency(name, method, path string, t time.Duration)
+}
+
+func SetupGlobalReporter(r Reporter) {
```
shazbert, line 13: comment please

exchanges/request/request.go
```diff
@@ -150,7 +150,20 @@ func (r *Requester) doRequest(ctx context.Context, endpoint EndpointLimit, newRe
 			}
 		}
 
+		reporter := r.reporter
+		if reporter == nil {
+			// no request reporter defined, fallback to global
+			reporter = globalReporter
+		}
```
shazbert, line 157: If globalReporter is a fall back, maybe it might be better setting the `Global Reporter` in New() and then override with the requester option? Unless you want to assign and change the global reporter at different times, then it might need a RWMutex for protection on the `globalReporter var`

shazbert (changes requested):
Thanks for updating, one small change, going off the comments by @gloriousCode. Let us know what you think.

gloriousCode (approved):
Thanks for making all of those changes! The setup allows for validation + protection if we pursue this feature further.

I setup a global one:
`[DEBUG] | LOG |  06/12/2021 09:32:44  | FTX GET https://ftx.com/api/wallet/deposit_address/BTC?method=sol 261.9964ms`

As well as one specifically for binance:
`[DEBUG] | EXCHANGE |  06/12/2021 09:33:02  | HELLO I AM LOCAL Binance GET https://fapi.binance.com/fapi/v1/ticker/24hr? 140.2946ms`

tACK!

exchanges/request/request.go
lrascao: not quite sure if i got your idea, the use-case so far for an application that imports `gct` is to globally define a global reporter that applies to all exchanges that didn't ask for a specific reporter (all so far). We also want this behaviour to be disabled by default so as not to incur in additional request processing.

shazbert:
```diff
		@@ -39,6 +39,7 @@ func New(name string, httpRequester *http.Client, opts ...RequesterOption) *Requ
                retryPolicy: DefaultRetryPolicy,
                maxRetries:  MaxRetryAttempts,
                timedLock:   timedmutex.NewTimedMutex(DefaultMutexLockTimeout),
+               reporter:    globalReporter,
        }

        for _, o := range opts {
@@ -150,18 +151,12 @@ func (r *Requester) doRequest(ctx context.Context, endpoint EndpointLimit, newRe
                        }
                }

-               reporter := r.reporter
-               if reporter == nil {
-                       // no request reporter defined, fallback to global
-                       reporter = globalReporter
-               }
-
                start := time.Now()

                resp, err := r.HTTPClient.Do(req)

-               if reporter != nil {
-                       reporter.Latency(r.Name, p.Method, p.Path, time.Since(start))
+               if r.reporter != nil {
+                       r.reporter.Latency(r.Name, p.Method, p.Path, time.Since(start))
                }

                if retry, checkErr := r.retryPolicy(resp, err); checkErr != nil {
```

So add the global reporter to new, which will be overridden by the requester option (if passed in), I am assuming we set up the global report prior to getting a new requester though. If this is not the case then disregard it. But this will just need a !nil comparative instead of var declaration and then checking that and then assigning the global. Let me know what you think.

lrascao: ah right, got it now, that is indeed cleaner.

>. I am assuming we set up the global report prior to getting a new requester though.

that is exactly the case

shazbert (approved):
tested approved, thanks for making those changes. I set a global reporter and it worked and then I set a custom reporter on an exchange and that overrides the global. 🚀

gloriousCode (approved):
tACK! Thanks for making those last changes!

thrasher- (approved):
Nice feature! tACK:

image

PR #858, opened 2021-12-15
engine/okgroup: fix panic when calling GetDepositAddress()

shazbert:
# PR Description

Repairs panic when no wallet information was returned via `GetDepositAddress` and a nil address was returned with a nil error. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/helpers.go
+++ b/engine/helpers.go
@@ -691,10 +691,7 @@ func (bot *Engine) GetCryptocurrencyDepositAddressesByExchange(exchName string) 
 func (bot *Engine) GetExchangeCryptocurrencyDepositAddress(ctx context.Context, exchName, accountID, chain string, item currency.Code, bypassCache bool) (*deposit.Address, error) {
 	if bot.DepositAddressManager != nil && bot.DepositAddressManager.IsSynced() && !bypassCache {
 		resp, err := bot.DepositAddressManager.GetDepositAddressByExchangeAndCurrency(exchName, chain, item)
-		if err != nil {
-			return nil, err
-		}
-		return &resp, nil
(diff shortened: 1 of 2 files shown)


engine/helpers.go
```diff
@@ -782,6 +782,12 @@ func (bot *Engine) GetAllExchangeCryptocurrencyDepositAddresses() map[string]map
 							err)
 						continue
 					}
+					if depositAddr == nil {
+						log.Errorf(log.Global, "%s failed to get cryptocurrency deposit address for %s. Err: Deposit address came back as nil.\n",
+							exchName,
+							cryptocurrency)
+						continue
+					}
```
gloriousCode, line 790: I'm not sure we should be adding this section. While the panic was unfortunate, it did highlight a bug in our code, where this would just hide it unless you are looking over logs. Typically for functions across our codebase that have pointer responses, we rely on errors to highlight any issues, including if `nil` is a problem.

I'm not totally against this addition though, but if you are to keep it, you'll need to add it to the two calls to `GetDepositAddress` above

gloriousCode (changes requested):
Good spotting! 👀

engine/helpers.go
shazbert: I eyeballed our implementations across wrapper files and we haven't done any other mixing from what I can see. I will remove this check.

gloriousCode (approved):
tACK! Thanks for making that change

thrasher- (approved):
Nice catch, ACK!

PR #869, opened 2022-01-10
exchanges: make ErrPairNotFound an exportable error

lrascao:
Stop erroring out when receiving execution reports from currency pairs
that have not been enabled. In the use-case where multiple services consume market data from different currency pairs on the same account it happens that eg. a service handling ETH/USDT gets execution reports from a separate service handling BTC/USDT on the same account, no big side effects but error rates increase somewhat damaging observability effectiveness.

## Type of change

## How has this been tested

- [X] go test ./... -race
- [X] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -47,6 +47,9 @@ var (
 
 	errEndpointStringNotFound = errors.New("endpoint string not found")
 	errTransportNotSet        = errors.New("transport not set, cannot set timeout")
+
+	// ErrPairNotFound is an error message for when unable to find a currency pair
+	ErrPairNotFound = errors.New("pair not found")
 )
 
 func (b *Base) checkAndInitRequester() {
@@ -458,7 +461,8 @@ func (b *Base) GetRequestFormattedPairAndAssetType(p string) (currency.Pair, ass
 			}
 		}
 	}
-	return response, "", errors.New("pair not found: " + p)
+	return response, "",
+		fmt.Errorf("%s %w", p, ErrPairNotFound)
 }
 
 // GetAvailablePairs is a method that returns the available currency pairs


exchanges/exchange.go
```diff
@@ -47,6 +47,8 @@ var (
 
 	errEndpointStringNotFound = errors.New("endpoint string not found")
 	errTransportNotSet        = errors.New("transport not set, cannot set timeout")
+
+	ErrPairNotFound = errors.New("pair not found")
```
shazbert, line 51: Comment please or un-export.

exchanges/exchange.go
```diff
@@ -458,7 +460,7 @@ func (b *Base) GetRequestFormattedPairAndAssetType(p string) (currency.Pair, ass
 			}
 		}
 	}
-	return response, "", errors.New("pair not found: " + p)
+	return response, "", ErrPairNotFound
```
shazbert, line 463: Please wrap for other calls outside of scope. 
```go
return response, "", fmt.Errorf("%s %w", p, ErrPairNotFound)
```

shazbert (changes requested):
Thanks for opening this PR, I am thinking maybe in this instance we can add in a function specifically for this use case that uses Available Pairs instead of Enabled Pairs. So we don't suppress errors, the other changes can still occur. Let me know if this will cause issues on your end.

exchanges/exchange.go
lrascao: > Thanks for opening this PR, I am thinking maybe in this instance we can add in a function specifically for this use case that uses Available Pairs instead of Enabled Pairs. So we don't suppress errors, the other changes can still occur. Let me know if this will cause issues on your end.

Bit on the fence about using `AvailablePairs`, nice thing about the current setup is that we spin up the GCT-wrapping service without the need for knowledge of unrelated pairs, if we want it to handle only BTC and ETH that's really only it needs to know, on startup the service tells GCT that it only cares about BTC/ETH and nothing else.

exchanges/exchange.go
lrascao: this is actually being used [here](https://github.com/thrasher-corp/gocryptotrader/pull/869/files#diff-e17060c728c86d82e2bb7909ca3677a3af1c272d317f799dbe9a3c37dc13c2f1R232) to actively ignore this specific error and none other

gloriousCode (changes requested):

exchanges/binance/binance_websocket.go
```diff
@@ -227,6 +228,11 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 				remainingAmount := data.Data.Quantity - data.Data.CumulativeFilledQuantity
 				pair, assetType, err := b.GetRequestFormattedPairAndAssetType(data.Data.Symbol)
 				if err != nil {
+					// ignore execution reports from disabled pairs instead of erroring out
+					if errors.Is(err, exchange.ErrPairNotFound) {
+						return nil
+					}
```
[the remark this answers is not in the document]
lrascao: yeah, that's a fair point, i agree. I'll remove this check and move it over to our end.

gloriousCode (changes requested):
Thank you for making that change! Please add a comment to `ErrPairNotFound` and its all good from me 🎉

gloriousCode (approved):
tACK! Thanks for understanding and making that change

thrasher- (approved):
Thanks @lrascao

shazbert (approved):
Changes ACK.

PR #935, opened 2022-04-28
engine: Add websocket data handler register function

shazbert:
# PR Description

Allows to easily register a data handler method for websocket data if needed for external strategy inspection. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -948,3 +948,13 @@ func (bot *Engine) SetupExchanges() error {
 func (bot *Engine) WaitForInitialCurrencySync() error {
 	return bot.currencyPairSyncer.WaitForInitialSync()
 }
+
+// RegisterWebsocketRoutineHandlerInterceptor registers an externally defined
+// interceptor function for diverting and handling websocket notifications
+// across all enabled exchanges.
+func (bot *Engine) RegisterWebsocketRoutineHandlerInterceptor(fn Interceptor) error {
+	if bot == nil {
+		return fmt.Errorf("%T %w", bot, ErrNilSubsystem)
+	}
+	return bot.websocketRoutineManager.registerInterceptor(fn)
+}
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -949,12 +949,13 @@ func (bot *Engine) WaitForInitialCurrencySync() error {
 	return bot.currencyPairSyncer.WaitForInitialSync()
 }
 
-// RegisterWebsocketRoutineHandlerInterceptor registers an externally defined
-// interceptor function for diverting and handling websocket notifications
-// across all enabled exchanges.
-func (bot *Engine) RegisterWebsocketRoutineHandlerInterceptor(fn Interceptor) error {
+// RegisterWebsocketDataHandler registers an externally defined data handler
+// for diverting and handling websocket notifications across all enabled
+// exchanges. InterceptorOnly as true will purge all other registered handlers
+// (including default) bypassing all other handling.
+func (bot *Engine) RegisterWebsocketDataHandler(fn WebsocketDataHandler, interceptorOnly bool) error {
 	if bot == nil {
 		return errNilBot
 	}
-	return bot.websocketRoutineManager.registerInterceptor(fn)
+	return bot.websocketRoutineManager.registerInterceptor(fn, interceptorOnly)
 }
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -959,3 +959,12 @@ func (bot *Engine) RegisterWebsocketDataHandler(fn WebsocketDataHandler, interce
 	}
 	return bot.websocketRoutineManager.registerWebsocketDataHandler(fn, interceptorOnly)
 }
+
+// SetDefaultWebsocketDataHandler sets the default websocket handler and
+// removing all pre-existing handlers
+func (bot *Engine) SetDefaultWebsocketDataHandler() error {
+	if bot == nil {
+		return errNilBot
+	}
+	return bot.websocketRoutineManager.setWebsocketDataHandler(bot.websocketRoutineManager.websocketDataHandler)
+}
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -957,5 +957,5 @@ func (bot *Engine) RegisterWebsocketDataHandler(fn WebsocketDataHandler, interce
 	if bot == nil {
 		return errNilBot
 	}
-	return bot.websocketRoutineManager.registerInterceptor(fn, interceptorOnly)
+	return bot.websocketRoutineManager.registerWebsocketDataHandler(fn, interceptorOnly)
 }
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -954,7 +954,7 @@ func (bot *Engine) WaitForInitialCurrencySync() error {
 // across all enabled exchanges.
 func (bot *Engine) RegisterWebsocketRoutineHandlerInterceptor(fn Interceptor) error {
 	if bot == nil {
-		return fmt.Errorf("%T %w", bot, ErrNilSubsystem)
+		return errNilBot
 	}
 	return bot.websocketRoutineManager.registerInterceptor(fn)
 }
--- a/engine/websocketroutine_manager.go
+++ b/engine/websocketroutine_manager.go
@@ -115,9 +115,9 @@ func (m *websocketRoutineManager) websocketRoutine() {
 						log.Errorf(log.WebsocketMgr, "%v", err)
 					}
 
-					if atomic.LoadInt32(&m.started) != 0 {
-						m.wg.Add(1)
-						go m.websocketDataReceiver(ws)
+					err = m.websocketDataReceiver(ws)
+					if err != nil {
+						log.Errorf(log.WebsocketMgr, "%v", err)
 					}
 
 					err = ws.FlushChannels()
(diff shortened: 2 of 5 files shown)


engine/websocketroutine_manager_test.go
```diff
@@ -257,3 +258,51 @@ func TestWebsocketRoutineManagerHandleData(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestRegisterInterceptorWithFunctionality(t *testing.T) {
+	var m *websocketRoutineManager
```
gloriousCode, line 263: parallel

engine/websocketroutine_manager_test.go
```diff
@@ -257,3 +258,51 @@ func TestWebsocketRoutineManagerHandleData(t *testing.T) {
...
+	go func() {
+		err = m.WebsocketDataHandler("", 1336)
+		if !errors.Is(err, nil) {
+			fmt.Println(err)
+			return
+		}
+
+		err = m.WebsocketDataHandler("", "intercepted")
+		if !errors.Is(err, nil) {
+			fmt.Println(err)
```
gloriousCode, line 300: Pass in `t` and `t.Error` please

engine/websocketroutine_manager.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/log"
 )
 
+var errNilInterceptorFunction = errors.New("interceptor function is nil")
```
gloriousCode, line 20: `_types` already has the errors, please move

engine/websocketroutine_manager.go
```diff
@@ -334,3 +343,20 @@ func (m *websocketRoutineManager) printAccountHoldingsChangeSummary(o account.Ch
...
+	if m == nil {
+		return fmt.Errorf("%T %w", m, ErrNilSubsystem)
+	}
+
+	if fn == nil {
+		return errNilInterceptorFunction
+	}
+
+	m.mu.Lock()
+	m.interceptor = fn
```
gloriousCode, line 359: What if there already has been an interceptor registered? It should be an error and you should also include a function to deregister

engine/engine.go
```diff
@@ -948,3 +948,13 @@ func (bot *Engine) SetupExchanges() error {
 func (bot *Engine) WaitForInitialCurrencySync() error {
 	return bot.currencyPairSyncer.WaitForInitialSync()
 }
+
+// RegisterWebsocketRoutineHandlerInterceptor registers an externally defined
+// interceptor function for diverting and handling websocket notifications
+// across all enabled exchanges.
+func (bot *Engine) RegisterWebsocketRoutineHandlerInterceptor(fn Interceptor) error {
```
gloriousCode, line 955: Could also change the signature to `RegisterWebsocketRoutineHandlerInterceptor(fn Interceptor, interceptorOnly bool)`

So that way, in `(m *websocketRoutineManager) WebsocketDataReceiver(ws *stream.Websocket)`, you can do:
```go
			m.mu.RLock()
			if m.interceptor != nil {
				err = m.interceptor(ws.GetName(), data)
				if err != nil {
				     log.Errorf(...)
				}
				if m.interceptorOnly {
					m.mu.RUnlock()
					continue
				}
			}
			m.mu.RUnlock()
```

engine/websocketroutine_manager.go
```diff
@@ -159,6 +162,12 @@ func (m *websocketRoutineManager) WebsocketDataHandler(exchName string, data int
 			exchName)
 	}
 
+	m.mu.RLock()
```
gloriousCode, line 165: Unrelated to your change, but in the function `WebsocketDataReceiver(ws *stream.Websocket) {`, we go: `m.wg.Add(1)`, could you instead add this before the go function is called inside `websocketRoutine`? eg:
```go
				if ws.IsEnabled() {
					err = ws.Connect()
					if err != nil {
						log.Errorf(log.WebsocketMgr, "%v", err)
					}
					m.wg.Add(1)
					go m.WebsocketDataReceiver(ws)
					err = ws.FlushChannels()
					if err != nil {
						log.Errorf(log.WebsocketMgr, "Failed to subscribe: %v", err)
					}
				}
```

gloriousCode (changes requested):
This one is a little harder for me to appreciate because I have concerns about how this impacts current GCT operations, but for now I've just done a code review. I've thrown some extra design ideas in here as well, but not 100% on those either. After this round, I'll be benchmarking things to check that GCT is not impacted by this change. 

Ultimately, I'd prefer something which if an external function was to be used, its setup and handled on the external side, rather than inside GCT. If GCT runs as normal, it shouldn't be constantly performing checks for something that in most usecases, does not actually do. We get many websocket updates per second, so constantly doing an `RLock` and `nil` check per update for an optional feature is something I'm not too big on

engine/websocketroutine_manager.go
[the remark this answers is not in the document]
shazbert: Yeah this is a good idea 👍

engine/websocketroutine_manager.go
```diff
@@ -131,34 +137,32 @@ func (m *websocketRoutineManager) websocketRoutine() {
 
 // WebsocketDataReceiver handles websocket data coming from a websocket feed
 // associated with an exchange
-func (m *websocketRoutineManager) WebsocketDataReceiver(ws *stream.Websocket) {
-	if m == nil || atomic.LoadInt32(&m.started) == 0 {
-		return
-	}
```
gloriousCode, line 137: Why was this section removed?

gloriousCode (changes requested):

I've run some benchmarks on this and I'm happy with how things are looking: https://gist.github.com/gloriousCode/8348398d7ac4e94a5effacbb3e16a29c

Just two more things I think

engine/websocketroutine_manager.go
shazbert: I just moved them to the calling function [here](https://github.com/thrasher-corp/gocryptotrader/pull/935/files/ede080485037d2fa41bd4b8e6a6edf2de3b77042#diff-ced91e5af7a15f7c959c32b7b5a2ffb26bd9771a25799af41c45acc02de1d3e5R118). The nil check is already done there in that scope and I unexported the `webscoketDataReceiver` itself.

engine/engine.go
```diff
@@ -948,3 +948,14 @@ func (bot *Engine) SetupExchanges() error {
 func (bot *Engine) WaitForInitialCurrencySync() error {
 	return bot.currencyPairSyncer.WaitForInitialSync()
 }
+
+// RegisterWebsocketDataHandler registers an externally defined data handler
+// for diverting and handling websocket notifications across all enabled
+// exchanges. InterceptorOnly as true will purge all other registered handlers
+// (including default) bypassing all other handling.
+func (bot *Engine) RegisterWebsocketDataHandler(fn WebsocketDataHandler, interceptorOnly bool) error {
```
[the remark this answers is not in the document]
shazbert: I am having a hard time conceptualizing the need for this because looking at it from an external perspective, if I have double registered, I would kill the program and fix the mistake, not hook into the engine and check if len(handlers) > n then set default handler via an exported function. Sorry if I am misunderstanding the idea.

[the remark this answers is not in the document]
gloriousCode: We're exporting a function which allows additional data handlers with no way of resetting it and has no way of validating passed in functions. If one is no longer needed, or something needs to change, the solution can't be to restart an application which manages people's orders with real money. We need a way of managing it, and I felt like resetting was the simplest solution out of the ones I've made

engine/websocketroutine_manager.go
gloriousCode: I think it would be best to keep it, if it's called outside of the normal path, we have a function that is stuck

gloriousCode (changes requested):
Thanks for making that change! One 🤞 more 🤞 thing 🤞

engine/websocketroutine_manager.go
```diff
@@ -356,12 +372,24 @@ func (m *websocketRoutineManager) registerWebsocketDataHandler(fn WebsocketDataH
...
+// setWebsocketDataHandler sets a single websocket data handler, removing all
+// pre-existing handlers.
+func (m *websocketRoutineManager) setWebsocketDataHandler(fn WebsocketDataHandler) error {
+	if m == nil {
+		return fmt.Errorf("%T %w", m, ErrNilSubsystem)
+	}
+	if fn == nil {
+		return errNilWebsocketDataHandlerFunction
+	}
+	m.dataHandlers = []WebsocketDataHandler{fn}
```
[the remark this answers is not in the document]
shazbert: 🦅

gloriousCode (approved):
tACK! Thanks for working with me on all my requests. I think this feature is much stronger now. Tested:
```go
err := man.registerWebsocketDataHandler(func(_ string, _ interface{}) error { log.Infof(log.Global, "LOL"); return nil }, false)
	if err != nil {
		return nil, err
	}
	go func() {
		time.Sleep(time.Second * 10)
		_ = man.setWebsocketDataHandler(man.websocketDataHandler)
	}()
	return man, man.registerWebsocketDataHandler(man.websocketDataHandler, false)
```

```go
[INFO]  | TICKER |  09/05/2022 14:49:18  | CoinbasePro websocket BTC-USD SPOT: TICKER: Last $33668.06000000 Ask $33668.06000000 Bid $33665.74000000 High $34938.41000000 Low $33256.00000000 Volume 20846.55946674
[INFO]  | LOG |  09/05/2022 14:49:18  | LOL
```

thrasher- (approved):
tACK, nice addition @shazbert !

PR #939, opened 2022-05-05
alert: Add optimizations

shazbert:
# PR Description

* Add sync pool for re-use between instances. 
* generate a specific persistent routine for alerting when needed, instead of every time an alert is required.
* Preaallocate channels in separate blocking routines to reduce friction.
* In hold routine don't select and default out as we are re-using that channel on each wait. 

NOTES: It's hard to tell from the benchmarks the performance gain on the waiting routine side, there is initial impedance generated by the pool interaction and this is not reusing the channels in benchmarks.

Draft for initial review. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/alert/alert.go
```diff
@@ -18,66 +31,95 @@ type Notice struct {
...
 	// the opposing compare and swap in method wait. This function can return
 	// freely when an alert operation is in process.
-	if !atomic.CompareAndSwapUint32(&n.sema, 1, 2) {
+	if !atomic.CompareAndSwapUint32(&n.sema, active, alerting) {
 		// Return if no waiting routines or currently alerting.
 		return
 	}
-	go n.actuate()
+
+	if n.toActuatorRoutine == nil {
```
gloriousCode, line 52: Should this be len instead of nil? Because the `actuate()` go routine will finish, but it won't be nil anymore.

Or should `actuate()` be `for { select {` ?

exchanges/alert/alert.go
```diff
@@ -5,6 +5,19 @@ import (
...
+// external routines waiting on a state change.
+var pool = sync.Pool{New: func() interface{} { return make(chan bool) }}
+
+const (
+	inactive = uint32(iota)
+	active
+	alerting
+
+	dataToActuatorDefaultBuffer = 1
+	preAllocCommsDefaultBuffer  = 5
```
gloriousCode, line 18: Should these be config variables?

gloriousCode (changes requested):
Hello Shazbert, these are my initial findings of your draft

exchanges/alert/alert.go
shazbert: `dataToActuatorDefaultBuffer` probably not, `preAllocCommsDefaultBuffer`; maybe?

exchanges/alert/alert.go
shazbert: Actuate() is persistent (never returns) so once the channel is set and the routine is scheduled it will continue operating for the life of the program.

exchanges/alert/alert.go
```diff
@@ -18,66 +31,95 @@ type Notice struct {
...
-		select {
-		case ch <- true:
-		default:
-		}
+		n.wg.Done()
+		ch <- true
 	}
-	n.wg.Done()
-	close(ch)
+	pool.Put(ch)
```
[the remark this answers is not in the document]
shazbert: You are correct, that's what can happen, but it's not really how It's intended to be used. For example, how I intended it to be cleanly used 😆:

in a routine.
```go
	for {
		select {
		case <-orderbook.Wait(nil):
			err = m.Check(orderbook)
			if err != nil {
				log.Errorln(monitorLog, err)
			}
		case <-m.shutdown:
			return
		}
	}
```

As for sharing across multiple instances of Notice as a global pool that was entirely intentional to reduce overall running allocations.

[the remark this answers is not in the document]
gloriousCode: I'll test it out! If there isn't some programmatic solution to this or that the above is fine, I think a new readme.md would be a good solution to highlight the potential issue and show how to properly use it

exchanges/alert/alert.go
shazbert: @gloriousCode preAllocCommsDefaultBuffer has been linked to engine settings and readme has been added for usage.

cmd/documentation/exchanges_templates/alert.tmpl
```diff
@@ -0,0 +1,61 @@
+{{define "exchanges alert" -}}
+{{template "header" .}}
+## Alert
+
++ This package allows for multiple routines to wait for a state change on any required data.
```
gloriousCode, line 5: Please update this with more serious example code. Add in writing that this package uses a pool of channels for alerts and then add in writing its the reason to not capture the returning channel because it can lead to false alerts as a result

exchanges/alert/alert.go
```diff
@@ -1,23 +1,50 @@
...
+	preAllocBufferSize = PreAllocCommsDefaultBuffer
+	mu                 sync.RWMutex
+
+	errInvalidBufferSize = errors.New("invalid buffer size cannot be equal or less than zero")
 )
 
+// SetPreAllocationCommsBuffer sets buffer size of the pre-allocated comms.
+func SetPreAllocationCommsBuffer(size int) error {
+	if size <= 0 {
+		return errInvalidBufferSize
```
gloriousCode, line 32: While you have wrapped the err at the engine level, better here
```suggestion
		return fmt.Errorf("%w received %v", errInvalidBufferSize size)
```

exchanges/alert/alert_test.go
```diff
@@ -108,3 +109,37 @@ func BenchmarkWait(b *testing.B) {
 		n.Wait(nil)
 	}
 }
+
+// not needed in binary
```
gloriousCode, line 113: rm or update to something like 
```suggestion
// getSize checks the buffer size for testing purposes
```

exchanges/alert/alert.go
```diff
@@ -1,23 +1,50 @@
...
-var pool = sync.Pool{New: func() interface{} { return make(chan bool) }}
-
 const (
 	inactive = uint32(iota)
 	active
 	alerting
 
 	dataToActuatorDefaultBuffer = 1
-	preAllocCommsDefaultBuffer  = 5
+	PreAllocCommsDefaultBuffer  = 5
```
gloriousCode, line 15: Internalise. I know you're trying to set this via startup, but you don't want to introduce races here by having it exported that anything can change it at anytime. You have the `SetPreAllocationCommsBuffer` function to cover this

gloriousCode (changes requested):
Thanks for making those changes. Few more things

exchanges/alert/alert.go
shazbert: This is a const and I haven't exported the `var preAllocBufferSize` either.

gloriousCode (commented):
Thanks for making those changes! Everything looks good now. Just going to make a benchmark test for it 🤙

exchanges/request/request.go
```diff
@@ -385,14 +385,6 @@ func isVerbose(ctx context.Context, verbose bool) bool {
...
-	val := ctx.Value(contextVerboseFlag)
-	if val == nil {
-		return false
-	}
-
-	isCtxVerbose, ok := val.(bool)
-	if !ok {
-		return false
-	}
+	isCtxVerbose, _ := ctx.Value(contextVerboseFlag).(bool)
```
gloriousCode, line 388: I don't know why you introduced changes here, but there is no guarantee that the value is a bool when casting, please re-introduce `ok`

```suggestion
isCtxVerbose, ok := ctx.Value(contextVerboseFlag).(bool)
	if !ok {
		return false
	}
```

gloriousCode (changes requested):
I've written a benchmark which you can look at and verify yourself [here](https://gist.github.com/gloriousCode/cca2f5478460e20454adeea269b5ee55)
From that, I can conclude that there are no significant speed improvements. However there is a halving of memory allocations which I think would become significant through the runtime of GCT with so many Alert systems running simultaneously. So I'm happy with this

Just one thing that you introduced

gloriousCode (approved):
I change my conclusion after testing my nit and see that the bool without `ok` just defaults to false anyway.

tACK!

thrasher- (approved):
tACK, based on supplied exchanges

PR #946, opened 2022-05-10
orders: adds method to retrieve snapshot of order execution limit values

shazbert:
## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/order/limits.go
```diff
@@ -228,23 +233,8 @@ func (e *ExecutionLimits) CheckOrderExecutionLimits(a asset.Item, cp currency.Pa
 // Limits defines total limit values for an associated currency to be checked
 // before execution on an exchange
 type Limits struct {
```
gloriousCode, line 235: I'm making an assumption here, but it looks like the purpose of this PR is to get a Limit without the potential of an end user directly modifying that limit.

Since Go passes by value, then I propose a simplification. We remove `Limits` and change `ExecutionLimits` to:
```go
type ExecutionLimits struct {
	m   map[asset.Item]map[*currency.Item]map[*currency.Item]MinMaxLevel
	mtx sync.RWMutex
}
```
This way, each exchange can keep their internal `MinMaxLevel` and when you `GetOrderExecutionLimits(...)`, it will just pass the struct by value and prevents an end user to modify the exchange's map value. Only through `LoadLimits` does it get modified

This reduces `protected` locking and removes the need for `GetSnapshot` as it already is one when calling `exch. GetOrderExecutionLimits(...)` to get latest

Having tested and linted this myself, I can confirm that I do not get the linting issue of "consider passing by pointer because its a big struct"

I've created a diff file to look over if you'd like: https://gist.github.com/gloriousCode/aea12694835125af8d8ec4327571996f  Also, the engine test fail will resolve once you merge master 🤙 

Let me know what you think.

shazbert: Yeah, I see your point in that. The main reason I kept it this way, for now, was 1) to keep a limit pointer so I didn't need to lock down the entire map on each order/orderbook change for pricing/amount adjustments. This is happening quite frequently for one asset pair ~10ms if not less. 2) And just easy access to the field values until the limit pointer methods are more robust in adjustments. 

Im calling this and assigning to a struct:
image

Then calling `GetSnapshot` and passing the pointer around to adjustment function handlers associated with that specific asset monitor.

shazbert: @gloriousCode I have stopped returning a pointer for MinMaxLevels and added some notes on functionality. Let me know if this is sufficient.

gloriousCode: Thanks for making the changes that allow `GetSnapshot` to now run faster than the triple-map lookup. Though I'm still a little lost on what you said before I continue.

I don't understand why you'd be calling GetSnapshot every 10ms if Order execution limits are set only once in `Run()` and is never updated again. 
- Are you trying to start loading the limits before the exchange has started loading which is why you need to constantly call it?
- Are you updating the limits yourself calling `UpdateOrderExecutionLimits` in a routine?
  - I would suggest instead updating `UpdateOrderExecutionLimits`  to return the limits as well so you can utilise them yourself. Because you'd then know that the only time the limits change is when your routine calls it
 
Another idea would be to have a `exch.GetAllExecutionLimits()` function so you can get everything in one go and store them, knowing that they won't change and you do this before you start doing your application's routines

shazbert: Yeah, I was being silly, I don't know why I am doing that, to be honest, I will be changing over to what you suggest. I was using that functionality like a placeholder for a future update because I wanted a limits pointer that will be protected by an RWMutex and utilize methods for changing values shared across multiple routines monitoring different order book changes. Maybe that's not the best approach.

gloriousCode (approved):
tACK! Thanks for working with me on this

thrasher- (approved):
Thanks for making that humongous change 😆

gloriousCode (approved):
utaCK! readability++

PR #948, opened 2022-05-11
orders: Add derive modify struct method from order.Detail 

shazbert:
# PR Description

* Easier population of struct data with limited guesswork via secondary method `DeriveModifyResponse`
* Standardize websocket responses to order.Detail
* restrict Modify struct size
* order.ModifyResponse for obvious response differentiation

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/websocketroutine_manager.go
+++ b/engine/websocketroutine_manager.go
@@ -213,7 +213,6 @@ func (m *websocketRoutineManager) WebsocketDataHandler(exchName string, data int
 		}
 		m.syncer.PrintOrderbookSummary(base, "websocket", nil)
 	case *order.Detail:
-		m.printOrderSummary(d)
 		if !m.orderManager.Exists(d) {
 			err := m.orderManager.Add(d)
 			if err != nil {
--- a/engine/websocketroutine_manager.go
+++ b/engine/websocketroutine_manager.go
@@ -231,17 +231,7 @@ func (m *websocketRoutineManager) WebsocketDataHandler(exchName string, data int
 				return err
 			}
 		}
-	case *order.Modify:
-		m.printOrderChangeSummary(d)
-		od, err := m.orderManager.GetByExchangeAndID(d.Exchange, d.ID)
-		if err != nil {
-			return err
-		}
-		od.UpdateOrderFromModify(d)
-		err = m.orderManager.UpdateExistingOrder(od)
-		if err != nil {
-			return err
-		}
+		m.printOrderSummary(d)
 	case order.ClassificationError:
 		return fmt.Errorf("%w %s", d.Err, d.Error())
 	case stream.UnhandledMessageWarning:
@@ -277,29 +267,6 @@ func (m *websocketRoutineManager) FormatCurrency(p currency.Pair) currency.Pair 
 	}
 	return p.Format(m.currencyConfig.CurrencyPairFormat.Delimiter,
 		m.currencyConfig.CurrencyPairFormat.Uppercase)
-}
-
-// printOrderChangeSummary this function will be deprecated when a order manager
-// update is done.
-func (m *websocketRoutineManager) printOrderChangeSummary(o *order.Modify) {
-	if m == nil || atomic.LoadInt32(&m.started) == 0 || o == nil {
-		return
-	}
-
-	log.Debugf(log.WebsocketMgr,
-		"Order Change: %s %s %s %s %s %s OrderID:%s ClientOrderID:%s Price:%f Amount:%f Executed Amount:%f Remaining Amount:%f",
-		o.Exchange,
-		o.AssetType,
-		o.Pair,
-		o.Status,
-		o.Type,
-		o.Side,
-		o.ID,
-		o.ClientOrderID,
-		o.Price,
-		o.Amount,
-		o.ExecutedAmount,
-		o.RemainingAmount)
 }
 
 // printOrderSummary this function will be deprecated when a order manager
--- a/engine/websocketroutine_manager.go
+++ b/engine/websocketroutine_manager.go
@@ -238,6 +238,7 @@ func (m *websocketRoutineManager) websocketDataHandler(exchName string, data int
 			if err != nil {
 				return err
 			}
+			m.printOrderSummary(d, false)
 		} else {
 			od, err := m.orderManager.GetByExchangeAndID(d.Exchange, d.ID)
 			if err != nil {
@@ -249,8 +250,8 @@ func (m *websocketRoutineManager) websocketDataHandler(exchName string, data int
 			if err != nil {
 				return err
 			}
+			m.printOrderSummary(d, true)
 		}
-		m.printOrderSummary(d)
 	case order.ClassificationError:
 		return fmt.Errorf("%w %s", d.Err, d.Error())
 	case stream.UnhandledMessageWarning:
@@ -290,12 +291,19 @@ func (m *websocketRoutineManager) FormatCurrency(p currency.Pair) currency.Pair 
 
 // printOrderSummary this function will be deprecated when a order manager
 // update is done.
-func (m *websocketRoutineManager) printOrderSummary(o *order.Detail) {
+func (m *websocketRoutineManager) printOrderSummary(o *order.Detail, isUpdate bool) {
 	if m == nil || atomic.LoadInt32(&m.started) == 0 || o == nil {
 		return
 	}
+
+	orderNotif := "New Order:"
+	if isUpdate {
+		orderNotif = "Order Change:"
+	}
+
 	log.Debugf(log.WebsocketMgr,
-		"New Order: %s %s %s %s %s %s OrderID:%s ClientOrderID:%s Price:%f Amount:%f Executed Amount:%f Remaining Amount:%f",
+		"%s %s %s %s %s %s %s OrderID:%s ClientOrderID:%s Price:%f Amount:%f Executed Amount:%f Remaining Amount:%f",
+		orderNotif,
 		o.Exchange,
 		o.AssetType,
 		o.Pair,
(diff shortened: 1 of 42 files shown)


exchanges/bithumb/bithumb_wrapper.go
```diff
@@ -527,15 +527,16 @@ func (b *Bithumb) ModifyOrder(ctx context.Context, action *order.Modify) (*order
...
-	return &order.Modify{
-		Exchange:  action.Exchange,
-		AssetType: action.AssetType,
-		Pair:      action.Pair,
-		ID:        o.Data[0].ContID,
-		Price:     float64(int64(action.Price)),
-		Amount:    action.Amount,
-		Side:      action.Side,
-	}, nil
+	resp, err := action.DeriveModifyResponse()
```
gloriousCode, line 530: Unrelated, but I can't even find documentation on the ability to modify an order

exchanges/ftx/ftx_wrapper.go
```diff
@@ -682,12 +682,15 @@ func (f *FTX) SubmitOrder(ctx context.Context, s *order.Submit) (order.SubmitRes
...
+func (f *FTX) ModifyOrder(ctx context.Context, action *order.Modify) (*order.ModifyResponse, error) {
 	if err := action.Validate(); err != nil {
 		return nil, err
 	}
 
-	if action.TriggerPrice != 0 {
+	var id string
+	switch {
+	case action.TriggerPrice != 0:
+		var a TriggerOrderData
```
gloriousCode, line 693: I think it would be best to set more details from the response with all of these functions. Including price/amount to account for any issues

exchanges/bitmex/bitmex_wrapper.go
```diff
@@ -609,34 +609,30 @@ func (b *Bitmex) SubmitOrder(ctx context.Context, s *order.Submit) (order.Submit
...
 
-	return &order.Modify{
-		Exchange:  action.Exchange,
-		AssetType: action.AssetType,
-		Pair:      action.Pair,
-		ID:        o.OrderID,
-		Price:     action.Price,
-		Amount:    float64(params.OrderQty),
-	}, nil
+	resp.OrderID = o.OrderID
```
gloriousCode, line 634: I think it would be best to include as many details as possible from the order response, including price and amount rather than trusting the request attempt implicitly

exchanges/bitfinex/bitfinex_wrapper.go
```diff
@@ -675,35 +675,33 @@ func (b *Bitfinex) SubmitOrder(ctx context.Context, o *order.Submit) (order.Subm
...
 // ModifyOrder will allow of changing orderbook placement and limit to
 // market conversion
-func (b *Bitfinex) ModifyOrder(ctx context.Context, action *order.Modify) (*order.Modify, error) {
+func (b *Bitfinex) ModifyOrder(ctx context.Context, action *order.Modify) (*order.ModifyResponse, error) {
 	if err := action.Validate(); err != nil {
 		return nil, err
 	}
 
+	if !b.Websocket.CanUseAuthenticatedWebsocketForWrapper() {
+		return nil, common.ErrNotYetImplemented
```
gloriousCode, line 684: Place above validate

exchanges/bitfinex/bitfinex_wrapper.go
```diff
@@ -675,35 +675,33 @@ func (b *Bitfinex) SubmitOrder(ctx context.Context, o *order.Submit) (order.Subm
...
-		return &order.Modify{
-			Exchange:  action.Exchange,
-			AssetType: action.AssetType,
-			Pair:      action.Pair,
-			ID:        action.ID,
-			Price:     action.Price,
-			Amount:    action.Amount,
-		}, err
+
+	request := WsUpdateOrderRequest{
```
gloriousCode, line 692: rn due to package conflict

exchanges/bithumb/bithumb_wrapper.go
```diff
@@ -527,15 +527,16 @@ func (b *Bithumb) ModifyOrder(ctx context.Context, action *order.Modify) (*order
...
+	if err != nil {
+		return nil, err
+	}
+
+	if len(o.Data) > 0 {
+		resp.OrderID = o.Data[0].ContID
```
gloriousCode, line 536: Set price, amount, fee

exchanges/poloniex/poloniex_wrapper.go
```diff
@@ -587,16 +587,12 @@ func (p *Poloniex) ModifyOrder(ctx context.Context, action *order.Modify) (*orde
...
-		Price:             action.Price,
-		Amount:            action.Amount,
-		PostOnly:          action.PostOnly,
-		ImmediateOrCancel: action.ImmediateOrCancel,
-	}, nil
+	modResp, err := action.DeriveModifyResponse()
+	if err != nil {
+		return nil, err
+	}
+	modResp.OrderID = strconv.FormatInt(resp.OrderNumber, 10)
```
gloriousCode, line 594: Include fee

engine/websocketroutine_manager.go
```diff
@@ -299,29 +288,6 @@ func (m *websocketRoutineManager) FormatCurrency(p currency.Pair) currency.Pair
 		m.currencyConfig.CurrencyPairFormat.Uppercase)
 }
 
-// printOrderChangeSummary this function will be deprecated when a order manager
-// update is done.
-func (m *websocketRoutineManager) printOrderChangeSummary(o *order.Modify) {
```
gloriousCode, line 304: Please keep this and update it. You've changed all WS modifies to Detail and every message says the order is new. Update handling to print this when modifying an existing order and use the other one for printing a new order

gloriousCode (changes requested):
I'm fine with this one, but I would like to see more details from response populated instead of assuming from the request info

exchanges/bithumb/bithumb_wrapper.go
shazbert: I have no keys to test this either, I am thinking about removing this functionality.

exchanges/poloniex/poloniex_wrapper.go
shazbert: I am not sure about adding this as it seems to be a fee rate in the docs [here](https://docs.poloniex.com/#moveorder). I don't have funds and actively modifying orders on Poloniex. So I don't know how that relates to the executed trades when the order has been modified and whether that needs to be handled more appropriately. 

In your other nits you suggest adding additional fields as well for this PR.  I didn't want to add extra fields but reduce fields that weren't being used in the code base. Maybe we can defer this for now until I can actively use modify like `Btc Markets` because of nuanced differences between exchanges?

As for not implicitly trusting the price/amounts data that is good point but I wanted to keep the state the same as before. Like bitmex for example we were using the modify field price and the params.OrderQty which was generated from the modify struct to make the return struct anyway. Maybe add a trello item for this?

gloriousCode: Yeah, given my spiel for the `order.Detail`, I'm okay with the removal of fields here - though I think fee is important and will likely be added back. The separation of `order.Modify` and having a `order.ModifyReposonse` is good too. I've been inconsistent with my responses in these PRs, but the two are on a fine line of containing information that could be misleading which is what I'm trying to avoid - though its a lot less with this Modify PR versus the Detail one. 

The main thing though is that we send `order.ModifyResponse` to the order manager and update our internal order records. By using the request information to update internally and not the real server-side data, we're updating our internal records of an order with information that may not be accurate. This was why I focused on fee information - if you change the price and amount of an order, the fee that you'd pay on the order would be changed too. If it were just a response struct that you'd hold and then call `GetOrderInfo` afterwards and update with that, then I wouldn't mind so much.

BUT, you're right, this PR is mostly keeping the functionality the same, so a Trello ticket will suffice.

exchanges/bitmex/bitmex_wrapper.go
gloriousCode, line 634: Include last updated time and remaining amount

exchanges/ftx/ftx_wrapper.go
```diff
@@ -698,46 +701,34 @@ func (f *FTX) ModifyOrder(ctx context.Context, action *order.Modify) (*order.Mod
...
-		Pair:      action.Pair,
-		ID:        strconv.FormatInt(o.ID, 10),
-		Price:     action.Price,
-		Amount:    action.Amount,
-	}, err
+	resp, err := action.DeriveModifyResponse()
+	if err != nil {
+		return nil, err
+	}
+	resp.OrderID = id
```
gloriousCode, line 730: Include remaining amount

exchanges/order/orders.go
```diff
@@ -244,24 +244,16 @@ func (d *Detail) UpdateOrderFromDetail(m *Detail) {
 
 // UpdateOrderFromModify Will update an order detail (used in order management)
 // by comparing passed in and existing values
-func (d *Detail) UpdateOrderFromModify(m *Modify) {
+func (d *Detail) UpdateOrderFromModify(m *ModifyResponse) {
```
gloriousCode, line 247:
```suggestion
func (d *Detail) UpdateOrderFromModifyResponse(m *ModifyResponse) {
```

gloriousCode (changes requested):
Thanks for making those changes! Just only minor things now without the larger changes

exchanges/poloniex/poloniex_wrapper.go
shazbert: Thank you sir! 💐

gloriousCode (approved):

thrasher- (approved):
Another helpful method! Bithumb doesn't allow non-corporate foreign users anymore so can't be tested but reviewing their API docs shows no modify ability.

PR #1011, opened 2022-08-19
Backtester: mini bug fixes

gloriousCode:
# PR Description
While working on my other branch I noticed there were some strategies not working. Then I noticed they didn't work on `master` either.

This fixes some typos with errors and the report template. ~I set `err = nil` after each of the `errors.Is` processing exceptions to prevent this kind of fall-through from happening again, but I'm happy to revert those if you don't like it.~ edit: this is no longer the case

Fixes # (issue)

## Type of change

## How has this been tested
- [x] Testing dca-api-candles-exchange-level-funding.strat
- [x] Testing rsi-api-candles.strat
- [x] Testing t2b2-api-candles-exchange-funding.strat
- [x] go test ./... -race

## Checklist

### DIFF
--- a/backtester/config/config_test.go
+++ b/backtester/config/config_test.go
@@ -930,26 +930,12 @@ func TestGenerateConfigForRSIAPICustomSettings(t *testing.T) {
 				MakerFee: &makerFee,
 				TakerFee: &takerFee,
 			},
-			{
-				ExchangeName: testExchange,
-				Asset:        asset.Spot,
-				Base:         currency.ETH,
-				Quote:        currency.USDT,
-				SpotDetails: &SpotDetails{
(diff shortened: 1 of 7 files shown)


shazbert (changes requested):
Changes look good 👍, there is a minor issue that I might have found with `rsi-api-candles.strat`, it doesn't sell ETH reserves when RSI is above 70. So there is no trading activity, it shows there is no balance for ETH when there is a signal.

shazbert (approved):
Thanks man, that looks good now!

thrasher- (approved):
tACK - old RSI:

image

PR:

image

Cash and carry example looks fine too.

PR #1023, opened 2022-09-08
Backtester: Live trading upgrades

gloriousCode:
image

# PR Description

The Backtester's live pair trading feature was only limited to one pair and did not support many features. This PR updates the live trading data source to run the same strategies as all other data types while moving real order trading closer to being usable.

### Live pair trading
- Added new strategy `binance-live-cash-and-carry.strat` which can perform cash and carry in real time.. Once Binance futures is supported
  -  Existing strategy `dca-candles-live.strat` has been updated and works nicely
- Added new subsystem that retrieves candles, wallet balances and collateral at intervals you define
- Adds new "close all positions on stop" feature as the strategies have no end-point, and strategies like "cash and carry" require the selling of positions to finish
   - This is now achieved by sending the stop command to the RPC server
- Reworks many systems to deal with appending data, instead of it all being pre-set

### Real order trading
- Updates livedata config to allow slice of exchange credentials in order to retrieve wallet balance and collateral values and place orders on your behalf
- It uses funding data from your actual API account. So if you set the config to $1000, but you only have $1 left in your trading wallet on Binance from your sweet strategies, the real order trader will only utilise the $1

### Other bits and bobs
- Added new "close all tasks on close" feature, so if you have many tasks running, they will be closed when you hit ctrl+c
  - If a strategy is configured to close all positions on stop, it will still run that before exiting the application
- Added extra functionality to `order.Submit` to allow a request to make an additional request to retrieve fees. FTX used to not return fees for orders, but you can gather fee details from an additional request
- Changed `map[string]map[asset.Item]map[currency.Pair]` to `map[string]map[asset.Item]map[*currency.Item]map[*currency.Item]` because maps do not use the `Equal` function, and instead any difference in formatting will return as `!ok`
- Fixed a cool bug where the function `RemoveDuplicates` wasn't _reaaaaally_ removing duplicates

### Something very important to consider
- ~As per [this](https://trello.com/c/NcwdOtKy/166-existing-backtester-bugs) trello ticket, sizing restrictions do not work for cash and carry. So if you decide to test a strategy with real money, you _must_ use a subaccount with limited funding, otherwise it will happily eat your money~ This is now fixed
- Please be careful running live strategies with your API keys, it will use your money and you will likely take some losses. Ensure you have limits set. I would recommend making a subaccount with limited funds to test this
[...]

PR #1049, opened 2022-09-26
ftx: fix null currency.Pair UnmarshalJSON

geseq:
# PR Description

I'm not sure this is the best place for this fix, but with FTX on some submit order calls (spot, with IOC), the returned data includes both `market` with the asset pair and `future` with `null`.

The `null` ends up in currency pair `UnmarshalJSON` resulting in an error for a successful SubmitOrder call.
I am not entirely sure if this will be a breaking change

If this looks ok at face value I'm happy to add some unit tests[

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Manually

## Checklist

### DIFF
--- a/currency/pair_methods.go
+++ b/currency/pair_methods.go
@@ -34,6 +34,11 @@ func (p *Pair) UnmarshalJSON(d []byte) error {
 		return err
 	}
 
+	if pair == "" {
+		*p = EMPTYPAIR
+		return nil
+	}
+
 	newPair, err := NewPairFromString(pair)
 	if err != nil {
 		return err
--- a/currency/pair_test.go
+++ b/currency/pair_test.go
@@ -72,6 +72,9 @@ func TestPairUnmarshalJSON(t *testing.T) {
 	}
 
 	encoded, err = json.Marshal(EMPTYPAIR)
+	if err != nil {
+		t.Fatal(err)
+	}
 	err = json.Unmarshal(encoded, &unmarshalHere)
 	if err != nil {
 		t.Fatal("Pair UnmarshalJSON() error", err)
--- a/currency/pair_test.go
+++ b/currency/pair_test.go
@@ -70,6 +70,19 @@ func TestPairUnmarshalJSON(t *testing.T) {
 		t.Errorf("Pairs UnmarshalJSON() error expected %s but received %s",
 			configPair, unmarshalHere)
 	}
+
+	encoded, err = json.Marshal(EMPTYPAIR)
+	err = json.Unmarshal(encoded, &unmarshalHere)
+	if err != nil {
+		t.Fatal("Pair UnmarshalJSON() error", err)
+	}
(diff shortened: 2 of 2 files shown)


currency/pair_methods.go
```diff
@@ -34,6 +34,11 @@ func (p *Pair) UnmarshalJSON(d []byte) error {
 		return err
 	}
 
+	if pair == "" {
```
gloriousCode, line 37: Please add a test case for `TestPairUnmarshalJSON` with
```go
encoded, err = json.Marshal(EMPTYPAIR)
	err = json.Unmarshal(encoded, &unmarshalHere)
	if err != nil {
		t.Fatal("Pair UnmarshalJSON() error", err)
	}
```

gloriousCode (changes requested):
Hello!
When speaking specifically about the FTX submit order, I made the following changes in another PR #1023 That field was definitely annoying, even with futures orders and that change fixes up the issue. 

image

I also share the concern of it breaking in allowing more currency pair slips to come through, and now needing an extra `IsEmpty()` check. I'll need to do more testing - replacing your `return nil` with `os.Exit(-1)` has not yielded any unexpected exits though. It may be simpler to just change the `currency.Pair` property to `string`

shazbert (commented):
Happy with this change. 👍 Agree with the tests then should be good to go.

gloriousCode (approved):
tACK! The only area that I identified a potential EMPTYPAIR slip-through spot was websocket order processing. However, its still preferable to store and track an order with an empty pair than it would be to error as it unmarshalls the JSON with a nil pair entry.

All other areas of concern for me had some form of checking or handled the situation appropriately. So thank you for your contribution!

currency/pair_test.go
```diff
@@ -70,6 +70,22 @@ func TestPairUnmarshalJSON(t *testing.T) {
...
+	err = json.Unmarshal(encoded, &unmarshalHere)
+	if err != nil {
+		t.Fatal("Pair UnmarshalJSON() error", err)
+	}
+	err = json.Unmarshal([]byte("null"), &unmarshalHere)
+	if err != nil {
+		t.Fatal("Pair UnmarshalJSON() error", err)
+	}
+	if unmarshalHere != EMPTYPAIR {
+		t.Fatalf("Expected EMPTPAIR got: %s", unmarshalHere)
```
thrasher-, line 87:
```suggestion
		t.Fatalf("Expected EMPTYPAIR got: %s", unmarshalHere)
```
😄

thrasher- (changes requested):
One minor nit and then this looks G2G 👍

shazbert (approved):
Changes ACK!

thrasher- (approved):

thrasher-:
Thanks @geseq !

Issue #901, opened 2022-03-15
Websocket stuck


idoall:
# New Issue

Websocket stuck

When switching to a usable proxy, the websocket will get stuck when trying connect again.
## Context

Please provide any relevant information about your setup. This is important in case the issue is not reproducible except for under certain conditions.

* Operating System:
mac 11.5
* GoCryptoTrader version (`gocryptotrader -version`):
GoCryptoTrader v0.1 amd64 go1.17.6 pre-release.
## Current Behavior
```
When the proxy network switches, websockt will output "binance websocket has been disconnected. Reason: read TCP 10.17.0.6:65306 - > 52.68.124.244:9443: use of closed network connection"
```

1. Use a timer to detect websocket every 5 minutes
2.
```
If ws.Isconnected() {
    return
}else{
    ws.connect()
}
```

### Failure Logs

```
binance websocket has been disconnected. Reason: read TCP 10.17.0.6:65306 - > 52.68.124.244:9443: use of closed network connection
```

thrasher-:
Hey @idoall, are you still experiencing this issue after #1059 was merged in?

thrasher-:
Closing this due to being stale but please feel free to re-open this if you're still experiencing an issue.

Issue #761, opened 2021-08-24
Binance Websocket Reconnection Deadlock


yangrq1018:
# New Issue

## Context

As far as from what I learned from Slack, `exchanges/stream` has a `Websocket.trafficMonitor` method that fires a routine which checks periodically the websocket is being read data from. When a timeout occurs and no data is read from, it attempts to call `w.Shutdown()`. Shutdown has a piece of code like this

```golang
close(w.ShutdownC)
w.Wg.Wait()
```

This seems to cause a deadlock problem:
1. `Binance.wsReadData` is holding Wg. It doesn't check if `w.ShutdownC` is closed at all in the ininifte loop.
2. `Websocket.trafficMonitor` is holding Wg. It waits on itself as `w.Wg.Done()` is called AFTER `w.Shutdown()` returns.

* Operating System:
Windows 10 x64

* GoCryptoTrader version (`gocryptotrader -version`):
On revision number 3467914ab763e33e4b088f21f1637329b9176f93

## Expected Behavior
I expect `w.Shutdown()` to return rather than causing a deadlock.

## Current Behavior
Besides the deadlock mentioned above, it also blocks `Websocket.Connect` which tries to acquire `Websocket.m` that is held by Shutdown. So what i am seeing is:

1. suppose the websocket is inactive for more than two seconds, exceeding timeout
2. trafficMonitor detects it and tries to call Shutdown. Shutdown blocks. Even if the user calls Websocket.Connect by himself, it blocks forever too.

I could work on a PR, but I am not very available recently, progress could be slow.

## Failure Information (for bugs)

Please help by providing information about the failure. If it is not a bug, please remove the rest of this template.

### Steps to Reproduce

config: 

```json
{
  "name": "Binance",
   "enabled": true,
   "verbose": false,
   "httpTimeout": 15000000000,
   "websocketResponseCheckTimeout": 30000000,
   "websocketResponseMaxLimit": 7000000000,
   "websocketTrafficTimeout": 2000000000,
   ...
}
```

Set `websocketTrafficTimeout` to  2 seconds. Start binance websocket. Don't subscribe to anything (actually I subscribed to order update only but it is very infrequent obviously). 

Add some logging to `websocket.go`

```golang
				if !w.IsConnecting() && w.IsConnected() {
					log.Infof(log.WebsocketMgr, "attempt shutdown")
					err := w.Shutdown()
					log.Infof(log.WebsocketMgr, "shutdown returns")
					if err != nil {
[...]

Issue #683, opened 2021-05-04
error: Websocket already connected


mshogin:
That comes from https://github.com/thrasher-corp/gocryptotrader/blob/master/exchanges/stream/websocket.go#L187

`
0  0x0000000000cfe8bb in github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).Connect
   at ./exchanges/stream/websocket.go:172
1  0x00000000015c4f2d in github.com/thrasher-corp/gocryptotrader/engine.(*Engine).WebsocketRoutine.func1
   at ./engine/routines.go:237
2  0x00000000004857c1 in runtime.goexit
   at /usr/local/go/src/runtime/asm_amd64.s:1371

0  0x0000000000cfe8bb in github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).Connect
   at ./exchanges/stream/websocket.go:172
1  0x00000000015b8b2a in github.com/thrasher-corp/gocryptotrader/engine.(*ExchangeCurrencyPairSyncer).Start
   at ./engine/syncer.go:520
2  0x00000000004857c1 in runtime.goexit
   at /usr/local/go/src/runtime/asm_amd64.s:1371

`

OS
`Linux mshogin-laptop 5.8.0-44-generic #50~20.04.1-Ubuntu SMP Wed Feb 10 21:07:30 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

go
`go env                                             
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/mshogin/.cache/go-build"
GOENV="/home/mshogin/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/mshogin/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/mshogin/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GOVCS=""
GOVERSION="go1.16"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
[...]

PR #1059, opened 2022-10-13
websocket: fix shutdown deadlock

geseq:
# PR Description

I've been noticing some reconnection issues and tracking it down, it looks like there's a condition that causes deadlock during websocket shutdown. 

Since this is in the traffic monitor which doesn't seem to be doing anything other than monitoring for traffic this seems fairly safe to do.

Not sure how to write a reliable unit test.

might be the fix for #901, #761 and possibly #683 as the issues are all similar to what I was noticing in my logs

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -568,6 +568,7 @@ func (w *Websocket) trafficMonitor() {
 						w.trafficTimeout)
 				}
 				trafficTimer.Stop()
+				w.Wg.Done() // without this the w.Shutdown() call below will deadlock
 				if !w.IsConnecting() && w.IsConnected() {
 					err := w.Shutdown()
 					if err != nil {
@@ -577,7 +578,6 @@ func (w *Websocket) trafficMonitor() {
 					}
 				}
 				w.setTrafficMonitorRunning(false)
-				w.Wg.Done()
 				return
 			}
 
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -568,8 +568,8 @@ func (w *Websocket) trafficMonitor() {
 						w.trafficTimeout)
 				}
 				trafficTimer.Stop()
-				w.Wg.Done() // without this the w.Shutdown() call below will deadlock
 				w.setTrafficMonitorRunning(false)
+				w.Wg.Done() // without this the w.Shutdown() call below will deadlock
 				if !w.IsConnecting() && w.IsConnected() {
 					err := w.Shutdown()
 					if err != nil {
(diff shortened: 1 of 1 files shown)


gloriousCode (approved):
Well that's a little bit funny :sweat_smile: Thank you for your research and fixing the issue.

I wrote the following test to confirm you suspicions:
```go
func TestTrafficTimeout(t *testing.T) {
	w := &Websocket{
		trafficTimeout: 0,
		ShutdownC:      make(chan struct{}),
		Wg:             &sync.WaitGroup{},
	}
	w.Wg.Add(1)
	w.connected = true
	go func() {
		defer w.Wg.Done()
		w.trafficMonitor()
	}()
	w.Wg.Wait()
}
```
This test will not pass on the old way because it gets stuck waiting during `Shutdown()`
I'm not requesting you add this test, because it only checks to see that it doesn't get stuck

tACK! Thank you for your contribution!

exchanges/stream/websocket.go
```diff
@@ -577,7 +578,6 @@ func (w *Websocket) trafficMonitor() {
 					}
 				}
 				w.setTrafficMonitorRunning(false)
```
shazbert, line 580: I think it might better to move this above the Done() call just so in the unlikely event when a traffic monitor starts up again and this isn't set to false. It might hang differently. Let me know what you think?

Other than that looks good! Thanks for checking this out.

shazbert (changes requested):
🚀

exchanges/stream/websocket.go
geseq: Makes sense. I’ve updated it.

exchanges/stream/websocket.go
```diff
@@ -568,6 +568,8 @@ func (w *Websocket) trafficMonitor() {
 						w.trafficTimeout)
 				}
 				trafficTimer.Stop()
+				w.Wg.Done() // without this the w.Shutdown() call below will deadlock
+				w.setTrafficMonitorRunning(false)
```
shazbert, line 572: Sorry, can you please swap these around so they are in the same call queue as before?

shazbert (changes requested):
Just one minor change.

exchanges/stream/websocket.go
geseq: Done

gloriousCode (approved):

shazbert (approved):
Thank you for checking this out, physically terminated websocket connection and reconnect occurred. ACK. 🚀

thrasher- (approved):
Nice catch @geseq !

PR #1072, opened 2022-11-01
requester: Remove httpclient.Do errors from latency measurement

geseq:
# PR Description

When http client.Do fails I don't think it makes sense to measure latency since this is due to things like connectivity issues rather than measuring the latency of the request-response cycle.

I thought about passing around the resp/err to the Latency call to let this be checked down the line, but ultimately I don't really see a use case for it as the error is passed down the line anyway. 

Happy to hear differing opinions.

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Manually by closing device connection

## Checklist

### DIFF
--- a/exchanges/request/request.go
+++ b/exchanges/request/request.go
@@ -165,7 +165,7 @@ func (r *Requester) doRequest(ctx context.Context, endpoint EndpointLimit, newRe
 
 		resp, err := r._HTTPClient.do(req)
 
-		if r.reporter != nil && err != nil {
+		if r.reporter != nil && err == nil {
 			r.reporter.Latency(r.name, p.Method, p.Path, time.Since(start))
 		}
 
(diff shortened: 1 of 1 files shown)


shazbert (approved):
LGTM - tested approved.

gloriousCode (approved):
tACK! I personally prefer "passing around the resp/err to the Latency call to let this be checked down the line" to let an individual reporter implementation choose what to do. However, I'm not actually using the reporter, the errors returned don't seem to be something for the latency reporter to be reporting on, and if this were to change, its a very simple fix to make later. So I'm happy with this

thrasher- (approved):
Your reasoning sounds fine, no differing opinion from I. Thanks!

PR #1091, opened 2022-11-30
kline/exchanges: automatic creation of unsupported candle intervals

shazbert:
# Automatic Creation of Unsupported Candle Intervals

image

This feature enables the automatic creation of unsupported candle intervals that are not natively supported by the exchange. This is achieved using wrapper functionality.

For example, if 4hr candles are needed but the exchange only supports 1hr candles, this feature will automatically create 4 hour candles using 1 hour candles and the wrapper.

Reasoning on why this implementation: [HERE](https://github.com/thrasher-corp/gocryptotrader/pull/1091#discussion_r1050159834)

## What Has Been Done
# Exchanges
* Pointer return added for `kline.Item`
* Wrapper functions have been standardized and refactored, as well as exchange intervals and limits
* Bug fixes have been implemented
* Extended functions that do not use date ranges have been deprecated, trello items have been added [HERE](https://trello.com/c/xGlzmgmO/177-re-implement-gethistoriccandlesextended-functionality-for-exchanges) 
    - BTSE
    - GateIO
    - Huobi
    - Kraken
    - Poloniex
    - Bithumb
* Helper functions have been added to the `kline.Request` to reduce implementation complexities when sorting and dealing with malformed candles
## kline package
* `Request` and `RequestExtended` types have been added for standard request checking, processing, and complexity reduction
* `ExchangeIntervals` type added for matching and constructing intervals from supported intervals
* Methods `RemoveDuplicates`, `RemoveOutsideRange`, and `SetHasDataFromCandles` have been optimized
* `TotalCandlesPerInterval` function has been refactored
* Functions `ConvertToNewInterval` and `CreateKline` have been optimized and refactored
* `CalculateCandleDateRanges` function has been preallocated
* Adds method `addPadding` to automatically fill out zero value candles in the event the endpoint only returns candles with trading data.

This is in preparation for an upcoming upgrade that will enable the backtester to retrieve custom candles, such as 4hr, 6hr, and 12hr candles, for use with a Simple Moving Average (SMA).

What needs to be integrated with backtester [Reference](https://github.com/thrasher-corp/gocryptotrader/pull/1091#discussion_r1048994844)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

[...]

PR #1107, opened 2022-12-23
cli: rpc time parsing

gloriousCode:
# PR Description
I noticed this when parsing time for #1091 that the start end times looked more funky than I'd like them to be. Turns out the implementation of `negateLocalOffset` was off. 

When running `gethistoriccandles` it would be returning candle data hours outside of my expected timeframe. Its harder to measure since `gethistoriccandles` truncates the time, but it was well over 11 hours + truncated time. I was expecting candle data up to late 22nd Dec/early 23rd Dec. Instead I was getting 1pm on the 22nd. See below:
```
 "start": "2022-12-20 21:00:00",
 "end": "2022-12-22 13:00:00",
```

where now it is in line with my expectations where it is in line with the truncated time
```
 "start": "2022-12-21 08:00:00",
 "end": "2022-12-23 00:00:00",
```

This also matches `negateLocalOffsetTS`'s implementation

## Type of change

## How has this been tested
Printing times in `commands.go` and matching them up in `rpcserver.go`
I was doing [these changes via gist](https://gist.github.com/gloriousCode/7261f4e5311d95799aa70a30ac256f7a)

### DIFF
--- a/cmd/gctcli/helpers.go
+++ b/cmd/gctcli/helpers.go
@@ -44,13 +44,9 @@ func closeConn(conn *grpc.ClientConn, cancel context.CancelFunc) {
 // so this function will minus the offset from the local sent time
 // to allow for proper use at RPCServer
 func negateLocalOffset(t time.Time) string {
-	_, offset := time.Now().Zone()
 	return t.In(time.UTC).Format(common.SimpleTimeFormat)
 }
 
-}
-
 func negateLocalOffsetTS(t time.Time) *timestamppb.Timestamp {
-	_, offset := time.Now().Zone()
-	return timestamppb.New(t.Add(time.Duration(-offset) * time.Second))
+	return timestamppb.New(t.In(time.UTC))
 }
--- a/cmd/gctcli/helpers.go
+++ b/cmd/gctcli/helpers.go
@@ -45,9 +45,7 @@ func closeConn(conn *grpc.ClientConn, cancel context.CancelFunc) {
 // to allow for proper use at RPCServer
 func negateLocalOffset(t time.Time) string {
 	_, offset := time.Now().Zone()
-	loc := time.FixedZone("", -offset)
-
-	return t.In(loc).Format(common.SimpleTimeFormat)
+	return t.Add(-time.Duration(offset) * time.Second).Format(common.SimpleTimeFormat)
 }
 
 func negateLocalOffsetTS(t time.Time) *timestamppb.Timestamp {
--- a/cmd/gctcli/helpers.go
+++ b/cmd/gctcli/helpers.go
@@ -45,7 +45,9 @@ func closeConn(conn *grpc.ClientConn, cancel context.CancelFunc) {
 // to allow for proper use at RPCServer
 func negateLocalOffset(t time.Time) string {
 	_, offset := time.Now().Zone()
-	return t.Add(-time.Duration(offset) * time.Second).Format(common.SimpleTimeFormat)
+	return t.In(time.UTC).Format(common.SimpleTimeFormat)
+}
+
 }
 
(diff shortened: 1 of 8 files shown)


cmd/gctcli/helpers.go
```diff
@@ -45,9 +45,7 @@ func closeConn(conn *grpc.ClientConn, cancel context.CancelFunc) {
 // to allow for proper use at RPCServer
 func negateLocalOffset(t time.Time) string {
 	_, offset := time.Now().Zone()
-	loc := time.FixedZone("", -offset)
-
-	return t.In(loc).Format(common.SimpleTimeFormat)
+	return t.Add(-time.Duration(offset) * time.Second).Format(common.SimpleTimeFormat)
```
thrasher-, line 48: How about this?

```suggestion
	return t.In(time.UTC).Format(common.SimpleTimeFormat)
}

```

gloriousCode: I could have sworn that there was a reason why we didn't do this........ But its matching
cmd:
```
2022-12-23 15:59:00 +1100 AEDT
```
rpcserver
```
 2022-12-23 15:59:00 +1100 AEDT
```

# 🌞

shazbert (approved):
tACK but a suggestion below: 
```diff
diff --git a/cmd/gctcli/helpers.go b/cmd/gctcli/helpers.go
index b23eb8794..e8412ebeb 100644
--- a/cmd/gctcli/helpers.go
+++ b/cmd/gctcli/helpers.go
@@ -35,18 +35,14 @@ func closeConn(conn *grpc.ClientConn, cancel context.CancelFunc) {
        }
 }

-// negateLocalOffset helps negate the offset of time generation
-// when the unix time gets to rpcserver, it no longer is the same time
-// that was sent as it handles it as a UTC value, even though when
-// using starttime it is generated as your local time
-// eg 2020-01-01 12:00:00 +10 will convert into
-// 2020-01-01 12:00:00 +00 when at RPCServer
-// so this function will minus the offset from the local sent time
-// to allow for proper use at RPCServer
+// negateLocalOffset converts a time value to a string in UTC format, correcting
+// for any local time offset.
 func negateLocalOffset(t time.Time) string {
-       return t.In(time.UTC).Format(common.SimpleTimeFormat)
+       return t.UTC().Format(common.SimpleTimeFormat)
 }

+// negateLocalOffsetTS converts a time value to a *timestamppb.Timestamp in UTC
+// format, correcting for any local time offset.
 func negateLocalOffsetTS(t time.Time) *timestamppb.Timestamp {
-       return timestamppb.New(t.In(time.UTC))
+       return timestamppb.New(t.UTC())
 }
```

thrasher-:
One other thing I noticed is that:

1) Our presentation of time variables uses time.Now() which will default to the local time (this is perfect)
2) However, since we use time.Parse with no timezone, this translates a supplied local time into a UTC time which is incorrect. (Example here: https://github.com/gloriousCode/gocryptotrader/blob/great-mate-negate/cmd/gctcli/commands.go#L4136). As a result, an improvement would be to switch all the time.Parses to `time.ParseInLocation` with `time.Local` as the last param to be consistent with our displayed user local times. Then negateOffset will ensure correct conversion to UTC time before usage with the RPC server

gloriousCode:
I've used `ParseInLocation` to parse the `time.Now()`'s in the CLI properly. The implementation of `timestamppb.New` uses Unix() so there was no need for `negateLocalOffsetPb(...)` and so I've removed it. I've also changed the timestamp parsing in the CLI and RPCserver side to include the timezone so that it can be parsed accurately and so I've removed `negateLocalOffset(...)`

thrasher- (approved):
Changes ACK, nice stuff!

shazbert (approved):
parsing works, looking good!

PR #1129, opened 2023-02-01
common: update Errors type

shazbert:
# PR Description

Unexport errors type to force inbuilt error interface usage across GCT, allows a better method of error type matching across multiple errors in line with the errors package handling instead of first match it will iterate across the entire list. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/backtester/eventhandlers/statistics/printresults.go
+++ b/backtester/eventhandlers/statistics/printresults.go
@@ -1,6 +1,7 @@
 package statistics
 
 import (
+	"errors"
 	"fmt"
 	"sort"
 	"time"
@@ -71,8 +72,7 @@ func (s *Statistic) PrintTotalResults() {
 // grouped by time to allow a clearer picture of events
 func (s *Statistic) PrintAllEventsChronologically() {
 	log.Info(common.Statistics, common.CMDColours.H1+"------------------Events-------------------------------------"+common.CMDColours.Default)
-	var errs gctcommon.Errors
-	var err error
+	var errs error
 	var results []eventOutputHolder
 	for _, exchangeMap := range s.ExchangeAssetPairStatistics {
 		for _, assetMap := range exchangeMap {
@@ -81,25 +81,26 @@ func (s *Statistic) PrintAllEventsChronologically() {
 					for i := range currencyStatistic.Events {
 						var result string
 						var tt time.Time
+						var err error
 						switch {
 						case currencyStatistic.Events[i].FillEvent != nil:
 							result, err = s.CreateLog(currencyStatistic.Events[i].FillEvent)
 							if err != nil {
-								errs = append(errs, err)
+								errs = gctcommon.AppendError(errs, err)
 								continue
 							}
 							tt = currencyStatistic.Events[i].FillEvent.GetTime()
 						case currencyStatistic.Events[i].SignalEvent != nil:
 							result, err = s.CreateLog(currencyStatistic.Events[i].SignalEvent)
 							if err != nil {
-								errs = append(errs, err)
+								errs = gctcommon.AppendError(errs, err)
 								continue
 							}
 							tt = currencyStatistic.Events[i].SignalEvent.GetTime()
 						case currencyStatistic.Events[i].DataEvent != nil:
 							result, err = s.CreateLog(currencyStatistic.Events[i].DataEvent)
 							if err != nil {
-								errs = append(errs, err)
+								errs = gctcommon.AppendError(errs, err)
 								continue
 							}
 							tt = currencyStatistic.Events[i].DataEvent.GetTime()
@@ -121,10 +122,10 @@ func (s *Statistic) PrintAllEventsChronologically() {
 			log.Info(common.Statistics, results[i].Events[j])
 		}
 	}
-	if len(errs) > 0 {
+	if errs != nil {
 		log.Info(common.Statistics, common.CMDColours.Error+"------------------Errors-------------------------------------"+common.CMDColours.Default)
-		for i := range errs {
-			log.Error(common.Statistics, errs[i].Error())
+		for err := errors.Unwrap(errs); err != nil; err = errors.Unwrap(errs) {
+			log.Error(common.Statistics, err.Error())
 		}
 	}
 }
@@ -206,7 +207,6 @@ func (s *Statistic) CreateLog(data common.Event) (string, error) {
 
 // PrintResults outputs all calculated statistics to the command line
 func (c *CurrencyPairStatistic) PrintResults(e string, a asset.Item, p currency.Pair, usingExchangeLevelFunding bool) {
-	var errs gctcommon.Errors
 	sort.Slice(c.Events, func(i, j int) bool {
 		return c.Events[i].Time.Before(c.Events[j].Time)
 	})
@@ -301,12 +301,6 @@ func (c *CurrencyPairStatistic) PrintResults(e string, a asset.Item, p currency.
 		realised = last.PNL.GetRealisedPNL()
 		log.Infof(common.CurrencyStatistics, "%s Final Unrealised PNL: %s", sep, convert.DecimalToHumanFriendlyString(unrealised.PNL, 8, ".", ","))
 		log.Infof(common.CurrencyStatistics, "%s Final Realised PNL: %s", sep, convert.DecimalToHumanFriendlyString(realised.PNL, 8, ".", ","))
-	}
-	if len(errs) > 0 {
-		log.Info(common.CurrencyStatistics, common.CMDColours.Error+"------------------Errors-------------------------------------"+common.CMDColours.Default)
-		for i := range errs {
-			log.Error(common.CurrencyStatistics, errs[i].Error())
-		}
 	}
 }
 
--- a/common/common.go
+++ b/common/common.go
@@ -461,8 +461,8 @@ func AppendError(original, incoming error) error {
 	return errSliceP
 }
 
-// multiError holds all the errors as a slice, this is unexported so it forces
-// ibuilt error handling.
+// multiError holds all the errors as a slice, this is unexported, so it forces
+// inbuilt error handling.
 type multiError struct {
 	loadedErrors []error
 	offset       *int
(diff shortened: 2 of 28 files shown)


gloriousCode:
I'll get to reviewing this later, but I sympathise with your timing. Golang 1.20 released today contains new error wrapping functionality with `errors.Join` https://pkg.go.dev/errors@master#Join The example snippet shows that `errors.Is` succeeds with checking individual errors within the wrapped error along with wrapping multiple errors via fmt.Errorf("%w %w", err1, err2)`. Though since 1.20 has been released only today, I don't know when GCT/our Actions suite will support this

common/common.go
```diff
@@ -433,28 +434,74 @@ func InArray(val, array interface{}) (exists bool, index int) {
...
 
-// Errors defines multiple errors
-type Errors []error
+// AppendError appends error in a more idiomatic way. This can start out as a
+// standard error e.g. err := errors.New("random error")
+// err = AppendError(err, errors.New("another random error"))
+func AppendError(original, incoming error) error {
+	if incoming == nil {
+		if _, file, no, ok := runtime.Caller(1); ok {
+			log.Errorf(log.Global, "AppendError() failed: incoming error is nil %s#%d\n", file, no)
```
gloriousCode, line 443: As fun as this is, this causes friction for say, appending an error where you don't yet know it something like 
`err = gctcommon.AppendError(err, okx.SendHTTPRequest(wow))`
Investigating the caller like this seems very debug thing to do. Are you doing this to dictate that one should not use this function like the snippet above? Even then, it seems somewhat unnecessary to call out and could be skipped and I'd rather you shame me in PR comments

For the sake of testing, I ran a benchmark:
```
BenchmarkButts-10    	 3082971	       387.7 ns/op	     288 B/op	       5 allocs/op
```

Without this `runtime` usage:
```
BenchmarkButts-10    	566423924	         2.035 ns/op	       0 B/op	       0 allocs/op
```

common/common.go
```diff
@@ -433,28 +434,74 @@ func InArray(val, array interface{}) (exists bool, index int) {
...
+		}
+		if errSlice, ok := original.(*multiError); ok {
+			errSlice.offset = 0
+		}
```
gloriousCode, line 447: Looks like errSlice is declared and you set offset = 0 in both paths. 
You could kinda sorta do the following, 
```
errSliceP, ok := original.(*multiError)
	if ok {
		errSliceP.offset = 0
	}
	if incoming == nil {
		return original // Skip append - continue as normal.
	}
	if !ok {
		// This assumes that a standard error is passed in and we can want to
		// track it and add additional errors.
		errSliceP = &multiError{}
		if original != nil {
			errSliceP.errors = append(errSliceP.errors, original)
		}
	}
	if incomingSlice, ok := incoming.(*multiError); ok {
		// Join slices if needed.
		errSliceP.errors = append(errSliceP.errors, incomingSlice.errors...)
	} else {
		errSliceP.errors = append(errSliceP.errors, incoming)
	}
	return errSliceP
```
But it doesn't flow as nicely. So this is pretty optional

backtester/eventhandlers/statistics/printresults.go
```diff
@@ -121,10 +122,10 @@ func (s *Statistic) PrintAllEventsChronologically() {
 			log.Info(common.Statistics, results[i].Events[j])
 		}
 	}
-	if len(errs) > 0 {
+	if errs != nil {
 		log.Info(common.Statistics, common.CMDColours.Error+"------------------Errors-------------------------------------"+common.CMDColours.Default)
-		for i := range errs {
-			log.Error(common.Statistics, errs[i].Error())
+		for err := errors.Unwrap(errs); err != nil; err = errors.Unwrap(errs) {
+			log.Error(common.Statistics, err.Error())
```
gloriousCode, line 128: This will duplicate print errors. a wrapped errs using err.Error() will print the comma delimited list. So for each error within errs, will print the whole list.
See this test:
```go
func TestLoop(t *testing.T) {
	t.Parallel()
	errOne := errors.New("one")
	errTwo := errors.New("two")
	errThree := errors.New("three")
	err := AppendError(errOne, errTwo)
	errs := AppendError(err, errThree)
	errs = AppendError(errs, errThree)
	for err := errors.Unwrap(errs); err != nil; err = errors.Unwrap(errs) {
		t.Log(err.Error())
	}
}
```
output:
```    
    common_test.go:822: one, two, three, three
    common_test.go:822: one, two, three, three
    common_test.go:822: one, two, three, three
```

gloriousCode (changes requested):
This works nicely! Given what we have discussed about forcing compatibility to go 1.20, I think this works well enough for now

common/common.go
shazbert: Well the design of it was supposed to be like var errs error and then you check the err for not nil then append if necessary. What you are doing above cleans up a bit of code though. So I am happy to just skip nils, I just wanted to be quite strict and alert and log if it was nil for things like if `err == nil { }` which was done by accident.

common/common.go
```diff
@@ -433,28 +433,78 @@ func InArray(val, array interface{}) (exists bool, index int) {
...
 }
 
-// Unwrap implements interface behaviour for errors.Is() matching NOTE: only
-// returns first element.
-func (e Errors) Unwrap() error {
-	if len(e) == 0 {
-		return nil
+// Unwrap increments the offset so errors.Is() can be called to its individual
+// error for correct matching.
+func (e *multiError) Unwrap() error {
```
gloriousCode, line 490: I'm getting a little confused on your implementation. I'm unsure how its working within Go
Check out the following, this does not touch your code:
```go
func TestUnwrap34(t *testing.T) {
	errOne := errors.New("one")
	errTwo := errors.New("two: ")
	errThree := errors.New("three: ")

	errTwoOnErrOne := fmt.Errorf("%v %w", errTwo, errOne)
	errThreeOnErrTwoOnErrOne := fmt.Errorf("%v %w", errThree, errTwoOnErrOne)
	errTwoUnwrappedFromThree := errors.Unwrap(errThreeOnErrTwoOnErrOne)
	t.Log(errTwoUnwrappedFromThree)
	errOneUnwrappedFromTwo := errors.Unwrap(errTwoUnwrappedFromThree)
	t.Log(errOneUnwrappedFromTwo)
	if errors.Is(errOneUnwrappedFromTwo, errOne) {
		t.Log("it sure is")
	}
}

```
However, using your way
```go
func TestUnwrap600(t *testing.T) {
	errOne := errors.New("one")
	errTwo := errors.New("two: ")
	errThree := errors.New("three: ")

	errTwoOnErrOne := AppendError(errOne, errTwo)
	errThreeOnErrTwoOnErrOne := AppendError(errTwoOnErrOne, errThree)
	errTwoUnwrappedFromThree := errors.Unwrap(errThreeOnErrTwoOnErrOne)
	if errors.Is(errTwoUnwrappedFromThree, errThree) {
		t.Logf("should I be error three?")
	}
}

```

gloriousCode (changes requested):
I think I'm confusing myself now, so here's my working. I'd appreciate the clarification

common/common.go
shazbert: Yeah I didn't see it necessary to remove the error when unwrapping, I guess I can return a minimized struct when calling unwrap if you want?

gloriousCode: That kind of logic is fine, we're doing a multi error after all. But then, why isn't the error also an error one?
```
func TestUnwrap600(t *testing.T) {
	errOne := errors.New("one")
	errTwo := errors.New("two: ")
	errThree := errors.New("three: ")

	errTwoOnErrOne := AppendError(errOne, errTwo)
	errThreeOnErrTwoOnErrOne := AppendError(errTwoOnErrOne, errThree)
	errTwoUnwrappedFromThree := errors.Unwrap(errThreeOnErrTwoOnErrOne)
	if !errors.Is(errTwoUnwrappedFromThree, errOne) {
		t.Logf("why aren't I err one?")
	}
}
```

shazbert: The way it unwraps it increments until it hits the last one then resets. So its designed to be unwrapped in errors package when calling errors.Is()

gloriousCode: I think I got a bit lost. Don't worry about this one, however the other comment may be a little bug to fix. Then I think its all good

common/common.go
```diff
@@ -433,28 +433,78 @@ func InArray(val, array interface{}) (exists bool, index int) {
...
+	e.offset++
+	if e.offset == len(e.loadedErrors) {
```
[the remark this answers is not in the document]
shazbert: 🦅

common/common.go
```diff
@@ -433,28 +433,78 @@ func InArray(val, array interface{}) (exists bool, index int) {
...
+	}
+	return errSliceP
+}
 
-// Error implements error interface
-func (e Errors) Error() string {
-	if len(e) == 0 {
-		return ""
+// multiError holds all the errors as a slice, this is unexported so it forces
+// ibuilt error handling.
```
gloriousCode, line 465:
```suggestion
// multiError holds all the errors as a slice, this is unexported, so it forces
// inbuilt error handling.
```

common/common.go
```diff
@@ -433,28 +433,78 @@ func InArray(val, array interface{}) (exists bool, index int) {
...
+		errSliceP.loadedErrors = append(errSliceP.loadedErrors, incoming)
+	}
+	return errSliceP
+}
 
-// Error implements error interface
-func (e Errors) Error() string {
-	if len(e) == 0 {
-		return ""
+// multiError holds all the errors as a slice, this is unexported so it forces
```
gloriousCode, line 464: Thoughts on moving the definition to before its used in the function above?

gloriousCode (changes requested):
I'm happy with all the work you've done :sunglasses: just two incredibly minor things

gloriousCode (approved):
```go
if errors.Is(multiErr, nil) {
    panic("tACK!")
}
```

thrasher- (approved):
tACK, nice helper func!

PR #1175, opened 2023-04-29
OKX: Fix balance using funding account

gbjk:
Fixes `UpdateAccountInfo` to use Trading account balances instead of Funding account.

It looks like this was an oversight.
No other exchange uses the funding account for balance and it wouldn't make sense.
Again, this aims to be minimally invasive and backwards compatible, so GetBalance is left alone to preserve compat.

There's one change that could be considered breaking, `EquityOfCurrency` is now just `Equity`

## Type of change

## How has this been tested

- [x] go test ./... -race
- [ ] golangci-lint run

## Checklist

### DIFF
--- a/exchanges/okx/okx_wrapper.go
+++ b/exchanges/okx/okx_wrapper.go
@@ -527,22 +527,22 @@ func (ok *Okx) UpdateAccountInfo(ctx context.Context, assetType asset.Item) (acc
 	if !ok.SupportsAsset(assetType) {
 		return info, fmt.Errorf("%w: %v", asset.ErrNotSupported, assetType)
 	}
-	balances, err := ok.GetBalance(ctx, "")
+	accountBalances, err := ok.GetNonZeroBalances(ctx, "")
 	if err != nil {
 		return info, err
 	}
-	currencyBalance := make([]account.Balance, len(balances))
-	for i := range balances {
-		free := balances[i].AvailBal
-		locked := balances[i].FrozenBalance
-		currencyBalance[i] = account.Balance{
-			Currency: currency.NewCode(balances[i].Currency),
-			Total:    balances[i].Balance,
-			Hold:     locked,
-			Free:     free,
+	currencyBalances := []account.Balance{}
+	for i := range accountBalances {
+		for j := range accountBalances[i].Details {
+			currencyBalances = append(currencyBalances, account.Balance{
+				Currency: currency.NewCode(accountBalances[i].Details[j].Currency),
+				Total:    accountBalances[i].Details[j].EquityOfCurrency.Float64(),
+				Hold:     accountBalances[i].Details[j].FrozenBalance.Float64(),
+				Free:     accountBalances[i].Details[j].AvailableBalance.Float64(),
+			})
 		}
 	}
-	acc.Currencies = currencyBalance
+	acc.Currencies = currencyBalances
 	acc.AssetType = assetType
 	info.Accounts = append(info.Accounts, acc)
 	creds, err := ok.GetCredentials(ctx)
--- a/exchanges/okx/okx.go
+++ b/exchanges/okx/okx.go
@@ -90,7 +90,7 @@ const (
 	oneClickRepayHistory          = "trade/one-click-repay-history"
 	oneClickRepay                 = "trade/one-click-repay"
 
-	// Funding orders routes
+	// Funding account routes
 	assetCurrencies    = "asset/currencies"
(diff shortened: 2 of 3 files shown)


exchanges/okx/okx_wrapper.go
```diff
@@ -520,22 +520,23 @@ func (ok *Okx) UpdateAccountInfo(ctx context.Context, assetType asset.Item) (acc
...
+	currencyBalances := []account.Balance{}
+	for _, a := range accountBalances {
+		// Not clear from API docs why we'd get multiple records back in this array; Asked Okx for clarification.
+		for _, d := range a.Details {
+			currencyBalances = append(currencyBalances, account.Balance{
+				Currency: currency.NewCode(d.Currency),
+				Total:    d.EquityOfCurrency.Float64(),
+				Hold:     d.FrozenBalance.Float64(),
+				Free:     d.AvailableBalance.Float64(),
+			})
```
gloriousCode, line 536:
```suggestion
	for i := range accountBalances {
		// Not clear from API docs why we'd get multiple records back in this array; Asked Okx for clarification.
		for j := range accountBalances[i].Details {
			currencyBalances = append(currencyBalances, account.Balance{
				Currency: currency.NewCode(accountBalances[i].Details[j].Currency),
				Total:    accountBalances[i].Details[j].EquityOfCurrency.Float64(),
				Hold:     accountBalances[i].Details[j].FrozenBalance.Float64(),
				Free:     accountBalances[i].Details[j].AvailableBalance.Float64(),
			})
```
I find via benchmarking that using a slice index rather than creating a temp local copy var is consistently more performant. Not super important

exchanges/okx/okx_wrapper.go
```diff
@@ -520,22 +520,23 @@ func (ok *Okx) UpdateAccountInfo(ctx context.Context, assetType asset.Item) (acc
...
-		locked := balances[i].FrozenBalance
-		currencyBalance[i] = account.Balance{
-			Currency: currency.NewCode(balances[i].Currency),
-			Total:    balances[i].Balance,
-			Hold:     locked,
-			Free:     free,
+	currencyBalances := []account.Balance{}
+	for _, a := range accountBalances {
+		// Not clear from API docs why we'd get multiple records back in this array; Asked Okx for clarification.
+		for _, d := range a.Details {
```
gloriousCode, line 530: Just seems a standard for their API and I don't think this comment is necessary

gloriousCode (changes requested):
Thanks for the catch and thanks for the fix!

shazbert (commented):
LGTM once other nits are addressed, ACK. Nice work!

exchanges/okx/okx_wrapper.go
gbjk: That's true, but note that it's not clear if `accountBalances` (`i`) will ever range. It might.
So we could range `i` and `j` to pre-alloc `currencyBalances`' slice cap, and then use len() or something,
but that's going to eat away the optimization time anyway.

Overall I figured the simplicity was more important here.

Let me know if I misunderstood.

exchanges/okx/okx_wrapper.go
gbjk: Happy to remove the comment.

Don't think we want to ignore the possibility of handling the array, though. Agreed?

gbjk:
Comments resolved.
Since they are all fish-in-a-barrel I've rebase fixup'd them against master for FF merging if easier.

exchanges/okx/okx_wrapper.go
gbjk: @gloriousCode This comment really caught me off guard, because I was sure at a gopher conference long ago I'd seen the complete opposite both explained and demonstrated.

Obviously it was true when you benched it, but I'm confident it hasn't been true since around go 1.8 at least.

My [https://go.dev/play/p/4_0cDM6Lcnk](own benchmark) shows alias vars are always faster on go 1.17 through to 1.20:
```
BenchmarkLocalAlias-8           1000000000               0.0000017 ns/op
BenchmarkIndex-8                1000000000               0.0000022 ns/op
```

Will raise in slack for better discussion.

shazbert (approved):
changes ACK. Thanks heaps for the PR!

thrasher- (approved):
tACK, matches balance in trading account. Thanks for the PR @gbjk !

image

gloriousCode (approved):
tACK! Thanks for the fix and showing me how little money I have in okx

PR #1186, opened 2023-05-06
OKX: Support exchange order limits

gbjk:
Adds support for PriceStepIncrementSize and MinAmount limits

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] ```go test -run TestUpdateOrderExecutionLimits exchanges/okx/*.go``` 

## Checklist

### DIFF
--- a/exchanges/okx/okx_test.go
+++ b/exchanges/okx/okx_test.go
@@ -1849,8 +1849,11 @@ func TestUpdateTradablePairs(t *testing.T) {
 
 func TestUpdateOrderExecutionLimits(t *testing.T) {
 	t.Parallel()
-	if err := ok.UpdateOrderExecutionLimits(context.Background(), asset.Spot); err != nil {
-		t.Error("Okx UpdateOrderExecutionLimits() error", err)
+
+	for _, a := range ok.GetAssetTypes(false) {
+		if err := ok.UpdateOrderExecutionLimits(context.Background(), a); err != nil {
+			t.Error("Okx UpdateOrderExecutionLimits() error", err)
+		}
 	}
 
 	for _, tt := range []struct {
--- a/exchanges/okx/okx_test.go
+++ b/exchanges/okx/okx_test.go
@@ -1847,6 +1847,35 @@ func TestUpdateTradablePairs(t *testing.T) {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


exchanges/okx/okx_test.go
```diff
@@ -1847,6 +1847,35 @@ func TestUpdateTradablePairs(t *testing.T) {
 	}
 }
 
+func TestUpdateOrderExecutionLimits(t *testing.T) {
+	t.Parallel()
+	if err := ok.UpdateOrderExecutionLimits(context.Background(), asset.Spot); err != nil {
```
shazbert, line 1852: could we please load all assets

```go
	assets := ok.GetAssetTypes(false)
	for x := range assets {
		if err := ok.UpdateOrderExecutionLimits(context.Background(), assets[x]); err != nil {
			t.Error("Okx UpdateOrderExecutionLimits() error", err)
		}
	}
```

and then check each one. Just so if the API changes in the future for one asset this might let us know where the issue is.

shazbert (changes requested):
Thanks for another awesome PR. Just one thing.

exchanges/okx/okx_test.go
gbjk: Agreed. I was a bit lazy with the reviews.

gbjk: That's done. ✅

gbjk: Actually, _and then check each one_.

Okay, yeah, can do.

gbjk:
I'll add further tests for more limits and all assets later today.

gbjk:
Okay. I think that's reasonably "_far enough_" on testing for okx's limit support.
It tests all asset types and it ensures at least one pair is tested.
It's possibly fragile due to the unpredictable futures and options, which might vary step size.
If that turns out to be the case we can home in on a specific pair contract and maturity date.
Probably not necessary though.

shazbert (approved):
tACK, looking good!

gloriousCode (approved):
tACK! Thanks for your contribution gbjk

thrasher- (approved):
tACK, thanks @gbjk 🚀

PR #1193, opened 2023-05-12
orders: Fix manager not processing immediately

gbjk:
This ensures that orders are processed as soon as orderManager claims to have started.
Specifically this will fix an eager call to CancelAllOrders not finding anything until after the first Interval run.

DriveBy:     orders: Fix reset timer call

This fix is small niggle.
The existing code was more complicated than it needed to be, whilst also ignoring the return of `timer.Reset` and the conditional need to drain the channel. It's unlikely but possible during transient resource starvation.
See [Timer docs](https://pkg.go.dev/time#Timer.Reset)

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -72,8 +73,12 @@ func (m *OrderManager) Start() error {
 	}
 	log.Debugln(log.OrderMgr, "Order manager starting...")
 	m.shutdown = make(chan struct{})
+	go func() {
+		m.processOrders()
+		m.run()
+	}()
+	runtime.Gosched()
 	m.orderStore.wg.Add(1)
-	go m.run()
 	return nil
 }
 
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -113,21 +113,16 @@ func (m *OrderManager) gracefulShutdown() {
 // run will periodically process orders
 func (m *OrderManager) run() {
 	log.Debugln(log.OrderMgr, "Order manager started.")
-	timer := time.NewTimer(orderManagerDelay)
 	for {
 		select {
 		case <-m.shutdown:
 			m.gracefulShutdown()
-			if !timer.Stop() {
-				<-timer.C
-			}
 			m.orderStore.wg.Done()
 			log.Debugln(log.OrderMgr, "Order manager shutdown.")
 			return
-		case <-timer.C:
+		case <-time.After(orderManagerInterval):
 			// Process orders go routine allows shutdown procedures to continue
 			go m.processOrders()
-			timer.Reset(orderManagerDelay)
 		}
 	}
 }
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"runtime"
 	"sort"
 	"strings"
 	"sync"
@@ -73,12 +72,8 @@ func (m *OrderManager) Start() error {
 	}
 	log.Debugln(log.OrderMgr, "Order manager starting...")
 	m.shutdown = make(chan struct{})
-	go func() {
-		m.processOrders()
-		m.run()
-	}()
-	runtime.Gosched()
 	m.orderStore.wg.Add(1)
+	go m.run()
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


gloriousCode (commented):
Good stuff on finding out why we don't know about orders for 10+ seconds after startup

engine/order_manager.go
```diff
@@ -108,21 +109,16 @@ func (m *OrderManager) gracefulShutdown() {
...
 		case <-m.shutdown:
 			m.gracefulShutdown()
-			if !timer.Stop() {
-				<-timer.C
-			}
 			m.orderStore.wg.Done()
 			log.Debugln(log.OrderMgr, "Order manager shutdown.")
 			return
-		case <-timer.C:
+		case <-time.After(orderManagerInterval):
```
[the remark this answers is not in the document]
gbjk: Happy to separate this out from the fix, because it's not atomic, it needs discussion, and the PR won't be atomic this 
 way.
 
The channel not draining during shutdown won't be an issue because it won't prevent or delay shutdown.
It just won't be GC'd, but that's a minor leak during a shutdown (IMO)

The reasoning for the issue I'm pointing to is that Reset *isn't* called directly.
The race would be that 10s occurs between the go m.processOrders and the Reset, due to context switching and then starvation or suspension.

I'm really not fussed though. If it's not a slam dunk let's drop it, it's a super edge case and very unlikely.
And the result wouldn't even be that bad.

Yanking it from this PR. Shout if I've swayed you and I should submit it in a separate PR 👍

engine/order_manager.go
```diff
@@ -73,6 +73,7 @@ func (m *OrderManager) Start() error {
 	log.Debugln(log.OrderMgr, "Order manager starting...")
 	m.shutdown = make(chan struct{})
 	m.orderStore.wg.Add(1)
+	m.processOrders()
```
gloriousCode, line 76: After our discussions, I do not mind the use of the `time.After`, however, the thing I'm concerned with here is that `m.ProcessOrders()` is not called via a go routine. In likely the majority of cases, this shouldn't be an issue, but if you have performed many orders for many currencies for many exchanges, this lack of `go` call blocks the application from closing and from cancelling orders on closure. If a user is trading frequently, the impacts of such limitations are large.

See the following: 
```go
        tt := time.Now()
	m.processOrders()
	fmt.Println(time.Since(tt))
```
Populating a few exchange API keys of my own, even with quite limited trading, I got `3.569271958s`. However, when FTX was alive, this function for me would take > 10 seconds and I would not consider myself a super-active trader.

I am concerned about how it can block versus the recommendation I gave in the previous review. 

An alternative could also be to do this before `m.orderStore.wg.Add(1)`, so then its not blocking the closure

gbjk:
@gloriousCode Okay, so I've put them both in the same goro, attempted to ensure that it starts before the order store wg gets added to, and then continue.
Happy to simplify, since I know this introduces some paradigms new to the codebase.

engine/order_manager.go
```diff
@@ -72,8 +73,12 @@ func (m *OrderManager) Start() error {
 	}
 	log.Debugln(log.OrderMgr, "Order manager starting...")
 	m.shutdown = make(chan struct{})
+	go func() {
+		m.processOrders()
+		m.run()
```
gloriousCode, line 78: Both being part of the same go routine technically still has the same issue, though it does alleviate startup. `m.processOrders()` has to finish before `m.run()` can begin and only via `m.run()`'s shutdown case does it execute `m.orderStore.wg.Done()` which allows GCT to shutdown with its `wg.Wait()`. Given the exact right circumstances, the application cannot shutdown until all orders for all exchanges have been processed

I've not encountered a situation where this is blocking. I think my fears were to do with the futures tracking which is not supported by any exchanges right now and that feature is much heavier in order cataloging. When that feature comes back, I may relook at this setup, but it's fine for now.

gloriousCode (approved):
tACK

engine/order_manager.go
gbjk: My solution to this for the last few years has been a ctx passed with a bubble down cancel and everything can check ctx.Err() or add `<-ctx.Done()` to an existing select in order to cancel.
I'd have that anywhere like inside process Orders.
Everything is aware then if it needs to abort cleanly.

( And then add another context.WithCancel for actions lower down that need to be discretely cancelled )

gbjk: @gloriousCode So are you suggesting we goro both independently?

gloriousCode: I was just outlining the issues and potential solutions with my review approval, you don't need to do anything. Your code investigations demonstrate that I'll need to look more thoroughly at the order manager

shazbert (approved):
tested ACK. Thanks.

thrasher- (approved):

PR #1196, opened 2023-05-15
dispatcher: prevent job limit errors with no subscribers

gloriousCode:
# PR Description
The GCT error log + dispatcher frequently errors for the following:
`Failed to get REST ticker. Error: dispatcher jobs at limit [100] current worker count [10]. Spawn more workers via --dispatchworkers=x, or increase the jobs limit via --dispatchjobslimit=x`
Not only does it look like something has gone wrong when it hasn't, like retrieving a ticker, its not really describing the issue

The current pipeline of `master` goes through the following:
- Receive ticker update
- Assign ticker a job ID if not exists
- Publish ticker as job to publisher
- Send job to jobs channel
- Receive job
- If any `routes` exist, push job data to subscribers
- If jobs channel is at capacity, throw an error for the calling function. eg Call `UpdateTicker` and receive an error unrelated to retrieving the ticker

This occurs for every ticker, orderbook update and account update, even when no subscription has been set, so we frequently see the error I mention above because we cannot push another ticker job to `d.jobs` while that channel is busy receiving and doing nothing. This creates noise and allows for jobs where a subscription that actually has been setup, eg via the CLI: `go run . gettickerstream kucoin btc-usdt spot` to get dropped per `dispatcher.go:175`

## Fixes
By introducing a check on whether `d.routes[id]` actually has any subscribers, we save on needing to push updates to a channel only to be dismissed. This reduces the time spent in the dispatcher, frees up memory space of `d.jobs`, reduces the amount of `dispatcher jobs at limit [100]` errors

## Type of change

## How has this been tested
Tested that subscriptions still function, that the error doesn't occur, that the new checks introduced do not add significant overhead

- [x] go test ./... -race
- [x] golangci-lint run
- [x] `TestHasListener`

## Checklist


dispatch/dispatch.go
```diff
@@ -216,6 +233,24 @@ func (d *Dispatcher) publish(id uuid.UUID, data interface{}) error {
...
+	if id.IsNil() {
+		return false, errChannelIsNil
+	}
+	d.rMtx.RLock()
+	defer d.rMtx.RUnlock()
+	if d.routes == nil {
+		return false, fmt.Errorf("%w routes property unset", errDispatcherNotInitialized)
+	}
+	routes := d.routes[id]
+	return len(routes) > 0, nil
```
shazbert, line 251: I would actually rm this function and move this to publish. Then change the field in `job` struct from ID to be this route slice, if length is zero then return nil like before. That way if there is routes it doesn't have a consecutive map lookup in the relayer routine. OR you could just change the default job slice to around ~10k length to then allow it to catch up. The map lookup adds small processing time to this caller routine and compounds across a bunch of orderbooks and ticker websocket updates etc. Up to you.

shazbert (changes requested):
Just a suggestion.

dispatch/dispatch.go
gloriousCode: I did test your first solution which came with its own complications with closed channels and panics and to protect that would have ended up worse.

The reason I don't want to just increase the length is that it increases the impacts behind the scenes - The jobs being created and sent into channels, routes being taken up unnecessarily and potentially taking away from actual subscriptions to publish and kicking the can a little bit.

I think I've come up with an even simpler solution! 0.6316n/s impact vs the 13n/s I had initially. Let me know what you think

dispatch/dispatch.go
```diff
@@ -287,6 +298,7 @@ func (d *Dispatcher) unsubscribe(id uuid.UUID, usedChan <-chan interface{}) erro
 		pipes[i] = pipes[len(pipes)-1]
 		pipes[len(pipes)-1] = nil
 		d.routes[id] = pipes[:len(pipes)-1]
+		atomic.AddInt32(&d.subscriberCount, -1)
```
shazbert, line 301: Suggestion: maybe a check to only decrement if its above 0? If you think that's unnecessary all good just resolve.

shazbert (approved):
Works fine tACK.

dispatch/dispatch.go
gloriousCode: I like the thought to ensure smooth running, but if it manages to go negative, it'll be a nice indicator to highlight that something else is amiss

thrasher- (approved):
Nice stuff! The benchmark results are 🚀

PR #1206, opened 2023-05-26
exchanges: Update OKCoin V5 exchange support

samuael:
This pull request incorporated a code implementation for version 5 of the **OKCoin** exchange, including support for REST and web socket. Extensive testing was done on all the Websocket and Wrapper functions. The code adheres to the style guidelines used by other exchanges of GCT. Additionally, I resolved some errors with the assistance of golangci-lint. Furthermore, I implemented endpoint methods that are currently unsupported by the GCT wrapper, to be utilized in the future.

Fixes # (issue)

## Type of change

- [ x ] Bug fix (non-breaking change which fixes an issue)
- [ x ] New feature (non-breaking change which adds functionality)
- [ x ] This change requires a documentation update

## How has this been tested

- [ x ] go test ./... -race
- [ x ] golangci-lint run
- [ x ] Test X

## Checklist

- [ x ] I have performed a self-review of my own code
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ x ] My changes generate no new warnings
- [ x ] I have added tests that prove my fix is effective or that my feature works


exchanges/okcoin/okcoin_types.go
```diff
@@ -1522,291 +525,840 @@ type WebsocketErrorResponse struct {
...
+}
+
+// ChannelInfo represents a channel information
+type ChannelInfo struct {
+	ChannelID            string       `json:"chanId"`
+	Currency             string       `json:"ccy"`
+	DepositQuota         string       `json:"depQuota"`
+	MinDeposit           okcoinNumber `json:"minDep"`
+	WithdrawalQuota      string       `json:"wdQuota"`
+	MinWithdrawal        string       `json:"minWd"`
```
shazbert, line 1150: float64 and below

exchanges/okcoin/okcoin_types.go
```diff
@@ -1522,291 +525,840 @@ type WebsocketErrorResponse struct {
...
+	FillTime                   okcoinMilliSec `json:"fillTime"`
+	InstrumentID               string         `json:"instId"`
+	InstrumentType             string         `json:"instType"`
+	Leverage                   okcoinNumber   `json:"lever"`
+	OrderID                    string         `json:"ordId"`
+	OrderType                  string         `json:"ordType"`
+	ProfitAndLoss              string         `json:"pnl"`
+	PosSide                    string         `json:"posSide"`
+	Price                      float64        `json:"px,string"`
+	Rebate                     string         `json:"rebate"`
```
shazbert, line 1235: float64?

exchanges/okcoin/okcoin_types.go
```diff
@@ -1522,291 +525,840 @@ type WebsocketErrorResponse struct {
...
+	InstrumentID             string         `json:"instId"`
+	InstrumentType           string         `json:"instType"`
+	Leverage                 string         `json:"lever"`
+	MoveTriggerPrice         okcoinNumber   `json:"moveTriggerPx"`
+	OrderID                  string         `json:"ordId"`
+	OrdPrice                 okcoinNumber   `json:"ordPx"`
+	OrderType                string         `json:"ordType"`
+	PosSide                  string         `json:"posSide"`
+	PriceLimit               okcoinNumber   `json:"pxLimit"`
+	PriceSpread              string         `json:"pxSpread"`
```
shazbert, line 1345: float64 and below.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -39,34 +83,53 @@ func (o *OKCoin) WsConnect() error {
...
+	go o.funnelWebsocketConn(o.Websocket.AuthConn)
+	o.Websocket.AuthConn.SetupPingHandler(stream.PingHandler{
+		Delay:       time.Second * 25,
+		Message:     []byte("ping"),
+		MessageType: websocket.TextMessage,
+	})
+	systemTime, err := o.GetSystemTime(context.Background())
+	if err != nil {
+		systemTime = time.Now().UTC()
+	}
```
shazbert, line 129: I don't think we should match with an external systems clock for sync.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -75,59 +138,126 @@ func (o *OKCoin) WsLogin(ctx context.Context) error {
...
+			if err != nil {
+				select {
+				case o.Websocket.DataHandler <- err:
+				default:
+					log.Errorf(log.WebsocketMgr,
+						"%s websocket handle data error: %v",
+						o.Name,
+						err)
+				}
+			}
```
shazbert, line 205: Remove?

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -75,59 +138,126 @@ func (o *OKCoin) WsLogin(ctx context.Context) error {
...
 		}
 	}
 }
 
 // WsHandleData will read websocket raw data and pass to appropriate handler
 func (o *OKCoin) WsHandleData(respRaw []byte) error {
+	if string(respRaw) == "pong" {
```
shazbert, line 212: bytes.Equal with a byte slice global.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -246,43 +532,50 @@ func (o *OKCoin) wsProcessOrder(respRaw []byte) error {
...
+	if err != nil {
+		return err
+	}
+	for x := range response {
+		pair, err := currency.NewPairFromString(response[x].InstrumentID)
+		if err != nil {
+			return err
 		}
-
 		o.Websocket.DataHandler <- &ticker.Price{
```
shazbert, line 564: suggestion: send a slice instead

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -621,34 +710,85 @@ func (o *OKCoin) GenerateDefaultSubscriptions() ([]stream.ChannelSubscription, e
 	var subscriptions []stream.ChannelSubscription
 	assets := o.GetAssetTypes(true)
 	for x := range assets {
+		if assets[x] != asset.Spot {
+			continue
+		}
```
shazbert, line 715: Seems like we can drop the for loop here as no other asset is being used currently? Or is there a pending TODO?

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -127,26 +122,22 @@ func (o *OKCoin) SetDefaults() {
...
-					// offset to the Asia/Shanghai time zone. This may lead to
-					// issues with candle quality and conversion as the
-					// intervals may be broken up. Therefore the below intervals
-					// are constructed from hourly candles.
-					// kline.IntervalCapacity{Interval: kline.SixHour},
-					// kline.IntervalCapacity{Interval: kline.TwelveHour},
-					// kline.IntervalCapacity{Interval: kline.OneDay},
-					// kline.IntervalCapacity{Interval: kline.ThreeDay},
-					// kline.IntervalCapacity{Interval: kline.OneWeek},
+					kline.IntervalCapacity{Interval: kline.SixHour},
```
shazbert, line 125: 💐

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -330,37 +332,43 @@ func (o *OKCoin) UpdateTradablePairs(ctx context.Context, forceUpdate bool) erro
...
+	enabledPairs, err := o.GetEnabledPairs(asset.Spot)
+	if err != nil {
+		return err
+	}
+	for p := range enabledPairs {
+		for i := range tickers {
+			cp, err := currency.NewPairFromString(tickers[i].InstrumentID)
+			if err != nil {
+				return err
+			}
```
shazbert, line 351: Use DeriveFrom method on available pairs and return an error. Then use contains method on the enabled pairs below. This is a bit more performant than calling new pair from string and it catches some malformed pairs.

shazbert (changes requested):
Great start for updating this!

exchanges/okcoin/okcoin_wrapper.go
samuael: ?

exchanges/okcoin/okcoin_convert.go
```diff
@@ -0,0 +1,78 @@
...
+func (a *okcoinNumber) UnmarshalJSON(data []byte) error {
+	var value interface{}
+	err := json.Unmarshal(data, &value)
+	if err != nil {
+		return err
+	}
+	switch val := value.(type) {
+	case string:
+		if val == "" {
+			*a = okcoinNumber(0)
```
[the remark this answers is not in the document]
samuael: I added this because if we try to **_unmarshal_** a JSON data into a struct more than once and if the _last_  JSON has "" empty string in that okcoinNumber field, we have to reset the value of that okcoinNumber instance to 0.

exchanges/okcoin/okcoin_types.go
```diff
@@ -1,1453 +1,465 @@
...
+	Bids      [][]okcoinNumber `json:"bids"` // [ Price, Quantity, depreciated, number of orders at the price ]
+	Timestamp okcoinMilliSec   `json:"ts"`
+}
+
+func (a *WebsocketOrderBook) prepareOrderbook() {
+	asks := [][]okcoinNumber{}
+askLoop:
+	for x := range a.Asks {
+		if a.Asks[x][1].Float64() != 0 {
+			for i := 0; i < len(asks); i++ {
```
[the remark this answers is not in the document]
samuael: > 1. If there is the same price, compare the size. If the size is 0, delete this depth data. If the size changes, replace the original data.
> 2. If there is no same price, sort by price (bid in descending order, ask in ascending order), and insert the depth information into the full load.

exchanges/okcoin/okcoin_types.go
```diff
@@ -1,1453 +1,465 @@
...
+		timestamp, err = strconv.ParseInt(a.Data[x][0], 10, 64)
+		if err != nil {
+			return nil, err
+		}
+		candlesticks[x] = stream.KlineData{
+			AssetType: asset.Spot,
+			Pair:      cp,
+			Timestamp: time.UnixMilli(timestamp),
+			Exchange:  exchangeName,
+		}
```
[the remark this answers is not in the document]
samuael: I don't think I understand your point

exchanges/okcoin/okcoin_websocket.go
samuael:
```
case []ticker.Price:
		for x := range d {
			if m.syncer.IsRunning() {
				err := m.syncer.Update(exchName,
					d[x].Pair,
					d[x].AssetType,
					SyncItemTicker,
					nil)
				if err != nil {
					return err
				}
			}
			err := ticker.ProcessTicker(&d[x])
			if err != nil {
				return err
			}
			m.syncer.PrintTickerSummary(&d[x], "websocket", err)
		}
```
If so, I also add this to **_websocketroutine_manager.go_**

exchanges/okcoin/okcoin_websocket.go
samuael: No, I left it like that Incase other asset types are added. But, I think it is not necessary.
Removed!

exchanges/okcoin/okcoin_convert.go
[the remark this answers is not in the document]
shazbert: Oh right, if you can add a comment for that, that would be awesome. 👍

exchanges/okcoin/okcoin_wrapper.go
shazbert: Flowers for your awesome update.

exchanges/okcoin/okcoin_websocket.go
shazbert: Yes please

exchanges/okcoin/okcoin_test.go
```diff
@@ -77,1251 +80,1670 @@ func TestStart(t *testing.T) {
...
-	update := `{"table":"spot/depth","action":"update","data":[{"instrument_id":"BTC-USD","asks":[["3864.6786","0",0],["3864.9852","0",0],["3865.9994","0.48402971",1],["3866.4004","0.001",1],["3866.7995","0.3273",2],["3867.4566","0",0],["3867.7031","0.025",1],["3868.0436","0",0],["3868.346","0",0],["3868.3695","0.051",1],["3870.9243","0.642",1],["3874.9942","0.51751796",1],["3875.7057","0",0],["3939","0.001",1]],"bids":[["3864.55","0.0565449",1],["3863.8282","0",0],["3863.8153","0",0],["3863.7898","0.01320077",1],["3863.4807","0.02112123",1],["3863.3002","0.04233533",1],["3863.1717","0.03379397",1],["3863.0685","0.04438179",1],["3863.0286","0.7362564",1],["3862.9912","0.06773651",1],["3862.8626","0.05407035",1],["3862.7595","0.07101087",1],["3862.313","0.3756",2],["3862.1848","0.012",1],["3862.0734","0",0],["3861.8391","0.025",1],["3861.7888","0",0],["3856.6716","0.38893641",1],["3768","0",0],["3766.66","0",0],["3766","0",0],["3765.58","0",0],["3762.58","0",0],["3761","0",0],["3760.8031","0",0],["3760.7777","0",0]],"timestamp":"2019-03-06T23:19:18.239Z","checksum":-1587788848}]}`
```
gloriousCode, line 956: I am noting the removal of orderbook checksum tests. I would like to see them return because they are still relevant

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -605,89 +594,160 @@ func (o *OKCoin) FetchAccountInfo(ctx context.Context, assetType asset.Item) (ac
...
+	var err error
+	req.Pair, err = o.FormatExchangeCurrency(req.Pair, asset.Spot)
+	if err != nil {
+		return nil, err
+	}
+	err = req.Validate()
+	if err != nil {
+		return nil, err
+	}
+	request := &AmendTradeOrderRequestParam{
```
gloriousCode, line 736: rename request as it collides with a package name

Please do this across this file

gloriousCode (changes requested):
Thanks for updaing this. Here is an initial high level review to catch the easy things

README.md
```diff
@@ -40,7 +40,7 @@ Join our slack to discuss all things related to GoCryptoTrader! [GoCryptoTrader
 | ItBit | Yes | NA | No |
 | Kraken | Yes | Yes | NA |
 | Lbank | Yes | No | NA |
-| OKCoin International | Yes | Yes | No |
+| OKCOIN International | Yes | Yes | No |
```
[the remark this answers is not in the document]
samuael: Alright, sir.
Change applied!

currency/forexprovider/exchangerate.host/README.md
```diff
@@ -1,4 +1,4 @@
-# GoCryptoTrader package Exchangerate.host
+# GoCryptoTrader package Exchangerate.Host
```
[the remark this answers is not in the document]
samuael: I am not sure🤔. but I will revert it.

exchanges/okcoin/okcoin_test.go
```diff
@@ -77,1251 +80,1670 @@ func TestStart(t *testing.T) {
...
+			Price:     123456.44,
+			Amount:    123,
+		})
+	if err != nil {
 		t.Error(err)
 	}
 }
 
-func TestCancelAllExchangeOrders(t *testing.T) {
+func TestCancelOrder(t *testing.T) {
```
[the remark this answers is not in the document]
samuael: This is because the web-socket endpoint sends a short message like "Operation failed" instead of "order not found" or such similar error messages. Same for the above. After I fix the **_CancelBatchOrder_** method, It is returning `okcoin websocket error code: 1 message: Operation failed.` without a specific error message.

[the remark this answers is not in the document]
gloriousCode: I brought these up because the other functions were much more specific about not being able to find the order rather than operation failed. I can reverify, but my conclusion was that anything with that error had something else going wrong

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -652,12 +646,13 @@ func (o *OKCoin) AppendWsOrderbookItems(entries [][]interface{}) ([]orderbook.It
...
 }
 
-// CalculatePartialOrderbookChecksum alternates over the first 25 bid and ask
+// CalculateChecksum alternates over the first 25 bid and ask
 // entries from websocket data. The checksum is made up of the price and the
 // quantity with a semicolon (:) deliminating them. This will also work when
 // there are less than 25 entries (for whatever reason)
 // eg Bid:Ask:Bid:Ask:Ask:Ask
-func (o *OKCoin) CalculatePartialOrderbookChecksum(orderbookData *WebsocketOrderBook) (uint32, error) {
+func (o *OKCoin) CalculateChecksum(orderbookData *WebsocketOrderBook) (uint32, error) {
```
shazbert, line 654: NOTE 4 ME: Check.

shazbert (changes requested):
Thanks for updating that. Things I have found on recent changes.

exchanges/okcoin/okcoin_ratelimit.go
```diff
@@ -0,0 +1,467 @@
...
+	case masterAccountsManageTransfersBetweenSubaccountEPL:
+		return r.MasterAccountsManageTransfersBetweenSubaccount.Wait(ctx)
+	case getTickersEPL:
+		return r.GetTickers.Wait(ctx)
+	case getTickerEPL:
+		return r.GetTicker.Wait(ctx)
+	case getOrderbookEPL:
+		return r.GetOrderbook.Wait(ctx)
+	case getOrderliteBookEPL:
+		return r.GetOrderliteBook.Wait(ctx)
```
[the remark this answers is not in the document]
samuael: And now they said it is removed from the API. So, I removed the rate limiter.

exchanges/okcoin/okcoin_test.go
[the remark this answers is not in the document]
samuael: I tried those endpoints with an `"Operation failed"` by disabling the WebSocket; the errors were like `order not found` or `insufficient Balance` and some similar messages.

If needed I can also add unit test which tests the REST endpoint by disabling the web-socket

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -198,11 +190,21 @@ func (o *OKCoin) Setup(exch *config.Exchange) error {
 	if err != nil {
 		return err
 	}
-
-	return o.Websocket.SetupNewConnection(stream.ConnectionSetup{
+	err = o.Websocket.SetupNewConnection(stream.ConnectionSetup{
 		RateLimit:            okcoinWsRateLimit,
 		ResponseCheckTimeout: exch.WebsocketResponseCheckTimeout,
 		ResponseMaxLimit:     exch.WebsocketResponseMaxLimit,
+		URL:                  okCoinWebsocketURL,
```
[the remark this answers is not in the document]
samuael: I think I don't get this.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -666,64 +772,117 @@ func (o *OKCoin) Unsubscribe(channelsToUnsubscribe []stream.ChannelSubscription)
...
+				return err
+			}
+		} else {
+			chunk, err = json.Marshal(request)
+			if err != nil {
+				return err
+			}
 		}
 
 		if len(chunk) > maxConnByteLen {
```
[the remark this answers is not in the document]
gloriousCode: Thanks! 😄 I foolishly didn't consider commas in the number

exchanges/okcoin/okcoin_test.go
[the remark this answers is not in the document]
gloriousCode: Thank you for investigating! That won't be necessary so long as the functions that the wrapper functions call have their own tests which looks to be the case 🤙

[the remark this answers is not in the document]
gloriousCode: Actually, there's sub error messages for websocket issues:
```
{"id":"189817431","op":"order","code":"1","msg":"","data":[{"tag":"","ordId":"","clOrdId":"yeneOrder","sCode":"51008","sMsg":"Order failed. Insufficient USD balance in account"}]}
```
Could you include the sMsg error in the response rather than just the `code` of `1`?

exchanges/okcoin/okcoin_wrapper.go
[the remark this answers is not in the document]
gloriousCode: Sorry for the confusion. For other exchanges, the URL here should be retrieved via 
`ePoint, err := o.API.Endpoints.GetURL(exchange.WebsocketSpot)`

You have already set the URL via the call inside `SetDefaults()` with 
```go
err = o.API.Endpoints.SetDefaultEndpoints(map[exchange.URL]string{
		exchange.RestSpot:      okcoinAPIURL,
		exchange.WebsocketSpot: okcoinWebsocketURL,
	})
```

What this does is allow a user to input their own URL in their config to override the default one set if a different URL needs to be used

exchanges/okcoin/okcoin_wrapper.go
gloriousCode: Thanks for that. There are two more inside of the websocket file:
image

exchanges/okcoin/okcoin_wrapper.go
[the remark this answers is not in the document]
samuael: I see. Thanks, sir.

exchanges/okcoin/okcoin_test.go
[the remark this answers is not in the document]
samuael: Ohh, I don't know why I didn't see this.
Change applied. Thanks!

exchanges/okcoin/okcoin_test.go
samuael: The order book data checksum value they send with the push data may not be correct. I tried multiple combinations following their rule but couldn't get the result.

gloriousCode: Even more of a reason to have tests to demonstrate that it doesn't work. Would you like some help going over it?

gloriousCode: I've sent you a diff.diff to apply that should get the checksum validation working. _It works on my machine_ 🤞

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1656 @@ import (
...
+		return nil, errMissingInstrumentID
 	}
-	var resp []WithdrawalHistoryResponse
-	return resp, o.SendHTTPRequest(ctx, exchange.RestSpot, http.MethodGet, accountSubsection, requestURL, nil, &resp, true)
+	params := url.Values{}
+	params.Set("instId", instrumentID)
+	var err error
+	if interval != kline.Interval(0) {
+		var intervalString string
+		intervalString, err = intervalToString(interval, false)
```
gloriousCode, line 183: Please make the `bool` of `false` a param, because wrapper functions `GetHistoricCandles` and `GetHistoricCandlesExtended` need it to be `true` for UTC alignment

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1656 @@ import (
...
+func (o *Okcoin) GetCandlestickHistory(ctx context.Context, instrumentID string, after, before time.Time, bar kline.Interval, limit int64) ([]CandlestickData, error) {
+	if instrumentID == "" {
+		return nil, errMissingInstrumentID
+	}
+	params := url.Values{}
+	params.Set("instId", instrumentID)
+	var err error
+	if bar != kline.Interval(0) {
+		var intervalString string
+		intervalString, err = intervalToString(bar, false)
```
gloriousCode, line 216: Please make the `bool` of `false` a param, because wrapper functions `GetHistoricCandles` and `GetHistoricCandlesExtended` need it to be `true` for UTC alignment

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,236 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+	var resp WebsocketAccount
+	err := json.Unmarshal(respRaw, &resp)
+	if err != nil {
+		return err
+	}
+	o.Websocket.DataHandler <- &resp
+	return nil
+}
+
+func (o *Okcoin) wsProcessOrderbook(respRaw []byte) error {
```
gloriousCode, line 331: I have recieved the following:
` exchange okcoin websocket error - websocket orderbook buffer error: update bid/ask targets cannot be nil`

The error could be more specific, which pair? which asset? Is the error a result of our processing? Or is it that okcoin is sending bad data?

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -29,39 +28,36 @@ import (
...
 			return nil, err
 		}
 	}
-
 	return exchCfg, nil
 }
 
 // SetDefaults method assigns the default values for OKCoin
-func (o *OKCoin) SetDefaults() {
+func (o *Okcoin) SetDefaults() {
```
gloriousCode, line 52: Is `Margin` supported by the websocket? If not, you'll need to add 
```go
	err = b.DisableAssetWebsocketSupport(asset.Margin)
	if err != nil {
	}
```
So you don't get the following:
```
[WARN]  | SYNC | 14/07/2023 11:15:42 | okcoin USDC-USDT SPOT: No orderbook update after 1m0s, switching from websocket to rest
[WARN]  | SYNC | 14/07/2023 11:15:42 | okcoin BTC-USD MARGIN: No ticker update after 1m0s, switching from websocket to rest
[ERROR] | SYNC | 14/07/2023 11:15:42 | Failed to get REST ticker. Error: unsupported asset type, asset: margin
```

gloriousCode (changes requested):
Nice work Sam!

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -5,7 +5,6 @@ import (
 	"fmt"
```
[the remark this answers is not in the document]
samuael: Well Implemented!

exchanges/order/order_types.go
```diff
@@ -71,6 +71,8 @@ type Submit struct {
 	// RetrieveFeeDelay some exchanges take time to properly save order data
 	// and cannot retrieve fees data immediately
 	RetrieveFeeDelay time.Duration
+	// TradeMode specifies the trading mode for margin and non-margin orders: see okcoin_wrapper.go
+	TradeMode string
```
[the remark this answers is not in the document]
samuael: Note For S*

Yes, Non-Margin mode: 'cash' is also missing from the list of margin types. So, That would be fine if you can add "cash" alongside 'isolated', and 'cross'.

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1656 @@ import (
...
+	if limit > 0 {
+		params.Set("limit", strconv.FormatInt(limit, 10))
+	}
+	var resp []candlestickItemResponse
+	err = o.SendHTTPRequest(ctx, exchange.RestSpot, getCandlesticksEPL, http.MethodGet, typeMarket, common.EncodeURLValues("candles", params), nil, &resp, false)
+	if err != nil {
+		return nil, err
+	}
+	return ExtractCandlesticks(resp)
 }
```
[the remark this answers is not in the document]
samuael: I think I wrote about this endpoint before. The order book lite endpoint is removed from their API. It is just the documentation is not updated.
Screenshot 2023-07-16 at 12 01 19 PM

exchanges/okcoin/okcoin_websocket.go
samuael: They have such push data below.
`{"arg":{"channel":"books","instId":"BTC-USD"},"action":"update","data":[{"asks":[],"bids":[],"ts":"1689510449762","checksum":1085589561}]}`

I added a check to skip such data. Thanks for pointing out.

exchanges/okcoin/okcoin_wrapper.go
samuael: Change applied, sir. Thanks for reminding me.

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -29,46 +28,46 @@ import (
...
 	o.API.CredentialsValidator.RequiresClientID = true
-
 	requestFmt := &currency.PairFormat{Uppercase: true, Delimiter: currency.DashDelimiter}
 	configFmt := &currency.PairFormat{Uppercase: true, Delimiter: currency.DashDelimiter}
 	err := o.SetGlobalPairsManager(requestFmt, configFmt, asset.Spot, asset.Margin)
 	if err != nil {
 		log.Errorln(log.ExchangeSys, err)
 	}
-
+	err = o.DisableAssetWebsocketSupport(asset.Margin)
```
gloriousCode, line 67: Is margin actually supported? It actually doesn't seem like it. The documentation constantly has the following:
`Not enabled. Please disregard.`

Endpoints such as `https://www.okcoin.com/api/v5/public/instruments?instType=MARGIN` return no data, your ticker & orderbook data is sourced from SPOT data and there is no margin type for websocket

Unfortunately, I am unable to log in to okcoin on any of my accounts/new account to fully verify, but I'm leaning towards its not supported.

gloriousCode (changes requested):
Thanks for your updates! Two main things I've come across today

exchanges/okcoin/okcoin.go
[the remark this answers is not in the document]
gloriousCode: Thanks Sam for doing that!

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,236 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+			TradeID:       resp.Data[x].TradeID,
+			Price:         resp.Data[x].FillPrice,
+			Amount:        resp.Data[x].Size,
+		}
+	}
+	o.Websocket.DataHandler <- algoOrder
+	return nil
+}
+
+func (o *Okcoin) wsProcessAccount(respRaw []byte) error {
```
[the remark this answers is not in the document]
gloriousCode: This could still use a coverage test unless there is something wrong?

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,236 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
 	}
 	return nil
 }
 
-// wsProcessTickers converts ticker data and sends it to the datahandler
-func (o *OKCoin) wsProcessTickers(respRaw []byte) error {
-	var response WebsocketTickerData
-	err := json.Unmarshal(respRaw, &response)
+// wsProcessInstruments converts instrument data and sends it to the datahandler
+func (o *Okcoin) wsProcessInstruments(respRaw []byte) error {
```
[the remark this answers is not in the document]
gloriousCode: This could still use a coverage test unless there is something wrong?

exchanges/okcoin/okcoin_wrapper.go
samuael: I added that because the exchange uses margin(trade) modes like "cash", "isolated", "cross"

gloriousCode: I'll try and get access to okcoin today to see if it has any margin abilities

gloriousCode: image
Confirmed in writing. Thanks for already removing support!

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -1125,71 +1069,172 @@ func (o *OKCoin) GetFeeByType(ctx context.Context, feeBuilder *exchange.FeeBuild
...
+		if err != nil {
+			return err
+		}
+		limits = append(limits, order.MinMaxLevel{
+			Asset:                  a,
+			Pair:                   pair,
+			PriceStepIncrementSize: instrumentsList[index].TickSize.Float64(),
+			MinimumBaseAmount:      instrumentsList[index].MinSize.Float64(),
+			MaxIcebergParts:        instrumentsList[index].MaxIcebergSz.Int64(),
+			MarketMinQty:           instrumentsList[index].MaxMarketSize.Float64(),
```
gloriousCode, line 1232: This is for minimum quantity, please rm this line

exchanges/okcoin/okcoin_ws_trade.go
```diff
@@ -0,0 +1,191 @@
...
+	"reflect"
+	"strconv"
+	"sync"
+
+	"github.com/thrasher-corp/gocryptotrader/exchanges/stream"
+)
+
+var (
+	waitingSignatureLock sync.Mutex
+	waitingSignatures    = []string{}
```
gloriousCode, line 16: I'm not seeing any removal of signatures in code. This will cause an endlessly appending list which will get slower over time

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,268 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+			o.Websocket.DataHandler <- eventResponse
+		case "error":
+			waitingSignatureLock.Lock()
+			for x := range waitingSignatures {
+				if strings.Contains(dataResponse.Message, waitingSignatures[x]) {
+					o.Websocket.Match.IncomingWithData(waitingSignatures[x], respRaw)
+					return nil
+				}
+			}
+			waitingSignatureLock.Unlock()
```
gloriousCode, line 256: Why do we only want to match the response with errors? I'm not seeing `waitingSignatures` match anywhere else.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,268 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
-			if o.Websocket.Match.Incoming("login") {
-				o.Websocket.SetCanUseAuthenticatedEndpoints(eventResponse.Success)
+		switch eventResponse.Event {
+		case "login":
+			if o.Websocket.Match.IncomingWithData("login", respRaw) {
+				o.Websocket.SetCanUseAuthenticatedEndpoints(eventResponse.Code == "0")
+			}
+		case "subscribe", "unsubscribe":
+			o.Websocket.DataHandler <- eventResponse
+		case "error":
```
gloriousCode, line 248: With `waitingSignatures` I'm not seeing any possibility of the event type being "error":
https://www.okcoin.com/docs-v5/en/#websocket-api-trade-cancel-order

gloriousCode (changes requested):
Thanks for those updates! I can confirm okcoin successfully resubscribes upon a timeout due to illiquidity

exchanges/okcoin/okcoin_ws_trade.go
samuael: That's a good suggestion thanks. I will use map[string]struct{} for that.

exchanges/okcoin/okcoin_websocket.go
samuael: Yes, I had to remove the signature from the list above when response data with a non-empty ID is found like below.
```
if dataResponse.ID != "" {
		waitingSignatureLock.Lock()
		delete(waitingSignatures, dataResponse.ID)
		waitingSignatureLock.Unlock()
		if !o.Websocket.Match.IncomingWithData(dataResponse.ID, respRaw) {
			return fmt.Errorf("couldn't match incoming message with id: %s and operation: %s", dataResponse.ID, dataResponse.Operation)
		}
		return nil
}
```
👍🏽.

exchanges/okcoin/okcoin_websocket.go
samuael: Embarrassingly, The error event I tried to handle is found to be for "subscriptions", not for the WebSocket "trade". The subscriptions WebSocket class doesn't use their signature, therefore, I removed the `waitingSignatureLock, waitingSignature` implementation.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -55,10 +55,10 @@ const (
...
 
 var defaultSubscriptions = []string{
-	wsTickers,
-	wsCandle1D,
-	wsTrades,
-	wsOrderbooks,
+	// wsTickers,
+	// wsCandle1D,
+	// wsTrades,
+	// wsOrderbooks,
```
gloriousCode, line 61: 👀

samuael: reverted🥲

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1658 @@ import (
...
+	return resp, o.SendHTTPRequest(ctx, exchange.RestSpot, requestQuotesEPL, http.MethodPost, typeOtc, "rfq/quote", arg, &resp, true)
 }
 
-// GetSpotAllTokenPairsInformation Get the last traded price, best bid/ask price, 24 hour trading volume and more info of all trading pairs.
-func (o *OKCoin) GetSpotAllTokenPairsInformation(ctx context.Context) ([]GetSpotTokenPairsInformationResponse, error) {
-	requestURL := instruments + "/" + tickerData
-	var resp []GetSpotTokenPairsInformationResponse
-	return resp, o.SendHTTPRequest(ctx, exchange.RestSpot, http.MethodGet, tokenSubsection, requestURL, nil, &resp, false)
+// PlaceRFQOrder submit RFQ order
+func (o *Okcoin) PlaceRFQOrder(ctx context.Context, arg *PlaceRFQOrderRequest) ([]RFQOrderResponse, error) {
```
gloriousCode, line 819: Missing field `clTReqTs`
https://www.okcoin.com/docs-v5/en/#rest-api-otc-desk-rfq-place-order

exchanges/okcoin/okcoin_types.go
```diff
@@ -1533,291 +471,930 @@ type WebsocketErrorResponse struct {
...
+	Side                        string        `json:"side"`
+	RfqSize                     float64       `json:"rfqSz,string"` // Amount
+	RfqSzCurrency               currency.Code `json:"rfqSzCcy"`     // Token
+	ClientDefinedQuoteRequestID string        `json:"clQReqId,omitempty"`
+	ClientRequestTimestamp      string        `json:"clQReqTs,omitempty"`
+}
+
+// RFQQuoteResponse query current market quotation information
+type RFQQuoteResponse struct {
+	QuoteTimesamp okcoinTime              `json:"quoteTs"`
```
gloriousCode, line 945:
```suggestion
	QuoteTimestamp okcoinTime              `json:"quoteTs"`
```

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1658 @@ import (
...
+		params.Set("chanId", channelID)
+	}
+	if !ccy.IsEmpty() {
+		params.Set("ccy", ccy.String())
+	}
+	if withdrawalState != "" {
+		params.Set("state", withdrawalState)
+	}
+	if withdrawalID != "" {
+		params.Set("depId", withdrawalID)
```
gloriousCode, line 1003: Should be `wdId`:
https://www.okcoin.com/docs-v5/en/#rest-api-fiat-get-withdrawal-history

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1658 @@ import (
...
+	}
+	if a.Size <= 0 {
+		return fmt.Errorf("%w, size: %f", errInvalidAmount, a.Size)
+	}
+	return nil
+}
 
-	err := o.SendHTTPRequest(ctx, exchange.RestSpot, http.MethodPost, marginTradingSubsection, batchOrders, request, &resp, true)
+// PlaceOrder to place a trade order.
+func (o *Okcoin) PlaceOrder(ctx context.Context, arg *PlaceTradeOrderParam) (*TradeOrderResponse, error) {
```
gloriousCode, line 1179: Client order id is optional:
https://www.okcoin.com/docs-v5/en/#rest-api-trade-place-order
Use omitempty

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1658 @@ import (
...
+			return nil, err
 		}
 	}
+	var resp []TradeOrderResponse
+	return resp, o.SendHTTPRequest(ctx, exchange.RestSpot, placeTradeMultipleOrdersEPL, http.MethodPost, typeTrade, "batch-orders", &args, &resp, true)
+}
 
-	return resp, orderErrors
+// CancelTradeOrder cancels a single trade order
+func (o *Okcoin) CancelTradeOrder(ctx context.Context, arg *CancelTradeOrderRequest) (*TradeOrderResponse, error) {
```
gloriousCode, line 1215: orderID and clientOtderID are optional, your work here is correct, just add omitempty to their fields in the request

gloriousCode (changes requested):
Thanks for those changes. I have gone through the incredibly fun process of verifying all REST endpoints. Next review will verify websocket and wrappers and then everything should be good to go!

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -24,8 +25,53 @@ import (
...
+	wsOrder       = "orders"
+	wsOrdersAlgo  = "orders-algo"
+	wsAlgoAdvance = "algo-advance"
+)
+
+var defaultSubscriptions = []string{
+	wsTickers,
+	wsCandle1D,
+	wsTrades,
+	wsOrderbooks,
```
gloriousCode, line 61: I think this could be changed to `wsOrderbookTickByTickL400` because its faster

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -299,468 +500,326 @@ func (o *OKCoin) wsProcessTrades(respRaw []byte) error {
...
+		)
+	}
+	for s := range channels {
+		switch channels[s] {
+		case wsInstruments:
+			subscriptions = append(subscriptions, stream.ChannelSubscription{
+				Channel: channels[s],
+				Asset:   asset.Spot,
+			})
+		case wsTickers, wsTrades, wsOrderbooks, wsOrderbooksL5, wsOrderbookL1, wsOrderbookTickByTickL50,
```
gloriousCode, line 642: Having switched the default orderbook to `wsOrderbookTickByTickL400`, I'm not getting orderbooks, please ensure the various endpoints work when subscribed

exchanges/okcoin/okcoin_ws_trade.go
```diff
@@ -0,0 +1,182 @@
...
+		Operation: operation,
+	}
+	if reflect.TypeOf(data).Kind() == reflect.Slice {
+		req.Arguments = data
+	} else {
+		req.Arguments = []interface{}{data}
+	}
+	var byteData []byte
+	var err error
+	if authenticated {
```
gloriousCode, line 154: Please add a `// TODO: ratelimits for websocket`

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,260 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
-		var oType order.Type
-		var oSide order.Side
-		var oStatus order.Status
-		oType, err = order.StringToOrderType(resp.Data[i].Type)
+	cp, err := currency.NewPairFromString(resp.Arg.InstrumentID)
+	if err != nil {
+		return err
+	}
+
+	algoOrder := make([]fill.Data, len(resp.Data))
```
gloriousCode, line 289: Shouldn't this be order.Detail? The endpoint updates when an order is placed, which doesn't necessarily mean a fill eg a limit order

Also, `websocketroutine_manager` disregards `fills.Data`, we definitely want to know this information

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,260 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+	var resp WebsocketAdvancedAlgoOrder
+	err := json.Unmarshal(respRaw, &resp)
+	if err != nil {
+		return err
+	}
+	o.Websocket.DataHandler <- resp
+	return nil
+}
+
+func (o *Okcoin) wsProcessAlgoOrder(respRaw []byte) error {
```
gloriousCode, line 268: Same here, please process response into an order.Detail to send to the websocketroutine_manager

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,150 +238,260 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
-		return order.PendingCancel, nil
-	default:
-		return order.UnknownStatus, fmt.Errorf("%v not recognised as order status", num)
+func (o *Okcoin) wsProcessAdvancedAlgoOrder(respRaw []byte) error {
+	var resp WebsocketAdvancedAlgoOrder
+	err := json.Unmarshal(respRaw, &resp)
+	if err != nil {
+		return err
+	}
+	o.Websocket.DataHandler <- resp
```
gloriousCode, line 264: And here too, order.Detail

gloriousCode (changes requested):
Here is my review of the websocket implementation

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -24,8 +25,53 @@ import (
...
+	wsAlgoAdvance = "algo-advance"
+)
+
+var defaultSubscriptions = []string{
+	wsTickers,
+	wsCandle1D,
+	wsTrades,
+	wsOrderbooks,
+	wsStatus,
+}
```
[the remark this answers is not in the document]
samuael: It would be great to have such kind of implementation.

exchanges/okcoin/okcoin_websocket.go
samuael: this channel wsOrderbookTickByTickL400= "books-l2-tbt" requires authentication. It is not public.

exchanges/okcoin/okcoin_websocket.go
samuael: Change not implemented because of the reason above.

exchanges/okcoin/okcoin_websocket.go
samuael: Alright, sir. Changes applied.

exchanges/okcoin/okcoin_websocket.go
gloriousCode: > Only API users who are VIP4 and above in trading fee tier are allowed to subscribe to "books-l2-tbt" 400 depth channels 

Ahh yes, I do not have enough money for it 😆 😢 sorry for false-request

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,584 +247,650 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
 		// Temp type to evaluate max byte len after a marshal on batched unsubs
-		temp := WebsocketEventRequest{
-			Operation: operation,
+		copy(temp.Arguments, subscriptionRequest.Arguments)
+		copy(authTemp.Arguments, authRequest.Arguments)
+		argument := map[string]string{
+			"channel": subs[i].Channel,
+		}
+		if subs[i].Params != nil {
+			if currency, okay := subs[i].Params["ccy"]; okay {
```
gloriousCode, line 799: rename `currency` to something else to avoid package collision

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -300,85 +300,89 @@ func (o *OKCoin) Run(ctx context.Context) {
...
+	if !o.SupportsAsset(a) {
+		return nil, fmt.Errorf("%w, asset: %v", asset.ErrNotSupported, a)
+	}
+	prods, err := o.GetInstruments(ctx, "SPOT", "")
+	if err != nil {
+		return nil, err
+	}
+	pairs := make([]currency.Pair, len(prods))
+	for x := range prods {
+		var pair currency.Pair
```
gloriousCode, line 313:
```go
If prods[x].State != "live" {
    continue
}
```

exchanges/okcoin/okcoin_test.go
```diff
@@ -54,1306 +80,1943 @@ func TestStart(t *testing.T) {
...
 	t.Parallel()
-	expectedResult := exchange.AutoWithdrawCryptoText + " & " + exchange.NoFiatWithdrawalsText
-	withdrawPermissions := o.FormatWithdrawPermissions()
-	if withdrawPermissions != expectedResult {
-		t.Errorf("Expected: %s, Received: %s", expectedResult, withdrawPermissions)
+	sharedtestvalues.SkipTestIfCredentialsUnset(t, o)
+	var getOrdersRequest = order.MultiOrderRequest{
+		Type:      order.AnyType,
+		AssetType: asset.Spot,
+		Side:      order.AnySide,
```
gloriousCode, line 1393:
```suggestion
		Side:      order.AnySide,
		Pairs:     currency.Pairs{spotTradablePair},
```

exchanges/okcoin/okcoin_websocket.go
gloriousCode: This one still looks like it needs to become an order.Detail

gloriousCode (changes requested):
Nice work Sam! I think this is all that's left from me. 

I've now also verified your wrapper funcs, checked your updated ws updates and checked for any running errors while running GCT.

Unfortunately, OKcoin is not available to me to transfer funds to, but I have verified your implementations to the API for orders

exchanges/okcoin/okcoin_convert.go
```diff
@@ -82,5 +82,8 @@ func (a *okcoinNumber) UnmarshalJSON(data []byte) error {
 // Float64 returns a float64 value from okcoinNumber instance.
 func (a *okcoinNumber) Float64() float64 { return float64(*a) }
 
+// Int64 returns a int64 value from okcoinNumber instance.
+func (a *okcoinNumber) Int64() int64 { return int64(*a) }
```
shazbert, line 86: These methods don't need pointer receivers

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -295,27 +264,104 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+			ClientOrderID:   resp.Data[d].ClOrdID,
+			Type:            oType,
+			Side:            oSide,
+			Status:          oStatus,
+			AssetType:       asset.Spot,
+			Date:            resp.Data[d].CreationTime.Time(),
+			LastUpdated:     resp.Data[d].PushTime.Time().UTC(),
+			Pair:            cp,
+		}
+		o.Websocket.DataHandler <- algoOrders
```
shazbert, line 312: This is pushing incomplete data, can you please shift this out of the for loop scope.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -615,7 +643,7 @@ func (o *OKCoin) wsProcessTrades(respRaw []byte) error {
 }
 
 // wsProcessCandles converts candle data and sends it to the data handler
-func (o *OKCoin) wsProcessCandles(respRaw []byte) error {
+func (o *Okcoin) wsProcessCandles(respRaw []byte) error {
```
shazbert, line 646: Reasoning for changing the struct name `Okcoin` -> `Okcoin` ?

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -631,17 +659,11 @@ func (o *OKCoin) wsProcessCandles(respRaw []byte) error {
 
 // AppendWsOrderbookItems adds websocket orderbook data bid/asks into an
 // orderbook item array
-func (o *OKCoin) AppendWsOrderbookItems(entries [][]interface{}) ([]orderbook.Item, error) {
+func (o *Okcoin) AppendWsOrderbookItems(entries [][]okcoinNumber) ([]orderbook.Item, error) {
```
shazbert, line 662: entries [][]okcoinNumber -> entries [][2]okcoinNumber reduces panic chance

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -783,14 +830,14 @@ func (o *OKCoin) handleSubscriptions(operation string, subs []stream.ChannelSubs
...
+		copy(temp.Arguments, subscriptionRequest.Arguments)
 		copy(authTemp.Arguments, authRequest.Arguments)
 		argument := map[string]string{
 			"channel": subs[i].Channel,
 		}
 		if subs[i].Params != nil {
-			if currency, okay := subs[i].Params["ccy"]; okay {
-				argument["ccy"], okay = (currency).(string)
+			if ccy, okay := subs[i].Params["ccy"]; okay {
+				argument["ccy"], okay = (ccy).(string)
```
shazbert, line 840: `(ccy).(string)` -> `ccy.(string)`

exchanges/okcoin/okcoin_wrapper.go
```diff
@@ -310,47 +310,47 @@ func (o *OKCoin) FetchTradablePairs(ctx context.Context, a asset.Item) (currency
 	}
 	pairs := make([]currency.Pair, len(prods))
 	for x := range prods {
+		if prods[x].State != "live" {
+			continue
```
shazbert, line 314: Continuing here will leave holes in the above allocation, so please change `pairs := make([]currency.Pair, len(prods))` to `pairs := make([]currency.Pair, 0,  len(prods))` then append.

shazbert (changes requested):
Looks good just a few things. Then should be good to go.

exchanges/okcoin/okcoin_websocket.go
samuael: 😲. Alright, sir.

exchanges/okcoin/okcoin_websocket.go
samuael: 1. It was your suggestion.
2. On their website, they use 'Okcoin' not 'OKCoin'

samuael:
I hope this will be the last update round and proceed to Kucoin and the others.

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -108,27 +169,66 @@ func (o *OKCoin) WsReadData() {
...
+			wsOrderbookTickByTickL50:
+			return o.wsProcessOrderbook(respRaw, dataResponse.Arguments.Channel)
+		case wsStatus:
+			var resp WebsocketStatus
+			err = json.Unmarshal(respRaw, &resp)
+			if err != nil {
+				return err
+			}
+			for x := range resp.Data {
+				systemStatus := fmt.Sprintf("%s %s on system %s %s service type\nFrom %s\nTo %s\n", systemStateString(resp.Data[x].State), resp.Data[x].Title, resp.Data[x].System, systemStatusServiceTypeString(resp.Data[x].ServiceType), resp.Data[x].Begin.Time().String(), resp.Data[x].End.Time().String())
```
shazbert, line 213: Can we please remove the new lines as this seems like an oversight:

image

shazbert (changes requested):
Thanks for the changes. Found last things.

shazbert (approved):
Thanks @samuael! tACK.

thrasher- (changes requested):
Brilliant stuff @samuael ! Just some minor nits

thrasher- (changes requested):
Thanks for making those changes! Now that the exchange wrapper standards has been re-instated, it has picked up two issues:

https://github.com/thrasher-corp/gocryptotrader/actions/runs/5822870239/job/15788479118#step:7:242

exchanges/okcoin/okcoin_test.go
```diff
@@ -1357,7 +1357,7 @@ func TestGetHistoricCandles(t *testing.T) {
 
 func TestGetHistoricCandlesExtended(t *testing.T) {
 	t.Parallel()
-	_, err := o.GetHistoricCandlesExtended(context.Background(), spotTradablePair, asset.Spot, kline.OneMin, time.Now().Add(-time.Hour*3), time.Now())
+	_, err := o.GetHistoricCandlesExtended(context.Background(), spotTradablePair, asset.Spot, kline.OneDay, time.Now().Add(-time.Hour*24*10), time.Now())
```
shazbert, line 1360:
```diff
diff --git a/exchanges/okcoin/okcoin_wrapper.go b/exchanges/okcoin/okcoin_wrapper.go
index aa4c0c166..922e3739f 100644
--- a/exchanges/okcoin/okcoin_wrapper.go
+++ b/exchanges/okcoin/okcoin_wrapper.go
@@ -1191,7 +1191,12 @@ func (o *Okcoin) GetHistoricCandlesExtended(ctx context.Context, pair currency.P
        timeSeries := make([]kline.Candle, 0, req.Size())
        for x := range req.RangeHolder.Ranges {
                var candles []CandlestickData
-               candles, err = o.GetCandlestickHistory(ctx, req.RequestFormatted.String(), req.RangeHolder.Ranges[x].Start.Time, req.RangeHolder.Ranges[x].End.Time, interval, 0)
+               candles, err = o.GetCandlestickHistory(ctx,
+                       req.RequestFormatted.String(),
+                       req.RangeHolder.Ranges[x].Start.Time.Add(-time.Nanosecond), // Time is not inclusive.
+                       req.RangeHolder.Ranges[x].End.Time,
+                       interval,
+                       0)
                if err != nil {
                        return nil, err
                }
```

Missing first candle.

shazbert (changes requested):
LGTM One final thing from me on new changes.

samuael:
I added this last fix on the method **_GetHistoricCandlesExtended_**  by trying other querying methods not written in the documentation. I found that after trying all possible combinations of correct valid but non-data-loading queries. So the change in "before", and "after" to "start" and "end" is based on that. The mistake was in their documentation.😉.

shazbert (approved):
Changes ACK. Nice work 👍

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,584 +247,690 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
-			RemainingAmount:   resp.Data[i].Size - resp.Data[i].LastFillQty,
-			Exchange:          o.Name,
-			OrderID:           resp.Data[i].OrderID,
-			Type:              oType,
-			Side:              oSide,
-			Status:            oStatus,
-			AssetType:         o.GetAssetTypeFromTableName(resp.Table),
-			Date:              resp.Data[i].CreatedAt,
-			Pair:              pair,
+			return err
```
gloriousCode, line 292: Instead of using `return err` here, could you use something like:
```go
	b.Websocket.DataHandler <- order.ClassificationError{
					Exchange: b.Name,
					OrderID:  notification.Data[i].OrderID,
					Err:      err,
				}
```
This is mostly because rather than returning an error for order 1 of 10 (for example) its ultimately best to know about orders, even if they don't have the correct order type (for example) . At least then we can query the order via orderID and get the right info

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,584 +247,690 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+		oType, err = order.StringToOrderType(resp.Data[a].OrderType)
+		if err != nil {
+			return err
 		}
-	}
+		var status order.Status
+		status, err = order.StringToOrderStatus(resp.Data[a].State)
+		if err != nil {
+			return err
+		}
```
gloriousCode, line 342: Instead of using `return err` here, could you use something like:
```go
	b.Websocket.DataHandler <- order.ClassificationError{
					Exchange: b.Name,
					OrderID:  notification.Data[i].OrderID,
					Err:      err,
				}
```
This is mostly because rather than returning an error for order 1 of 10 (for example) its ultimately best to know about orders, even if they don't have the correct order type (for example) . At least then we can query the order via orderID and get the right info

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,584 +247,690 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+		oType, err = order.StringToOrderType(resp.Data[x].OrderType)
 		if err != nil {
 			return err
 		}
-		klineData.Volume, err = strconv.ParseFloat(response.Data[i].Candle[5], 64)
+		var status order.Status
+		status, err = order.StringToOrderStatus(resp.Data[x].State)
 		if err != nil {
 			return err
 		}
```
gloriousCode, line 391: Instead of using `return err` here, could you use something like:
```go
	b.Websocket.DataHandler <- order.ClassificationError{
					Exchange: b.Name,
					OrderID:  notification.Data[i].OrderID,
					Err:      err,
				}
```
This is mostly because rather than returning an error for order 1 of 10 (for example) its ultimately best to know about orders, even if they don't have the correct order type (for example) . At least then we can query the order via orderID and get the right info

exchanges/okcoin/okcoin_test.go
```diff
@@ -1357,7 +1357,7 @@ func TestGetHistoricCandles(t *testing.T) {
 
 func TestGetHistoricCandlesExtended(t *testing.T) {
 	t.Parallel()
-	_, err := o.GetHistoricCandlesExtended(context.Background(), spotTradablePair, asset.Spot, kline.OneMin, time.Now().Add(-time.Hour*3), time.Now())
+	_, err := o.GetHistoricCandlesExtended(context.Background(), spotTradablePair, asset.Spot, kline.ThreeMonth, time.Now().Add(-time.Hour*24*3700), time.Now().Add(-time.Hour*24*3000))
```
gloriousCode, line 1360: I think I'd prefer something less hardcoded with:
```go
	start := time.Now().Add(-kline.ThreeMonth.Duration() * 3).Truncate(kline.ThreeMonth.Duration())
	end := start.Add(kline.ThreeMonth.Duration())
	_, err := o.GetHistoricCandlesExtended(context.Background(), spotTradablePair, asset.Spot, kline.ThreeMonth, start, end)
	if err != nil {
		t.Errorf("%s GetHistoricCandlesExtended() error: %v", o.Name, err)
	}
```

gloriousCode (changes requested):
Nice work Sam, just a couple of extra ones from me

gloriousCode (approved):
tACK! Nice work Sam!

I approve this with the caveat that I do not have a verified account on OKCoin as they require me to be from the US. I've verified the endpoints match documentation and have tested all non-authenticated endpoints

shazbert (approved):
Ch ACK! Thanks @samuael

samuael:
Thank you all for your contribution.

exchanges/okcoin/okcoin.go
```diff
@@ -12,613 +12,1634 @@ import (
...
-	okCoinStandardRequestRate = 6
-	apiPath                   = "api/"
-	okCoinAPIURL              = "https://www.okcoin.com/" + apiPath
-	okCoinAPIVersion          = "/v3/"
-	okCoinExchangeName        = "OKCOIN International"
-	okCoinWebsocketURL        = "wss://real.okcoin.com:8443/ws/v3"
+	apiPath                   = "/api/"
+	okcoinAPIURL              = "https://www.okcoin.com"
+	okcoinAPIVersion          = apiPath + "v5/"
+	okcoinExchangeName        = "okcoin"
```
thrasher-, line 29:
```suggestion
	okcoinExchangeName        = "Okcoin"
```

thrasher- (changes requested):
For config compatibility, please add a "OKCOIN International" -> "Okcoin" check here:

https://github.com/thrasher-corp/gocryptotrader/blob/master/config/config.go#L803

And enhance the test coverage: https://github.com/thrasher-corp/gocryptotrader/blob/master/config/config_test.go#L1337

exchanges/okcoin/okcoin.go
samuael: Change applied, sir!

gloriousCode (approved):
tACK latest changes

shazbert (approved):
tACK changes

thrasher- (approved):
Great work @samuael ! 💯

exchanges/okcoin/okcoin_websocket.go
```diff
@@ -147,584 +247,726 @@ func (o *OKCoin) WsHandleData(respRaw []byte) error {
...
+	}
+	channel := resp.Arg.Channel
+	length := len(resp.Data[0].Asks) + len(resp.Data[0].Bids)
+	var snapshot bool
+	if channel == wsOrderbooks && length >= 100 ||
+		channel == wsOrderbooksL5 ||
+		channel == wsOrderbookTickByTickL50 ||
+		channel == wsOrderbookTickByTickL400 && length >= 400 ||
+		channel == wsOrderbookTickByTickL50 && length >= 50 {
+		snapshot = true
```
Beadko, line 492: Hi @samuael, shouldn't we be using resp.Action == snapshot? 

I don't think we should ignore being told that it's a snapshot, and if the orderBook is < 50 deep, we wouldn't process it correctly.

Was there a problem with using action?

samuael: Hi @Beadko the reason why I used such conditions to classify the data as snapshot or update is the following description.
Screenshot 2023-12-28 at 2 02 45 PM

samuael: I am reviewing it. I'll update you when I finish with it.

Beadko: no worries, I am working in this area at the moment anyway, so don't mind taking a look at it. Just wanted to double check first :)

samuael: Created a new PR [here](https://github.com/thrasher-corp/gocryptotrader/pull/1436)

PR #1194, opened 2023-05-12
CommunicationsConfig: Fix Telegram can't receive messages

shanhuhai5739:
# PR Description
After Telegram configures the robot, it still cannot receive the transaction order message. After reading the code, it is found that `AuthorizedClients []int64` is not referenced and assigned and cannot push the PushEvent message.

Fixes # (issue)
When the main process starts for the first time, obtain the Telegram robot information, initialize and allocate the `AuthorizedClients` object, and successfully trigger the `PushEvent` message when the transaction order event occurs

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions/AppVeyor with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1208, opened 2023-05-30
telegram: Fix pushing events to an authorised client list

thrasher-:
# PR Description

Takes the work done in https://github.com/thrasher-corp/gocryptotrader/pull/1194 by @shanhuhai5739 and adds an authorised client list so only authenticated users can receive pushed events.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Telegram

## Checklist

### DIFF
--- a/cmd/documentation/communications_templates/telegram.tmpl
+++ b/cmd/documentation/communications_templates/telegram.tmpl
@@ -33,7 +33,7 @@ developed by Telegram Messenger LLP
 			Enabled:           true,
 			Verbose:           false,
 			VerificationToken: "token",
-			AuthorisedClients: "pepe,bob",
+			AuthorisedClients: map[string]int64{"pepe": 0}, // 0 represents a placeholder for the user's ID
 		},
 	}
 
@@ -49,7 +49,6 @@ via Telegram:
 /start			- Will authenticate your ID
 /status			- Displays the status of the bot
 /help			- Displays current command list
-/settings		- Displays current bot settings
 ```
 
 ### Please click GoDocs chevron above to view current GoDoc information for this package
--- a/cmd/documentation/communications_templates/telegram.tmpl
+++ b/cmd/documentation/communications_templates/telegram.tmpl
@@ -17,7 +17,10 @@ developed by Telegram Messenger LLP
 
 	+ [Enable via configuration](https://github.com/thrasher-corp/gocryptotrader/tree/master/config#enable-communications-via-config-example)
 
-	+ Individual package example below:
+	+ See the individual package example below. NOTE: For privacy considerations, it's not possible to directly request a user's ID through the 
+	Telegram Bot API unless the user interacts first. The user must message the bot directly. This allows the bot to identify and save the user's ID. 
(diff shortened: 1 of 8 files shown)


communications/telegram/telegram.go
```diff
@@ -141,14 +148,18 @@ func (t *Telegram) InitialConnect() error {
 		return errors.New(resp.Description)
 	}
 
-	warmWelcomeList := make(map[string]int64)
 	for i := range resp.Result {
-		if resp.Result[i].Message.From.ID != 0 {
-			warmWelcomeList[resp.Result[i].Message.From.UserName] = resp.Result[i].Message.From.ID
+		if resp.Result[i].Message.From.UserName != "" && resp.Result[i].Message.From.ID != 0 {
+			_, ok := t.AuthorisedClients[resp.Result[i].Message.From.UserName]
+			if !ok {
```
gloriousCode, line 154: I think this could be on `verbose` unless its a bit more important than I realise to highlight? eg me spamming 64 messages in a chatroom while unauthorised prints this error 64 times

gloriousCode (changes requested):
It works! Just one verbose thing

communications/telegram/telegram.go
thrasher-: Security related messages should be relayed by default without verbosity enabled, however if a user spams your bot multiple times, then that shouldn't spam the logs. I've opted to instead report all unique unauth user warnings and then it's up to the user to ban repeat offenders

gloriousCode (approved):
tACK! I like the solution to my spam crimes

shazbert (approved):
utACK. telegram is not working.

cmd/documentation/communications_templates/telegram.tmpl
```diff
@@ -20,19 +20,22 @@ developed by Telegram Messenger LLP
 	+ Individual package example below:
```
gloriousCode, line 20: Perhaps highlight that the zero placeholder will be replaced by the bot once the authclient is verified. Just because of the difficulty of actually getting the ID otherwise.

gloriousCode (changes requested):
The updates work! I think just a small text change to help a user understand whats required when setting up their config

gloriousCode (approved):
image

communications/telegram/telegram.go
```diff
@@ -25,17 +27,15 @@ const (
 	methodGetUpdates  = "getUpdates"
 	methodSendMessage = "sendMessage"
 
-	cmdStart    = "/start"
-	cmdStatus   = "/status"
-	cmdHelp     = "/help"
-	cmdSettings = "/settings"
+	cmdStart  = "/start"
```
shazbert, line 30: Suggestion: Would change time from local to UTC instead as server could be in diff zones.

shazbert (approved):
Just one suggestion. Approved.

shazbert (approved):
Thank you sir!

gloriousCode (approved):
I appreciate the config upgrade, tack

PR #1255, opened 2023-07-06
Kraken: Load order execution limits for tradable pairs

Beadko:
## Type of change

## How has this been tested

- [x] go test ./... -race
- [ ] golangci-lint run
- [x] Test Kraken Limit Execution

## Checklist

### DIFF
--- a/exchanges/kraken/kraken_test.go
+++ b/exchanges/kraken/kraken_test.go
@@ -116,7 +116,6 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 	}
 
 	for assetItem, limitTests := range tests {
-
 		if err := k.UpdateOrderExecutionLimits(context.Background(), assetItem); err != nil {
 			t.Errorf("Error fetching %s pairs for test: %v", assetItem, err)
 		}
--- a/exchanges/kraken/kraken_test.go
+++ b/exchanges/kraken/kraken_test.go
@@ -99,7 +99,7 @@ func TestWrapperGetServerTime(t *testing.T) {
 	}
 }
 
-func TestExecutionLimits(t *testing.T) {
+func TestUpdateOrderExecutionLimits(t *testing.T) {
 	t.Parallel()
 
 	type limitTest struct {
@@ -116,9 +116,8 @@ func TestExecutionLimits(t *testing.T) {
 	}
 
 	for assetItem, limitTests := range tests {
-		// FetchTradablePairs updates Pair Limits in the exchange
-		_, err := k.FetchTradablePairs(context.Background(), assetItem)
-		if err != nil {
+
+		if err := k.UpdateOrderExecutionLimits(context.Background(), assetItem); err != nil {
 			t.Errorf("Error fetching %s pairs for test: %v", assetItem, err)
 		}
 
--- a/exchanges/kraken/kraken_test.go
+++ b/exchanges/kraken/kraken_test.go
@@ -99,6 +99,46 @@ func TestWrapperGetServerTime(t *testing.T) {
 	}
 }
 
+func TestExecutionLimits(t *testing.T) {
+	t.Parallel()
+
+	type limitTest struct {
+		pair currency.Pair
+		step float64
+		min  float64
+	}
+
+	tests := map[asset.Item][]limitTest{
+		asset.Spot: {
+			{currency.NewPair(currency.ETH, currency.USDT), 0.01, 0.01},
+			{currency.NewPair(currency.XBT, currency.USDT), 0.1, 0.0001},
+		},
+	}
+
+	for assetItem, limitTests := range tests {
+		// FetchTradablePairs updates Pair Limits in the exchange
+		_, err := k.FetchTradablePairs(context.Background(), assetItem)
+		if err != nil {
+			t.Errorf("Error fetching %s pairs for test: %v", assetItem, err)
+		}
+
+		for _, limitTest := range limitTests {
+			limits, err := k.GetOrderExecutionLimits(assetItem, limitTest.pair)
+			if err != nil {
+				t.Errorf("Kraken GetOrderExecutionLimits() error during TestExecutionLimits; Asset: %s Pair: %s Err: %v", assetItem, limitTest.pair, err)
+				continue
+			}
+			if got := limits.PriceStepIncrementSize; got != limitTest.step {
+				t.Errorf("Kraken UpdateOrderExecutionLimits wrong PriceStepIncrementSize; Asset: %s Pair: %s Expected: %v Got: %v", assetItem, limitTest.pair, limitTest.step, got)
+			}
+
+			if got := limits.MinimumBaseAmount; got != limitTest.min {
+				t.Errorf("Kraken UpdateOrderExecutionLimits wrong MinAmount; Pair: %s Expected: %v Got: %v", limitTest.pair, limitTest.min, got)
+			}
+		}
+	}
+}
+
 func TestFetchTradablePairs(t *testing.T) {
 	t.Parallel()
 	_, err := k.FetchTradablePairs(context.Background(), asset.Futures)
--- a/exchanges/kraken/kraken_wrapper.go
+++ b/exchanges/kraken/kraken_wrapper.go
@@ -346,6 +346,9 @@ func (k *Kraken) Run(ctx context.Context) {
 
 // UpdateOrderExecutionLimits sets exchange execution order limits for an asset type
 func (k *Kraken) UpdateOrderExecutionLimits(ctx context.Context, a asset.Item) error {
+	if a != asset.Spot {
+		return common.ErrNotYetImplemented
+	}
 	if _, err := k.FetchTradablePairs(ctx, a); err != nil {
 		return fmt.Errorf("%s failed to load %s pair execution limits. Err: %s", k.Name, a, err)
 	}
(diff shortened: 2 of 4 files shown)


exchanges/kraken/kraken_wrapper.go
```diff
@@ -382,8 +389,18 @@ func (k *Kraken) FetchTradablePairs(ctx context.Context, a asset.Item) (currency
...
+				Asset:                  a,
+				Pair:                   pair,
+				PriceStepIncrementSize: info.TickSize,
+				MinimumBaseAmount:      info.OrderMinimum,
+			})
 			pairs = append(pairs, pair)
 		}
+
+		if err := k.LoadLimits(limits); err != nil {
+			log.Errorf(log.ExchangeSys, "%s Error loading exchange limits: %v", k.Name, err)
```
gloriousCode, line 402: Hello! Thank you for adding this.
We have an interface function `UpdateOrderExecutionLimits` which would be a better place for this!
Checkout Binance for their implementation as a guide

`UpdateOrderExecutionLimits` gets called when an exchange is setup, so you wouldn't need to also make the modifications you've done in the `Run` function

gloriousCode (changes requested):
Thanks for taking the time to add this functionality!  I'd like it if you could update this so that it gets called inside our standard interface function of `UpdateOrderExecutionLimits`

exchanges/kraken/kraken_wrapper.go
gbjk: @gloriousCode 
I agree with implementing the Interface function. We missed that.

However: Are you sure that `UpdateOrderExecutionLimits` gets called when an exchange is setup?
I think that's only true if the exchange wrapper itself does so, like binance wrapper and okx wrapper,
and they actually do it in Run.

The quandary here with kraken's api was not wanting to call fetch symbols twice, because of the delay, performance impact, etc.
I really can't see a nice way around it.
What I think we'll do is keep the implementation that FetchTradingPairs does the load, but then move the else statement to call UpdateOrderExecutionLimits and just have that call Fetch Pairs.
Best compromise, implements the interface but doesn't introduce extra load if autoUpdate is on.

gloriousCode: Yeah I misread the code with the calling of `UpdateOrderExecutionLimits`, it does get called in `Run`. What you describe as fixes should be good 🚀

Beadko: that's done

gloriousCode: Thanks for that. There's a linter issue still for a blank line in kraken_test.go.

A test in our exchange_wrapper_standards testing suite is failing because OrderLimits does not expect a partial implementation. I'll create a PR to add `CheckOrderExecutionLimits` to the ignore list

Beadko: the whitespace is now fixed

exchanges/kraken/kraken_wrapper.go
```diff
@@ -358,6 +370,8 @@ func (k *Kraken) FetchTradablePairs(ctx context.Context, a asset.Item) (currency
 		}
 
 		pairs = make([]currency.Pair, 0, len(symbols))
+		limits := make([]order.MinMaxLevel, 0, len(symbols))
```
shazbert, line 373: We are doing 2 calls anyway when we update, might as well have this dedicated to the `UpdateOrderExecutionLimits` function because it is common across the other exchange. I know there is going duplication of code but it might be better just from a code maintenance perspective.

shazbert (changes requested):
Thank you for opening this PR and updating the exchange limits for Kraken, just two critiques. 🚀

exchanges/kraken/kraken_test.go
```diff
@@ -99,6 +99,44 @@ func TestWrapperGetServerTime(t *testing.T) {
...
+	t.Parallel()
+
+	type limitTest struct {
+		pair currency.Pair
+		step float64
+		min  float64
+	}
+
+	tests := map[asset.Item][]limitTest{
+		asset.Spot: {
```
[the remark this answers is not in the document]
gbjk: Yep, that seems okay.
I didn't want to do a partial implementation but it's better than nothing.

exchanges/kraken/kraken_wrapper.go
gbjk: Okay, so you want to not have UpdateOrderExecutionLimits call FetchTradalePairs?
But still have FetchTradable pairs set limits as well?
That seems redundant to me. There's no negative effect of calling FetchTradablePairs other than having to comment to explain.

On the other hand if you're suggesting duplicating a call to get the symbols then I don't agree at all.
We shouldn't make 2 duplicate API calls literally back to back just to avoid this, right?

exchanges/kraken/kraken_test.go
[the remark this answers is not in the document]
gbjk: @Beadko Suggestion here is that we have UpdateOrderExecutionLimits return common.ErrNotYetImplemented if the asset isn't spot.
That's a good solution.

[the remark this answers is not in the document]
Beadko: Added common.ErrNotYetImplemented

exchanges/kraken/kraken_wrapper.go
```diff
@@ -328,60 +328,105 @@ func (k *Kraken) Run(ctx context.Context) {
...
 	if err != nil {
-		log.Errorf(log.ExchangeSys,
-			"%s failed to update tradable pairs. Err: %s",
-			k.Name,
-			err)
+		return pairs, err
 	}
+
+	for _, info := range pairInfo {
+		if info.Status != "online" {
```
shazbert, line 385: This is not a change request. Just noticed this and not created by you. All these removed assets have their status as `reduce_only`. 

```go
[DEBUG] | EXCHANGE | 14/07/2023 10:07:59 | Kraken Updating available pairs [SPOT] - Removed: [TRX_ETH KEEP_USD TRIBE_EUR XTZ_ETH ETC_XBT TRIBE_USD EOS_ETH KEEP_EUR REP_XBT REP_EUR XBT_AED ETH_AED ETC_ETH].
```

shazbert (approved):
Thank you Bea! This PR looks great and works as intended, tACK!

exchanges/kraken/kraken_wrapper.go
```diff
@@ -328,60 +328,105 @@ func (k *Kraken) Run(ctx context.Context) {
...
+		if err := k.UpdateOrderExecutionLimits(ctx, a); err != nil && err != common.ErrNotYetImplemented {
+			log.Errorln(log.ExchangeSys, err.Error())
+		}
+	}
+}
+
+// UpdateOrderExecutionLimits sets exchange execution order limits for an asset type
+func (k *Kraken) UpdateOrderExecutionLimits(ctx context.Context, a asset.Item) error {
+	if a != asset.Spot {
+		return common.ErrNotYetImplemented
```
samuael, line 350: `common.ErrNotYetImplemented` 🤔Why?

samuael: Cool. I was about to suggest asset.ErrNotSupported. But, this one makes sense. Thanks for reply @beadko

exchanges/kraken/kraken_test.go
```diff
@@ -99,6 +99,44 @@ func TestWrapperGetServerTime(t *testing.T) {
...
+			{currency.NewPair(currency.XBT, currency.USDT), 0.1, 0.0001},
+		},
+	}
+
+	for assetItem, limitTests := range tests {
+		if err := k.UpdateOrderExecutionLimits(context.Background(), assetItem); err != nil {
+			t.Errorf("Error fetching %s pairs for test: %v", assetItem, err)
+		}
+
+		for _, limitTest := range limitTests {
```
gloriousCode, line 123: I think it would have been better for code readability's sake to name the `limitTest` var in this loop to something different to the inline struct you defined earlier to avoid collisions

exchanges/kraken/kraken_wrapper.go
```diff
@@ -328,60 +328,105 @@ func (k *Kraken) Run(ctx context.Context) {
...
+			Asset:                  a,
+			Pair:                   pair,
+			PriceStepIncrementSize: info.TickSize,
+			MinimumBaseAmount:      info.OrderMinimum,
+		})
 	}
 
-	err := k.UpdateTradablePairs(ctx, forceUpdate)
+	if err := k.LoadLimits(limits); err != nil {
+		return fmt.Errorf("%s Error loading %s exchange limits: %v", k.Name, a, err)
```
gloriousCode, line 370: If you use `%w` instead of `%v` of `%s` for errors when using `fmt.Errorf`, you get to wrap the original error type instead of declaring a new one. While Its not important in this instance, it can allow for easier debugging to identify the type of problem, especially if the error text is lacking

For instance, if `limits` contained an unsupported asset, you can still check if the returned error was asset.ErrNotSupported, but when using `%s` you cannot without string inspection

gloriousCode (approved):
tACK! Thank you for your contribution Beadko and co!

shazbert (changes requested):
The wrapping of errors is standard across GCT now. Here's [glorious codes critique](https://github.com/thrasher-corp/gocryptotrader/pull/1255/files#r1266248379). Once this is done then I will approve.

exchanges/kraken/kraken_wrapper.go
Beadko: that is completed

shazbert (approved):
Changes ACK. 🚀

thrasher- (approved):
tACK, thanks @Beadko !

PR #1286, opened 2023-07-29
Order: Fix AmountStepIncrementSize conforms check

gbjk:
Was using max instead of min, and using either should be redundant anyway

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


gloriousCode (approved):
Nice pickup

shazbert (approved):
tACK. 🦅 Nice pickup!

thrasher- (approved):
Great catch 👀

PR #1292, opened 2023-08-01
Bitfinex:  Improve websocket type assertion checks for order processing

### LINKED ISSUE #1288: Bitfinex: WS cancel stuck waiting for confirmation
Given a cancellation request such as:
```
[[0,"oc",null,{"id":123642403113}]]
```
And an ack:
```
[0,"n",[1690698879385,"oc-req",null,null,[123642403113,null,136725840,"tBTCUST",1690698868393,169
0698868394,0.0008,0.0008,"EXCHANGE LIMIT",null,null,null,0,"ACTIVE",null,null,29145,0,0,0,null,null,null,0,0,null,null,null,"API>BFX",null,null,{}],null,"SUCCESS","Submitted for cancellation; waiting for confirmation (ID: 123642403113)."]]
```
and confirmation
```
[0,"oc",[123642403113,null,136725840,"tBTCUST",1690698868393,1690698879386,0.0008,0.0008,"EXCHANG
E LIMIT",null,null,null,0,"CANCELED",null,null,29145,0,0,0,null,null,null,0,0,null,null,null,"API>BFX",null,null,{}]]
```

We eventually get a response to cancelOrder:
`Bitfinex websocket connection: timeout waiting for response with signature: 123642403113"`
Where the signature is the order number.

It seems to not be accepting the cancellation ack/conf.

gbjk:
Fixes #1288

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/bitfinex/bitfinex_websocket.go
+++ b/exchanges/bitfinex/bitfinex_websocket.go
@@ -916,7 +916,9 @@ func (b *Bitfinex) handleWSNotification(d []interface{}, respRaw []byte) error {
 			}
 		case strings.Contains(channelName, wsOrderNewRequest):
 			if data[2] != nil {
-				if cid, ok := data[2].(float64); ok && cid > 0 {
+				if cid, ok := data[2].(float64); !ok {
+					return common.GetTypeAssertError("float64", data[2])
+				} else if cid > 0 {
 					if b.Websocket.Match.IncomingWithData(int64(cid), respRaw) {
 						return nil
 					}
@@ -926,7 +928,9 @@ func (b *Bitfinex) handleWSNotification(d []interface{}, respRaw []byte) error {
 		case strings.Contains(channelName, wsOrderUpdateRequest),
 			strings.Contains(channelName, wsOrderCancelRequest):
 			if data[0] != nil {
-				if id, ok := data[0].(float64); ok && id > 0 {
+				if id, ok := data[0].(float64); !ok {
+					return common.GetTypeAssertError("float64", data[0])
+				} else if id > 0 {
 					if b.Websocket.Match.IncomingWithData(int64(id), respRaw) {
 						return nil
 					}
--- a/exchanges/bitfinex/bitfinex_websocket.go
+++ b/exchanges/bitfinex/bitfinex_websocket.go
@@ -917,7 +917,7 @@ func (b *Bitfinex) handleWSNotification(d []interface{}, respRaw []byte) error {
 		case strings.Contains(channelName, wsOrderNewRequest):
 			if data[2] != nil {
 				if cid, ok := data[2].(float64); !ok {
-					return common.GetTypeAssertError("float64", data[2])
+					return common.GetTypeAssertError("float64", data[2], channelName+" cid")
 				} else if cid > 0 {
 					if b.Websocket.Match.IncomingWithData(int64(cid), respRaw) {
 						return nil
@@ -929,7 +929,7 @@ func (b *Bitfinex) handleWSNotification(d []interface{}, respRaw []byte) error {
 			strings.Contains(channelName, wsOrderCancelRequest):
 			if data[0] != nil {
 				if id, ok := data[0].(float64); !ok {
-					return common.GetTypeAssertError("float64", data[0])
+					return common.GetTypeAssertError("float64", data[0], channelName+" id")
 				} else if id > 0 {
 					if b.Websocket.Match.IncomingWithData(int64(id), respRaw) {
 						return nil


exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -658,11 +658,19 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
 								b.Websocket.DataHandler <- offer
 							}
 						}
-					case strings.Contains(channelName, wsOrderNewRequest),
-						strings.Contains(channelName, wsOrderUpdateRequest),
```
shazbert, line 662: Prefer the assertion to be checked a little more explicitly. 
```diff
diff --git a/exchanges/bitfinex/bitfinex_websocket.go b/exchanges/bitfinex/bitfinex_websocket.go
index 49ddad0b3..31aa092b2 100644
--- a/exchanges/bitfinex/bitfinex_websocket.go
+++ b/exchanges/bitfinex/bitfinex_websocket.go
@@ -660,7 +660,12 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
                                                }
                                        case strings.Contains(channelName, wsOrderNewRequest):
                                                if data[2] != nil {
-                                                       if cid, ok := data[2].(float64); ok && cid > 0 {
+                                                       cid, ok := data[2].(float64)
+                                                       if !ok {
+                                                               return common.GetTypeAssertError("float64", data[2])
+                                                       }
+
+                                                       if cid > 0 {
                                                                if b.Websocket.Match.IncomingWithData(int64(cid), respRaw) {
                                                                        return nil
                                                                }
@@ -670,7 +675,12 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
                                        case strings.Contains(channelName, wsOrderUpdateRequest),
                                                strings.Contains(channelName, wsOrderCancelRequest):
                                                if data[0] != nil {
-                                                       if id, ok := data[0].(float64); ok && id > 0 {
+                                                       id, ok := data[0].(float64)
+                                                       if !ok {
+                                                               return common.GetTypeAssertError("float64", data[0])
+                                                       }
+
+                                                       if id > 0 {
                                                                if b.Websocket.Match.IncomingWithData(int64(id), respRaw) {
                                                                        return nil
                                                                }
```

shazbert (changes requested):
Well done this works. Just one nit.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Sure. Can do. ~I didn't touch it since it was already there but that's no excuse. In fact, strikethrough.~
Though I'll still prefer to scope variables wherever possible as well. Saves so many lives.

gbjk: And Thanks 🙏

gbjk: Nit / Challenge: Feels like GetTypeAssertError could be common.AssertType and return the correct type OR an error.
Would require Generics, but we're requiring go 1.20 anyway so shrugapotamus 🤷 🦛

gbjk:
Strange as this might seem, this branch had 1 nit from @shazbert which didn't make it into the main refactor.
Whilst there are other assertions that could/should be caught, I don't want to lose this one.

Rebased on master already

thrasher- (approved):

shazbert (approved):
changes ACK!

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -916,7 +916,9 @@ func (b *Bitfinex) handleWSNotification(d []interface{}, respRaw []byte) error {
 			}
 		case strings.Contains(channelName, wsOrderNewRequest):
 			if data[2] != nil {
-				if cid, ok := data[2].(float64); ok && cid > 0 {
+				if cid, ok := data[2].(float64); !ok {
+					return common.GetTypeAssertError("float64", data[2])
```
thrasher-, line 920:
```suggestion
					return common.GetTypeAssertError("float64", data[2], "wsOrderNewRequest cid")
```

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -926,7 +928,9 @@ func (b *Bitfinex) handleWSNotification(d []interface{}, respRaw []byte) error {
 		case strings.Contains(channelName, wsOrderUpdateRequest),
 			strings.Contains(channelName, wsOrderCancelRequest):
 			if data[0] != nil {
-				if id, ok := data[0].(float64); ok && id > 0 {
+				if id, ok := data[0].(float64); !ok {
+					return common.GetTypeAssertError("float64", data[0])
```
thrasher-, line 932:
```suggestion
					return common.GetTypeAssertError("float64", data[0], "wsOrderUpdateRequest/wsOrderCancelRequest id")
```

thrasher- (commented):
Upon second thought, would be better to annotate what stream event triggered the type assert failure so it's more helpful for the user

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Think better to use chanName for both of these, since it's non-internal and consumers can look at BFX api docs.
Now I'm looking at it, these strings.Contains all seem superfluous; equality checks should work.
Not going to fix that now.

thrasher- (approved):
Changes ACK, thanks!

gloriousCode (approved):
tACK! 

```
// my forced error
type assert failure from string to float64 for: on-req cid
```

PR #1310, opened 2023-08-08
Bitfinex: Fix ws Trades Fees on te

gbjk:
The ws channel for authenticated Trades sends two types of update:
* te, Trade Executed
* tu, Trade Execution Update

Only the second one contains fee information.
[See the docs](https://docs.bitfinex.com/reference/ws-auth-trades)

This commit fixes:
`exchange Bitfinex websocket error - unable to type assert trade fee`
after an executed market trade on the te update

Side-note: I was tempted to type eventType but I've deliberately avoided yaking too much on this.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


gbjk:
Merged as part of #1317

PR #1309, opened 2023-08-08
stream/buffer: Reduces map lookups by using key struct

shazbert:
# PR Description

Uses a key instead of multiple map to map lookups. This yields slightly better performance, with no additional allocations. There was also an obstruction due to a potential slow go schedular read on the data handler channel. Opted to ~~drop alert for improved websocket reader update through put.~~  buffer the channel to allow for that hand over to occur.

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions/AppVeyor with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

## Benchmark - Edit by thrasher-

```
var multiMapHolder = make(map[currency.Code]map[currency.Code]map[asset.Item]*orderbookHolder)

// 62019669	        19.40 ns/op	       0 B/op	       0 allocs/op
func BenchmarkMultiMapHolder(b *testing.B) {
	var c1 currency.Code
	var c2 currency.Code
	var a asset.Item
	for x := 0; x < 3000; x++ {
		c1 = currency.NewCode(strconv.FormatInt(int64(x), 10))
		c2 = currency.NewCode(strconv.FormatInt(int64(x), 10))

		m1, ok := multiMapHolder[c1]
		if !ok {
			m1 = make(map[currency.Code]map[asset.Item]*orderbookHolder)
			multiMapHolder[c1] = m1
		}

		m2, ok := m1[c2]
		if !ok {
			m2 = make(map[asset.Item]*orderbookHolder)
			m1[c2] = m2
		}

[...]

PR #1316, opened 2023-08-10
stream: Add processor for concurrent processing of websocket data

shazbert:
# PR Description

* Adds processor (POC) type for streaming readers so that a read can be handled quickly and appropriately in a concurrent manner.
* Adds method `Process` to take in a `stream.Key` struct that allows matching to a spawned routine that directly handles that update for future. The worker has a a buffer in front of it so that if for any reason there is a slow down it does not impede the direct reads of the websocket handler. 
* changes buffer.go orderbook holder's map main mutex to RW mutex. This allows quick direct access to individual data without the overhead of a a standard mutex. This allows for the processor to access the underlying data concurrently. 

NOTE: This is only implemented for GATEIO due to its large pair count before undertaking support across all exchanges. 

To test in gateio_websocket.go
```diff
diff --git a/exchanges/gateio/gateio_websocket.go b/exchanges/gateio/gateio_websocket.go
index 64e7c1139..ba1b25a4d 100644
--- a/exchanges/gateio/gateio_websocket.go
+++ b/exchanges/gateio/gateio_websocket.go
@@ -49,9 +49,9 @@ const (
 )

 var defaultSubscriptions = []string{
-       spotTickerChannel,
-       spotCandlesticksChannel,
-       spotTradesChannel,
+       // spotTickerChannel,
+       // spotCandlesticksChannel,
+       // spotTradesChannel,
        spotOrderbookTickerChannel,
 }

@@ -94,15 +94,35 @@ func (g *Gateio) generateWsSignature(secret, event, channel string, dtime time.T
        return hex.EncodeToString(mac.Sum(nil)), nil
 }

+var startTime = time.Now()
+var wow time.Duration
+var count int
+var ops int
+
 // wsReadConnData receives and passes on websocket messages for processing
 func (g *Gateio) wsReadConnData() {
+       go func() {
+               for {
+                       time.Sleep(time.Second)
+                       fmt.Printf("Current operating throughput: %v operations per second.\n", ops)
+                       ops = 0
+               }
+       }()
        defer g.Websocket.Wg.Done()
        for {
                resp := g.Websocket.Conn.ReadMessage()
                if resp.Raw == nil {
                        return
                }
+               count++
+               ops++
+               start := time.Now()
                err := g.wsHandleData(resp.Raw)
+               since := time.Since(start)
+               if since > wow {
+                       wow = since
+                       fmt.Printf("WOW: %v ms COUNT: %d SINCE: %s\n", since.Milliseconds(), count, time.Since(startTime).String())
+               }
                if err != nil {
                        g.Websocket.DataHandler <- err
                }
```

```cmd
.\gocryptotrader --tickersync=false --enableallpairs
```

My testing: 

* Master branch: Operations per second was comparable to this branch but there was linear friction observed over time which exceeded 100ms in some cases, so everything in the websocket buffer was now out by 100ms (which is not ideal) until the websocket data handler could effectively catch up, and this would occur quite consistently.
* This branch: Observed that this update achieved better latency performance ~5ms at the worst end over 5mins of operation. There is caveats to initial start with go routine generation and also the handover between a reader routine to the processor which will need to be benchmarked and ironed out but at this situation I am seeing some robust improvements from what it was. Also to note the buffer size of 10 I have only seen maxed out at start up hasn't yet filled up and complained over time.

Dependancy Required: #1309 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/stream/processor.go
```diff
@@ -0,0 +1,116 @@
...
+			ch = make(chan func() error, w.chanBufferSize)
+			w.routes[key] = ch
+			close(wait)
+			w.mtx.Unlock()
+			for pending := range ch {
+				err := pending()
+				if err != nil {
+					w.dataHandler <- err
+				}
+			}
```
gloriousCode, line 101: Does this need to have consideration for shutdown? Either websocket enable/disable or GCT level?

exchanges/stream/buffer/buffer_test.go
```diff
@@ -57,7 +57,10 @@ func createSnapshot() (holder *Orderbook, asks, bids orderbook.Items, err error)
 		ob:           newBook,
 	}
 	err = holder.LoadSnapshot(book)
-	return
+	if errors.Is(err, errDataHandlerReaderSlow) { // expected return error
+		err = nil
```
gloriousCode, line 61: You're not actually using error here or above

gloriousCode (changes requested):
There are a few linter issues to fix up. Otherwise, I quite like your implementation here. It creates your special lane for orderbook processing with potentially more to add. I haven't yet benchmarked performance in a meaningful way yet.

I implemented this for OKx orderbook snapshot and updates and I'm happy that checksum calculation still works as it processes things synchronously despite your PR description and my initial fears when I didn't understand what was going on.

exchanges/stream/processor.go
shazbert: Depends, blocking routines are ~2->3 kb so about 30k pairs running would be around 90 megs then future we stack ticker, order, other processing it could get bigger but at the moment its got a small foot print? I think the added complexity is not really needed just yet. The processor will be persistent if you enable/disable, if we wanted to purge that then I would say yeah we should shut down all the routines. 

As for the name: executeJob(), processTask(), handleJob(), runTask(), performAction() ?

gloriousCode: executeJob() sounds good

exchanges/stream/processor.go
```diff
@@ -0,0 +1,117 @@
...
+			ch = make(chan func() error, w.chanBufferSize)
+			w.routes[key] = ch
+			close(wait)
+			w.mtx.Unlock()
+			for processWebsocketTask := range ch {
+				err := processWebsocketTask()
+				if err != nil {
+					w.dataHandler <- err
+				}
+			}
```
gloriousCode, line 102: This comment is a bit theoretical because its making assumptions about future uses. As I said last time, I actually quite like having dedicated routes to process messages. 

Depending on how this goes moving forward  I anticipate two problems with this design:
**1)** We have a system to allow the overriding of the DataHandler to allow a user to have custom processes for how they wish to handle WS updates. This system circumvents that. For the specific use-case you've done here with orderbooks, it doesn't apply because we use `LoadSnapshot(...)` and `Update(...)` which already circumvent the datahandler. 
If you expand this to tickers or orders, we process those via the DataHandler. 
I would no longer have the ability to process these updates and means that it can only be done via GCT and requires an update to `main` GCT to change. Which leads to the next part

**2)** That it can't be toggled. Depending on this feature and how it is going to be accomplished, I think this is something that should be a feature and if not enabled, sends data to the datahandler. 
This all would require a documentation update in terms of having a readme.md for it.

This isn't something to really address in this PR though, because its not being utilised in this way. But you've clearly designed it in such a way as to allow for future expansion. If this is only going to stay to help process orderbooks asynchronously, it could potentially be renamed to be more specific to the orderbook.

Thank you for listening to my TED talk

exchanges/stream/processor.go
```diff
@@ -0,0 +1,117 @@
...
+		chanBufferSize: maxChanBuffer,
+		dataHandler:    dataHandler,
+	}, nil
+}
+
+// Process spawns a new worker for a key if one does not already exist. It
+// will then queue the function to be processed. If the channel is full, it
+// will block until a slot is available. This tries to alleviate websocket
+// reader blocking issues.
+func (w *Processor) Process(key Key, fn func() error) error {
```
gloriousCode, line 71: While it does `process`, you are sending a function to a queue to be executed, so I think you should highlight in the naming that it isn't just executing the function you send to it immediately. It would be weird to receive the error "channel full" when I ask the websocket processor to process.

I think `QueueFunction` should work. This was mainly highlighted when looking at GateIO being `Processor.Process(...)` which looks funny. A user can tell from `Processor.QueueFunction` that it should be Processed by the Processor.

gloriousCode (changes requested):
Thanks for making those changes! I've got some small requests and one comment on the future of this system

exchanges/stream/processor.go
shazbert: That is a good suggestion 🖖

exchanges/stream/processor.go
shazbert: For point 1), I understand your concern about the ability to override the DataHandler for custom processing. However, I'd like to clarify that for tickers as an example, we perform pre-processing within the WebSocket handle data function. This includes reading data from the connection, unmarshalling the byte array into our declared type, and allocating memory for a ticker.Price struct. With the changes proposed, we are essentially moving this unmarshalling and struct allocation process to a dedicated routine. The key point here is that the processed data will still be pushed directly to the DataHandler in all cases. Therefore, while it may seem like a shift, it doesn't bypass the DataHandler's processing within the WebSocket routine manager.

Regarding point 2), Toggling is a good point, If for some reason you want to read and processing linearly then I can see this as acceptable.

gloriousCode: > With the changes proposed, we are essentially moving this unmarshalling and struct allocation process to a dedicated routine

Does that mean you will be [moving the json unmarshalling as part of the processor.Process](https://github.com/thrasher-corp/gocryptotrader/pull/1316/files#diff-c737e2208dda52b4019818edc7d9c8ccc622b1e6a230baa26c4a1cadd455e833R281)? 

The other part of what you say regarding ticker is only theoretical because its the DataHandler that processes tickers at the moment, adding this to it would only be adding an additional async queue to then process the ticker in a sync channel on the data handler - but this is all theoretical because the work for tickers doesn't exist yet. If you're saying that everything will continue to be pushed to the sync data handler, then there's no real reason for a enable/disable, its just a discussion on what your intent is with this design.

shazbert: > Does that mean you will be [moving the json unmarshalling as part of the processor.Process]

I need the data for the key so I cannot under this design. 

> adding this to it would only be adding an additional async queue to then process the ticker in a sync channel on the data handler 

Yeah a cost benefit would need to be figured out in this case.

gloriousCode (changes requested):
Fix lint and I'll do one more review, but I think it should be good to go/expand from

exchanges/stream/processor.go
```diff
@@ -0,0 +1,128 @@
...
+		dataHandler:    dataHandler,
+	}, nil
+}
+
+// QueueFunction spawns a new worker for a key if one does not already exist. It
+// will then queue the function to be processed. If the channel is full, it
+// will block until a slot is available. This tries to alleviate websocket
+// reader blocking issues.
+func (w *Processor) QueueFunction(key Key, fn func() error) error {
+	if key == (Key{}) {
```
gloriousCode, line 76: Feel free to click resolve, but since Processor is a pointer property, having a nil check here could prevent a panic. But its likely something that would only come up if someone setup a test incorrectly since our default path in GCT is to go through `websocket.Setup` where this gets setup. I just can't not comment on this 😄

gloriousCode (approved):
Nice work! I'm excited for its use 😎

gloriousCode (approved):

shazbert:
Closing due current external trading branch average processing time is around ~30 microseconds from reader, processing to orderbook package map. This is on a EC2 instance (anecdotal to that specific branch also doesn't use engine at all). Also this adds in unneeded complexity.

PR #1302, opened 2023-08-06
Bitfinex: Fix ws Unsubscribe and Resubscribe

gbjk:
Unsubscribe needed to use the channel id.
Resubscribe needs to have the original subscription params.

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] Bug fix (non-breaking change which fixes an issue)

## How has this been tested

- [x] go test ./... -race
- [] golangci-lint run

PR #1311, opened 2023-08-08
Bitfinex: Fix WS handling of executed trades

gbjk:
Fixes:
`[ERROR] | WEBSOCKET | ... | 'EXECUTED @ 29298.0(-0.0008)' unrecognised order status`

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestWSHandleOrder

PR #1317, opened 2023-08-10
Bitfinex: Refactor wsUpdate handling

gbjk:
This commit doesn't break out all the sub-updates, but attempts to do something about the unmanagable size of ws update handling

**Note: This PR stacks on top of #1292, #1302 #1310**

## Type of change

- [x] OCD appeasement

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1073,6 +494,617 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
+	}
+
+	return fmt.Errorf("%s unhandled channel update: %s", b.Name, c.Channel)
+}
+
+func (b *Bitfinex) handleWSBookUpdate(c *stream.ChannelSubscription, d []interface{}) error {
+	var newOrderbook []WebsocketBook
+	obSnapBundle, ok := d[1].([]interface{})
+	if !ok {
+		return errors.New("orderbook interface cast failed")
```
shazbert, line 516: Could you please standardise type assertion error with `common.GetTypeAssertError` across just your changes.

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1073,6 +494,617 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
+			tradeHolder = append(tradeHolder, wsTrade)
+		}
+	case 3:
+		event, ok := d[1].(string)
+		if !ok {
+			return errors.New("unable to type assert data event")
+		}
+		if event != wsFundingTradeUpdate &&
+			event != wsTradeExecutionUpdate {
+			return nil
```
shazbert, line 822: This might better as an error return?

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1073,6 +494,617 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
+			}
+		}
+	}
+	return nil
+}
+
+func (b *Bitfinex) handleWSPositionSnapshot(d []interface{}) error {
+	snapBundle, ok := d[2].([]interface{})
+	if !ok || len(snapBundle) == 0 {
+		return nil
```
shazbert, line 963: return error

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1073,6 +494,617 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
+	}
+	if _, ok := snapBundle[0].([]interface{}); !ok {
+		return nil
```
shazbert, line 966: return error

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1073,6 +494,617 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
+		snapshot[i] = position
+	}
+	b.Websocket.DataHandler <- snapshot
+	return nil
+}
+
+func (b *Bitfinex) handleWSPositionUpdate(d []interface{}) error {
+	positionData, ok := d[2].([]interface{})
+	if !ok || len(positionData) == 0 {
+		return nil
```
shazbert, line 1016: return error

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1073,6 +494,617 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
+		return errors.New("unable to type assert position leverage")
+	}
+	b.Websocket.DataHandler <- position
+	return nil
+}
+
+func (b *Bitfinex) handleWSTradeUpdate(d []interface{}, eventType string) error {
+	tradeData, ok := d[2].([]interface{})
+	if !ok || len(tradeData) <= 4 {
+		return nil
```
shazbert, line 1058: return error

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1494,7 +1529,7 @@ func (b *Bitfinex) WsUpdateOrderbook(p currency.Pair, assetType asset.Item, book
 // GenerateDefaultSubscriptions Adds default subscriptions to websocket to be handled by ManageSubscriptions()
 func (b *Bitfinex) GenerateDefaultSubscriptions() ([]stream.ChannelSubscription, error) {
 	var wsPairFormat = currency.PairFormat{Uppercase: true}
-	var channels = []string{wsBook, wsTrades, wsTicker, wsCandles}
+	var channels = []string{wsBook}
```
shazbert, line 1532: Please reinstate the default subscriptions

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1622,11 +1669,43 @@ func (b *Bitfinex) WsSendAuth(ctx context.Context) error {
...
+func (b *Bitfinex) WsAddSubscriptionChannel(chanID int, channel, symbol string) error {
+	assetType, pair, err := assetPairFromSymbol(symbol)
+	if err != nil {
+		return err
+	}
+
+	var c *stream.ChannelSubscription
+	s := b.Websocket.GetSubscriptions()
+	for i := range s {
+		if strings.EqualFold(s[i].Channel, channel) && s[i].Currency.Equal(pair) {
```
shazbert, line 1682: Might need to distinguish between asset types as well

shazbert (changes requested):
Thanks for opening this PR. Just some things I noticed over first pass. 🚀

exchanges/bitfinex/bitfinex_test.go
```diff
@@ -1276,7 +1289,7 @@ func TestWsCandleResponse(t *testing.T) {
 }
 
 func TestWsOrderSnapshot(t *testing.T) {
-	b.WsAddSubscriptionChannel(0, "account", "")
+	_ = b.WsAddSubscriptionChannel(0, "account", "")
```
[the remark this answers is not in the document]
gbjk: These shouldn't even have been here. I dropped the ball. They're no longer necessary.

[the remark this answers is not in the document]
gbjk: ( And they're only _ *because* they were transitory )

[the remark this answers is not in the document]
gbjk: Fixed by removing them

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Yep.

gbjk: FWIW, I just moved these around.

gbjk: Actually, this is a huge request, and not something I introduced.
I've reached my yak limit on this. I'll add it for a separate task.

gbjk: Also: I think we'd use one of:
* [array to struct](https://eagain.net/articles/go-json-array-to-struct/)
* And there's a note from @thrasher- to consider b2dfef87989d1ecbe35822e0cd973bb47ada150a

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Agreed. Done.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Needs breaking into 2 but yeah, can do.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Actually ... I have no idea why this is here. Deleting it.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Done.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: I can return error on failed conversion but returning error for tradeData < 4 is more intrusive than this refactor should be doing.
If it's an (existing) issue let's open a separate issue

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Whooooooa!
That's horrific. Sorry, really bad mistake. Would have happened during rebase.

Thanks for the catch 🙏

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Agreed.
Note however that the [existing CS Equal method](https://github.com/thrasher-corp/gocryptotrader/blob/master/exchanges/stream/websocket.go#L956-L960) does not.

gbjk: Ah, right. I didn't do this because there's no guarantee at all that they'll match.
I had conflicting use-cases of empty asset and asset.Spot.
I backed away from it as a result.

Tests are passing now, so it was an issue in my runtime.

I'll make the change as requested and then revisit if I have issues again.

Done 👍

gbjk:
@shazbert Done.

Declined the type assertion fix - It's unchanged from original, and another big chore.
Also: Better ways to handle it outlined.

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1348,8 +1385,9 @@ func (b *Bitfinex) wsHandleOrder(data []interface{}) {
 		}
 	}
 	if data[13] != nil {
-		if ordStatus, ok := data[13].(string); ok {
-			oStatus, err := order.StringToOrderStatus(ordStatus)
+		if combinedStatus, ok := data[13].(string); ok {
+			statusParts := strings.Split(combinedStatus, " @ ")
```
gloriousCode, line 1389: Do you have a reference for this change? The split is fine, I'm just wondering what this is doing here

gloriousCode (commented):
This is lovely. Thank you for putting some extra attention to Bitfinex. My actual change requests are really quite tiny

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -218,476 +220,19 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
-				}
-			}
-
-			return b.AddTradesToBuffer(trades...)
+			return fmt.Errorf("unable to locate chanID: %d", chanID)
 		}
 
-		if authResp, ok := d[1].(string); ok {
-			switch authResp {
+		if eventType, ok := d[1].(string); ok {
```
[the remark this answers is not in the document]
gbjk: Thanks, but I'll need help here.
Why do you want to avoid localising ok here?
Whenever there's an option to limit scope I want to, since it also reduces risk.
I know this shadows the ok from chanF. That's also kind of the point too.

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1622,11 +1674,43 @@ func (b *Bitfinex) WsSendAuth(ctx context.Context) error {
...
+	s := b.Websocket.GetSubscriptions()
+	for i := range s {
+		if strings.EqualFold(s[i].Channel, channel) && s[i].Currency.Equal(pair) && s[i].Asset == assetType {
+			c = &s[i]
+			break
+		}
+	}
+
+	if c == nil {
+		log.Errorf(log.ExchangeSys,
```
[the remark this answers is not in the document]
gbjk: Thanks.
I thought about this a fair bit.
Outside of tests, I think it actually is an error always, but not enough to stop us falling back to a default empty channel to track _something_.

Happy to handle this issue afterwards if you want me to.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Yes 🎉 
This change is one of the rolled in PRs. See #1311

exchanges/bitfinex/bitfinex_websocket.go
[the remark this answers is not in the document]
gloriousCode: The values don't really need to be protected like that inside this function, the previous usage of it is the parse the exact same information. So we're checking if `d[1]` is a string twice and assigning it to a temp variable twice.

The shadow declaration is something my IDE likes to highlight and something I should get out of the habit of commenting on, the practice of what you're doing is something I like. So I wouldn't actually worry about changing it unless any other change requests come up.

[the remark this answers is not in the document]
gbjk: This actually triggered me to fix L194. I hadn't seen that.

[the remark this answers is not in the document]
gbjk: I was too gentle. Let me man-handle this a little more.
I know I'm risking alimony but I can't just ignore it.
Just one commit more for @shazbert to review too. Soz!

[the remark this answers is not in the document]
gbjk: 👍  Okay. That's better.
This triggered a slight rampage, which was probably overdue anyway

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -182,897 +184,929 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
+	if f, ok := d[2].(float64); !ok {
+		return common.GetTypeAssertError("float64", d[2], "checksum")
+	} else {
+		token = int(f)
+	}
+
+	var seqNo int64
+	if f, ok := d[3].(float64); !ok {
+		return common.GetTypeAssertError("float64", d[3], "seqNo")
+	} else {
```
gloriousCode, line 506: I think you're going to need to `//nolint:indent-error-flow // using temporary variable requires else statement` (or nolint:revive) otherwise this will likely make all our PR linting fail

gloriousCode (changes requested):
Everything appears to be working well, I like the improvements and thanks for the latest changes! I can only request changes for using `//nolint`

exchanges/bitfinex/bitfinex_websocket.go
shazbert: Or could even do this:

```go
seqNo , ok := d[3].(float64)
if !ok {
	return common.GetTypeAssertError("float64", d[3], "seqNo")
}
```
image

shazbert (changes requested):

shazbert (approved):
LGTM tACK!

gloriousCode (approved):
Thanks for the lint update and thanks for combining everything into this one PR. tACK

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -1949,3 +2031,34 @@ subSort:
...
+	switch len(pairInfo) {
+	case 1:
+		newPair := pairInfo[0]
+		if newPair[0] == 'f' {
+			assetType = asset.MarginFunding
+		}
+		symbol = newPair[1:]
+	case 2:
+		assetType = asset.Margin
+		symbol = symbol[1:]
```
thrasher-, line 2052: Note that a spot/margin asset can both contain a ":" delimiter. A pattern I've noticed for bitfinex spot/margin pairs is that anything where the base + quote len > 6, requires it. With derivatives, the pair always has one but GCT does not support it on websocket. As it stands, any pair with a len > 6 will throw a subscription issue so I've made a patch to address this.

[1.patch](https://github.com/thrasher-corp/gocryptotrader/files/12450131/1.patch)

Next up, with the delimiter, this would return an enabled spot pair like DOGE:USD or LUNA2:USD as a margin asset. As a separate issue with this, margin assets with a len of 6 like IOTUSD would not contain a delimiter in its returned symbol which this would currently report it as a spot asset.

thrasher- (changes requested):
Some very nice improvements, thank you sir @gbjk ! Found some issues:

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Thanks.
I've applied the patch and it makes sense.
Looking into assetPairFromSymbol now.

gbjk: Okay. Think that's solved it.
Tests added.

gbjk: Nope. More to do. I'll reply when I'm done.

gbjk: Good to go now. Disabled Margin assets for WS until we solve the collisions, multiplex the responses (🤮 ) or something else.

thrasher- (changes requested):
I still noticed subscription errors for the added test case in this patch. I also made subscription of pairs > len 6 always contain a delimiter for all subscription types, because this was causing issues for the assetPairFromSymbol func if there was a pair like DOGEUSDT which didn't contain a delimiter (DOGEUSDT and DOGE:USDT are both valid in Bitfinex's subscription case and will return either, depending on the subscribed pair format. NOTE: the changes in assetPairFromSymbol I've made could be done on the key wsUpdate handling side, either way is fine with me.
Patch diff:

[1.patch](https://github.com/thrasher-corp/gocryptotrader/files/12471422/1.patch)

gbjk:
@thrasher- Fixed and improved. Thanks for catching that.
I've aimed to be more specific without adding any new codewhiff.

exchanges/bitfinex/bitfinex_wrapper.go
```diff
@@ -85,6 +85,13 @@ func (b *Bitfinex) SetDefaults() {
 	if err != nil {
 		log.Errorln(log.ExchangeSys, err)
 	}
+
+	// Margin WS Currently not fully implemented and causes subscription collisions with spot
+	err = b.DisableAssetWebsocketSupport(asset.Margin)
```
thrasher-, line 90: Now that this is added, will just need a check added to the GenerateDefaultSubscriptions() method and the syncer will take care of the rest:

https://github.com/gbjk/gocryptotrader/blob/bugfix/bfx_ws_refactor/exchanges/bitfinex/bitfinex_websocket.go#L1540

		if !b.IsAssetWebsocketSupported(assets[i]) {
			continue
		}

thrasher- (changes requested):
Thanks for making those changes, subs are working great! This is a tACK from me after these two minor changes

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -146,33 +145,38 @@ func (b *Bitfinex) wsHandleData(respRaw []byte) error {
...
-					}
+			symbol, ok := d["symbol"].(string)
+			if !ok {
+				key, ok := d["key"].(string)
+				if !ok {
+					return fmt.Errorf("subscribed to channel but no symbol or key: %v", channel)
+				}
+				if channel != wsCandles {
+					// status channel not implemented at all yet.
+					return errors.Wrapf(common.ErrNotYetImplemented, "%v channel subscription keys", channel)
```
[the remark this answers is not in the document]
gbjk: Agreed. pkg/errors has been my default since [this go blog post](https://go.dev/blog/go1.13-errors) and I hadn't really considered before if it actually adds anything. It doesn't seem to.

exchanges/bitfinex/bitfinex_wrapper.go
gbjk: Gotcha.

gbjk:
@thrasher- Changes applied.

thrasher- (approved):
tACK, nice stuff!

PR #1319, opened 2023-08-11
Bitstamp: Fix websocket orderbook snapshot blanking LastUpdate

gbjk:
Drive-By: Remove:
`Bitstamp BTC/USDT spot orderbook has zero bid price, filtering.` It's unactionable and happens every snapshot right now

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/bitstamp/bitstamp_websocket.go
+++ b/exchanges/bitstamp/bitstamp_websocket.go
@@ -254,7 +254,7 @@ func (b *Bitstamp) wsUpdateOrderbook(update *websocketOrderBook, p currency.Pair
 		Bids:            make(orderbook.Items, len(update.Bids)),
 		Asks:            make(orderbook.Items, len(update.Asks)),
 		Pair:            p,
-		LastUpdated:     time.Unix(update.Timestamp, 0),
+		LastUpdated:     time.UnixMicro(update.Microtimestamp),
 		Asset:           assetType,
 		Exchange:        b.Name,
 		VerifyOrderbook: b.CanVerifyOrderbook,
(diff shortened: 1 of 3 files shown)


shazbert (approved):
tACK. Happy for this to go, we could even wrap it in a verbose check for future monitoring so we can remove if they ever update it but that seems unlikely.

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -309,6 +309,7 @@ func (b *Bitstamp) seedOrderBook(ctx context.Context) error {
 			VerifyOrderbook: b.CanVerifyOrderbook,
 			Bids:            make(orderbook.Items, len(orderbookSeed.Bids)),
 			Asks:            make(orderbook.Items, len(orderbookSeed.Asks)),
+			LastUpdated:     time.Now(),
```
shazbert, line 312: Just noticed this, can we use the orderbookSeed timestamp so its closer to the matching engine time. and also in `wsUpdateOrderbook` we can use `time.UnixMicro(update.Microtimestamp)` for the `lastupdate` field.

shazbert (changes requested):
Just noticed something on changes.

exchanges/bitstamp/bitstamp_websocket.go
gbjk: Agreed. On it.

shazbert (approved):
tACK. Thanks you!

gloriousCode (approved):
tACK! RIP log spam

thrasher- (approved):
tACK, thanks @gbjk !

PR #1328, opened 2023-08-26
Orderbook: Add GetDepth to Base

gbjk:
Base.GetDepth returns the concrete book of which Base is a copy This is probably useful for immutably monitoring orderbook health and state whereas FetchOrderbook would trigger a refresh.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/orderbook/orderbook.go
+++ b/exchanges/orderbook/orderbook.go
@@ -201,9 +201,7 @@ func (s *Service) Retrieve(exchange string, p currency.Pair, a asset.Item) (*Bas
 	return book.Retrieve()
 }
 
-// GetDepth returns the concrete book of which Base is a copy
-// This is probably useful for immutably monitoring orderbook health and state
-// whereas FetchOrderbook would trigger a refresh.
+// GetDepth returns the concrete book allowing the caller to stream orderbook changes
 func (b *Base) GetDepth() (*Depth, error) {
 	return service.GetDepth(b.Exchange, b.Pair, b.Asset)
 }
--- a/exchanges/orderbook/orderbook.go
+++ b/exchanges/orderbook/orderbook.go
@@ -17,8 +17,7 @@ func Get(exchange string, p currency.Pair, a asset.Item) (*Base, error) {
 	return service.Retrieve(exchange, p, a)
 }
 
-// GetDepth returns a Depth pointer allowing the caller to stream orderbook
-// changes
+// GetDepth returns a Depth pointer allowing the caller to stream orderbook changes
 func GetDepth(exchange string, p currency.Pair, a asset.Item) (*Depth, error) {
 	return service.GetDepth(exchange, p, a)
 }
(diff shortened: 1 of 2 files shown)


exchanges/orderbook/orderbook.go
```diff
@@ -202,6 +201,13 @@ func (s *Service) Retrieve(exchange string, p currency.Pair, a asset.Item) (*Bas
 	return book.Retrieve()
 }
 
+// GetDepth returns the concrete book of which Base is a copy
+// This is probably useful for immutably monitoring orderbook health and state
+// whereas FetchOrderbook would trigger a refresh.
```
gloriousCode, line 206:
```suggestion
// GetDepth returns the concrete book allowing the caller to stream orderbook changes
```
FetchOrderbook desn't necessarily trigger a refresh

gloriousCode (changes requested):
Base already has access to the fields to make the call for `GetDepth`, but this is still a nice enough helper to avoid needing to provide params anyway.

exchanges/orderbook/orderbook.go
```diff
@@ -202,6 +201,13 @@ func (s *Service) Retrieve(exchange string, p currency.Pair, a asset.Item) (*Bas
...
+func (b *Base) GetDepth() (*Depth, error) {
```
shazbert, line 207: Add test please

shazbert (changes requested):
One nit.

exchanges/orderbook/orderbook.go
gbjk: Okay. So s/ would / might / and add context.

gbjk:
@gloriousCode 
> Base already has access to the fields to make the call for `GetDepth`, but this is still a nice enough helper to avoid needing to provide params anyway.

Base does, but nothing outside it does, right?
In the case of a consumer of the Bitfinex exchange, calling FetchOrderbook formats the pair, and then fetches the orderbook.
You can't call GetDepth without knowing the pair, and you can't find out if the orderbook is valid without risking a fetch.
There's a few other options, but this felt easiest.

exchanges/orderbook/orderbook.go
gbjk: Done. Thanks.

exchanges/orderbook/orderbook.go
gbjk: Done, exactly as per suggestion.

gbjk:
All comments addressed.

We discussed why this method is helpful, and with some refactoring/improvement/something in bitfinex maybe we could get away with calling FormatExchangePair before calling Fetch ever, or something.
But it's a can of worms so for now this solves my need and isn't too code-rotty.

exchanges/orderbook/orderbook_test.go
```diff
@@ -294,6 +294,38 @@ func TestGetDepth(t *testing.T) {
...
+	if result, err := GetDepth("Exchange", c, asset.Spot); err != nil {
+		t.Errorf("failed to get orderbook. Error %s", err)
+	} else if !result.pair.Equal(c) {
+		t.Errorf("Mismatched pairs: %v %v", result.pair, c)
+	}
+
+	if c, err := currency.NewPairFromStrings("BTC", "DOGE"); err != nil {
+		t.Error(err)
+	} else {
+		if _, err = GetDepth("Exchange", c, asset.Futures); !errors.Is(err, errCannotFindOrderbook) {
```
gloriousCode, line 323: Shouldn't this be calling `base.GetDepth()`?

gbjk: But ... then it might fail! 😉

gbjk: Fixed. Thanks! 🙏

gbjk:
Ah. This is failing when run as a whole suite. Will work out why.
Probably service is populated by prev test

gbjk:
Yep. It was that. Fixed.

shazbert (approved):
tACK. Thanks.

gloriousCode (approved):

PR #1333, opened 2023-09-05
Bitstamp: Add auth channel subscription handling

gbjk:

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestWsOrderUpdate


exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+	case "order_deleted":
+		if o.RemainingAmount == 0 && o.Amount > 0 {
+			status = order.Filled
+		} else {
+			status = order.Cancelled
 		}
-	default:
-		b.Websocket.DataHandler <- stream.UnhandledMessageWarning{Message: b.Name + stream.UnhandledMessage + string(respRaw)}
 	}
+
```
gbjk, line 211: orderType could be derived from status == New and Buy & price == 9+e9 or Sell & price == 0
But it would only be true for the first update and it really doesn't feel worth the risk
Consumers are going to have to merge to original request anyway

gbjk:
Note the switch of the PairManager to a pointer is explained in [this commit message](https://github.com/thrasher-corp/gocryptotrader/pull/1333/commits/223ac4945bec11bc1880f55c8127deb4ab641210) for discussion if necessary

gbjk:
Hold off review. Erroring heavily and needs some TLC.

gbjk:
Okay. Sorted.
I reversed direction and took the path that impacts production code the least

currency/manager.go
```diff
@@ -45,6 +45,18 @@ func (p *PairsManager) GetAssetTypes(enabled bool) asset.Items {
...
+func (p *PairsManager) Clone() (PairsManager, error) {
+	var pN PairsManager
+	j, err := json.Marshal(p)
+	if err == nil {
+		err = json.Unmarshal(j, &pN)
+	}
+
+	return pN, err //nolint:govet // copylocks not relevant; We're cloning
+}
+
```
gloriousCode, line 59: I think I'd prefer the removal of this function, see my other comments. If you do keep this, it will need a test

exchanges/bitstamp/bitstamp.go
```diff
@@ -43,7 +43,14 @@ const (
...
 	bitstampAPITradingPairsInfo   = "trading-pairs-info"
-	bitstampOHLC                  = "ohlc"
+	bitstampAPIWSAuthToken        = "websockets_token"
+	bitstampAPIWSTrades           = "live_trades"
+	bitstampAPIWSOrders           = "live_orders"
+	bitstampAPIWSOrderbook        = "order_book"
+	bitstampAPIWSMyOrders         = "my_orders"
+	bitstampAPIWSMyTrades         = "my_trades"
+
+	bitstampOHLC = "ohlc"
```
gloriousCode, line 53: Seems like this guy could rejoin his friends

exchanges/bitstamp/bitstamp_types.go
```diff
@@ -246,3 +257,20 @@ type OHLCResponse struct {
 		} `json:"ohlc"`
 	} `json:"data"`
 }
+
+type websocketOrderResponse struct {
+	websocketResponse
+	Order websocketOrderData `json:"data"`
+}
+type websocketOrderData struct {
```
gloriousCode, line 265:
```suggestion

type websocketOrderData struct {
```

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+	}
+	return nil
+}
 
-		p, err := currency.NewPairFromFormattedPairs(currencyPair, enabledPairs, pFmt)
-		if err != nil {
-			return err
-		}
+func (b *Bitstamp) handleWSOrderbook(_ context.Context, wsResp *websocketResponse, msg []byte) error {
+	if wsResp.pair == currency.EMPTYPAIR {
```
gloriousCode, line 138:
```suggestion
	if wsResp.pair.IsEmpty() {
```

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+func (b *Bitstamp) handleWSTrade(_ context.Context, wsResp *websocketResponse, msg []byte) error {
+	if !b.IsSaveTradeDataEnabled() {
+		return nil
+	}
 
-		p, err := currency.NewPairFromFormattedPairs(currencyPair, enabledPairs, pFmt)
-		if err != nil {
-			return err
-		}
+	if wsResp.pair == currency.EMPTYPAIR {
```
gloriousCode, line 156:
```suggestion
	if wsResp.pair.IsEmpty() {
```

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+		return nil
+	}
+
+	r := &websocketOrderResponse{}
+	if err := json.Unmarshal(msg, &r); err != nil {
+		return err
+	}
+
+	o := r.Order
+	if o.ID == 0 {
```
gloriousCode, line 192: I think you should only be returning if both the clientorderid and id is not populated

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -344,3 +413,58 @@ func (b *Bitstamp) seedOrderBook(ctx context.Context) error {
...
+	switch len(authParts) {
+	case 1:
+		// Not an auth channel
+	case 3:
+		chanName = authParts[1]
+	default:
+		return fmt.Errorf("channel name does not contain exactly 0 or 2 hyphens: %v", r.Channel)
+	}
+
+	parts := strings.Split(chanName, currency.UnderscoreDelimiter)
```
gloriousCode, line 445: Would be nicer to have the consistency of "-" or currency.DashDelimiter etc instead of both

exchanges/sharedtestvalues/sharedtestvalues.go
```diff
@@ -147,3 +150,38 @@ func ForceFileStandard(t *testing.T, pattern string) error {
 	}
 	return nil
 }
+
+// TestFixtureToDataHandler takes a new empty exchange and configures a new websocket handler for it, and squirts the json path contents to it
```
gloriousCode, line 154: > squirts

😭

exchanges/sharedtestvalues/sharedtestvalues.go
```diff
@@ -147,3 +150,38 @@ func ForceFileStandard(t *testing.T, pattern string) error {
...
+// It accepts a reader function, which is probably e.wsHandleData but could be anything
+func TestFixtureToDataHandler(t *testing.T, seed, e exchange.IBotExchange, fixturePath string, reader func([]byte) error) {
+	b := e.GetBase()
+	p, err := seed.GetBase().CurrencyPairs.Clone()
+	if err != nil {
+		t.Error(err)
+		return
+	}
+	b.CurrencyPairs = p //nolint:govet // copylocks not relevant; We just cloned it
```
gloriousCode, line 163:
```go
	newBase := e.GetBase()
	seedBase := seed.GetBase()

	newBase.CurrencyPairs.RequestFormat = seedBase.CurrencyPairs.RequestFormat
	newBase.CurrencyPairs.ConfigFormat = seedBase.CurrencyPairs.ConfigFormat
	newBase.CurrencyPairs.UseGlobalFormat = seedBase.CurrencyPairs.UseGlobalFormat
	newBase.CurrencyPairs.Pairs = seedBase.CurrencyPairs.Pairs
```

Given this is for a test, I don't think `Clone` as a func is necessary and can only just copy what is relevant. It also removes nolints and copying of locks

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -344,3 +413,58 @@ func (b *Bitstamp) seedOrderBook(ctx context.Context) error {
...
+	if err != nil {
+		return err
+	}
+
+	p, err := currency.NewPairFromFormattedPairs(symbol, enabledPairs, pFmt)
+	if err != nil {
+		return err
+	}
+
+	r.pair = p
```
gloriousCode, line 466:
```suggestion
	enabledPairs, err := b.GetEnabledPairs(asset.Spot)
	if err != nil {
		return err
	}

	r.pair, err = enabledPairs.DeriveFrom(symbol)
```

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -344,3 +413,58 @@ func (b *Bitstamp) seedOrderBook(ctx context.Context) error {
...
+		return nil
+	}
+
+	chanName := r.Channel
+	authParts := strings.Split(r.Channel, "-")
+	switch len(authParts) {
+	case 1:
+		// Not an auth channel
+	case 3:
+		chanName = authParts[1]
```
gloriousCode, line 440: Technically the pair is not required for the subscription:
```
[DEBUG] | WEBSOCKET | 07/09/2023 09:40:29 | Bitstamp websocket connection: message received: {"event":"bts:subscription_succeeded","channel":"private-my_orders-3215483","data":{}}
[ERROR] | WEBSOCKET | 07/09/2023 09:40:29 | exchange Bitstamp websocket error - unable to parse currency pair from wsResponse.Channel: channel name does not contain exactly 2 underscores: private-my_orders-3215483
[DEBUG] | WEBSOCKET | 07/09/2023 09:40:29 | Bitstamp websocket connection: message received: {"event":"bts:subscription_succeeded","channel":"private-my_trades-3215483","data":{}}
```
While I prefer the way you've setup the subscriptions in code to do it with a currencypair, a user is still able to subscribe to endpoints on their own and this will error out

gloriousCode (changes requested):
Nice work! Thanks for adding this

exchanges/bitstamp/bitstamp.go
gbjk: If he behaves.

But if I find him smoking crack behind the bike shed again then that's the final warning!

exchanges/bitstamp/bitstamp_websocket.go
gbjk: Agreed.
Note: I couldn't seem to make an order without getting an ID, and their API accepts their order ID and never clientID for subsequent actions.
Without an ID, I don't think we can meaningfully interact with this order.
However I accept that it's not this function's job to decide that, and if the subscriber can updateDetails and make sense on them, good luck with that.

exchanges/bitstamp/bitstamp_websocket.go
gbjk: We're not setting orderType ( Market vs Limit )

We *could* try to work it out, but it's messy, it'd be inconsistent, it's not documented, and I think it's not worth the risk of trying.

They use price 999999999 for a market buy. I really don't want to try to hang trading decisions off an undocumented feature like that.

exchanges/bitstamp/bitstamp_websocket.go
gbjk: Agreed. I'll fix this and add a test.

Also: Yet another undocumented feature. It [explicitly says you have to supply currencyPair](https://www.bitstamp.net/websocket/v2/).

exchanges/bitstamp/bitstamp_websocket.go
gbjk: Oooh. Much better. Thank you 🙏

exchanges/bitstamp/bitstamp_websocket.go
gloriousCode: Oh you don't really need to worry then. I just saw that it worked and didn't check that it wasn't supported sorry

gbjk: Oh, yeah.... cos if we can't get the pair from the channel name, it's not provided in any other way.

Think I'll sniff this a bit more. Want to see what we get back with this sub...

gbjk: I'll investigate it separately from this PR, actually.

exchanges/sharedtestvalues/sharedtestvalues.go
gbjk: Sure, I think that makes sense.

gbjk: Irony: The reason I went to use Clone was because I reasoned we should lock the seed base to copy it.
And then forgot to do so!

Fixed.

currency/manager.go
gbjk: Done.
I've punished you by adding 2 new methods instead 😈

gbjk: Alternative is to expose the mutex ... actually I prefer that more.
Imma change it.

exchanges/bitstamp/bitstamp.go
```diff
@@ -53,6 +59,8 @@ const (
 // Bitstamp is the overarching type across the bitstamp package
 type Bitstamp struct {
 	exchange.Base
+
+	wsAuthToken websocketAuthResponse
```
shazbert, line 63: Can RM this doesn't look like its being used?

exchanges/bitstamp/bitstamp_type_convert.go
```diff
@@ -27,3 +80,47 @@ func (p *TradingPair) UnmarshalJSON(data []byte) error {
...
+	if err := json.Unmarshal(data, a); err != nil {
+		return err
+	}
+
+	o.Datetime = a.Datetime.Time()
+	o.Microtimestamp = a.Microtimestamp.Time()
+	o.Side = a.Side.Side()
+
+	return nil
+}
```
shazbert, line 126: Can RM this and the test. 
```diff
diff --git a/exchanges/bitstamp/bitstamp.go b/exchanges/bitstamp/bitstamp.go
index 5f32e9d99..e30f98e3a 100644
--- a/exchanges/bitstamp/bitstamp.go
+++ b/exchanges/bitstamp/bitstamp.go
@@ -59,8 +59,6 @@ const (
 // Bitstamp is the overarching type across the bitstamp package
 type Bitstamp struct {
        exchange.Base
-
-       wsAuthToken websocketAuthResponse
 }

 // GetFee returns an estimate of fee based on type of transaction
diff --git a/exchanges/bitstamp/bitstamp_test.go b/exchanges/bitstamp/bitstamp_test.go
index 67b025de9..790795dff 100644
--- a/exchanges/bitstamp/bitstamp_test.go
+++ b/exchanges/bitstamp/bitstamp_test.go
@@ -696,27 +696,6 @@ func TestParseTime(t *testing.T) {
        }
 }

-func TestWebsocketOrderDataUnmarshalJSON(t *testing.T) {
-       t.Parallel()
-
-       var o websocketOrderData
-       if err := o.UnmarshalJSON([]byte(`{"id":1658394968748032,"id_str":"1658394968748032","client_order_id":"balrog-42","order_type":1,"datetime":"1231006505","microtimestamp":"1231006505314159","amount":0.0004,"amount_str":"0.00040000","amount_traded":"0.0001","amount_at_create":"0.00060000","price":28265,"price_str":"28265","trade_account_id":12345678}`)); err != nil {
-               t.Error(err)
-       } else {
-               genesis := time.UnixMicro(1231006505314159)
-               assert.Equal(t, int64(1658394968748032), o.ID, "ID")
-               assert.Equal(t, "1658394968748032", o.IDStr, "IDStr")
-               assert.Equal(t, "balrog-42", o.ClientOrderID, "ClientOrderID")
-               assert.Equal(t, genesis.Truncate(time.Second), o.Datetime, "Datetime")
-               assert.Equal(t, genesis, o.Microtimestamp, "Microtimestamp")
-               assert.Equal(t, order.Sell, o.Side, "Side")
-               assert.Equal(t, 0.0006, o.Amount, "Amount")
-               assert.Equal(t, 0.0001, o.ExecutedAmount, "ExecutedAmount")
-               assert.Equal(t, 0.0004, o.RemainingAmount, "RemainingAmount")
-               assert.Equal(t, 28265.0, o.Price, "Price")
-       }
-}
-
 func TestWsSubscription(t *testing.T) {
        pressXToJSON := []byte(`{
                "event": "bts:subscribe",
diff --git a/exchanges/bitstamp/bitstamp_type_convert.go b/exchanges/bitstamp/bitstamp_type_convert.go
index a39187e7d..e0af07f1c 100644
--- a/exchanges/bitstamp/bitstamp_type_convert.go
+++ b/exchanges/bitstamp/bitstamp_type_convert.go
@@ -103,24 +103,3 @@ func (s *orderSide) UnmarshalJSON(data []byte) error {
 func (s *orderSide) Side() order.Side {
        return order.Side(*s)
 }
-
-func (o *websocketOrderData) UnmarshalJSON(data []byte) error {
-       type Alias websocketOrderData
-       a := &struct {
-               *Alias
-               Side           orderSide      `json:"order_type"`
-               Datetime       datetime       `json:"datetime"`
-               Microtimestamp microTimestamp `json:"microtimestamp"`
-       }{
-               Alias: (*Alias)(o),
-       }
-       if err := json.Unmarshal(data, a); err != nil {
-               return err
-       }
-
-       o.Datetime = a.Datetime.Time()
-       o.Microtimestamp = a.Microtimestamp.Time()
-       o.Side = a.Side.Side()
-
-       return nil
-}
diff --git a/exchanges/bitstamp/bitstamp_types.go b/exchanges/bitstamp/bitstamp_types.go
index 800744816..ece0246c6 100644
--- a/exchanges/bitstamp/bitstamp_types.go
+++ b/exchanges/bitstamp/bitstamp_types.go
@@ -6,7 +6,6 @@ import (
        "time"

        "github.com/thrasher-corp/gocryptotrader/currency"
-       "github.com/thrasher-corp/gocryptotrader/exchanges/order"
 )

 // Transaction types
@@ -264,14 +263,14 @@ type websocketOrderResponse struct {
 }

 type websocketOrderData struct {
-       ID              int64      `json:"id"`
-       IDStr           string     `json:"id_str"`
-       ClientOrderID   string     `json:"client_order_id"`
-       RemainingAmount float64    `json:"amount"`
-       ExecutedAmount  float64    `json:"amount_traded,string"` // Not Cumulative; Partial fill amount
-       Amount          float64    `json:"amount_at_create,string"`
-       Price           float64    `json:"price"`
-       Side            order.Side `json:"order_type"`
-       Datetime        time.Time  `json:"datetime"`
-       Microtimestamp  time.Time  `json:"microtimestamp"`
+       ID              int64          `json:"id"`
+       IDStr           string         `json:"id_str"`
+       ClientOrderID   string         `json:"client_order_id"`
+       RemainingAmount float64        `json:"amount"`
+       ExecutedAmount  float64        `json:"amount_traded,string"` // Not Cumulative; Partial fill amount
+       Amount          float64        `json:"amount_at_create,string"`
+       Price           float64        `json:"price"`
+       Side            orderSide      `json:"order_type"`
+       Datetime        datetime       `json:"datetime"`
+       Microtimestamp  microTimestamp `json:"microtimestamp"`
 }
diff --git a/exchanges/bitstamp/bitstamp_websocket.go b/exchanges/bitstamp/bitstamp_websocket.go
index c3b5de1a0..1f01f9aaa 100644
--- a/exchanges/bitstamp/bitstamp_websocket.go
+++ b/exchanges/bitstamp/bitstamp_websocket.go
@@ -220,10 +220,10 @@ func (b *Bitstamp) handleWSOrder(_ context.Context, wsResp *websocketResponse, m
                Exchange:        b.Name,
                OrderID:         o.IDStr,
                ClientOrderID:   o.ClientOrderID,
-               Side:            o.Side,
+               Side:            o.Side.Side(),
                Status:          status,
                AssetType:       asset.Spot,
-               Date:            o.Microtimestamp,
+               Date:            o.Microtimestamp.Time(),
                Pair:            wsResp.pair,
        }
```

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+		b.Websocket.DataHandler <- stream.UnhandledMessageWarning{Message: b.Name + stream.UnhandledMessage + string(respRaw)}
+	}
+	return nil
+}
 
-		p, err := currency.NewPairFromFormattedPairs(currencyPair, enabledPairs, pFmt)
-		if err != nil {
-			return err
-		}
+func (b *Bitstamp) handleWSOrderbook(_ context.Context, wsResp *websocketResponse, msg []byte) error {
```
shazbert, line 137: rm param `_ context.Context,`

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
-			return err
-		}
+	return b.wsUpdateOrderbook(&wsOrderBookTemp.Data, wsResp.pair, asset.Spot)
+}
 
-		enabledPairs, err := b.GetEnabledPairs(asset.Spot)
-		if err != nil {
-			return err
-		}
+func (b *Bitstamp) handleWSTrade(_ context.Context, wsResp *websocketResponse, msg []byte) error {
```
shazbert, line 151: rm param `_ context.Context,`

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+		AssetType:    asset.Spot,
+		Exchange:     b.Name,
+		Price:        wsTradeTemp.Data.Price,
+		Amount:       wsTradeTemp.Data.Amount,
+		Side:         side,
+		TID:          strconv.FormatInt(wsTradeTemp.Data.ID, 10),
+	})
+}
+
+func (b *Bitstamp) handleWSOrder(_ context.Context, wsResp *websocketResponse, msg []byte) error {
```
shazbert, line 181: rm param `_ context.Context,`

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+	if wsResp.channelType != bitstampAPIWSMyOrders {
+		return nil
```
shazbert, line 183: This could return an error if its something that's not set or completely different.

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -101,137 +117,190 @@ func (b *Bitstamp) wsHandleData(respRaw []byte) error {
...
+	}
+
+	r := &websocketOrderResponse{}
+	if err := json.Unmarshal(msg, &r); err != nil {
+		return err
+	}
+
+	o := r.Order
```
shazbert, line 191: Rm as we don't need to allocate a new struct

shazbert (changes requested):
Nice improvement! 🐦

exchanges/bitstamp/bitstamp_test.go
```diff
@@ -750,11 +772,85 @@ func TestWsOrderbook2(t *testing.T) {
...
+					assert.Equal(t, asset.Spot, v.AssetType, "AssetType")
+					assert.Equal(t, currency.NewPairWithDelimiter("BTC", "USD", "/"), v.Pair, "Pair")
+					assert.Equal(t, 0.0, v.ExecutedAmount, "ExecutedAmount")
+					assert.Equal(t, 999999999.0, v.Price, "Price") // Market Buy Price
+					// Note: Amount is 0 for market order create messages, oddly
+				case 2:
+					assert.Equal(t, "1658864794234880", v.OrderID, "OrderID")
+					assert.Equal(t, order.PartiallyFilled, v.Status, "Status")
+					assert.Equal(t, 0.00038667, v.Amount, "Amount")
+					assert.Equal(t, 0.00000001, v.RemainingAmount, "RemainingAmount") // During live tests we consistently got back this Sat remaining
```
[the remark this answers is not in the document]
gbjk: Yeah. I thought it was a one off, but it happens so consistently. Absurd.

currency/manager_types.go
```diff
@@ -14,7 +14,7 @@ type PairsManager struct {
 	UseGlobalFormat            bool        `json:"useGlobalFormat,omitempty"`
 	LastUpdated                int64       `json:"lastUpdated,omitempty"`
 	Pairs                      FullStore   `json:"pairs"`
-	m                          sync.RWMutex
+	Mutex                      sync.RWMutex
```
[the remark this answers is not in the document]
gbjk: Yep. Done and Thanks.

exchanges/bitstamp/bitstamp_test.go
```diff
@@ -750,11 +772,85 @@ func TestWsOrderbook2(t *testing.T) {
...
+					assert.Equal(t, order.Cancelled, v.Status, "Status") // Even though they probably consider it filled, Deleted + PartialFill = Cancelled
+					assert.Equal(t, 0.00038667, v.Amount, "Amount")
+					assert.Equal(t, 0.00000001, v.RemainingAmount, "RemainingAmount")
+					assert.Equal(t, 0.00038666, v.ExecutedAmount, "ExecutedAmount")
+					assert.Equal(t, 25862.0, v.Price, "Price")
+				case 4:
+					assert.Equal(t, "1658870500933632", v.OrderID, "OrderID")
+					assert.Equal(t, order.New, v.Status, "Status")
+					assert.Equal(t, order.Sell, v.Side, "Side")
+					assert.Equal(t, 0.0, v.Price, "Price") // Market Sell Price
```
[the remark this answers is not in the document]
gbjk: Yep. Shame, nevermind.

exchanges/bitstamp/bitstamp_type_convert.go
gbjk: Agreed.

gbjk: bf91a235 and 1c1cc7d5

exchanges/bitstamp/bitstamp.go
gbjk: 1c1cc7d5dc7c1c6f34b6068d59e670c9b3eff0d9

exchanges/bitstamp/bitstamp_websocket.go
gbjk: Yeah, okay.
When I finally crusade for universal context it won't make any difference if a few places already had it, especially if they were ignoring it.
Fixed: 5aedf4ad

exchanges/bitstamp/bitstamp_websocket.go
gbjk: I was going to comment to explain WHY we return here.
Then I realised it's an anti-pattern. It's not handleWSOrder's job to know what channel it got an order from, that's delegation of responsibilty.
So I've moved the responsibility up to the caller and added a comment.
See 681dbf21
Resolving assuming you'll be happy but please reopen if not.

exchanges/bitstamp/bitstamp_websocket.go
gbjk: Hmmph. Okay.
I think we're over optimising in situations like this, but I'll make-it-so in-lieu of benching it.

That said, I'm so used to "everything is a pointer" that I forget that there are so many value structs here too,
and that certainly adds weight to your request.

Fixed: 9a171776

gloriousCode (changes requested):
Looking good!

exchanges/sharedtestvalues/sharedtestvalues.go
```diff
@@ -147,3 +150,40 @@ func ForceFileStandard(t *testing.T, pattern string) error {
...
+func TestFixtureToDataHandler(t *testing.T, seed, e exchange.IBotExchange, fixturePath string, reader func([]byte) error) {
+	b := e.GetBase()
+	seedBase := seed.GetBase()
+
+	seedBase.CurrencyPairs.Mutex.RLock()
+	b.CurrencyPairs.RequestFormat = seedBase.CurrencyPairs.RequestFormat
+	b.CurrencyPairs.ConfigFormat = seedBase.CurrencyPairs.ConfigFormat
+	b.CurrencyPairs.UseGlobalFormat = seedBase.CurrencyPairs.UseGlobalFormat
+	b.CurrencyPairs.Pairs = seedBase.CurrencyPairs.Pairs
+	seedBase.CurrencyPairs.Mutex.RUnlock()
```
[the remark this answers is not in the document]
gbjk: @gloriousCode ff8717f4
Switched to using assert, but otherwise used as-is. Thanks again.

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -344,3 +411,47 @@ func (b *Bitstamp) seedOrderBook(ctx context.Context) error {
 	}
 	return nil
 }
+
+// fetchWSAuth Retrieves a userID and auth-token from REST for subscribing to a websocket channel
+// The token life-expectancy is only about 60s; use it immediately and do not store it
+func (b *Bitstamp) fetchWSAuth(ctx context.Context) (*websocketAuthResponse, error) {
```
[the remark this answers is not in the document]
gbjk: Assuming you're happy for it to be `sharedtestvalues.SkipTestIfCredentialsUnset` ?

shazbert (approved):
Changes ACK. Thanks!

exchanges/bitstamp/bitstamp_websocket.go
```diff
@@ -344,3 +411,47 @@ func (b *Bitstamp) seedOrderBook(ctx context.Context) error {
...
+		// Not an auth channel
+	case 3:
+		chanName = authParts[1]
+	default:
+		return fmt.Errorf("channel name does not contain exactly 0 or 2 hyphens: %v", r.Channel)
+	}
+
+	parts := strings.Split(chanName, "_")
+	if len(parts) != 3 {
+		return fmt.Errorf("%v: channel name does not contain exactly 2 underscores: %v", errWSPairParsingError, r.Channel)
```
gloriousCode, line 445:
```suggestion
		return fmt.Errorf("%w: channel name does not contain exactly 2 underscores: %v", errWSPairParsingError, r.Channel)
```

exchanges/bitstamp/bitstamp_websocket.go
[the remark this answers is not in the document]
gloriousCode: Yes please. A dedicated test and not just covered as part of another test. Also since its an API implementation, I think it should be exported

gloriousCode (changes requested):
Thanks for the updates, I think this should be my last review with comments

exchanges/sharedtestvalues/sharedtestvalues.go
```diff
@@ -157,12 +157,7 @@ func TestFixtureToDataHandler(t *testing.T, seed, e exchange.IBotExchange, fixtu
 	b := e.GetBase()
 	seedBase := seed.GetBase()
 
-	seedBase.CurrencyPairs.Mutex.RLock()
-	b.CurrencyPairs.RequestFormat = seedBase.CurrencyPairs.RequestFormat
-	b.CurrencyPairs.ConfigFormat = seedBase.CurrencyPairs.ConfigFormat
-	b.CurrencyPairs.UseGlobalFormat = seedBase.CurrencyPairs.UseGlobalFormat
-	b.CurrencyPairs.Pairs = seedBase.CurrencyPairs.Pairs
-	seedBase.CurrencyPairs.Mutex.RUnlock()
+	assert.NoError(t, b.CurrencyPairs.Load(&seedBase.CurrencyPairs), "Loading currency pairs should not error")
```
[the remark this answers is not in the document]
gbjk: * Agreed. In some places the functionality isn't clear enough, and I agree to breaking it out to err =.
So I'll do so here.
* assert.NoError and assert.Nil are the same. It's simply used for clarity. I'll use NoError 
* We don't *have* to add text to either Nil or NoError. We just should.
I try to stick to clearly descriptive errors which will be clear to newcomers when the test fails.

Fixed with 9e295cff

exchanges/bitstamp/bitstamp_websocket.go
gbjk: Thanks

exchanges/bitstamp/bitstamp_websocket.go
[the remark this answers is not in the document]
gbjk: - [x] Mock passing test
- [ ] Mock failing test ⛔ 
- [x] Live passing test manually
- [x] Changed to Public method

I've tried relatively hard do get a failing mock test but it would involve significant changes to mocks.
We request a fixed path, the only variables are the key and the signature, and the mocks ignore those values during matching.

A couple of options:
* Move the url to a var instead of const and change it to force the mock to use a different path
* Change the mock server

But for now I thought this is probably enough. It's more than we have for many other tests.

I've checked that we won't panic, etc with bad creds live.

88f65116a8b00daef112951ba15fc21b3f7cbc60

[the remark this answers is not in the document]
gloriousCode: A failing mock test is definitely not a requirement. I wouldn't even say a mock test is required for an authenticated function. Thank you for your efforts though, its still appreciated 🤙

gloriousCode (approved):
Hurray! Thanks for that, tACK!

shazbert (approved):
re-tACK

To note: 

```
[ERROR] | PORTFOLIO | 04/10/2023 14:32:41 | Error encountered retrieving exchange account info for Bitstamp. Error Bitstamp unsuccessful HTTP status code: 403 raw response: {"status": "error", "reason": "Invalid nonce", "code": "API0004"}, authenticated request failed
[ERROR] | ORDER | 04/10/2023 14:32:41 | Unable to get active orders for Bitstamp and asset type spot: Bitstamp unsuccessful HTTP status code: 403 raw response: {"status": "error", "reason": "Invalid nonce", "code": "API0004"}, authenticated request failed
```

Might be a timeclock issues this is sporadic.

thrasher- (approved):
tACK, great stuff @gbjk !

image

PR #1336, opened 2023-09-07
OrderManager: Fix race condition in submit with ws

gbjk:
If the ws sees the order before processSubmittedOrder then it will have assigned it an internal order id already and added it to the store. Don't treat that as an error. Instead just use the newer ws details

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestSubmitOrderAlreadyInStore

### DIFF
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -589,6 +589,14 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 		return nil, err
 	}
 
+	if err := m.orderStore.add(detail.CopyToPointer()); err == ErrOrdersAlreadyExists {
+		// Streamed by ws before we got here. Details from ws supecede since they are more recent.
+		detail = m.orderStore.getByDetail(detail)
+	} else if err != nil {
+		// Non-fatal error. OrderStore doesn't mean it wasn't submitted and consumers shouldn't get an error
+		log.Errorf(log.OrderMgr, "unable to add %v order %v to orderStore: %s", detail.Exchange, detail.OrderID, err)
+	}
+
 	msg := fmt.Sprintf("Exchange %s submitted order ID=%v [Ours: %v] pair=%v price=%v amount=%v quoteAmount=%v side=%v type=%v for time %v.",
 		detail.Exchange,
 		detail.OrderID,
@@ -606,13 +614,7 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 		m.orderStore.commsManager.PushEvent(base.Event{Type: "order", Message: msg})
 	}
 
-	err = m.orderStore.add(detail.CopyToPointer())
-	if err != nil {
-		return nil, fmt.Errorf("unable to add %v order %v to orderStore: %s",
-			detail.Exchange, detail.OrderID, err)
-	}
-
-	return &OrderSubmitResponse{Detail: detail, InternalOrderID: id.String()}, nil
+	return &OrderSubmitResponse{Detail: detail, InternalOrderID: detail.InternalOrderID.String()}, nil
 }
 
 // processOrders iterates over all exchange orders via API
@@ -1046,18 +1048,24 @@ func (s *store) upsert(od *order.Detail) (*OrderUpsertResponse, error) {
 
 // exists verifies if the orderstore contains the provided order
 func (s *store) exists(det *order.Detail) bool {
+	return s.getByDetail(det) != nil
+}
+
+// getByDetail fetches an order from the store and returns it
+// returns nil if not found
+func (s *store) getByDetail(det *order.Detail) *order.Detail {
 	if det == nil {
-		return false
+		return nil
 	}
 	s.m.RLock()
 	defer s.m.RUnlock()
 	exchangeOrders := s.Orders[strings.ToLower(det.Exchange)]
-	for x := range exchangeOrders {
-		if exchangeOrders[x].OrderID == det.OrderID {
-			return true
+	for _, o := range exchangeOrders {
+		if o.OrderID == det.OrderID {
+			return o
 		}
 	}
-	return false
+	return nil
 }
 
 // Add Adds an order to the orderStore for tracking the lifecycle
@@ -1070,7 +1078,7 @@ func (s *store) add(det *order.Detail) error {
 	if err != nil {
 		return err
 	}
-	if s.exists(det) { // TODO: Error on conflict; remove unnecessary locking.
+	if s.exists(det) {
 		return ErrOrdersAlreadyExists
 	}
 
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -1073,19 +1073,24 @@ func (s *store) add(det *order.Detail) error {
 	if det == nil {
 		return errNilOrder
 	}
+
 	name := strings.ToLower(det.Exchange)
-	_, err := s.exchangeManager.GetExchangeByName(name)
-	if err != nil {
+	if _, err := s.exchangeManager.GetExchangeByName(name); err != nil {
 		return err
 	}
-	if s.exists(det) {
-		return ErrOrdersAlreadyExists
+
+	s.m.Lock()
+	defer s.m.Unlock()
+
+	// Inline copy of getByDetail to avoid possible lock races
+	for _, o := range s.Orders[name] {
+		if o.OrderID == det.OrderID {
+			return ErrOrdersAlreadyExists
+		}
 	}
 
 	// Untracked websocket orders will not have internalIDs yet
 	det.GenerateInternalOrderID()
-	s.m.Lock()
-	defer s.m.Unlock()
 	s.Orders[name] = append(s.Orders[name], det)
 	if !det.AssetType.IsFutures() {
 		return nil
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -593,7 +593,7 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 		// Streamed by ws before we got here. Details from ws supersede since they are more recent.
 		detail = m.orderStore.getByDetail(detail)
 	} else if err != nil {
-		// Non-fatal error. OrderStore doesn't mean it wasn't submitted and consumers shouldn't get an error
+		// Non-fatal error: Unable to store order, but error does not need to be returned to caller
 		log.Errorf(log.OrderMgr, "unable to add %v order %v to orderStore: %s", detail.Exchange, detail.OrderID, err)
 	}
 
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -590,7 +590,7 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 	}
 
 	if err := m.orderStore.add(detail.CopyToPointer()); errors.Is(err, ErrOrdersAlreadyExists) {
-		// Streamed by ws before we got here. Details from ws supecede since they are more recent.
+		// Streamed by ws before we got here. Details from ws supersede since they are more recent.
 		detail = m.orderStore.getByDetail(detail)
 	} else if err != nil {
 		// Non-fatal error. OrderStore doesn't mean it wasn't submitted and consumers shouldn't get an error
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -1062,7 +1062,7 @@ func (s *store) getByDetail(det *order.Detail) *order.Detail {
 	exchangeOrders := s.Orders[strings.ToLower(det.Exchange)]
 	for _, o := range exchangeOrders {
 		if o.OrderID == det.OrderID {
-			return o
+			return o.CopyToPointer()
 		}
 	}
 	return nil
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -594,7 +594,7 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 		detail = m.orderStore.getByDetail(detail)
 	} else if err != nil {
 		// Non-fatal error: Unable to store order, but error does not need to be returned to caller
-		log.Errorf(log.OrderMgr, "unable to add %v order %v to orderStore: %s", detail.Exchange, detail.OrderID, err)
+		log.Errorf(log.OrderMgr, "Unable to add %v order %v to orderStore: %s", detail.Exchange, detail.OrderID, err)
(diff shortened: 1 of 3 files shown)


engine/order_manager.go
```diff
@@ -589,6 +589,14 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 		return nil, err
 	}
 
+	if err := m.orderStore.add(detail.CopyToPointer()); err == ErrOrdersAlreadyExists {
```
gloriousCode, line 592: Please use `errors.Is(...)`. We wrap errors and using `==` will not be able to handle that. Even if its fine now, better to properly handle it now than to rely on this and have it fail silently when it doesn't need to

```go
func TestError(t *testing.T) {
	err1 := fmt.Errorf("%w hello", errEmptyOrderIDs)
	if err1 != errEmptyOrderIDs {
		t.Fatal("fail")
	}
}
```

```go
func TestWrapError(t *testing.T) {
	err1 := fmt.Errorf("%w hello", errEmptyOrderIDs)
	if !errors.Is(err1, errEmptyOrderIDs) {
		t.Fatal("fail")
	}
}
```

engine/order_manager_test.go
```diff
@@ -27,6 +30,17 @@ type omfExchange struct {
...
+
+func TestMain(m *testing.M) {
+	var err error
+	btcusdPair, err = currency.NewPairFromString("BTCUSD")
+	if err != nil {
+		log.Fatal(err)
+	}
+	os.Exit(m.Run())
+}
+
```
gloriousCode, line 43:
```suggestion
var btcusdPair = currency.NewPair(currency.BTC, currency.USD)

```

engine/order_manager_test.go
```diff
@@ -645,6 +649,37 @@ func TestSubmit(t *testing.T) {
...
+	submitResp, err := submitReq.DeriveSubmitResponse("batman.obvs")
+	assert.Nil(t, err, "Deriving a SubmitResp should not error")
+
+	id, err := uuid.NewV4()
+	assert.Nil(t, err, "uuid should not error")
+	d, err := submitResp.DeriveDetail(id)
+	assert.Nil(t, err, "Derive Detail should not error")
+
+	d.ClientOrderID = "SecretSquirrelSauce"
+	assert.Nil(t, m.orderStore.add(d), "Adding an order should not error")
```
gloriousCode, line 674:
```suggestion
        err = m.orderStore.add(d)
	assert.Nil(t, err, "Adding an order should not error")
```
I find this more readable

engine/order_manager.go
```diff
@@ -1046,18 +1048,24 @@ func (s *store) upsert(od *order.Detail) (*OrderUpsertResponse, error) {
 
 // exists verifies if the orderstore contains the provided order
 func (s *store) exists(det *order.Detail) bool {
+	return s.getByDetail(det) != nil
+}
+
+// getByDetail fetches an order from the store and returns it
+// returns nil if not found
+func (s *store) getByDetail(det *order.Detail) *order.Detail {
```
gloriousCode, line 1056: Please add dedicated test to this function

gloriousCode (changes requested):
Thanks for this. The race was an annoyance

engine/order_manager.go
gbjk: Yep, completely agree

engine/order_manager_test.go
gbjk: Thanks. Will DriveBy this somewhere else too then.

engine/order_manager_test.go
gbjk: Thanks. I flipflopped on this so happy to take a subjective view.

engine/order_manager.go
gbjk: Done.

engine/order_manager_test.go
```diff
@@ -1658,3 +1683,24 @@ func TestProcessFuturesPositions(t *testing.T) {
...
+		ClientOrderID: "DuskyLeafMonkey",
+	}
+	concise := &order.Detail{
+		Exchange: od.Exchange,
+		OrderID:  od.OrderID,
+	}
+	assert.Nil(t, m.orderStore.getByDetail(od), "Fetching a non-stored order should return nil")
+	assert.Nil(t, m.orderStore.add(od), "Adding the details should not error")
+	assert.Same(t, m.orderStore.getByDetail(od), od, "Retrieve by full details should return the same pointer")
+	assert.Same(t, m.orderStore.getByDetail(concise), od, "Retrieve by other details should return the original pointer")
```
shazbert, line 1705: Suggestion: I prefer not to expose the pointer in the exchange order list. Just update the original allocated type. Maybe you have a better idea. 🤷 

```diff
diff --git a/engine/order_manager.go b/engine/order_manager.go
index 84fb9266a..b4649f2a3 100644
--- a/engine/order_manager.go
+++ b/engine/order_manager.go
@@ -1062,7 +1062,7 @@ func (s *store) getByDetail(det *order.Detail) *order.Detail {
        exchangeOrders := s.Orders[strings.ToLower(det.Exchange)]
        for _, o := range exchangeOrders {
                if o.OrderID == det.OrderID {
-                       return o
+                       *det = *o
                }
        }
        return nil
```

shazbert (commented):
Looks good to me. Just one potential suggestion for you to look over to consider.

engine/order_manager_test.go
shazbert: You would need to return the det p btw whoopsies.

gbjk: It was intentional, yes.
1) It's more expensive to return copies.
2) For most of my use cases, I want the order to be updated as changes happen to it.
If I don't return a pointer, I'll become incredibly inefficient constantly regrabbing the order all the time.

TL;DR: This actually provides functionality I struggled to live without.

gloriousCode: I agree with Shazbert. Returning `true` or `false` previously already did the job of seeing if an order exists.

I'm thinking more work would need to be done before we allow the direct order to be returned in the manner that you describe. The order pointer is protected down at that level with mutexes (which don't exist on the order.Detail itself) and also prevents meddling which is why we chose to do all the copying. Having the ability for an order to be modified without knowing the source of the modification is a concern, lowers trust and could lead to some incorrect conclusions being made.

The work done with say the `orderbook.Depth` is a good example of being able to hold an orderbook without being able to harm the underlying linked list and keep getting updates.

gbjk: Understood guys; I'll turn it into a value copy return for now then.

gbjk: @shazbert Strong disagreement about method of gutting and thralling the pointer that was passed in, but I think that's immaterial.
I can achieve what you've asked for 👍

gbjk: Tests will explicitly state behaviour for this

gbjk:
@gloriousCode Changes implemented.
I don't seem to have a "re-request review" available, but if you could do the honours please 🙏

engine/order_manager.go
```diff
@@ -1070,7 +1078,7 @@ func (s *store) add(det *order.Detail) error {
 	if err != nil {
 		return err
 	}
-	if s.exists(det) { // TODO: Error on conflict; remove unnecessary locking.
+	if s.exists(det) {
```
shazbert, line 1081: Maybe reintroduce the `TODO: remove unnecessary RW locking.` There is a duplication potential just by looking at the code, if this function is called by different routines for the same order ID which is unlikely because the function is not exported but it highlights it might need a possible redesign.

shazbert (changes requested):
Looks good just one pedantic request. Thanks! 🚀

engine/order_manager.go
gbjk: Okay, so to be clear:
We're about to Write lock it anyway, therefore you'd like to optimise not RLocking inside exists/getByDetails.
So you'd like to leave a TODO for that later.
I considered this. I'm pretty sure it's a premature optimisation, since add() is a write method.
I don't like TODOs that aren't coupled with actionable task numbers. They nearly never get done.
Suppose we did open an issue for this. I can't see it ever getting to the top of anyone's list.

TL;DR: It was a considered removal of the TODO because it's non-trivial, and premopt.

I'll readd it if you still want, though 😄

shazbert: It's not an optimisation per se it's more of a protection, when a RWMutex is not write locked the calling function doesn't semaquire. So just from a quick observation I may be wrong but concurrently if you call exists (read lock) and two routines for example could slip through and call write lock `s.m.Lock()` below which will act as a turnstile gate thus potentially producing a doubling of append calls. 

You are correct TODOs are not the best. 

Option 1) Re add TODO or Option 2) Fix the potential issue while you are there, which is remove the ReadLock and Unlock calls from `exists` and insert them into the scope of the calling function.

gbjk: Gotcha. You're right, I wasn't understanding your concern, which is definitely valid.

Whenever I feel I'm being annoyingly verbose it pays off.

Will Make-It-So.

gbjk: Tried Option 2 but it tasted awful because we're not *in* the store when we'd be fondling it's mutex.
I tried to live with it but I ended up rocking back and forth on the floor of the shower.
Instead I've opted to internalise it.
Is ee25cb74 acceptable?

shazbert: Yeah that is a way better approach thanks. 🚀  Thinking about it last night I don't think we should be having a RWmutex anyway, it puts a lot of burden on the writers. I will mull this over.

shazbert (approved):
Thanks for the changes. tack.

engine/order_manager.go
```diff
@@ -589,6 +589,14 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 		return nil, err
 	}
 
+	if err := m.orderStore.add(detail.CopyToPointer()); errors.Is(err, ErrOrdersAlreadyExists) {
+		// Streamed by ws before we got here. Details from ws supecede since they are more recent.
+		detail = m.orderStore.getByDetail(detail)
```
gloriousCode, line 594: This is mostly a comment on the "superseded" part and I'm not requesting a change.
This is a little risky, but for the most part its true. For ZB and Kraken however, we actually lose detail by discarding the REST response. However, I will rely on `processOrders` to ensure orders are up to date and not request a change.

engine/order_manager_test.go
```diff
@@ -645,6 +638,38 @@ func TestSubmit(t *testing.T) {
 	}
 }
 
+// TestSubmitOrderAlreadyInStore ensures that if an order is submitted, but the WS sees the conf before processSubmittedOrder
+// then we don't error that it was there already, and we merge together the details
```
gloriousCode, line 642: You're not really merging together details, like I said above, you're discarding them and this test can demonstrate that (by not reusing the same struct which wouldn't occur with a REST order and WS update). Please update comment

engine/order_manager.go
```diff
@@ -589,6 +589,14 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
...
+	} else if err != nil {
+		// Non-fatal error. OrderStore doesn't mean it wasn't submitted and consumers shouldn't get an error
```
gloriousCode, line 596: > OrderStore doesn't mean it wasn't submitted

? I'm fine with the rest, but that isn't really clear to me

gloriousCode (changes requested):
Just minor commentary. Thank you for updating things

engine/order_manager.go
gbjk: I see you've learnt not to bait me with challenges :joy:

engine/order_manager.go
gbjk: I had several cases where Submitting an order through OrderManager resulted in a an error storing it, and that was all we got back.
The actual order was made. The response details had order details, including an orderID.
But instead we just returned nil and a formatted error and the consumer wasn't able to programatically recover, or anything.

The point here is to say that when orderManager Submits an order, storing it is secondary to success, and it will not error on failure to store.
The alternative would be to return the detail + the err, but it would be non-idiomatic Go to expect the consumer to catch err != nil and yet still use returned detail at all.

I tried both for a while and landed on this.

engine/order_manager_test.go
gbjk: Agreed and Done: 69043784

I've simply removed the tail of this sentence.
I haven't gone so far as to add a failure test to *prove* we lose the detalis, since that's not a feature we want to protect against regressing.
If someone fixes that, AND the ClientOrderID is still correct (which _is_ tested) then we're happy.

engine/order_manager.go
gloriousCode: What I was getting at was the comment wasn't clear 😄 Yes it can be strange to submit an order and get an error back about how it couldn't be stored which has little to do with the order itself, so logging it is preferable and can be tracked and sorted later.

Perhaps Something like:
`// Non-fatal error: Unable to store order, but error does not need to be returned to caller`

gbjk: Agreed. Done. 7f472996

gloriousCode (approved):
tACK!

shazbert (approved):
Changes ACK.

engine/order_manager.go
```diff
@@ -589,6 +589,14 @@ func (m *OrderManager) processSubmittedOrder(newOrderResp *order.SubmitResponse)
 		return nil, err
 	}
 
+	if err := m.orderStore.add(detail.CopyToPointer()); errors.Is(err, ErrOrdersAlreadyExists) {
+		// Streamed by ws before we got here. Details from ws supersede since they are more recent.
+		detail = m.orderStore.getByDetail(detail)
+	} else if err != nil {
+		// Non-fatal error: Unable to store order, but error does not need to be returned to caller
+		log.Errorf(log.OrderMgr, "unable to add %v order %v to orderStore: %s", detail.Exchange, detail.OrderID, err)
```
thrasher-, line 597:
```suggestion
		log.Errorf(log.OrderMgr, "Unable to add %v order %v to orderStore: %s", detail.Exchange, detail.OrderID, err)
```

Pedantic nit, to be consistent with all other log messages now that this isn't returned as an error

gbjk: Thank you. Fixed [c8235a5](https://github.com/thrasher-corp/gocryptotrader/pull/1336/commits/c8235a56b7223f0d3020daf1e7f84c2cdd95a663)

thrasher- (approved):
Thanks for making that change!

PR #1337, opened 2023-09-07
Engine: Fix panic on OrderManager configured off

gbjk:
It's okay for OrderManager to not be enabled; It's configurable.
Remove the WSM setup protection and move it to runtime using IsRunning.

Left the WSM deps so they can be fixed as apporpriate to each.

TestWebsocketRoutineManagerSetup tests that passing a nil value returns errNilOrderManager;
However that's not actually what would happen when order manager is configured off.

The arguments to setupWebsocketRoutineManager are interface types.
When a nil pointer of interface type is passed, it does NOT equal nil.
nil is a primitive type.
A nil pointer of interface type has the value (nil;type).
See [the Go lang spec](https://golang.org/ref/spec#Comparison_operators):
```
Interface values are comparable. Two interface values are equal if they have identical dynamic types and equal dynamic values or if both have value nil.
```

So that means that whilst this test was passing, because it was sending
in a real nil value and comparing it to a real nil, that's not what
would happen at runtime. At runtime the bot.OrderManager would be a nil
pointer to a concrete type *OrderManager, and so not comparible to nil.

The first commit just fixes that oversight, and explains the often
misunderstood mechanics of comparing interface types to nil.

In practical terms this means that the tests assert that the WSRM would
not run without a OM, but in fact it actually would. And panic later.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestWebsocketRoutineManagerSetup


engine/websocketroutine_manager.go
```diff
@@ -287,6 +284,9 @@ func (m *WebsocketRoutineManager) websocketDataHandler(exchName string, data int
 		}
 		m.syncer.PrintOrderbookSummary(base, "websocket", nil)
 	case *order.Detail:
+		if !m.orderManager.IsRunning() {
```
gloriousCode, line 287: I think you should add this to `case []order.Detail` below

gloriousCode (changes requested):
Thanks for the detailed PR description!

engine/websocketroutine_manager.go
gbjk: Agreed and fixed!

shazbert (approved):
Tested ACK. Thanks.

engine/websocketroutine_manager.go
```diff
@@ -287,6 +284,9 @@ func (m *WebsocketRoutineManager) websocketDataHandler(exchName string, data int
...
+			return nil
+		}
 		if !m.orderManager.Exists(d) {
 			err := m.orderManager.Add(d)
 			if err != nil {
```
gloriousCode, line 292: While I understand what you're saying regarding the `nil` and the interface side of things, I'm unable to actually get a panic. Are you able to expand on that?
I'm running `--ordermanager=false`, it is not setup, and it is able to correctly fail `!m.orderManager.Exists(d)` on account of `m == nil`. Then after `err := m.orderManager.Add(d)` I receive the error 
image
which is only hit from
```go
func (m *OrderManager) Add(o *order.Detail) error {
	if m == nil {
		return fmt.Errorf("order manager %w", ErrNilSubsystem)
	}
```

I'm happy with the changes, but I want to ensure there is no panic remaining, but don't experience a panic on `master` myself

gbjk: I'll have to hunt back; this work was actually a few months ago originally.
It's not impossible I'm wrong it was panicing, and it was just throwing this error. Let me check.

gbjk: Yep; I can't figure it out either. I had a note that it panics.
All public methods inside it protect against nil. I can't see why or how it'd panic.

Ironically I broke one of my commit message rules.
Instead of the message saying what the bug would look like and how to reproduce, I focused on the intricasies of nil interface arguments.

As you say, though, this is valid either way, even if it doesn't panic.

gloriousCode (approved):
Thank you for verifying that no panic is reproducible, I'm happy with the changes. tACK!

thrasher- (approved):
tACK, thanks for the explanation + changes

thrasher-:
🚀 Happy 1337 PR 🎉

PR #1354, opened 2023-09-21
orderbook: Add optional orderbook.Item string fields for potential checksum calculations

shazbert:
# PR Description

* Adds addition fields for saving of raw strings which don't lose precision. 
* This optimises the checksum validation path, less stuff to do. 

NOTE: Can push this out to other exchanges if this is tested to be true and correct. Scoped it to only Kraken for now. 

Fixes # (issue)

* This attempts to fix a bug for luna pairs due to checksum failing intermittently. Storing "XXXXXXX.79999999" using function `strconv.ParseFloat("XXXXXXX.79999999", 64)` for example sometimes creates a float64 that automatically rounds `XXXXXXX.8`. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/kraken/kraken_websocket.go
```diff
@@ -941,8 +943,10 @@ func (k *Kraken) wsProcessOrderBookPartial(channelData *WebsocketChannelData, as
 		}
 
 		base.Bids[i] = orderbook.Item{
-			Amount: amount,
-			Price:  price,
+			Amount:    amount,
+			StrAmount: bids[1].(string),
+			Price:     price,
+			StrPrice:  bids[0].(string),
```
gloriousCode, line 949: An acceptable spot for nolint given that you wouldn't get here due to the above `ParseFloats`

gloriousCode (changes requested):
You'll need to address lints. Also, Exchange_Wrapper_standards has died here. Normally ~400 seconds in github actions, died because > 10 minutes, the issue might be the `hugeparam` related as the slowdown 🤷  or just a once off.

exchanges/kraken/kraken_websocket.go
```diff
@@ -927,22 +941,36 @@ func (k *Kraken) wsProcessOrderBookPartial(channelData *WebsocketChannelData, as
...
+		timeData, err := strconv.ParseFloat(tdStr, 64)
 		if err != nil {
 			return err
 		}
 
 		base.Bids[i] = orderbook.Item{
-			Amount: amount,
-			Price:  price,
+			Amount:    amount,
+			StrAmount: amountStr,
```
gloriousCode, line 971: The only thing I can think of is that perhaps it would be good to have an error - likely at LoadSnapshot()/Update() level - to comment specifically on if the Str versions are populated and the Float versions are not. We still require the float versions to process things and output certain results. I suppose you could do the conversions further down the line if float not populated, but I think that becomes complicated.

gloriousCode (changes requested):
Fix test ` depth_test.go:114: struct value options not set for field checksumStringRequired`,and then I think this is good to go. I think its the simplest solution to such lameness.

Could also merge master again to reduce _some_ test failures

gloriousCode (approved):

thrasher-:
It indeed is true and correct: https://go.dev/play/p/WcK8G3Q_gFq

This offers a nice optional solution for this issue and allows a user to use the string representation as they please. Ultimate solution would be an int/decimal handler but this is good enough for now

thrasher- (approved):
Nice stuff @shazbert

PR #1373, opened 2023-10-16
websocket: configurable subscriptions concept

gloriousCode:
# PR Description
This is a design for having customisable websocket subscriptions. Its only implemented on Binance and Kucoin and I want feedback on it before I do it for all exchanges. I think the two exchanges show the flexibility of the design, they are applied quite differently

If this design is fine, I'll be adding
- Readme updates (some channels like ` "/market/ticker:%s"` might need explaining)
- Testing
- All exchange support
- If necessary for external users, a way to set subs on setup without a config

Another aspect of this is that it is trying to change things minimally. There are certainly extra features I'd desire, but I feel like the important part of this is to _stop subscribing to candles_ because I don't want to

## Type of change

### DIFF
--- a/config/config_types.go
+++ b/config/config_types.go
@@ -184,6 +184,7 @@ type Exchange struct {
 	WebsocketResponseCheckTimeout time.Duration          `json:"websocketResponseCheckTimeout"`
 	WebsocketResponseMaxLimit     time.Duration          `json:"websocketResponseMaxLimit"`
 	WebsocketTrafficTimeout       time.Duration          `json:"websocketTrafficTimeout"`
+	WebsocketSubscriptions        WebsocketSubscriptions `json:"websocketSubscriptions"`
 	ConnectionMonitorDelay        time.Duration          `json:"connectionMonitorDelay"`
 	ProxyAddress                  string                 `json:"proxyAddress,omitempty"`
 	BaseCurrencies                currency.Currencies    `json:"baseCurrencies"`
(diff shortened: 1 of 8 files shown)


config/config_types.go
```diff
@@ -214,6 +215,12 @@ type Exchange struct {
 	WebsocketURL                     *string              `json:"websocketUrl,omitempty"`
 }
 
+// WebsocketSubscriptions holds the websocket subscriptions for an exchange
+type WebsocketSubscriptions struct {
+	Authenticated   map[string][]string `json:"authenticatedSubscriptions"`
```
shazbert, line 220: Would be cool to have a struct with enabled bool as well so we can on off whatever is supported.

shazbert (commented):
Good first start, only one thing I could think of. I did try and generalise it but its more of an inconvenience  
image


Also with function `GenerateSubscriptions()` we could have params public, private map[asset.Item][]string which the handling of what's working can be abstracted away from the wrapper implementations, reduce code and then you can define it externally, then in the wrapper you just range over the incoming items. But I am speculating 🤷

config/config_types.go
gloriousCode: Yeah great stuff. I like your ideas and they _shouldn't_ blow out too much more than this. I didn't want to try and do a catch all, because I'd rather improvements happen when needed and this not taking too long. 

Will look at the enabled/disabled style with predefined toggles with maybe a string slice for ones outside of that.

 I'll see how I go before looking at wrapper funcs though to minimise the amount of changes

gloriousCode:
I'm going to close this one. While I knew GK was going to work on this same concept, I  wanted a quick one as a transition because I kept running into needing to change `defaultSubscriptions`. The suggestions here put it a lot more in the direction of where G K is already going and is not worth continuing with due to double the amount of people working on the same concept

PR #1379, opened 2023-10-23
requester: purge max request jobs

shazbert:
# PR Description

* Removes friction point while tracking how many HTTP outbound jobs are run in parallel. This is generally not needed because of the rate limiting systems in place and allows you to spawn pending routines which will queue and exceed this amount especially in the confines of rolling window data on start-up (e.g. fee synchronisation).

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/bybit/bybit_live_test.go
```diff
@@ -11,7 +11,6 @@ import (
 	"testing"
```
gloriousCode, line 11: Please remove in mock test

exchanges/binance/binance_mock_test.go
```diff
@@ -12,7 +12,6 @@ import (
```
gloriousCode, line 12: Please remove in live test

gloriousCode (changes requested):
I think it is reasonable for a user to be able to set their own limits/manage request frequency. I don't think the current system of throwing away the request is the best way of handling it. I can see room for improvement for such a system to manage burstiness, but if that ever happens - PRs welcome
So in that sense, I'm comfortable with the removal of this system

gloriousCode (approved):
tACK m8

thrasher- (approved):
tACK. This was originally intended to cap a person from spamming thousands of go routines and subsequent requests by accident but most of the important exchanges support a rate limiter now. I'd still say we should have some kind of limit on the ones which don't but can be addressed in a separate PR.

PR #1395, opened 2023-11-10
Engine: Remove Default Forex Provider

gbjk:
# PR Description

Previously `ExchangeRateHost` would be the default provider if no other porvider was enabled and configured.
It is no longer allowing free access without keys, and after chasing a viable "free default" a few times, we will now default to no forex unless configured.

Also fixes display of erroring rates to avoid showing `0.00`, instead now showing `?.??` to distinguish small conversions from no conversion performed.

Fixes [Trello 206: Make engine's currency converter optional and upgrade exchangerate.host](https://trello.com/c/TLftCdB3/206-make-engines-currency-converter-optional-and-upgrade-exchangeratehost)

## Type of change

- [x] Bug fix
- [x] Behaviour change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


currency/storage.go
```diff
@@ -571,6 +571,9 @@ func (s *Storage) GetExchangeRates() (Conversions, error) {
 func (s *Storage) SeedForeignExchangeRates() error {
 	s.fxRates.mtx.Lock()
 	defer s.fxRates.mtx.Unlock()
+	if s.fiatExchangeMarkets == nil {
```
gloriousCode, line 574: Same here, it is a problem to want to seed rates with no market to seed from

currency/storage.go
```diff
@@ -167,22 +171,9 @@ func (s *Storage) RunUpdater(overrides BotOverrides, settings *Config, filePath
...
-			log.Warnf(log.Currency, "No valid foreign exchange providers configured. Defaulting to %s.", DefaultForexProviderExchangeRatesAPI)
-			fxSettings = append(fxSettings, base.Settings(settings.ForexProviders[x]))
-		}
-	}
-
-	if len(fxSettings) == 0 {
+		log.Warnln(log.Currency, "No foreign exchange providers enabled, currency conversion will not be available")
 		s.mtx.Unlock()
-		return errNoForeignExchangeProvidersEnabled
+		return nil
```
gloriousCode, line 176: I'm happy with this being nil, I'm only putting this here because I believe it should error elsewhere

currency/storage_test.go
```diff
@@ -30,219 +30,92 @@ func TestMain(m *testing.M) {
...
+	// Exchanges which do not error with a bad APIKey on startup
+	for _, n := range []string{"OpenExchangeRates", "ExchangeRateHost"} {
+		c.ForexProviders = AllFXSettings{{Name: n, Enabled: true, APIKey: ""}}
+		err = newStorage.RunUpdater(overideForProvider(n), c, tempDir)
+		assert.NoErrorf(t, err, "%s should not error and silently exit without running with no api keys", n)
+		assert.Nil(t, newStorage.fiatExchangeMarkets, "Forex should not be enabled with no providers")
+		c.ForexProviders = AllFXSettings{{Name: n, Enabled: true, APIKey: "sudo shazam!"}}
+		err = newStorage.RunUpdater(overideForProvider(n), c, tempDir)
+		assert.NoError(t, err, "%s should not error on Setup with a bad apikey", n)
+		assert.NotNil(t, newStorage.fiatExchangeMarkets, "Forex should not be enabled with no providers")
```
gloriousCode, line 85: image
Inaccurate commentary? Or more specific check required?

currency/storage_test.go
```diff
@@ -30,219 +30,92 @@ func TestMain(m *testing.M) {
...
+		{Name: "ExchangeRates"}, // Old Default
+		{Name: "OpenExchangeRates", APIKey: "shazam?"},
+		{Name: "ExchangeRateHost"}, // Old Default
+	}
+
+	// Regression test for old defaults which were enabled when in settings and nothing else was enabled and configured
+	err = newStorage.RunUpdater(BotOverrides{}, c, tempDir)
+	assert.NoError(t, err, "RunUpdater should not error")
+	assert.Nil(t, newStorage.fiatExchangeMarkets, "Forex should not be enabled with no providers") // Proxy for testing ForexEnabled
+	assert.False(t, c.ForexProviders[0].Enabled, "Old Default ExchangeRates should not have defualted to enabled with no enabled overrides")
```
gloriousCode, line 100:
```suggestion
	assert.False(t, c.ForexProviders[0].Enabled, "Old Default ExchangeRates should not have defaulted to enabled with no enabled overrides")
```

gloriousCode (changes requested):
I'm currently struggling to actually get any forex data from multiple providers 😭 So here are code comments and I'll see how things go once I can get a forex provider working too

currency/storage.go
```diff
@@ -74,14 +73,21 @@ func (s *Storage) SetDefaults() {
...
+func ForexEnabled() bool {
+	return storage.fiatExchangeMarkets != nil
 }
 
 // RunUpdater runs the foreign exchange updater service. This will set up a JSON
 // dump file and keep foreign exchange rates updated as fast as possible without
 // triggering rate limiters, it will also run a full cryptocurrency check
 // through coin market cap and expose analytics for exchange services
 func (s *Storage) RunUpdater(overrides BotOverrides, settings *Config, filePath string) error {
+	s.fiatExchangeMarkets = nil
```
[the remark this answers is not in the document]
gbjk: Yes, essentially, the reason for it is because of tests *and* because RunUpdater is _currently_ only run once, but I don't want to introduce a beartrap for someone later who might have reason to call it twice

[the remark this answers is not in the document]
gbjk: Seeing what you mean about the lock, I agree. But I also don't want to have all those return errors ignore an already set and now should-be-invaild provider.

As an experiment, how about this: 71a7eedf

The overlap of lock disenganging and reengaging going into ForeignExchangeUpdater remedies my discomfort about that tightly coupled lock dependency anyway.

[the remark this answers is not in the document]
gbjk: Yeah, okay. That won't fly.
Manifestly Raceable with the lock getting stolen before the goro runs.

I guess this, then: 830b301

It has the downside of skipping the errors, but I guess that's the lesser of 2 wevils 🐁

currency/storage.go
```diff
@@ -546,6 +543,9 @@ func (s *Storage) GetDefaultForeignExchangeRates() (Conversions, error) {
 func (s *Storage) SeedDefaultForeignExchangeRates() error {
 	s.fxRates.mtx.Lock()
 	defer s.fxRates.mtx.Unlock()
+	if s.fiatExchangeMarkets == nil {
+		return nil
```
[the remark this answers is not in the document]
gbjk: Agreed.
Fixed in ae48ee8

currency/storage.go
gbjk: Agreed.
Fixed in ae48ee8

currency/storage_test.go
gbjk: Inaccurate Human. Bad Human.
Thanks for spotting.

Fixed 8085342

currency/storage_test.go
gbjk: Thanks.
Fixed 9374f9b

gbjk:
@gloriousCode I know you'd started review of this, but I've added mock provider on top of it now, just in one commit:
da0320e9

gbjk:
@gloriousCode Ready for review

currency/forexprovider/currencylayer/currencylayer_test.go
```diff
@@ -1,137 +1,129 @@
...
 
-// please set your API key here for due diligence testing NOTE be aware you will
-// minimize your API calls using this test.
-const (
-	APIkey   = ""
-	Apilevel = 0
+// Either set your API key here or in env var for integration testing
+var (
+	apiKey      = ""
+	apiKeyLevel = 0
```
gloriousCode, line 18: So many functions are behind a check of 
```go
	if c.APIKeyLvl >= AccountBasic {
		return 0, errors.New("insufficient API privileges, upgrade to basic to use this function")
	}
```
It was very confusing, but for anyone else that tests, leave this as `0` 😄

currency/storage_test.go
```diff
@@ -30,219 +32,88 @@ func TestMain(m *testing.M) {
...
+	err = newStorage.RunUpdater(BotOverrides{}, c, tempDir)
+	assert.NoError(t, err, "RunUpdater should not error")
+	assert.Nil(t, newStorage.fiatExchangeMarkets, "Forex should not be enabled with no providers") // Proxy for testing ForexEnabled
+	assert.False(t, c.ForexProviders[0].Enabled, "Old Default ExchangeRates should not have defaulted to enabled with no enabled overrides")
+}
 
-	if !mainConfig.ForexProviders[2].PrimaryProvider {
-		t.Fatal("actual default provider should be designated as primary provider")
-	}
+func overideForProvider(n string) BotOverrides {
```
gloriousCode, line 104:
```suggestion
func overrideForProvider(n string) BotOverrides {
```

gloriousCode (changes requested):
Lovely, thanks for making those changes!
```[INFO]	 | SYNC | 17/11/2023 15:49:37 | BTC Markets websocket BTC-AUD SPOT ORDERBOOK: Bids len: 983 Amount: 340407.074076 BTC. Total value: $7195586.81 USD ($11127399.35 AUD) Asks len: 819 Amount: 107.693626 BTC. Total value: $6188629.76 USD ($9570220.83 AUD)```

I question BTC Markets' orderbook right now, but the conversion is working well

File: common/common.go [hunk 9]
```diff
@@ -73,6 +73,8 @@ var (
 	// ErrNoResponse is returned when a response has no entries/is empty
 	// when one is expected
 	ErrNoResponse = errors.New("no response")
+	// ErrAPIKeyRequired is returned whenever any service requires an API key
+	ErrAPIKeyRequired = errors.New("API key is required")
```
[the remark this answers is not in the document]
gbjk: Thanks. Was made redundant when EXH got removed.
I'd still like to go on a mission to unify this across all uses, but that's for another day.

Fixed 53dbc38e

currency/storage.go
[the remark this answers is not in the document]
gbjk: 🤷  Yikes. Must have gone wrong in a merge because 830b301 clearly shows it removed 🤷 

Fixed (again) 6eab25c

currency/forexprovider/currencylayer/currencylayer_test.go
gbjk: Yeah. I'd actually like to remove APIKeyLvl across the board, because I'm not seeing anywhere we'd base a decision tree off it that would be much better than letting the user hit the API's own error.
That is: Yes, we can pre-empt the error, but unless we're going to save requests that the engine or syncer would make wastefully, then there more downside in trying to clientside their pricing plan.
As I've seen it so far, every feature locked in API levels beyond basic is not a feature we'd use *automatically*, and the user can manage their own API calls.
Anyway. Out of scope and not tackling.

currency/storage_test.go
gbjk: Thanks.

Fixed ddcadd5a

engine/sync_manager.go
```diff
@@ -697,12 +697,12 @@ func printCurrencyFormat(price float64, displayCurrency currency.Code) string {
...
+	var conv string
 	if origPrice > 0 {
-		var err error
-		conv, err = currency.ConvertFiat(origPrice, origCurrency, displayCurrency)
-		if err != nil {
-			log.Errorf(log.SyncMgr, "Failed to convert currency: %s", err)
+		if convFloat, err := currency.ConvertFiat(origPrice, origCurrency, displayCurrency); err != nil {
+			conv = "?.??"
+		} else {
+			conv = fmt.Sprintf("%.2f", convFloat)
```
shazbert, line 705: Amounts get dropped, this allows it to reinstate it. 
```diff
diff --git a/engine/sync_manager.go b/engine/sync_manager.go
index b8afaf602..22db70952 100644
--- a/engine/sync_manager.go
+++ b/engine/sync_manager.go
@@ -704,6 +704,8 @@ func printConvertCurrencyFormat(origPrice float64, origCurrency, displayCurrency
                } else {
                        conv = fmt.Sprintf("%.2f", convFloat)
                }
+       } else {
+               conv = "0.00"
        }

        displaySymbol, err := currency.GetSymbolByCurrencyName(displayCurrency)
```

shazbert (changes requested):
Looks good only one thing from me! 👍 x 1000

engine/sync_manager.go
gbjk: Yep. Thanks.

Fixed f22a191b

shazbert (approved):
Thanks for that, looks good. ACK

gloriousCode (approved):

currency/forexprovider/currencylayer/currencylayer_test.go
```diff
@@ -1,137 +1,129 @@
...
 func TestGetSupportedCurrencies(t *testing.T) {
-	err := setup()
-	if err != nil {
-		t.Fatal("CurrencyLayer GetSupportedCurrencies error", err)
+	t.Parallel()
+	skipIfNoAPIKey(t)
+	currs, err := c.GetSupportedCurrencies()
+	if assert.NoError(t, err, "GetSupportedCurrencies should not error") {
+		assert.Contains(t, currs, "AUD", "AUD is a valid currency")
+		assert.Contains(t, currs, "USD", "USD is a valid currency") // Might fail in the near future
```
thrasher-, line 62: Cheeky 😆

currency/mock_provider_test.go
```diff
@@ -0,0 +1,40 @@
...
+func (m *MockProvider) Setup(base.Settings) error { return nil }
+func (m *MockProvider) IsEnabled() bool           { return true }
+func (m *MockProvider) IsPrimaryProvider() bool   { return true }
+func (m *MockProvider) GetSupportedCurrencies() ([]string, error) {
+	return storage.defaultFiatCurrencies.Strings(), nil
+}
+func (m *MockProvider) GetRates(baseCurrency, symbols string) (map[string]float64, error) {
+	c := map[string]float64{}
+	for _, s := range strings.Split(symbols, ",") {
+		// The year is 2027; The USD is nearly worthless. The world reserve currency is eggs.
```
thrasher-, line 36: Even more cheeky 😆

thrasher- (approved):
tACK using config_example, valid provider and invalid api key for a provider; all looks good! Nice stuff

image

PR #1400, opened 2023-11-16
Engine: Fix panic when exchange errors

gbjk:
BTSE is currently failing startup.
This fix simply prevents that from panicing.

## Type of change

## How this has been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/engine/helpers_test.go
+++ b/engine/helpers_test.go
@@ -17,6 +17,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/common/convert"
 	"github.com/thrasher-corp/gocryptotrader/common/file"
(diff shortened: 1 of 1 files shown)


gloriousCode (changes requested):
I think I'd prefer a more thorough fix. BTSE actually cannot run at all because of this error, its not about this one helper test.

BTSE has added some jank surrounding the currency PIT. We build a currency using {{base}} {{quote}} and not from {{symbol}} which misses the issue and ends up creating a duplicate, preventing BTSE from being setup

[So I would propose this fix](https://gist.github.com/gloriousCode/4effc0594b1f0b465549b550ecd72ee6). Yes, its annoying to need to do specific code like this and you are free to improve it, but this will allow BTSE to actually start as an exchange. You can test this by trying to retrieve the ticker for PIT-USD vs M_PIT-USD and see that BTSE will return an error for PIT-USD (it will also 301 when you try to view it on the site)

With the diff applied, the helper test works, but I also don't have a problem with your fix, its just too narrow

gbjk:
Yep; I'm working on BTSE.

I still recommend we fix this issue separately/atomically.
We don't want it reverted or confused with the BTSE issue, and it's a standalone issue with the engine test.

gloriousCode (approved):
tACK! Big fan of not panicking

engine/helpers_test.go
```diff
@@ -1372,10 +1373,8 @@ func TestNewSupportedExchangeByName(t *testing.T) {
 func TestNewExchangeByNameWithDefaults(t *testing.T) {
 	t.Parallel()
 
-	_, err := NewExchangeByNameWithDefaults(context.Background(), "meow")
-	if !errors.Is(err, ErrExchangeNotFound) {
-		t.Fatalf("received: '%v' but expected: '%v'", err, ErrExchangeNotFound)
-	}
+	_, err := NewExchangeByNameWithDefaults(context.Background(), "moarunlikelymeow")
```
shazbert, line 1376: 🐈

shazbert (approved):
LGTM! ACK

thrasher- (approved):
Thanks @gbjk !

PR #1401, opened 2023-11-19
BTSE: Fix duplicate pair errors on Million pairs (M_*)

gbjk:
BTSE listed Pitbull token with two symbols:
PIT-USD and M_PIT-USD for millons of PIT / USD.

They listed these pairs for about a week before the pairs were removed without announcement.
Nevertheless, these fixes are still valid, since they protect us against several similar edge cases and bugs.

The native token is not tradable, so we ignore them and get a base of M_PIT because that's what later APIs will accept

* Improves test coverage for BTSE
* Adds common.ExcludeError()
* Adds tests for Multi-Error
* Abstracts Locked market / HasLiquidity
* Adds a skip for spot markets not open for trading, implementing the TODO regarding OTC
* Improves Update Ticker to process all viable tickers and collect errors
* UpdateTickers tests won't fail on locked markets

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


exchanges/btse/btse_test.go
```diff
@@ -977,52 +702,41 @@ func TestGetLatestFundingRates(t *testing.T) {
...
+}
+
+var updatePairsGuard sync.Once
+
+func updatePairsOnce(tb testing.TB) {
+	tb.Helper()
+	updatePairsGuard.Do(func() {
+		err := b.UpdateTradablePairs(context.Background(), true)
+		assert.NoError(tb, err, "UpdateTradablePairs should not error")
+	})
```
gbjk, line 741: I've left this in because I've used it for an improved test around using pairs from GetEnabledPairs() for ticker and orderbook tests to ensure futures pairs are useable directly.
I've moved that work out for now, but feel this is useful enough to remain lurking for when I pick that back up.

common/common.go
```diff
@@ -537,6 +537,25 @@ func (e *multiError) Is(incoming error) bool {
 	return false
 }
 
+// ExcludeError returns a new error without any errors.Is(err)
```
gloriousCode, line 540: > without any errors.Is(err)

I don't understand since you use it twice in this function

exchanges/btse/btse_test.go
gloriousCode: I like this a lot. The multiple calls to `UpdateTradablePairs` slows down testing

exchanges/btse/btse_test.go
```diff
@@ -597,95 +424,76 @@ func TestSubmitOrder(t *testing.T) {
...
 
 func TestCancelOrder(t *testing.T) {
 	t.Parallel()
 	sharedtestvalues.SkipTestIfCredentialsUnset(t, b, canManipulateRealOrders)
 
-	_, err := b.CancelExistingOrder(context.Background(), "", testSPOTPair, "")
-	if err != nil {
-		t.Fatal(err)
-	}
+	// TODO: Surely this cannot work
```
gloriousCode, line 459: For tests behind `SkipTestIfCredentialsUnset` with `canManipulateRealOrders`, we don't expect the test details to be accurate, the user should fill it themselves, if that's what you're `TODO`ing about. If that's the case, I think you should be removing the TODO

exchanges/btse/btse.go
```diff
@@ -615,3 +615,24 @@ func (b *BTSE) calculateTradingFee(ctx context.Context, feeBuilder *exchange.Fee
...
+		if s.Active && s.HasLiquidity() && s.IsMillions() {
+			pairs[strings.TrimPrefix(s.Symbol, "M_")] = struct{}{}
+		}
+	}
+	return pairs
+}
+
+// HasLiquidity returns if a market pair has a bid or ask != 0
+func (m *MarketPair) HasLiquidity() bool {
+	return m.LowestAsk != 0 && m.HighestBid != 0
```
gloriousCode, line 632: You say `or` in the comment but require both Ask and Bid to be populated

gloriousCode (changes requested):
Thanks for this! I like a lot of these improvements, just basic commentary from me

exchanges/btse/btse.go
```diff
@@ -615,3 +615,24 @@ func (b *BTSE) calculateTradingFee(ctx context.Context, feeBuilder *exchange.Fee
 func parseOrderTime(timeStr string) (time.Time, error) {
 	return time.Parse(time.DateTime, timeStr)
 }
+
+// MillionPairs returns a map of symbol names which have a IsMillion equivalent
+func (m *MarketSummary) MillionPairs() map[string]struct{} {
```
shazbert, line 620: Can drop pointer receiver

exchanges/btse/btse.go
```diff
@@ -615,3 +615,24 @@ func (b *BTSE) calculateTradingFee(ctx context.Context, feeBuilder *exchange.Fee
...
+	pairs := map[string]struct{}{}
```
shazbert, line 621: map[string]struct{}{} ->  map[string]bool 
 
 Instead of this: `if _, ok := mPairs[l.Symbol]; ok {` 

Can be: `if mPairs[l.Symbol] {`

shazbert (changes requested):
Good stuff! Just some minor things from me.

common/common_test.go
```diff
@@ -881,3 +882,27 @@ func TestGenerateRandomString(t *testing.T) {
 		t.Error("GenerateRandomString() unexpected test validation result")
 	}
 }
+
+func TestMultiError(t *testing.T) {
+	t.Parallel()
+
+	e1 := errors.New("Inconsistent gravity")
+	e2 := errors.New("Barely marginal interest in your story")
```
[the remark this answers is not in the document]
gbjk: Fixed 6340b4a7

exchanges/btse/btse_test.go
gbjk: Thanks for the clarification.
I afraid I don't agree that it's a reasonable expectation for contributors to change the order number, but even if it was, it's definitely not clear.
It's unusual to expect any manual changes to source code, including tests, to get things to work.
A .env file, a flag, or a env var are usual. I've never seen the likes of "fill in your creds / order number here" in Go code before.
In these cases, we should *make* the orders we expect to cancel, so that the test suites are self-fulfilling.
That's what the TODOS are for.
Now that you've clarified 459, I've change these TODO line to match: 3cda30fd

exchanges/btse/btse.go
gbjk: Yes, I know that 🙂 

Maybe you already understood the background here, but just in case:
There was a lot of discussion about this already in the Go community.
For existence only checks, it's lighter weight to use empty structs than bools and check existence.
( IIRC [5% faster and 10% lighter memory](https://itnext.io/set-in-go-map-bool-and-map-struct-performance-comparison-5315b4b107b) )
We're conveying the existence of a million pair, not the truthiness of it, and we don't populate keys which are false.
Leaving unfound keys to default to false as you've suggested _works_, but it's implicit instead of explicit.
Either implementation is reasonable or justifiable.

This is a hot pathway, so I chose to be lighter.

Anyway, I'll change it if you still want.

exchanges/btse/btse.go
gbjk: Can drop, yes, obviously 🙂 

However it's lighter to use the pointer here, right? ( By the 2 extra fields a slice by value contains )

What's the benefit of switching to pass-by-value here?

exchanges/btse/btse_test.go
gloriousCode: Your views aside, your original todo was not going to get merged into `master` as it was not helpful for contributors and was more akin to a personal note. I'm glad that the context I gave you allowed you to improve your commentary

gbjk: Yep; Sorry, These comments as they stood shouldn't have been committed at all. I should have been clearer about that in my reply.
Thanks for catching them 🙏

exchanges/btse/btse.go
gbjk: Thank you

Fixed e7e5b13f

exchanges/btse/btse.go
gbjk: Fixed e78a3b5d

common/common.go
gbjk: Is this clearer:
```
// ExcludeError returns a new error excluding any errors matching excl
// For a standard error it will either return the error unchanged or nil
// For multiError it will remove any errors matching excl and return the remaining errors or nil
```

Fixed 5770e95e

Please reopen if that didn't help. I might need some more help understanding what's not clear.

common/common.go
```diff
@@ -537,6 +537,25 @@ func (e *multiError) Is(incoming error) bool {
...
+// If there are no other errors it will return nil
+func ExcludeError(err, excl error) error {
+	mE, ok := err.(*multiError)
+	if !ok {
+		if errors.Is(err, excl) {
+			return nil
```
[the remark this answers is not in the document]
gbjk: Thanks.
I did have that thought about how Unwrap is _normally_ used and support for multi-wrapped errors which aren't our own.
I'm thinking that it *should* be viable to handle all errors which support Unwrap the same way.

To answer your question: Nope, I'm explicitly trying to find the remaining errors after acceptable errors are removed. This is a pattern I tend to use a lot, but it's implemented slightly differently in my own version of an error implementation.

I'll swing at it a bit and come back to you.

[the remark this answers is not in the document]
gbjk: I'm back from the rabbit hole 🐇 🕳️ 
Thanks for the prod to do better. This really needed a properly idiomatic solution consistent with the direction of the go standard library.

```
Common: Improve AppendError and ExcludeError

This change switches from a stateful multiError to caring more about the
Unwrap() []error interface, the same as [go standard
lib](https://github.com/golang/go/blob/go1.21.4/src/errors/wrap.go#L54-L68)

Notably, if we implement Unwrap() []error and do NOT implement Is() then
we get free compatibility with the core functions.

The only distateful thing here is needing to deeply unwrap fmt.Errorf
errors, since they don't flatten. I can't see any way around that
```

Fixed 32f57a0a

Sidenote: I'm left wondering why errors doesn't expose some sort of Unwrappable interface, since both errors and fmt seem to be sharing this inside ~joke~ behaviour 🤷 
Oh well. If it changes tests will tell us.

shazbert (changes requested):
Nice improvements 🚀, only thing left is when I run the instance it falls over on rest fetching and initial sync never completes.

```
[ERROR] | SYNC | 27/11/2023 10:01:20 | Failed to get REST ticker. Error: BTSE PIT-USD bid equals ask this is a crossed or locked market, BTSE PIT-USDC bid equals ask this is a crossed or locked market, BTSE PIT-USDT bid equals ask this is a crossed or locked market
```

gbjk:
Yeah. I'm not happy to see this every time:
```
[DEBUG] | EXCHANGE | 27/11/2023 10:49:47 | BTSE Updating available pairs [SPOT] - Added: [M_PIT-USD M_PIT-USDC M_PIT-USDT].
[DEBUG] | EXCHANGE | 27/11/2023 10:49:47 | BTSE Updating available pairs [SPOT] - Removed: [M_PIT-USD M_PIT-USDC M_PIT-USDT].
```

gbjk:
@shazbert Linter fixed

Just adding coverage for the config unmarshal and then see about fixing the locked market blocker to initial sync.

gbjk:
> Nice improvements 🚀, only thing left is when I run the instance it falls over on rest fetching and initial sync never completes.
> 
> ```
> [ERROR] | SYNC | 27/11/2023 10:01:20 | Failed to get REST ticker. Error: BTSE PIT-USD bid equals ask this is a crossed or locked market, BTSE PIT-USDC bid equals ask this is a crossed or locked market, BTSE PIT-USDT bid equals ask this is a crossed or locked market
> ```

Fixed b5bb64b8

gbjk:
@shazbert That's everything addressed I think.
Rebased, linter fixed, and locked markets isolated to the locked pairs.

The issues with locked markets disappeared, but I think the fix here is self-evident anyway.

currency/manager.go
```diff
@@ -458,6 +458,40 @@ func (p *PairsManager) getPairStoreRequiresLock(a asset.Item) (*PairStore, error
 	return pairStore, nil
 }
 
+// UnmarshalJSON implements json.Unmarshal
+// Pairs.Unmarshal doesn't know what the delimiter is, so uses the first punctuation rune
+// This method uses the configFormat delimiter to improve that guess
+func (p *PairsManager) UnmarshalJSON(d []byte) error {
```
shazbert, line 464: Can strip the methods by converting to alias straight away so it doesn't recursively call. Then we don't need to assign. Maybe add a comment as to why. 🤷 
```diff
diff --git a/currency/manager.go b/currency/manager.go
index 304c9de03..cf98db0c5 100644
--- a/currency/manager.go
+++ b/currency/manager.go
@@ -462,12 +462,10 @@ func (p *PairsManager) getPairStoreRequiresLock(a asset.Item) (*PairStore, error
 // Pairs.Unmarshal doesn't know what the delimiter is, so uses the first punctuation rune
 // This method uses the configFormat delimiter to improve that guess
 func (p *PairsManager) UnmarshalJSON(d []byte) error {
-       type Alias PairsManager
-       var t Alias
-       if err := json.Unmarshal(d, &t); err != nil {
+       type alias *PairsManager
+       if err := json.Unmarshal(d, alias(p)); err != nil {
                return err
        }
-       *p = PairsManager(t) //nolint:govet // Mutex copying of lexical private alias is okay
        for _, s := range p.Pairs {
                cf := s.ConfigFormat
                if cf == nil {
```

shazbert (changes requested):
Looks good just one thing! 🚀

currency/manager.go
gbjk: This is super cute; Thank you 🙏 
Pattern Squirrelled.

Fixed c6b57d02

shazbert (changes requested):
Thanks for the changes, on startup:

```
GoCryptoTrader v0.1 amd64 go1.21.1 pre-release.
This version is pre-release and is not intended to be used as a production ready trading framework or bot - use at your own risk.
Copyright (c) 2014-2023 The GoCryptoTrader Developers.

GitHub: https://github.com/thrasher-corp/gocryptotrader
Trello: https://trello.com/b/ZAhMhpOy/gocryptotrader
Slack:  https://join.slack.com/t/gocryptotrader/shared_invite/enQtNTQ5NDAxMjA2Mjc5LTc5ZDE1ZTNiOGM3ZGMyMmY1NTAxYWZhODE0MWM5N2JlZDk1NDU0YTViYzk4NTk3OTRiMDQzNGQ1YTc4YmRlMTk
Issues: https://github.com/thrasher-corp/gocryptotrader/issues

2023/12/04 10:34:12 Loading config file C:\Users\meow\AppData\Roaming\GoCryptoTrader\config.json..
2023/12/04 10:34:12 Unable to initialise bot engine. Error: failed to load config. Err: fatal error opening C:\Users\meow\AppData\Roaming\GoCryptoTrader\config.json file. Error: error reading config delimiter: [_] not found in currencypair string: BTC-USDT
```

gbjk:
@shazbert Yep. I found issues in the existing config_example.json and configtest.json as well.

Your issue is harder to find than those, since `BTC-USDT` is everywhere.

I've made the assumption that we should error on bad config.

If we want to give the user more hints about where the config issue is, we're going to need to add an UnmarshalJSON protection at the exchange level so we can tell them what exchange to look at.

I'll start to look at adding info so the error would bubble up as:
`error in exchange config for BitGremlin: error reading config delimiter: [_] not found in currencypair string: BTC-USDT in margin available pairs`

If you think we shouldn't error on the bad config, let me know.

gbjk:
> Now I am failing on Bybit. But nice changes. 😆

Thanks.
So your config for Bybit must be wrong somewhere, right.
Mine has delimiter "-" set for all asset types, and no overrides below:
```
    "configFormat": {
     "uppercase": true,
     "delimiter": "-"
    },
```
No sign of a `_` in my config for Bybit.
I presume your config *does* have a `_`, right?

You've requested changes with your last review, but it's not clear to me what change you'd like :)

shazbert (approved):
As it stands this is good to go.

currency/manager.go
```diff
@@ -458,6 +458,33 @@ func (p *PairsManager) getPairStoreRequiresLock(a asset.Item) (*PairStore, error
...
+				nP, err := NewPairDelimiter(p[j].String(), cf.Delimiter)
+				if err != nil {
+					return fmt.Errorf("%s.%s.%s: %w", a, []string{"enabled", "available"}[i], p[j], err)
+				}
+				p[j] = nP
+			}
+		}
+	}
+	return nil
+}
```
gloriousCode, line 486: This is nice to highlight things, but its placement, where an error leads to GCT being unable to start could be an issue such as the errors raised from `NewPairDelimiter`.
For example, ITBIT has an incorrect config format set in `itbit_wrapper.go:64` with the inclusion of a `DashDelimiter`. However, no currencies in the config contain one, which causes `NewPairDelimiter` to fail, and GCT unable to start. Removing it from the config will only re-add it on next startup, reestablishing the issue. To reproduce this yourself, enable ITBIT and run GCT

For this PR to be merged, you will need to ensure that all set config formats work for all exchanges and their assets or consider calling `SetDelimitersFromConfig` from a place which doesn't impact the entire startup of the application. If a user has an encrypted config, they will have to start again by running this because they can't modify their config formats to fix the fatal error on startup.

currency/manager.go
```diff
@@ -458,6 +458,33 @@ func (p *PairsManager) getPairStoreRequiresLock(a asset.Item) (*PairStore, error
...
+// SetDelimitersFromConfig ensures that the pairs adhere to the configured delimiters
+// Pairs.Unmarshal doesn't know what the delimiter is, so uses the first punctuation rune
+func (p *PairsManager) SetDelimitersFromConfig() error {
+	for a, s := range p.Pairs {
+		cf := s.ConfigFormat
+		if cf == nil {
+			cf = p.ConfigFormat
+		}
+		if cf == nil {
+			continue
```
gloriousCode, line 470: All exchanges have configs set either globally or at the asset level, `continue` and `return nil` seems a strange response to a missing field. It seems like a testing convenience when you are asking it to set something and it doesn't

exchanges/btse/btse_wrapper.go
```diff
@@ -350,7 +352,23 @@ func (b *BTSE) UpdateTicker(ctx context.Context, p currency.Pair, a asset.Item)
...
+			Ask:          ticks[0].LowestAsk,
+			Bid:          ticks[0].HighestBid,
+			Low:          ticks[0].Low24Hr,
+			Last:         ticks[0].Last,
+			Volume:       ticks[0].Volume,
+			High:         ticks[0].High24Hr,
+			ExchangeName: b.Name,
+			AssetType:    a})
+	}
+	if err != nil {
```
gloriousCode, line 371: I don't understand why checking the length of the response is a higher priority than checking the returned error,  dumping whatever BTSE has said about the situation.

common/common_test.go
```diff
@@ -881,3 +882,54 @@ func TestGenerateRandomString(t *testing.T) {
 		t.Error("GenerateRandomString() unexpected test validation result")
 	}
 }
+
```
gloriousCode, line 885: The test `TestErrors` now fails in this PR:
`    common_test.go:757: targets not achieved`

gloriousCode (changes requested):
Thanks for your Error improvements from last review!

currency/manager.go
gbjk: Would you suggest we return error when it's nil, then?

currency/manager.go
gbjk: Understood.

I considered whether this should be a fatal startup issue or not, and kept coming back to "It should be" because otherwise we have a fundamental config issue and it's too easily ignored on startup.

Yes, It was my intention to fix any issues arising from this in config_example.json and testconfig.json.

Regarding ITBIT: I didn't/don't get an issue, because config_example.json has:
```
    "configFormat": {
     "uppercase": true,
     "delimiter": "-"
    },
```
So the SetDefault version doesn't get a look-in.
Best shot I get is when I turn it's `useGlobalFormat` to false, and then I get:
`2023/12/12 15:30:18 Unable to initialise bot engine. Error: failed to load config. Err: fatal error checking config values. Error: no exchanges enabled`

Could you share the config I can use to replicate what you see?

In the meantime, I can also just fix the itbit_wrapper, right?

Your last comment is more worrying, I guess. I'm not sure how they'd go about fixing any other issues with config either, not having experimented with it.

I'm happy to switch the fatal error to a warning, which would mitigate the issues, but leave us more vulnerable.
And/Or We could disable the exchange like we do in other situations.

exchanges/btse/btse_wrapper.go
gbjk: Yeah. That's just wrong.
Suspect it got abstracted several times and then ended up like this.

Fixed

common/common_test.go
gbjk: Tried to work out how TestErrors ended up back in here. `git-reflog` and `git-stash` show no signs of the change.
I definitely gutted it and moved TestMultiError into there, so I must have not got as far as committing it for some reason.
Everything I moved into TestMultiError is definitely covering it all as I remember.

Oh well.

Fixed and pushed

currency/manager.go
gbjk: ITBit: I couldn't reproduce because this issue happens on the *second* run, after the config is updated.
I was running with -dryrun so never got the config update. 🤦 

Fixed: 45b9d362

gbjk: We've discussed this and we're agreed now that whilst erroring isn't great, it's the necessary evil here.
Resolving.

I'm going to ensure that Binance is okay with it's [special handling](https://github.com/thrasher-corp/gocryptotrader/blob/master/exchanges/binance/binance_wrapper.go#L313-L360)

currency/manager.go
gbjk: Due to `CheckPairConsistency` we should never get here.
```
Unable to initialise bot engine. Error: failed to load config. Err: fatal error checking config values. Error: no exchanges enabled
Exchange BTSE: CheckPairConsistency error: exchange BTSE pair config format for asset type spot is nil
```
So we can safely make this an error, and if anyone removes/breaks `CheckPairConsistency` then it'll help.

gloriousCode: Sounds good! Thanks for following that up 🕵️

gbjk: Fixed 8f7559457

gbjk:
Breaking out the review of #836 changes into a separate task.
Those codepaths become redundant, but I don't think we should attempt to bloat the scope of this PR any more than it's already pigletesque stature 🐷 

This PR addresses the breaking tests and that's it's focus.

shazbert (changes requested):
Thanks for the changes. There seems to be a pending [test failure](https://github.com/thrasher-corp/gocryptotrader/actions/runs/7190800911/job/19584504254?pr=1401#step:7:2792)
Then everything else looks good to go. Well done.

gbjk:
@shazbert Fixed

One option was to re-instate delimiter setting on initial json.Unmarshal again, but not error.
That'd mean that any time someone unmarshals but doesn't check it'd be okay.
But it's wasteful duplication, and it's just the test that was failing. So I've just called the delimiter setting logic in the test instead.

shazbert (approved):
tACK. Nice work.

gloriousCode (approved):
tACK! Thanks for making those changes!

thrasher- (approved):
tACK, nice improvements across the board

PR #1405, opened 2023-11-25
Okx: Fix GetInsuranceFundInformation "" resp 

gbjk:
When amt is "" GetInsuranceFundInformation was erroring.
amt is [documented](https://www.okx.com/docs-v5/en/#public-data-rest-api-get-insurance-fund) to return empty when type is `regular_update`

Added some more test coverage, impinged by the fields not actually being populated for currency.

Fixes [Trello 211](https://trello.com/c/gLjdz1Ky/211-okx-getinsurancefundinformation-test-failure)

Also quickly formats configtest.json with jq. Because this is a small and quick PR to get such a thing done in.
git diff -w or github ignore whitespaces should show no changes.
Any conflicts, just accept --ours and reformat it. Or not.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestGetInsuranceFundInformation


shazbert (approved):
Nice catch.

Suggestion: I think it would be good as a catch all for this issue is to convert this pattern to using `convert.StringToFloat64`.
```
FieldName float64          `json:"fieldTag,string"`
```

But doesn't need to be attempted in this PR.

thrasher- (approved):
Thanks @gbjk !

PR #1415, opened 2023-12-09
Types: Switch convert.StringToFloat64 to types.Number

gbjk:
This change mostly just renames the type.

convert package and StringToFloat64 represent actions, not types,
and make it misleading to use outside of the API context,
especially when using it for a Float64ToString operation.

This is intended as a first step towards unified type helpers appropriate across GCT in general, not just in exchange APIs.

## Type of change

- [x] Non-Breaking change (fix or feature that would cause existing functionality to not work as expected)

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/types/number.go
+++ b/types/number.go
@@ -1,0 +1,73 @@
+package types
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/shopspring/decimal"
+)
+
+var errInvalidNumberValue = errors.New("invalid value for Number type")
+
+// Number represents a floating point number, and implements json.Unmarshaller and json.Marshaller
+type Number float64
+
+// UnmarshalJSON implements json.Unmarshaler
+func (f *Number) UnmarshalJSON(data []byte) error {
+	switch c := data[0]; c { // From json.decode literalInterface
+	case 'n', 't', 'f': // null, true, false
+		return fmt.Errorf("%w: %s", errInvalidNumberValue, data)
+	case '"': // string
+		if len(data) < 2 || data[len(data)-1] != '"' {
+			return fmt.Errorf("%w: %s", errInvalidNumberValue, data)
+		}
+		data = data[1 : len(data)-1] // Naive Unquote
+	default: // Should be a number
+		if c != '-' && (c < '0' || c > '9') { // Invalid json syntax
+			return fmt.Errorf("%w: %s", errInvalidNumberValue, data)
+		}
+	}
+
+	if len(data) == 0 {
+		*f = Number(0)
+		return nil
+	}
+
+	val, err := strconv.ParseFloat(string(data), 64)
+	if err != nil {
+		return fmt.Errorf("%w: %s", errInvalidNumberValue, data) // We don't use err; We know it's not valid and errInvalidNumberValue is clearer
+	}
+
+	*f = Number(val)
+
+	return nil
+}
+
+// MarshalJSON implements json.Marshaler by formatting to a json string
+// 1337.37 will marshal to "1337.37"
+// 0 will marshal to an empty string: ""
+func (f Number) MarshalJSON() ([]byte, error) {
+	if f == 0 {
+		return []byte(`""`), nil
+	}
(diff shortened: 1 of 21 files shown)


types/number.go
```diff
@@ -0,0 +1,71 @@
...
+func (f Number) Float64() float64 {
+	return float64(f)
+}
+
+// Int64 returns the truncated integer component of the number
+func (f Number) Int64() int64 {
+	// It's likely this is sufficient, since Numbers probably have not had floating point math performed on them
+	// However if issues arise then we can switch to math.Round
+	return int64(f)
+}
```
shazbert, line 66: Suggestion: I would probably just RM this so we don't need to worry about this case. Clients can just type convert by int64(Number) easily enough if they need?

shazbert (changes requested):
This looks good, I like the code improvements. Kucoin test needs to be fixed and only one suggestion. Thanks heaps.

types/number.go
```diff
@@ -0,0 +1,71 @@
...
+func (f *Number) UnmarshalJSON(data []byte) error {
+	switch c := data[0]; c { // From json.decode literalInterface
+	case 'n', 't', 'f': // null, true, false
+		return fmt.Errorf("%w: %s", errInvalidNumberValue, data)
+	case '"': // string
+		if len(data) < 2 || data[len(data)-1] != '"' {
+			return fmt.Errorf("%w: %s", errInvalidNumberValue, data)
+		}
+		data = data[1 : len(data)-1] // Naive Unquote
+	default: // Should be a number
```
gloriousCode, line 26: Thanks for adding this, would have been my first comment 🎉

types/number.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return fmt.Errorf("%w: %s", errInvalidNumberValue, data) // We don't use err; We know it's not valid and errInvalidNumberValue is clearer
+	}
+
+	*f = Number(val)
+
+	return nil
+}
+
+// MarshalJSON implements json.Marshaler
```
gloriousCode, line 47: Could you please add here saying that it marshalls as a string? Its the only thing the `convert.StringToFloat` kinda sorta implied, but this loses. One might use a `type.Number` not realising its marshalled into `"1337"`

I'd also accept it being added  to the type declaration instead

gloriousCode (changes requested):
After lint and test fixes, just one thing from me I think

gbjk:
I'm breaking [2d7aac3](https://github.com/thrasher-corp/gocryptotrader/pull/1415/commits/2d7aac33bd8ade020778b547e7331f863e8f3e2c) out into a separate task, along with the other types.

I've already encountered several places where I'm reusing StringToFloat64 which will need this rebase if after.

Let's keep this on track and get it merged, then I'll address the others with less of a blunderbuss to my head. 🔫

types/number.go
gbjk: Intended for use in okcoin `MaxIcebergParts`
I'm keen to centralise the behaviour early so that if we do need to `math.Round` then it's easy and in one place.

TL;DR: Noted, but I'd like to keep it.

types/number.go
gbjk:
```
// MarshalJSON implements json.Marshaler by formatting to a json string
// 1337.37 will marshal to "1337.37"
// 0 will marshal to an empty string: ""
```
Fixed a7bd1dff

gloriousCode (approved):
Thanks for updating that comment. tACK!

shazbert (approved):
Changes look great! Thanks. tACK!.

gbjk:
Rebased and new StringToFloats changed over:
```
perl -pi -E 's{convert.StringToFloat64}{types.Number}g' exchanges/**/*.go
git status -s | sed 's/...//' | xargs goimports -w
```
There's a fixup commit which I'll rebase down now.

thrasher- (approved):
tACK 78612498509fd97ed88a0def2e974fae6e146aa3, nice!

PR #1424, opened 2023-12-16
Exchanges: Remove bespoke pair upgrade handling and abstract Start/Run

gbjk:
## Remove bespoke pair upgrade handling
Now redundant behind #1401. These paths should never be met.
Several legacy coin upgrade paths being deprecated as well: ZUSD and CNY
Expecting any users with bad config from 3+ years ago would have to reset anyway.

Also: At the time the intention of this was to upgrade the config format.
However now, instead, it'd mostly serve to reset enabled pairs if there's a config mistake, which doesn't feel right.

## Abstract Start and Run

Both Start and Run were pretty much cargo culted across the exchanges.
The bespoke pair upgrade handling masked this, as well as inconsistencies, but when whittled down the only time there was a difference it was a missing behaviour that should be added.

For clarity about this design pattern:
* If we had a `*Base.Run` then any methods called inside it would not call `*exchange.UpdateOrderExecutionLimits` but always only call `*Base.UpdateOrderExecutionLimits`. Because once Run has been dispatched to the embedded field Base, there's no way back to the containing struct.
* By calling functions which receive interface arguments, we get to dispatch first to `exchange.Method`, if it exists, otherwise `base.Method`
* Run is not currently a method, because we wanted to have the behaviour above ☝️ Any option that would end up calling methods on a `*Base` would lose the method inheritence
* We might consider renaming exchange.Run to `CommonBootstrap` and then calling `Bootstrap` on b to allow it to do anything else
* We could also consider a type switch on an interface, and calling the exchange `Bootstrap` if it can do it, otherwise calling `CommonBootstrap`

All of this is simply to say: We're not penned in or inflexible as a result of this abstraction.

## Type of change

- [ ] Non-Breaking change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run

### DIFF
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -1874,3 +1874,52 @@ func (b *Base) ParallelChanOp(channels []subscription.Subscription, m func([]sub
 
 	return errs
 }
+
+// Start starts an exchange bootstrap process in the background
+func Start(ctx context.Context, b IBotExchange, wg *sync.WaitGroup) error {
+	if wg == nil {
+		return fmt.Errorf("%T %w", wg, common.ErrNilPointer)
+	}
+	wg.Add(1)
+	go func() {
+		Run(ctx, b)
+		wg.Done()
+	}()
+	return nil
+}
+
+// Run performs a few Bootstrapping actions for the exchange
+func Run(ctx context.Context, b IBotExchange) {
+	if b.IsVerbose() {
+		if b.GetSupportedFeatures().Websocket {
+			wsURL := ""
+			wsEnabled := false
+			if w, err := b.GetWebsocket(); err == nil {
+				wsURL = w.GetWebsocketURL()
+				wsEnabled = w.IsEnabled()
+			}
+			log.Debugf(log.ExchangeSys, "%s Websocket: %s. (url: %s)", b.GetName(), common.IsEnabled(wsEnabled), wsURL)
+		} else {
+			log.Debugf(log.ExchangeSys, "%s Websocket: Unsupported", b.GetName())
+		}
+		b.PrintEnabledPairs()
+	}
+
+	a := b.GetAssetTypes(true)
+	for x := range a {
+		if err := b.UpdateOrderExecutionLimits(ctx, a[x]); err != nil && !errors.Is(err, common.ErrNotYetImplemented) {
+			log.Errorf(log.ExchangeSys, "%s failed to set exchange order execution limits. Err: %v", b.GetName(), err)
+		}
+	}
+
+	if b.GetEnabledFeatures().AutoPairUpdates {
+		if err := b.UpdateTradablePairs(ctx, false); err != nil {
+			log.Errorf(log.ExchangeSys, "%s failed to update tradable pairs. Err: %s", b.GetName(), err)
+		}
+	}
+}
+
+// IsVerbose returns if the exchange is set to verbose
+func (b *Base) IsVerbose() bool {
+	return b.Verbose
+}
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -46,8 +46,6 @@ const (
 	DefaultWebsocketResponseMaxLimit = time.Second * 7
 	// DefaultWebsocketOrderbookBufferLimit is the maximum number of orderbook updates that get stored before being applied
 	DefaultWebsocketOrderbookBufferLimit = 5
-	// ResetConfigPairsWarningMessage is displayed when a currency pair format in the config needs to be updated
-	ResetConfigPairsWarningMessage = "%s Enabled and available pairs for %s reset due to config upgrade, please enable the ones you would like to use again. Defaulting to %v"
 )
 
 var (
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -1901,28 +1901,24 @@ func Bootstrap(ctx context.Context, b IBotExchange) error {
 		b.PrintEnabledPairs()
 	}
 
-	errC := make(chan error, 16) // Blocking isn't a dealbreaker, but a little buffer is nice
-	var wg sync.WaitGroup
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		a := b.GetAssetTypes(true)
-		for x := range a {
-			if aErr := b.UpdateOrderExecutionLimits(ctx, a[x]); aErr != nil && !errors.Is(aErr, common.ErrNotYetImplemented) {
-				errC <- fmt.Errorf("failed to set exchange order execution limits: %w", aErr)
-			}
+	if b.GetEnabledFeatures().AutoPairUpdates {
+		if err := b.UpdateTradablePairs(ctx, false); err != nil {
+			return fmt.Errorf("failed to update tradable pairs: %w", err)
 		}
-	}()
+	}
 
-	wg.Add(1)
-	go func() {
-		defer wg.Done()
-		if b.GetEnabledFeatures().AutoPairUpdates {
-			if aErr := b.UpdateTradablePairs(ctx, false); aErr != nil {
-				errC <- fmt.Errorf("failed to update tradable pairs: %w", aErr)
+	a := b.GetAssetTypes(true)
+	var wg sync.WaitGroup
+	errC := make(chan error, len(a))
+	for i := range a {
+		wg.Add(1)
+		go func(a asset.Item) {
+			defer wg.Done()
+			if err := b.UpdateOrderExecutionLimits(ctx, a); err != nil && !errors.Is(err, common.ErrNotYetImplemented) {
+				errC <- fmt.Errorf("failed to set exchange order execution limits: %w", err)
 			}
-		}
-	}()
+		}(a[i])
+	}
 	wg.Wait()
 	close(errC)
 
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -1881,9 +1881,9 @@ func (b *Base) ParallelChanOp(channels []subscription.Subscription, m func([]sub
 // * Print debug startup information
 // * UpdateOrderExecutionLimits
 // * UpdateTradablePairs
-func Bootstrap(ctx context.Context, b IBotExchange) {
-	if continueBootstrap := b.Bootstrap(ctx); !continueBootstrap {
-		return
+func Bootstrap(ctx context.Context, b IBotExchange) error {
+	if continueBootstrap, err := b.Bootstrap(ctx); !continueBootstrap || err != nil {
+		return err
 	}
 
 	if b.IsVerbose() {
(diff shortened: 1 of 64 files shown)


exchanges/exchange.go
```diff
@@ -1752,3 +1751,48 @@ func (b *Base) MatchSymbolCheckEnabled(symbol string, a asset.Item, hasDelimiter
...
+	wg.Add(1)
+	go func() {
+		Run(ctx, b)
+		wg.Done()
+	}()
+	return nil
+}
+
+// Run performs a few Bootstrapping actions for the exchange
+func Run(ctx context.Context, b IBotExchange) {
```
shazbert, line 1769: Suggestion: We could use another wait group and parallel every REST request here to make it a bit more snappier if you want.

shazbert (changes requested):
Nice work, I always enjoy seeing excess code fat removed. For the point `exchange.Run to CommonBootstrap`  IMO I am happy for that to be changed. 
The only thing I would consider with a type switch and checking the interface it can lead to an obfuscation. Just having it as a forced wrapper function might be better, so devs know it's there. 

Linter issues need to be fixed but did a quick test of all exchanges and looks good. Awesome stuff.

exchanges/exchange.go
gbjk: Good suggestion

Fixed 905d6555

gbjk:
* Linter fixed
* Revised pattern
* Simplified waitgroups

exchanges/exchange.go
```diff
@@ -1752,3 +1751,64 @@ func (b *Base) MatchSymbolCheckEnabled(symbol string, a asset.Item, hasDelimiter
...
+	}()
+
+	wg.Wait()
+}
+
+// Bootstrap is a fallback method for exchange startup actions
+// Exchange authors should overide this if they wysh to customise startup actions
+// Return true to all default Bootstrap actions to occur afterwars
+// or false to signal that no further bootstrapping should occur
+func (b *Base) Bootstrap(_ context.Context) (continueBootstrap bool) {
```
shazbert, line 1806: Can we return an error for this method, so we don't need to log everything wrapper side if implemented?

shazbert (changes requested):
Just one thing on new changes. Linters need a look at as well. Then lgtm.

exchanges/exchange.go
gbjk: I thought the same thing myself more than once.
If we both thought it, Let's make it so.
On it like a car bonnet.

gbjk: Fixed abac8c89

Note Well: Behaviour change; We will now return errors all the way up if Bootstrap errors.
I think this is a good thing.

shazbert (changes requested):
Linters need to be fixed, then this looks good to go. Thanks for the changes.

gbjk:
Codespell fixed 5044aca6 and 4e118364
Linter fixed 7daf81f2

exchanges/exchange.go
```diff
@@ -1752,3 +1751,72 @@ func (b *Base) MatchSymbolCheckEnabled(symbol string, a asset.Item, hasDelimiter
...
+				errC <- fmt.Errorf("failed to set exchange order execution limits: %w", aErr)
+			}
+		}
+	}()
+
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if b.GetEnabledFeatures().AutoPairUpdates {
+			if aErr := b.UpdateTradablePairs(ctx, false); aErr != nil {
```
shazbert, line 1794: It was throwing a bunch of errors when update order execution limits was trying to match symbols with loaded pairs. This helps to minimise that issue. Total loading is down to ~9s which is pretty good. 
```diff
diff --git a/exchanges/exchange.go b/exchanges/exchange.go
index 5e906ce37..706af3811 100644
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -1774,6 +1774,12 @@ func Bootstrap(ctx context.Context, b IBotExchange) error {
                b.PrintEnabledPairs()
        }

+       if b.GetEnabledFeatures().AutoPairUpdates {
+               if err := b.UpdateTradablePairs(ctx, false); err != nil {
+                       return err
+               }
+       }
+
        errC := make(chan error, 16) // Blocking isn't a dealbreaker, but a little buffer is nice
        var wg sync.WaitGroup
        wg.Add(1)
@@ -1787,15 +1793,6 @@ func Bootstrap(ctx context.Context, b IBotExchange) error {
                }
        }()

-       wg.Add(1)
-       go func() {
-               defer wg.Done()
-               if b.GetEnabledFeatures().AutoPairUpdates {
-                       if aErr := b.UpdateTradablePairs(ctx, false); aErr != nil {
-                               errC <- fmt.Errorf("failed to update tradable pairs: %w", aErr)
-                       }
-               }
-       }()
        wg.Wait()
        close(errC)
```

shazbert (changes requested):
Thanks for making that change, initial start is nice and snappy. 
download

exchanges/exchange.go
gbjk: Fixed 1763395e

shazbert (approved):
**LOVELY** 😍 tACK.

exchanges/exchange.go
```diff
@@ -1876,3 +1874,67 @@ func (b *Base) ParallelChanOp(channels []subscription.Subscription, m func([]sub
...
+	for i := range a {
+		wg.Add(1)
+		go func(a asset.Item) {
+			defer wg.Done()
+			if err := b.UpdateOrderExecutionLimits(ctx, a); err != nil && !errors.Is(err, common.ErrNotYetImplemented) {
+				errC <- fmt.Errorf("failed to set exchange order execution limits: %w", err)
+			}
+		}(a[i])
+	}
+	close(errC)
```
thrasher-, line 1918: This needs a wg.Wait() before it otherwise we'll panic on sending to a closed channel

exchanges/exchange.go
```diff
@@ -1876,3 +1874,67 @@ func (b *Base) ParallelChanOp(channels []subscription.Subscription, m func([]sub
...
+	}
+
+	if b.IsVerbose() {
+		wsURL := ""
+		wsEnabled := false
+		if w, err := b.GetWebsocket(); err != nil {
+			wsURL = w.GetWebsocketURL()
+			wsEnabled = w.IsEnabled()
+		}
+		log.Debugf(log.ExchangeSys, "%s Websocket: %s. (url: %s).\n", b.GetName(), common.IsEnabled(wsEnabled), wsURL)
```
thrasher-, line 1896: Not all exchanges support websocket so we can check to see if the exchange supports it and only print the above if it does so

thrasher- (changes requested):
Very nice! Like the simplification. Looks like there's a spelling + linter issue. Also noticed that now we're updating Bybit's order execution limits, it is throwing a bunch of errors on startup:
image

shazbert (changes requested):
Forgot to request a change on the cmd/exchange_template/Wrapper_file.tmpl we can remove `start` and `run`

exchanges/exchange.go
gbjk: Yeah. That's a horrific oversight.
Must have been accidentally deleted, I'd assume.
Fixed d99caa20

exchanges/exchange.go
gbjk: It's also wrong AF: `[DEBUG] | EXCHANGE | 30/01/2024 12:13:57 | Bybit Websocket: Disabled. (url: ).`
🤷 
On it.

gbjk: Yeah. That was stupid. Should have been `err == nil`.

Hope it's okay, but I've decided to be specific that the websocket is unsupported. We're being verbose, after all.

Fixed 3cccf3c0

gbjk:
> Forgot to request a change on the cmd/exchange_template/Wrapper_file.tmpl we can remove `start` and `run`

Fixed 93627919

exchanges/exchange.go
```diff
@@ -1876,3 +1874,72 @@ func (b *Base) ParallelChanOp(channels []subscription.Subscription, m func([]sub
...
+	for e := range errC {
+		err = common.AppendError(err, e)
+	}
+
+	return err
+}
+
+// Bootstrap is a fallback method for exchange startup actions
+// Exchange authors should override this if they wish to customise startup actions
+// Return true or an error to all default Bootstrap actions to occur afterwars
```
gbjk, line 1935: Huh. `afterwars`

gbjk: Fixed 72019b8d

gbjk:
✅ Linter fixed
✅ Spelling fixed (again, somehow 🤦 )
✅ Rogue Alphapoint `Start` moved to `Bootstrap`
✅ Exchange template updated to remove `Start` and `Run`
✅ Verbose Debug fixed and updated to support Unsupported

shazbert (changes requested):
Just one last thing I think:

```diff
diff --git a/cmd/exchange_template/wrapper_file.tmpl b/cmd/exchange_template/wrapper_file.tmpl
index 4d04fd4c0..d57f96b54 100644
--- a/cmd/exchange_template/wrapper_file.tmpl
+++ b/cmd/exchange_template/wrapper_file.tmpl
@@ -3,8 +3,6 @@ package {{.Name}}

 import (
        "context"
-       "fmt"
-       "sync"
        "time"

        "github.com/thrasher-corp/gocryptotrader/common"
```

gbjk:
> Just one last thing I think:
> 
> ```diff
> diff --git a/cmd/exchange_template/wrapper_file.tmpl b/cmd/exchange_template/wrapper_file.tmpl
> index 4d04fd4c0..d57f96b54 100644
> --- a/cmd/exchange_template/wrapper_file.tmpl
> +++ b/cmd/exchange_template/wrapper_file.tmpl
> @@ -3,8 +3,6 @@ package {{.Name}}
> 
>  import (
>         "context"
> -       "fmt"
> -       "sync"
>         "time"
> 
>         "github.com/thrasher-corp/gocryptotrader/common"
> ```

Fixed a348fc18

shazbert (changes requested):
Quick testing there might be an issue when testing the keys.

```
[WARN]  | EXCHANGE | 30/01/2024 17:03:46 | BTC Markets: Cannot validate credentials, authenticated support has been disabled, Error: BTC Markets unsuccessful HTTP status code: 401 raw response: {"code":"InvalidAPIKey","message":"invalid api key"}, authenticated request failed
[DEBUG] | EXCHANGE | 30/01/2024 17:03:46 | BTC Markets: Exchange support: Enabled (Authenticated API support: Enabled - Verbose mode: Disabled).
```

AUTH support is still enabled and then it tries to connect auth services then breaks.

gbjk:
Seems the same for me on master:
```
[WARN]  | EXCHANGE | 30/01/2024 15:08:03 | BTC Markets: Cannot validate credentials, authenticated support has been disabled, Error: BTC Markets unsuccessful HTTP status code: 401 raw response: {"code":"InvalidAPIKey","message":"invalid api key"}, authenticated request failed
[DEBUG] | EXCHANGE | 30/01/2024 15:08:03 | BTC Markets: Exchange support: Enabled (Authenticated API support: Enabled - Verbose mode: Disabled).
```

I can't reproduce the "and then breaks" part - But that might depend on how you're seeing it break.

shazbert (approved):
Thanks for checking that out, it seems to be an isolated issue on BTCM on Master. tACK.

gloriousCode (approved):
tACK! Handled pre-engine config, all exchanges tested, benchmarked

thrasher- (approved):
tACK with the Bybit issue added to Trello.

gbjk:
Rebased on master ready for merge.

PR #1417, opened 2023-12-10
exchanges/futures: Implement open interest

gloriousCode:
# PR Description
This PR adds support for open interest for exchanges that support it. Its an important metric for measuring futures trading

## Type of change
- [x] New feature (non-breaking change which adds functionality)

## How has this been tested
- [x] go test ./... -race
- [x] golangci-lint run
- [x] Coverage of all new functionality

## Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally and on Github Actions/AppVeyor with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #1442, opened 2024-01-09
Binance: Fix GetHistoricTrades failing live

gbjk:
This test was failing with `-tags mock_test_off` because the API imposes a default limit of 500 now, so we'd just get 500 trades and no batching.

If the user doesn't provide a limit, that means we need to assume batching

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestGetHistoricTrades


thrasher- (approved):
tACK with mock tests off, the other two live test failures will be resolved once #1417 is synced with master

shazbert (approved):
tACK! nice 🦅

PR #1444, opened 2024-01-10
Binance: Subscribe response handling

gbjk:
* ~Add a generic WS mock and instance fixture for re-use~
* Add listSubscriptions API support
* Handle Subscribe and Unsubscribe response errors
This is sub-optimal, because it turns out that no matter what we do, we can't verify the subscriptions.
A bad subscription name not only doesn't error, but it's actually returned by listSubscriptions.
All this really achieves is ensuring that the message is actually received and acknowledged before marking a subscription as successful.
Maybe they'll fix the API at some point.

~I'll roll the mock WS handling out to other exchanges in a separate PR.~

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/common/file/archive/zip_test.go
+++ b/common/file/archive/zip_test.go
@@ -3,8 +3,12 @@ package archive
 import (
 	"archive/zip"
 	"errors"
+	"io/fs"
 	"path/filepath"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestUnZip(t *testing.T) {
@@ -33,60 +37,39 @@ func TestUnZip(t *testing.T) {
 
 func TestZip(t *testing.T) {
 	tempDir := t.TempDir()
-	singleFile := filepath.Join("..", "..", "..", "testdata", "configtest.json")
 	outFile := filepath.Join(tempDir, "out.zip")
-	err := Zip(singleFile, outFile)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err := Zip(filepath.Join("..", "..", "..", "testdata", "configtest.json"), outFile)
+	require.NoError(t, err, "Zip should not error")
 	o, err := UnZip(outFile, tempDir)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if len(o) != 1 {
-		t.Fatalf("expected 1 files to be extracted received: %v ", len(o))
-	}
+	require.NoError(t, err, "UnZip should not error")
+	assert.Len(t, o, 1, "Should extract 1 file")
 
-	folder := filepath.Join("..", "..", "..", "testdata", "http_mock")
+	folder := filepath.Join("..", "..", "..", "testdata", "gctscript")
 	outFolderZip := filepath.Join(tempDir, "out_folder.zip")
 	err = Zip(folder, outFolderZip)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err, "Zip should not error")
 	o, err = UnZip(outFolderZip, tempDir)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err, "UnZip should not error")
 	var found bool
 	for i := range o {
-		if filepath.Base(o[i]) == "binance.json" {
+		if filepath.Base(o[i]) == "timer.gct" {
 			found = true
 		}
 	}
-	if !found {
-		t.Fatal("could not find file in zip")
-	}
-
-	if expected := 6; len(o) < expected {
-		t.Fatalf("expected at least %v files to be extracted, received: %v ", expected, len(o))
-	}
+	assert.True(t, found, "Should find a gctscript in the zip")
+	assert.GreaterOrEqual(t, len(o), 6, "Should extract at least 6 files")
 
 	folder = filepath.Join("..", "..", "..", "testdata", "invalid_file.json")
-	outFolderZip = filepath.Join(tempDir, "invalid.zip")
-	err = Zip(folder, outFolderZip)
-	if err == nil {
-		t.Fatal("expected IsNotExistError on invalid file")
-	}
+	err = Zip(folder, filepath.Join(tempDir, "invalid.zip"))
+	assert.ErrorIs(t, err, fs.ErrNotExist, "Zip should error correctly")
 
 	addFilesToZip = addFilesToZipTestWrapper
 	folder = filepath.Join("..", "..", "..", "testdata", "http_mock")
 	outFolderZip = filepath.Join(tempDir, "error_zip.zip")
 	err = Zip(folder, outFolderZip)
-	if err == nil {
-		t.Fatal("expected Zip() to fail due to invalid addFilesToZipTestWrapper()")
-	}
+	assert.ErrorContains(t, err, "specific error", "Zip should error correctly")
 }
 
 func addFilesToZipTestWrapper(_ *zip.Writer, _ string, _ bool) error {
-	return errors.New("error")
+	return errors.New("specific error")
 }
--- a/config/config.go
+++ b/config/config.go
@@ -1850,9 +1850,18 @@ func (c *Config) UpdateConfig(configPath string, newCfg *Config, dryrun bool) er
 	return c.LoadConfig(configPath, dryrun)
 }
 
-// GetConfig returns a pointer to a configuration object
+// GetConfig returns the global shared config instance
 func GetConfig() *Config {
-	return &Cfg
+	m.Lock()
+	defer m.Unlock()
+	return &cfg
+}
+
+// SetConfig sets the global shared config instance
+func SetConfig(c *Config) {
+	m.Lock()
+	defer m.Unlock()
+	cfg = *c
 }
 
 // RemoveExchange removes an exchange config
--- a/exchanges/binance/binance_websocket.go
+++ b/exchanges/binance/binance_websocket.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/buger/jsonparser"
 	"github.com/gorilla/websocket"
 	"github.com/thrasher-corp/gocryptotrader/currency"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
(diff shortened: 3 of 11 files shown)


gbjk:
Rebased on master 🔁

gbjk:
Rebased and ready 👍

exchanges/binance/binance_websocket.go
```diff
@@ -601,55 +604,108 @@ func channelName(s *subscription.Subscription) (string, error) {
...
 		}
 	}
-	b.Websocket.RemoveSubscriptions(channelsToUnsubscribe...)
+
+	if err != nil {
+		err = fmt.Errorf("%w: %w; Channels: %s", stream.ErrUnsubscribeFailure, err, strings.Join(cNames, ", "))
+		b.Websocket.DataHandler <- err
+	}
+
+	b.Websocket.RemoveSubscriptions(chans...)
```
shazbert, line 683: If we error above should we be removing subs? Maybe consider just returning above on error.

exchanges/binance/binance_websocket.go
```diff
@@ -601,55 +604,108 @@ func channelName(s *subscription.Subscription) (string, error) {
...
+
 	return nil
 }
 
+// listSubscriptoins returns the current list of subscribed channels
+func (b *Binance) listSubscriptions() ([]string, error) {
```
shazbert, line 689: This isn't being used in the current work flow or being tested? Any chance of integrating this and adding it to the sub/unsub?

shazbert (changes requested):
Cool stuff, just some things I noticed.

exchanges/binance/binance_websocket.go
```diff
@@ -548,83 +552,160 @@ func (b *Binance) UpdateLocalBuffer(wsdp *WebsocketDepthStream) (bool, error) {
...
+		Params: cNames,
+		ID:     id,
+	}
+
+	respRaw, err := b.Websocket.Conn.SendMessageReturnResponse(id, req)
+	if err == nil {
+		if _, d, _, rErr := jsonparser.Get(respRaw, "result"); rErr != nil {
+			err = rErr
+		} else if d != jsonparser.Null { // null is the only expected and acceptable response
+			err = errUnknownError
```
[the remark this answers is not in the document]
gbjk: I've seen no evidence we will ever get any indication at all that the sub wasn't sucessful, even when it's insane.
We'll get result:null back, we'll get it back in listSubs, it'll just be a non-operational subscription to the sexual activity feed of giraffes in the Serengeti 🤷 

Point being: I don't wanna go guessing what result is going to look like when we've got no evidence of it.

However I agree we need to put an RemoveSub in ( that's what you meant by unsub ? ) below, right?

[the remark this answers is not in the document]
shazbert: Okay, how about we add the respRaw to the error fmt.Errorf("%w: %s") just for a little more context if it does change?

In this function we already add the the subscription, but its hard to justify unsubscribing even if there is a malformed response from what you are saying about the animal kingdoms and such. 

In my other comment about the function `listSubscriptions` can we add that in to this function and cross check what's potentially been subscribed?

Also in this function it returns an error but down below in `unsubscribeFromChan` it doesn't return an error at the end of the function, should it be?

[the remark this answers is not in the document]
gbjk: Fixed 521c0cc3

exchanges/binance/binance_websocket.go
gbjk: I thought about that too. I was on the fence.

Let's remove it.

Fixed d30ebb3b

exchanges/binance/binance_websocket.go
gbjk: See commit message 0c584dd9

gbjk: Note @thrasher-  @gloriousCode 👆 This is one of my best arguments for not squashing.
If I want to implement something, and then roll it back but show how it was done, I do it like this, and that commit in the history becomes important, even though it was reverted inside the PR.
Don't have a solution other than maybe labelling some PRs to be FF merged instead of squash merged

shazbert: @gbjk I think in future it's best to have those caveats included as a warning or note on the actual function comment, instead of diving into git history. 

If what you say is true and the fact that its not included in any other function, it's not exported and its not tested. Just remove it completely, it's dead code. Unless you want to use it as a library function export it and test it and add commentary.

gbjk: Okay; I'll remove it.

Regarding comments versus commits, I think both are warranted. The comment should explain why, but the commit would give a viable implementation should things change. A better approach is probably to punt it into a mothball/ branch though, and I'll do that now instead.

shazbert (approved):
Good stuff nothing else from me. tACK.

gloriousCode (changes requested):
Excellent

exchanges/binance/binance_websocket.go
```diff
@@ -601,52 +604,87 @@ func channelName(s *subscription.Subscription) (string, error) {
...
```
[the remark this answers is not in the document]
gbjk: Fixed fa7a3e8f

Sorry for the delay. Encountered a panic along the way.

gloriousCode (changes requested):
Thanks for making those changes! Just need to fix up the linters

gbjk:
Linter satisfied.
Several new commits that need reviewing though. Should be obvious.

config/config.go
```diff
@@ -1850,9 +1850,14 @@ func (c *Config) UpdateConfig(configPath string, newCfg *Config, dryrun bool) er
...
-// GetConfig returns a pointer to a configuration object
+// GetConfig returns the global shared config instance
 func GetConfig() *Config {
-	return &Cfg
+	return &cfg
+}
+
+// SetConfig sets the global shared config instance
+func SetConfig(c *Config) {
+	cfg = *c
```
gloriousCode, line 1860: In its current use, there is no need to use the mutex `m`. I can foresee this being a race at some other time and suggest using `m` to protect from that like it protects elsewhere in this package. I'm not preventing merge since its pre-existing

gloriousCode (approved):
tACK!

config/config.go
gbjk: I'll do that. Trying not to get too into it because I'm sure this is a 🧵 I'll keep pulling at...

Fixed 26d22b18

Avoided doing RWMutex and fixing other uses at the same time.

gloriousCode (approved):

shazbert (approved):
Ch ACK. Nice work!

thrasher- (changes requested):
Tested subscriptions and working good, nice! Just a consistency nit

thrasher- (changes requested):
Also noticed this test error due to the testdata file move: https://github.com/thrasher-corp/gocryptotrader/actions/runs/7751182517/job/21139393179?pr=1444#step:9:144

exchanges/binance/binance_websocket.go
```diff
@@ -601,52 +604,87 @@ func channelName(s *subscription.Subscription) (string, error) {
...
```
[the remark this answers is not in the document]
gbjk: Thanks.
Fixed 5d035327

exchanges/binance/binance_websocket.go
```diff
@@ -601,52 +604,87 @@ func channelName(s *subscription.Subscription) (string, error) {
...
+
+	req := WsPayload{
+		Method: wsUnsubscribeMethod,
+		Params: cNames,
+		ID:     id,
+	}
+
+	respRaw, err := b.Websocket.Conn.SendMessageReturnResponse(id, req)
+	if err == nil {
+		if _, d, _, rErr := jsonparser.Get(respRaw, "result"); rErr != nil {
```
thrasher-, line 674:
```suggestion
		if v, d, _, rErr := jsonparser.Get(respRaw, "result"); rErr != nil {
```

To make it easy to resolve after the below

gbjk:
> Also noticed this test error due to the testdata file move: https://github.com/thrasher-corp/gocryptotrader/actions/runs/7751182517/job/21139393179?pr=1444#step:9:144

Fixed 0f631528

exchanges/binance/binance_websocket.go
gbjk: Sorry, already fixed but got delayed pushing

Fixed aadd1a38

common/file/archive/zip_test.go
```diff
@@ -33,60 +36,39 @@ func TestUnZip(t *testing.T) {
...
 	}
-	if !found {
-		t.Fatal("could not find file in zip")
-	}
-
-	if expected := 6; len(o) < expected {
-		t.Fatalf("expected at least %v files to be extracted, received: %v ", expected, len(o))
-	}
+	assert.True(t, found, "Should find a gctscript in the zip")
+	assert.Equal(t, 9, len(o), "Should extract 6 files")
```
thrasher-, line 59:
```suggestion
	assert.Equal(t, 9, len(o), "Should extract 9 files")
```

OR

```suggestion
	assert.GreaterOrEqual(t, len(o), 6, "Should extract at least 6 files")
```

The second gives leeway in case scripts are added in the future which I believe was the original intent. Happy with either.

common/file/archive/zip_test.go
```diff
@@ -33,60 +36,39 @@ func TestUnZip(t *testing.T) {
...
 
 	folder = filepath.Join("..", "..", "..", "testdata", "invalid_file.json")
-	outFolderZip = filepath.Join(tempDir, "invalid.zip")
-	err = Zip(folder, outFolderZip)
-	if err == nil {
-		t.Fatal("expected IsNotExistError on invalid file")
-	}
+	err = Zip(folder, filepath.Join(tempDir, "invalid.zip"))
+	assert.ErrorContains(t, err, "no such file or directory", "Zip should error correctly")
```
thrasher-, line 63:
```suggestion
	assert.ErrorIs(t, err, fs.ErrNotExist, "Zip should error with fs.ErrNotExist")
```

This is cross platform error friendly as Billy G's tests are currently unhappy

thrasher- (changes requested):
Should be good after these 2

common/file/archive/zip_test.go
gbjk: Thanks. I went hunting for what to use and didn't quickly find it.

Fixed  6b59bda0

common/file/archive/zip_test.go
gbjk: Fixed 76a4cd64

thrasher- (approved):
tACK, thank you for making those changes!

PR #1451, opened 2024-01-16
Kraken: Improve ticker batch performance

gloriousCode:
# PR Description
Noticed during PPROF that kraken batch tickers was responsible for a big delay (> 10 seconds). Thankfully Kraken has updated their API to handle better batch support, which I've now added.

This improves performance by > 7x. Here are my benchmarks and tests that I used:

|Bench|ns|b/op|alloc|
|---|---|---|---|
|BenchmarkUpdateTicker-10    |   	       7745914458 ns/op	|29208788320 B/op	 |2698082 allocs/op|
|BenchmarkUpdateTickersNew-10    	|       1017438583 ns/op	 |2045088 B/op	    |8991 allocs/op|

🎉 

```go

func BenchmarkUpdateTickers(b *testing.B) {
	b.ReportAllocs()
	ap, err := k.GetAvailablePairs(asset.Spot)
	if err != nil {
		b.Fatal(err)
	}
	err = k.CurrencyPairs.StorePairs(asset.Spot, ap, true)
	if err != nil {
		b.Fatal(err)
	}
	for i := 0; i < b.N; i++ {
		err = k.UpdateTickers(context.Background(), asset.Spot)
		if err != nil {
			b.Error(err)
		}
	}
}

// Obviously this won't work just by switching to the branch and running this
// you would need to grab the old function and have them both in tandem
// just wanted to show my working, okay?
func BenchmarkUpdateTickersNew(b *testing.B) {
	b.ReportAllocs()
	ap, err := k.GetAvailablePairs(asset.Spot)
	if err != nil {
		b.Fatal(err)
	}
	err = k.CurrencyPairs.StorePairs(asset.Spot, ap, true)
	if err != nil {
		b.Fatal(err)
	}
	for i := 0; i < b.N; i++ {
		err = k.UpdateTickersNew(context.Background(), asset.Spot)
		if err != nil {
			b.Error(err)
		}
	}
}
```

Updated tests to ensure all pairs have tickers. Though Kraken does return zeros for some ticks, but by its status is "online" and matches available pairs, so I consider it fine.

Fixes # (issue)
REALLY SLOW TICKS

## How has this been tested
- [x] `go test ./... -race`
- [x] `golangci-lint run`
- [x] `TestNewPairFromSymbol`
- [x] Updated `TestUpdateTickers`

## Checklist


exchanges/kraken/kraken_wrapper.go
```diff
@@ -485,53 +485,52 @@ func (k *Kraken) UpdateTradablePairs(ctx context.Context, forceUpdate bool) erro
 	return k.EnsureOnePairEnabled()
 }
 
+func (k *Kraken) newPairFromSymbol(symbol string, item asset.Item) (currency.Pair, error) {
+	cp, err := k.MatchSymbolWithAvailablePairs(symbol, item, item == asset.Futures)
+	if err != nil {
+		if errors.Is(err, currency.ErrPairNotFound) {
```
shazbert, line 491: Might be better to invert this check and return the error, just so if there is any funny things going on with that function above instead of potentially returning an empty pair and nil. 

```
if !errors.Is(err, currency.ErrPairNotFound) {
   return err
}
// Lookup here.
```

shazbert (changes requested):
Only one thing I noticed. Looking good 🚀

exchanges/kraken/kraken_wrapper.go
gloriousCode: Much nicer!

shazbert (approved):
Exchange CurrencyPairSyncer initial sync took 3.7173014s [666 sync items]. 😈 

tACK.

gloriousCode:
soz @shazbert just added some extra fields. You prompted it though so not my fault

shazbert (approved):
CH tACK. Well done.

exchanges/kraken/kraken_wrapper.go
```diff
@@ -485,53 +485,56 @@ func (k *Kraken) UpdateTradablePairs(ctx context.Context, forceUpdate bool) erro
...
-					Open:         t.Open,
-					Pair:         pairs[i],
-					ExchangeName: k.Name,
-					AssetType:    a})
-				if err != nil {
-					return err
+		for c, t := range tickers {
+			cp, err := k.newPairFromSymbol(c, a)
+			if err != nil {
+				if errors.Is(err, currency.ErrPairNotFound) {
```
thrasher-, line 517: This check is not needed since it's replicated in `newPairFromSymbol`

thrasher- (changes requested):
Nice! One nit

thrasher- (changes requested):
Looks like there's a bunch of test errors popping up https://github.com/thrasher-corp/gocryptotrader/actions/runs/7604600459/job/20712567636?pr=1451#step:9:607

thrasher- (approved):
tACK!

PR #1456, opened 2024-01-24
engine/gRPC proxy: Fix mux regression and add test coverage

thrasher-:
# PR Description

A user spotted a regression in Slack that the mux wasn't set for the gRPC proxy handler. This PR adds it back and adds test coverage.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/engine/rpcserver.go
+++ b/engine/rpcserver.go
@@ -220,6 +220,7 @@ func (s *RPCServer) authClient(handler http.Handler) http.Handler {
 		if !ok || username != s.Config.RemoteControl.Username || password != s.Config.RemoteControl.Password {
 			w.Header().Set("WWW-Authenticate", `Basic realm="restricted"`)
 			http.Error(w, "Access denied", http.StatusUnauthorized)
+			log.Warnf(log.GRPCSys, "gRPC proxy server unauthorised access attempt. IP: %s Path: %s\n", r.RemoteAddr, r.URL.Path)
 			return
 		}
 		handler.ServeHTTP(w, r)
--- a/engine/rpcserver.go
+++ b/engine/rpcserver.go
@@ -172,12 +172,14 @@ func StartRPCServer(engine *Engine) {
 
 // StartRPCRESTProxy starts a gRPC proxy
 func (s *RPCServer) StartRPCRESTProxy() {
-	log.Debugf(log.GRPCSys, "gRPC proxy server support enabled. Starting gRPC proxy server on http://%v.\n", s.Config.RemoteControl.GRPC.GRPCProxyListenAddress)
+	log.Debugf(log.GRPCSys, "gRPC proxy server support enabled. Starting gRPC proxy server on https://%v.\n", s.Config.RemoteControl.GRPC.GRPCProxyListenAddress)
 
 	targetDir := utils.GetTLSDir(s.Settings.DataDir)
-	creds, err := credentials.NewClientTLSFromFile(filepath.Join(targetDir, "cert.pem"), "")
+	certFile := filepath.Join(targetDir, "cert.pem")
+	keyFile := filepath.Join(targetDir, "key.pem")
+	creds, err := credentials.NewClientTLSFromFile(certFile, "")
 	if err != nil {
-		log.Errorf(log.GRPCSys, "Unabled to start gRPC proxy. Err: %s\n", err)
+		log.Errorf(log.GRPCSys, "Unable to start gRPC proxy. Err: %s\n", err)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


engine/rpcserver_test.go
```diff
@@ -4139,3 +4148,99 @@ func TestGetOpenInterest(t *testing.T) {
...
+
+	fakeTime := time.Now().Add(-time.Hour)
+	e.uptime = fakeTime
+
+	StartRPCServer(e)
+
+	// Give the proxy time to start
+	time.Sleep(time.Millisecond * 500)
+
+	targetDir := utils.GetTLSDir(common.GetDefaultDataDir(""))
```
gloriousCode, line 4184: How about a temp dir instead of going with a user's files?
```go
tempDir := filepath.Join(os.TempDir(), "gct-temp-tls")
	cleanup := func() {
		if err := os.RemoveAll(tempDir); err != nil {
			t.Errorf("unable to remove temp dir %s, manual deletion required", tempDir)
		}
	}
```

engine/rpcserver_test.go
```diff
@@ -4139,3 +4148,99 @@ func TestGetOpenInterest(t *testing.T) {
...
+	require.True(t, ok, "AppendCertsFromPEM should return true")
+	client := &http.Client{Transport: &http.Transport{TLSClientConfig: &tls.Config{RootCAs: caCertPool, MinVersion: tls.VersionTLS12}}}
+
+	req, err := http.NewRequestWithContext(context.Background(), http.MethodGet, "https://localhost:"+strconv.Itoa(gRPCProxyPort)+"/v1/getinfo", http.NoBody)
+	require.NoError(t, err, "NewRequestWithContext should not error")
+	req.SetBasicAuth("bobmarley", "Sup3rdup3rS3cr3t")
+	resp, err := client.Do(req)
+	require.NoError(t, err, "Do should not error")
+	defer resp.Body.Close()
+
```
gloriousCode, line 4200: One extra scenario for this one:
```go
req, err = http.NewRequestWithContext(context.Background(), http.MethodGet, "https://localhost:"+strconv.Itoa(gRPCProxyPort)+"/v1/getinfo", http.NoBody)
	require.NoError(t, err, "NewRequestWithContext should not error")
	req.SetBasicAuth("wif", "bonk")
	resp, err = client.Do(req)
	require.NoError(t, err, "Do should not error")
	require.Equal(t, resp.StatusCode, 401, "incorrect credentials should return 401")
```

gloriousCode (changes requested):
Nice, thanks for actioning this!

engine/rpcserver_test.go
thrasher-: Yup, everything else was mocked so this can be too 👍

gloriousCode (changes requested):
Lovely test improvements!
Just noticed on the easy example response, with the move to using https:
image
Want to update the string to `https`?

thrasher-:
Yeppers! Pooshed https://github.com/thrasher-corp/gocryptotrader/pull/1456/commits/f9ac86df52a00f8070fb8ae5c37bde167a63d5b6

gloriousCode (approved):
tACK!

brook-w:
Please continue to add HTTP CORS support, in order to better use it in the web （吴兴）
@thrasher-

engine/rpcserver.go
```diff
@@ -200,16 +202,30 @@ func (s *RPCServer) StartRPCRESTProxy() {
...
 	log.Debugln(log.GRPCSys, "gRPC proxy server started!")
 }
 
+func (s *RPCServer) authClient(handler http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		username, password, ok := r.BasicAuth()
+		if !ok || username != s.Config.RemoteControl.Username || password != s.Config.RemoteControl.Password {
+			w.Header().Set("WWW-Authenticate", `Basic realm="restricted"`)
+			http.Error(w, "Access denied", http.StatusUnauthorized)
+			return
```
shazbert, line 223: Suggestion: It might be better to log an authentication failure server side as well.

shazbert (approved):
Looking good one suggestion. tACK! 🚀

shazbert:
> Please continue to add HTTP CORS support, in order to better use it in the web （吴兴） Can be configured in config.json. I think this function is very practical @thrasher-

@brook-w Good suggestion but out of scope for this as this will probably need a middleware handler and integration with config, PR's are welcome.

engine/rpcserver.go
thrasher-: Added with a note: by default we bind to localhost so you'd be notified if something on your machine is trying to gain access to it (with ip reported being localhost). However, it would be good if someone changes the listen address and in that case the IP log would be more useful

shazbert (approved):
Rude tACK

```
[WARN]  | GRPC | 30/01/2024 12:52:45 | gRPC proxy server unauthorised access attempt. IP: 127.0.0.1:59740 Path: /favicon.ico
```

PR #1457, opened 2024-01-26
Websockets: Fix subscription failure on reconnect

gbjk:
If a websocket is disconnected then the subscription map was left with old subscriptions, causing:
`Okx websocket: subscription failure, channel already subscribed for ...`

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -270,6 +270,8 @@ func (w *Websocket) Connect() error {
 			w.exchangeName)
 	}
 
+	w.subscriptions = subscriptionMap{}
+
 	w.dataMonitor()
 	w.trafficMonitor()
 	w.setConnectingStatus(true)


exchanges/stream/websocket.go
```diff
@@ -270,6 +270,8 @@ func (w *Websocket) Connect() error {
 			w.exchangeName)
 	}
 
+	w.subscriptions = subscriptionMap{}
```
gloriousCode, line 273:
```suggestion
        w.subscriptionMutex.Lock()
	w.subscriptions = subscriptionMap{}
	w.subscriptionMutex.Unlock()
```
Arguable whether its needed, but its used anywhere else it gets touched

gloriousCode (changes requested):
Happy with the solution given the pathway of a disconnect never touches `Shutdown` to reset subs

exchanges/stream/websocket.go
gbjk: Agreed.

Fixed c71460f8

gloriousCode (approved):
tACK!

shazbert (approved):
image

Cool beans

thrasher- (approved):
Thanks for investigating + resolving this

PR #1464, opened 2024-02-01
Binance: Fix UpdateOrderExecutionLimits margin test

gbjk:
Calls for limits for margin asset were not doing anything if spot is enabled.
This caused intermittent test failures when the sequence meant margin was tested before spot, since the limits wouldn't be loaded.

But it also highlighted that API users calling Update Limits for margin outside the context of a loop on all assets would not get an update. Also intermittently when the loop of UpdateLimits on assets puts margin first

The unfortunate side to this is that we can't avoid having to hit the API twice, because UpdateOrderExecutionLimits can never know if it's being called in isolation or not. Caching could fix that, but it'd be a new paradigm in this context.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/binance/binance.go
+++ b/exchanges/binance/binance.go
@@ -7,8 +7,10 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"slices"
 	"sort"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/thrasher-corp/gocryptotrader/common"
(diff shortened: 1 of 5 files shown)


gbjk:
CodeCoverage changes reviewed and acceptable.
No changes in coverage introduced that I can spot.
Think the delta is due to change in total lines maybe.

exchanges/binance/binance.go
```diff
@@ -1209,72 +1211,65 @@ func (b *Binance) MaintainWsAuthStreamKey(ctx context.Context) error {
...
-			case "SPOT":
-				assets = append(assets, asset.Spot)
-			case "MARGIN":
-				assets = append(assets, asset.Margin)
-			default:
-				// "LEVERAGED", "TRD_GRP_003", "TRD_GRP_004", "TRD_GRP_005" etc are unused permissions
-				// for spot exchange limits
-			}
+
+		if !slices.Contains(s.Permissions, strings.ToUpper(a.String())) {
```
shazbert, line 1233: 🔥

shazbert (approved):
Good stuff, nothing from me. tACK!

exchanges/binance/binance.go
gbjk: Wait. [This shouldn't work with 1.20](https://pkg.go.dev/slices?tab=versions)
😢 
I don't actually get how it's passing.

gbjk: go.mod needed updating
Opened #1467

gloriousCode, line 1233: Only thing I'd suggest here is to move `strings.ToUpper(...)` out of the `for` so you don't need to process it for every symbol and for each item in `s.Permissions`

gloriousCode (changes requested):
Thanks for the fun fixes!

exchanges/binance/binance.go
gbjk: Agreed.

Fixed 1ee33f73

gloriousCode (approved):
tACK! 🐎💨

shazbert (approved):
Ch ACK.

thrasher- (approved):
Thank you for looking into this and resolving it 🚀

gbjk:
Squashed and rebased for merging

PR #1466, opened 2024-02-04
Websocket: Various refactors and test improvements

gbjk:
* Websocket: Remove IsInit
IsInit was basically the same as IsConnected.
Any time Connect was called both would be set to true.
Any time we had a disconnect they'd both be set to false
Shutdown() incorrectly didn't setInit(false)
* SetProxyAddress simplified to only reconnect a connected Websocket.
Any other state means it hasn't been Connected, or it's about to
reconnect anyway.
There was no handling for IsConnecting previously, either, so I've wrapped
SetProxyAddress behind the main mutex.
* Websocket: Improve and simplify tests
* Simplify state (init, disc, connecting, connected) to a single state field
* Move WebsocketNotEnabled to a real ErrWebsocketNotEnabled

This PR started because of inconsistencies noticed with IsInit and state from a okx panic bug.
It kinda broke out containment at that point and went native. KSozThx.

## Type of change

- [x] Refactoring (non-breaking change)

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -65,8 +65,10 @@ var (
 	errChannelAlreadySubscribed             = errors.New("channel already subscribed")
 	errInvalidChannelState                  = errors.New("invalid Channel state")
 	errSameProxyAddress                     = errors.New("cannot set proxy address to the same address")
-	errNoConnectFunc                        = errors.New("connect func not set")
-	errAlreadyConnected                     = errors.New("already connected")
+	errNoConnectFunc                        = errors.New("websocket connect func not set")
+	errAlreadyConnected                     = errors.New("websocket already connected")
+	errCannotShutdown                       = errors.New("websocket cannot shutdown")
+	errAlreadyReconnecting                  = errors.New("websocket in the process of reconnection")
 )
 
 var globalReporter Reporter
@@ -266,10 +268,10 @@ func (w *Websocket) Connect() error {
 		return errors.New(WebsocketNotEnabled)
 	}
 	if w.IsConnecting() {
-		return fmt.Errorf("%v Websocket already attempting to connect", w.exchangeName)
+		return fmt.Errorf("%v %w", w.exchangeName, errAlreadyReconnecting)
 	}
 	if w.IsConnected() {
-		return fmt.Errorf("%v Websocket already connected", w.exchangeName)
+		return fmt.Errorf("%v %w", w.exchangeName, errAlreadyConnected)
 	}
 
 	w.subscriptionMutex.Lock()
@@ -444,12 +446,12 @@ func (w *Websocket) Shutdown() error {
 	defer w.m.Unlock()
 
 	if !w.IsConnected() {
-		return fmt.Errorf("%v websocket: cannot shutdown %w", w.exchangeName, ErrNotConnected)
+		return fmt.Errorf("%v %w: %w", w.exchangeName, errCannotShutdown, ErrNotConnected)
 	}
 
 	// TODO: Interrupt connection and or close connection when it is re-established.
 	if w.IsConnecting() {
-		return fmt.Errorf("%v websocket: cannot shutdown, in the process of reconnection", w.exchangeName)
+		return fmt.Errorf("%v %w: %w ", w.exchangeName, errCannotShutdown, errAlreadyReconnecting)
 	}
 
 	if w.verbose {
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -477,10 +477,11 @@ func (w *Websocket) Shutdown() error {
 	w.subscriptions = subscriptionMap{}
 	w.subscriptionMutex.Unlock()
 
+	w.setState(disconnected)
+
 	close(w.ShutdownC)
 	w.Wg.Wait()
 	w.ShutdownC = make(chan struct{})
-	w.setState(disconnected)
 	if w.verbose {
 		log.Debugf(log.WebsocketMgr, "%v websocket: completed websocket shutdown", w.exchangeName)
 	}
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -24,24 +24,20 @@ const (
 	defaultTrafficPeriod = time.Second
 )
 
+// Public websocket errors
 var (
-	// ErrSubscriptionNotFound defines an error when a subscription is not found
-	ErrSubscriptionNotFound = errors.New("subscription not found")
-	// ErrSubscribedAlready defines an error when a channel is already subscribed
-	ErrSubscribedAlready = errors.New("duplicate subscription")
-	// ErrSubscriptionFailure defines an error when a subscription fails
-	ErrSubscriptionFailure = errors.New("subscription failure")
-	// ErrSubscriptionNotSupported defines an error when a subscription channel is not supported by an exchange
+	ErrWebsocketNotEnabled      = errors.New("websocket not enabled")
+	ErrSubscriptionNotFound     = errors.New("subscription not found")
+	ErrSubscribedAlready        = errors.New("duplicate subscription")
+	ErrSubscriptionFailure      = errors.New("subscription failure")
 	ErrSubscriptionNotSupported = errors.New("subscription channel not supported ")
-	// ErrUnsubscribeFailure defines an error when a unsubscribe fails
-	ErrUnsubscribeFailure = errors.New("unsubscribe failure")
-	// ErrChannelInStateAlready defines an error when a subscription channel is already in a new state
-	ErrChannelInStateAlready = errors.New("channel already in state")
-	// ErrAlreadyDisabled is returned when you double-disable the websocket
-	ErrAlreadyDisabled = errors.New("websocket already disabled")
-	// ErrNotConnected defines an error when websocket is not connected
-	ErrNotConnected = errors.New("websocket is not connected")
+	ErrUnsubscribeFailure       = errors.New("unsubscribe failure")
+	ErrChannelInStateAlready    = errors.New("channel already in state")
+	ErrAlreadyDisabled          = errors.New("websocket already disabled")
+	ErrNotConnected             = errors.New("websocket is not connected")
+)
 
+var (
 	errAlreadyRunning                       = errors.New("connection monitor is already running")
 	errExchangeConfigIsNil                  = errors.New("exchange config is nil")
 	errWebsocketIsNil                       = errors.New("websocket is nil")
@@ -265,7 +261,7 @@ func (w *Websocket) Connect() error {
 	defer w.m.Unlock()
 
 	if !w.IsEnabled() {
-		return errors.New(WebsocketNotEnabled)
+		return ErrWebsocketNotEnabled
 	}
 	if w.IsConnecting() {
 		return fmt.Errorf("%v %w", w.exchangeName, errAlreadyReconnecting)
@@ -490,7 +486,7 @@ func (w *Websocket) Shutdown() error {
 // FlushChannels flushes channel subscriptions when there is a pair/asset change
 func (w *Websocket) FlushChannels() error {
 	if !w.IsEnabled() {
-		return fmt.Errorf("%s websocket: service not enabled", w.exchangeName)
+		return fmt.Errorf("%s %w", w.exchangeName, ErrWebsocketNotEnabled)
 	}
 
 	if !w.IsConnected() {
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -40,19 +40,23 @@ var (
 var (
 	errAlreadyRunning                       = errors.New("connection monitor is already running")
 	errExchangeConfigIsNil                  = errors.New("exchange config is nil")
+	errExchangeConfigEmpty                  = errors.New("exchange config is empty")
 	errWebsocketIsNil                       = errors.New("websocket is nil")
 	errWebsocketSetupIsNil                  = errors.New("websocket setup is nil")
 	errWebsocketAlreadyInitialised          = errors.New("websocket already initialised")
+	errWebsocketAlreadyEnabled              = errors.New("websocket already enabled")
 	errWebsocketFeaturesIsUnset             = errors.New("websocket features is unset")
 	errConfigFeaturesIsNil                  = errors.New("exchange config features is nil")
 	errDefaultURLIsEmpty                    = errors.New("default url is empty")
 	errRunningURLIsEmpty                    = errors.New("running url cannot be empty")
 	errInvalidWebsocketURL                  = errors.New("invalid websocket url")
-	errExchangeConfigNameUnset              = errors.New("exchange config name unset")
+	errExchangeConfigNameEmpty              = errors.New("exchange config name empty")
 	errInvalidTrafficTimeout                = errors.New("invalid traffic timeout")
+	errTrafficAlertNil                      = errors.New("traffic alert is nil")
 	errWebsocketSubscriberUnset             = errors.New("websocket subscriber function needs to be set")
 	errWebsocketUnsubscriberUnset           = errors.New("websocket unsubscriber functionality allowed but unsubscriber function not set")
 	errWebsocketConnectorUnset              = errors.New("websocket connector function not set")
+	errReadMessageErrorsNil                 = errors.New("read message errors is nil")
 	errWebsocketSubscriptionsGeneratorUnset = errors.New("websocket subscriptions generator function needs to be set")
 	errClosedConnection                     = errors.New("use of closed network connection")
 	errSubscriptionsExceedsLimit            = errors.New("subscriptions exceeds limit")
@@ -65,6 +69,7 @@ var (
 	errAlreadyConnected                     = errors.New("websocket already connected")
 	errCannotShutdown                       = errors.New("websocket cannot shutdown")
 	errAlreadyReconnecting                  = errors.New("websocket in the process of reconnection")
+	errConnSetup                            = errors.New("error in connection setup")
 )
 
 var globalReporter Reporter
@@ -107,7 +112,7 @@ func (w *Websocket) Setup(s *WebsocketSetup) error {
 	}
 
 	if s.ExchangeConfig.Name == "" {
-		return errExchangeConfigNameUnset
+		return errExchangeConfigNameEmpty
 	}
 	w.exchangeName = s.ExchangeConfig.Name
 	w.verbose = s.ExchangeConfig.Verbose
@@ -196,22 +201,22 @@ func (w *Websocket) Setup(s *WebsocketSetup) error {
 // SetupNewConnection sets up an auth or unauth streaming connection
 func (w *Websocket) SetupNewConnection(c ConnectionSetup) error {
 	if w == nil {
-		return errors.New("setting up new connection error: websocket is nil")
+		return fmt.Errorf("%w: %w", errConnSetup, errWebsocketIsNil)
 	}
 	if c == (ConnectionSetup{}) {
-		return errors.New("setting up new connection error: websocket connection configuration empty")
+		return fmt.Errorf("%w: %w", errConnSetup, errExchangeConfigEmpty)
 	}
 
 	if w.exchangeName == "" {
-		return errors.New("setting up new connection error: exchange name not set, please call setup first")
+		return fmt.Errorf("%w: %w", errConnSetup, errExchangeConfigNameEmpty)
 	}
 
 	if w.TrafficAlert == nil {
-		return errors.New("setting up new connection error: traffic alert is nil, please call setup first")
+		return fmt.Errorf("%w: %w", errConnSetup, errTrafficAlertNil)
 	}
 
 	if w.ReadMessageErrors == nil {
-		return errors.New("setting up new connection error: read message errors is nil, please call setup first")
+		return fmt.Errorf("%w: %w", errConnSetup, errReadMessageErrorsNil)
 	}
 
 	connectionURL := w.GetWebsocketURL()
@@ -314,7 +319,7 @@ func (w *Websocket) Connect() error {
 // Note that connectionMonitor will be responsible for shutting down the websocket after disabling
 func (w *Websocket) Disable() error {
 	if !w.IsEnabled() {
-		return fmt.Errorf("%w for exchange '%s'", ErrAlreadyDisabled, w.exchangeName)
+		return fmt.Errorf("%s %w", w.exchangeName, ErrAlreadyDisabled)
 	}
 
 	w.setEnabled(false)
@@ -324,8 +329,7 @@ func (w *Websocket) Disable() error {
 // Enable enables the exchange websocket protocol
 func (w *Websocket) Enable() error {
 	if w.IsConnected() || w.IsEnabled() {
-		return fmt.Errorf("websocket is already enabled for exchange %s",
-			w.exchangeName)
+		return fmt.Errorf("%s %w", w.exchangeName, errWebsocketAlreadyEnabled)
 	}
 
 	w.setEnabled(true)
@@ -490,7 +494,7 @@ func (w *Websocket) FlushChannels() error {
 	}
 
 	if !w.IsConnected() {
-		return fmt.Errorf("%s websocket: service not connected", w.exchangeName)
+		return fmt.Errorf("%s %w", w.exchangeName, ErrNotConnected)
 	}
 
 	if w.features.Subscribe {
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -128,7 +128,7 @@ func (w *Websocket) Setup(s *WebsocketSetup) error {
 	if s.ExchangeConfig.Features == nil {
 		return fmt.Errorf("%s %w", w.exchangeName, errConfigFeaturesIsNil)
 	}
-	w.enabled = s.ExchangeConfig.Features.Enabled.Websocket
+	w.setEnabled(s.ExchangeConfig.Features.Enabled.Websocket)
 
 	if s.Connector == nil {
 		return fmt.Errorf("%s %w", w.exchangeName, errWebsocketConnectorUnset)
@@ -388,9 +388,7 @@ func (w *Websocket) connectionMonitor() error {
 	if w.checkAndSetMonitorRunning() {
 		return errAlreadyRunning
 	}
-	w.fieldMutex.RLock()
 	delay := w.connectionMonitorDelay
-	w.fieldMutex.RUnlock()
 
 	go func() {
 		timer := time.NewTimer(delay)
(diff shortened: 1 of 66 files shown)


gloriousCode (changes requested):
From a high level, looks quite good.
I don't understand this point:
> There's no handling for IsConnecting previously

And does this mean that the panic is fixed, is it still present?
> from a okx panic bug

exchanges/stream/websocket_types.go
```diff
@@ -25,13 +23,21 @@ const (
...
+
 // Websocket defines a return type for websocket connections via the interface
 // wrapper for routine processing
 type Websocket struct {
 	canUseAuthenticatedEndpoints bool
 	enabled                      bool
-	Init                         bool
-	connected                    bool
+	state                        state
 	connecting                   bool
```
shazbert, line 41: connecting field in this struct can be removed

shazbert (changes requested):
One thing from me, tested and everything else looks good. 👍

exchanges/stream/websocket_test.go
```diff
@@ -92,416 +97,267 @@ type dodgyConnection struct {
...
-		t.Error("It's not")
-	}
-
-	isADisconnectionError = IsDisconnectionError(&net.OpError{
-		Err: errors.New("errText"),
-	})
-	if !isADisconnectionError {
-		t.Error("It is")
-	}
+	assert.False(t, IsDisconnectionError(errors.New("errorText")), "IsADisconnectionError should return false")
```
thrasher-, line 204:
```suggestion
	assert.False(t, IsDisconnectionError(errors.New("errorText")), "IsDisconnectionError should return false")
```

For this one and the blow

thrasher- (changes requested):
Nice! Only minor things. The assert package sure gets rid of a lot of testing boilerplate code

gbjk:
> From a high level, looks quite good. I don't understand this point:
> 
> > There's no handling for IsConnecting previously

In the context of the SetProxyAddress changes.
I've added a bulletpoint to make this context clearer in the PR description and clarified the mutex.
Specifically the issue is that a reconnecting websocket could have it's proxy changed and maybe race or end up in an undefined state.

> And does this mean that the [okx] panic is fixed, is it still present?

Yes. I think the whole thing started when I was trying to debug it and found the connecting state being left in a strange situation during unwinding one of the connections.
At the end of finishing this work I was so far from that work that I needed to step back and clear my desk of this one so I can focus on the okx panic in isolation.

exchanges/stream/websocket.go
```diff
@@ -98,7 +103,7 @@ func (w *Websocket) Setup(s *WebsocketSetup) error {
 		return errWebsocketSetupIsNil
 	}
 
-	if !w.Init {
+	if w.state != uninitialised {
```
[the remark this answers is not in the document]
gbjk: Thanks! I agree

Fixed 75a9718a

Includes a DriveBy of comment spotted in the neighbourhood.

Raises the question if Setup should lock the main w.m though ... Thoughts?

exchanges/stream/websocket_types.go
gbjk: Absolutely. Thought it was removed. Thanks for catching.

Fixed ac5fe67b

exchanges/stream/websocket_test.go
gbjk: This is so funny. I was copying the test messages and thinking "IsADisc... is a stupid name but I'm not going to fix it, I've done enough" ... so karma dictates I should learn from being lazy - it was wrong after all anyway!

Fixed 52ebf55a

shazbert (approved):
Thanks for making that change. Cheers 🍻  tACK.

exchanges/stream/websocket.go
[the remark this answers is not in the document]
gbjk: Okay. I'm going to put a man mutex around Setup then.

Fixed b65f102d

shazbert (approved):
ch ACK.

gloriousCode (changes requested):
Last thing I noticed

exchanges/stream/websocket.go
```diff
@@ -593,24 +573,19 @@ func (w *Websocket) trafficMonitor() {
...
-					log.Warnf(log.WebsocketMgr,
-						"%v websocket: has not received a traffic alert in %v. Reconnecting",
-						w.exchangeName,
-						w.trafficTimeout)
+					log.Warnf(log.WebsocketMgr, "%v websocket: has not received a traffic alert in %v. Reconnecting", w.exchangeName, w.trafficTimeout)
 				}
 				trafficTimer.Stop()
 				w.setTrafficMonitorRunning(false)
 				w.Wg.Done() // without this the w.Shutdown() call below will deadlock
 				if !w.IsConnecting() && w.IsConnected() {
```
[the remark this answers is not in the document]
gbjk: Right, yep.
I had to think about this for a second.
We don't have any traffic, but we're trying to reconnect already, so we shouldn't do anything.
However I think this implies a bigger change to handle Connecting properly in this context.

And now I'm looking at it my eye is caught by the "Routine pausing mechanism" briefly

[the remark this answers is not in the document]
gbjk: Fixed cff39913
> Websocket: Fix and simplify traffic monitor
> 
> trafficMonitor had a check throttle at the end of the for loop to stop it just gobbling the (blocking) trafficAlert channel non-stop.
> That makes sense, except that nothing is sent to the trafficAlert channel if there's no listener.
> So that means that it's out by one second on the trafficAlert, because any traffic received during the pause is doesn't try to send a traffic alert.
> 
> The unstopped timer is deliberately leaked for later GC when shutdown.
> It won't delay/block anything, and it's a trivial memory leak during an infrequent event.
> 
> Deliberately Choosing to recreate the timer each time instead of using Stop, drain and reset

Also accidentally 1c7441c

> Websocket: Use atomics instead of mutex
> 
> This was spurred by looking at the setState call in trafficMonitor and
> the effect on blocking and efficiency.
> With the new atomic types in Go 1.19, and the small types in use here,
> atomics should be safe for our usage. bools should be truly atomic,
> and uint32 is atomic when the accepted value range is less than one byte/uint8 since
> that can be written atomicly by concurrent processors.
> Maybe that's not even a factor any more, however we don't even have to worry enough to check.
> 
> 

Tomorrow I'll check if the existing tests of trafficMonitor are sufficient, since this was a heftier rewrite than intended.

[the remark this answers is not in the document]
gbjk: Tests extended to cover all trafficMonitor behaviours.
Also caught several bugs in doing so. Embarrassingly large ones. Go tests! 🎉 ( Excuse the pun )

Rebased on master to get latest linter versions on this PR.

gbjk:
For clarity on re-reviewing:
* No previous PR fixes have been squashed, but they have been rebased on master to get the linter
* The only commits needing re-review should be: 208ca57a, c3edef2c and 3e19e83e
( 208ca57a Is resequenced and really trivial, but mentioning it for full disclosure )

exchanges/stream/websocket_test.go
```diff
@@ -92,416 +97,343 @@ type dodgyConnection struct {
...
 // override websocket connection method to produce a wicked terrible error
 func (d *dodgyConnection) Connect() error {
-	return errors.New("cannot connect due to some dastardly reason")
+	return fmt.Errorf("cannot connect: %w", errDastardlyReason)
+}
+
+func TestMain(_ *testing.M) {
+	// Change trafficCheckInterval for TestTrafficMonitorTimeout before parallel tests to avoid racing
+	trafficCheckInterval = 50 * time.Millisecond
 }
```
shazbert, line 111: Or the tests don't actually run
```suggestion
func TestMain(t *testing.M) {
	// Change trafficCheckInterval for TestTrafficMonitorTimeout before parallel tests to avoid racing
	trafficCheckInterval = 50 * time.Millisecond
	os.Exit(t.Run())
}
```

exchanges/stream/websocket.go
```diff
@@ -388,34 +387,25 @@ func (w *Websocket) connectionMonitor() error {
...
 					}
 				}
 				if w.verbose {
-					log.Debugf(log.WebsocketMgr,
-						"%v websocket: connection monitor exiting\n",
-						w.exchangeName)
+					log.Debugf(log.WebsocketMgr, "%v websocket: connection monitor exiting", w.exchangeName)
 				}
 				timer.Stop()
 				w.setConnectionMonitorRunning(false)
```
shazbert, line 411: Seems like we never actually set connection monitor running as true anyywhere, might be able to drop this method completely if not used?

exchanges/stream/websocket.go
```diff
@@ -576,183 +555,116 @@ func (w *Websocket) trafficMonitor() {
 	w.Wg.Add(1)
 
 	go func() {
-		var trafficTimer = time.NewTimer(w.trafficTimeout)
-		pause := make(chan struct{})
+		t := time.NewTimer(w.trafficTimeout)
```
shazbert, line 558: suggestion: don't need a timer can pivot around a time variable, it could potential t.Stop calls and drain, but either ways yours works 🤷 so can ignore.

shazbert (changes requested):
Cool beans! Nice work 🖖

exchanges/stream/websocket_test.go
```diff
@@ -18,6 +18,7 @@ import (
```
gloriousCode, line 18: Once `TestMain` is executing tests again, I got the following race
```
==================
WARNING: DATA RACE
Write at 0x000104f1c970 by goroutine 10:
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.TestTrafficMonitorTimeout()
      /Users/a/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket_test.go:252 +0xb28
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1742 +0x40

Previous read at 0x000104f1c970 by goroutine 66:
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).trafficMonitor.func1()
      /Users/a/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:569 +0x74

Goroutine 10 (running) created at:
  testing.(*T).Run()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1742 +0x5e4
  testing.runTests.func1()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:2161 +0x80
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1689 +0x180
  testing.runTests()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:2159 +0x6e0
  testing.(*M).Run()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:2027 +0xb6c
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.TestMain()
      /Users/a/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket_test.go:112 +0x44
  main.main()
      _testmain.go:111 +0x294

Goroutine 66 (running) created at:
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).trafficMonitor()
      /Users/a/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:557 +0xd0
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).Connect()
      /Users/a/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:285 +0x314
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.TestConnectionMessageErrors()
      /Users/a/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket_test.go:315 +0x740
  testing.tRunner()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1689 +0x180
  testing.(*T).Run.gowrap1()
      /opt/homebrew/Cellar/go/1.22.0/libexec/src/testing/testing.go:1742 +0x40
==================
--- FAIL: TestTrafficMonitorTimeout (0.58s)
    testing.go:1398: race detected during execution of test
--- FAIL: TestGenerateMessageID (0.63s)
    testing.go:1398: race detected during execution of test
--- FAIL: TestDial (0.00s)
    --- FAIL: TestDial/test1 (1.14s)
        testing.go:1398: race detected during execution of test
--- FAIL: TestSendMessage (0.00s)
    --- FAIL: TestSendMessage/test1 (1.17s)
        testing.go:1398: race detected during execution of test
--- FAIL: TestWebsocketConnectionShutdown (1.18s)
    testing.go:1398: race detected during execution of test
--- FAIL: TestSendMessageWithResponse (1.18s)
    testing.go:1398: race detected during execution of test
--- FAIL: TestLatency (2.18s)
    testing.go:1398: race detected during execution of test
--- FAIL: TestSetupPingHandler (2.56s)
    testing.go:1398: race detected during execution of test
FAIL
exit status 1
```

gloriousCode (changes requested):
The changes all appear pretty good. Off to the races though 🏎️ 🏇 💨

exchanges/stream/websocket_connection.go
```diff
@@ -240,7 +227,7 @@ func (w *WebsocketConnection) ReadMessage() Response {
 
 	select {
 	case w.Traffic <- struct{}{}:
-	default: // causes contention, just bypass if there is no receiver.
+	default: // Non-Blocking write ensuses 1 buffered signal per trafficCheckInterval to avoid flooding
```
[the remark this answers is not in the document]
gbjk: Don't mind me. I'm just here to take comments and introduce typos into them 😂 

Fixed 61910caa

exchanges/stream/websocket_test.go
gbjk: Bwahahaha! Well. Fixed the race, didn't it?

FWIW, `m` should be called `m`.
Not like I'm on firm ground to be slinging corrections though!

Fixed 56b62692 and 7dd00dc9

exchanges/stream/websocket.go
gbjk: I don't see a clear path to that suggestion.
If t becomes time.Time then line 579 needs to use a channel somehow.
If 579 becomes time.After() it'd need to subtract Now() from the time declared at 558, which feels messy.
I scanned through time pkg for any other channel methods than time.After() and couldn't see any.

Happy to change it if there's an improvement, but I can't see it yet.

Note: I tried variations with timer and ticker several times. This was the least convoluted solution I could spot.

exchanges/stream/websocket.go
gbjk: We set it running [here](https://github.com/thrasher-corp/gocryptotrader/blob/4bcf97340b4492d5b01378933f45a877f70bf2a5/exchanges/stream/websocket.go#L387) using `checkAndSetMonitorRunning`

exchanges/stream/websocket_test.go
gbjk: Already fixed 7dd00dc9 😄 
It was due to the other setting of `trafficCheckInterval`.
Sloppy last minute fix from me. Sorry.

shazbert (approved):
Thanks for making those changes. Looks good tACK.

exchanges/stream/websocket.go
```diff
@@ -576,183 +555,116 @@ func (w *Websocket) trafficMonitor() {
...
 				return
-			case <-w.TrafficAlert:
-				if !trafficTimer.Stop() {
-					select {
-					case <-trafficTimer.C:
-					default:
+			case <-time.After(trafficCheckInterval):
+				select {
+				case <-w.TrafficAlert:
+					w.setState(connected)
```
gloriousCode, line 572: I think the only foil to this is how you've changed `w.TrafficAlert` to be a buffered channel. In that context, you actually don't know if the state is connected when you receive a message here, just that at the time the traffic alert was sent, you were connected.

In `connectionMonitor`, when a disconnection error message is received, you set the state to `disconnected`, but no further action is taken until the next cycle until `timer.C` does a check of the connection status and then it will reconnect.

From that, it sounds like there is the possibility of the `TrafficAlert` buffered channel, setting the state to `connected` while disconnected and things attempt to carry on.

When you call `ReadMessage` on a disconnected websocket, my testing says you don't continue to receive `IsDisconnectionError` errors, instead you get the following, so you don't get back to the state "disconnected" until `websocketTrafficTimeout` which default is 30 seconds. So it will eventually reconnect
image

I feel like I'm thinking about this too much though. I like your buffered channel change. I'm not asking you to change this. What are your thoughts on what I'm saying?

gbjk: I think what you're saying makes sense.

Before: Traffic alerts would be sent only when listening, with the impact that we'd miss traffic that might turn out to be important.
Now: Traffic alerts aren't atomic, so we don't know if the sequencing is valid.

Also, I've just realised:
When the traffic Timeout Timer goes off, we might already have traffic, but we're not checking for it.

I'm going to think about this. The tests I have should allow me to iterate a solution.

I already had a thought about solving this problem differently anyway, so I'll take another swing at it.
It means firing a collector goro, which is fine.

I'll get back to you shortly.

gbjk: This hypothesis should be testable.

Split the existing test into behaviours with d6918391 and I'll add a new one now for this behaviour described, which should fail.

gbjk: Fixed 74437455

By never setting the state to connected we avoid any of these issues, and really there's no reason to.

I've reduced the trafficCheckInterval to 100ms.

We also check to make sure any pending traffic signals are seen when we get the trafficTimeout, in case it's out of sequence with trafficCheckInterval

exchanges/stream/websocket_test.go
```diff
@@ -221,33 +222,48 @@ func TestTrafficMonitorTimeout(t *testing.T) {
...
+			assert.Truef(c, ws.IsConnected(), "state should still be connected; Check #%d", i)
 			assert.Emptyf(c, ws.TrafficAlert, "trafficAlert channel should be drained; Check #%d", i)
-		}, 2*trafficCheckInterval, patience, "trafficAlert should be read and state connected; Check #%d", i)
+		}, 2*trafficCheckInterval, patience, "trafficAlert should be read; Check #%d", i)
 	}
 
-	// Behaviour: Shuts down websocket and exits on timeout
 	require.EventuallyWithT(t, func(c *assert.CollectT) {
 		assert.Equal(c, disconnected, ws.state.Load(), "websocket must be disconnected")
 		assert.False(c, ws.IsTrafficMonitorRunning(), "trafficMonitor shound be shut down")
```
gloriousCode, line 232:
```suggestion
		assert.False(c, ws.IsTrafficMonitorRunning(), "trafficMonitor should be shut down")
```

exchanges/stream/websocket_test.go
```diff
@@ -221,33 +222,48 @@ func TestTrafficMonitorTimeout(t *testing.T) {
...
 	ws.trafficMonitor()
-	assert.True(t, ws.IsTrafficMonitorRunning(), "traffic monitor should be running")
+	require.True(t, ws.IsTrafficMonitorRunning(), "traffic monitor should be running")
 	require.Equal(t, connecting, ws.state.Load(), "websocket must be connecting")
 	<-time.After(4 * ws.trafficTimeout)
 	require.Equal(t, connecting, ws.state.Load(), "websocket must still be connecting after several checks")
 	ws.state.Store(connected)
 	require.EventuallyWithT(t, func(c *assert.CollectT) {
 		assert.Equal(c, disconnected, ws.state.Load(), "websocket must be disconnected")
 		assert.False(c, ws.IsTrafficMonitorRunning(), "trafficMonitor shound be shut down")
```
gloriousCode, line 254:
```suggestion
		assert.False(c, ws.IsTrafficMonitorRunning(), "trafficMonitor should be shut down")
```

gloriousCode (changes requested):
I agree with your conclusions, nice work

exchanges/stream/websocket.go
gloriousCode: Now that its not impacting the connected status, I can understand re-increasing the CheckInterval again. I assume you changed it to this for performance reasons (since we used to act on every single message received)

exchanges/stream/websocket.go
shazbert: This is what I meant:
```diff
diff --git a/exchanges/stream/websocket.go b/exchanges/stream/websocket.go
index 7d1d402d1..f77ad308a 100644
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -555,27 +555,26 @@ func (w *Websocket) trafficMonitor() {
        w.Wg.Add(1)

        go func() {
-               t := time.NewTimer(w.trafficTimeout)
+               defer func() {
+                       w.setTrafficMonitorRunning(false)
+                       w.Wg.Done()
+                       if !w.IsConnected() {
+                               return
+                       }
+                       if err := w.Shutdown(); err != nil {
+                               log.Errorf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown err: %s", w.exchangeName, err)
+                       }
+               }()
+
+               next := time.Now().Add(w.trafficTimeout)
                for {
                        select {
                        case <-w.ShutdownC:
                                if w.verbose {
                                        log.Debugf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown message received", w.exchangeName)
                                }
-                               t.Stop()
-                               w.setTrafficMonitorRunning(false)
-                               w.Wg.Done()
                                return
                        case <-time.After(trafficCheckInterval):
-                               select {
-                               case <-w.TrafficAlert:
-                                       if !t.Stop() {
-                                               <-t.C
-                                       }
-                                       t.Reset(w.trafficTimeout)
-                               default:
-                               }
-                       case <-t.C:
                                checkAgain := w.IsConnecting()
                                select {
                                case <-w.TrafficAlert:
@@ -583,20 +582,15 @@ func (w *Websocket) trafficMonitor() {
                                default:
                                }
                                if checkAgain {
-                                       t.Reset(w.trafficTimeout)
+                                       next = time.Now().Add(w.trafficTimeout)
+                                       break
+                               }
+                               if time.Now().Before(next) {
                                        break
                                }
                                if w.verbose {
                                        log.Warnf(log.WebsocketMgr, "%v websocket: has not received a traffic alert in %v. Reconnecting", w.exchangeName, w.trafficTimeout)
                                }
-                               w.setTrafficMonitorRunning(false) // Cannot defer lest Connect is called after Shutdown but before deferred call
-                               w.Wg.Done()                       // Without this the w.Shutdown() call below will deadlock
-                               if w.IsConnected() {
-                                       err := w.Shutdown()
-                                       if err != nil {
-                                               log.Errorf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown err: %s", w.exchangeName, err)
-                                       }
-                               }
                                return
                        }
                }
diff --git a/exchanges/stream/websocket_test.go b/exchanges/stream/websocket_test.go
index b099adca1..0ce64edca 100644
--- a/exchanges/stream/websocket_test.go
+++ b/exchanges/stream/websocket_test.go
@@ -280,7 +280,7 @@ func TestTrafficMonitorShutdown(t *testing.T) {
        case <-time.After(1 * trafficCheckInterval):
        }

-       close(ws.ShutdownC)
+       require.NoError(t, ws.Shutdown())

        <-time.After(2 * trafficCheckInterval)
        assert.False(t, ws.IsTrafficMonitorRunning(), "traffic monitor should be shutdown")
```

exchanges/stream/websocket_test.go
```diff
@@ -92,416 +98,358 @@ type dodgyConnection struct {
...
+
+	wgReady := make(chan bool)
+	go func() {
+		ws.Wg.Wait()
+		close(wgReady)
+	}()
+	select {
+	case <-wgReady:
+		require.Failf(t, "", "WaitGroup should be blocking still")
+	case <-time.After(1 * trafficCheckInterval):
```
shazbert, line 280:
```suggestion
	case <-time.After(trafficCheckInterval):
```

shazbert (changes requested):
Looking good! 👍 Just one pedantic thing

exchanges/stream/websocket_test.go
gbjk: Thanks! 🤦 

Fixed d0cd94ee

exchanges/stream/websocket_test.go
```diff
@@ -221,33 +222,48 @@ func TestTrafficMonitorTimeout(t *testing.T) {
...
+func TestTrafficMonitorShutdown(t *testing.T) {
+	t.Parallel()
+	ws := NewWebsocket()
+	err := ws.Setup(defaultSetup)
+	require.NoError(t, err, "Setup must not error")
+
+	ws.ShutdownC = make(chan struct{})
+	ws.trafficTimeout = 50 * time.Millisecond
 	ws.state.Store(connected)
 	ws.trafficTimeout = time.Minute
```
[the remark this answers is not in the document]
gbjk: I knew I lost that somewhere. Thanks!

Fixed d0cd94ee

exchanges/stream/websocket_test.go
gbjk: Fixed d0cd94ee

gbjk:
@shazbert Thanks!
I like your direction there. A couple of notes though:

1. As it is now I'm pretty sure the defer will race because you have Shutdown 👉  ShutdownC 👉  Shutdown:
➡️ Shutdown
:arrow_right: Wg.Wait
:arrow_right: context switch to trafficMonitor
:arrow_right: ShutdownC
:arrow_right: defer
🚨  Still connected at this point
:arrow_right: Call Shutdown again (and block on mutex)
:arrow_right: Other Shutdown gets to finish and set disconnected
:arrow_right: Back to defer's Shutdown call again, now unblocked
:arrow_right: IsConnected now false, return error
⛔  defer logs the error

I think there's no harm to moving the setState(disconnected) in Shutdown to before close(ShutdownC) because it better represents the state. If that causes something to call Connect then that's fine because w.m is still Locked.

2. My bigger problem with your pattern is the calls to time.Now() every cycle. Channels and timers are cheap and delegated to OS, but calling time.Now is much more expensive, I think. I guess we'll have to bench it 🤷 

3. On using w.Shutdown() instead of ShutdownC: where there's a dependency on a channel like this, I believe it's more unity to use the dependency than call a method, so that you're not unit testing more units than you need to. trafficMonitor depends on the channel, not on Shutdown(), and so where possible I'll use the channel.

exchanges/stream/websocket_test.go
gbjk: Fixed e040f17b

shazbert:
> @shazbert Thanks! I like your direction there. A couple of notes though:
> 
> 1. As it is now I'm pretty sure the defer will race because you have Shutdown 👉  ShutdownC 👉  Shutdown:
>    ➡️ Shutdown
>    ➡️ Wg.Wait
>    ➡️ context switch to trafficMonitor
>    ➡️ ShutdownC
>    ➡️ defer
>    🚨  Still connected at this point
>    ➡️ Call Shutdown again (and block on mutex)
>    ➡️ Other Shutdown gets to finish and set disconnected
>    ➡️ Back to defer's Shutdown call again, now unblocked
>    ➡️ IsConnected now false, return error
>    ⛔  defer logs the error
> 
> > I think there's no harm to moving the setState(disconnected) in Shutdown to before close(ShutdownC) because it better represents the state. If that causes something to call Connect then that's fine because w.m is still Locked.
> > 👆  Fixed [89be3d8](https://github.com/thrasher-corp/gocryptotrader/commit/89be3d85773d6732bac5be290a95ee9bcfe56945)
> 
> 2. My bigger problem with your pattern is the calls to time.Now() every cycle. Channels and timers are cheap and delegated to OS, but calling time.Now is much more expensive, I think. I guess we'll have to bench it 🤷
> 3. On using w.Shutdown() instead of ShutdownC: where there's a dependency on a channel like this, I believe it's more unity to use the dependency than call a method, so that you're not unit testing more units than you need to. trafficMonitor depends on the channel, not on Shutdown(), and so where possible I'll use the channel.

Yeah all good points, I will let you decide on how you want to proceed. All valid and at the end of the day, yours works.

gbjk:
I've used [this](https://go.dev/play/p/zNv1613xjOs) to sniff the cost of time.Now().Add() versus channels.
```
BenchmarkAddTime-8      14308972                82.98 ns/op
BenchmarkChannel-8      38685134                30.75 ns/op
```
Even if we reduce the bench to just doing a time.Now().Add(it) and a t.Reset(it) we end up with 40 vs 35.

Finally I've looked at the defered cleanup.
In essence we add 2 lines for the defer, Run a Stop when we don't need to, and and run a Connected and Shutdown check when we don't need to from ShutdownC.
Or put another way: Two lines are common, but we waste time in both directions, and we add 2 lines for the defer.

So I'm going to leave it as is 😄

gloriousCode (approved):

shazbert (approved):
ch ack. Thanks 👍

thrasher- (approved):
tACK, nice improvements!

gbjk:
Rebased on master and squashed commits.

PR #1467, opened 2024-02-05
Deps: Require go 1.21

gbjk:
Our build toolchain is using 1.21 since #1313
This hides issues on 1.20 from go.mod
Notably uses of implied Generics and slices package seem okay, but would fail for anyone running 1.20 locally

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/thrasher-corp/gocryptotrader
 
-go 1.20
+go 1.21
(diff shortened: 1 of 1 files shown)


thrasher- (approved):
Thanks!

shazbert (approved):
LGTM!

PR #1471, opened 2024-02-06
Websocket: Prevent panic on Disconnect/Connect

gbjk:
Previously we'd set the websocket to disconnected when *either* of the
Conn/AuthConn got a disconnect message.

This meant that:
* The other connection was still functioning
* A user would be free to call Connect()
* wsReadData() may not have exited

Any call to Connect would be acceptable, and we'd probably get a panic
from the underlying shared/re-used gorilla websocket:
`panic: runtime error: slice bounds out of range [:13501] with capacity 8192`
when a new wsReadData goro is started and the old tries to ReadMessage
as well, overallocating the buffer.

This wouldn't normally occur because trafficMonitor would see traffic
(on either connection) and then set the websocket to being connected
again.

The solution is to treat a Disconnect on either websocket as a call to
Shutdown the whole websocket, and then trafficMonitor can reconnect it
properly.

Unit Testing for this has been difficult. So far I've had to rely on a
disruption inside websocket's connectionMonitor itself to reproduce the
panic, but from there it's been very consistent.

Depends on merge of #1466

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -351,7 +351,7 @@ func (w *Websocket) dataMonitor() {
 			w.setDataMonitorRunning(false)
 			w.Wg.Done()
 		}()
-		full := 0
+		dropped := 0
 		for {
 			select {
 			case <-w.ShutdownC:
@@ -359,15 +359,16 @@ func (w *Websocket) dataMonitor() {
 			case d := <-w.DataHandler:
 				select {
 				case w.ToRoutine <- d:
-					if full {
-						full = false
+					if dropped != 0 {
+						log.Infof(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer recovered; %d messages were dropped", w.exchangeName, dropped)
+						dropped = 0
 					}
 				default:
-					if !full {
-						full = true
+					if dropped == 0 {
 						// If this becomes prone to flapping we could drain the buffer, but that's extreme and we'd like to avoid it if possible
-						log.Warnf(log.WebsocketMgr, "%s exchange backlog in websocket ToRoutine consumer channel; dropping messages", w.exchangeName)
+						log.Warnf(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer full; dropping messages", w.exchangeName)
 					}
+					dropped++
 				}
 			}
 		}
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -348,18 +348,10 @@ func (w *Websocket) dataMonitor() {
 
 	go func() {
 		defer func() {
-			for {
-				// Bleeds data from the websocket connection if needed
-				select {
-				case <-w.DataHandler:
-				default:
-					w.setDataMonitorRunning(false)
-					w.Wg.Done()
-					return
-				}
-			}
+			w.setDataMonitorRunning(false)
+			w.Wg.Done()
 		}()
-
+		full := 0
 		for {
 			select {
 			case <-w.ShutdownC:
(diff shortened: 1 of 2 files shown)


gbjk:
Rebased on master now that dependency is merged.

Ready to review. Super simple change, despite the verbose explanation.

It'd be nice to keep the commit message on merge if possible 🙏

shazbert (approved):
Nice stuff tACK.

exchanges/stream/websocket.go
```diff
@@ -347,34 +347,24 @@ func (w *Websocket) dataMonitor() {
...
-				default:
-					w.setDataMonitorRunning(false)
-					w.Wg.Done()
-					return
-				}
-			}
-		}()
-
+		defer w.setDataMonitorRunning(false)
+		defer w.Wg.Done()
```
shazbert, line 351: `Done()` will execute before `setDataMonitorRunning()` might want to swap these just to be on the safe side.

shazbert (changes requested):
One thing I noticed on latest changes. Others looks good. 👍

exchanges/stream/websocket.go
gbjk: Good point. I doubt that it'd race, but I'm not sure. Best to be safe.

Fixed 61ac0e5e

exchanges/stream/websocket.go
```diff
@@ -347,34 +347,24 @@ func (w *Websocket) dataMonitor() {
...
 			select {
 			case <-w.ShutdownC:
 				return
 			case d := <-w.DataHandler:
 				select {
 				case w.ToRoutine <- d:
-				case <-w.ShutdownC:
-					return
+					if full {
+						full = false
```
gloriousCode, line 361: Part of me desires knowing just how many messages are not being processed due blockage. `verbose` does let me see ws messages, but this does not let me know when processing is back on. Would you consider a `if verbose` log to say its back?

gloriousCode (changes requested):
Thank you for working with me yesterday to address things. One minor request which is up for debate 😄

exchanges/stream/websocket.go
gbjk: I did consider it, yes!
My issue with that is not wanting to do more processing than necessary for every single message going through.

What I had instead was a dropped counter, and a verbose check about it coming back.

Fixed 34bf736d

Note: Haven't had time to test this _at all_

shazbert (approved):
This is pretty much at a mergeable state, optional nits.  chACK.

gloriousCode (approved):
Thanks for the changes. I'm happy with the test fixes, the logs and the improvements to the handling. It further highlights how the connections don't need to be intertwined, but that is for another day
tACK

exchanges/stream/websocket.go
```diff
@@ -347,35 +347,26 @@ func (w *Websocket) dataMonitor() {
...
-				default:
-					w.setDataMonitorRunning(false)
-					w.Wg.Done()
-					return
-				}
-			}
-		}()
-
+		defer w.Wg.Done()
+		defer w.setDataMonitorRunning(false)
```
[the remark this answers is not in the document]
gbjk: I considered that too, so let's do it.

Fixed dd4c0d52

exchanges/stream/websocket.go
```diff
@@ -347,35 +347,26 @@ func (w *Websocket) dataMonitor() {
...
-					log.Warnf(log.WebsocketMgr, "%s exchange backlog in websocket processing detected", w.exchangeName)
-					select {
-					case w.ToRoutine <- d:
-					case <-w.ShutdownC:
-						return
+					if dropped == 0 {
+						// If this becomes prone to flapping we could drain the buffer, but that's extreme and we'd like to avoid it if possible
+						log.Warnf(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer full; dropping messages", w.exchangeName)
 					}
+					dropped++
```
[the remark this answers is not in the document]
gbjk: Understand where you're coming from, but I think starting to co-mingle modular code responsibilties when linkage between them fails is a very dark path.
We'd be better later ensuring that `websocketDataReceiver` is highly *unlikely* to block reading using goros or something. Anyway, I think that's definitely another day.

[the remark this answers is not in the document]
shazbert: 👍

shazbert (approved):
chACK.

thrasher- (approved):
Like the extra verbosity and thanks for tracking this issue down!

PR #1358, opened 2023-10-01
Kraken: Subscription improvements

gbjk:
* Move UpdateTradablePairs to a test
    _This should not happen in TestMain just because it's a requirement for
    most tests. It should just be a non-parallel early test.
    That allows unit tests to run quickly without calling it, and allows
    proper clear failures_
* Separate auth and non-auth ws tests
* Move SeedAssets from Setup to Start
   _Having SeedAssets in Setup is cruel and unusual because it calls the
    API. Most other interactive data seeding happens in Start.
    This made it so that fixing and creating unit tests for Kraken was
    painfully slow, particularly on flaky internet._
* Use Websocket subscriptionChannels instead of local slice
* Remove ChannelID - Deprecated in docs
   _I went both ways on this N+ times. ChannelID is great, but the bottom line is that it's deprecated, and not available in auth channels. By moving away from it we futureproof, and there's no downside really. Note that with or without_
* Simplify ping handlers and hardcodes message
* Add Depth as configurable orderbook channel param
* Simplify auth/non-auth channel updates

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Improvement (non-functional improvements)

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

PR #1484, opened 2024-02-19
Common: Fix fmt msg lost in AppendError().Error()

gbjk:
Retain the .msg field of a go fmt.Errorf .msg field returned by .Error() when wrapping multiple errors.
This fixes a situation where a nested stack of errors would lose formatting information, which is often used to supply identifying context.
e.g.
```
err = fmt.Errorf("%w `%s`: %w", errParsingField, fieldName,
parsingError)
```

This isn't really an issue with our implementation; Calling Unwrap() on a fmt.Errorf() which returns a wrapErrors will lose that formatting. Our issue was that we were using just Unwrap() to bind together our chain-of-custody.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -463,9 +463,20 @@ func InArray(val, array interface{}) (exists bool, index int) {
 	return
 }
 
+// fmtError holds a formatted msg and the errors which formatted it
+type fmtError struct {
+	errs []error
+	msg  string
+}
+
 // multiError holds errors as a slice
 type multiError struct {
 	errs []error
+}
+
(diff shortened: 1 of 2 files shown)


gbjk:
Note: This commit is from #1358 and has been reviewed already by everyone

shazbert (approved):
tACK.

thrasher- (approved):
ACK

gloriousCode (approved):
reTACK?!

PR #1495, opened 2024-03-08
Coinut: Fix websocket not parsing instruments, improve nonce matching and error handling

gbjk:
WsConnect is calling GetInstruments, and when that fails, erroring out and not subscribing to anything.

The response to get instruments is an object {}, which skips the IncomingWithData check in wsReadData for arrays []. The check in wsHandleData depended on client_ord_id, but I figure so long as there's a nonce which matches, we can palm it off.

This results simutaneously in having to move the login handler back to it's waiting nonce-parser, and also simplifying/deduping it.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/coinut/coinut_websocket.go
+++ b/exchanges/coinut/coinut_websocket.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/buger/jsonparser"
 	"github.com/gorilla/websocket"
 	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/common/crypto"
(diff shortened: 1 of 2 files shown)


exchanges/coinut/coinut_websocket.go
```diff
@@ -61,7 +62,7 @@ func (c *COINUT) WsConnect() error {
 	err = c.wsAuthenticate(context.TODO())
```
gloriousCode, line 62:
```go
if c.IsWebsocketAuthenticationSupported() {
		err = c.wsAuthenticate(context.TODO())
		if err != nil {
			c.Websocket.SetCanUseAuthenticatedEndpoints(false)
			log.Errorln(log.WebsocketMgr, c.Name+" "+err.Error())
		}
	}
```
May as well while you're here 😄

gloriousCode (approved):
tACK! Replicated the issue and appreciated the fixes

exchanges/coinut/coinut_websocket.go
gbjk: Sure.

Fixed bb9cd142

shazbert (approved):
Don't have keys, tested locally and public websocket service is still working. Changes look good.

thrasher- (approved):
tACK, thanks!

PR #1502, opened 2024-03-11
Websocket: Simplify the connectionMonitor

Beadko:
# PR Description

Currently, the WebSocket connectionMonitor is adding concurrency within the function. According to [google styleguide](https://google.github.io/styleguide/go/decisions#synchronous-functions), the caller should add the concurrency to the function.

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestConnectionMonitor

## Checklist

### DIFF
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -526,7 +526,7 @@ func (w *Websocket) trafficMonitor() {
 			t.Stop()
 			w.Wg.Done()
 			return
-		// this case ensures the timer is reset as close to when the traffic is received
+		// This case ensures the timer is reset as close to when the traffic is received as possible
 		case <-time.After(trafficCheckInterval):
 			select {
 			case <-w.TrafficAlert:
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -291,7 +291,8 @@ func (w *Websocket) Connect() error {
 		return fmt.Errorf("%v Error connecting %w", w.exchangeName, err)
 	}
 	w.setState(connected)
-	if !w.connectionMonitorRunning.CompareAndSwap(false, true) {
+
+	if w.connectionMonitorRunning.CompareAndSwap(false, true) {
 		go w.connectionMonitor()
 	}
 
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -335,39 +335,29 @@ func (w *Websocket) Enable() error {
 
 // dataMonitor monitors job throughput and logs if there is a back log of data
 func (w *Websocket) dataMonitor() {
-	if w.IsDataMonitorRunning() {
-		return
-	}
-	w.setDataMonitorRunning(true)
 	w.Wg.Add(1)
-
-	go func() {
-		defer func() {
-			w.setDataMonitorRunning(false)
-			w.Wg.Done()
-		}()
-		dropped := 0
-		for {
+	defer w.Wg.Done()
+	dropped := 0
+	for {
+		select {
+		case <-w.ShutdownC:
+			return
+		case d := <-w.DataHandler:
 			select {
-			case <-w.ShutdownC:
-				return
-			case d := <-w.DataHandler:
-				select {
-				case w.ToRoutine <- d:
-					if dropped != 0 {
-						log.Infof(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer recovered; %d messages were dropped", w.exchangeName, dropped)
-						dropped = 0
-					}
-				default:
-					if dropped == 0 {
-						// If this becomes prone to flapping we could drain the buffer, but that's extreme and we'd like to avoid it if possible
-						log.Warnf(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer full; dropping messages", w.exchangeName)
-					}
-					dropped++
+			case w.ToRoutine <- d:
+				if dropped != 0 {
+					log.Infof(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer recovered; %d messages were dropped", w.exchangeName, dropped)
+					dropped = 0
 				}
+			default:
+				if dropped == 0 {
+					// If this becomes prone to flapping we could drain the buffer, but that's extreme and we'd like to avoid it if possible
+					log.Warnf(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer full; dropping messages", w.exchangeName)
+				}
+				dropped++
 			}
 		}
-	}()
+	}
 }
 
 // connectionMonitor ensures that the WS keeps connecting
@@ -617,15 +607,6 @@ func (w *Websocket) setTrafficMonitorRunning(b bool) {
 // IsTrafficMonitorRunning returns status of the traffic monitor
 func (w *Websocket) IsTrafficMonitorRunning() bool {
 	return w.trafficMonitorRunning.Load()
-}
-
-func (w *Websocket) setDataMonitorRunning(b bool) {
-	w.dataMonitorRunning.Store(b)
-}
-
-// IsDataMonitorRunning returns status of data monitor
-func (w *Websocket) IsDataMonitorRunning() bool {
-	return w.dataMonitorRunning.Load()
 }
 
 // CanUseAuthenticatedWebsocketForWrapper Handles a common check to
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -280,6 +280,7 @@ func (w *Websocket) Connect() error {
 	w.subscriptions = subscriptionMap{}
 	w.subscriptionMutex.Unlock()
 
+	w.Wg.Add(2)
 	go w.dataMonitor()
 	go w.trafficMonitor()
 	w.setState(connecting)
@@ -335,7 +336,6 @@ func (w *Websocket) Enable() error {
 
 // dataMonitor monitors job throughput and logs if there is a back log of data
 func (w *Websocket) dataMonitor() {
-	w.Wg.Add(1)
 	defer w.Wg.Done()
 	dropped := 0
 	for {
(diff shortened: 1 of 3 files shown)


exchanges/stream/websocket.go
```diff
@@ -376,63 +373,57 @@ func (w *Websocket) dataMonitor() {
 }
 
 // connectionMonitor ensures that the WS keeps connecting
-func (w *Websocket) connectionMonitor() error {
-	if w.checkAndSetMonitorRunning() {
```
shazbert, line 380: `checkAndSetMonitorRunning` method and `errAlreadyRunning` can be removed completely.

exchanges/stream/websocket.go
```diff
@@ -376,63 +373,57 @@ func (w *Websocket) dataMonitor() {
...
-		return errAlreadyRunning
-	}
+func (w *Websocket) connectionMonitor() {
+	w.setConnectionMonitorRunning(true)
 	delay := w.connectionMonitorDelay
-
-	go func() {
-		timer := time.NewTimer(delay)
-		for {
+	timer := time.NewTimer(delay)
```
shazbert, line 379: Optionally you can apply this diff which removes the methods on methods approach:
```diff
diff --git a/exchanges/stream/websocket.go b/exchanges/stream/websocket.go
index 954c0c4ad..511ca12c0 100644
--- a/exchanges/stream/websocket.go
+++ b/exchanges/stream/websocket.go
@@ -34,7 +34,6 @@ var (

 // Private websocket errors
 var (
-       errAlreadyRunning                       = errors.New("connection monitor is already running")
        errExchangeConfigIsNil                  = errors.New("exchange config is nil")
        errExchangeConfigEmpty                  = errors.New("exchange config is empty")
        errWebsocketIsNil                       = errors.New("websocket is nil")
@@ -292,7 +291,7 @@ func (w *Websocket) Connect() error {
        }
        w.setState(connected)

-       if !w.IsConnectionMonitorRunning() {
+       if !w.connectionMonitorRunning.Swap(true) {
                go w.connectionMonitor()
        }

@@ -374,7 +373,6 @@ func (w *Websocket) dataMonitor() {

 // connectionMonitor ensures that the WS keeps connecting
 func (w *Websocket) connectionMonitor() {
-       w.setConnectionMonitorRunning(true)
        delay := w.connectionMonitorDelay
        timer := time.NewTimer(delay)
        for {
@@ -394,7 +392,7 @@ func (w *Websocket) connectionMonitor() {
                                log.Debugf(log.WebsocketMgr, "%v websocket: connection monitor exiting", w.exchangeName)
                        }
                        timer.Stop()
-                       w.setConnectionMonitorRunning(false)
+                       w.connectionMonitorRunning.Store(false)
                        return
                }
                select {
@@ -628,19 +626,6 @@ func (w *Websocket) IsTrafficMonitorRunning() bool {
        return w.trafficMonitorRunning.Load()
 }

-func (w *Websocket) checkAndSetMonitorRunning() (alreadyRunning bool) {
-       return !w.connectionMonitorRunning.CompareAndSwap(false, true)
-}
-
-func (w *Websocket) setConnectionMonitorRunning(b bool) {
-       w.connectionMonitorRunning.Store(b)
-}
-
-// IsConnectionMonitorRunning returns status of connection monitor
-func (w *Websocket) IsConnectionMonitorRunning() bool {
-       return w.connectionMonitorRunning.Load()
-}
-
 func (w *Websocket) setDataMonitorRunning(b bool) {
        w.dataMonitorRunning.Store(b)
 }
diff --git a/exchanges/stream/websocket_test.go b/exchanges/stream/websocket_test.go
index 6f150359c..701f2a8d0 100644
--- a/exchanges/stream/websocket_test.go
+++ b/exchanges/stream/websocket_test.go
@@ -571,13 +571,14 @@ func TestConnectionMonitor(t *testing.T) {
        ws.exchangeName = "hello"
        ws.Wg = &sync.WaitGroup{}
        ws.setEnabled(true)
+       ws.connectionMonitorRunning.Store(true)
        go ws.connectionMonitor()
        require.EventuallyWithT(t, func(c *assert.CollectT) {
-               assert.True(c, ws.IsConnectionMonitorRunning(), "IsConnectionMonitorRunning should return true")
+               assert.True(c, ws.connectionMonitorRunning.Load(), "IsConnectionMonitorRunning should return true")
        }, 5*time.Second, 10*time.Millisecond, "ConnectionMonitor must be running")
        ws.setEnabled(false)
        require.EventuallyWithT(t, func(c *assert.CollectT) {
-               assert.False(c, ws.IsConnectionMonitorRunning(), "IsConnectionMonitorRunning should return false")
+               assert.False(c, ws.connectionMonitorRunning.Load(), "IsConnectionMonitorRunning should return false")
        }, 5*time.Second, 10*time.Millisecond, "ConnectionMonitor must not be running")
 }
```

shazbert (changes requested):
@Beadko Fine work. This is a much better approach. Just some nits and feedback.

exchanges/stream/websocket.go
gbjk: Actually, I don't think we need an atomic at all.
* The only thing that starts it is Connect
* Connect is protected by a full struct mutex
* And bools are inherently atomic if you're not using compareAndSwap anyway, not that this matters given ☝️ 

I'm also curious about the state of any time we find we run Connect, it gets this far, but connectionMonitor is still running.
To me that implies that it's the *last* connectionMonitor, we're in a race, and it's about to exit.
Prove me wrong?

shazbert: Yeah good point, we can probably remove it then we might be able to add it to the waitgroup?

gbjk: Yep, that works best.
* Remove all "IsRunning" flags and checks
* Add to waitgroup
* Double check if this is already how the partner in crime `dataMonitor` works. They should follow the same pattern

shazbert: Agreed 🦅 Nice.

exchanges/stream/websocket.go
Beadko: removed

exchanges/stream/websocket.go
Beadko: I have removed the flag for the connection monitor and will cover dataMonitor in a separate PR. Is that ok?

gbjk: I think this is really a coherent chunk of work: Simplify Websocket monitor routines.
So should be one PR.

shazbert (commented):
@Beadko Just waiting on these changes that was [laid out from gk](https://github.com/thrasher-corp/gocryptotrader/pull/1502/files#r1520784798)

shazbert (changes requested):
Thanks for the update, this is deadlocked by dataMonitor not being called in a go routine. Please update and retest. Once that is updated it then it looks good to go.

exchanges/stream/websocket.go
```diff
@@ -340,99 +335,76 @@ func (w *Websocket) Enable() error {
 
 // dataMonitor monitors job throughput and logs if there is a back log of data
 func (w *Websocket) dataMonitor() {
-	if w.IsDataMonitorRunning() {
-		return
-	}
-	w.setDataMonitorRunning(true)
 	w.Wg.Add(1)
```
shazbert, line 338: please remove this and add this to the calling function

exchanges/stream/websocket.go
```diff
@@ -545,59 +517,49 @@ func (w *Websocket) FlushChannels() error {
 // 1 slot buffer means that connection will only write to trafficAlert once per trafficCheckInterval to avoid read/write flood in high traffic
 // Otherwise we Shutdown the connection after trafficTimeout, unless it's connecting. connectionMonitor is responsible for Connecting again
 func (w *Websocket) trafficMonitor() {
-	if w.IsTrafficMonitorRunning() {
-		return
-	}
-	w.setTrafficMonitorRunning(true)
 	w.Wg.Add(1)
```
shazbert, line 520: And here as well, remove this and move to the calling function

exchanges/stream/websocket.go
```diff
@@ -340,99 +335,76 @@ func (w *Websocket) Enable() error {
...
+	}
 }
 
 // connectionMonitor ensures that the WS keeps connecting
-func (w *Websocket) connectionMonitor() error {
-	if w.checkAndSetMonitorRunning() {
-		return errAlreadyRunning
-	}
+func (w *Websocket) connectionMonitor() {
 	delay := w.connectionMonitorDelay
```
shazbert, line 365: remove variable and use `w.connectionMonitorDelay`  in `new timer` and `reset` directly

exchanges/stream/websocket.go
```diff
@@ -545,59 +517,49 @@ func (w *Websocket) FlushChannels() error {
...
+	for {
+		select {
+		case <-w.ShutdownC:
+			if w.verbose {
+				log.Debugf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown message received", w.exchangeName)
+			}
+			t.Stop()
+			w.Wg.Done()
+			return
+		case <-time.After(trafficCheckInterval):
```
shazbert, line 531: This case seems to just bleed the traffic alert buffer by 1 and continues, this seems to already be handled below by the timer, could probably remove this case altogether?

shazbert (changes requested):
Thanks for doing removing the deadlock. There some more things that need to be done and then it should be good to go. Well done.

There are some race issues in the [ci build](https://github.com/thrasher-corp/gocryptotrader/actions/runs/8962752632/job/24612055648?pr=1502#step:9:440)

exchanges/stream/websocket.go
gbjk: Need to add a comment to explain why it's here.
It's about the cadence.
If the timeout went off, and we reset the timer at that point, there's no relationship between when we got traffic and the *next* traffic timeout.
This case is to ensure that as close as possible we reset the timer when we get traffic to avoid giving double timeouts.
e.g. If we removed this check:
10 second timeout
Traffic arrives in first 1 second and not again after
9 seconds later we'd reset for another 10 seconds

shazbert (changes requested):
Thanks for the changes, there is a deadlock introduced in the tests. Then everything else looks good to go.

exchanges/stream/websocket.go
shazbert: Gotcha thanks!

shazbert (approved):
@Beadko Nice clean! Thanks. tACK.

thrasher-:
Closing this as this is no longer required with @shazbert's websocket changes. Thank you anyway!

PR #1503, opened 2024-03-13
Tests: Abstract UpdatePairsOnce

gbjk:
Abstracts updatePairsOnce for use in several branches including #1482

## Type of change

- [x] Non-functional improvement

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/internal/testing/exchange/exchange.go
+++ b/internal/testing/exchange/exchange.go
@@ -1,6 +1,7 @@
 package exchange
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"log"
@@ -141,7 +142,7 @@ func SetupWs(tb testing.TB, e exchange.IBotExchange) {
 	setupWsMutex.Lock()
 	defer setupWsMutex.Unlock()
 
-	if _, ok := setupWsOnce[e]; ok {
+	if setupWsOnce[e] {
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


internal/testing/exchange/exchange.go
```diff
@@ -157,3 +158,23 @@ func SetupWs(tb testing.TB, e exchange.IBotExchange) {
...
+var updatePairsOnce = make(map[exchange.IBotExchange]bool)
+
+// UpdatePairsOnce ensures pairs are only updated once in parallel tests
+func UpdatePairsOnce(tb testing.TB, e exchange.IBotExchange) {
+	tb.Helper()
+
+	updatePairsMutex.Lock()
+	defer updatePairsMutex.Unlock()
+
+	if _, ok := updatePairsOnce[e]; ok {
```
shazbert, line 172:
```suggestion
	if updatePairsOnce[e] {
```

gloriousCode (approved):
Kind of a fun chain now. Can't properly review #1482 since it already requires dependencies and contains those changes, the changes that I'd request here might be things fixed in #1482 and don't want to review #1482 until this is merged in 😄 

I can confirm that this works nicely though and will benefit other PRs, tACK

shazbert (changes requested):
one pedantic nitters, tested and looks good for merge.

internal/testing/exchange/exchange.go
gbjk: On the fence about this, since I prefer existence tests most of the time.
OTOH, it doesn't hurt and it's clearer and this isn't a hot path.

Fixed ed22da9c

shazbert (approved):
ch ACK. Thanks 🖖

thrasher- (approved):
Lovely shared helper func + cleanup ❤️

PR #1505, opened 2024-03-16
Kraken: Fix wsCancelOrders not erroring with order id

gbjk:
We were using the "cancel many" facility of the Kraken api. However since that doesn't actually report errors individually, it seems saner to just multiplex over it.
We were going to get N+ responses anyway. Might as well send N+ requests

One thing I'm unhappy about is the change of Key/Secret for this to work.
I really don't like that there's special values for the auth values which get silently ignored, such as Test, Mock, Key, Secret, etc.
So I've ended up using something adhoc, and putting a require in the test to ensure it's obvious if the auth-conn isn't connected.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] TestWsCancelOrders

### DIFF
--- a/exchanges/kraken/kraken_test.go
+++ b/exchanges/kraken/kraken_test.go
@@ -2,6 +2,7 @@ package kraken
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"log"
@@ -17,7 +18,6 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/common/convert"
 	"github.com/thrasher-corp/gocryptotrader/common/key"
-	"github.com/thrasher-corp/gocryptotrader/config"
 	"github.com/thrasher-corp/gocryptotrader/core"
 	"github.com/thrasher-corp/gocryptotrader/currency"
 	exchange "github.com/thrasher-corp/gocryptotrader/exchanges"
@@ -31,10 +31,11 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/exchanges/stream"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/subscription"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
+	testexch "github.com/thrasher-corp/gocryptotrader/internal/testing/exchange"
 	"github.com/thrasher-corp/gocryptotrader/portfolio/withdraw"
 )
 
-var k = &Kraken{}
+var k *Kraken
 var wsSetupRan bool
 
 // Please add your own APIkeys to do correct due diligence testing.
@@ -44,31 +45,21 @@ const (
 	canManipulateRealOrders = false
 )
 
-// TestSetup setup func
 func TestMain(m *testing.M) {
-	k.SetDefaults()
-	cfg := config.GetConfig()
-	err := cfg.LoadConfig("../../testdata/configtest.json", true)
-	if err != nil {
+	k = new(Kraken)
+	if err := testexch.TestInstance(k); err != nil {
 		log.Fatal(err)
 	}
-	krakenConfig, err := cfg.GetExchangeConfig("Kraken")
-	if err != nil {
-		log.Fatal(err)
-	}
-	krakenConfig.API.AuthenticatedSupport = true
-	krakenConfig.API.Credentials.Key = apiKey
-	krakenConfig.API.Credentials.Secret = apiSecret
-	k.Websocket = sharedtestvalues.NewTestWebsocket()
-	err = k.Setup(krakenConfig)
-	if err != nil {
-		log.Fatal(err)
-	}
-	err = k.UpdateTradablePairs(context.Background(), true)
-	if err != nil {
-		log.Fatal(err)
+	if apiKey != "" && apiSecret != "" {
+		k.API.AuthenticatedSupport = true
+		k.SetCredentials(apiKey, apiSecret, "", "", "", "")
 	}
 	os.Exit(m.Run())
+}
+
+func TestUpdateTradablePairs(t *testing.T) {
+	t.Parallel()
+	testexch.UpdatePairsOnce(t, k)
 }
 
 func TestGetCurrentServerTime(t *testing.T) {
@@ -139,6 +130,7 @@ func TestUpdateTicker(t *testing.T) {
 
 func TestUpdateTickers(t *testing.T) {
 	t.Parallel()
+	testexch.UpdatePairsOnce(t, k)
 	ap, err := k.GetAvailablePairs(asset.Spot)
 	require.NoError(t, err)
 	err = k.CurrencyPairs.StorePairs(asset.Spot, ap, true)
@@ -1253,7 +1245,9 @@ func TestGetWSToken(t *testing.T) {
 }
 
 func TestWsAddOrder(t *testing.T) {
-	setupWsTests(t)
+	t.Parallel()
+	sharedtestvalues.SkipTestIfCredentialsUnset(t, k, canManipulateRealOrders)
+	testexch.SetupWs(t, k)
 	_, err := k.wsAddOrder(&WsAddOrderRequest{
 		OrderType: order.Limit.Lower(),
 		OrderSide: order.Buy.Lower(),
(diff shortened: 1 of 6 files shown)


exchanges/kraken/kraken_test.go
```diff
@@ -1265,11 +1255,45 @@ func TestWsAddOrder(t *testing.T) {
...
+	if err != nil {
+		return err
+	}
+	return w.WriteMessage(websocket.TextMessage, respJSON)
+}
+
+func TestWsCancelOrders(t *testing.T) {
+	t.Parallel()
+
+	k := testexch.MockWsInstance[Kraken](t, curryWsMockUpgrader(t, mockWsCancelOrders)) //nolint:govet // Intentional shadow to avoid future copy/paste mistakes
```
gloriousCode, line 1286: Is there a way I've missed to set these to verbose? Not a change request, just something I want, but perhaps I've missed it.

exchanges/kraken/kraken_test.go
```diff
@@ -1265,11 +1255,45 @@ func TestWsAddOrder(t *testing.T) {
...
+	require.True(t, k.IsWebsocketAuthenticationSupported(), "WS must be authenticated")
+
+	err := k.wsCancelOrders([]string{"RABBIT", "BATFISH", "SQUIRREL", "CATFISH", "MOUSE"})
+	assert.ErrorIs(t, err, errCancellingOrder, "Should error cancelling order")
+	assert.ErrorContains(t, err, "BATFISH", "Should error containing txn id")
+	assert.ErrorContains(t, err, "CATFISH", "Should error containing txn id")
+	assert.ErrorContains(t, err, "[EOrder:Unknown order]", "Should error containing server error")
+
+	err = k.wsCancelOrders([]string{"RABBIT", "SQUIRREL", "MOUSE"})
+	assert.NoError(t, err, "Should not error without bad txn id")
```
gloriousCode, line 1296:
```suggestion
	assert.NoError(t, err, "Should not error with valid ids")
```

exchanges/kraken/kraken_test.go
```diff
@@ -1265,11 +1255,45 @@ func TestWsAddOrder(t *testing.T) {
...
-	if err := k.wsCancelOrders([]string{"1337"}); err != nil {
-		t.Error(err)
+func mockWsCancelOrders(msg []byte, w *websocket.Conn) error {
+	var req WsCancelOrderRequest
+	if err := json.Unmarshal(msg, &req); err != nil {
+		return err
+	}
+	if req.Event != krakenWsCancelOrder {
+		return w.WriteMessage(websocket.TextMessage, []byte(fmt.Sprintf(`{"event":"subscriptionStatus","reqid":%d,"status":"ok"}`, req.RequestID)))
+	}
```
gloriousCode, line 1265: This does not get hit, is there something you're trying to do?

testdata/configtest.json
```diff
@@ -1850,7 +1850,7 @@
         "credentialsValidator": {
           "requiresKey": true,
           "requiresSecret": true,
-          "requiresBase64DecodeSecret": true
+          "requiresBase64DecodeSecret": false
```
gloriousCode, line 1853: Why has this been changed?

internal/testing/exchange/exchange.go
```diff
@@ -103,29 +106,38 @@ func MockWSInstance[T any, PT interface {
 	}
 
 	b.Features.Subscriptions = []*subscription.Subscription{}
+	// Exchanges which don't support subscription conf; Can be removed when all exchanges support sub conf
+	b.Websocket.GenerateSubs = func() ([]subscription.Subscription, error) { return []subscription.Subscription{}, nil }
```
gloriousCode, line 110: Carry-over from something else? kraken and binance both have it set. Does not seem relevant

exchanges/kraken/kraken_test.go
```diff
@@ -44,33 +45,19 @@ const (
...
-	krakenConfig.API.Credentials.Secret = apiSecret
-	k.Websocket = sharedtestvalues.NewTestWebsocket()
-	err = k.Setup(krakenConfig)
-	if err != nil {
-		log.Fatal(err)
-	}
-	err = k.UpdateTradablePairs(context.Background(), true)
-	if err != nil {
+	k = new(Kraken)
+	if err := testexch.TestInstance(k); err != nil {
```
gloriousCode, line 50: Doing this and leaving empty api key strings above will need to be amended.
This seems related to this snippet:
> One thing I'm unhappy about is the change of Key/Secret for this to work.
I really don't like that there's special values for the auth values which get silently ignored, such as Test, Mock, Key, Secret, etc.
So I've ended up using something adhoc, and putting a require in the test to ensure it's obvious if the auth-conn isn't connected.

But using the old testmain, your `TestWsCancelOrders` test still functions. I'm a bit unsure specifically what you're referring to in that snippet anyway

gloriousCode (changes requested):
The cancel orders change works well 🎉

exchanges/kraken/kraken_test.go
gbjk: Nope, not currently. It's a pretty tight box.

I tend to switch on verbose in configtest.json and then just run one test.

exchanges/kraken/kraken_types.go
```diff
@@ -721,7 +721,7 @@ type WsCancelOrderRequest struct {
 type WsCancelOrderResponse struct {
 	Event        string `json:"event"`
 	Status       string `json:"status"`
-	ErrorMessage string `json:"errorMessage"`
+	ErrorMessage string `json:"errorMessage,omitempty"`
```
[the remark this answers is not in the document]
gbjk: Agreed. But 
If it's not needed then I'll drop this.
It probably was at one point and then I added the msg != "", as you say, and then I muggle-dropped-code-rot.
Probably wasn't using jsonparser at first.

Fixed d353410e

exchanges/kraken/kraken_test.go
gbjk: Thanks.
I mean ... it wasn't _wrong_.
I've literally been giggling at this for 20 straight seconds 😂 

Should it be like Surgeons General or Courts Martial?
```
-	assert.NoError(t, err, "Should not error without bad txn id")
+	assert.NoError(t, err, "Should not error with non-bad txns id")
```
Okay. I'm done. Promise.

Fixed 55e24547

exchanges/kraken/kraken_test.go
gbjk: Yeah, it *was* a protection against when subscriptions were firing.
But then I fixed that with internal/testing/exchange/exchange.go:108-110 and  ...
... dropped this common loot 🤷 

Fixed 596c9cce

testdata/configtest.json
gbjk: Also Change Rot.
I'd been struggling to not have AuthenticatedWebsocketApiSupport re-disabled on config loading.
So I'd been using a Key/Secret and setting it to true in configtest.json.
That had (obvious) ramifications, so I ended up deciding that all mock tests should get to auth anyway.

Fixed 029769f1

internal/testing/exchange/exchange.go
gbjk: We're disabling default subscriptions for both config sub exchanges and for hardcoded ones.
If we don't do this then any mock server would have to also handle the subscriptions.
And obviously GenerateSubscriptions and WsSubscribe are unit tested separately anyway.

If you mean that not both these methods of disabling the Subs are needed as it stands, then I definitely want to keep them, because this MockWsInstance should be available to work for any exchange immediately and not bear-trap any new users.

Does this help? 8cc728fd

gloriousCode: Sure! I don't think I absorbed what was going on

exchanges/kraken/kraken_test.go
gbjk: Thanks. Nope, this was just a mistake, the apiKeys should have been respected.
( _It would be passive-aggressive-AF to yank this mechanism without discussing further with team members who have said it's their current methodology_ )

Fixed 4f95ad49

gbjk:
I'm not liking those failures in kraken's `TestWsOpenOrders`; Sniffing it now

gbjk:
TestWsOpenOrders errors were unrelated. This is good to review.

gloriousCode (approved):

As an extra, [would you consider this diff](https://gist.github.com/gloriousCode/630e1e8257cb7c3eb659bd60aa2291c7)? Otherwise, I can just put on another PR. I just wanted to get rid of the func `setupWsTests` while I was here

gbjk:
Already in the [next upcoming branch](https://github.com/gbjk/gocryptotrader/tree/feature/kraken_ws_chans_v3).

I've got significant testing of Subscribe in there without mocking, so I'm on the fence about adding a mock for it if we don't have to ( perversely ).

Anyway, Let's touch on this when that PR lands.

exchanges/kraken/kraken_websocket.go
```diff
@@ -1383,43 +1333,60 @@ func (k *Kraken) wsAddOrder(request *WsAddOrderRequest) (string, error) {
...
 func (k *Kraken) wsCancelOrders(orderIDs []string) error {
+	wg := sync.WaitGroup{}
+	wg.Add(len(orderIDs))
+	errC := make(chan error, len(orderIDs))
+	for _, id := range orderIDs {
+		go func(id string) {
+			defer wg.Done()
+			if err := k.wsCancelOrder(id); err != nil {
+				errC <- err
+			}
```
shazbert, line 1347:
```suggestion
			errC <- k.wsCancelOrder(id)
```

shazbert (approved):
Good stuff. tACK.

exchanges/kraken/kraken_test.go
```diff
@@ -44,33 +45,23 @@ const (
...
 		log.Fatal(err)
 	}
-	err = k.UpdateTradablePairs(context.Background(), true)
-	if err != nil {
-		log.Fatal(err)
+	if apiKey != "" && apiSecret != "" {
+		k.API.AuthenticatedSupport = true
+		k.SetCredentials(apiKey, apiSecret, "", "", "", "")
 	}
```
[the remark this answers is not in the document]
gbjk: Actually, I'm leaving this alone completely because I'd like to significantly improve key management soon.
Someone else could move it into TestInstance in the meantime, but I know if I start touching it I won't be able to stop.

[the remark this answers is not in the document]
shazbert: 👍 😆

exchanges/kraken/kraken_websocket.go
gbjk: My first reaction:
> I think I'd prefer to avoid sending nils down the channel just to either ignore them at 1356 or have wasteful calls to  AppendError

Then I realised when I said "Wasteful calls to AppendError" I realised I must be being defensive, because I've recently argued for that.

And then I looked at it and realised this is a pattern I've done too many times. So:

Fixed 2ed96411

I'll write tests now for it, but wanted to sling it up for any fast rejections

shazbert: Already approved my man, whatever you want to do. 🤷 😄

shazbert (approved):
Changes are lovely. Thanks.

thrasher- (approved):
Nice work

gbjk:
Rebased and ready to roll 😄

PR #1501, opened 2024-03-11
Subscription Pairs and Encapsulation

gbjk:
* Move Subscriptions to their own encapsulated store, list, and package
This work starts the process of disentangling subscriptions from websocket, since a subscription is an exchange concept, and we want the collection of subscriptions separate from each websocket so we can start to separate auth, non-auth, assets, private and multiplexed websockets
* Move Subscription Pair to Pairs
For some exchanges we've immediately made full use of this.
For others we've ported over the existing one-pair method, and will revist it when we tackle subscription configuration
* Subscription methods all use pointers, and mutexes exist on the store and on the subscriptions to allow this
* Subscription configuration support for Bitmex and CoinbasePro
* Fixes and improvements to many exchanges' subscription logic as a driveby
* Linter: Disable testify.Len
* Tests: Fix WsAuth turned off by config checking
* Tests: TestFixtureToDataHandler preserve WS
* Websocket: Log actual sent message when Verbose
 
## Type of change

- [x] Non-backwards compatible feature

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

PR #1516, opened 2024-04-14
Kraken Subscription Improvements

gbjk:
* Add Websocket SendMessageReturnResponses
*  Fix convert.TimeFromUnixTimestampDecimal using local
* Separate auth and non-auth ws tests
* Move SeedAssets from Setup to Start
   _Having SeedAssets in Setup is cruel and unusual because it calls the
    API. Most other interactive data seeding happens in Start.
    This made it so that fixing and creating unit tests for Kraken was
    painfully slow, particularly on flaky internet._
* Use Websocket subscriptionChannels instead of local slice
* Remove ChannelID - Deprecated in docs
   _I went both ways on this N+ times. ChannelID is great, but the bottom line is that it's deprecated, and not available in auth channels. By moving away from it we futureproof, and there's no downside really. Note that with or without_
* Simplify ping handlers and hardcodes message
* Add Depth as configurable orderbook channel param
* Simplify auth/non-auth channel updates

Depends on #1501 and #1505

## Type of change

- [x] Improvement (non-functional improvements)

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


gbjk:
Replaced by #1587

PR #1522, opened 2024-04-17
Kraken: Fix GetLatestFundingRates requiring pair enabled

gbjk:
APIs like this should not require the pair to be enabled.
Fixes intermittent test failure on cmd/exchange_wrapper_standards:
```
exchange_wrapper_standards_test.go:227: Kraken Func 'GetLatestFundingRates' Error: 'pair not enabled FF-TEST-DELIM-solusd_240628'. Inputs: [context.Background {futures FF-TEST-DELIM-solusd_240628 true}].
```

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/cmd/exchange_wrapper_standards/exchange_wrapper_standards_test.go
+++ b/cmd/exchange_wrapper_standards/exchange_wrapper_standards_test.go
@@ -185,17 +185,15 @@ func executeExchangeWrapperTests(ctx context.Context, t *testing.T, exch exchang
 		var assetLen int
 		for y := 0; y < method.Type().NumIn(); y++ {
 			input := method.Type().In(y)
-			if input.AssignableTo(assetParam) ||
-				input.AssignableTo(orderSubmitParam) ||
-				input.AssignableTo(orderModifyParam) ||
-				input.AssignableTo(orderCancelParam) ||
-				input.AssignableTo(orderCancelsParam) ||
-				input.AssignableTo(pairKeySliceParam) ||
-				input.AssignableTo(getOrdersRequestParam) ||
-				input.AssignableTo(pairKeySliceParam) {
-				// this allows wrapper functions that support assets types
-				// to be tested with all supported assets
-				assetLen = len(assetParams) - 1
+			for _, t := range []reflect.Type{
+				assetParam, orderSubmitParam, orderModifyParam, orderCancelParam, orderCancelsParam, pairKeySliceParam, getOrdersRequestParam, latestRateRequest,
+			} {
+				if input.AssignableTo(t) {
+					// this allows wrapper functions that support assets types
+					// to be tested with all supported assets
+					assetLen = len(assetParams) - 1
+					break
+				}
 			}
 		}
 		tt := time.Now()
--- a/currency/manager_test.go
+++ b/currency/manager_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/common/convert"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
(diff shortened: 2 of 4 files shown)


gloriousCode:
>APIs like this should not require the pair to be enabled

Wrapper functions tend to respect enabled pairs over available pairs. This has had some back and forth over a long time. To that end, if you were to comply what you're saying, you are after a considerable change to many implementations across many exchanges.

So with current approach applied, it would initially appear as though `exchange_wrapper_standards` has an intermittent issue with enabling pairs. However, upon inspection its because I did not add a flag to test the endpoint against all asset types.

As for the reason it failed when it tested the correct asset type, it is because the developer (me) did the inappropriate implementation of not formatting a pair before calling `String()` on it, so `-TEST-DELIM-` remained, which is not an available pair within Kraken and so it fails the test.

So in that sense I've made a diff of the changes to apply for fixing `exchange_wrapper_standards` handling of `GetLatestFundingRates` and bow my head in shame for the double test bugs introduced

gbjk:
> if you were to comply what you're saying, you are after a considerable change to many implementations across many exchanges.

Agreed. This has come up a few times with @thrasher-.

The direction I'm following:
* API method calls which don't require stateful data (e.g. Not relying on a stored Ticker, Orderbook, etc) __MAY__ support non-enabled but available pairs
* If they get in our way in the meantime, we should change them, depending on cost
* But we're not guaranteeing API compatibility that all such API methods will support available pairs _yet_ and there's no active task to go change them all

So on this task, I'll add your fix, but keep the addition to allow FundingRates on Available pairs

cmd/exchange_wrapper_standards/exchange_wrapper_standards_test.go
```diff
@@ -185,17 +185,15 @@ func executeExchangeWrapperTests(ctx context.Context, t *testing.T, exch exchang
...
-				input.AssignableTo(orderModifyParam) ||
-				input.AssignableTo(orderCancelParam) ||
-				input.AssignableTo(orderCancelsParam) ||
-				input.AssignableTo(pairKeySliceParam) ||
-				input.AssignableTo(getOrdersRequestParam) ||
-				input.AssignableTo(pairKeySliceParam) {
-				// this allows wrapper functions that support assets types
-				// to be tested with all supported assets
-				assetLen = len(assetParams) - 1
+			for _, t := range []reflect.Type{
```
shazbert, line 188: 🔥

shazbert (approved):
good stuff tACK.

currency/manager_test.go
```diff
@@ -574,6 +575,41 @@ func TestFullStoreUnmarshalMarshal(t *testing.T) {
...
+
+	_, err = pm.IsPairEnabled(cp, asset.Item(1337))
+	assert.ErrorIs(t, err, asset.ErrNotSupported, "Should error when asset is not supported")
+
+	pm.Pairs[asset.PerpetualSwap] = &PairStore{}
+	_, err = pm.IsPairEnabled(cp, asset.PerpetualSwap)
+	assert.ErrorIs(t, err, ErrAssetIsNil, "Should error when store AssetEnabled is nil")
+
+	_, err = pm.IsPairEnabled(EMPTYPAIR, asset.PerpetualSwap)
+	assert.ErrorIs(t, err, ErrCurrencyPairEmpty, "Should error when currency pair is empty")
```
gloriousCode, line 610: I am going to assume a typo, or I'm not understanding why you're testing something else 😄 
```suggestion
	ok, err = pm.IsPairAvailable(NewPair(SAFE, MOONRISE), asset.Spot)
	require.NoError(t, err, "pm.IsPairAvailable must not error")
	assert.False(t, ok, "pm.IsPairAvailable should return correct value for an non-existent")

	ok, err = pm.IsPairAvailable(cp, asset.Futures)
	require.NoError(t, err, "pm.IsPairAvailable must not error")
	assert.False(t, ok, "pm.IsPairAvailable should return false for a disabled asset type")

	cp = NewPairWithDelimiter("XRP", "DOGE", "-")
	ok, err = pm.IsPairAvailable(cp, asset.Spot)
	require.NoError(t, err, "pm.IsPairAvailable must not error")
	assert.False(t, ok, "pm.IsPairAvailable should return false for non-existent pair")

	_, err = pm.IsPairAvailable(cp, asset.PerpetualSwap)
	assert.ErrorIs(t, err, ErrAssetNotFound, "Should error when asset is not found")

	_, err = pm.IsPairAvailable(cp, asset.Item(1337))
	assert.ErrorIs(t, err, asset.ErrNotSupported, "Should error when asset is not supported")

	pm.Pairs[asset.PerpetualSwap] = &PairStore{}
	_, err = pm.IsPairAvailable(cp, asset.PerpetualSwap)
	assert.ErrorIs(t, err, ErrAssetIsNil, "Should error when store AssetEnabled is nil")

	_, err = pm.IsPairAvailable(EMPTYPAIR, asset.PerpetualSwap)
	assert.ErrorIs(t, err, ErrCurrencyPairEmpty, "Should error when currency pair is empty")
```

gloriousCode (changes requested):
Thank you for listening to my TED Talk and implementing those changes. Also glad they didn't break all the other `GetLatestFundingRates` standards tests.

currency/manager_test.go
gbjk: Damn. Yeah. Thanks.
I went to abstract it into closure over a method and test either of them, and then backed away from the crazy glue. In my haste to escape the intoxication of relentless abstraction I forgot to actually do it *right* 🤦 

Fixed ba314737

gloriousCode (approved):
tACK!

thrasher- (approved):
Thanks for looking into this and 👍 on direction!

PR #1523, opened 2024-04-17
go: Fix darwin/arm64: toolchain not available version directive

gemscng:
# PR Description

go build will give following error.

```
go: download go1.22 for darwin/arm64: toolchain not available
```

## Type of change
Refer to this solution
https://github.com/golang/go/issues/65568#issuecomment-1954876836

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/thrasher-corp/gocryptotrader
 
-go 1.22
+go 1.22.0
 
 require (
 	github.com/buger/jsonparser v1.1.1


thrasher- (approved):
Thank you @gemscng ! This was a fun read as well https://github.com/golang/go/issues/62278#issuecomment-1698829945

gloriousCode (approved):
1.Thanks.0 gemscng!

shazbert (approved):
good stuff thanks for the PR!

PR #1534, opened 2024-05-05
Tests: Various race fixes and move TestFixtureToDataHandler

gbjk:
* Fix a instance pair manager race intermittently causing Bitfinex and Kraken to fail
* Fix PairsManager.Load breaking matcher
* Fix UpdatePairsOnce not working across instances
* Move and Simplify FixtureToDataHandler
* Rename PairStore.copy to PairStore.clone and replace various builtin `copy` with slices.Clone
* Add Fullstore.clone and PairFormat.clone

## Type of change

- [x] Test fixes and improvements

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/currency/manager.go
+++ b/currency/manager.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/thrasher-corp/gocryptotrader/common"
@@ -65,10 +66,9 @@ func (p *PairsManager) Get(a asset.Item) (*PairStore, error) {
 	defer p.mutex.RUnlock()
 	c, ok := p.Pairs[a]
 	if !ok {
-		return nil,
-			fmt.Errorf("cannot get pair store, %v %w", a, asset.ErrNotSupported)
+		return nil, fmt.Errorf("cannot get pair store, %v %w", a, asset.ErrNotSupported)
 	}
-	return c.copy()
+	return c.clone(), nil
 }
 
 // Match returns a currency pair based on the supplied symbol and asset type
@@ -94,24 +94,16 @@ func (p *PairsManager) Store(a asset.Item, ps *PairStore) error {
 	if !a.IsValid() {
 		return fmt.Errorf("%s %w", a, asset.ErrNotSupported)
 	}
-	cpy, err := ps.copy()
-	if err != nil {
-		return err
+	if ps == nil {
+		return errPairStoreIsNil
 	}
 	p.mutex.Lock()
+	defer p.mutex.Unlock()
 	if p.Pairs == nil {
-		p.Pairs = make(map[asset.Item]*PairStore)
+		p.Pairs = FullStore{}
 	}
-	p.Pairs[a] = cpy
-	if p.matcher == nil {
-		p.matcher = make(map[key]*Pair)
-	}
-	for x := range cpy.Available {
-		p.matcher[key{
-			Symbol: cpy.Available[x].Base.Lower().String() + cpy.Available[x].Quote.Lower().String(),
-			Asset:  a}] = &cpy.Available[x]
-	}
-	p.mutex.Unlock()
+	p.Pairs[a] = ps.clone()
+	p.reindex()
 	return nil
 }
 
@@ -145,9 +137,7 @@ func (p *PairsManager) GetPairs(a asset.Item, enabled bool) (Pairs, error) {
 	}
 
 	if !enabled {
-		availPairs := make(Pairs, len(pairStore.Available))
-		copy(availPairs, pairStore.Available)
-		return availPairs, nil
+		return slices.Clone(pairStore.Available), nil
 	}
 
 	lenCheck := len(pairStore.Enabled)
@@ -157,8 +147,7 @@ func (p *PairsManager) GetPairs(a asset.Item, enabled bool) (Pairs, error) {
 
 	// NOTE: enabledPairs is declared before the next check for comparison
 	// reasons within exchange update pairs functionality.
-	enabledPairs := make(Pairs, lenCheck)
-	copy(enabledPairs, pairStore.Enabled)
+	enabledPairs := slices.Clone(pairStore.Enabled)
 
 	err := pairStore.Available.ContainsAll(pairStore.Enabled, true)
 	if err != nil {
@@ -176,8 +165,6 @@ func (p *PairsManager) StoreFormat(a asset.Item, pFmt *PairFormat, config bool) 
 		return errPairFormatIsNil
 	}
 
-	cpy := *pFmt
-
 	p.mutex.Lock()
 	defer p.mutex.Unlock()
 
@@ -192,9 +179,9 @@ func (p *PairsManager) StoreFormat(a asset.Item, pFmt *PairFormat, config bool) 
 	}
 
 	if config {
-		pairStore.ConfigFormat = &cpy
+		pairStore.ConfigFormat = pFmt.clone()
 	} else {
-		pairStore.RequestFormat = &cpy
+		pairStore.RequestFormat = pFmt.clone()
 	}
 	return nil
 }
@@ -205,11 +192,6 @@ func (p *PairsManager) StorePairs(a asset.Item, pairs Pairs, enabled bool) error
 	if !a.IsValid() {
 		return fmt.Errorf("%s %w", a, asset.ErrNotSupported)
 	}
-
-	// NOTE: Length check not needed in this scenario as it has the ability to
-	// remove the entire stored list if needed.
-	cpy := make(Pairs, len(pairs))
-	copy(cpy, pairs)
 
 	p.mutex.Lock()
 	defer p.mutex.Unlock()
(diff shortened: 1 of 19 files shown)


currency/manager.go
```diff
@@ -208,8 +205,7 @@ func (p *PairsManager) StorePairs(a asset.Item, pairs Pairs, enabled bool) error
 
 	// NOTE: Length check not needed in this scenario as it has the ability to
 	// remove the entire stored list if needed.
-	cpy := make(Pairs, len(pairs))
-	copy(cpy, pairs)
+	cpy := slices.Clone(pairs)
```
shazbert, line 208: suggestion: can rm the var and comment and inline to assignment.

currency/pair.go
```diff
@@ -118,6 +118,16 @@ func (f PairFormat) Format(pair Pair) string {
...
 }
 
+// clone returns a clone of the PairFormat
+func (f *PairFormat) clone() *PairFormat {
+	if f == nil {
+		return nil
+	}
+	c := &PairFormat{}
+	*c = *f
+	return c
```
shazbert, line 128:
```suggestion
	c := *f
	return &c
```

currency/manager.go
```diff
@@ -435,34 +431,27 @@ func (p *PairsManager) SetAssetEnabled(a asset.Item, enabled bool) error {
...
+	p.BypassConfigFormatUpgrades = seed.BypassConfigFormatUpgrades
+	p.UseGlobalFormat = seed.UseGlobalFormat
+	p.LastUpdated = seed.LastUpdated
+	p.Pairs = seed.Pairs.clone()
+	p.RequestFormat = seed.RequestFormat.clone()
+	p.ConfigFormat = seed.ConfigFormat.clone()
+	p.matcher = map[key]*Pair{}
+
+	// Re-index matcher
+	for a, fs := range p.Pairs {
```
shazbert, line 449: Suggestion: This is done a few times could make a method for it.

currency/manager.go
```diff
@@ -539,21 +528,18 @@ func (fs FullStore) MarshalJSON() ([]byte, error) {
...
 		assetEnabled = convert.BoolPtr(*ps.AssetEnabled)
 	}
 
-	enabled := make(Pairs, len(ps.Enabled))
-	copy(enabled, ps.Enabled)
-
-	avail := make(Pairs, len(ps.Available))
-	copy(avail, ps.Available)
+	enabled := slices.Clone(ps.Enabled)
+	avail := slices.Clone(ps.Available)
```
shazbert, line 542: Suggestion: can remove these declarations and use them in the Pairs Store allocation and below you can remove rFmt and cFmt and just use the methods you made.

shazbert (changes requested):
Looks good to me, only some pedantic things, that you can resolve at your choosing. Nice.

currency/manager.go
```diff
@@ -157,8 +155,7 @@ func (p *PairsManager) GetPairs(a asset.Item, enabled bool) (Pairs, error) {
 
 	// NOTE: enabledPairs is declared before the next check for comparison
 	// reasons within exchange update pairs functionality.
-	enabledPairs := make(Pairs, lenCheck)
-	copy(enabledPairs, pairStore.Enabled)
+	enabledPairs := slices.Clone(pairStore.Enabled)
 
 	err := pairStore.Available.ContainsAll(pairStore.Enabled, true)
```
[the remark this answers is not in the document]
gbjk: Fixed 23773fd

currency/manager.go
gbjk: Yep. In doing so I'm going to remove the comment, which needed rewording at best, but feels implicit to me.

Fixed ef4e43c

currency/pair.go
gbjk: 🤦 
Fixed d6480c1

currency/manager.go
gbjk: This exposed that the matcher index was never flushed when storing pairs.
We have three options:
1. Blow away the whole index each time regardless of asset updated
2. Restructure the key to make it fast to blow away just one asset.
3. Walk the current map and delete by asset first

Since store pair is relatively infrequent, I've opted to blow away the whole index each time, for simplicity.

Fixed 9005629

shazbert: 🔥 Nice 👍

currency/manager.go
gbjk: Fixed 30361ff1

shazbert (changes requested):
Just some 🏃 racey features left over [here](https://github.com/thrasher-corp/gocryptotrader/actions/runs/8962919898/job/24612474164#step:9:511)

gbjk:
Fixed with 2e9e7d67 and 1481244a
In both cases I've shied away from going and changing everything to remove the (very) thin wrappers around CurrencyManager methods
Also haven't gone hunting for other examples of intrusion (yet)

currency/manager.go
```diff
@@ -186,6 +177,43 @@ func (p *PairsManager) StoreFormat(a asset.Item, pFmt *PairFormat, config bool)
 	return nil
 }
 
+// GetFormat returns the pair format in a concurrent safe manner
+func (p *PairsManager) GetFormat(a asset.Item, r bool) (PairFormat, error) {
```
shazbert, line 181: Can you expand the `r` name to something else a little easier for somebody to understand? This also needs a test.
 
suggestion: Also can consolidate this down a little bit, example below, it calls p.Get just to verify asset in every case, would need to change the error to something more generic for this to work:

```go
	ps, err := p.Get(a)
	if err != nil {
		return EMPTYFORMAT, err
	}
	var outbound *PairFormat
	if p.UseGlobalFormat {
		outbound = p.ConfigFormat
		if r {
			outbound = p.RequestFormat
		}
	} else {
		outbound = ps.ConfigFormat
		if r {
			outbound = ps.RequestFormat
		}
	}
	if outbound == nil {
		return EMPTYFORMAT, fmt.Errorf("global: %v request %v: %w", p.UseGlobalFormat, r, ErrAssetConfigFormatIsNil)
	}
	return *outbound, nil

```

exchanges/gateio/gateio.go
```diff
@@ -3753,8 +3753,12 @@ func (g *Gateio) IsValidPairString(currencyPair string) bool {
 	if len(currencyPair) < 3 {
 		return false
 	}
-	if strings.Contains(currencyPair, g.CurrencyPairs.RequestFormat.Delimiter) {
-		result := strings.Split(currencyPair, g.CurrencyPairs.RequestFormat.Delimiter)
+	pf, err := g.CurrencyPairs.GetFormat(asset.Spot, true)
+	if err != nil {
+		return false
```
shazbert, line 3758: I am just pointing out that something ungodly is going down if this is erroring, so just a _suggestion_ to return an error along with the bool. But up to you.

shazbert (changes requested):
Good stuff, thanks for addressing my change requests. There's a couple of error on the [CI build](https://github.com/thrasher-corp/gocryptotrader/actions/runs/8965928427/job/24620397195?pr=1534#step:9:341)

currency/manager.go
gbjk: 💯 can do
I had the same thoughts myself, started to do exactly this, but then thought I should rein myself in.
You've given me carte blanche so I'll have at it.
I wanted to simplify the error, as well. Seemed like overkill.

gbjk: Fixed 907d016

Thanks for the suggested simplification.
I ensured that we don't try to get the asset if we're using Global, and I kept a much simpler (existing) error for nil, which should be an edge case anyway.

exchanges/gateio/gateio.go
gbjk: Thought about it.
But I think that shouldn't be the responsibility of this function, and I want to keep `IsXYZ` functions slim.
According to the data we have, this is definitely *not* a valid string 😭 😂

currency/manager_test.go
```diff
@@ -857,3 +857,60 @@ func TestPairManagerSetDelimitersFromConfig(t *testing.T) {
...
+
+// TestGetFormat exercises PairsManager GetFormat
+func TestGetFormat(t *testing.T) {
+	t.Parallel()
+
+	m := PairsManager{
+		UseGlobalFormat: true,
+		ConfigFormat: &PairFormat{
+			Uppercase: true,
+			Delimiter: "🦄",
```
shazbert, line 869: 🔥

shazbert (changes requested):
Changes look good 👍 can proceed with rebase.

gbjk:
@shazbert Following discussion with @thrasher- I've switched it back to using a new `PairsManager.IsAssetSupported`.

Fixed 93fee81

Everything else was a rebase of what you've already seen.

shazbert (approved):
Thanks for the changes. tACK! 🚀

shazbert (approved):
Changes ACK.

currency/manager.go
```diff
@@ -539,39 +538,30 @@ func (fs FullStore) MarshalJSON() ([]byte, error) {
 	return json.Marshal(temp)
 }
 
-// copy copies and segregates pair store from internal and external calls.
-func (ps *PairStore) copy() (*PairStore, error) {
+// clone returns a deep clone of the PairStore
+func (ps *PairStore) clone() *PairStore {
 	if ps == nil {
-		return nil, errPairStoreIsNil
+		return nil
```
gloriousCode, line 544: The removal of the error here and making returning nil an acceptable situation is something that could lead to panics as not much code is prepared for that. 

At present it looks like only a botched test via `Load` will lead to this which is acceptable and addressable by the offending user, but I still need to point out because 🦕

gloriousCode (approved):
tACK!

currency/manager.go
gbjk: Understood.

thrasher- (approved):
tACK, nice work! NOTE: BTSE issues are being worked on in a separate branch.

PR #1539, opened 2024-05-10
bugfixes: Backtester credentials, Binance ExecutionLimits, Mock recording

gloriousCode:
# PR Description
David on Slack has been informing us of issues with the Backtester. This fix uses up-to-date credential checking functions instead of returning an error. 

Before I could fix that I found [Binance changed the json for SPOT/MARGIN execution limits](https://binance-docs.github.io/apidocs/spot/en/#change-log) to use [[][]permissionSets](https://binance-docs.github.io/apidocs/spot/en/#exchange-information)

Upon fixing that I went to update our json mock data, and that showed that #1444 introduced a compatibility change with recording to the new path

I also came across duplicate errors _right next to eachother_ and have addressed that

## To test the Backtester
- Modify `/backtester/config/strategyexamples/dca-candles-live.strat`
- Change exchange to Binance and currency to BTC-USDT (or whatever)
- Add your credentials
- Set `real-orders` to `true`
- I would also encourage you to  set `maximum-total` to a tiny figure
- open cmd in `/backtester` execute `go run .`
- open another cmd in `/backtester/btcli/` execute `go run . executestrategyfromfile -p="config/strategyexamples/dca-candles-live.strat"`
- You should then get a success response instead of an error about credentials (if your credentials are correct ;)

Fixes # (issue)

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/backtester/engine/setup.go
+++ b/backtester/engine/setup.go
@@ -843,7 +843,7 @@ func (bt *BackTest) loadData(cfg *config.Config, exch gctexchange.IBotExchange, 
 				cfg.DataSettings.Interval)
 		}
 		err = bt.exchangeManager.Add(exch)
-		if err != nil {
+		if err != nil && !errors.Is(err, engine.ErrExchangeAlreadyLoaded) {
 			return nil, err
 		}
 		err = bt.LiveDataHandler.AppendDataSource(&liveDataSourceSetup{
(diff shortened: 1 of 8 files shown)


shazbert (approved):
Tested. ACK.

backtester/engine/setup.go
```diff
@@ -964,10 +965,13 @@ func setExchangeCredentials(cfg *config.Config, base *gctexchange.Base) error {
 			cfg.DataSettings.LiveData.ExchangeCredentials[i].Keys.PEMKey,
 			cfg.DataSettings.LiveData.ExchangeCredentials[i].Keys.OneTimePassword,
 		)
-		validated := base.AreCredentialsValid(context.TODO())
-		base.API.AuthenticatedSupport = validated
-		if !validated {
-			return fmt.Errorf("%v %w", base.GetName(), errInvalidCredentials)
+		creds, err := base.GetCredentials(context.TODO())
```
thrasher-, line 968:
```suggestion
		_, err := base.GetCredentials(context.TODO())
		if err != nil {
		          return err
		 }
```

And can drop the VerifyAPICredentials below since in this context `GetCredentials` will check them

thrasher- (changes requested):
Thanks for resolving this! Just one minor change and then this is a tACK:

image

thrasher- (approved):

PR #1556, opened 2024-05-29
currency/exchanges: Add bespoke exchange translator and pair matching helper

shazbert:
# PR Description

* Adds translator helpers for specific currency matching operations that are bespoke to each exchange. This can be overridden client side and implemented ad-hoc but a hardcoded pass when needed can be introduced as is done with Kukoin, bybit & GateIO. 

Thanks. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


currency/translation.go
```diff
@@ -20,3 +20,83 @@ var translations = map[*Item]Code{
...
+type PairsWithTranslation struct {
+	Pairs      Pairs
+	Translator Translator
+}
+
+// keyPair defines an immutable pair for lookup purposes
+type keyPair struct {
+	Base  *Item
+	Quote *Item
+}
```
gloriousCode, line 70: Add this to the key package as `key.Pair`

currency/translation.go
```diff
@@ -20,3 +20,83 @@ var translations = map[*Item]Code{
...
+
+// FindMatchingPairsBetween returns all pairs that match the incoming pairs.
+// Translator is used to convert exchange specific currency codes to common
+// currency codes used in lookup process. The pairs are not modified. So that
+// the original pairs are returned for deployment to the specific exchange.
+// NOTE: Translator is optional and can be nil. Translator can be obtained from
+// the exchange implementation by calling Base() method and accessing Features
+// and Translation fields.
+func FindMatchingPairsBetween(this, that PairsWithTranslation) map[Pair]Pair {
```
gloriousCode, line 79: Only clunk I can see with this is that it doesn't offer the ability to backtranslate. You need to call this function twice to create two maps to do so. eg you cannot do this without two calls `FindMatchingPairsBetween` with the params in opposite order XBT-USDCM -> BTC-USDC -> XBT-USDCM

Its not a required change, but I think its some user-friendliness to add both translation keys in there. Its only one thing to call and hold for a user wishing to translate between things

currency/translation_test.go
```diff
@@ -34,3 +36,97 @@ func TestGetTranslation(t *testing.T) {
...
+
+	translations := NewTranslations(map[Code]Code{
+		XBT:  BTC,
+		XETH: ETH,
+		XDG:  DOGE,
+		USDM: USD,
+		BTCM: BTC,
+	})
+
+	matchingPairs = FindMatchingPairsBetween(PairsWithTranslation{spotPairs, nil}, PairsWithTranslation{futuresPairs, translations})
```
gloriousCode, line 93: I feel this test could be improved upon by actually using the `matchingPairs` response to ensure that all expected pairs from `futuresPairs` is contained instead of a len check. Could also solidify your design of this translator by adding a scenario with delimiters to show that you can translate between different one, so then the test will fail if someone ever attempts to change that - stronger than a comment above the function

currency/translation_test.go
```diff
@@ -34,3 +36,97 @@ func TestGetTranslation(t *testing.T) {
...
+	}
+
+	matchingPairs := FindMatchingPairsBetween(PairsWithTranslation{spotPairs, nil}, PairsWithTranslation{futuresPairs, nil})
+	require.Len(t, matchingPairs, 1)
+
+	if !matchingPairs[ltcusd].Equal(ltcusd) {
+		t.Error("FindMatchingPairsBetween: matching pair not found")
+	}
+
+	translations := NewTranslations(map[Code]Code{
```
gloriousCode, line 85: you have declared a global with this name, rn [for annoying shadow](https://upload.wikimedia.org/wikipedia/en/4/41/ShadowTheHedgehogSA2.png)

gloriousCode (changes requested):
Cool concept

currency/translation.go
shazbert: I think that might be a cycle import issue but I will give it a crack.

gloriousCode: Oh yeah, that's okay then

currency/translation.go
shazbert: Yeah I can add that functionality and split `Translate` method to `TranslateToCommon` and `TranslateToExchange` or add a TODO and open another PR when it's actually being used in the library.

gloriousCode: I didn't realise this was waiting on my response. Any of those is fine

exchanges/bybit/bybit_wrapper.go
```diff
@@ -74,6 +74,32 @@ func (by *Bybit) SetDefaults() {
...
+				currency.NewCode("1000BEER"):        currency.NewCode("BEER"),
+				currency.NewCode("1000BONK"):        currency.BONK,
+				currency.NewCode("1000BTT"):         currency.BTT,
+				currency.NewCode("1000FLOKI"):       currency.FLOKI,
+				currency.NewCode("1000IQ50"):        currency.NewCode("IQ50"),
+				currency.NewCode("1000LUNC"):        currency.LUNC,
+				currency.NewCode("1000PEPE"):        currency.PEPE,
+				currency.NewCode("1000RATS"):        currency.NewCode("RATS"),
+				currency.NewCode("1000TURBO"):       currency.NewCode("TURBO"),
+				currency.NewCode("1000XEC"):         currency.XEC,
```
gloriousCode, line 98: Unlike XBT->BTC I don't consider 10000000AIDOGE to be the same as 1 AIDOGE. That seems more like this function is about returning the underlying currency of a futures contract currency, not translating it. If this is about returning underlying currencies I think things should be named more towards that as we already have existing language and features for that

gloriousCode (changes requested):

exchanges/bybit/bybit_wrapper.go
shazbert: I disagree the preceding number is just a price multiplier. You are still trading the same asset. It's not really about returning underlying currency as its converting a specific representation into its base currency. The primary goal of the function is to enable a more seamless matching between spot and futures or across exchanges.

Either way though, the current design allows a client to externally modify this list and if you still deem it a pinch point I will RM these translations and I will have them externally supported.

gloriousCode: I don't know why I came to that conclusion the other day because I've dealt with them being price multipliers myself 🤙 Fine to keep 🌈

gloriousCode (approved):
tACKeroo

currency/translation_test.go
```diff
@@ -34,3 +36,130 @@ func TestGetTranslation(t *testing.T) {
...
+	t.Parallel()
+	translationsTest := NewTranslations(map[Code]Code{
+		XBT:  BTC,
+		XETH: ETH,
+		XDG:  DOGE,
+		USDM: USD,
+	})
+	require.NotNil(t, translations)
+
+	if !translationsTest.Translate(XBT).Equal(BTC) {
```
thrasher-, line 50:
```suggestion
	assert.Equal(BTC, translationsTest.Translate(XBT), "Translate should return BTC")
	assert.Equal(LTC, translationsTest.Translate(LTC), "Translate should return LTC")
```

currency/translation_test.go
```diff
@@ -34,3 +36,130 @@ func TestGetTranslation(t *testing.T) {
...
+		NewPair(XETH, USDM).Format(PairFormat{Delimiter: "DELIMITER"}),
+		NewPair(XETH, BTCM),
+		ltcusd.Format(PairFormat{Delimiter: "DELIMITER"}), // exact match
+		NewPair(XRP, USDM), // no match
+	}
+
+	matchingPairs := FindMatchingPairsBetween(PairsWithTranslation{spotPairs, nil}, PairsWithTranslation{futuresPairs, nil})
+	require.Len(t, matchingPairs, 1)
+
+	if !matchingPairs[ltcusd].Equal(ltcusd) {
```
thrasher-, line 81: assert.Equal(ltcusd, matchingPairs[ltcusd], "Pairs should match")

thrasher- (changes requested):
Just some consistency issues with mix and matching between assert and Go's native tester

Also with this change, we could potentially rid the need for Kraken's asset translator

currency/translation_test.go
shazbert: [resolved](https://github.com/thrasher-corp/gocryptotrader/pull/1556/commits/59d61926939e3cbe3653e3414cf217582f3e9a47)

thrasher- (approved):
Nice stuff!

PR #1579, opened 2024-07-12
Kucoin: Add subscription templating and various fixes

gbjk:
### Kucoin
* Add subscription templating
* Support (undocumented) Batching for most websocket endpoints
* Remove support for deprecated margin funding book endpoint
* Remove unimplemented market snapshot by currency endpoint
 * Add subscription documentation, replacing in-situ boilerplate docs
 * Use ticker:all for more than 10 pairs, otherwise use specific pairs
 * Renames channel variables for consistency and clarity
 * Add Assets to existing sub configs
 * Consolidates Futures ticker and orderbook sub config into asset.All

 ### Subscription Templates
 * Add `BatchSize` support
 * Fix subscription Pairs not respected
 * `$.AssetPairs` only receives relevant assets for the subscription
 * Fixes and tests for separator trimming
 
 ### Common
 * Add `Batch` and `Sort` to common
 * Remove error from `pairs.Remove`
 * Add `common.SortStrings`
 * Variadic `pairs.Add` and `pairs.Remove`

## Stacked Dependencies
- [x] #1582 
- [x] #1583

## Type of change

- [x] Feature

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


common/common.go
```diff
@@ -653,3 +654,25 @@ func GetTypeAssertError(required string, received interface{}, fieldDescription
...
+func Batch[S ~[]E, E any](blobs S, batchSize int) (batches []S) {
+	var j int
+	for i := 0; i < len(blobs); i += batchSize {
+		j += batchSize
+		if j >= len(blobs) {
+			j = len(blobs)
+		}
+		batches = append(batches, blobs[i:j])
+	}
+	return
```
shazbert, line 668:
```suggestion
	if batchSize <= 1 {
		return
	}
	batches = make([]S, 0, (len(blobs)+batchSize-1)/batchSize)
	var j int
	for i := 0; i < len(blobs); i += batchSize {
		j += batchSize
		if j >= len(blobs) {
			j = len(blobs)
		}
		batches = append(batches, blobs[i:j])
	}
	return
```

Awesome function! Fixes panic and infinite loop issues and reduces allocations.

common/common_test.go
```diff
@@ -856,3 +856,23 @@ func TestErrorCollector(t *testing.T) {
...
+}
+
+type A int
+
+func (a A) String() string {
+	return strconv.Itoa(int(a))
+}
+
+func TestSortStrings(t *testing.T) {
+	assert.Equal(t, []A{1, 2, 5, 6}, SortStrings([]A{6, 2, 5, 1}))
```
shazbert, line 877: Just thinking out loud, not a change request cause it's doing exactly what it should be doing but numbers I don't think is a good idea to be sorted as a test if you put in a 10 (10, 2, 5, 1) it would output (1, 10, 2, 5) unless we had some string conversion handling as well?

currency/manager.go
```diff
@@ -289,11 +289,12 @@ func (p *PairsManager) DisablePair(a asset.Item, pair Pair) error {
 		return err
 	}
 
-	enabled, err := pairStore.Enabled.Remove(pair)
-	if err != nil {
-		return err
+	if !pairStore.Enabled.Contains(pair, true) {
```
shazbert, line 292: Suggestion: If you want to minimise walk
```go
enabledLen := len(pairStore.Enabled)
	pairStore.Enabled = pairStore.Enabled.Remove(pair)

	if enabledLen == len(pairStore.Enabled) {
		return fmt.Errorf("%w %s", ErrPairNotFound, pair)
	}
```

currency/pairs.go
```diff
@@ -199,28 +199,25 @@ func (p Pairs) GetPairsByCurrencies(currencies Currencies) Pairs {
...
+			n = append(n, pN)
 		}
 	}
-	return nil, fmt.Errorf("%s %w", pair, ErrPairNotFound)
+	return n
 }
 
-// Add adds a specified pair to the list of pairs if it doesn't exist
+// Add adds pairs to the list of pairs ignoring duplicates
 func (p Pairs) Add(pairs ...Pair) Pairs {
```
shazbert, line 214: Looking at benchmark this is indeed more performant but if p Pairs is pre allocated with a capacity greater that the incoming slice then you are modifiying the original slice. It's an edge case that was included in the previous tests.

currency/pairs.go
```diff
@@ -199,28 +199,25 @@ func (p Pairs) GetPairsByCurrencies(currencies Currencies) Pairs {
...
 
-// Remove removes the specified pair from the list of pairs if it exists
-func (p Pairs) Remove(pair Pair) (Pairs, error) {
-	pairs := slices.Clone(p)
-	for x := range p {
-		if p[x].Equal(pair) {
-			return append(pairs[:x], pairs[x+1:]...), nil
+// Remove removes the specified pairs from the list of pairs if they exist
+func (p Pairs) Remove(rem ...Pair) Pairs {
+	var n Pairs
```
shazbert, line 204: `n := make(Pairs, 0, len(p))`

exchanges/exchange.go
```diff
@@ -1823,19 +1820,14 @@ func (b *Base) ParallelChanOp(channels subscription.List, m func(subscription.Li
 		return errBatchSizeZero
 	}
 
-	var j int
-	for i := 0; i < len(channels); i += batchSize {
-		j += batchSize
-		if j >= len(channels) {
-			j = len(channels)
-		}
+	for _, b := range common.Batch(channels, batchSize) {
```
shazbert, line 1823: 🔥

exchanges/kucoin/kucoin_test.go
```diff
@@ -1999,6 +2001,20 @@ func verifySubs(tb testing.TB, subs subscription.List, a asset.Item, prefix stri
...
+
+func init() {
+	for _, pp := range [][]string{
+		{"BTC", "USDT", "-"}, {"ETH", "BTC", "-"}, {"ETH", "USDT", "-"}, {"LTC", "USDT", "-"}, // Spot
+		{"ETH", "BTC", "-"}, {"LTC", "USDT", "-"}, {"SOL", "USDC", "-"}, {"TRX", "BTC", "-"}, // Margin
+		{"ETH", "USDCM", ""}, {"SOL", "USDTM", ""}, {"XBT", "USDCM", ""}, // Futures
+	} {
+		pair := currency.NewPairWithDelimiter(pp[0], pp[1], pp[2])
+		pair.Delimiter = pp[2]
+		subPairs = append(subPairs, pair)
```
shazbert, line 2014:
```suggestion
		subPairs = append(subPairs, currency.NewPairWithDelimiter(pp[0], pp[1], pp[2]))
```

exchanges/kucoin/kucoin_test.go
```diff
@@ -1999,6 +2001,20 @@ func verifySubs(tb testing.TB, subs subscription.List, a asset.Item, prefix stri
 	}
 }
 
+var subPairs currency.Pairs
```
shazbert, line 2004: verify subs above seems unused and there are test failures in this file.

shazbert (changes requested):
Nice work, nothing really jumps out, just some things I noticed and there are some ci errors that need to be resolved. 🔥

common/common.go
gbjk: batchSize 1 should return a new Clone of the slice, right?
So 660 should be `<= 0`

shazbert: Yeah good point. 🙏

gbjk: Fixed bd970c0

Can Haz Tests.

common/common_test.go
gbjk: Also thinking out loud in reply:
I think that's a hypothetical that _if_ our test code input was different, then the test would fail.
But fixing it would mean changing `A`, and wouldn't change SortStrings implementation, and if it failed it'd be because someone changed the test.
So I reckon it's okay :)

currency/manager.go
gbjk: Huh. Actually fails the test 😮 
So, that's interesting.

currency/pairs.go
gbjk: I don't tend to pre-alloc if we don't know the size, but yeah, that's fine.
I can always `Shrink`

currency/manager.go
gbjk: Fixed 132650e

currency/pairs.go
gbjk: Fixed 99967da

currency/pairs.go
gbjk: I don't see evidence for this edge case.
Yes, the previous test proved that it wasn't modified.
But that doesn't prove it would be modified otherwise.
I added a test for this, used slices.Grow, and then removed the slices.Clone (or Clip) and my test still passed.
I forced the issue by changing the original to be sure.
Pass-by-value and then append does not modify the original.

HOWEVER: If you're meaning that we'll *return* a slice where modifying it would change the original, then agreed. But the existing tests don't check for that, right?

Anyway, I have tests and a Clone to cover all bases now.

shazbert: Yeah it's a bit of a weird one and hard to see I should have gotten @cranktakular to break it out a bit better. It was central to this block:

```go
	// Test whether the original pairs have been modified
	pairsWithExtraBaggage := make(Pairs, 0, len(p)+3)
	pairsWithExtraBaggage = append(pairsWithExtraBaggage, p...)
	brain := NewPair(BRAIN, USD)
	// Edge case: using preallocated pairsWithExtraBaggage as base for two independent slices
	// will leak changes across underlying arrays.
	withBrainPairs := pairsWithExtraBaggage.Add(NewPair(BTC, LTC), brain)
	fmt.Println("withBrainPairs", withBrainPairs)
	assert.Equal(t, brain, withBrainPairs[len(withBrainPairs)-1])
	badger := NewPair(BADGER, USD)
	withBadgerPairs := pairsWithExtraBaggage.Add(NewPair(BTC, LTC), badger)
	fmt.Println("withBadgerPairs", withBadgerPairs)
	fmt.Println("withBrainPairs second check", withBrainPairs)  // This will be modified with magic and contain badger
	fmt.Println("pairsWithExtraBaggage", pairsWithExtraBaggage) // This won't be modified, so its hard to test.
	assert.Equal(t, badger, withBadgerPairs[len(withBadgerPairs)-1])
	assert.Equal(t, brain, withBrainPairs[len(withBrainPairs)-1])
```

Adding this block and reverting the changes should pass the tests again.

gbjk: Fixed 7604877

gbjk: Also Fixed cc1e73b

All our issues are symptoms of underlying array not re-allocated.
But we'd also fall into an issue of changes within the arrays contaminating across siblings.
So the fix and the test is just to Clone() it and ensure it doesn't cross-contaminate.

exchanges/kucoin/kucoin_test.go
gbjk: Yeah. That's some first class code rot there.

Fixed 64ba0b3

exchanges/kucoin/kucoin_test.go
gbjk: Thanks

Fixed 54f927f

exchanges/kucoin/kucoin_types.go
```diff
@@ -956,19 +956,6 @@ type WsPriceIndicator struct {
 	Value       float64              `json:"value"`
 }
```
shazbert, line 958: remove error `errSubscriptionPairRequired` top of file please.

shazbert (changes requested):
Looking great just some minor niggles, tested websocket.

common/common.go
```diff
@@ -653,3 +654,24 @@ func GetTypeAssertError(required string, received interface{}, fieldDescription
 	}
 	return fmt.Errorf("%w from %T to %s%s", ErrTypeAssertFailure, received, required, description)
 }
+
+// Batch takes a slice type and converts it into a slice of slices
+func Batch[S ~[]E, E any](blobs S, batchSize int) []S {
+	if batchSize <= 0 {
```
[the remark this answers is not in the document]
gbjk: Fixed 1ee3c34e

common/common.go
```diff
@@ -653,3 +654,24 @@ func GetTypeAssertError(required string, received interface{}, fieldDescription
...
+		return []S{}
+	}
+	batches := make([]S, 0, (len(blobs)+batchSize-1)/batchSize)
+	for batchSize < len(blobs) {
+		blobs, batches = blobs[batchSize:], append(batches, blobs[0:batchSize:batchSize])
+	}
+	return append(batches, blobs)
```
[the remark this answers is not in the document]
gbjk: Could do. But I don't think we'll gain very much at all in performance for our troubles.
If anything.

common/common.go
```diff
@@ -653,3 +654,24 @@ func GetTypeAssertError(required string, received interface{}, fieldDescription
...
 }
+
+// Batch takes a slice type and converts it into a slice of slices
+func Batch[S ~[]E, E any](blobs S, batchSize int) []S {
+	if batchSize <= 0 {
+		return []S{}
+	}
+	batches := make([]S, 0, (len(blobs)+batchSize-1)/batchSize)
+	for batchSize < len(blobs) {
+		blobs, batches = blobs[batchSize:], append(batches, blobs[0:batchSize:batchSize])
```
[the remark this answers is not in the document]
gbjk: Fixed 4d1f17b1

Yes, I agree. I actually find it harder to read with 0 in there.
When slices omit low or high, I don't have to think about whether that means "starting after 0" or not.
It's intuitively and immediately "from the start".

Worth noting this is actually straight out of the [official go documentation](https://go.dev/wiki/SliceTricks#batching-with-minimal-allocation).

exchanges/kucoin/kucoin_types.go
gbjk: Fixed b090d967

shazbert (approved):
tACK! Nice work.

shazbert (approved):
Changes ACK.

currency/pairs.go
```diff
@@ -199,28 +199,26 @@ func (p Pairs) GetPairsByCurrencies(currencies Currencies) Pairs {
...
-// Add adds a specified pair to the list of pairs if it doesn't exist
+// Add adds pairs to the list of pairs ignoring duplicates
 func (p Pairs) Add(pairs ...Pair) Pairs {
-	merge := append(slices.Clone(p), pairs...)
-	var filterInt int
-	for x := len(p); x < len(merge); x++ {
-		if !merge[:len(p)+filterInt].Contains(merge[x], true) {
-			merge[len(p)+filterInt] = merge[x]
-			filterInt++
+	p = slices.Clone(p)
```
gloriousCode, line 215:
```suggestion
	c := slices.Clone(p)
```
Has no impact on allocs or performance. Even though the function doesn't have a pointer receiver, why allow the idea that you're doing anything with `p` when `p` isn't `p` and has no impact on `p`?
I know, I know, just `git gud`, but also removes two of three yellow squiggles

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -6,9 +6,11 @@ import (
 	"errors"
```
gloriousCode, line 6:
```
[ERROR] | WEBSOCKET | 29/07/2024 13:32:51 | Kucoin websocket: subscription failure, ticker all : subscription template containing PairSeparator with must contain either specific Asset or AssetSeparator, orderbook all : subscription template containing PairSeparator with must contain either specific Asset or AssetSeparator
```
I have the following pairs enabled:
```json
 "spot": {
      "assetEnabled": true,
      "enabled": "BTC-USDT,AVA-USDT,FET-BTC,FET-ETH,ANKR-BTC,XMR-BTC,XMR-ETH,MTV-BTC,MTV-ETH",
```

gloriousCode (changes requested):
Was looking into the 10+ symbols -> sub ticker:all and ran into the error, so I'll pause reviewing for now to get it looked at. The other two things aren't really much of an issue

exchanges/kucoin/kucoin_websocket.go
gbjk: @gloriousCode Yep, I could reproduce this before my recent backported fix, and it's gone with my latest fix.

Fixed 4782a10

common/common.go
```diff
@@ -653,3 +654,29 @@ func GetTypeAssertError(required string, received interface{}, fieldDescription
 	}
 	return fmt.Errorf("%w from %T to %s%s", ErrTypeAssertFailure, received, required, description)
 }
+
+// Batch takes a slice type and converts it into a slice of slices
+func Batch[S ~[]E, E any](blobs S, batchSize int) []S {
+	if batchSize <= 0 || len(blobs) == 0 {
+		return []S{}
```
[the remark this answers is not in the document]
gbjk: > but you've put a batchSize == 0 check for Kucoin
I didn't understand this.

It feels to me like if someone writes Batch(b, 0) then they wouldn't be any more correct to expect a batchSize of blobs than an empty resultset.
I'm leaning heavily towards leaving this as-it-is because returning a slice with an empty slice in it feels even more wrong.

currency/pairs.go
gbjk: My only argument against this, in general, is the bear trap:
When we switch to using `c` as suggested ( Or `n`-ew is what I'd normally do) we run the risk of people making future mistakes by using `p` instead of `c`.

What's "three yellow squiggles" you mention?

common/common.go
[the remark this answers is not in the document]
shazbert:
```go
output := Batch(inputSlice, maxAmountCanBeZero)
```

```go
output := []type{inputSlice}
if  maxAmountCannotBeZero > 0 {
output = Batch(inputSlice, maxAmountCannotBeZero )
}
```

I see both scenarios being valid. Look at dem lines saved though 😆.

currency/pairs.go
gloriousCode: I generally view non-pointer receivers as something you read from, since you can't modify the underlying, you are modifying a fake `p`, so your other foot gets in the bear trap, where you can think you are modifying `p` when you're not.
2 squiggles are `Assignment to the method receiver doesn't propagate to other calls ` and the other is the usual complaint about having struct functions with pointer and non-pointer receivers which go docs don't like, but I don't mind

common/common.go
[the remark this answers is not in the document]
gloriousCode: > I'm leaning heavily towards leaving this as-it-is because returning a slice with an empty slice in it feels even more wrong

For a scenario where len(blob) > 0 and batch is a var that happens to be zero, one gets an empty batch, must check the batch, then use blob instead. That's the only scenario I didn't like. I'm fine with it staying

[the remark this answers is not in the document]
gbjk: > where you need to put checking code around the function call like you do with Kucoin
Can you point to that code?

Cos I don't think I have any that does that.
The only thing we have is the (documented) switch that will make it use `ticker:all` if you have more than 10 pairs enabled.

currency/pairs.go
gbjk: Swayed 👍 

Fixed 10ce34b7

common/common.go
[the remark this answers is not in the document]
gbjk: TL;DR:
* Yes, `batchSize <= 0` should act like len(blobs) as suggested.
* *But* len(blobs) 0 should still produce an empty slice, not a slice with one empty slice in it

### Discussion
That really comes down to what you think a batch size of 0 should do.
A batch size of 1 means: Give me slices with 1 elements in each slice.
My first thought was: Therefore batch size of 0 should not swing completely the other direction and act as batchsize is len(blobs).
The logically correct behaviour feels like:
> You asked for batches with zero elements. Here's the empty slice you asked for.

But then I thought: Why would the batchSize be dynamic?
Most likely, because it's from config.
And if it's from config, what does an empty batch size mean?
( Or even a negative one, in muggle-minded people? )
It would mean "don't batch it", which supports the requested change.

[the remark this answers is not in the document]
gbjk: Fixed 85567a25

Added behavioural tests in case any of this regresses.
Clarified function comment, since it was not clear if batchSize is the size of the batches or the number of batches.

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -502,7 +465,30 @@ func (ku *Kucoin) processFuturesMarkPriceAndIndexPrice(respData []byte, instrume
...
 
-func (ku *Kucoin) processFuturesOrderbookLevel5(respData []byte, instrument string) error {
+// ensureFuturesOrderbookSnapshotLoaded makes sure an initial futures orderbook snapshot is loaded
+func (ku *Kucoin) ensureFuturesOrderbookSnapshotLoaded(symbol string) error {
+	fetchedFuturesOrderbookMutex.Lock()
+	defer fetchedFuturesOrderbookMutex.Unlock()
+	if fetchedFuturesOrderbook[symbol] {
+		return nil
+	}
+	fetchedFuturesOrderbook[symbol] = true
```
gloriousCode, line 475: Not huge, but shouldn't this be at the end in the event of an error in the 4 functions below? Its only ensuring that things run once per pair otherwise

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -1777,3 +1580,147 @@ func (ku *Kucoin) CalculateAssets(topic string, cp currency.Pair) ([]asset.Item,
...
+		if i, err := intervalToString(s.Interval); err == nil {
+			return i
+		}
+	}
+	return ""
+}
+
+// currencies returns the currencies from all pairs in an asset
+// Updates the AssetPairs map parameter to contain only those currencies as Base items
+func assetCurrencies(s *subscription.Subscription, ap map[asset.Item]currency.Pairs) currency.Currencies {
```
gloriousCode, line 1679: Up to you, but I think it would be nice for the commentary to say something about how `ap` is generated in `fillAssetPairs` `/subscriptions/list.go`. It doesn't have a direct line from there to here

Also `currencies`->`assetCurrencies`

exchanges/subscription/template.go
```diff
@@ -63,86 +68,121 @@ func (l List) ExpandTemplates(e iExchange) (List, error) {
...
+	}
+
+	subs := List{}
+	for i, assetChannels := range assetRecords {
+		a := assets[i]
+		pairs := subCtx.AssetPairs[a]
+
+		batchSize := len(pairs) // Default to all pairs in one batch
+		if b := strings.Split(assetChannels, subCtx.BatchSize); len(b) > 2 {
+			return nil, fmt.Errorf("%w for %s", errTooManyBatchSize, a)
```
gloriousCode, line 149: Would you mind either pm or here, expanding on what's happening? For the test scenario that hits here, there's an intentional addition of more batches, but then I don't understand the _why_ of the `Default to all pairs in one batch` and this part
You're using a hardcoded len of 2 here and below, why? Why not store that in a var with a more descriptive term? In the below section, you change if there is a batch indicator, but you only want one?

gloriousCode (changes requested):
Thanks for making those changes

common/common_test.go
```diff
@@ -856,3 +856,36 @@ func TestErrorCollector(t *testing.T) {
...
+	assert.Len(t, b[0], 3)
+	assert.Len(t, b[3], 1)
+
+	b[0][0] = 42
+	assert.Equal(t, 1, s[0], "Changing the batches must not change the source")
+
+	require.NotPanics(t, func() { Batch(s, -1) }, "Must not panic on negative batch size")
+	done := make(chan any, 1)
+	go func() { done <- Batch(s, 0) }()
+	require.Eventually(t, func() bool { return len(done) > 0 }, time.Second, time.Millisecond, "Batch 0 must not hang")
```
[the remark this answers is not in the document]
gbjk: Yes. Previously it would hang. This is a regression test.

See L860
```
// TestBatch ensures the Batch function does not regress into common behavioural faults if implementation changes
```

I struggled slightly because I was considering saying "Regression test" in 874 where you're asking this, but it felt off because no merge of the failure has happened. We're protecting against a regression that's within the scope of the PR.

common/common.go
```diff
@@ -653,3 +654,34 @@ func GetTypeAssertError(required string, received interface{}, fieldDescription
...
+		blobs, batches[i] = blobs[batchSize:], blobs[:batchSize:batchSize]
+		i++
+	}
+	if len(blobs) > 0 {
+		batches[i] = blobs
+	}
+	return batches
+}
+
+// SortStrings takes a slice of fmt.Stringer implementers and returns a new sorted slice
```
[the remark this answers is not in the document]
gbjk: What kind of monster would default to descending? 👀 

Fixed 13da5afa

common/common_test.go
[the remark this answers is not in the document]
gloriousCode: I complain if there's no commentary, but I also complain when I don't _read_ commentary. There is no winning for you

exchanges/kucoin/kucoin_test.go
```diff
@@ -1973,29 +1974,15 @@ func TestPushData(t *testing.T) {
...
-		suffix := strings.TrimPrefix(sub.Channel, prefix)
-		if len(expected) == 0 {
-			assert.Empty(tb, suffix, "Sub for asset %s with prefix %s should have no symbol suffix", a.String(), prefix)
-		} else {
-			currs := strings.Split(suffix, ",")
-			assert.ElementsMatch(tb, currs, expected, "Currencies should match in sub for asset %s with prefix %s", a.String(), prefix)
-		}
+var subPairs currency.Pairs
+
+func init() {
```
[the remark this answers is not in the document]
gbjk: 1. I use multiple init functions next to the thing they're acting upon because action close to usage is clearer. I think it's one of the go tenents, in fact. So have as many init() funcs as you have things that you want to init next to where they are used.

2. No reason to do this at all. It's reducible complexity; At one point subPairs was used elsewhere.
Thanks for spotting 🙏 

Fixed bc20106a

exchanges/kucoin/kucoin_test.go
```diff
@@ -2490,14 +2483,90 @@ func TestProcessMarketSnapshot(t *testing.T) {
...
+	err = ku.CurrencyPairs.StorePairs(asset.Spot, avail[:320], true)
+	require.NoError(t, err, "StorePairs must not error")
+
+	ku.Features.Subscriptions = subscription.List{{Asset: asset.Spot, Channel: subscription.AllTradesChannel}}
+	subs, err = ku.generateSubscriptions()
+	require.NoError(t, err, "generateSubscriptions must not error")
+	require.Len(t, subs, 4, "Must get 4 subs")
+
+	err = ku.Subscribe(subs)
+	require.ErrorContains(t, err, "exceed max subscription count limitation of 300 per session", "Subscribe to MarketSnapshot must error above connection symbol limit")
```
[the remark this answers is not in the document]
gbjk: Well, yeah. That's the point of this. 😄 

If they allow more, we want to fail, so we can increase our allowance (or probably the ticker:all switch at 10)

exchanges/kucoin/kucoin_test.go
```diff
@@ -2490,14 +2483,90 @@ func TestProcessMarketSnapshot(t *testing.T) {
...
+	require.NoError(t, err, "GetAvailablePairs must not error")
+
+	err = ku.CurrencyPairs.StorePairs(asset.Spot, avail[:500], true)
+	require.NoError(t, err, "StorePairs must not error")
+
+	ku.Features.Subscriptions = subscription.List{{Asset: asset.Spot, Channel: subscription.TickerChannel}}
+
+	subs, err := ku.generateSubscriptions()
+	require.NoError(t, err, "generateSubscriptions must not error")
+	require.Len(t, subs, 1, "Must generate one subscriptions")
```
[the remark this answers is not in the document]
gbjk: Fixed fa49f70f

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -28,60 +31,48 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/log"
 )
 
-var fetchedFuturesSnapshotOrderbook map[string]bool
+var fetchedFuturesOrderbookMutex sync.Mutex
+var fetchedFuturesOrderbook map[string]bool
```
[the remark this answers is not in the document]
gbjk: Yes. Just not in this PR 😄 
I was *really* trying not to get into this bit, cos I sensed I'd start pulling and end up with a lot more 🧶 
I've got it on my list for a rainy day or high priority after all exchanges are configurable.

exchanges/kucoin/kucoin_websocket.go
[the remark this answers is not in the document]
gbjk: Fixed 9c496153

More to come:
* Tests for this case in template
* And can make kucoin's template robust against this, and also simplify the handling of assetCurrencies by making it something that currency.Pairs knows how to do

exchanges/kucoin/kucoin_websocket.go
gbjk: Hmmm.
So my issue with this is the back-off we'd need.
If we get orderbook updates, and we haven't loaded the snapshot, and we don't say we've loaded it before we error, we're going to keep looping around in an expensive stream-rate-driven loop.

If it's okay, I'll swerve fixing this issue until I come back and deal with this better entirely.

I think you already noted: I didn't introduce this issue you're citing; it's just been moved from line 294 to here.

exchanges/kucoin/kucoin_websocket.go
gbjk: I think that's okay and right, because of separation of responsibilities.
It's not down to anything below `expandTemplates` to know *how* it got the maps.
Especially since we're definitely going to move that.

But I did see that it's not clear *why* we're updating the AssetPairs map, so I've documented that.

Fixed 9b85f620

exchanges/subscription/template.go
gbjk: I think perhaps the cleanest explanation might belong in the docs for subscriptions:
Fixed 1e271b4b

Let me explain here, and then we can work out together how/where/what to document for others.
* BatchSize is a directive, which is optional.
* If the template doesn't use a batch size, then there is only one batch, which includes all pairs.
Hence line 147.
* We expect only one batchSize directive per assetRecord. So when we split on it, we have three possible situations:
 * We got more than 2 elements, meaning more than one directive, which is an error `too many BatchSize directives`
 * We got exactly 2 elements. The first part is the asset record, and the second part is the batch size. So we assign them like that
 * Or if we get only 1 element from the split, it means it doesn't include a batchSize. So the batch size remains the length of all the pairs. EXCEPT if the pairs are going to be expanded, and we're not batching, then each batch is one pair. The opposite of the default on line 147. So the two cases to consider to understand this EXCEPT are a subscription for All assets, which either wants one subscription for all pairs, and so batch size is len(pairs), OR a subscriptins that wants one subscription for each pair, and so batch size is 1.

Does that help explain the logic here?
If so, was there a specific "Ah Ha" moment in that verbose description I should make sure is in the comments next to the code?

exchanges/kucoin/kucoin_websocket.go
gloriousCode: Yeah that's fair 👍

exchanges/kucoin/kucoin_websocket.go
gloriousCode: Yeah it can be added to the pile. Its less about subscription templating

exchanges/subscription/template.go
gloriousCode: Thanks for writing that out! These are definitely the underlying rules I was seeking. 
Trying to think it through:
- I think `errTooManyBatchSize` turns to `errOneBatchSizePerAssetPair` to make the rule more solid
- The line `return nil, fmt.Errorf("%w for %s: Got %d; Expected %d", errPairRecords, a, len(pairLines), len(batches)) adds something like `pair line length must match batchSize number` again to highlight the rules of template generation

I think that addresses my confusions... Hopefully 😄 If batchSize is present, there can only be one per assetPair in the template. That if there is an batchSize, the count of the slices of pairLines must match the number of batches.

gbjk: Fixed edde8d01e

I've added an overall description at the top instead of trying to inline the comments as we go.
I think that this section lends itself better to a contextual block than code+comment style.

On item:
2. The issues that cause this can be very varied, so I don't want to suggest that this is about batchSize, especially when it isn't used most of the time. This code reads "batchSize centric", but that's really only because if there might be one, it's easier to normalise all paths to that. On that basis I've added a comment, but left the error alone.

gloriousCode: I'm happy with that, thanks!

cmd/documentation/exchanges_templates/subscription.tmpl
```diff
@@ -20,12 +20,30 @@ The template is provided with a single context structure:
   AssetPairs     map[asset.Item]currency.Pairs
```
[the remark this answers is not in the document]
gbjk: Fixed dffac5b2

exchanges/subscription/list.go
```diff
@@ -85,7 +85,7 @@ func fillAssetPairs(ap assetPairs, a asset.Item, e iExchange) error {
 	if err != nil {
```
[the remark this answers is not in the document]
gbjk: I think we mentioned this before.
Not intentional, no, but I think we're going to ignore it for now.
I'll  sniff it though.

cmd/documentation/exchanges_templates/subscription.tmpl
```diff
@@ -25,9 +25,21 @@ The template is provided with a single context structure:
 
 Subscriptions may fan out many channels for assets and pairs, to support exchanges which require individual subscriptions.  
 To allow the template to communicate how to handle its output it should use the provided directives:
-- AssetSeparator should be added at the end of each section related to assets
+- AssetSeparator should be added at the end of each section related to assets:
```
shazbert, line 28: `:` has been added and below was this intentional?

shazbert (approved):
Pretty much ch tACK. One question on the template.

shazbert:
Also there is a CI test issue: [here](https://github.com/thrasher-corp/gocryptotrader/actions/runs/10265536809/job/28401940242?pr=1579#step:10:642)

exchanges/subscription/list.go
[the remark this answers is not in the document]
gloriousCode: Thanks for resolving that one ⭐

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -1777,3 +1580,157 @@ func (ku *Kucoin) CalculateAssets(topic string, cp currency.Pair) ([]asset.Item,
...
+
+const subTplText = `
+{{- removeSpotFromMargin $.S $.AssetPairs -}}
+{{- if isCurrencyChannel $.S }}
+	{{ channelName $.S $.S.Asset -}} : {{- (assetCurrencies $.S $.AssetPairs).Join -}}
+{{- else if isSymbolChannel $.S }}
+	{{ range $asset, $pairs := $.AssetPairs }}
+		{{- with $name := channelName $.S $asset }}
+			{{- if and (eq $name "/market/ticker") (gt (len $pairs) 10) -}}
+				{{- $name -}} :all
```
gloriousCode, line 1721: I know you hate me at this point but I have to test things 😅 
Here's this:
```
[ERROR] | WEBSOCKET | 07/08/2024 10:16:01 | Kucoin websocket: subscription failure, ticker all : subscription template did not generate the expected number of pair records for spot: Got 1; Expected 17
```
Here's some config snippets:
```json
 "futures": {
      "assetEnabled": true,
      "enabled": "XBT_USDTM,ETH_USDTM",
```

```json
  "margin": {
      "assetEnabled": true,
      "enabled": "BTC-USDT",
```

```json
 "spot": {
      "assetEnabled": true,
      "enabled": "AVA-USDT,FET-BTC,FET-ETH,ANKR-BTC,XMR-BTC,XMR-ETH,MTV-BTC,MTV-ETH,CRO-BTC,MTV-USDT,KMD-BTC,KMD-USDT,TEL-USDT,TT-USDT,AERGO-USDT,XMR-USDT,TRX-KCS",
```

only enabled sub:
```json
 {
      "enabled": true,
      "channel": "ticker",
```

gloriousCode (changes requested):
Thanks for making those changes, happy with your work. Just found another stinky bug though

cmd/documentation/exchanges_templates/subscription.tmpl
gbjk: Fixed b9fe4d4b

exchanges/kucoin/kucoin_websocket.go
gbjk: Fixed cfb62def

Lost the batchSize for :all at some point.

gbjk:
> Also there is a CI test issue: [here](https://github.com/thrasher-corp/gocryptotrader/actions/runs/10265536809/job/28401940242?pr=1579#step:10:642)

🤦  Thanks. Sorry I didn't spot that.

Fixed 857ac7ad

exchanges/kucoin/kucoin_websocket.go
gbjk: Also added a test for this:

Fixed 7e118b29

gbjk: Okay, so this was a real edge case.
xpandPairs was set to true because futures and margin needed to expandPairs.
So testing this with asset: spot would have a different result from asset: All

The *second* correct fix for this was to also move the xpandPairs down into AssetPairs

Fixed 51ddad04 and 65ee7043

cmd/documentation/exchanges_templates/subscription.tmpl
shazbert: thanks!

shazbert (approved):
Lovely looking good!

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -94,12 +85,27 @@ var (
 	maxWSOrderbookWorkers = 10
 )
 
+var defaultSubscriptions = subscription.List{
```
gloriousCode, line 88: I'm okay with this one being knocked back, but there is spot candle data support which does work:
```go
	{Enabled: false, Asset: asset.Spot, Channel: subscription.CandlesChannel, Interval: kline.OneMin},
```

```json
 {
      "enabled": true,
      "channel": "candles",
      "asset": "spot",
      "interval": "1m"
     }
```

```
[DEBUG] | WEBSOCKET | 07/08/2024 15:08:55 | Kucoin websocket connection: message received: {"topic":"/market/candles:ATOM-USDT_1min","type":"message","subject":"trade.candles.update","data":{"symbol":"ATOM-USDT","candles":["1723007280","4.9799","4.9803","4.9807","4.9799","37.3","185.77914072"],"time":1723007335395578690}}
```
In GCT we don't do anything with it though, it can be utilised externally, but I'm okay with it not being here

gloriousCode (changes requested):
One comment on error message and one optional thing, then I'm done

currency/manager.go
```diff
@@ -289,11 +289,14 @@ func (p *PairsManager) DisablePair(a asset.Item, pair Pair) error {
...
 
-	enabled, err := pairStore.Enabled.Remove(pair)
-	if err != nil {
-		return err
+	enabledLen := len(pairStore.Enabled)
+
+	pairStore.Enabled = pairStore.Enabled.Remove(pair)
+
+	if enabledLen == len(pairStore.Enabled) {
+		return fmt.Errorf("%w 2 %s", ErrPairNotFound, pair)
```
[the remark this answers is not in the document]
gbjk: Fixed 01a9f9f9

exchanges/kucoin/kucoin_websocket.go
gbjk: I wasn't generally adding available subscriptions to this list currently.
Exception if it's disabled for some sort of documentable reason that might interest people.

gloriousCode (approved):
tACK!

shazbert (approved):
Ch ack

gbjk:
Squashed down for final review 😄 
Zero changes in the squash though.

common/common.go
```diff
@@ -653,3 +654,34 @@ func GetTypeAssertError(required string, received interface{}, fieldDescription
 	}
 	return fmt.Errorf("%w from %T to %s%s", ErrTypeAssertFailure, received, required, description)
 }
+
+// Batch takes a slice type and converts it into a slice of containing slices of length batchSize, and any remainder in the final batch
+// batchSize <= 0 will return the entire input slice in one batch
+func Batch[S ~[]E, E any](blobs S, batchSize int) []S {
```
thrasher-, line 660: Now that this is introduced, we can rid SplitStringSliceByLimit which BTC Markets is the only consumer of.

thrasher- (changes requested):
Just one minor nit on the introduction of the new Batch helper. It makes a pre-existing helper func obsolete.

Tests, functionality for Kucoin and config uprade all worked well. Nice stuff! Appreciate the strings.HasPrefix improvement as well on the websocket handling topic info!

common/common.go
gbjk: Nice spot 🙏 

Fixed 7c3d7fb1

Sorry it's not a straight drop in, but couldn't tolerate the strangeness going on there with tempMap.

thrasher- (approved):
Thanks for making those changes

PR #1580, opened 2024-07-15
websocket/gateio:  Support multi connection management and integrate with GateIO

shazbert:
### Multiple Connections for Asset Endpoints
- **Feature**: Implemented the ability to spawn multiple WebSocket connections for different asset endpoints.
- **Design Decision**: 
  - Opted for a separate subscription holder per connection due to import cycle issues with a unified subscription manager.
  - This allows for future scalability when exchange constraints limit subscriptions per connection.

### Connection-Coupled Functionality
- **Approach**: Coupled specific functionality with the connection itself, along with its own subscription holder.
- **Benefit**: Enables tracking multiple connections efficiently, facilitating scaling when exchanges restrict subscriptions per connection.

### Reader Routine in Streaming Package
- **Implementation**: Created a reader routine in the streaming package to avoid code duplication.
- **Future Work**: This will initiate a purge across wrappers to standardize this routine.

### Context Awareness
- **Progress**: Began integrating context awareness throughout the WebSocket functionality.
- **Enhancement**: Added a `DialContext` method to `WebsocketConnection` to leverage the supported context.

---

### WebSocket Support
- **Added Support**: 
  - Futures
  - Dated Futures
  - Options
- **Note**: Margin, cross margin, and BTC-margined dated futures not included due to PR size and complexity.

### Bug Fixes
- **Delivery Futures Handling**: 
  - Fixed issue where all available pairs were fetched instead of only enabled pairs.
- **Payload Generation**: 
  - Addressed issues in `generateDeliveryFuturesPayload` where different settlements were handled incorrectly.
  - Separated connections for different settlements to flatten functions for consistency.

---

🐛 fix:

* [Routine leak on ping handler](https://github.com/thrasher-corp/gocryptotrader/pull/1580/commits/2d2f8722dd81fa13724dbec1d348777206b96cf4)
* [potential Connection leak without reader](https://github.com/thrasher-corp/gocryptotrader/pull/1580/commits/39191c865734207b8a836819d9f36825b20d9efa)

## Type of change

## How has this been tested

- [I ran it 🏃 instead] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/stream/websocket_connection.go
```diff
@@ -192,6 +224,9 @@ func (w *WebsocketConnection) IsConnected() bool {
 
 // ReadMessage reads messages, can handle text, gzip and binary
 func (w *WebsocketConnection) ReadMessage() Response {
+	if w.Connection == nil {
+		return Response{}
+	}
```
gloriousCode, line 229: Has this happened? Is there a scenario I've missed?

gloriousCode (changes requested):
I'll only offer very light commentary right now. 

There are things I'd do differently from this first look, but it doesn't work to comment in that way because that could blow out having an implementation where this looks to try and minimise the amount of changes while having backwards compatability.

I'll try and focus on things to fix up and any holes I find

exchanges/gateio/gateio_websocket.go
```diff
@@ -723,14 +712,14 @@ func (g *Gateio) handleSubscription(event string, channelsToSubscribe subscripti
...
-func (g *Gateio) generatePayload(event string, channelsToSubscribe subscription.List) ([]WsInput, error) {
+func (g *Gateio) generatePayload(ctx context.Context, conn stream.Connection, event string, channelsToSubscribe subscription.List) ([]WsInput, error) {
 	if len(channelsToSubscribe) == 0 {
 		return nil, errors.New("cannot generate payload, no channels supplied")
 	}
 	var creds *account.Credentials
 	var err error
 	if g.Websocket.CanUseAuthenticatedEndpoints() {
-		creds, err = g.GetCredentials(context.TODO())
+		creds, err = g.GetCredentials(ctx)
```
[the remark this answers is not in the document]
shazbert: Yeah as long as the connection accepts auth it will subscribe to it, the only issue is outbound messaging which I will work on when I integrate an exchange which supports it and this design is stable if that is okay.

exchanges/stream/websocket.go
```diff
@@ -279,37 +284,134 @@ func (w *Websocket) Connect() error {
...
+		return nil
 	}
 
-	subs, err := w.GenerateSubs() // regenerate state on new connection
-	if err != nil {
-		return fmt.Errorf("%s websocket: %w", w.exchangeName, common.AppendError(ErrSubscriptionFailure, err))
+	// hasStableConnection is used to determine if the websocket has a stable
+	// connection. If it does not, the websocket will be set to disconnected.
+	hasStableConnection := false
+	defer w.setStateFromHasStableConnection(&hasStableConnection)
```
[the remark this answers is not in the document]
shazbert: Yeah that is much simpler, I will shutdown all open connections and clean up if there is an error connecting or there is a subscription error. We can update it later if this is too over the top.

exchanges/stream/websocket_connection.go
shazbert: Yeah caused by me and an incorrect configuration. If this is called before dialing it panics.

gbjk: TL;DR: I think we should take the warning labels off the boxes.

I think panicing is the best thing to do there.
Discussed a few times with @thrasher-, and if developer error is the only way to cause something, and it happens during setup/startup, then that's fine.
The failure they'd get is that Connection is nil.
If they've managed to cause that, they'll be able to understand it as well.

gbjk (commented):
This first comment from me covers the overall architecture.

I don't think that a single websocket should have more than one URL behind/inside it.
If we break that, then what do we really mean when we say "this is a websocket".

So I think we could/should take this opportunity to:
* Add the concept of a websockets router/manager ( `exchange/websocket/router/` )
* Exchanges which use multiple URLs will have multiple websockets
* Websockets in turn have routers for connection fanning based on limits (and anything else I might be forgetting)
* subscriptions should be separate, and routed separately, as `exchange/subscription`
* subscription reconnection should behave statelessly, not needing to know about what conn it was on that just got disconnected
* N+ across websockets and websocket connections; Handling should be identical whether we have one or more
* Radical step: Break out the idea of subscription chans, and we make it so that messages for a subscription go directly to a specific subscription handler. The downside of that would be out-of-sequence handling of unrelated channels, but that feels okay

Note: Yes, I wrote `exchange/`; Radical, right!

So overall, I agree with your problem, but I feel you can be more bullish in rearchitecting this.

You mentioned that your design decision related to a per-connection subscription store was related to import cycles. We can use an interface to avoid the cycle, and possibly localising the parts of the interface that you want by compositing smaller interfaces.
If you want to use something, it makes sense to define what you want to use yourself, rather than saying "I want a subscription and all that entails". e.g. "Give me anything that can behave like this", means a static and limited scope that we be more adaptive.

You can engage in as much or as little of this as you want. You can ignore it entirely :)

shazbert:
> This first comment from me covers the overall architecture.

Thanks for the suggestions. I appreciate the thought you've put into the overall architecture and the proposed changes you outlined should help provide a more logical structured framework. 

I'd like to suggest for now, we proceed with the current implementation @thrasher- . Just so we can merge the PR and make it usable right away and can get Sam to start rolling that across the exchanges straight away. In the meantime we can start planning iterations to gradually evolve the architecture towards the model you've outlined?

So plan is, after this is merged I want to:
   * Make connections and subscriptions concurrent
   * Have N+ subs per "connection/websocket/whatever name its going to be"

I'd like to defer localizing the interface for the subscription package, just for now. I don't want to start blowing out complexity for this PR, there's already a bit of moving parts. Perhaps you could open a supplementary PR to address this when you have the time? 

Thanks again for your input!

gbjk:
Sounds good buddy 👍

exchanges/stream/websocket.go
```diff
@@ -127,47 +133,52 @@ func (w *Websocket) Setup(s *WebsocketSetup) error {
...
-	if w.connectionMonitorDelay <= 0 {
-		w.connectionMonitorDelay = config.DefaultConnectionMonitorDelay
-	}
-	w.Unsubscriber = s.Unsubscriber
-
-	if s.GenerateSubscriptions == nil {
-		return fmt.Errorf("%s %w", w.exchangeName, errWebsocketSubscriptionsGeneratorUnset)
-	}
-	w.GenerateSubs = s.GenerateSubscriptions
+	if !w.useMultiConnectionManagement {
```
gbjk, line 138: Are we saying that we can't make all websockets use a single connection inside connection manager straight away, and remove the concept of "multi or not" straight away?

I can't see the connections being that invasive yet.

exchanges/stream/stream_types.go
```diff
@@ -36,9 +39,44 @@ type ConnectionSetup struct {
...
+	// Handler defines the function that will be called when a message is
+	// received from the exchange's websocket server. This function should
+	// handle the incoming message and pass it to the appropriate data handler.
+	Handler func(ctx context.Context, incoming []byte) error
+}
+
+// ConnectionCandidate contains the connection setup details to be used when
+// attempting a new connection. It also contains the subscriptions that are
+// associated with the specific connection.
+type ConnectionCandidate struct {
```
gbjk, line 71: The name Candidate feels wrong to me.
It has Setup, Subscriptions and Connection.
Maybe it's a ConnectionWrapper or something.

Candidate implies something that doesn't feel true; That this is one of several options which may or may not be chosen.

exchanges/stream/websocket.go
shazbert: Maybe I am not understanding the correct implementation of what you are trying to say. If you have time send me a diff here on the code block or open a PR against my branch and I will get to it monday. If it's a simple change then fo shiz.

exchanges/stream/stream_types.go
shazbert: Can do, the names I generate are literally the first thing that pops in my head. ask @gloriousCode 😆

exchanges/stream/websocket_test.go
```diff
@@ -1101,34 +1273,83 @@ func TestEnable(t *testing.T) {
...
-	err = web.SetupNewConnection(ConnectionSetup{URL: "urlstring", Authenticated: true})
+	err = web.SetupNewConnection(&ConnectionSetup{URL: "urlstring", Authenticated: true})
 	assert.NoError(t, err, "SetupNewConnection should not error")
+
+	// Test connection candidates for multi connection tracking.
+	multi := NewWebsocket()
+	set := *defaultSetup
+
+	// Values below are now not necessary as this will be set per connection
+	// candidate in SetupNewConnection.
```
gbjk, line 1307: I don't understand this comment - If not necessary, why not remove?

shazbert: This was before when I was using the connector function as a pivot to use instead of the bool `UseMultiConnectionManagement` I will remove and the setting of nil below.

gbjk (changes requested):
Only thing from me is to try to get away from the branching logic around the use-multi.
If we can't harmonise, I don't yet understand why.
Other than that, I didn't spot anything.

exchanges/stream/websocket.go
```diff
@@ -826,12 +1072,19 @@ func (w *Websocket) SubscribeToChannels(subs subscription.List) error {
...
-		w.subscriptions = subscription.NewStore()
+	var subscriptionStore **subscription.Store
+	if candidate, ok := w.connections[conn]; ok {
+		subscriptionStore = &candidate.Subscriptions
+	} else {
+		subscriptionStore = &w.subscriptions
+	}
+
+	if *subscriptionStore == nil {
+		*subscriptionStore = subscription.NewStore()
```
gloriousCode, line 1087: Could you explain the need for the pointer to a pointer situation?

exchanges/stream/websocket.go
```diff
@@ -283,27 +321,139 @@ func (w *Websocket) Connect() error {
...
+		err = w.connectionManager[i].Setup.Connector(context.TODO(), conn)
+		if err != nil {
+			multiConnectFatalError = fmt.Errorf("%v Error connecting %w", w.exchangeName, err)
+			break
 		}
+
+		w.Wg.Add(1)
+		go w.Reader(context.TODO(), conn, w.connectionManager[i].Setup.Handler)
+
+		w.connections[conn] = &w.connectionManager[i]
```
gloriousCode, line 416: There is a degree of disconnect between `w.connections` and `w.connectionManager.` Is there a reason you don't try to assign them and remove them at the same time?
I see on Shutdown, you clear `w.connections` while leaving `w.connectionManager`, but I'm not sure I'm getting the distinction between the two. I took them both as more of a design of a sliceMap, so you can instant lookup connections via a map, but can quickly iterate over them in a slice

exchanges/stream/websocket.go
```diff
@@ -486,40 +653,90 @@ func (w *Websocket) FlushChannels() error {
...
+			newsubs, err := w.connectionManager[x].Setup.GenerateSubscriptions()
+			if err != nil {
+				if errors.Is(err, asset.ErrNotEnabled) {
+					continue
+				}
+				return err
+			}
+			if len(newsubs) != 0 {
+				// Purge subscription list as there will be conflicts
+				w.connections[w.connectionManager[x].Connection].Subscriptions.Clear()
```
gloriousCode, line 734: We can see from how you've designed this, is that `w.connectionManager` is the source of truth, being that the map `w.connections` contains a pointer to the slice entry. You already have `w.connectionManager[x].Subscriptions` why not use that?

exchanges/stream/websocket.go
```diff
@@ -283,27 +321,139 @@ func (w *Websocket) Connect() error {
...
+
+		w.connectionManager[i].Connection = conn
 	}
 
-	subs, err := w.GenerateSubs() // regenerate state on new connection
-	if err != nil {
-		return fmt.Errorf("%s websocket: %w", w.exchangeName, common.AppendError(ErrSubscriptionFailure, err))
+	if multiConnectFatalError != nil {
+		// Roll back any successful connections and flush subscriptions
+		for conn, candidate := range w.connections {
```
gloriousCode, line 437: Unless there is a reason otherwise, we shouldn't ever really iterate over the map if the slice contains the same fields. Should you not be using `w.connectionManager[x].Connection.Shutdown`?

exchanges/stream/websocket_connection.go
gloriousCode: I'm happier with a middleground. Fix the issue that caused your panic and RM this section. eg have a block on `(w *Websocket) Reader` until `Connected` is true

exchanges/stream/websocket_connection.go
```diff
@@ -192,25 +203,27 @@ func (w *WebsocketConnection) IsConnected() bool {
...
-				default:
-					// bypass if there is no receiver, as this stops it returning
-					// when shutdown is called.
-					log.Warnf(log.WebsocketMgr,
-						"%s failed to relay error: %v",
-						w.ExchangeName,
-						err)
-				}
+		// Any error condition will return nil response which will return the
+		// reader routine and the connection will hang with no readers. This has
```
gloriousCode, line 212: Could you please rewrite this for clarity? An error return is a nil return?

exchanges/stream/websocket_connection.go
```diff
@@ -228,18 +241,12 @@ func (w *WebsocketConnection) ReadMessage() Response {
 	case websocket.BinaryMessage:
 		standardMessage, err = w.parseBinaryResponse(resp)
 		if err != nil {
-			log.Errorf(log.WebsocketMgr,
-				"%v websocket connection: parseBinaryResponse error: %v",
-				w.ExchangeName,
-				err)
-			return Response{}
+			log.Errorf(log.WebsocketMgr, "%v websocket connection: parseBinaryResponse error: %v", w.ExchangeName, err)
+			return Response{Raw: []byte(``)} // Non-nil response to avoid the reader returning on this case.
```
gloriousCode, line 245: This seems like a patchup that could have unintended consequences. What situation caused this?

gloriousCode (changes requested):
Thanks for making those changes. Most of these comments are down to two things, the slice+map design and the double pointer situation. Not necessarily things that need to change, but I do want to understand them more

exchanges/stream/websocket.go
shazbert: Just keeps the code more concise, there is nothing that it adds just removes a bunch of doubled up code when it is nil and needs a new store.

exchanges/stream/websocket_connection.go
shazbert: No situation, I just noticed this would return the reader routine, the connection seems like it would still be valid, the websocket frame is intact so "should" 😆 be able to keep reading the next messages but continue to error if the binary response is garbage. I can roll this back if you so desire?

exchanges/stream/websocket.go
shazbert: I am not really understanding assigning and removing at the same time sorry. 

As for the shutdown I clear the connections because they will drop in shutdown and not need to be used, the connection manager holds a singular ConnectionCandidate (will be named to ConnectionWrapper now) you can think of it basically  as a URL holder with handlers (which I can name that as well) then from that you can generate and spawn any amount of connections based off that. The connections map is purely for subscription management specifically, but will only come into play when there are many connections holding different subscriptions that need to be scaled and managed. But If for now this is too much for this PR, I can drop that map entirely iterate over the manager itself and when I do upgrade the system to spawn connections based of subscriptions I can reinstate?

gloriousCode: Thank you for your details!
I was approaching the Connection slice and map as like a sliceMap configuration, a way to quickly look up a connection via map, while being able to iterate over all connections in a slice.
What you're saying is that isn't the case and in that sense, the assigning things at the same time, doesn't really apply.
On next review I'll try to understand your design with these details in mind and comment from there

exchanges/stream/websocket_connection.go
shazbert: Purged this contentious issue. 😆

exchanges/stream/websocket_connection.go
```diff
@@ -293,7 +292,11 @@ func (w *WebsocketConnection) Shutdown() error {
 		return nil
 	}
 	w.setConnectedStatus(false)
-	return w.Connection.UnderlyingConn().Close()
+	w.writeControl.Lock()
+	defer w.writeControl.Unlock()
+	defer w.Connection.NetConn().Close() // Ungraceful close as backup
+	// Gracefully close the connection
+	return w.Connection.WriteMessage(websocket.CloseMessage, websocket.FormatCloseMessage(websocket.CloseNormalClosure, ""))
```
gbjk, line 299: TL;DR: Just close it.

I'm afraid these are meaningless together.
If you're not going to wait for the response, you haven't closed the connection gracefully.
From the perspective of the server they saw:
"_Hey, We're going to close the webso..._" <slam>🚪 💨 

I'm assuming you read [this discussion](https://github.com/gorilla/websocket/issues/448) as the starting point for this, but most notably the last two comments seem relevant.

I think all servers are fine with you just closing it. They don't expect any better.
There's no information gained by signalling you're going to shut down.
It's not like they'd say "we had this order message for you, and we were hoarding it, but here, have it now".

exchanges/stream/websocket_test.go
```diff
@@ -1448,3 +1445,36 @@ func TestCheckSubscriptions(t *testing.T) {
...
+	defer conn.Close()
+
+	for {
+		messageType, message, err := conn.ReadMessage()
+		if err != nil {
+			if websocket.IsUnexpectedCloseError(err, websocket.CloseNormalClosure) {
+				panic(err)
+			}
+			break
+		}
```
gbjk: I think you could use / move `internal/testing/exchange/exchange.go` for most of this.
Your `WsMockFunc` would just do the below.

`internal/testing/websocket/mock.go` might be a good new home to share the mock upgrader.

exchanges/stream/websocket_types.go
```diff
@@ -50,12 +50,12 @@ type Websocket struct {
 	m                            sync.Mutex
 	connector                    func() error
 
-	// ConnectionManager contains the connection candidates and the current
+	// connectionWrapper contains the connection wrappers and the current
```
gbjk, line 53: Did you mean connectionManager?


Also: I don't think this comment adds any useful information about this _private_ variable :)

exchanges/stream/websocket_types.go
```diff
@@ -50,12 +50,12 @@ type Websocket struct {
...
 	// connections
-	connectionManager []ConnectionCandidate
+	connectionManager []ConnectionWrapper
 	// Connections contains the current connections with their associated
 	// connection candidates
```
gbjk, line 57: Associated wrappers :)

gbjk (changes requested):
Minor changes.
I've put "Request Changes" in order to see how a broken out `WsMockUpgrader` shakes out.

exchanges/stream/websocket_connection.go
shazbert: That's fair I didn't read the discussion just saw an abnormal close and thought to change it out. I guess I will just close it then 🤷

exchanges/stream/websocket_test.go
shazbert: Good suggestion will do

exchanges/stream/websocket_types.go
shazbert: yep

exchanges/stream/websocket_types.go
shazbert: 👀

internal/testing/websocket/mock.go
```diff
@@ -0,0 +1,57 @@
...
+
+		require.NoError(tb, err, "ReadMessage should not error")
+
+		err = wsHandler(p, c)
+		assert.NoError(tb, err, "WS Mock Function should not error")
+	}
+}
+
+// WsWithEcho is a simple echo function after a read
+func WsWithEcho(p []byte, c *websocket.Conn) error {
```
gbjk, line 51: Minor suggestion:
`EchoHandler` is a good description for this, given the package it's in, and what it's doing.

exchanges/stream/websocket.go
```diff
@@ -449,6 +453,14 @@ func (w *Websocket) Connect() error {
 		}
 		clear(w.connections)
 		w.setState(disconnectedState) // Flip from connecting to disconnected.
+
+		// Drain residual error in the single buffered channel, this mitigates
+		// the cycle when `Connect` is called again and the connectionMonitor
+		// starts but there is an old error in the channel.
+		select {
+		case <-w.ReadMessageErrors:
```
gbjk, line 461: Is there definitely only one?
Otherwise could range drain it.

gbjk (approved):

samuael:
https://github.com/shazbert/gocryptotrader/blob/6376a12b18b7db8ab6d7ab68864af89e8ab06342/exchanges/okx/okx_websocket.go#L254

This function is very similar to WsConnect, so it might be a good idea to treat WsAuth as a single connection and use the SetupConnection method call to initialize it. Additionally, you could consider moving the dialer and context information inside the function.

samuael:
https://github.com/shazbert/gocryptotrader/blob/6376a12b18b7db8ab6d7ab68864af89e8ab06342/exchanges/stream/websocket.go#L901

Just something to consider for the next round of improvements: we can eliminate these methods by treating each connection as a single instance. This instance can be appended to the slice of ConnectionWrapper, thereby reducing the complexities created by **_Conn_** and **_AuthConn_**

shazbert:
> https://github.com/shazbert/gocryptotrader/blob/6376a12b18b7db8ab6d7ab68864af89e8ab06342/exchanges/okx/okx_websocket.go#L254
> 
> This function is very similar to WsConnect, so it might be a good idea to treat WsAuth as a single connection and use the SetupConnection method call to initialize it. Additionally, you could consider moving the dialer and context information inside the function.

Good suggestion, I think that would be the direction to go when implementing. I have started a proof of concept to grow connections based off the max subscriptions per connection and I think I have already done what you have suggested. Will send you a link to it here when it's ready for your look over.

shazbert:
> https://github.com/shazbert/gocryptotrader/blob/6376a12b18b7db8ab6d7ab68864af89e8ab06342/exchanges/stream/websocket.go#L901
> 
> Just something to consider for the next round of improvements: we can eliminate these methods by treating each connection as a single instance. This instance can be appended to the slice of ConnectionWrapper, thereby reducing the complexities created by **_Conn_** and **_AuthConn_**

Is this specifically pointing to the change in SetWebsocketURL?

exchanges/stream/websocket.go
```diff
@@ -410,11 +414,17 @@ func (w *Websocket) Connect() error {
 			break
 		}
 
+		w.connections[conn] = &w.connectionManager[i]
+		w.connectionManager[i].Connection = conn
+
+		if !conn.IsConnected() {
+			multiConnectFatalError = fmt.Errorf("%s websocket: [conn:%d] [URL:%s] failed to connect", w.exchangeName, i+1, conn.URL)
+			break
```
gloriousCode, line 422: Why in this order? Why not only set things when you know they're successful? Seems like this check should be before setting it

exchanges/stream/websocket_connection.go
```diff
@@ -203,15 +203,14 @@ func (w *WebsocketConnection) IsConnected() bool {
...
-		// Any error condition will return nil response which will return the
-		// reader routine and the connection will hang with no readers. This has
-		// to be handed over to w.readMessageErrors if there is an active
-		// connection.
+		// Any error condition will return a Response{Raw: nil, Type: 0} which
+		// will force the reader routine to return. The connection will hang
+		// with no reader routine and its buffer will be written to from the
+		// active websocket connection. This should be handed over to
+		// `w.readMessageErrors` and managed by 'connectionMonitor' which needs
+		// to flush, reconnect and resubscribe the connection.
```
gloriousCode, line 213: Should? It does?

exchanges/stream/websocket_connection.go
gloriousCode: I think its suitable enough

gloriousCode (changes requested):
Things are looking good, but it failed the simple test of being disconnected and not reconnecting:
edit: what a fool! I was testing okx. So that has its own issues 😆 Let me retry with gateio

gloriousCode (changes requested):
The good part is that it works and that I haven't yet managed to make it fall over. Its naturally more complex, but to interact with the system via wrapper setup is logical. It also doesn't add a hefty weight to everything which is nice

exchanges/stream/websocket.go
```diff
@@ -784,16 +1016,38 @@ func (w *Websocket) GetName() string {
...
+	if candidate, ok := w.connections[conn]; ok {
+		if candidate.Subscriptions == nil {
+			return nil // No channels to unsubscribe from is not an error
+		}
+		for _, s := range channels {
+			if candidate.Subscriptions.Get(s) == nil {
+				return fmt.Errorf("%w: %s", subscription.ErrNotFound, s)
+			}
+		}
+		return candidate.Setup.Unsubscriber(context.TODO(), conn, channels)
```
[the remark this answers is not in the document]
shazbert: It's hard to consolidate due to the unsubscriber method parameters from the original to the new multi connection signature. The originals will be removed eventually when everything is using the multi connection manager. Maybe I am missing what you are asking, here's what I did just to make:

```go
// UnsubscribeChannels unsubscribes from a list of websocket channel
func (w *Websocket) UnsubscribeChannels(conn Connection, channels subscription.List) error {
	if len(channels) == 0 {
		return nil // No channels to unsubscribe from is not an error
	}
	if candidate, ok := w.connections[conn]; ok {
		return w.unsubscribe(candidate.Subscriptions, channels, func(channels subscription.List) error {
			return candidate.Setup.Unsubscriber(context.TODO(), conn, channels)
		})
	}
	return w.unsubscribe(w.subscriptions, channels, func(channels subscription.List) error {
		return w.Unsubscriber(channels)
	})
}

func (w *Websocket) unsubscribe(store *subscription.Store, channels subscription.List, unsubscribe func(channels subscription.List) error) error {
	if store == nil {
		return nil // No channels to unsubscribe from is not an error
	}
	for _, s := range channels {
		if err := store.Remove(s); err != nil {
			if store.Get(s) == nil {
				return fmt.Errorf("%w: %s", subscription.ErrNotFound, s)
			}
		}
	}
	return unsubscribe(channels)
}
```

So the foot print is pretty much unchanged

exchanges/stream/websocket.go
```diff
@@ -495,40 +679,90 @@ func (w *Websocket) FlushChannels() error {
...
+				return err
+			}
+			if len(newsubs) != 0 {
+				// Purge subscription list as there will be conflicts
+				w.connectionManager[x].Subscriptions.Clear()
+				err = w.SubscribeToChannels(w.connectionManager[x].Connection, newsubs)
+				if err != nil {
+					return err
+				}
+			}
```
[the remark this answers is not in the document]
shazbert: Here as well has a similar issue [with abstraction](https://github.com/thrasher-corp/gocryptotrader/pull/1580#discussion_r1737923281) maybe I am getting which could be the case, could you please provide an example.

exchanges/stream/websocket.go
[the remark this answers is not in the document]
gloriousCode: It's less about footprint and more about repeated code. I definitely didn't  _make_ something, I just spotted two sections doing the same thing and thought it could be abstracted. What you posted looks nice. I can give a more thorough look on my next pass

exchanges/okx/okx_websocket.go
```diff
@@ -538,10 +538,10 @@ func (ok *Okx) handleSubscription(operation string, subscriptions subscription.L
 
 	channels = append(channels, authChannels...)
 	if operation == operationUnsubscribe {
-		return ok.Websocket.RemoveSubscriptions(channels...)
+		return ok.Websocket.RemoveSubscriptions(ok.Websocket.Conn, channels...)
```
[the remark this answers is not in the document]
shazbert: Looks like it's working for me.
```
go run ./... pair disable --exchange='okx' --pairs="ape-usdc" --asset=spot
{
 "status": "success"
}
```

```
go run ./... pair enable --exchange='okx' --pairs="ape-usdc" --asset=spot
{
 "status": "success"
}
```

Originally there was no differentiation between authenticated and unauthenticated connections. One of the reasons why I opted for a nil when I opened this PR because it was specifically not going to be matched. It should just be defaulting to what it was doing prior.

exchanges/stream/websocket.go
shazbert: Ahh okay the core issue is when `subscriptionStore` is nil. We then call:
```go
subscriptionStore = subscription.NewStore()
```

But that still needs to be set somewhere and when this function returns it just disappears. So what the ** means is that we are setting up a pointer to a field value.

exchanges/stream/stream_types.go
```diff
@@ -10,11 +10,13 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
```
gloriousCode, line 10: A somewhat optional request is perhaps this is a good point to add a `readme.md` to the stream folder

gloriousCode (changes requested):
I think this is nearly good to go. I mostly just want to see further testing and documentation. The overall design and implementation is a good improvement to GCT 📈

exchanges/stream/stream_types.go
shazbert: Happy to do it, what would you like in there?

exchanges/gateio/gateio_wrapper.go
```diff
@@ -205,31 +195,101 @@ func (g *Gateio) Setup(exch *config.Exchange) error {
...
+		Features:                     &g.Features.Supports.WebsocketCapabilities,
+		FillsFeed:                    g.Features.Enabled.FillsFeed,
+		TradeFeed:                    g.Features.Enabled.TradeFeed,
+		UseMultiConnectionManagement: true,
+	})
+	if err != nil {
+		return err
+	}
+	// Spot connection
+	err = g.Websocket.SetupNewConnection(&stream.ConnectionSetup{
```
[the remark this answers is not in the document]
shazbert: If we do that then if we enable, disable or re-enable an asset then we might need to call setup again. I can't think of a better way of doing it. Let me know your thoughts on how to over come that little obstacle because what I have done is not the best.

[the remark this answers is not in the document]
gloriousCode: I don't think even in its current form, the scenario you speak of doesn't work. If it is important to you, then you already need to change code. t present.
Enabling an asset does not call `exch.Websocket.Connect`, the websocketroutinemanager already considers the websocket connected, it only checks for ws.enabled. I don't find it as important to consider, especially since it doesn't work.

In fact, you raising this has demonstrated a bug!!!!!!
```
gctcli % go run . disableexchange gateio 
{
 "status": "success"
}%                        
```

```
[DEBUG] | WEBSOCKET | 12/09/2024 16:16:42 | GateIO websocket: running connection monitor cycle
[DEBUG] | WEBSOCKET | 12/09/2024 16:16:42 | GateIO websocket: shutting down websocket
[DEBUG] | WEBSOCKET | 12/09/2024 16:16:42 | GateIO websocket: completed websocket shutdown
[INFO]  | EXCHANGE | 12/09/2024 16:16:42 | gateio exchange unloaded successfully.
[INFO]  | WEBSOCKET | 12/09/2024 16:16:55 | GateIO Websocket connected to wss://api.gateio.ws/ws/v4/
[DEBUG] | WEBSOCKET | 12/09/2024 16:16:55 | GateIO websocket connection: sending message [{"time":1726121815,"id":5,"channel":"spot.tickers","event":"subscribe","payload":["BTC_USDT"]}]
...
```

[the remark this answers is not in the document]
gloriousCode: As for the asset enabled aspect. I think I would just like to see something more explicit than relying on `GenerateSubs()`returning asset error to prevent a ws from connecting.
I don't mind it if Gateio setup stays the same ultimately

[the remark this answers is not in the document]
shazbert: I removed the asset error handling in recent commits and defaulted to what spot was doing.

exchanges/stream/stream_types.go
shazbert: Added a basic [something](https://github.com/thrasher-corp/gocryptotrader/pull/1580/commits/818584fbb4d65c78eb442ab8043ee41405e94928)

Add what more you want when you can.

exchanges/gateio/gateio_wrapper.go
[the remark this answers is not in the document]
shazbert: I fixed the flushing in the most recent commit and should be good to go, I can't consolidate the code down to what it was before as it is too different from previous handling to multiconnection handling so I reverted that and expanded test functionality in `TestFlushChannels`; [check here](https://github.com/thrasher-corp/gocryptotrader/pull/1580/commits/0709ba30eaab1b702eb19bc0230dc4100ac55215)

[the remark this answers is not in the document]
shazbert: As for the unloading and loaded the exchange via RPC there seems to be a bit more in engine that needs to be done to support this; might have to open another PR to fix this issue.

exchanges/gateio/gateio_ws_delivery_futures.go
```diff
@@ -78,13 +78,17 @@ func (g *Gateio) GenerateDeliveryFuturesDefaultSubscriptions() (subscription.Lis
...
 		)
 	}
-	pairs, err := g.GetEnabledPairs(asset.DeliveryFutures)
-	if err != nil {
-		return nil, err
-	}
 
 	var subscriptions subscription.List
 	for i := range channelsToSubscribe {
+		pairs, err := g.GetEnabledPairs(asset.DeliveryFutures)
```
gloriousCode, line 85: I think this needs re-looking at. Its now in a loop with an expensive call and if the pairs aren't enabled, continue and call the function again?

My guess here is that you want to return `nil, nil`. So I think you can just do that outside of this loop

gloriousCode (changes requested):
I appreciate the asset CLI fix!

exchanges/gateio/gateio_ws_delivery_futures.go
shazbert: what 😆

shazbert: y i do dis?

exchanges/stream/websocket_test.go
```diff
@@ -287,22 +275,11 @@ func TestTrafficMonitorShutdown(t *testing.T) {
...
 
-func TestIsDisconnectionError(t *testing.T) {
-	t.Parallel()
-	assert.False(t, IsDisconnectionError(errors.New("errorText")), "IsDisconnectionError should return false")
-	assert.True(t, IsDisconnectionError(&websocket.CloseError{Code: 1006, Text: "errorText"}), "IsDisconnectionError should return true")
-	assert.False(t, IsDisconnectionError(&net.OpError{Err: errClosedConnection}), "IsDisconnectionError should return false")
-	assert.True(t, IsDisconnectionError(&net.OpError{Err: errors.New("errText")}), "IsDisconnectionError should return true")
-}
-
 func TestConnectionMessageErrors(t *testing.T) {
```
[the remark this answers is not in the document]
shazbert: [fixed](https://github.com/thrasher-corp/gocryptotrader/pull/1580/commits/33c4128bc5ad380ae17e289a2926d1b4b510a770)

exchanges/stream/websocket_test.go
```diff
@@ -1050,22 +1201,20 @@ func TestFlushChannels(t *testing.T) {
...
-	w.GenerateSubs = problemFunc
-	err = w.FlushChannels() // error on full subscribeToChannels
+	w.GenerateSubs = func() (subscription.List, error) { return nil, errDastardlyReason } // error on generateSubs
+	err = w.FlushChannels()                                                               // error on full subscribeToChannels
 	assert.ErrorIs(t, err, errDastardlyReason, "FlushChannels should error correctly on GenerateSubs")
 
-	w.GenerateSubs = noSub
-	err = w.FlushChannels() // No subs to sub
+	w.GenerateSubs = func() (subscription.List, error) { return nil, nil } // No subs to sub
+	err = w.FlushChannels()                                                // No subs to sub
```
gloriousCode, line 1209:
```
    websocket_test.go:1210: 
        	Error Trace:	/Users/scottg/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket_test.go:1210
        	            	test websocket already connected
        	Test:       	TestFlushChannels
        	Messages:   	Flush Channels should not error
```

I got this, then also debugged it. So it kinda goes through:
- FlushChannels() 
    - Connect()
       -  Hits"
   ```go
	if w.IsConnected() {
		return fmt.Errorf("%v %w", w.exchangeName, errAlreadyConnected)
	}
    ```

My guess is the connection monitor kicks in too quickly. Please just sanity check to see that its not going to cause problems for things. For the test itself is likely to increase the timer

exchanges/stream/websocket.go
```diff
@@ -410,15 +573,21 @@ func (w *Websocket) connectionMonitor() error {
...
 						}
 					}
 				}
+				// Speedier reconnection, instead of waiting for the next cycle.
+				if w.IsEnabled() && (!w.IsConnected() && !w.IsConnecting()) {
+					if connectErr := w.Connect(); connectErr != nil {
+						log.Errorln(log.WebsocketMgr, connectErr)
+					}
+				}
+				w.DataHandler <- err // hand over the error to the data handler (shutdown and reconnection is priority)
```
gloriousCode, line 590: The test issue could be from here. The FlushChannels calls `Shutdown`, should a connectionmonitor restart a connection from there?

gloriousCode (changes requested):
I'm pretty happy now with this work 🌈 Just one thing came up in recent testing

exchanges/stream/websocket_test.go
gloriousCode: It also comes up in Actions: 
https://github.com/thrasher-corp/gocryptotrader/actions/runs/10950367635/job/30405470763?pr=1580#step:8:313
https://github.com/thrasher-corp/gocryptotrader/actions/runs/10950367635/job/30405471659?pr=1580#step:10:389

gloriousCode: Another Actions test death:
https://github.com/thrasher-corp/gocryptotrader/actions/runs/10950367635/job/30405471508?pr=1580#step:10:395

exchanges/stream/websocket.go
```diff
@@ -1292,3 +1103,149 @@ func drain(ch <-chan error) {
...
+			}
+		}
+		t.Reset(w.connectionMonitorDelay)
+	}
+	return false
+}
+
+// monitorTraffic monitors to see if there has been traffic within the trafficTimeout time window. If there is no traffic
+// the connection is shutdown and will be reconnected by the connectionMonitor routine.
+func (w *Websocket) monitorTraffic() func() bool {
```
gloriousCode, line 1212: I tried simulating a disconnect on Binance and got the following:
```
panic: close of closed channel

goroutine 950 [running]:
github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*WebsocketConnection).Shutdown(0x14000b12000)
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket_connection.go:275 +0x70
github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).shutdown(0x14000264800)
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:553 +0x364
github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).Shutdown(0x0?)
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:512 +0xb8
github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).observeTraffic.func1()
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:1233 +0x24
created by github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).observeTraffic in goroutine 427
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:1232 +0x1b4
exit status 2
```

This does not occur on `master` branch. I have not been able to re-do this

gloriousCode: I had a different recreation on closing the application:
```
^C[INFO]  | SYNC | 30/09/2024 12:09:56 | GateIO websocket BTC-USDT SPOT ORDERBOOK: Bids len: 1 Amount: 2.933270 BTC. Total value: 189864.407233 Asks len: 1 Amount: 0.045910 BTC. Total value: 2971.66248
[INFO]  | LOG | 30/09/2024 12:09:56 | Captured interrupt, shutdown requested.
[DEBUG] | LOG | 30/09/2024 12:09:56 | Engine shutting down..
[DEBUG] | ORDER | 30/09/2024 12:09:56 | Order manager shutting down...
[DEBUG] | PORTFOLIO | 30/09/2024 12:09:56 | Portfolio manager shutting down...
[DEBUG] | CONNECTION | 30/09/2024 12:09:56 | Connection manager shutting down...
[DEBUG] | CONNECTION | 30/09/2024 12:09:56 | Connection manager stopped.
[DEBUG] | PORTFOLIO | 30/09/2024 12:09:56 | Portfolio manager shutdown.
[DEBUG] | ORDER | 30/09/2024 12:09:56 | Order manager shutdown.
[DEBUG] | DISPATCH | 30/09/2024 12:09:56 | Dispatch manager shutting down...
[DEBUG] | DISPATCH | 30/09/2024 12:09:56 | Dispatch manager shutdown
[DEBUG] | EXCHANGE | 30/09/2024 12:09:56 | Currency state manager shutting down...
[DEBUG] | EXCHANGE | 30/09/2024 12:09:56 | Currency state manager shutdown.
panic: close of closed channel

goroutine 677 [running]:
github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*WebsocketConnection).Shutdown(0x14000c06000)
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket_connection.go:275 +0x70
github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).shutdown(0x140001fc800)
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:553 +0x364
github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).Shutdown(0x14000406be0?)
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/stream/websocket.go:512 +0xb8
github.com/thrasher-corp/gocryptotrader/exchanges.(*Base).Shutdown(0x1400035a008)
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/exchange.go:1644 +0x2c
github.com/thrasher-corp/gocryptotrader/engine.(*ExchangeManager).Shutdown.func1(0x14000cfe040, 0x14000cfe030, {0x1017f2e20, 0x1400035a008})
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/engine/exchange_manager.go:159 +0x3c
created by github.com/thrasher-corp/gocryptotrader/engine.(*ExchangeManager).Shutdown in goroutine 1
	/Users/WIF/go/src/github.com/thrasher-corp/gocryptotrader/engine/exchange_manager.go:158 +0x1f8
exit status 2
```

gloriousCode (changes requested):
Thanks for addressing those issues! I think there's a potential double-shut/close happening

gloriousCode (approved):
tACK!

exchanges/stream/websocket.go
```diff
@@ -37,7 +37,6 @@ var (
 var (
 	errAlreadyRunning                       = errors.New("connection monitor is already running")
```
thrasher-, line 38: No longer in use

exchanges/stream/websocket.go
```diff
@@ -53,6 +52,7 @@ var (
 	errWebsocketSubscriberUnset             = errors.New("websocket subscriber function needs to be set")
 	errWebsocketUnsubscriberUnset           = errors.New("websocket unsubscriber functionality allowed but unsubscriber function not set")
 	errWebsocketConnectorUnset              = errors.New("websocket connector function not set")
+	errWebsocketDataHandlerUnset            = errors.New("websocket data handler not set")
 	errReadMessageErrorsNil                 = errors.New("read message errors is nil")
 	errWebsocketSubscriptionsGeneratorUnset = errors.New("websocket subscriptions generator function needs to be set")
 	errClosedConnection                     = errors.New("use of closed network connection")
```
thrasher-, line 58: No longer in use

thrasher- (changes requested):
Been testing this for a few hours and appears to be working well! Just some unused vars with the changes of this PR

thrasher- (approved):

PR #1581, opened 2024-07-17
stream/match: Reduce complexity and limit locking when match occurs

shazbert:
# PR Description

Inline lock and cleanup when match occurs, reduce the need to lock and close unless there is a timeout. 
* Adds context cancellation handling and cleanup.
* Adds in @gbjk changes to the mix, because prior did not work for all cases Thankyou and appreciate 🙏 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/stream/websocket_connection.go
```diff
@@ -22,34 +22,32 @@ import (
 // SendMessageReturnResponse will send a WS message to the connection and wait
 // for response
 func (w *WebsocketConnection) SendMessageReturnResponse(signature, request interface{}) ([]byte, error) {
```
gloriousCode, line 24: I have an extra request. Could you update `TestSendMessageReturnResponse` to the following?
```go
func TestSendMessageReturnResponse(t *testing.T) {
	t.Parallel()
	wc := &WebsocketConnection{
		Verbose:          true,
		URL:              "wss://ws.kraken.com",
		ResponseMaxLimit: time.Second * 5,
		Match:            NewMatch(),
	}
	if wc.ProxyURL != "" && !useProxyTests {
		t.Skip("Proxy testing not enabled, skipping")
	}

	err := wc.Dial(&dialer, http.Header{})
	if err != nil {
		t.Fatal(err)
	}

	go readMessages(t, wc)

	request := testRequest{
		Event: "subscribe",
		Pairs: []string{currency.NewPairWithDelimiter("XBT", "USD", "/").String()},
		Subscription: testRequestData{
			Name: "ticker",
		},
		RequestID: wc.GenerateMessageID(false),
	}

	_, err = wc.SendMessageReturnResponse(request.RequestID, request)
	if err != nil {
		t.Error(err)
	}

	// with timeout
	wc.ResponseMaxLimit = 1
	_, err = wc.SendMessageReturnResponse("123", request)
	assert.ErrorIs(t, err, errMatchTimeout, "SendMessageReturnResponse should error when request ID not found")
}
```
Where errMatchTimeout is
```go
var errMatchTimeout = errors.New("websocket connection: timeout waiting for response with signature")
```

gloriousCode (changes requested):
I like the simplification, though it is slower, but lighter overall.
Here's my benchmarks (in binance_test.go).
```go
func BenchmarkButts(bb *testing.B) {
	mock := func(msg []byte, w *websocket.Conn) error {
		return b.wsHandleData([]byte(`{"stream":"btcusdt@kline_1m","data":{}, "id":1337}`))
	}
	b := testexch.MockWsInstance[Binance](bb, testexch.CurryWsMockUpgrader(bb, mock))
	bb.ReportAllocs()
	for i := 0; i < bb.N; i++ {
		b.Websocket.Match.Set(1337)
		_, _ = b.Websocket.Conn.SendMessageReturnResponse(1337, []byte(`{"id":1337}`))
	}
}
```
```
// branch
BenchmarkButts-10    	 8486716	       131.0 ns/op	      64 B/op	       3 allocs/op
BenchmarkButts-10    	 8792042	       131.1 ns/op	      64 B/op	       3 allocs/op
BenchmarkButts-10    	 8709994	       131.4 ns/op	      64 B/op	       3 allocs/op
// master
BenchmarkButts-10    	11024931	       104.9 ns/op	      72 B/op	       4 allocs/op
BenchmarkButts-10    	11190993	       102.3 ns/op	      72 B/op	       4 allocs/op
BenchmarkButts-10    	11247758	        99.61 ns/op	      72 B/op	       4 allocs/op
```

exchanges/stream/websocket_connection.go
gloriousCode: Hello Shazbert, thank you for making those changes, however please rename the test as it is out-of-sync with the function name

gloriousCode (approved):
tACK! 🌈

exchanges/stream/stream_match.go
```diff
@@ -5,76 +5,60 @@ import (
...
 	m.mu.Lock()
+	defer m.mu.Unlock()
 	if _, ok := m.m[signature]; ok {
-		m.mu.Unlock()
-		return Matcher{}, errors.New("signature collision")
+		return nil, errSignatureCollision
 	}
-	// This is buffered so we don't need to wait for receiver.
-	ch = make(chan []byte, 1)
+	ch := make(chan []byte, 1) // This is buffered so we don't need to wait for receiver.
```
gbjk, line 51: #1587  ( and it's 3 parent PRs 😂 ) make this buffer configurable, so that we can capture a stream of matches for an ID.
This happens with Kraken returning N+ subscription responses with the same ID to a single request.
We know how many we want to expect.

exchanges/stream/stream_match.go
```diff
@@ -5,76 +5,60 @@ import (
...
-			return false
-		}
-		return true
+	if !ok {
+		return false
 	}
-	return false
+	ch <- data
+	close(ch)
+	delete(m.m, signature)
```
gbjk, line 40: Related to the comment below, we can't close it and delete it if we're allowing N+1

gbjk (changes requested):
Suggest we merge the 1587 changes into this.
Would that work for you?
It's just the commit 65dd631

exchanges/stream/stream_match.go
```diff
@@ -5,76 +5,60 @@ import (
...
-	C   chan []byte
-	sig interface{}
-	m   *Match
-}
-
 // Incoming matches with request, disregarding the returned payload
-func (m *Match) Incoming(signature interface{}) bool {
+func (m *Match) Incoming(signature any) bool {
 	return m.IncomingWithData(signature, nil)
 }
```
gbjk, line 27: I think we should remove Incoming.
It's only used in tests.
You can always move it to test pkg I think.

exchanges/stream/stream_match.go
```diff
@@ -9,21 +9,22 @@ var errSignatureCollision = errors.New("signature collision")
...
+	m  map[any]Incoming
 	mu sync.Mutex
 }
 
-// Incoming matches with request, disregarding the returned payload
-func (m *Match) Incoming(signature any) bool {
-	return m.IncomingWithData(signature, nil)
+// Incoming is a sub-type that handles incoming data
+type Incoming struct {
+	count          int // Number of responses expected
```
gbjk, line 26: Nit: Maybe `expected` is better?
Because it'd still be truthful as it decreases.
And count is the opposite.

And then you wouldn't need the comment, which is what caused me to even read this closely.

exchanges/stream/stream_match.go
```diff
@@ -9,21 +9,22 @@ var errSignatureCollision = errors.New("signature collision")
...
+	waitingRoutine chan<- []byte
```
gbjk, line 27: This isn't a waitingRoutine; We actually don't know anything at all about the consumers/readers, which is good.

So I think we should just call it `c`, which has a lot of precedent in go already.

exchanges/stream/stream_match.go
```diff
@@ -35,21 +36,27 @@ func (m *Match) IncomingWithData(signature any, data []byte) bool {
 	if !ok {
 		return false
 	}
-	ch <- data
-	close(ch)
-	delete(m.m, signature)
+	ch.waitingRoutine <- data
+	ch.count--
+	if ch.count == 0 {
```
gbjk, line 41: Nit:
```
if ch.count--; ch.count == 0 {
```

exchanges/stream/stream_match.go
```diff
@@ -35,21 +36,27 @@ func (m *Match) IncomingWithData(signature any, data []byte) bool {
...
+		close(ch.waitingRoutine)
+		delete(m.m, signature)
+	}
 	return true
 }
 
 // Set the signature response channel for incoming data
-func (m *Match) Set(signature any) (<-chan []byte, error) {
+func (m *Match) Set(signature any, bufSize int) (<-chan []byte, error) {
```
gbjk, line 49: I was thinking we might want to uint this to avoid the below

exchanges/stream/stream_match.go
shazbert: Good point!

gbjk (commented):
Minor nits.

I think the thing that needs addressing are: erroring on < 0 size (or make it a uint).

Everything else is subjective.

Aside:
I find uints annoying when I'm a library user. But as an author it makes so much more sense to just type safe the values you want to use.

exchanges/stream/stream_match.go
shazbert: 🔥 yesssssssss!

exchanges/stream/stream_match.go
gbjk: You can't call it with a buf size of 0.
It'd mean you're asking for a blocking channel, right?
We wouldn't want someone to be able to do that would we? ( I really don't know. I think we don't want them to ).
Assuming that, we should error on values of <= 0 and we can accept ints

exchanges/stream/stream_match.go
```diff
@@ -35,21 +36,27 @@ func (m *Match) IncomingWithData(signature any, data []byte) bool {
...
+	if bufSize < 0 {
+		bufSize = 1
+	}
```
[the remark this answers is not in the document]
gbjk: And we should error on 0 as well. Callers shouldn't be able to make a blocking channel, should they?

exchanges/stream/stream_match.go
```diff
@@ -5,26 +5,28 @@ import (
 	"sync"
 )
 
-var errSignatureCollision = errors.New("signature collision")
+var (
+	errSignatureCollision            = errors.New("signature collision")
+	errBufferShouldBeGreaterThanZero = errors.New("buffer size should be greater than zero")
```
gbjk, line 10: Subjective comment:
```suggestion
	errInvalidBufferSize = errors.New("buffer size must be positive")
```

exchanges/stream/stream_match_test.go
```diff
@@ -15,6 +15,8 @@ func TestMatch(t *testing.T) {
 	match := NewMatch()
 	assert.False(t, match.IncomingWithData("hello", load), "Should not match an empty signature")
 
+	_, err := match.Set("hello", -0)
+	require.ErrorIs(t, err, errBufferShouldBeGreaterThanZero, "Should error on buffer size less than 0")
```
gbjk, line 19:
```suggestion
	_, err := match.Set("hello", 0)
	require.ErrorIs(t, err, errInvalidBufferSize, "Must error on zero buffer size")
	_, err := match.Set("hello", -1)
	require.ErrorIs(t, err, errInvalidBufferSize, "Must error on negative buffer size")
```

gbjk (approved):
Approved, but with suggestions.
`-0` is a bit weird though

exchanges/stream/stream_match.go
```diff
@@ -5,76 +5,69 @@ import (
...
 }
 
-// Cleanup closes underlying channel and deletes signature from map
-func (m *Matcher) Cleanup() {
-	m.m.mu.Lock()
-	close(m.C)
-	delete(m.m.m, m.sig)
-	m.m.mu.Unlock()
+// RemoveSignature removes the signature response from map and closes the channel.
+func (m *Match) RemoveSignature(signature any) {
```
gloriousCode, line 66: Please add test 😄

exchanges/stream/stream_match.go
```diff
@@ -5,76 +5,69 @@ import (
...
+	return &Match{m: make(map[any]incoming)}
 }
 
 // Match is a distributed subtype that handles the matching of requests and
 // responses in a timely manner, reducing the need to differentiate between
 // connections. Stream systems fan in all incoming payloads to one routine for
 // processing.
 type Match struct {
-	m  map[interface{}]chan []byte
+	m  map[any]incoming
```
gloriousCode, line 23:
```suggestion
	m  map[any]*incoming
```

Please look over your test `TestMatch` again. I've snipped irrelevant bits

```go
	ch, err := match.Set("hello", 2)
	assert.True(t, match.IncomingWithData("hello", []byte("hello")))
	assert.Equal(t, "hello", string(<-ch))
	assert.True(t, match.IncomingWithData("hello", load), "Should match with matching message and signature")
	assert.True(t, match.IncomingWithData("hello", load), "Should match with matching message and signature")

	assert.Len(t, ch, 2, "Channel should have 2 items")
```

Note that by not having a pointer in your map, `expected` never sets its `expected--`, so the _second_ time it goes in, `expected` is still `2` despite having a match already.

The scenario you've made in the test should not be able to occur because the signature should already be deleted by the third time you call `hello`

gloriousCode (changes requested):
I like the idea of the expected feature update, just needs a fix

exchanges/stream/stream_match.go
gloriousCode: I would still like a test

gloriousCode (changes requested):

exchanges/stream/stream_match.go
shazbert:
```go
	ch, err = match.Set("masterblaster", 1)
	require.NoError(t, err)
	match.RemoveSignature("masterblaster")
	garbage := <-ch // Should be closed and super slipery
	require.Empty(t, garbage)
```

Was [added in to satisfy the test](https://github.com/thrasher-corp/gocryptotrader/pull/1581/commits/ee8a35c42d8e1fec5070600356acd4905e23970a), so if the signature wasn't removed this would hang indefinitely. Would you like me to break it out further into its own test function?

gloriousCode: Yes please

shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1581/commits/44a3461eb09d183914463795d5695db5f83082b5)

exchanges/stream/stream_match_test.go
```diff
@@ -1,50 +1,53 @@
...
-		t.Fatal("Should not have matched")
-	}
-
-	nm := NewMatch()
-	// try to match with unset signature
-	if nm.Incoming("hello") {
-		t.Fatal("should not be able to match")
-	}
+	load := []byte("42")
+	assert.False(t, new(Match).IncomingWithData("hello", load), "Should not match an uninitilized Match")
```
gloriousCode, line 13:
```suggestion
	assert.False(t, new(Match).IncomingWithData("hello", load), "Should not match an uninitialised Match")
```

```suggestion
	assert.False(t, new(Match).IncomingWithData("hello", load), "Should not match an uninitialized Match")
```

gloriousCode (changes requested):
Thank you for testing your function. One tiny choose-your-own-adventure nit and then tACK

gloriousCode (approved):

thrasher- (approved):
tACK

PR #1582, opened 2024-07-17
Subscriptions: Fix subscription.Clone Params and Pairs

gbjk:
Fixes Clone() affecting the original and not the clone for Pairs and Params.
That actually meant it would pass tests, and kinda work, but not in the way expected, and definitely not thread safe on the original parent

## Type of change

### DIFF
--- a/exchanges/subscription/subscription.go
+++ b/exchanges/subscription/subscription.go
@@ -132,16 +132,14 @@ func (s *Subscription) Clone() *Subscription {
 		Enabled:          s.Enabled,
 		Channel:          s.Channel,
 		Asset:            s.Asset,
-		Params:           s.Params,
+		Params:           maps.Clone(s.Params),
 		Interval:         s.Interval,
 		Levels:           s.Levels,
 		Authenticated:    s.Authenticated,
 		state:            s.state,
-		Pairs:            s.Pairs,
+		Pairs:            slices.Clone(s.Pairs),
 		QualifiedChannel: s.QualifiedChannel,
 	}
-	c.Pairs = slices.Clone(s.Pairs)
-	c.Params = maps.Clone(s.Params)
 	s.m.RUnlock()
 	return c
 }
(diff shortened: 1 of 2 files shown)


exchanges/subscription/subscription.go
```diff
@@ -140,8 +140,8 @@ func (s *Subscription) Clone() *Subscription {
 		Pairs:            s.Pairs,
 		QualifiedChannel: s.QualifiedChannel,
```
gloriousCode, line 141:
```suggestion

```
Removes the double-assign and potentially what allowed us to get confused

gloriousCode (changes requested):
A little whoopsie that went unnoticed 🌈

exchanges/subscription/subscription.go
```diff
@@ -140,8 +140,8 @@ func (s *Subscription) Clone() *Subscription {
 		Pairs:            s.Pairs,
```
gloriousCode, line 140:
```suggestion
```
Removes the double-assign and potentially what allowed us to get confused.
Do the same for `Params` too

gloriousCode (approved):
tACK!

exchanges/subscription/subscription.go
```diff
@@ -132,16 +132,14 @@ func (s *Subscription) Clone() *Subscription {
...
 		Interval:         s.Interval,
 		Levels:           s.Levels,
 		Authenticated:    s.Authenticated,
 		state:            s.state,
-		Pairs:            s.Pairs,
+		Pairs:            slices.Clone(s.Pairs),
 		QualifiedChannel: s.QualifiedChannel,
 	}
-	s.Pairs = slices.Clone(s.Pairs)
-	s.Params = maps.Clone(s.Params)
```
shazbert, line 144: 😬

shazbert (approved):
Nice fix! tACK.

thrasher- (approved):
Nice gotcha fix + additional test coverage

PR #1583, opened 2024-07-18
Okx: Fix TestGetBlockTrade transient fail

gbjk:
Last block trade has fallen outside 7 days many times now. Switching to within 90 days, which appears safe.


gloriousCode (approved):
Not a fault of our implementation that no one wants to block trade. 90 days is serviceable, but I'd also accept the removal of the assert or a change to `NotEmpty`

gbjk:
> Not a fault of our implementation that no one wants to block trade. 90 days is serviceable, but I'd also accept the removal of the assert or a change to `NotEmpty`

Agreed. I considered NotEmpty. Wanted to keep *some* sanity validation in there, though.

shazbert (approved):
utACK.

thrasher- (approved):
Poor OKX, looks like this feature never really took off

PR #1586, opened 2024-07-21
Bitmex: Subscription templating

gbjk:
### Bitmex

* Add subscription templating
* Upgrade test config
  * Remove deprecated Upside/Downside Profit contracts

## Stacked Dependencies

- [x] #1579 

## Type of change

### DIFF
--- a/exchanges/bitmex/bitmex_websocket.go
+++ b/exchanges/bitmex/bitmex_websocket.go
@@ -11,7 +11,9 @@ import (
 	"text/template"
 	"time"
 
+	"github.com/buger/jsonparser"
 	"github.com/gorilla/websocket"
+	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/common/crypto"
 	"github.com/thrasher-corp/gocryptotrader/currency"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
@@ -25,7 +27,7 @@ import (
 )
 
 const (
-	bitmexWSURL = "wss://www.bitmex.com/realtime"
+	bitmexWSURL = "wss://www.bitmex.com/realtimemd"
 
 	// Public Subscription Channels
 	bitmexWSAnnouncement        = "announcement"
@@ -68,13 +70,11 @@ const (
 )
 
 var defaultSubscriptions = subscription.List{
-	{Enabled: true, Channel: bitmexWSAnnouncement},
 	{Enabled: true, Channel: bitmexWSOrderbookL2, Asset: asset.All},
 	{Enabled: true, Channel: bitmexWSTrade, Asset: asset.All},
 	{Enabled: true, Channel: bitmexWSAffiliate, Authenticated: true},
 	{Enabled: true, Channel: bitmexWSOrder, Authenticated: true},
 	{Enabled: true, Channel: bitmexWSMargin, Authenticated: true},
-	{Enabled: true, Channel: bitmexWSPrivateNotifications, Authenticated: true},
 	{Enabled: true, Channel: bitmexWSTransact, Authenticated: true},
 	{Enabled: true, Channel: bitmexWSWallet, Authenticated: true},
 	{Enabled: true, Channel: bitmexWSExecution, Authenticated: true, Asset: asset.PerpetualContract},
@@ -87,40 +87,50 @@ func (b *Bitmex) WsConnect() error {
 		return stream.ErrWebsocketNotEnabled
 	}
 	var dialer websocket.Dialer
-	err := b.Websocket.Conn.Dial(&dialer, http.Header{})
-	if err != nil {
+	if err := b.Websocket.Conn.Dial(&dialer, http.Header{}); err != nil {
 		return err
-	}
-
-	resp := b.Websocket.Conn.ReadMessage()
-	if resp.Raw == nil {
-		return errors.New("connection closed")
-	}
-	var welcomeResp WebsocketWelcome
-	err = json.Unmarshal(resp.Raw, &welcomeResp)
-	if err != nil {
-		return err
-	}
-
-	if b.Verbose {
-		log.Debugf(log.ExchangeSys,
-			"Successfully connected to Bitmex %s at time: %s Limit: %d",
-			welcomeResp.Info,
-			welcomeResp.Timestamp,
-			welcomeResp.Limit.Remaining)
 	}
 
 	b.Websocket.Wg.Add(1)
 	go b.wsReadData()
 
+	ctx := context.TODO()
+	if err := b.wsOpenStream(ctx, b.Websocket.Conn, wsPublicStream); err != nil {
+		return err
+	}
+
 	if b.Websocket.CanUseAuthenticatedEndpoints() {
-		err = b.websocketSendAuth(context.TODO())
-		if err != nil {
+		if err := b.websocketSendAuth(ctx); err != nil {
 			b.Websocket.SetCanUseAuthenticatedEndpoints(false)
 			log.Errorf(log.ExchangeSys, "%v - authentication failed: %v\n", b.Name, err)
 		}
 	}
 
+	return nil
+}
+
+const (
+	wsPublicStream  = "public"
+	wsPrivateStream = "private"
+	wsSubscribeOp   = "subscribe"
+	wsUnsubscribeOp = "unsubscribe"
+	wsMsgPacket     = 0
+	wsOpenPacket    = 1
+	wsClosePacket   = 2
+)
+
+func (b *Bitmex) wsOpenStream(ctx context.Context, c stream.Connection, name string) error {
+	resp, err := c.SendMessageReturnResponse(ctx, request.Unset, "open:"+name, []any{wsOpenPacket, name, name})
+	if err != nil {
+		return err
+	}
+	var welcomeResp WebsocketWelcome
+	if err := json.Unmarshal(resp, &welcomeResp); err != nil {
+		return err
+	}
+	if b.Verbose {
+		log.Debugf(log.ExchangeSys, "Successfully connected to Bitmex %s websocket API at time: %s Limit: %d", name, welcomeResp.Timestamp, welcomeResp.Limit.Remaining)
+	}
 	return nil
 }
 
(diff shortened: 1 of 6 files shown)


gbjk:
Test enabled pairs aren't stable.
Will fix in the morning. Not a big thing so leaving this as "reviewable".

shazbert (approved):
tACK.

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -65,9 +66,18 @@ const (
 	bitmexActionUpdateData  = "update"
 )
 
-var subscriptionNames = map[string]string{
-	subscription.OrderbookChannel: bitmexWSOrderbookL2,
-	subscription.AllTradesChannel: bitmexWSTrade,
+var defaultSubscriptions = subscription.List{
+	{Enabled: true, Channel: bitmexWSAnnouncement},
```
gloriousCode, line 70:
```
[DEBUG] | WEBSOCKET | 06/09/2024 09:12:01 | Bitmex websocket connection: sending message: {"op":"subscribe","args":["announcement","orderBookL2:UNI_USDT","orderBookL2:ETHUSD","orderBookL2:XBTH25","trade:UNI_USDT","trade:ETHUSD","trade:XBTH25"]}
[DEBUG] | WEBSOCKET | 06/09/2024 09:12:01 | Bitmex websocket connection: message received: {"status":400,"error":"Topic announcement is not available on this endpoint. Please see the documentation at https://www.bitmex.com/app/wsAPI.","meta":{},"request":{"op":"subscribe","args":["announcement","orderBookL2:UNI_USDT","orderBookL2:ETHUSD","orderBookL2:XBTH25","trade:UNI_USDT","trade:ETHUSD","trade:XBTH25"]}}
```
It highlights that we do not recognise WS errors properly since it comes through as `DEBUG`

gloriousCode:
```
[DEBUG] | WEBSOCKET | 06/09/2024 09:16:45 | Bitmex websocket connection: message received: {"status":400,"error":"Unknown table: privateNotifications. Please see the documentation at https://www.bitmex.com/app/wsAPI.","meta":{},"request":{"op":"subscribe","args":["affiliate","order","margin","privateNotifications","transact","wallet","execution:ETHUSD","position:ETHUSD"]}}
```

https://www.bitmex.com/app/wsAPI

> The following subscriptions are only available on the /realtimePlatform endpoint:
> "announcement",        // Site announcements
> "privateNotifications", // Individual notifications - currently not used (requires auth)

We are not connecting to the additional endpoint for those subscriptions. I am not asking you to set that up, but I don't think those two topics should be part of this PR

Since these messages came through as `DEBUG` I get how they were missed

gloriousCode (changes requested):

exchanges/bitmex/bitmex_websocket.go
gbjk: Fixed 0db29723

* Announcement and Private are no longer included
* Switched to using the multiplexer to make it safe to get auth and stream failures out with separately without racing
* Error handling, including auth

Had to rebase onto master to get `SendMessageReturnResponses` unfortunately.

gbjk:
Leaving the reconstructing label since I've broken many of the tests with the rewrite.

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -141,338 +152,321 @@ func (b *Bitmex) wsReadData() {
 }
 
 func (b *Bitmex) wsHandleData(respRaw []byte) error {
-	quickCapture := make(map[string]interface{})
-	err := json.Unmarshal(respRaw, &quickCapture)
+	// We don't need to know about errors, since we're using looking optimistically into the json
```
gloriousCode, line 155: It looks like there are missing words, or I've had too much caffeine and can't parse your comment
Plus you define and look at the error

Perhaps this rephrasing is for below?

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -141,338 +152,321 @@ func (b *Bitmex) wsReadData() {
...
+		return fmt.Errorf("unknown message format: %s", respRaw)
+	}
+	op, _ := jsonparser.GetString(msg, "request", "op")
+	errMsg, _ := jsonparser.GetString(msg, "error")
+	success, _ := jsonparser.GetBoolean(msg, "success")
+	version, _ := jsonparser.GetString(msg, "version")
+	switch {
+	case version != "":
+		op = "open"
+		fallthrough
```
gloriousCode, line 167: This fallthrough tripped me up because `errMsg` is `""` and `success` is `false` when this happens. That is on me, I just wanted to tell a cool story

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -24,7 +27,7 @@ import (
 )
 
 const (
-	bitmexWSURL = "wss://www.bitmex.com/realtime"
+	bitmexWSURL = "wss://www.bitmex.com/realtimemd"
```
gloriousCode, line 30: I think the only thing that this change highlights is that we're not using the URL system for `b.Websocket.SetupNewConnection` where we are using `bitmexWSURL` instead of `wsEndpoint`.
I bring this up because I received this warning from having an outdated URL in my config:
```
[WARN]  | EXCHANGE | 03/10/2024 11:30:20 | Bitmex: Config is overwriting default endpoint URL values from: [wss://www.bitmex.com/realtimemd] to: [wss://www.bitmex.com/realtime] for: [WebsocketSpotURL]
```
but then we don't do anything with it anyway. 

We don't use the functionality consistently anyway. The feature is there if an endpoint goes down we can use a different one without code-change.
So I will highlight it, see if anyone cares enough to make you do it now, or its something for the pile 🌈

gloriousCode (changes requested):
Thankyou for making those changes, just one thing about a comment

exchanges/bitmex/bitmex_websocket.go
gbjk: Fixed 5d0bc5d2

exchanges/bitmex/bitmex_websocket.go
gbjk: It's on my pile.
As I see it, this happens a lot, both in this context of urls and others, where the default value changes, but the user's left with the old default in local config.
One of the things we don't do well right now is to separate out default config from user config in a way that we don't break the user's config when we change defaults, since we flatten them down to the user's file indiscriminately.
I have a plan for this.

gloriousCode (approved):
Excellent I am happy. tACK

gbjk:
Rebased 🔁

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -131,338 +152,321 @@ func (b *Bitmex) wsReadData() {
...
 		if err != nil {
 			return err
 		}
-		switch decodedResp.Table {
-		case bitmexWSOrderbookL2, bitmexWSOrderbookL225, bitmexWSOrderbookL10:
-			var orderbooks OrderBookData
-			err = json.Unmarshal(respRaw, &orderbooks)
+		var trades []trade.Data
+		for i := range tradeHolder.Data {
+			if tradeHolder.Data[i].Price == 0 {
```
shazbert, line 218:
```suggestion
			if tradeHolder.Data[i].Size == 0 {
```

shazbert: Just going off from the commentary below

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -131,338 +152,321 @@ func (b *Bitmex) wsReadData() {
...
 			}
-			if len(orderbooks.Data) == 0 {
-				return fmt.Errorf("%s - Empty orderbook data received: %s", b.Name, respRaw)
+			var oSide order.Side
+			oSide, err = order.StringToOrderSide(tradeHolder.Data[i].Side)
+			if err != nil {
+				b.Websocket.DataHandler <- order.ClassificationError{
+					Exchange: b.Name,
+					Err:      err,
+				}
```
shazbert, line 235:
```suggestion
				return err
```

Usually only for actual orders matched with your account.

shazbert (changes requested):
Good stuff; just noticed these. everything else looks great. Thanks!

exchanges/bitmex/bitmex_websocket.go
gbjk: Good catch 👍 
OutOfScope: I'll handle this separately 😄 
I'm going to be a bit stricter on not contaminating unrelated fixes, especially not into features.

exchanges/bitmex/bitmex_websocket.go
gbjk: Fixed 3377e351

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -131,338 +152,292 @@ func (b *Bitmex) wsReadData() {
 }
 
 func (b *Bitmex) wsHandleData(respRaw []byte) error {
-	quickCapture := make(map[string]interface{})
-	err := json.Unmarshal(respRaw, &quickCapture)
+	var err error
```
gbjk, line 155: All linter placations are distasteful when it comes to err; ( _But this is my one_ ).
This one is just the least intrusive way of dealing with `gocritic.sloppyReassign` and `govet.shadow` fighting over whether it's simultaneously right and wrong to shadow a var.

exchanges/bitmex/bitmex_websocket.go
gbjk: Fixed #1685

shazbert (approved):
This is a tACK for current changes.

Out of scope issue when updating tradable pairs for index asset that I am highlighting it wigs out subs for some reason:

```
[ERROR] | WEBSOCKET | 22/10/2024 15:06:26 | subscription failure: Unknown or expired symbol: RBNUSDTPI8H. Please see the documentation at https://www.bitmex.com/app/wsAPI., Unknown or expired symbol: RBNUSDTPI8H. Please see the documentation at https://www.bitmex.com/app/wsAPI.
```

exchanges/bitmex/bitmex_websocket.go
```diff
@@ -66,46 +69,39 @@ const (
...
+	{Enabled: true, Channel: bitmexWSMargin, Authenticated: true},
+	{Enabled: true, Channel: bitmexWSTransact, Authenticated: true},
+	{Enabled: true, Channel: bitmexWSWallet, Authenticated: true},
+	{Enabled: true, Channel: bitmexWSExecution, Authenticated: true, Asset: asset.PerpetualContract},
+	{Enabled: true, Channel: bitmexWSPosition, Authenticated: true, Asset: asset.PerpetualContract},
 }
 
 // WsConnect initiates a new websocket connection
 func (b *Bitmex) WsConnect() error {
+	ctx := context.TODO()
```
thrasher-, line 86: Can move this down to before `b.wsOpenStream` so it's used when it's actually needed instead of here

thrasher- (changes requested):
Tested auth and unauth, unfortunately ran into this issue https://github.com/thrasher-corp/gocryptotrader/issues/1690 when testing but doesn't have anything to do with this PR.

Just one minor change and then this looks good to go. Nice stuff!

exchanges/bitmex/bitmex_websocket.go
gbjk: Agreed.
I'm usually putting it at the top because it *should* be passed in, and declaring it is kind of representative of that.
But equally it's wasteful to declare it before error returns.
So I've flip flopped a few times between: Up at the top, After early returns, and Just In Time.
I'll move it to where it's used here.

Fixed 9f8d70449

Immediate squash incoming.

thrasher- (approved):
Thanks!

PR #1589, opened 2024-07-25
stream/websocketconnection: drop and consolidate duplicate fields

shazbert:
# PR Description

* Drop fields associated with the websocket and pass reference to governing type.
* Drop method set proxy as proxy is held and maintained at the websocket level anyway
* Adds error to ping handler message output
* Shifts wg.Done to inside the go routine to minimise leak chance
* Privatise all websocket connection fields

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/stream/websocket_connection.go
```diff
@@ -12,17 +12,32 @@ import (
...
+// WebsocketConnection contains all the data needed to send a message to a WS
+// connection
+type WebsocketConnection struct {
+	// Gorilla websocket does not allow more than one goroutine to utilise writes methods
+	writeControl     sync.Mutex
+	connected        int32
+	connection       *websocket.Conn
+	rateLimit        int64
+	_URL             string
+	sharedContext    *Websocket
```
gloriousCode, line 32: Sorry I don't like `sharedContext`, doesn't seem fitting. `parentWebsocket` or `parent`

gloriousCode (changes requested):
Just one comment on naming. While in the context of this PR it all looks fine, I just want to make sure this change doesn't have other consequences on my next pass

exchanges/stream/websocket_connection.go
shazbert: Yeah good suggestion will change, but why you give me 2 choices, because I will always pick the less length one. 😆

exchanges/stream/websocket.go
```diff
@@ -232,18 +232,11 @@ func (w *Websocket) SetupNewConnection(c ConnectionSetup) error {
 	}
 
 	newConn := &WebsocketConnection{
-		ExchangeName:      w.exchangeName,
-		URL:               connectionURL,
-		ProxyURL:          w.GetProxyAddress(),
-		Verbose:           w.verbose,
-		ResponseMaxLimit:  c.ResponseMaxLimit,
-		Traffic:           w.TrafficAlert,
```
gloriousCode, line 240: I'm leaving this as just a comment for now. I think a TrafficAlert from parent only seems like a problematic decision for next steps. It feels like things like Traffic and Wg should be moving from `w` to the individual connection instead of a shared point

exchanges/stream/websocket_connection.go
```diff
@@ -202,13 +211,13 @@ func (w *WebsocketConnection) ReadMessage() Response {
 				// method on WebsocketConnection type has been called and can
 				// be skipped.
 				select {
-				case w.readMessageErrors <- err:
+				case w.parent.ReadMessageErrors <- err:
```
gloriousCode, line 214: This as well, if you're moving towards a `w` holding many `WebsocketConnections`, sharing `ReadMessageErrors` across them will be very jank and confusing to debug errors

gloriousCode (commented):
I'm leaving these as comments. If you resolve them, I'll approve. While this doesn't impact things now, its concern for the other PRs/future things.

shazbert:
> I'm leaving these as comments. If you resolve them, I'll approve. While this doesn't impact things now, its concern for the other PRs/future things which we cannot see/review from this PR itself

All those comments are the reason why I made this change in the first place. The state hasn't changed at all, except there is now less obfuscation about where the references are pointing too. They were all sharing the same wait group, traffic and readmessage errors channel. 

For the first point about the waitgroups, right now if there is a problem with 1 they will all shutdown or connect as a group anyway so it "shouldn't" be a an issue even in further multi connection updates. 

For the traffic channel I do see your point there and have it reconnect at a single connection basis. That is very valid. Defer for a future PR. 

For the debugging errors issue, I can wrap the URL and exchange name to the error. I think I have in another PR. But will do that now anyway.

shazbert:
Most critical items where added in multi currency this can be closed as it's a time issue with reviewing and doesn't add anything in terms of expanded functionality or bug fixes.

PR #1592, opened 2024-07-25
Subscriptions: Add List.Clone

gbjk:
This PR just adds Subscription List cloning.

It's broken out from #1579 so that other PRs can move along without such a big dependency chain.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/subscription/list_test.go
+++ b/exchanges/subscription/list_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/currency"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/kline"
 )
 
 // TestListStrings exercises List.Strings()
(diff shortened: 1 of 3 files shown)


shazbert (approved):
tACK! Good stuff.

exchanges/subscription/list_test.go
```diff
@@ -91,3 +91,12 @@ func TestAssetPairs(t *testing.T) {
...
 	}
 }
+
+func TestListClone(t *testing.T) {
+	t.Parallel()
+	l := List{{Channel: TickerChannel}, {Channel: OrderbookChannel}}
+	n := l.Clone()
+	assert.NotSame(t, n, l, "Slices must not be the same")
+	require.NotEmpty(t, n, "List must not be empty")
+	assert.NotSame(t, n[0], l[0], "Subscriptions must be cloned")
```
gloriousCode, line 101: I think it would be worthwhile to prove in some capacity that you are actually cloning things, that something is equal

gloriousCode (changes requested):
Cool

gloriousCode (approved):
Thanks! tACK!

exchanges/subscription/list_test.go
gbjk: Fixed aace80f

thrasher- (approved):
ACK, nice!

PR #1593, opened 2024-07-29
currency: fix edge case on NewCode where all string characters are digits

shazbert:
# PR Description

* Forces formatting to be upper case when only digits found, majority of storage pairs are upper case anyway. So this is a quick implementation, we could go as far as tracking it by specifically having a secondary bool because all digits are agnostic to casing. But I can leave that up to the reviewers discretion if this needs to happen. 

* optimise pair difference and make sure formatting check is inlined to for loop. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


currency/pairs_test.go
```diff
@@ -903,3 +903,112 @@ func TestPairsEqual(t *testing.T) {
...
+
+	diff, err = original.FindDifferences(compare, PairFormat{Delimiter: DashDelimiter, Uppercase: true})
+	require.NoError(t, err)
+	require.False(t, diff.FormatDifference)
+}
+
+// 2208139	       509.3 ns/op	     288 B/op	       2 allocs/op (current)
+//
+// 1614865	       712.5 ns/op	     336 B/op	       8 allocs/op (prev)
+func BenchmarkFindDifferences(b *testing.B) {
```
gloriousCode, line 1003: Good stuff

currency/code_test.go
```diff
@@ -456,6 +458,14 @@ func TestBaseCode(t *testing.T) {
 	}
 }
 
+func TestNewCodeFormatting(t *testing.T) {
+	require.True(t, NewCode("BTC").UpperCase)
+	require.False(t, NewCode("btc").UpperCase)
+	require.True(t, NewCode("BTC").Equal(NewCode("btc")))
+	require.True(t, NewCode("420").UpperCase)
+	require.False(t, NewCode("btc420").UpperCase)
+}
```
gloriousCode, line 467: Just seeking clarification, is it that you wish for digits to always be upper case? 
One can change that with
```go
	butts := NewCode("420")
	hi := butts.Lower()
	t.Log(hi.UpperCase)
```

```
    code_test.go:470: false
```

gloriousCode (changes requested):

currency/code_test.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1593/commits/ed3a1a7c9955f83d1bffc2e33dd1f9f574035cca)

currency/pairs.go
```diff
@@ -233,54 +235,70 @@ func (p Pairs) GetMatch(pair Pair) (Pair, error) {
...
-			(!p[x].Base.IsEmpty() && p[x].Base.UpperCase != pairFmt.Uppercase) ||
-			(!p[x].Quote.IsEmpty() && p[x].Quote.UpperCase != pairFmt.Uppercase) {
+		if p[x].hasFormatDifference(pairFmt) {
 			return true
 		}
 	}
 	return false
 }
 
+func (p Pair) hasFormatDifference(pairFmt PairFormat) bool {
```
gloriousCode, line 296: Please put in `pair.go`

gloriousCode (changes requested):
huge critical nit

currency/code.go
```diff
@@ -187,9 +187,9 @@ func (b *BaseCodes) Register(c string, newRole Role) Code {
 	}
 
 	var format bool
-	// Digits fool upper and lower casing. So find first letter and check case.
 	for x := range c {
 		if unicode.IsLetter(rune(c[x])) {
+			// Digits fool upper and lower casing. So find first letter and check case.
 			format = unicode.IsUpper(rune(c[x]))
```
gbjk, line 193: Interesting that this would have a strange outcome on a mixed case code, like sETH.
Feeling like this could be simplifier to use strings.ContainsFunc(... unicode.IsLetter() && unicode.IsUpper() )
Then we'd consider it uppercase if anything in it is uppercase.

currency/code.go
```diff
@@ -275,21 +282,27 @@ func (c Code) String() string {
...
 }
 
 // Upper converts the code to uppercase formatting
 func (c Code) Upper() Code {
-	c.UpperCase = true
+	if c.Item != nil && !c.Item.CaseSensitive {
+		return c
+	}
+	c.upperCase = true
 	return c
```
gbjk, line 306:
```suggestion
// Upper flags the Code to use UpperCase formatting, but does not change Symbol
// If Code cannot be uppercased then it will return Code undchanged
func (c Code) Upper() Code {
	if c.Item == nil || c.Item.Symbol != c.Item.Lower {
		// Symbol is uppercased already, so if they're equal, then there's no uppercase representation
		return c
	}
	c.UpperCase = true
	return c
}
```

gbjk (commented):
I came to this review because `apiserver.TestConfigAllJsonResponse` was failing due to Pairs.Format() setting digits to be marked as uppercase, and not matching what was sent through json.
( LoadConfig > CheckConfig > CheckExchangeConfigValues > CheckPairConsistency > GetAvailablePairs > pairs.Format )

I felt when I saw it that this introduced more complexity than necessary, and a fallacy, because we're allowing the idea that it's Uppercase to pervade, and CaseSensitive feels like derivable information that's rarely needed.

Instead of this, I felt we could just leave everything exactly as it is throughout.
Register sets the Code to be `UpperCase: false` if it's a digit, which is fine.

The only time we're going to get into trouble is if `Upper()` does something to it later. So we just avoid that by having it return the same thing. Which is what you've done. We just do it without the extra field by looking if/when someone asks us to. And it's a very fast comparison.

The gotcha here is that if you try to `GetFormat` from a pair like that, it's going to return mixed-case. I think at this point we can again derive that it's not caseSensitive and avoid throwing the error.

I'm easy. I just felt we could reduce the blast radius of this PR if you want to.

currency/code.go
gbjk:
```suggestion
	format := strings.ContainsFunc(c, func(r rune){ return unicode.IsLetter(r) && unicode.IsUpper(r) })
```

shazbert: 🔥 I like this. 👍

currency/code.go
shazbert: Hopefully this resolves the issue; [50191fd](https://github.com/thrasher-corp/gocryptotrader/pull/1593/commits/50191fdd58d5c0ca2cb4129fde7077be475dfd88) Thanks

gbjk: Suspect you meant 1aa4da4 🙂

currency/code.go
```diff
@@ -218,20 +212,13 @@ func (b *BaseCodes) Register(c string, newRole Role) Code {
 				}
 				stored[x].Role = newRole
 			}
-			if !stored[x].CaseSensitive {
-				format = true
-			}
-			return Code{Item: stored[x], upperCase: format}
+			return Code{Item: stored[x], upperCase: format || c == lower}
```
gbjk, line 215: Is there a use-case for when format is false, and c != lower ?
I would have thought not. Maybe I'm being stupid.

gbjk: Subject to this, I think we should be changing format if it *is* true, because we use this decision (`format || c == lower`) twice.
But I'm still skeptical about when it'd be exercised.
If we put that in a:
```
if !format && c != lower {
  format = true
}
```
then we can see if we can hit it or not with TestCoverage.

currency/code.go
```diff
@@ -186,17 +186,11 @@ func (b *BaseCodes) Register(c string, newRole Role) Code {
...
 
-	var format bool
-	for x := range c {
-		if unicode.IsLetter(rune(c[x])) {
-			// Digits fool upper and lower casing. So find first letter and check case.
-			format = unicode.IsUpper(rune(c[x]))
-			break
-		}
-	}
+	format := strings.ContainsFunc(c, func(r rune) bool { return unicode.IsLetter(r) && unicode.IsUpper(r) })
```
gbjk, line 189: `format` is feeling like a bad var name now.
`isUppercase` 🤷

currency/code.go
shazbert: Yes, Seems I have done a boo boo with respect to bitfinex tests.

currency/code.go
```diff
@@ -282,24 +269,26 @@ func (c Code) String() string {
...
 	}
 	return c.Item.Lower
 }
 
-// Lower converts the code to lowercase formatting
+// Lower flags the Code to use LowerCase formatting, but does not change Symbol
+// If Code cannot be lowercased then it will return Code unchanged
 func (c Code) Lower() Code {
-	if c.Item != nil && !c.Item.CaseSensitive {
+	if c.Item == nil || c.Item.Symbol == c.Item.Lower {
```
gbjk, line 281: 🚨 I believe this might introduce a bug.
```
c := Code{Item: Item{ Symbol: "lower", Lower: "lower": Role: DanceLikeA🐧 }}
c = c.Upper() // Now upperCase is true
c = c.Lower() // Now upperCase is still true
```
I don't see any downside in just doing:
```suggestion:
if c.Item == nil {
 return c
 }
 c.upperCase = false
```
I think it's probably faster to set a bool to false than to check equality.

currency/pair.go
```diff
@@ -150,11 +140,22 @@ func MatchPairsWithNoDelimiter(currencyPair string, pairs Pairs, pairFmt PairFor
...
+
+func (p Pair) hasFormatDifference(pairFmt PairFormat) bool {
+	return p.Delimiter != pairFmt.Delimiter ||
+		(p.Base.isCaseSensitive() && p.Base.upperCase != pairFmt.Uppercase) ||
+		(p.Quote.isCaseSensitive() && p.Quote.upperCase != pairFmt.Uppercase)
+}
+
+func (c Code) isCaseSensitive() bool {
+	if c.Item == nil {
+		return false
```
gbjk, line 158: This seems left over?
I felt like the original check should be okay here, personally. We just trust the `upperCase` field, and error if they're inconsistent.

gbjk (changes requested):
Thanks - Seems we're agreed we can simplify this 🎉 
One bug and one issue in GetFormatting

currency/pair.go
shazbert: It's mostly to do with just numbers for either a base or quote, where the upperCase field is irrelevant, so for things like "ETH-420" or "eth-420" this bypasses the check required. Unless you have another way to do things or you want me to mutate the pair on creation so that it always conforms to string formatting but that seems like a weird step or maybe I am just not understanding. <-- high probability. 😆

currency/code.go
shazbert: I have purged

currency/code.go
shazbert: I purged it completely a number will default to lower anyway.

currency/pair.go
gbjk: Actually, no, I think I'm good with this.
My thought was that if we don't *know* the casing, we can't derive the Formatting from it, therefore it's an error.
However I think your solution is better.
Just to check though:
When *might* we want to derive formatting from an ambiguously cased pair like `BTC-420`
If we don't have a current use-case, I'd vote for going for the simplest solution and erroring, removing the extra func for `isCaseSensitive` and leaving a comment saying that the option is there.
I normally push stuff like that to `mothball/feature_pair_getformat_ambiguous` to retrieve later.

So, up to you.

gbjk (approved):
Approved for the stuff I've engaged on.

Note: Haven't done a thorough review of tests.

currency/pair.go
shazbert: defer to @gloriousCode 😆

gloriousCode (changes requested):
Just some little things. I have not yet gotten my head around all the changes by gk yet, so I'll caffeine up for that later

currency/pair.go
gbjk: Resolving this:
I encountered this failure throughout ExchangeWrapperStandards, so it's neccessary. Happy with your change.

currency/pair.go
```diff
@@ -150,11 +140,22 @@ func MatchPairsWithNoDelimiter(currencyPair string, pairs Pairs, pairFmt PairFor
 
 // GetFormatting returns the formatting style of a pair
 func (p Pair) GetFormatting() (PairFormat, error) {
-	if p.Base.UpperCase != p.Quote.UpperCase {
-		return PairFormat{}, fmt.Errorf("%w casing mismatch", errPairFormattingInconsistent)
+	if p.Base.isCaseSensitive() && p.Quote.isCaseSensitive() && (p.Base.upperCase != p.Quote.upperCase) {
+		return EMPTYFORMAT, fmt.Errorf("%w casing mismatch", errPairFormattingInconsistent)
+	}
+	upper := p.Base.isCaseSensitive() && p.Base.upperCase || p.Quote.isCaseSensitive() && p.Quote.upperCase
```
gbjk, line 146: I think you don't need to check isCaseSensitive() like this.
In each of the `a && b` tests `b` can only be true if `a` is.
In which case, you only need to test `b`
```suggestion
	upper := p.Base.upperCase || p.Quote.upperCase
```

currency/pairs.go
```diff
@@ -233,48 +235,58 @@ func (p Pairs) GetMatch(pair Pair) (Pair, error) {
...
+	return PairDifference{New: newPairs, Remove: removedPairs, FormatDifference: formatDiff}, nil
 }
 
 // HasFormatDifference checks and validates full formatting across a pairs list
 func (p Pairs) HasFormatDifference(pairFmt PairFormat) bool {
 	for x := range p {
-		if p[x].Delimiter != pairFmt.Delimiter ||
-			(!p[x].Base.IsEmpty() && p[x].Base.UpperCase != pairFmt.Uppercase) ||
-			(!p[x].Quote.IsEmpty() && p[x].Quote.UpperCase != pairFmt.Uppercase) {
+		if p[x].hasFormatDifference(pairFmt) {
```
gbjk, line 289: I think this is just a return 
```suggestion
		return p[x].hasFormatDifference(pairFmt)
```

gbjk, line 289: Think the entire func can become:
```suggestion
return slices.ContainsFunc(p, func(pair Pair) bool { return pair.hasFormatDifference(pairFmt) })```

File: currency/pair.go (hunk 10 above)
Comment (shazbert) [reply currency/pair.go:146]: What a guy! 🔥

File: currency/manager.go [hunk 12]
```diff
@@ -497,7 +497,7 @@ func (p *PairsManager) SetDelimitersFromConfig() error {
 		}
 		for i, p := range []Pairs{s.Enabled, s.Available} {
 			for j := range p {
-				if p[j].Delimiter == cf.Delimiter {
+				if cf.Delimiter == "" || p[j].Delimiter == cf.Delimiter {
```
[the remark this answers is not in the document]
shazbert: Yeah I wasn't sure, I purged the entire function to see what actually goes on and its called in config in func `CheckExchangeConfigValues`. All pairs should contain a delimiter and the formatting will be overidden from the wrapper function in all cases. The config format and the request format are not configurable. So the only time it was needed was in a config upgrade but it would be flushed a delimiter set anyway. I will push a change up and see what breaks. Then you can call me a silly billy for doing such things.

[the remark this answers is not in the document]
shazbert: Yeah that broke everything to do with BTSE. 😆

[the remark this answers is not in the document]
shazbert: Edge case it seems when there is no config delimiter set. It doesn't do anything and fetch as normal and set as normal.
image

gloriousCode (approved):
tACK

currency/pairs_test.go
```diff
@@ -836,3 +841,112 @@ func TestPairsEqual(t *testing.T) {
 	assert.Equal(t, "USDT-BTC", orig[0].String(), "Equal Pairs should not effect original order or format")
 	assert.False(t, orig.Equal(Pairs{NewPair(DAI, XRP), NewPair(DAI, BTC), NewPair(USD, LTC)}), "UnEqual Pairs should return false")
 }
+
+func TestFindPairDifferences(t *testing.T) {
+	pairList, err := NewPairsFromStrings([]string{defaultPairWDelimiter, "ETH-USD", "LTC-USD"})
```
thrasher-, line 846: Inconsistency between using assert vs traditional Go testing style

thrasher- (changes requested):
Just two minor nits

thrasher- (approved):
Thanks for making those changes! tACK

PR #1598, opened 2024-07-31
websocket/gateio: Add request functions for websocket multi-connection [SPOT]

shazbert:
# PR Description

This adds initial functionality for outbound websocket requests on GateIO using multi connection management. 

Dependencies:

* #1580 
* #1581 
* #1615 
* #1733 

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


gloriousCode (changes requested):
This is very nice. Only a couple of things

Should this be rebased? I feel like you haven't made 129 commits in this PR branch (it is okay not to)

exchanges/stream/websocket_connection.go
```diff
@@ -352,6 +338,32 @@ func (w *WebsocketConnection) SendMessageReturnResponses(ctx context.Context, ep
 	return resps, err
 }
 
+// waitForResponses waits for N responses from a channel
+func (w *WebsocketConnection) waitForResponses(ctx context.Context, signature any, ch <-chan []byte, expected int, isFinalMessage ...Inspector) ([][]byte, error) {
```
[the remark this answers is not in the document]
shazbert: Happy to; What do you suggest?

exchanges/stream/websocket.go
```diff
@@ -1241,3 +1257,38 @@ func signalReceived(ch chan struct{}) bool {
 		return false
 	}
 }
+
+// GetConnection returns a connection by connection signature (defined in wrapper setup) for request and response
+// handling in a multi connection context.
+func (w *Websocket) GetConnection(connSignature any) (Connection, error) {
```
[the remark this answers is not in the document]
shazbert: I disagree this will add in more complexity than it needs to. The metadata can be extracted from the connection itself after an immutable key is associated with that connection type, I have tried the comparable type but it adds more complexity with no added benefit (unless you have an actual diff please show me of it).

With regards to asset type only that's not a bad idea but I have Bybit PR which is using a string as key for a Trade connection and a specific private connection for incoming data. 

As for more options for people using fancier signatures they can define a struct with fields with immutable keys right now if they so choose.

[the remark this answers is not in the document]
shazbert: What I can do in the meantime is do a comparable check on the keys being set if not nil using reflection as an option.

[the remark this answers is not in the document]
gloriousCode: I agree that `any`thing is more complex than `any`.  I cannot see anywhere one can extract the asset from a connection. A connection itself doesn't even have a signature. The only way I can see that you can retrieve the asset is from `Websocket.GetConnection` and _hope_ that the signature is simply an asset and if you get a connection back. 

So far the requirements you have told me for a connection signature is a `string` and an `asset.Item` and struct containing those two does not sound complex. I don't see how utilising reflection is simpler

I'm still open to using `any`, but I think I'll need more convincing. What I think is needed is a way to get connection signatures. At present you can only return a connection based on if you already know the signature in advance. Then you can dynamically determine the connection you want. At least with a type you have clarity and structure for what is expected of a signature. That's where I think `any` becomes more problematic than a type, but it can still work, just with more complexity. Being able to dynamically retrieve a connection I think would be to the feature's benefit.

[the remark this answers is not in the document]
shazbert: > I agree that anything is more complex than any. I cannot see anywhere one can extract the asset from a connection. A connection itself doesn't even have a signature. The only way I can see that you can retrieve the asset is from Websocket.GetConnection and hope that the signature is simply an asset and if you get a connection back.

Why do you need that anyway? There is no way to get a list of connections, everything is set and handled in wrapper? I must be confused on what your requesting as I have said above `I have tried the comparable type but it adds more complexity with no added benefit (unless you have an actual diff please show me of it)` So please send me an example or a diff and what system will need that if any. 

> So far the requirements you have told me for a connection signature is a string and an asset.Item and struct containing those two does not sound complex. I don't see how utilising reflection is simpler. At least with a type you have clarity and structure for what is expected of a signature.

This is the only check I have added which forces comparable on the key which is wrapper defined and not used externally [here](https://github.com/thrasher-corp/gocryptotrader/pull/1598/commits/511c78b43522acd6fe8b0b077896070439f2fc0f) This is only for setting the comparable key. This does not check every time you call get connection. 

> I'm still open to using any and maybe I'm misunderstanding everything you're saying. What I think is needed is a way to get connection signatures (even if its returning []any). At present you can only return a connection based on if you already know the signature in advance. The feature for returning signatures you can then dynamically determine the connection you want. That's where I think any becomes more problematic than a type, but it can still work, just with more complexity for the caller. Being able to dynamically retrieve connections I think would be to the feature's benefit.

There is yet to be any external request path introduced in this PR so what your requesting can't even be tested properly, everything is handle via wrapper functions, the decision to use the connection is consolidated to the wrapper and not external. Can you open PR based off this and we can go down that direction when there is a use case? I don't see an issue its just I am reluctant to add things in the off chance its going to be used. Or maybe I am still not understanding the request.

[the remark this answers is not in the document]
gloriousCode: I think this is a result of exporting functions like `AuthenticateSpot(ctx context.Context, conn stream.Connection) error` which is designed and developed specifically for the handling inside this package and makes no mention of it. When I'm seeing exported functions that ask for a `conn` with the only way to get a connection is to call _this exported function_ `GetConnection`, where the only way to retrieve a connection is to already know the hard-coded connection signature that can be of any type. I hope from that its a bit clearer about why I've been saying what I've been saying

I think the best way of closing this is unexporting those two functions `AuthenticateSpot` and `WebsocketLogin`. You are passing `AuthenticateSpot` as a function already and it closes the funniness a bit and ensures that your stream package functions are doing the handling of it all.

Otherwise, sounds like we need to talk IRL to get a better understanding of things 🌈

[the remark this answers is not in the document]
shazbert: [unexported](https://github.com/thrasher-corp/gocryptotrader/pull/1598/commits/d3343a7dbde3a935286a248b1bf19c8ed6524b99)

exchanges/gateio/gateio_websocket.go
```diff
@@ -109,11 +115,17 @@ func (g *Gateio) generateWsSignature(secret, event, channel string, t int64) (st
...
 	var push WsResponse
-	err := json.Unmarshal(respRaw, &push)
-	if err != nil {
+	if err := json.Unmarshal(respRaw, &push); err != nil {
 		return err
 	}
 
+	if push.RequestID != "" {
+		if !g.Websocket.Match.IncomingWithData(push.RequestID, respRaw) {
+			return fmt.Errorf("gateio_websocket.go error - unable to match requestID %v", push.RequestID)
```
gloriousCode, line 124: `stream.ErrNoMessageListener`

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+
+	var result WebsocketLoginResponse
+	return &result, json.Unmarshal(inbound.Data, &result)
+}
+
+// WebsocketOrderPlaceSpot places an order via the websocket connection. You can
+// send multiple orders in a single request. But only for one asset route.
+// So this can only batch spot orders or futures orders, not both.
+func (g *Gateio) WebsocketOrderPlaceSpot(ctx context.Context, batch []WebsocketOrder) ([]WebsocketOrderResponse, error) {
```
gloriousCode, line 87: This doesn't seem like its just for spot

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,284 @@
...
+
+	if len(batch) == 1 {
+		var singleResponse WebsocketOrderResponse
+		err := g.SendWebsocketRequest(ctx, "spot.order_place", asset.Spot, batch[0], &singleResponse, 2)
+		return []WebsocketOrderResponse{singleResponse}, err
+	}
+
+	var resp []WebsocketOrderResponse
+	err := g.SendWebsocketRequest(ctx, "spot.order_place", asset.Spot, batch, &resp, 2)
+	return resp, err
```
gloriousCode, line 120:
```suggestion
	var resp []WebsocketOrderResponse
	return resp, g.SendWebsocketRequest(ctx, "spot.order_place", asset.Spot, batch, &resp, 2)
```

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,284 @@
...
+	}
+	if pair.IsEmpty() {
+		return nil, currency.ErrCurrencyPairEmpty
+	}
+
+	params := &WebsocketOrderRequest{OrderID: orderID, Pair: pair.String(), Account: account}
+
+	var resp WebsocketOrderResponse
+	err := g.SendWebsocketRequest(ctx, "spot.order_cancel", asset.Spot, params, &resp, 1)
+	return &resp, err
```
gloriousCode, line 136:
```suggestion
	return &resp, g.SendWebsocketRequest(ctx, "spot.order_cancel", asset.Spot, params, &resp, 1)
```
you get the idea

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,284 @@
...
+		},
+	}
+
+	responses, err := conn.SendMessageReturnResponses(ctx, request.Unset, req.Payload.RequestID, req, expectedResponses, InspectPayloadForAck)
+	if err != nil {
+		return err
+	}
+
+	if len(responses) == 0 {
+		return errors.New("no responses received")
```
gloriousCode, line 249: `common.ErrNoResponse`

exchanges/stream/websocket.go
```diff
@@ -1241,3 +1263,38 @@ func signalReceived(ch chan struct{}) bool {
...
+	w.m.Lock()
+	defer w.m.Unlock()
+
+	if !w.IsConnected() {
+		return nil, ErrNotConnected
+	}
+
+	if !w.useMultiConnectionManagement {
+		return nil, fmt.Errorf("%s: multi connection management not enabled %w please use exported Conn and AuthConn fields", w.exchangeName, errCannotObtainOutboundConnection)
+	}
```
gloriousCode, line 1287: I note the need for the mutex, but put this above IsConnected. None of the other checks matter if the feature isn't enabled

exchanges/stream/websocket.go
```diff
@@ -1241,3 +1263,38 @@ func signalReceived(ch chan struct{}) bool {
...
+
+	// Opted to range and not have a map, as connection level wrappers will be limited.
```
gloriousCode, line 1289: I think this comment detracts here. Its either something for connectionManager property or can be rm'd

exchanges/stream/websocket_test.go
```diff
@@ -223,13 +223,17 @@ func TestConnectionMessageErrors(t *testing.T) {
 	assert.ErrorIs(t, err, errNoPendingConnections, "Connect should error correctly")
 
 	ws.useMultiConnectionManagement = true
+	ws.SetCanUseAuthenticatedEndpoints(true)
+	ws.verbose = true // NOTE: Intentional
```
gloriousCode, line 227: Why? Gotta say why in the comment. If its just trying to get coverage of a println I say remove. Its getting less than 1% coverage of the package for stuff that doesn't matter

exchanges/stream/stream_types.go
```diff
@@ -37,6 +37,9 @@ type Connection interface {
 	Shutdown() error
 }
 
+// Inspector is a hook that allows for custom message inspection
+type Inspector func([]byte) bool
```
gloriousCode, line 41:
```suggestion
// Inspector is used to verify messages via SendMessageReturnResponse
// Only one can used
// It inspects the []bytes websocket message and returns true if it is the appropriate message to action
type Inspector func([]byte) bool
```

exchanges/stream/websocket_connection.go
```diff
@@ -320,36 +320,50 @@ func (w *WebsocketConnection) SendMessageReturnResponses(ctx context.Context, ep
...
+	if w.Reporter != nil {
+		w.Reporter.Latency(w.ExchangeName, outbound, time.Since(start))
+	}
+
+	return resps, err
+}
+
+// waitForResponses waits for N responses from a channel
+func (w *WebsocketConnection) waitForResponses(ctx context.Context, signature any, ch <-chan []byte, expected int, messageInspector ...Inspector) ([][]byte, error) {
 	timeout := time.NewTimer(w.ResponseMaxLimit * time.Duration(expected))
```
gloriousCode, line 337:
```go
if len(messageInspector) > 1 {
    return nil, errOnlyOneMessageInspector
}
```

gloriousCode (changes requested):
I am pleased that connection hole was semi-sorted 🌈 
Just a few small things

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,278 @@
...
+	}{
+		Result: result,
+	}
+
+	return json.Unmarshal(inbound.Data, &to)
+}
+
+// InspectPayloadForAck checks the payload for an ack, it returns true if the
+// payload does not contain an ack. This will force the cancellation of further
+// waiting for responses.
```
gbjk, line 275: Think this should not be a public func.
Also suggested we make it a type and instance func

exchanges/stream/stream_types.go
```diff
@@ -76,6 +81,15 @@ type ConnectionSetup struct {
 	// This is useful for when an exchange connection requires a unique or
 	// structured message ID for each message sent.
 	BespokeGenerateMessageID func(highPrecision bool) int64
+	// Authenticate is a function that will be called to authenticate the
+	// connection to the exchange's websocket server. This function should
+	// handle the authentication process and return an error if the
+	// authentication fails.
+	Authenticate func(ctx context.Context, conn Connection) error
```
gbjk, line 88:
```suggestion
	// Authenticate will be called to authenticate the connection
	Authenticate func(ctx context.Context, conn Connection) error
```

exchanges/stream/websocket.go
```diff
@@ -85,12 +92,12 @@ func NewWebsocket() *Websocket {
...
-		subscriptions:     subscription.NewStore(),
-		features:          &protocol.Features{},
-		Orderbook:         buffer.Orderbook{},
-		connections:       make(map[Connection]*ConnectionWrapper),
+		ReadMessageErrors:   make(chan error, 1),
+		Match:               NewMatch(),
+		subscriptions:       subscription.NewStore(),
+		features:            &protocol.Features{},
+		Orderbook:           buffer.Orderbook{},
+		connectionToWrapper: make(map[Connection]*ConnectionWrapper),
```
gbjk, line 100: I think this is regressing; Didn't we previously have it over-named and went just down to `connections` ?

exchanges/stream/websocket.go
```diff
@@ -1241,3 +1263,37 @@ func signalReceived(ch chan struct{}) bool {
...
+	if !w.useMultiConnectionManagement {
+		return nil, fmt.Errorf("%s: multi connection management not enabled %w please use exported Conn and AuthConn fields", w.exchangeName, errCannotObtainOutboundConnection)
+	}
+
+	if !w.IsConnected() {
+		return nil, ErrNotConnected
+	}
+
+	for _, wrapper := range w.connectionManager {
+		if wrapper.Setup.WrapperDefinedConnectionSignature == connSignature {
```
gbjk, line 1290: Can we have a type implement a "IsAppropriate" function instead?
It'd give us more flexibility, avoid reflection, etc.
So all your examples for asset type would use AssetFilter(asset.Spot), for example, and that'd be a type which implements `IsAppropriate`

gbjk (changes requested):
Great work 🎉 
Initial feedback, suggestions, nitpicks.
No showstoppers, I don't think.

exchanges/gateio/gateio_test.go
```diff
@@ -3011,11 +3011,11 @@ func TestGetSettlementFromCurrency(t *testing.T) {
 	for _, assetType := range []asset.Item{asset.Futures, asset.DeliveryFutures, asset.Options} {
 		availPairs, err := g.GetAvailablePairs(assetType)
 		require.NoErrorf(t, err, "GetAvailablePairs for asset %s must not error", assetType)
-		for x := range availPairs {
-			t.Run(strconv.Itoa(x), func(t *testing.T) {
+		for x, pair := range availPairs {
+			t.Run(strconv.Itoa(x)+":"+assetType.String(), func(t *testing.T) {
```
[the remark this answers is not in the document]
shazbert: time to make a linter for it 😆

exchanges/gateio/gateio_websocket.go
```diff
@@ -109,14 +115,20 @@ func (g *Gateio) generateWsSignature(secret, event, channel string, t int64) (st
...
 	var push WsResponse
-	err := json.Unmarshal(respRaw, &push)
-	if err != nil {
+	if err := json.Unmarshal(respRaw, &push); err != nil {
 		return err
 	}
 
+	if push.RequestID != "" {
+		if !g.Websocket.Match.IncomingWithData(push.RequestID, respRaw) {
+			return fmt.Errorf("%w for requestID %v", stream.ErrNoMessageListener, push.RequestID)
```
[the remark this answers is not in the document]
shazbert: I have a suggestion we could standardise this at the stream package level and return an error, just wondering if you want this option instead that way if the requirement is there we just return the error and it will give back all the details without having to worry about it wrapper side, let me know what you think:

```go 
func (m *Match) IncomingWithDataMust(signature any, data []byte) error {
	if m.IncomingWithData(signature, data) {
		return nil
	}
	return fmt.Errorf("'%v' %w with data %v", signature, errSignatureNotMatched, string(data))
}
```

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,278 @@
...
+		return nil, errChannelEmpty
+	}
+
+	creds, err := g.GetCredentials(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	tn := time.Now().Unix()
+	msg := "api\n" + channel + "\n" + "\n" + strconv.FormatInt(tn, 10)
```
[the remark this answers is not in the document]
shazbert: Yeah I see what you are saying; we tend to concat everything because it's faster generally. So I will pass on this request if that is okay?

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		return nil, fmt.Errorf("%s: %s", wsErr.Errors.Label, wsErr.Errors.Message)
+	}
+
+	var result WebsocketLoginResponse
+	return &result, json.Unmarshal(inbound.Data, &result)
+}
+
+// WebsocketOrderPlaceSpot places an order via the websocket connection. You can
+// send multiple orders in a single request. But only for one asset route.
```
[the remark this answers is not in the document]
shazbert: Yeah 😆 Its more of me trying to map to the exchanges convention [example](https://www.gate.io/docs/developers/apiv4/ws/en/#order-place) and trying to tie it in then on the interface side have it more of a better convention. So in the event of trying to find the relation from their API docs Where is `OrderPlace` for example then you will find a few hits across some files in this package. So prefixed with websocket and tacked Spot on the end, this will also be different for futures trading functions in another PR. That was my reasoning not necessarily correct reasoning I might add. 😆 If you still want me to change I can.

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,278 @@
...
+func (g *Gateio) WebsocketOrderPlaceSpot(ctx context.Context, batch []WebsocketOrder) ([]WebsocketOrderResponse, error) {
+	if len(batch) == 0 {
+		return nil, errBatchSliceEmpty
+	}
+
+	for i := range batch {
+		if batch[i].Text == "" {
+			// For some reason the API requires a text field, or it will be
+			// rejected in the second response. This is a workaround.
```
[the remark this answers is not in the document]
shazbert: So you have an initial response which is the acknowledgement of receiving the order, then the second response is the confirmation of order details.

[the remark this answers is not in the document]
shazbert: I will use just this `// API requires Text field, or it will be rejected`

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,278 @@
...
+		}
+		if batch[i].Type == "limit" && batch[i].Price == "" {
+			return nil, errInvalidPrice
+		}
+	}
+
+	if len(batch) == 1 {
+		var singleResponse WebsocketOrderResponse
+		return []WebsocketOrderResponse{singleResponse}, g.SendWebsocketRequest(ctx, "spot.order_place", asset.Spot, batch[0], &singleResponse, 2)
+	}
```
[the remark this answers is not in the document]
shazbert: I think it could be nuanced in a way their matching engine operates and executes batches. If you send a single order it might have less overhead (maybe not 🤷), so if you include a single order as an outbound slice there could be differences (maybe not 🤷). I haven't added batch support in a market maker yet so I have no idea. I am still thinking we should keep the current mapping to their functionality for now, even though what you suggest is much much cleaner

exchanges/stream/websocket.go
shazbert: I don't remember to be honest, I think I had Connection Wrapper as Connection Candidate. 🤷 I will revert

exchanges/stream/websocket.go
shazbert: Can you send me a diff?

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,278 @@
...
+
+var (
+	errBatchSliceEmpty  = errors.New("batch cannot be empty")
+	errNoOrdersToCancel = errors.New("no orders to cancel")
+	errEdgeCaseIssue    = errors.New("edge case issue")
+	errChannelEmpty     = errors.New("channel cannot be empty")
+)
+
+// websocketLogin authenticates the websocket connection
+func (g *Gateio) websocketLogin(ctx context.Context, conn stream.Connection, channel string) (*WebsocketLoginResponse, error) {
```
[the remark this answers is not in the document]
gbjk: > I still need declare an anon func on the authenticate field so it will be something like this and in future the futures endpoints in the PR above this one for its connection:

Okay, ignore that, then. If you'd need an anonymous function and you're *not* injecting state into it, then what you've done is sanest.

> Also I don't know if we ever need the return value (APIKEY or Connection ID) but it gives me a clue that there is something else returned so I would rather for the moment keep it for potential future usage. 🤷

Nope, I still don't agree. Anything that's coming out of that API is something that websocketLogin should be handling.
You have no consumer for this (even conceivably, I think), so there's nobody else to process that response but the function itself. Any state it gets, it needs to find a place to store (WS or exch I guess).
But if you don't have a consumer for the information, don't return it 🙂 
TL;DR: Don't be a hoarder 😂 

> With regard to the dropping prefix that is understandable maybe we should have that as its own PR across exchanges for a standardisation pass? I have addressed the inconsistency though and will push that up soon.

Agreed - Out of scope and would be wrong to do it here.

exchanges/gateio/gateio_websocket.go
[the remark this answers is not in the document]
gbjk: Agreed.
Note: I think that `ErrNoMessageListener` is widespread and should replace `errSignatureNotMatched` in your example

exchanges/gateio/gateio_websocket_request_spot.go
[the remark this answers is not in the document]
gbjk: Yeah, that's much faster than I expected, as well.
```
BenchmarkConcat1-8      35654198                33.84 ns/op            7 B/op          0 allocs/op
BenchmarkConcat2-8      12242659                97.91 ns/op           40 B/op          3 allocs/op
```

exchanges/gateio/gateio_websocket_request_spot.go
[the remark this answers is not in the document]
gbjk: Yeah. Okay. I'm still going to vote for standardising more towards our naming, even accepting that this is the "client" side of the package. I'm still hoping to break that line a little more at some point so our client interface package and our GCT adapter package can be separate.

Explicitly: I'd still vote against introducing `Place` here, and use `Submit`.
Reasoning: GCT users who wanted to find this function are GCT centric, not GateIO centric. They're not looking at GateIO specs trying to work out what they can do with GCT, they're looking at GCT docs.

Not a change request. Still just a suggestion.

exchanges/stream/stream_types.go
```diff
@@ -37,10 +39,11 @@ type Connection interface {
 	Shutdown() error
 }
 
-// Inspector is used to verify messages via SendMessageReturnResponse
-// Only one can used
+// Inspector is used to verify messages via SendMessageReturnResponsesWithInspection
 // It inspects the []bytes websocket message and returns true if it is the appropriate message to action
-type Inspector func([]byte) bool
+type Inspector interface {
+	Inspect([]byte) bool
```
gbjk, line 45: Soz; Nitpick:
`Inspect` doesn't feel appropriate for the boolean response type.
I'd say an Inspector would decide if things are Interesting, or Valid, or Appropriate.
So I guess I'm asking for an adjective.
`Appropriate` feels right

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -261,18 +207,12 @@ func (g *Gateio) SendWebsocketRequest(ctx context.Context, channel string, connS
...
-	return json.Unmarshal(inbound.Data, &to)
+	return json.Unmarshal(inbound.Data, &funnelResult{Result: result})
 }
 
-// InspectPayloadForAck checks the payload for an ack, it returns true if the
+// Inspect checks the payload for an ack, it returns true if the
 // payload does not contain an ack. This will force the cancellation of further
 // waiting for responses.
-func InspectPayloadForAck(data []byte) bool {
+func (g *Gateio) Inspect(data []byte) bool {
```
gbjk, line 216: Whoa 🙁 
Kinda cute to abuse GateIO as the Inspector type, but no matter how you name this, you can't pollute the GateIO namespace with a func named anything like this.
I was expecting a private type which would do `Appropriate`
```
type wsRespAckInspector struct {}
func (_ *wsRespAckInspector) Appropriate(m []byte) bool {
...
}
```

exchanges/stream/websocket.go
gbjk: Sure. Will do.

gbjk: Okay. I'm wrong.

Direction was this: https://github.com/gbjk/gocryptotrader/commit/33bb31e4d

But we're just allowing a function to compare X to Y, in which case we could use exchange defined struct types ( or even stream package defined, like UrlAssetFilter )

We don't need to make it a method call because I can't see a case where it wouldn't just be filtering on equality, yet.

So, my only change request is to sanitise this name to `MessageFilter` or similar. `WrapperDefined` prefix is wrong, not least because Wrapper isn't actually a thing, other than a file suffix _convention_. But also because we don't case what defined or provided the filter. We just care that we have one.

exchanges/gateio/gateio_websocket_request_spot.go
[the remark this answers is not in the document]
gbjk: Okay. There's too much guessing in there for my liking, but I'll leave it be.

exchanges/gateio/gateio_websocket_request_spot.go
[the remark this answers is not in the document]
gbjk: Note: Resolve if you don't want to address :)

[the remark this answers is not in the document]
shazbert: I won't resolve just yet for others to check this out. @thrasher- @gloriousCode

exchanges/stream/stream_types.go
shazbert: For the use case I used `IsFinal` because it returns true if further processing is not required and updated the commentary [here](https://github.com/thrasher-corp/gocryptotrader/pull/1598/commits/6a2fbdec024a0e192a9255a6b23d9fd9eaf82338)

exchanges/stream/websocket.go
shazbert: [fixed](https://github.com/thrasher-corp/gocryptotrader/pull/1598/commits/cb965218c36e27c0547c98ccd2a2c9cb7e6ecb16)

exchanges/stream/stream_types.go
gbjk: That's good.

exchanges/gateio/gateio_websocket_request_spot.go
gbjk: Looks good.

gbjk (approved):
I'm happy.

2 conversations outstanding where I was asking for changes, but they're trivial.
Address or Ignore :)

shazbert:
This now requires dependency #1733 for spot trading rate limits

exchanges/gateio/gateio_websocket_request_spot.go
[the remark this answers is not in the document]
gloriousCode: `WebsocketSpotOrderPlace` I think most cleanly matches our standard of naming API functions after their entry in the API docs. If this were a wrapper function I'd want `Submit`

I also think I complained about the function name above without realising that it's the API doc's fault

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -261,18 +207,13 @@ func (g *Gateio) SendWebsocketRequest(ctx context.Context, channel string, connS
...
 	}
 
-	to := struct {
-		Result any `json:"result"`
-	}{
-		Result: result,
-	}
-
-	return json.Unmarshal(inbound.Data, &to)
+	return json.Unmarshal(inbound.Data, &funnelResult{Result: result})
```
gloriousCode, line 210: Something for another time is that I think jsonparser may be of use in the 
```
if err := json.Unmarshal(endResponse, &inbound); err != nil {
```
above this. All you do is check a status and then if failed, the error message. Might be able to save on a double unmarshal. But Sonic the json hedgehog might be faster I can't benchmark that

gloriousCode (changes requested):
I'm happy. Only one change which is just a rename. Once ratelimits is merged then I can do a final look

exchanges/stream/stream_match.go
```diff
@@ -47,6 +51,15 @@ func (m *Match) IncomingWithData(signature any, data []byte) bool {
 	return true
 }
 
+// EnsureMatchWithData validates that incoming data matches a request's signature.
+// If a match is found, the data is processed; otherwise, it returns an error.
+func (m *Match) EnsureMatchWithData(signature any, data []byte) error {
```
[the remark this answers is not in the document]
shazbert: 🤷 I obviously don't know how to word so it's not surprising.

exchanges/gateio/gateio_websocket_request_spot.go
shazbert: Yeah good point; I will do some tests on it when I get some time, from what I have seen previously its still lighter to unmarshal everything using sonic rather than even json parser for one item. But mem bad 🤯. Also it might add in overhead from a non error path situation if we look at it from that perspective. We shall see.

shazbert:
removed label because test fix was included in [recent commit](https://github.com/thrasher-corp/gocryptotrader/pull/1598/commits/9c56cbb413e7bca41aa42340d896cc10cb686561)

gloriousCode (approved):
tACK! Thanks Shazburtino

exchanges/gateio/gateio_websocket_request_spot.go
```diff
@@ -0,0 +1,219 @@
...
+	params := &WebsocketOrderRequest{OrderID: orderID, Pair: pair.String(), Account: account}
+
+	var resp WebsocketOrderResponse
+	return &resp, g.SendWebsocketRequest(ctx, spotCancelSingleOrderEPL, "spot.order_cancel", asset.Spot, params, &resp, 1)
+}
+
+// WebsocketOrderCancelAllByIDsSpot cancels multiple orders via the websocket
+func (g *Gateio) WebsocketOrderCancelAllByIDsSpot(ctx context.Context, o []WebsocketOrderBatchRequest) ([]WebsocketCancellAllResponse, error) {
+	if len(o) == 0 {
+		return nil, errNoOrdersToCancel
```
thrasher-, line 77: MTC

engine/engine.go
```diff
@@ -791,17 +791,11 @@ func (bot *Engine) LoadExchange(name string) error {
...
-			exch.GetName(),
-		)
 		err = exch.DisableRateLimiter()
 		if err != nil {
-			gctlog.Errorf(gctlog.ExchangeSys,
-				"Loaded exchange %s rate limiting cannot be turned off: %s.\n",
-				exch.GetName(),
-				err,
-			)
+			gctlog.Errorf(gctlog.ExchangeSys, "error disabling rate limiter for %s: %v", exch.GetName(), err)
```
thrasher-, line 796:
```suggestion
			gctlog.Errorf(gctlog.ExchangeSys, "%s error disabling rate limiter: %v", exch.GetName(), err)
```

To be consistent with all of our other log messages where the exchange or thing is mentioned first, followed by the message

thrasher- (changes requested):
Only two minor cosmetic things

exchanges/gateio/gateio_websocket_request_spot.go
shazbert: At first I was like wth is MTC 😆 - [fixed](https://github.com/thrasher-corp/gocryptotrader/pull/1598/commits/8f9e9535b27452ef7328066f9560df02036963fa)

exchanges/stream/stream_match_test.go
```diff
@@ -51,3 +51,18 @@ func TestRemoveSignature(t *testing.T) {
 		t.Fatal("Should be able to read from a closed channel")
 	}
 }
+
+func TestRequireMatchWithData(t *testing.T) {
+	t.Parallel()
+	match := NewMatch()
+	err := match.RequireMatchWithData("hello", []byte("world"))
+	require.ErrorIs(t, err, ErrSignatureNotMatched, "Should error on unmatched signature")
```
thrasher-, line 59:
```suggestion
	require.ErrorIs(t, err, ErrSignatureNotMatched, "Must error on unmatched signature")
```

exchanges/stream/stream_match_test.go
```diff
@@ -51,3 +51,18 @@ func TestRemoveSignature(t *testing.T) {
...
+	assert.Contains(t, err.Error(), "world", "Should contain the data in the error message")
+	assert.Contains(t, err.Error(), "hello", "Should contain the signature in the error message")
+
+	ch, err := match.Set("hello", 1)
+	require.NoError(t, err, "Set must not error")
+	err = match.RequireMatchWithData("hello", []byte("world"))
+	require.NoError(t, err, "Should not error on matched signature")
```
thrasher-, line 66:
```suggestion
	require.NoError(t, err, "Must not error on matched signature")
```

thrasher- (changes requested):
Thanks for making those changes. Minor should/must cosmetic things remaining

exchanges/gateio/gateio_websocket_request_spot.go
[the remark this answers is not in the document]
thrasher-: Standardisation of client side methods is great and the desired long term direction, though for now that's to be expected in the wrappers. What I want to see first is integration of all exchange API endpoints and then naming standards can be enforced across all exchanges after we've achieved that since not all exchanges support the same features/endpoints etc. We currently have many PlaceOrder methods across several exchanges which would all need updating as well.

For now, I'm fine with `Websocket(Place/Submit)SpotOrders` or WebsocketSpot(Place/Submit)Orders. I think OrderPlace is a bit funny, despite it being correlated to the API docs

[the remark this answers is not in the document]
shazbert: [Names have been changed](https://github.com/thrasher-corp/gocryptotrader/pull/1598/commits/544fbccdedbe9b76a80cde4383e3edfc1e184e49)

thrasher- (approved):
Changes ACK 👍

PR #1599, opened 2024-08-01
GateIO: Add Subscription configuration

gbjk:
### GateIO

* Add subscription configuration
* Upgrade test config

## New Library Dependencies

- [sprig](github.com/Masterminds/sprig/v3)   ( Also included in 4 other template PRs )

## Type of change


shazbert (changes requested):
Just needs rebase and we are good to go. 👍

gbjk:
Rebased on master. No issues during rebase.

shazbert (approved):
tACK.

exchanges/gateio/gateio_websocket.go
```diff
@@ -11,8 +11,10 @@ import (
 	"net/http"
 	"strconv"
 	"strings"
+	"text/template"
 	"time"
 
+	"github.com/Masterminds/sprig/v3"
```
gloriousCode, line 17: I'm just commenting here because this dependency isn't mentioned in this PR or its "stacked dependency" reference.
This is used here to access the function "hasPrefix" when generating subscription templates. 

Only thing I'll say is that we're adding a lot of dependencies when we could just add the following to your funcmap
```go
"hasPrefix":           func(substr string, str string) bool { return strings.HasPrefix(str, substr) },` 
```

Maybe I'll see more usages for it in other PRs. I'll see what you say

gloriousCode (commented):
It all works, good stuff

exchanges/gateio/gateio_websocket.go
gbjk: > Maybe I'll see more usages for it in other PRs. I'll see what you say

Exactly that. In isolation some of the sprig usages seem "_meh_", but taken as a whole they add up to "yeah, that's worth it". 
I think it's used in 4 of them, and it's `mustToJson` was the main driver.

gbjk: PR updated to mention it.

gbjk: @gloriousCode Just spotted this PR is tickless. Like a domestic cat.
Unresolving conversation.
If you're happy with this then approve 😄

gloriousCode (approved):
tACK! I'm happy with the changes and the inclusion of sprig 🎉 There is some rebasing to do, so I'll probably tACK again after

gbjk:
Clean rebase, zero issues.
Conflicts were just on go.sum and adding `request.Unset` into requests.

thrasher- (approved):
tACK auth and unauth subs 🚀

PR #1420, opened 2023-12-11
exchanges: Okx Update

samuael:
Selam, ወንድሞች.

# PR Description

This pull request incorporates modifications for the Okx exchange, encompassing changes to endpoint methods, rate limits, WebSocket subscriptions and handlers, wrapper functions, and conversion methods. Rigorous testing has been conducted on all facets of the functions to ensure their reliability. The code adheres to the GCT style standards, consistent with practices adopted by other exchanges. To enhance code quality, **_golangci-lint_** was utilized for linter testing, effectively rectifying a few errors. Furthermore, this implementation includes endpoint methods that the GCT wrapper did not initially support, ensuring compatibility for future use

The PR updated implementations for REST for trading and other endpoint methods, Web-socket streaming, and Trading through the web socket stream.

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as the item is complete.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and also, consider improving test coverage whilst working on a certain feature or package.

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test X

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions/AppVeyor with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1600, opened 2024-08-02
okx: decouple outbound trading functionality from multiplexer

shazbert:
# PR Description

* The multiplexer was a good intermediary but added in overhead and code complexity. This PR shifts the burden of websocket message match management to stream.Match using the function SendMessageReturnResponse.
* It also shifts the burden of JSON unmarshalling to the request calling routine and the message []byte is shifted via pointer, ~~which has a higher probability of improved websocket message handling~~.
* The function design attempts to mirror REST request flow. ~~This is probably going to be a multi-stage decoupling for multi-connection management.~~ All done thanks Sam.
* For matching signature, have opted to use ~~Unix timestamp~~ common.Counter. There was slight overhead from common.GenerateRandomString.
* Removes WsRequestSemaphore due to upstream changes incorporating request rate limiter.
*  Adds TODO to export request rate limiting functionality so that we can couple rate limits between REST and websocket connections.

* Waiting for #1420 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


samuael:
Thank you for your changes on the Okx exchange and for inviting me to review them. This PR is very optimized and very clean. The only proposal I have on this PR is to eliminate the Multiplexer by handling the incoming data by checking possible login error codes from the incoming data and handling the data based on that filter.

For instance, these are the possible incoming login error codes that can be used to filter login errors.
1. 60022, 60023, 60024, 60026, 63999, 60032, 60011, 60009, 60005, 60021, and 60031; as specified in here https://www.okx.com/docs-v5/en/#error-code-websocket-public

samuael:
[remove_multiplexer.patch](https://github.com/user-attachments/files/16474225/remove_multiplexer.patch)

This is the change I am referring to. You might need to adjust a few things, but I have attempted to implement my intended modifications to eliminate the multiplexer.

shazbert:
@samuael I added in your patch and did some adjustments to a type and the matching signature. Thanks heaps for looking into this.

samuael:
> @samuael I added in your patch and did some adjustments to a type and the matching signature. Thanks heaps for looking into this.

The pleasure is mine and thanks, Boss @shazbert. I hope to see this merged soon so that I can proceed with changes to the Okx Update PR.

exchanges/okx/okx_websocket_requests.go
```diff
@@ -0,0 +1,230 @@
...
+	id := strconv.FormatInt(time.Now().UnixNano(), 10) // TODO: use atomic counter
+
+	// Opted to not check scode here as some orders may be successful and some
+	// may not. So return everything to the caller to handle.
+	var orderResp []OrderData
+	return orderResp, ok.SendWebsocketRequest(id, "batch-amend-orders", args, &orderResp)
+}
+
+// SendWebsocketRequest sends a websocket request to the server
+func (ok *Okx) SendWebsocketRequest(id, operation string, payload, response any) error {
```
gloriousCode, line 178:
```suggestion
func (ok *Okx) SendAuthenticatedWebsocketRequest(id, operation string, payload, response any) error {
```

exchanges/okx/okx_websocket.go
```diff
@@ -554,8 +515,14 @@ func (ok *Okx) WsHandleData(respRaw []byte) error {
 		}
 		return fmt.Errorf("%w unmarshalling %v", err, respRaw)
 	}
-	if (resp.Event != "" && (resp.Event == "login" || resp.Event == "error")) || resp.Operation != "" {
-		ok.WsResponseMultiplexer.Message <- &resp
+	if resp.Event != "" &&
+		(resp.Event == "login" ||
+			(resp.Event == "error" &&
+				slices.Contains([]string{"60022", "60023", "60024", "60026", "63999", "60032", "60011", "60009", "60005", "60021", "60031"}, resp.Code))) {
```
gloriousCode, line 521: 🚨 🚑 🚨 
What kind of insane slice ordering is this? Its not like you were going down the list in okx docs and adding them 😆 
Would 6007 fit?

exchanges/okx/okx_websocket.go
```diff
@@ -554,8 +515,14 @@ func (ok *Okx) WsHandleData(respRaw []byte) error {
...
+		// find error codes and corresponding reasons: https://www.okx.com/docs-v5/en/#error-code-websocket-public
+		if !ok.Websocket.Match.IncomingWithData("login-response", respRaw) {
```
gloriousCode, line 523: Unrelated, but this is the kind of thing I was talking about in #1589 , this kind of generic message of `login-response` is going to get whacky if you have multiple authenticated okx websocket connections

gloriousCode (changes requested):
>higher probability of improved websocket message handling

👀 _well does it?_ Otherwise remove such a claim from your description 🚨 

Cool stuff otherwise

exchanges/okx/okx_websocket.go
shazbert: Yeah good point, So there is one option that might mitigate this, when there is concurrent connections being made (which is not currently happening at the moment). We can move the stream.Match type to be held by the connection itself, which I was thinking about doing anyway because you can localise your outbound subscription response matching to the connection. So you can have this generic message on each connection and it shouldn't impede anything else. But I would need to untie a bunch of websocket Match field calls in the wrappers, so I think it will need to be in a separate PR.

shazbert:
> > higher probability of improved websocket message handling
> 
> 👀 _well does it?_ Otherwise remove such a claim from your description 🚨
> 
> Cool stuff otherwise

It was speculation, I can remove that wording? So this PR removes what was happening before including the inline JSON calls in the websocket reader/handler, I haven't tested it in a benchmark. That was the reason I suggested it had an higher probability of improved websocket message handling. I can see now why those choice of words are not entirely correct.

exchanges/okx/okx_websocket.go
gloriousCode: I definitely wasn't asking to hold up this PR for that. It was just a good example of potential impacts of solidifying moving things off the individual connection

shazbert: #1613 Potentially addresses this.

shazbert:
Opted to label this as medium priority to highlight that it's complementary to #1420 and it's updates. This does not necessarily need to be merged first but it reduces a lot of code complexity.

exchanges/okx/okx_websocket_requests.go
```diff
@@ -0,0 +1,230 @@
...
+// access request level rate limits and update the rate limiter accordingly.
+func (ok *Okx) SendAuthenticatedWebsocketRequest(ctx context.Context, id, operation string, payload, response any) error {
+	if id == "" || operation == "" || payload == nil || response == nil {
+		return errInvalidWebsocketRequest
+	}
+
+	outbound := &struct {
+		ID        string      `json:"id"`
+		Operation string      `json:"op"`
+		Arguments interface{} `json:"args"`
```
gloriousCode, line 186:
```suggestion
		Arguments any `json:"args"`
```

exchanges/okx/okx_websocket.go
```diff
@@ -282,56 +275,17 @@ func (ok *Okx) WsAuth(ctx context.Context, dialer *websocket.Dialer) error {
...
-		case <-timer.C:
-			timer.Stop()
-			return fmt.Errorf("%s websocket connection: timeout waiting for response with an operation: %v",
-				ok.Name,
-				req.Operation)
-		}
-	}
+	var nothing interface{}
+	err = ok.SendAuthenticatedWebsocketRequest(ctx, "login-response", operationLogin, args, &nothing)
+	return err
```
gloriousCode, line 288:
```suggestion
	return ok.SendAuthenticatedWebsocketRequest(ctx, "login-response", operationLogin, args, &nothing)
```

exchanges/okx/okx_websocket_requests.go
```diff
@@ -0,0 +1,230 @@
...
+	}{
+		ID:        id,
+		Operation: operation,
+		Arguments: payload,
+	}
+
+	if ok.Verbose {
+		if payload, err := json.Marshal(outbound); err != nil {
+			log.Debugf(log.ExchangeSys, "%s sending outbound request via websocket: %v", ok.Name, string(payload))
+		}
```
gloriousCode, line 196: Only log that you're sending a payload when you fail to marshal the struct and don't include the error in the log?

exchanges/okx/okx_websocket.go
```diff
@@ -282,56 +275,17 @@ func (ok *Okx) WsAuth(ctx context.Context, dialer *websocket.Dialer) error {
...
-			continue
-		case <-timer.C:
-			timer.Stop()
-			return fmt.Errorf("%s websocket connection: timeout waiting for response with an operation: %v",
-				ok.Name,
-				req.Operation)
-		}
-	}
+	var nothing interface{}
+	err = ok.SendAuthenticatedWebsocketRequest(ctx, "login-response", operationLogin, args, &nothing)
```
gloriousCode, line 287:
```suggestion
	err = ok.SendAuthenticatedWebsocketRequest(ctx, "login-response", operationLogin, args, nil)
```

Remove the `response` `nil` check from `SendAuthenticatedWebsocketRequest` and you won't need to jank it up

gloriousCode (changes requested):
I AM _MOSTLY_ HAPPY THANK YOU

exchanges/okx/okx_websocket_requests.go
shazbert: Removed both logs after merge with master

exchanges/okx/okx_websocket_requests.go
```diff
@@ -0,0 +1,218 @@
...
+		Data: response,
+	}
+
+	if err := json.Unmarshal(incoming, &intermediary); err != nil {
+		return err
+	}
+
+	if intermediary.Code > 1 {
+		return fmt.Errorf("error code: %d message: %s", intermediary.Code, intermediary.Message)
+	}
```
gloriousCode, line 215: Order code 2 was previously handled. Your comments on partial success in those functions likely won't get hit with a response of 2

gloriousCode (changes requested):

exchanges/okx/okx_websocket_requests.go
shazbert: Yeah good catch; I bumped it to 2 so that anything beyond that can fail here, 1&2 codes don't actually need to fail here. None of my testing actually returned a code 2 that's why I must have forgotten to put it in. Example:

```
Okx wss://ws.okx.com:8443/ws/v5/private: Sending message: {"id":"1","op":"batch-orders","args":[{"instId":"BTC-USDT","tdMode":"cash","ccy":"USDT","side":"buy","ordType":"limit","sz":"0.0001","px":"-1"},{"instId":"BTC-USDT","tdMode":"cash","ccy":"USDT","side":"buy","ordType":"limit","sz":"0.0001","px":"20000"}]}
Okx wss://ws.okx.com:8443/ws/v5/private: Received response [1/1]: {"id":"1","op":"batch-orders","code":"2","msg":"","data":[{"tag":"","ts":"1728858915989","ordId":"","clOrdId":"","sCode":"51000","sMsg":"Parameter px  error"},{"tag":"","ts":"1728858915990","ordId":"1891010883570696192","clOrdId":"","sCode":"0","sMsg":"Order successfully placed."}],"inTime":"1728858915989134","outTime":"1728858915990899"}
```

Correct code 1 is a failure but the data holds the specifics of the error in sCode and sMsg and that can be checked in the calling function.

gloriousCode: I get that you're trying to put the onus on the caller, but you also don't give the caller the opportunity to do so since you only return the `Data` property and nothing else. Then with this:
```go
if len(orderResp) != 1 {
		return nil, errNoValidResponseFromServer
	}
```
There could be a `Message` and `Code` from the parent which could contain details but its instead thrown out and you don't know why.

shazbert: All code 1 and code 2 API examples don't produce a message; 1 is a total failure and 2 is a partial failure which have data returned with specific scode and smsg to each action. From testing anything above 2 which has a message component; I have only seen request errors pertaining to operation string, connection authentication issues and they don't send any data back for the caller. The only reason I added the length check was to protect against a possible panic. 

Happy to add in whatever you want though send me a diff on what you propose and the response you have found anything while testing and I will update it.

shazbert:
In favour of #1851

PR #1601, opened 2024-08-03
Bybit: Add subscription configuration for spot

gbjk:
## Bybit

* Adds subscription configuration
* Adds candles to the default subs

This PR does not touch any of the assets that aren't used by the engine by default.
It just fixes the default subs for the only available asset for the websocket as it would normally run.

## Stacked Dependencies

- [x] #1579 

## Type of change

### DIFF
--- a/exchanges/bybit/bybit_test.go
+++ b/exchanges/bybit/bybit_test.go
@@ -4,13 +4,16 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"slices"
 	"testing"
 	"time"
 
 	"github.com/gofrs/uuid"
+	"github.com/gorilla/websocket"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/common/key"
 	"github.com/thrasher-corp/gocryptotrader/currency"
 	exchange "github.com/thrasher-corp/gocryptotrader/exchanges"
@@ -22,8 +25,11 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/sharedtestvalues"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/stream"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/subscription"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
 	testexch "github.com/thrasher-corp/gocryptotrader/internal/testing/exchange"
+	testsubs "github.com/thrasher-corp/gocryptotrader/internal/testing/subscriptions"
+	testws "github.com/thrasher-corp/gocryptotrader/internal/testing/websocket"
 	"github.com/thrasher-corp/gocryptotrader/portfolio/withdraw"
 	"github.com/thrasher-corp/gocryptotrader/types"
 )
@@ -3682,4 +3688,91 @@ func TestGetCurrencyTradeURL(t *testing.T) {
 		require.NoError(t, err)
 		assert.NotEmpty(t, resp)
 	}
+}
+
+// TestGenerateSubscriptions exercises generateSubscriptions
+func TestGenerateSubscriptions(t *testing.T) {
+	t.Parallel()
+
+	b := new(Bybit)
+	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+
+	b.Websocket.SetCanUseAuthenticatedEndpoints(true)
+	subs, err := b.generateSubscriptions()
+	require.NoError(t, err, "generateSubscriptions must not error")
+	exp := subscription.List{}
+	for _, s := range b.Features.Subscriptions {
+		for _, a := range b.GetAssetTypes(true) {
+			if s.Asset != asset.All && s.Asset != a {
+				continue
+			}
+			pairs, err := b.GetEnabledPairs(a)
+			require.NoErrorf(t, err, "GetEnabledPairs %s must not error", a)
+			pairs = common.SortStrings(pairs).Format(currency.PairFormat{Uppercase: true, Delimiter: ""})
+			s := s.Clone() //nolint:govet // Intentional lexical scope shadow
+			s.Asset = a
+			if isSymbolChannel(channelName(s)) {
+				for i, p := range pairs {
+					s := s.Clone() //nolint:govet // Intentional lexical scope shadow
+					switch s.Channel {
+					case subscription.OrderbookChannel:
+						s.QualifiedChannel = fmt.Sprintf("%s.%d.%s", channelName(s), s.Levels, p)
+					default:
+						s.QualifiedChannel = channelName(s) + "." + p.String()
+					}
+					s.Pairs = pairs[i : i+1]
+					exp = append(exp, s)
+				}
+			} else {
+				s.Pairs = pairs
+				s.QualifiedChannel = channelName(s)
+				exp = append(exp, s)
+			}
+		}
+	}
+	testsubs.EqualLists(t, exp, subs)
+}
+
+func TestSubscribe(t *testing.T) {
+	t.Parallel()
+	b := new(Bybit)
+	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+	subs, err := b.Features.Subscriptions.ExpandTemplates(b)
+	require.NoError(t, err, "ExpandTemplates must not error")
+	b.Features.Subscriptions = subscription.List{}
+	testexch.SetupWs(t, b)
+	err = b.Subscribe(subs)
+	require.NoError(t, err, "Subscribe must not error")
+}
+
+func TestAuthSubscribe(t *testing.T) {
+	t.Parallel()
+	b := new(Bybit)
+	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+	b.Websocket.SetCanUseAuthenticatedEndpoints(true)
+	subs, err := b.Features.Subscriptions.ExpandTemplates(b)
+	require.NoError(t, err, "ExpandTemplates must not error")
+	b.Features.Subscriptions = subscription.List{}
+	success := true
+	mock := func(tb testing.TB, msg []byte, w *websocket.Conn) error {
+		tb.Helper()
+		var req SubscriptionArgument
+		require.NoError(tb, json.Unmarshal(msg, &req), "Unmarshal must not error")
+		require.Equal(tb, "subscribe", req.Operation)
+		msg, err = json.Marshal(SubscriptionResponse{
+			Success:   success,
+			RetMsg:    "Mock Resp Error",
+			RequestID: req.RequestID,
+			Operation: req.Operation,
+		})
+		require.NoError(tb, err, "Marshal must not error")
+		return w.WriteMessage(websocket.TextMessage, msg)
+	}
+	b = testexch.MockWsInstance[Bybit](t, testws.CurryWsMockUpgrader(t, mock))
+	b.Websocket.AuthConn = b.Websocket.Conn
+	err = b.Subscribe(subs)
+	require.NoError(t, err, "Subscribe must not error")
+	success = false
+	err = b.Subscribe(subs)
+	assert.ErrorContains(t, err, "Mock Resp Error", "Subscribe should error containing the returned RetMsg")
 }
--- a/exchanges/bybit/bybit_test.go
+++ b/exchanges/bybit/bybit_test.go
@@ -3715,6 +3715,8 @@ func TestGenerateSubscriptions(t *testing.T) {
 				for i, p := range pairs {
 					s := s.Clone() //nolint:govet // Intentional lexical scope shadow
 					switch s.Channel {
+					case subscription.CandlesChannel:
+						s.QualifiedChannel = fmt.Sprintf("%s.%.f.%s", channelName(s), s.Interval.Duration().Minutes(), p)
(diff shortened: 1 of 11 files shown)


shazbert (approved):
tACK.

exchanges/subscription/subscription.go
```diff
@@ -31,6 +31,7 @@ const (
 	AllTradesChannel = "allTrades"
 	MyTradesChannel  = "myTrades"
 	MyOrdersChannel  = "myOrders"
+	WalletChannel    = "wallet"
```
gbjk, line 34: Wondering if this should be MyWalletChannel 🤷

shazbert: That's more consistent with what you have so sure.

gbjk: Fixed 415c88fd

exchanges/bybit/bybit_websocket.go
```diff
@@ -7,15 +7,16 @@ import (
 	"net/http"
```
gloriousCode, line 7: With this in your opener:
>This PR does not:
>    Attempt to fix subscription error handling

And no further details means I don't know what issue(s) you are referring to

gloriousCode: As such, here's the error I'm encountering
```
[DEBUG] | WEBSOCKET | 10/10/2024 09:08:14 | Bybit websocket connection: message received: {"req_id":"128873076","success":false,"ret_msg":"Batch subscription partially succeeded and partially failed.Successful subscriptions are as follows:[order position execution]. Subscription to the following topics failed because the topic does not exist or there is a subscription conflict:[myWallet.BTCUSDT].","op":"subscribe","conn_id":"cmjom4cfnontv8pqvlug-28whu5"}
```

Now you have to tell me if this is not an issue 😄

exchanges/bybit/bybit_websocket.go
```diff
@@ -54,6 +55,24 @@ const (
 	websocketPrivate = "wss://stream.bybit.com/v5/private"
 )
 
+var defaultSubscriptions = subscription.List{
+	{Enabled: true, Asset: asset.Spot, Channel: subscription.TickerChannel},
+	{Enabled: true, Asset: asset.Spot, Channel: subscription.OrderbookChannel, Levels: 50},
+	{Enabled: true, Asset: asset.Spot, Channel: subscription.AllTradesChannel},
+	{Enabled: true, Asset: asset.Spot, Authenticated: true, Channel: subscription.MyOrdersChannel},
```
gloriousCode, line 62:
```suggestion
	{Enabled: true, Asset: asset.Spot, Channel: subscription.CandlesChannel, Interval: kline.OneHour},
	{Enabled: true, Asset: asset.Spot, Authenticated: true, Channel: subscription.MyOrdersChannel},
```

exchanges/bybit/bybit_websocket.go
```diff
@@ -54,6 +55,24 @@ const (
...
+	{Enabled: true, Asset: asset.Spot, Authenticated: true, Channel: chanPositions},
+	{Enabled: true, Asset: asset.Spot, Authenticated: true, Channel: chanExecution},
+}
+
+var subscriptionNames = map[string]string{
+	subscription.TickerChannel:    chanPublicTicker,
+	subscription.OrderbookChannel: chanOrderbook,
+	subscription.AllTradesChannel: chanPublicTrade,
+	subscription.MyOrdersChannel:  chanOrder,
+	subscription.MyTradesChannel:  chanExecution,
```
gloriousCode, line 73:
```suggestion
	subscription.MyTradesChannel:  chanExecution,
	subscription.CandlesChannel:   chanKline,
```

Proof of success:
```
[DEBUG] | WEBSOCKET | 10/10/2024 09:20:36 | Bybit websocket connection: sending message [{"req_id":"105987425","op":"subscribe","args":["tickers.BTCUSDT","orderbook.50.BTCUSDT","publicTrade.BTCUSDT","kline.60.BTCUSDT"]}]
...
[DEBUG] | WEBSOCKET | 10/10/2024 09:20:36 | Bybit websocket connection: message received: {"type":"snapshot","topic":"kline.60.BTCUSDT","data":[{"start":1728511200000,"end":1728514799999,"interval":"60","open":"60901.27","close":"60772.63","high":"60912.85","low":"60742.14","volume":"179.015874","turnover":"10888078.92165951","confirm":false,"timestamp":1728512435550}],"ts":1728512435550}
```

exchanges/bybit/bybit_test.go
```diff
@@ -3683,3 +3687,58 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+		}
+	}
+	testsubs.EqualLists(t, exp, subs)
+}
+
+func TestSubscribe(t *testing.T) {
+	t.Parallel()
+	b := new(Bybit)
+	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+	subs, err := b.Features.Subscriptions.ExpandTemplates(b)
```
gloriousCode, line 3738: Any ability to test generation of auth subs?

gloriousCode (changes requested):
tyty

exchanges/bybit/bybit_websocket.go
gbjk: This is an issue 😂 
It's due to the lack of a subscription name for `myWallet`.
What I don't understand at all is how you get a completely different error from me.
I get:
```
{"success":false,"ret_msg":"Invalid topic :[myWallet.BTCUSDT]","conn_id":"66a98329-eb29-480e-8423-5af14f82f730","req_id":"119798340","op":"subscribe"}
```

You got a partial subscription, but I got a failed subscription.
I caused mine by sending `myWallet.*` to a non-authed endpoint.
Did you get an ERROR message in your logs after your DEBUG message?

Fixed c95e40513

What I meant by my comment was that I'm not touching subscription mechanics at all until v5 lands, and maybe multi websockets. After that, I'll try to address partial subscribe failures.

exchanges/bybit/bybit_websocket.go
gbjk: Happy to add this.
You realise it wasn't enabled before, right?

I'll update the PR notes.

Fixed 0acd68ea6

exchanges/bybit/bybit_test.go
gbjk: That's tested above in TestGenerateSubscriptions.
Just double checked to be sure.

gloriousCode: I misspoke, I meant to say subscribe to auth subs

gbjk: For live testing, right?
I think if you just set apiKey and apiSecret and then run with -tags mock_test_off it should JFW.

Do you think we need a mock test for it?
I just realised that we hit the live server WS either way, I think :'(

gloriousCode: Excuse my mess, but how about something like this?
```go
func TestSubscribe(t *testing.T) {
	t.Parallel()
	b := new(Bybit)
	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
	b.Websocket.SetCanUseAuthenticatedEndpoints(true)
	subs, err := b.Features.Subscriptions.ExpandTemplates(b)
	require.NoError(t, err, "ExpandTemplates must not error")
	b.Features.Subscriptions = subscription.List{}
	mock := func(msg []byte, w *websocket.Conn) error {
		var req SubscriptionArgument
		require.NoError(t, json.Unmarshal(msg, &req), "Unmarshal should not error")
		require.Equal(t, req.Operation, "subscribe")
		msg, err := json.Marshal(SubscriptionResponse{
			Success:   true,
			RetMsg:    "OK",
			RequestID: req.RequestID,
			Operation: req.Operation,
		})
		require.NoError(t, err, "Marshal should not error")
		return w.WriteMessage(websocket.TextMessage, msg)
	}
	b = testexch.MockWsInstance[Bybit](t, testexch.CurryWsMockUpgrader(t, mock))
	b.Websocket.AuthConn = b.Websocket.Conn
	err = b.Subscribe(subs)
	require.NoError(t, err, "Subscribe must not error")
}
```

gbjk: Thank you 🙏 
Added failure case as well. Abused RetMsg very slightly in the positive case, but no big deal.

Fixed c5e16335

exchanges/bybit/bybit_test.go
```diff
@@ -3683,3 +3688,91 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+	require.NoError(t, err, "ExpandTemplates must not error")
+	b.Features.Subscriptions = subscription.List{}
+	testexch.SetupWs(t, b)
+	err = b.Subscribe(subs)
+	require.NoError(t, err, "Subscribe must not error")
+}
+
+func TestAuthSubcribe(t *testing.T) {
```
gloriousCode, line 3749: Critical issue
```suggestion
func TestAuthSubscribe(t *testing.T) {
```

exchanges/bybit/bybit_test.go
```diff
@@ -3683,3 +3688,91 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+		msg, err = json.Marshal(SubscriptionResponse{
+			Success:   success,
+			RetMsg:    "Mock Resp Error",
+			RequestID: req.RequestID,
+			Operation: req.Operation,
+		})
+		require.NoError(t, err, "Marshal must not error")
+		return w.WriteMessage(websocket.TextMessage, msg)
+	}
+	b = testexch.MockWsInstance[Bybit](t, testexch.CurryWsMockUpgrader(t, mock))
```
gloriousCode, line 3771: I don't know what's going on in linter land, but please check

gloriousCode (changes requested):
Thanks for adding my test. I'm good after this

exchanges/bybit/bybit_test.go
gbjk: Fixed [96d5997](https://github.com/thrasher-corp/gocryptotrader/pull/1601/commits/96d5997b66b28c66f1120e37ca2589daa2a60320)

exchanges/bybit/bybit_test.go
gbjk: Was a minor issue with upstream master moving the location and signature of `CurryWsMockUpgrader`
Since it was minimal, I've squashed it down and rebased.
That was necessary to fix this anyway, since the problem came from CI merging in master during testing.

exchanges/bybit/bybit_test.go
```diff
@@ -3683,3 +3689,92 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+}
+
+func TestSubscribe(t *testing.T) {
+	t.Parallel()
+	b := new(Bybit)
+	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+	subs, err := b.Features.Subscriptions.ExpandTemplates(b)
+	require.NoError(t, err, "ExpandTemplates must not error")
+	b.Features.Subscriptions = subscription.List{}
+	testexch.SetupWs(t, b)
```
gloriousCode, line 3745: Looks like we got a racy race car 🏎️ 
https://github.com/thrasher-corp/gocryptotrader/actions/runs/11490767919/job/31982037417?pr=1601#step:10:3624

gbjk: Unrelated to this PR.

Fixed c669bf02a

Also moved to a RWMutex at the same time

I don't believe this is a prod bug, because the orderbook mutex should protect it, so I'm not going to sling it outside this PR.

gloriousCode (approved):
I noted no loss when benchmarking using a RWMutex for the Depths

exchanges/orderbook/depth.go
```diff
@@ -46,7 +46,7 @@ type Depth struct {
 	// validationError defines current book state and why it was invalidated.
 	validationError error
 
-	m sync.Mutex
+	m sync.RWMutex
```
shazbert, line 49: There is a good amount of writes incoming from a websocket connection on this mutex, this adds an addition ~2ns per operation than a normal lock. Change not required though. As this stops the raciness and I haven't as yet had any issues in any real world trading. 

What I would probably do in the future if it does become a problem is revert it back to a normal lock and shift the stream/buffer package directly into the orderbook package where we can lock it out once and directly check the variables as we go.

shazbert (approved):
Looks good tACK. Can't wait for the conflict pain on my bybit branch.

exchanges/orderbook/depth.go
gbjk: For me, this isn't about raciness, but concurrency.
We can't control what users are doing with the methods available to them.
A RWMutex is the best way I can think of to protect write performance in the face of an unpredictable number of reads.
Even if we shift everything to one package, the previous implementations of an OB I've written have still needed a RWMutex over a straight mutex. Where I've tried to do away with the mutex by using a single select, I've found that it stopped-the-world for too long in several situations.
Maybe I'm missing your point.

shazbert: I don't think you are missing my point, we both want the same goal. What your saying makes sense for a read heavy load but the most important thing is write performance and I guess further testing can be done. 

> We can't control what users are doing with the methods available to them.
👻 They are dead to me. I have external operations that only use isolated wrapper functionality. If they are being silly billies then that's on them. 

> A RWMutex is the best way I can think of to protect write performance in the face of an unpredictable number of reads.
You're probably correct on this assumption. 

> Even if we shift everything to one package, the previous implementations of an OB I've written have still needed a RWMutex over a straight mutex. Where I've tried to do away with the mutex by using a single select, I've found that it stopped-the-world for too long in several situations.
What use case was this? Like is this for a market maker bot or similar system?

thrasher- (approved):
tACK public and private subs. Nice!
image

PR #1611, opened 2024-08-14
Kraken: Fix TestGetOpenInterest

gbjk:
We see daily failures on OpenInterest for Kraken.
This fix assumes that the issue might be related to volume of ETHUSD open interest, and switches the single pair test to XBTUSD instead

Also isolates the test from others, since we're changing stored pairs and we might be colliding on the global k

If we keep getting the issue after this, then we remove the NotEmpty errors I guess.

## Type of change

- [x] Test fix

### DIFF
--- a/exchanges/kraken/kraken_futures.go
+++ b/exchanges/kraken/kraken_futures.go
@@ -377,20 +377,46 @@ func (k *Kraken) SendFuturesAuthRequest(ctx context.Context, method, path string
 		}, nil
 	}
 
-	if err := k.SendPayload(ctx, request.Unset, newRequest, request.AuthenticatedRequest); err != nil {
-		return err
+	err = k.SendPayload(ctx, request.Unset, newRequest, request.AuthenticatedRequest)
+
+	if err == nil {
+		err = getFuturesErr(interim)
 	}
 
-	var resp genericFuturesResponse
-	if err := json.Unmarshal(interim, &resp); err != nil {
-		return fmt.Errorf("%w %w", request.ErrAuthRequestFailed, err)
-	} else if resp.Error != "" && resp.Result != "success" {
-		return fmt.Errorf("%w %v", request.ErrAuthRequestFailed, resp.Error)
+	if err == nil {
+		err = json.Unmarshal(interim, result)
 	}
 
-	if err := json.Unmarshal(interim, result); err != nil {
+	if err != nil {
 		return fmt.Errorf("%w %w", request.ErrAuthRequestFailed, err)
 	}
 
 	return nil
+}
+
+func getFuturesErr(msg json.RawMessage) error {
+	var resp genericFuturesResponse
+	if err := json.Unmarshal(msg, &resp); err != nil {
+		return err
+	}
+
+	// Result may be omitted entirely, so we don't test for == "success"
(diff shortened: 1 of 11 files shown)


gloriousCode (approved):
tACK!
This is one of those instances where I can't get the old test to fail or even find a CI failure of it. But the theory of using a shadow-kraken 🦑  and the more popular/less poopcoin pair should resolve things

gbjk:
> tACK! This is one of those instances where I can't get the old test to fail or even find a CI failure of it.

[here](https://github.com/thrasher-corp/gocryptotrader/actions/runs/10364575815/job/28690105595?pr=1586) and many others, but very inconsistent
```
--- FAIL: TestGetOpenInterest (12.05s)
    kraken_test.go:2271: 
        	Error Trace:	/home/runner/work/gocryptotrader/gocryptotrader/exchanges/kraken/kraken_test.go:2271
        	Error:      	Should NOT be empty, but was []
        	Test:       	TestGetOpenInterest
```

gbjk:
> Could you please look over and maybe handle the above, and check out the error response [here](https://docs.futures.kraken.com/#http-api-trading-v3-api-market-data-get-tickers)

Fixed 59b98a33

gbjk:
Following discussion, we're concluding the issue was nearly definitely the pairs on the shared k.

I've added the `errors` as requested by @shazbert . That span a little out of control but not awful.

gbjk:
Hold off, I've broken a few other tests. Will fix and reply when done.

gbjk:
Fixed a77fb448

Had to back away from equality testing Result == "success", because of the APIs that don't return it (Charts).
Considered using `jsonparser` so we can tell if the Result is present and then if it is require it to be success, but this will do for now.

shazbert (changes requested):
tACK, just a linter issue needs to be rectified then this looks good to go. Thanks for checking that out.

gbjk:
> tACK, just a linter issue needs to be rectified then this looks good to go. Thanks for checking that out.

Actually, looks like I dropped a line.

Fixed b3c0268a

shazbert (approved):
ch ACK! Nice work.

gloriousCode (approved):
tACK! While I haven't had the misfortune of getting multiple errors back from Kraken Futures, I appreciate the added support

exchanges/kraken/kraken_futures.go
```diff
@@ -377,20 +377,46 @@ func (k *Kraken) SendFuturesAuthRequest(ctx context.Context, method, path string
...
 
-	if err := json.Unmarshal(interim, result); err != nil {
+	if err != nil {
 		return fmt.Errorf("%w %w", request.ErrAuthRequestFailed, err)
 	}
 
 	return nil
 }
+
+func getFuturesErr(msg json.RawMessage) error {
```
thrasher-, line 397: Now that this has been split off, please add direct test coverage

thrasher- (changes requested):
Thanks for looking into this issue, just one nit

exchanges/kraken/kraken_futures.go
gbjk: Fixed 5370934c

thrasher- (approved):

PR #1613, opened 2024-08-15
stream/match: allow a single connection to maintain its own match lookup for multi-connection

shazbert:
# PR Description

* Attempts to decouple a connection from a global lookup table for outbound request to response matching. This allows for generic startup messages to be used across future multiple connections of the same type. This also minimises the chance of a signature conflict on outbound requests. Potentially reduces the semacquire slowness when dealing with many concurrently spawned and subscribed return responses. 

* This is backwards compatible. 

* Starts implementation in Gateio due to #1580 base. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


gloriousCode (changes requested):
Thank you for adding this. I would just like to see a test from gateio_test.go along similar lines to this: https://github.com/thrasher-corp/gocryptotrader/pull/1601#discussion_r1798683818
which would allow for your changes to be hit

shazbert:
> Thank you for adding this. I would just like to see a test from gateio_test.go along similar lines to this: [#1601 (comment)](https://github.com/thrasher-corp/gocryptotrader/pull/1601#discussion_r1798683818) which would allow for your changes to be hit

If you want that I will just wait until #1598 is merged it has all the auth and outbound inbound testing done there.

exchange/websocket/manager.go
```diff
@@ -397,7 +403,7 @@ func (m *Manager) getConnectionFromSetup(c *ConnectionSetup) *connection {
 		readMessageErrors:        m.ReadMessageErrors,
 		shutdown:                 m.ShutdownC,
 		Wg:                       &m.Wg,
-		Match:                    m.Match,
+		Match:                    match,
```
gloriousCode, line 406: I think this can now be unexported thanks to `(c *connection) RequireMatchWithData`
See diff:
https://gist.github.com/gloriousCode/663741b41fe73fa48c7c60f64bea2499

exchanges/gateio/gateio_test.go
```diff
@@ -2827,7 +2827,7 @@ const futuresCandlesticksPushData = `{"time": 1678469467, "time_ms": 16784694679
 
 func TestFuturesCandlestickPushData(t *testing.T) {
 	t.Parallel()
-	err := g.WsHandleFuturesData(t.Context(), []byte(futuresCandlesticksPushData), asset.Futures)
+	err := g.WsHandleFuturesData(t.Context(), nil, []byte(futuresCandlesticksPushData), asset.Futures)
```
gloriousCode, line 2830: Having `nil` here was my least favourite part of reviewing. I think I've gone over the potential panic scenarios and can conclude that it _shouldn't_ happen, but you've now allowed the possibility to happen.

We all see websocket as highly important functionality though and I have full trust that panics as a result of this change will be addressed with high priority

gloriousCode (approved):
tACK! Thanks for updating with tests and the package movement

exchange/websocket/manager.go
shazbert: Good idea! There's a couple exchange which have auth and unauth data funneling into the same ws data handler and no real way of differentiating between the two connections from that when it's not using multiconnection ws. I think I might defer this once we do a full upgrade to that across exchanges.

exchange/websocket/manager_test.go
```diff
@@ -1336,3 +1336,18 @@ func TestGetConnection(t *testing.T) {
...
+	err := ws.RequireMatchWithData(0, nil)
+	require.ErrorIs(t, err, ErrSignatureNotMatched)
+
+	ch, err := ws.Match.Set(0, 1)
+	require.NoError(t, err)
+
+	err = ws.RequireMatchWithData(0, []byte("test"))
+	require.NoError(t, err)
+
+	require.Equal(t, []byte("test"), <-ch)
```
gbjk: 🐒 
* I think this is an assert, not a require
* Won't this block badly if there's nothing there?
`d, ok :=`

gbjk (approved):
Good work 👍

thrasher- (approved):
LGTM!

exchange/websocket/manager_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1613/commits/70cea726c96e36df4bd98f251bed04660cdd112d

gbjk (approved):
Good work; Thanks

exchanges/gateio/gateio_websocket_option.go
```diff
@@ -295,14 +295,14 @@ func (e *Exchange) OptionsUnsubscribe(ctx context.Context, conn websocket.Connec
 }
 
 // WsHandleOptionsData handles options websocket data
-func (e *Exchange) WsHandleOptionsData(ctx context.Context, respRaw []byte) error {
+func (e *Exchange) WsHandleOptionsData(ctx context.Context, conn websocket.Connection, respRaw []byte) error {
 	push, err := parseWSHeader(respRaw)
```
samuael, line 299: Shouldn't we add a check for the `conn` here? In-case a nil value of `conn` is provided

samuael (approved):
The PR looks perfect. ✅ 

I just left one optional code suggestion

exchanges/gateio/gateio_websocket_option.go
shazbert: Happy to panic here on that, all parameters will be provided by the websocket manager from its reader routine, if it is nil, something very wrong is very wrong. 🦇 🧛 🦇

thrasher- (approved):
tACK, GateIO ws working well🌞

PR #1624, opened 2024-08-20
BTCMarkets: Add subscription configuration

gbjk:
### BTCMarkets

* Add subscription configuration
* Add `subscription.List.GroupByPairs` and `subscription.List.Authenticated`
* Group subscriptions by pairs to reduce subscription messages
* Rename default config exchange name to match package name
* Subscribe to heartbeat when not authenticated

## Type of change


exchanges/subscription/list.go
```diff
@@ -46,6 +46,22 @@ func (l List) GroupPairs() (n List) {
 	return s.List()
 }
 
+// GroupByPairs groups subscriptions which have the same Pairs
+func (l List) GroupByPairs() []List {
+	n := []List{}
+outer:
+	for _, a := range l {
+		for i, b := range n {
+			if a.Pairs.Equal(b[0].Pairs) {
```
shazbert, line 55: b "could be zero", any chance you could you could check length and panic with some extra info so *if* it does happen there's slightly more context?

shazbert (changes requested):
Nice work! Just one nit.

exchanges/subscription/list.go
gbjk: Maybe ... by "could be zero" you meant that if someone came in here and broke this code?

gbjk: Fixed b91e6115

panic is implicit if anyone anyone breaks this. Comment to guard against tripwire.

shazbert (approved):
ch ACK. Thanks.

internal/testing/exchange/exchange.go
```diff
@@ -38,11 +37,8 @@ func Setup(e exchange.IBotExchange) error {
 	if err != nil {
 		return fmt.Errorf("LoadConfig() error: %w", err)
 	}
-	parts := strings.Split(fmt.Sprintf("%T", e), ".")
```
gloriousCode, line 41: Just a curiosity, what use-case did this originally address?

exchanges/btcmarkets/btcmarkets_websocket.go
```diff
@@ -358,26 +359,33 @@ func (b *BTCMarkets) Subscribe(subs subscription.List) error {
 	}
 
 	var errs error
-	for _, s := range subs {
-		if baseReq.Key == "" && common.StringSliceContains(authChannels, s.Channel) {
-			if err := b.authWsSubscibeReq(baseReq); err != nil {
-				return err
+	if authed := subs.Private(); len(authed) > 0 {
```
gloriousCode, line 362: While I'd appreciate a test for this func, it doesn't have the same level of changes like in Bitstamp PR and so I will not force the issue

gloriousCode (approved):
tACK! Nice work

internal/testing/exchange/exchange.go
thrasher-: A way to get the exchange name from the type (e.g btcmarkets.BTCMarkets). The problem though is that for this exchange the name is set to "BTC Markets" via SetDefaults and thus engine also expects "btc markets" https://github.com/thrasher-corp/gocryptotrader/blob/master/engine/helpers.go#L1010

gbjk: Just getting exchange name, without calling `SetDefaults`. Which turned out to not work for several exchanges.

gbjk:
Fixed test wording inconsistency

Fixed https://github.com/gbjk/gocryptotrader/commit/f49c4eef4

exchanges/btcmarkets/btcmarkets_test.go
```diff
@@ -1134,3 +1136,33 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+	t.Parallel()
+	b := new(BTCMarkets)
+	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+	p := currency.Pairs{currency.NewPairWithDelimiter("BTC", "USD", "_"), currency.NewPairWithDelimiter("ETH", "BTC", "_")}
+	require.NoError(t, b.CurrencyPairs.StorePairs(asset.Spot, p, false))
+	require.NoError(t, b.CurrencyPairs.StorePairs(asset.Spot, p, true))
+	b.Websocket.SetCanUseAuthenticatedEndpoints(true)
+	require.True(t, b.Websocket.CanUseAuthenticatedEndpoints(), "CanUseAuthenticatedEndpoints must return true")
+	subs, err := b.generateSubscriptions()
+	require.NoError(t, err, "generateSubscriptions should not error")
```
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/6daa32bcc

exchanges/btcmarkets/btcmarkets_websocket.go
gbjk: I think I backed away from TestSubscribe because of the low priority of this exchange.
I have it in a distant TODO, though.

thrasher- (approved):
tACK and thanks for making those additional changes👍

shazbert (approved):
Nice work! tACK

```
[DEBUG] | WEBSOCKET | 05/12/2024 13:20:02 | New Order: BTC Markets spot BTC-AUD NEW LIMIT BID OrderID:14590027440 ClientOrderID: Price:0.000000 Amount:0.001000 Executed Amount:0.000000 Remaining Amount:0.001000
```

ws message doesn't have a price so that's not an issue

PR #1629, opened 2024-08-23
Subscriptions: Add List.Enabled, List.Public and List.Private

gbjk:
## Type of change

### DIFF
--- a/exchanges/subscription/list.go
+++ b/exchanges/subscription/list.go
@@ -113,3 +113,36 @@ func (l List) assetPairs(e iExchange) (assetPairs, error) {
 	}
 	return ap, nil
 }
+
+// Enabled returns a new list of only enabled subscriptions
+func (l List) Enabled() List {
+	n := make(List, 0, len(l))
+	for _, s := range l {
+		if s.Enabled {
+			n = append(n, s)
+		}
+	}
+	return slices.Clip(n)
+}
+
+// Private returns only subscriptions which require authentication
+func (l List) Private() List {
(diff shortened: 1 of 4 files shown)


shazbert (approved):
tACK.

exchanges/subscription/list.go
```diff
@@ -113,3 +113,36 @@ func (l List) assetPairs(e iExchange) (assetPairs, error) {
...
+	for _, s := range l {
+		if s.Enabled {
+			n = append(n, s)
+		}
+	}
+	return slices.Clip(n)
+}
+
+// Private returns only Authenticated subscriptions
+func (l List) Private() List {
```
gloriousCode, line 129: Why introduce the field as Authenticated and then use a different adjective? rabble rabble rabble

gloriousCode (approved):
tACK

exchanges/subscription/list.go
gbjk: I struggled with naming on this.
When I added the `Public` method, things became clearer.

My historical issue was that `Authenticated` might sound like "Subscriptions which have already been authenticated".
And in that sense, I now want to improve this comment.
Fixed c1eb4f5e1

But you're right. I think I'd like to rename the field from Authenticated to Private too.
It's just more invasive than I wanted to be right now. I was thinking I'd do it gently once this group of PRs has landed.

thrasher- (approved):
Some lovely helper methods. tACK!

PR #1480, opened 2024-02-14
Coinbase: Update exchange implementation

cranktakular:
# PR Description

Updating Coinbase to the Advanced Trade and Sign In With Coinbase APIs, as well as a few tiny fixes of other parts of the codebase found along the way.

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally and on Github Actions with my changes

The only tests which seem to be failing (exchanges/kucoin, database/models/postgres, config, and common/file) are parts that I haven't substantively changed, and which seem to be failing on master too.

PR #1635, opened 2024-08-27
Subscriptions: Relax subscription validation for non-existent pairs

gbjk:
The subscription pairs do not need to be validated as enabled or available. The check was just belt-and-braces and didn't have a specific use-case in mind.

Coinbase has a use-case for wanting to subscribe to BTC-USD when it's not enabled.

Moreover, it shouldn't be our job to check this. You want a sub expanded with these pairs? Fine. Done. If it doesn't work, you can work out why

## Type of change

### DIFF
--- a/exchanges/subscription/template_test.go
+++ b/exchanges/subscription/template_test.go
@@ -84,12 +84,17 @@ func TestExpandTemplates(t *testing.T) {
 	}
 	equalLists(t, exp, got)
 
+	// Users can specify pairs which aren't available, even across diverse assets
+	// Use-case: Coinbasepro user sub for futures BTC-USD would return all BTC pairs and all USD pairs, even though BTC-USD might not be enabled or available
+	p := currency.Pairs{currency.NewPairWithDelimiter("BEAR", "PEAR", "🐻")}
+	got, err = List{{Channel: "expand-pairs", Asset: asset.All, Pairs: p}}.ExpandTemplates(e)
+	require.NoError(t, err, "Must not error with fictional pairs")
+	exp = List{{Channel: "expand-pairs", QualifiedChannel: "spot-PEARBEAR-expand-pairs@0", Asset: asset.Spot, Pairs: p}}
+	equalLists(t, exp, got)
+
 	// Error cases
 	_, err = List{{Channel: "nil"}}.ExpandTemplates(e)
 	assert.ErrorIs(t, err, errInvalidTemplate, "Should get correct error on nil template")
-
-	_, err = List{{Channel: "single-channel", Asset: asset.Spot, Pairs: currency.Pairs{currency.NewPairWithDelimiter("NOPE", "POPE", "🐰")}}}.ExpandTemplates(e)
-	assert.ErrorIs(t, err, currency.ErrPairNotContainedInAvailablePairs, "Should error correctly when pair not available")
 
 	e.tpl = "errors.tmpl"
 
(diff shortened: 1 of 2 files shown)


shazbert (changes requested):
One thing then looks good to go.

exchanges/subscription/template_test.go
```diff
@@ -84,13 +84,18 @@ func TestExpandTemplates(t *testing.T) {
 	}
 	equalLists(t, exp, got)
 
+	// Users can specify pairs which aren't available, even across diverse assets
+	// Use-case: Coinbasepro user sub for futures BTC-USD would return all BTC pairs and all USD pairs, even though BTC-USD might not be enabled op available
```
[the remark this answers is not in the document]
gbjk: Fixed e49f7aff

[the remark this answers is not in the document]
gbjk: Fixed e49f7aff

shazbert (approved):
Thanks! tACK.

exchanges/subscription/template_test.go
```diff
@@ -84,13 +84,18 @@ func TestExpandTemplates(t *testing.T) {
 	}
 	equalLists(t, exp, got)
 
+	// Users can specify pairs which aren't available, even across diverse assets
+	// Use-case: Coinbasepro user sub for futures BTC-USD would return all BTC pairs and all USD pairs, even though BTC-USD might not be enabled or available
```
gloriousCode, line 88: I'm fine with the changes regarding pair validation, but I am a bit confused with this.
This PR does not mention #1480 as otherwise we don't have CBP futures support. The other thing is that GCT will do a purge of any unlisted pairs:
Before:
```json
  "futures": {
      "assetEnabled": true,
      "enabled": "BTC-USD",
      "available": "..."
     },
```
After:
`Exchange CoinbasePro: [futures] Removing enabled pair(s) [BTC-USD] from enabled pairs list, as it isn't located in the available pairs list.`

```json
     "futures": {
      "assetEnabled": true,
      "enabled": "BIT-25OCT24-CDE",
      "available": "..."
     },
```

`[INFO]  | EXCHANGE | 23/10/2024 12:30:29 | CoinbasePro Asset type FUTURES:
	 Enabled pairs: [BIT-25OCT24-CDE]`

Because of this, I want to know what scenario you are speaking of and how to check it so that I can be happy the situation is resolved 🎉

gbjk: I *think* #1480 would be the right reference.
@cranktakular  raised it with me directly.
> This came up in a Coinbase test for a websocket channel which doesn't distinguish between asset types; a request for "BTC-USD" will return all spot, perpetual futures, and expiring futures items associated with those two currencies.

So I don't know how he's going to make the pair end up in the subscription pairs.
He's free to have it happen during the template called from GenerateSubscriptions, or set the subscription that way programmatically using GCT as a library.

This PR just aims to get out of the user's way and not introduce unnecessary checks.

gloriousCode (approved):
I'm happy with the changes. Since #1480 is near the end, I'll seek clarification there and seek any amendments to that comment if necessary. Your work is already a part of 1480, however the `if len(s.Pairs) != 0 {` does not even get satisfied on CoinbasePro  🤷 

tACK

exchanges/subscription/template_test.go
gloriousCode: Thanks for providing details! I'll seek clarification in 1480

thrasher- (approved):
RIP🕯️NOPE POPE and 👋 BEAR PEAR

PR #1641, opened 2024-09-06
stream/okx: allow rate limit definitions to be used by the stream package

shazbert:
# PR Description

image

This allows for rate limits to be defined across connections if needed. 

OKx: 
* Share REST trading endpoint limits so that the websocket system can use them and not 429 our lives away.

Request:
* Add function GetRateLimiterDefinitions() to export request defined limiters to be shared across websocket implementation if needed.

stream:
* Allow the setting of a global rate limiting set across all potential connections which is backwards compatible
* Add param epl request.EndpointLimit to outbound request functions so that it can be matched to a specific rate limiter 
* Add commentary to Connection interface funcs
* Adjust writToConn function and remove secondary IsConnected() check as it will just error at a writeconn level anyway when using a closed connection. Also add coverage.

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/stream/websocket_types.go
```diff
@@ -121,6 +125,13 @@ type WebsocketSetup struct {
 	// MaxWebsocketSubscriptionsPerConnection defines the maximum number of
 	// subscriptions per connection that is allowed by the exchange.
 	MaxWebsocketSubscriptionsPerConnection int
+
+	// RateLimitDefinitions contains the rate limiters shared between WebSocket and REST connections for all endpoints.
+	// These rate limits take precedence over any rate limits specified in individual connection configurations.
+	// If no connection-specific rate limit is provided and the endpoint does not match any of these definitions,
+	// an error will be returned. However, if a connection configuration includes its own rate limit,
+	// it will fall back to that configuration’s rate limit without raising an error.
+	RateLimitDefinitions request.RateLimitDefinitions
```
gloriousCode, line 134: I was just thinking, why not use the requester itself? Check it out here
https://gist.github.com/gloriousCode/15e1e8a54cdbafec7add64e6cad183e4
Not a big deal though

exchanges/stream/websocket_connection.go
```diff
@@ -67,52 +73,60 @@ func (w *WebsocketConnection) SendJSONMessage(ctx context.Context, data interfac
...
 		if w.Verbose {
 			log.Debugf(log.WebsocketMgr, "%v websocket connection: sending message [%s]\n", w.ExchangeName, message)
 		}
 		return w.Connection.WriteMessage(messageType, message)
 	})
 }
 
-func (w *WebsocketConnection) writeToConn(ctx context.Context, writeConn func() error) error {
+func (w *WebsocketConnection) writeToConn(ctx context.Context, epl request.EndpointLimit, writeConn func() error) error {
 	if !w.IsConnected() {
```
gloriousCode, line 86: With the other removal, we could put this after the locks. The other one I'd argue is more important

gloriousCode (changes requested):
Good stuff. Two minor bits

exchanges/stream/websocket_types.go
shazbert: Yeah that isn't a bad idea, the current diff though, doesn't look like it will be backwards compatible did it pass tests? error: corrupt patch at line 171 🪦 for me 😭. 

I personally prefer it being segregated, instead of relying on a correctly configured Requester, the okx implementation seems like an edge case, most definitions will be done per connection from what I quickly looked at, so keeping it as a more generalised component might be better. If I ever use the stream package for whatever reason to make a custom websocket link with no REST implementation point of view.

exchanges/stream/websocket_connection.go
shazbert: Yeah I did think about this, the only reason I had for keeping the upper one is so routines don't hang consecutively on the rate limit sleep when there is no available connection. When you are writing to the connection it does the check anyway and determines if its connected 🤷 and it's hard to get coverage for it trying to time execution. I won't resolve it just yet, maybe there is something else to consider by other reviews.

exchanges/stream/websocket_types.go
gloriousCode: Yeah that reasoning is fine for me. I was thinking along the lines of you already have a requester by that point and it slightly simplified a bit of your code, but you're right in that it increased a dependency which was unnecessary

gloriousCode (approved):
tACK!

thrasher- (approved):
Changes ACK. Was curious about the PingHandler requiring a rate limit as I thought it would be excluded from requests however it's probably best to leave support for it just in a specific exchange limits it

PR #1648, opened 2024-09-12
types: Add Time type from Gateio and share across codebase

shazbert:
# PR Description

Consolidate and share similar time type. The unmarshal method is more performant than the convert type, it also is more consistent and fixes improper time handling e.g. bybit.

NOTE: So any integer string length that is under 10 will be rejected, wondering if I should change that, but everything we deal with crypto wise should be past 2009 anyway and if we go tradfi route then that will make sense as well pre 2001. Let me know what you guys want me to handle, or if this is a good state for now. 

Also had a fast path optimisation in one of the convert functions but is now being done under the hood in the strconv package, I have backtested it so no need to introduce:

```go
	if strconv.IntSize == 32 && len(s) >= 10 {
		i, err := strconv.ParseInt(s, 10, 64)
		if err != nil {
			return err
		}
		*t = microTimestamp(time.UnixMicro(i))
		return nil
	}

	// Has Fast path optimisation when int == 64
	i, err := strconv.Atoi(s)
	if err != nil {
		return err
	}
```

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


types/time.go
```diff
@@ -1,37 +1,31 @@
...
-	zero     = []byte(`0`)
-	emptyStr = []byte(`""`)
-	zeroStr  = []byte(`"0"`)
-)
-
 // Time represents a time.Time object that can be unmarshalled from a float64 or string.
 type Time time.Time
 
 // UnmarshalJSON deserializes json, and timestamp information.
 func (a *Time) UnmarshalJSON(data []byte) error {
```
gloriousCode, line 15:
```suggestion
func (t *Time) UnmarshalJSON(data []byte) error {
```

types/time.go
```diff
@@ -1,37 +1,31 @@
...
 )
 
-var (
-	zero     = []byte(`0`)
-	emptyStr = []byte(`""`)
-	zeroStr  = []byte(`"0"`)
-)
-
 // Time represents a time.Time object that can be unmarshalled from a float64 or string.
 type Time time.Time
```
gloriousCode, line 13: Please add 
```go
func (t Time) String() string {
	return t.Time().String()
}
```

gloriousCode: How about this too
```go
func (t Time) MarshalJSON() ([]byte, error) {
	return t.Time().MarshalJSON()
}
```

exchanges/bybit/bybit_test.go
```diff
@@ -3226,7 +3226,7 @@ func TestWsTicker(t *testing.T) {
 				assert.Equal(t, 6780.866843, v.Volume, "Volume should be correct")
 				assert.Equal(t, "BTC_USDT", v.Pair.String(), "Pair should be correct")
 				assert.Equal(t, asset.Spot, v.AssetType, "AssetType should be correct")
-				assert.Equal(t, int64(233366401000), v.LastUpdated.UnixMilli(), "LastUpdated should be correct")
+				assert.Equal(t, int64(2333664010000), v.LastUpdated.UnixMilli(), "LastUpdated should be correct")
```
gloriousCode, line 3229: Please relook at this test. The dates are far in the future. I don't mind if you're just updating the time with a more recent one, but it makes it look like an error

I have verified that the WS data is otherwise coming back properly

gloriousCode (changes requested):
Tested each exchange you implemented this for.... 😭

gloriousCode (approved):
tytytytACK

exchanges/gateio/gateio_types.go
```diff
@@ -16,7 +16,7 @@ const (
 	settleUSD  = "usd"
 	settleUSDT = "usdt"
 
-	// time in force variables
+	// types.Timein force variables
```
thrasher-, line 19:
```suggestion
	// types.Time time in force variables
```

types/time.go
```diff
@@ -43,30 +37,40 @@ func (a *Time) UnmarshalJSON(data []byte) error {
...
+func (t Time) Time() time.Time { return time.Time(t) }
+
+// String returns a string representation of the time.
+func (t Time) String() string {
+	return t.Time().String()
+}
+
+// MarshalJSON serializes the time to json.
+func (t Time) MarshalJSON() ([]byte, error) {
+	return t.Time().MarshalJSON()
```
thrasher-, line 75: Just a note here that we'll need to be careful on exchanges using this for outbound requests since it's unlikely all exchanges would use RFC 3339

thrasher- (changes requested):
A nice consolidation, tested with one minor nit

thrasher- (approved):
Changes ACK, thanks!

PR #1652, opened 2024-09-18
Binance: Fix occassional race failure on TestWsOrderExecutionReport

gbjk:
Fixes:
```
binance_test.go:2608: expected type order.Detail, found *orderbook.Depth
```

## Type of change

### DIFF
--- a/exchanges/binance/binance_test.go
+++ b/exchanges/binance/binance_test.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"os"
 	"reflect"
-	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


gloriousCode (approved):
tACK!
I'm happy with the change. I've got a [gist here](https://gist.github.com/gloriousCode/c2081642f9c646bd8143e5e20370ac9a) which should prevent future problems and removes the mutex, but your change addressing things just fine

shazbert (approved):
ch ACK.

gbjk:
> I've got a [gist here](https://gist.github.com/gloriousCode/c2081642f9c646bd8143e5e20370ac9a) which should prevent future problems and removes the mutex, but your change addressing things just fine

Thanks! I did wonder about where the other data came from, sniffed for other things consuming DataHandler, but didn't take it further.

Fixed c0f11540

shazbert (approved):
Ch ACK! Thanks!

thrasher- (approved):
Thank you @gbjk and @gloriousCode for the further enhancements!

PR #1655, opened 2024-09-22
Huobi: Fix intermittent UpdateTickers failure on expiry

gbjk:
## Huobi

* Collect and Return all errors from UpdateTickers
* Fix [intermittent failure of TestUpdateTickers](https://github.com/thrasher-corp/gocryptotrader/actions/runs/10917737674/job/30301656625?pr=1652) and implicitly a failure of tickers in runtime too
```
--- FAIL: TestUpdateTickers (2.29s)
    huobi_test.go:2879: 
        	Error Trace:	/home/runner/work/gocryptotrader/gocryptotrader/exchanges/huobi/huobi_test.go:2879
        	            	no ticker found huobi ETH-241227 futures
        	Test:       	TestUpdateTickers
FAIL
```
* Make contract type errors consistent
* Convert and Send lowercase contract type regardless of user input ( Option to reject uppercase. I'm in favour of that, personally )

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/huobi/huobi_futures.go
+++ b/exchanges/huobi/huobi_futures.go
@@ -9,7 +9,9 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"slices"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/thrasher-corp/gocryptotrader/common"
@@ -77,11 +79,12 @@ const (
 	fTriggerOrderHistory       = "/api/v1/contract_trigger_hisorders"
 
 	uContractOpenInterest = "/linear-swap-api/v1/swap_open_interest"
-
-	fContractDateFormat = "060102"
 )
 
-var errInvalidContractType = errors.New("invalid contract type")
+var (
+	errInvalidContractType        = errors.New("invalid contract type")
+	errInconsistentContractExpiry = errors.New("inconsistent contract expiry date codes")
+)
 
 // FGetContractInfo gets contract info for futures
 func (h *HUOBI) FGetContractInfo(ctx context.Context, symbol, contractType string, code currency.Pair) (FContractInfoData, error) {
@@ -90,11 +93,11 @@ func (h *HUOBI) FGetContractInfo(ctx context.Context, symbol, contractType strin
 	if symbol != "" {
 		params.Set("symbol", symbol)
 	}
-	if contractType != "" {
-		if !common.StringSliceCompareInsensitive(validContractTypes, contractType) {
-			return resp, errors.New("invalid contractType")
+	if t := strings.ToLower(contractType); t != "" {
+		if _, ok := contractExpiryNames[t]; !ok {
+			return resp, fmt.Errorf("%w: %v", errInvalidContractType, t)
 		}
-		params.Set("contract_type", contractType)
+		params.Set("contract_type", t)
 	}
 	if !code.IsEmpty() {
 		codeValue, err := h.FormatSymbol(code, asset.Futures)
@@ -129,11 +132,11 @@ func (h *HUOBI) FContractPriceLimitations(ctx context.Context, symbol, contractT
 	if symbol != "" {
 		params.Set("symbol", symbol)
 	}
-	if contractType != "" {
-		if !common.StringSliceCompareInsensitive(validContractTypes, contractType) {
-			return resp, fmt.Errorf("invalid contractType: %s", contractType)
+	if t := strings.ToLower(contractType); t != "" {
+		if _, ok := contractExpiryNames[t]; !ok {
+			return resp, fmt.Errorf("%w: %v", errInvalidContractType, t)
 		}
-		params.Set("contract_type", contractType)
+		params.Set("contract_type", t)
 	}
 	if !code.IsEmpty() {
 		codeValue, err := h.FormatSymbol(code, asset.Futures)
@@ -163,11 +166,11 @@ func (h *HUOBI) ContractOpenInterestUSDT(ctx context.Context, contractCode, pair
 		}
 		params.Set("pair", p)
 	}
-	if contractType != "" {
-		if !common.StringSliceCompareInsensitive(validContractTypes, contractType) {
-			return nil, errors.New("invalid contractType")
+	if t := strings.ToLower(contractType); t != "" {
+		if _, ok := contractExpiryNames[t]; !ok {
+			return nil, fmt.Errorf("%w: %v", errInvalidContractType, t)
 		}
-		params.Set("contract_type", contractType)
+		params.Set("contract_type", t)
 	}
 	if businessType != "" {
 		params.Set("business_type", businessType)
@@ -186,11 +189,11 @@ func (h *HUOBI) FContractOpenInterest(ctx context.Context, symbol, contractType 
 	if symbol != "" {
 		params.Set("symbol", symbol)
 	}
-	if contractType != "" {
-		if !common.StringSliceCompareInsensitive(validContractTypes, contractType) {
-			return resp, errors.New("invalid contractType")
+	if t := strings.ToLower(contractType); t != "" {
+		if _, ok := contractExpiryNames[t]; !ok {
+			return resp, fmt.Errorf("%w: %v", errInvalidContractType, t)
 		}
-		params.Set("contract_type", contractType)
+		params.Set("contract_type", t)
 	}
 	if !code.IsEmpty() {
 		codeValue, err := h.formatFuturesPair(code, true)
@@ -375,8 +378,9 @@ func (h *HUOBI) FQueryHisOpenInterest(ctx context.Context, symbol, contractType,
 	if symbol != "" {
 		params.Set("symbol", symbol)
 	}
-	if !common.StringSliceCompareInsensitive(validContractTypes, contractType) {
-		return resp, fmt.Errorf("%w %v", errInvalidContractType, contractType)
+	contractType = strings.ToLower(contractType)
+	if _, ok := contractExpiryNames[contractType]; !ok {
+		return resp, fmt.Errorf("%w: %v", errInvalidContractType, contractType)
 	}
 	params.Set("contract_type", contractType)
 	if !common.StringSliceCompareInsensitive(validPeriods, period) {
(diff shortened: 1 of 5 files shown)


exchanges/huobi/huobi_wrapper.go
```diff
@@ -288,16 +289,30 @@ func (h *HUOBI) FetchTradablePairs(ctx context.Context, a asset.Item) (currency.
...
 		pairs = make([]currency.Pair, 0, len(symbols.Data))
-		for c := range symbols.Data {
-			if symbols.Data[c].ContractStatus != 1 {
+		expiryCodeDates := map[string]currency.Code{}
+		for _, c := range symbols.Data {
+			if c.ContractStatus != 1 {
 				continue
 			}
-			pair, err := currency.NewPairFromString(symbols.Data[c].ContractCode)
+			pair, err := currency.NewPairFromString(c.ContractCode)
```
gloriousCode, line 297:
```suggestion
			pair, err = currency.NewPairFromString(c.ContractCode)
```

exchanges/huobi/huobi_wrapper.go
```diff
@@ -388,73 +404,73 @@ func (h *HUOBI) UpdateTickers(ctx context.Context, a asset.Item) error {
...
+			var err error
+			if ticks[i].Symbol != "" {
+				cp, err = currency.NewPairFromString(ticks[i].Symbol)
+				if err == nil {
+					cp, err = h.pairFromContractExpiryCode(cp)
 				}
-				cp, err = h.convertContractShortHandToExpiry(cp, time.Now())
-				if err != nil {
-					return err
+				if err == nil {
```
gloriousCode, line 431: I like this `err == nil`  less when you're inspecting for errors for one of three function calls below. It becomes more ambiguous about what you're checking for and when. Its not saving many line changes and now you're nesting code that doesn't really need to be further nested

exchanges/huobi/huobi_futures.go
```diff
@@ -90,11 +93,11 @@ func (h *HUOBI) FGetContractInfo(ctx context.Context, symbol, contractType strin
 	if symbol != "" {
 		params.Set("symbol", symbol)
 	}
-	if contractType != "" {
-		if !common.StringSliceCompareInsensitive(validContractTypes, contractType) {
-			return resp, errors.New("invalid contractType")
+	if t := strings.ToLower(contractType); t != "" {
```
gloriousCode, line 96: You'd make an empty string lowercase to avoid an extra line of code 😄 no change required

exchanges/huobi/huobi_test.go
```diff
@@ -2870,74 +2851,52 @@ func TestUpdateTickers(t *testing.T) {
...
-	cp, err = h.convertContractShortHandToExpiry(cp, tt)
-	assert.NoError(t, err)
-	assert.Equal(t, "210604", cp.Quote.String())
+	h := new(HUOBI) //nolint:govet // Intentional shadow to avoid future copy/paste mistakes
+	require.NoError(t, testexch.Setup(h), "Test Instance Setup must not fail")
 
-	cp = currency.NewPair(currency.BTC, currency.NewCode("CWif hat"))
-	_, err = h.convertContractShortHandToExpiry(cp, tt)
-	assert.ErrorIs(t, err, errInvalidContractType)
+	_, err := h.FetchTradablePairs(context.Background(), asset.Futures)
```
gloriousCode, line 2870: Since `UpdateTradablePairs` is called in `TestMain` you could assign it to your shadow h, or use your helper `UpdatePairsOnce`

exchanges/huobi/huobi_test.go
```diff
@@ -2870,74 +2851,52 @@ func TestUpdateTickers(t *testing.T) {
...
+		assert.Equal(t, currency.BTC, p.Base, "pair Base should be the same")
+		assert.Equal(t, exp, p.Quote, "pair Quote should be the same")
+		d, err := time.Parse("060102", p.Quote.String())
+		require.NoError(t, err, "currency code must be a parsable date")
+		require.Truef(t, d.After(n), "%s expiry must be today or after", cType)
+		switch cType {
+		case "CW", "NW":
+			require.True(t, d.Before(n.Add(24*time.Hour*14)), "%s expiry must be within 2 weeks", cType)
+		case "CQ", "NQ":
+			require.True(t, d.Before(n.Add(24*time.Hour*90*2)), "%s expiry must be within 2 months", cType)
```
gloriousCode, line 2897: Doesnt sound right for Next Quarter if its ever present. It would likely be further out. But then I also take issue with 180 days being within 2 months

gloriousCode (changes requested):
Thanks for looking into and resolving the issue. Only minor changes

exchanges/huobi/huobi_futures.go
```diff
@@ -1273,42 +1277,16 @@ func (h *HUOBI) formatFuturesPair(p currency.Pair, convertQuoteToExpiry bool) (s
 	return h.FormatSymbol(p, asset.Futures)
 }
 
-// convertContractShortHandToExpiry converts a contract shorthand eg BTC-CW into a full expiry date
-// eg BTC240329 to associate with tradable pair formatting
-func (h *HUOBI) convertContractShortHandToExpiry(pair currency.Pair, tt time.Time) (currency.Pair, error) {
```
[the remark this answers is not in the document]
gbjk: Happy to restore that. I must be missing how it worked though.
It's going to take one of CW, CQ, NW, NQ, and convert it.
For an expired contract ... what would that look like?
Can you give me a test case and I'll restore that functionality if possible.

exchanges/huobi/huobi_wrapper.go
gbjk: Imma do the opposite, hope that's okay.
As you know, I always try to keep var scope as small as possible.

Fixed 85442e6a

gbjk:
Seen in the wild today:
```
    huobi_test.go:2892:
                Error Trace:    /Users/gbjk/go/src/github.com/gbjk/gocryptotrader/exchanges/huobi/huobi_test.go:2892
                Error:          Should be true
                Test:           TestPairFromContractExpiryCode
                Messages:       CW expiry must be today or after
```

gbjk:
That's fixed. Was testing the date was `After` instead of `not Before`

exchanges/huobi/huobi_futures.go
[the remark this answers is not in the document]
gbjk: Now I get it. You'd pass it an old expiry time to get what you wanted.
In which case shall we keep both functions in the master branch maybe?

exchanges/huobi/huobi_test.go
gbjk: Could do. However for this test-case I wanted to avoid any indirectness, because we expect `FetchTradablePairs` to be doing the thing itself, and now.

exchanges/huobi/huobi_test.go
gbjk: Agreed. 2 quarters.

Fixed 8110c76d

exchanges/huobi/huobi_wrapper.go
gbjk: So you're asking for:
```
M exchanges/huobi/huobi_wrapper.go
@@ -423,14 +423,15 @@ func (h *HUOBI) UpdateTickers(ctx context.Context, a asset.Item) error {
 			var cp currency.Pair
 			var err error
 			if ticks[i].Symbol != "" {
-				cp, err = currency.NewPairFromString(ticks[i].Symbol)
-				if err == nil {
-					cp, err = h.pairFromContractExpiryCode(cp)
+				if cp, err = currency.NewPairFromString(ticks[i].Symbol); err != nil {
+					errs = common.AppendError(errs, err)
+					continue
 				}
-				if err == nil {
-					cp, _, err = h.MatchSymbolCheckEnabled(cp.String(), a, true)
+				if cp, err = h.pairFromContractExpiryCode(cp); err != nil {
+					errs = common.AppendError(errs, err)
+					continue
 				}
-				if err != nil {
+				if cp, _, err = h.MatchSymbolCheckEnabled(cp.String(), a, true); err != nil {
 					if !errors.Is(err, currency.ErrPairNotFound) {
 					}
```
That would look like 837fd824

See, I think personally I'd double down even further on this, saving the lines in the else statement, saving another 5 lines.
My thinking on this is pretty simple:
If the issue is that we don't find the pair, then we skip the error and the tick.
It really doesn't matter where that came from.

So I'd prefer this: a1cf27f9

I've left it with the later in the branch. Shout if it's still not acceptable, obviously.

exchanges/huobi/huobi_futures.go
gbjk: Note: If it was a hotter path I wouldn't do this.
I sniffed the cost, and it's not awful. It should be outta there ns.

exchanges/huobi/huobi_futures.go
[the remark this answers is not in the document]
gbjk: Nope. We're good. Resolving.

gloriousCode (approved):
tACK! tyty

exchanges/huobi/huobi_test.go
```diff
@@ -2674,49 +2671,33 @@ func TestGetAvailableTransferChains(t *testing.T) {
...
-	}
+	require.NoError(t, err)
+	assert.Equal(t, "BTC_CW", r)
+
+	p, err := h.FetchTradablePairs(context.Background(), asset.Futures)
+	require.NoError(t, err, "FetchTradablePairs must not error")
+	require.NotEmpty(t, p, "FetchTradablePairs must return pairs")
+
+	// test getting a tradable pair in the format of BTC210827 but make it lower case to test correct formatting
+	r, err = h.formatFuturesPair(p[0], false)
```
shazbert, line 2685:
```suggestion
	r, err = h.formatFuturesPair(p[0].Lower(), false)
```
As per commentary?

shazbert (changes requested):
LGTM. One minor thing.

exchanges/huobi/huobi_test.go
gbjk: Yoink. Thanks 🙏 

Fixed e8410a83

shazbert (approved):
Thanks good stuff. 🙏

thrasher- (approved):
tACK, all working well. Thanks for looking into this and resolving!

PR #1603, opened 2024-08-06
exchange/websocket, gateio: Rename/export package again, add websocket request functions for futures trading

shazbert:
# PR Description

* Adds functions for trading using websocket connection. 
* Correctly routes outbound requests through available connection dependant on asset settlement
* Integrates an independant wrapper function for websocket outbound orders

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1670, opened 2024-10-09
bybit: enable multiconnection handling across websocket endpoints

shazbert:
# PR Description

* Expand websocket handling to concurrently allow processing for CoinMarginedFutures, USDTMarginedFutures, USDCMarginedFutures, Options. 
* Implement dedicated connection for private data.

NOTE: This is a nice size PR for just functionality expansion, I will build outbound request functions for trading on top of this in a future PR. Here's the dependency: #1603

Fixes # 

* 🐛 ~DOGEUSDT wasn't being processed correctly so used `MatchSymbolWithAvailablePairs` because there was no delimiter.~ This was merged upstream.

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/bybit/bybit_websocket.go
```diff
@@ -792,3 +740,162 @@ const subTplText = `
...
+		return currency.EMPTYPAIR, 0, fmt.Errorf("category '%s' not supported for incoming symbol '%s'", category, symbol)
+	}
+	var err error
+	for _, a := range assets {
+		var cp currency.Pair
+		if cp, err = by.MatchSymbolWithAvailablePairs(symbol, a, hasPotentialDelimiter(a)); err == nil {
+			return cp, a, nil
+		}
+	}
+	return currency.EMPTYPAIR, 0, err
```
gloriousCode, line 900:
```go
	for _, a := range assets {
		cp, err := by.MatchSymbolWithAvailablePairs(symbol, a, hasPotentialDelimiter(a))
		if err != nil {
			return currency.EMPTYPAIR, 0, err
		}
		return cp, a, nil
	}
	return currency.EMPTYPAIR, 0, nil
```
I think its an improvement to readability to not try funny business with err handling and such

exchanges/bybit/bybit_websocket.go
```diff
@@ -792,3 +740,162 @@ const subTplText = `
...
+// TODO: Remove this function when template expansion is across all assets
+func (by *Bybit) generateAuthSubscriptions() (subscription.List, error) {
+	if !by.Websocket.CanUseAuthenticatedEndpoints() {
+		return nil, nil
+	}
+	var subscriptions subscription.List
+	for _, channel := range []string{chanPositions, chanExecution, chanOrder, chanWallet} {
+		subscriptions = append(subscriptions, &subscription.Subscription{
+			Channel: channel,
+			Pairs:   currency.Pairs{currency.EMPTYPAIR}, // This is a placeholder, the actual pair is not required for these channels
```
gloriousCode, line 860: Is pairs required to be filled? Have you tried not sending pairs?

gloriousCode (changes requested):
Nice work shazbert, I'm quite happy with a lot of the work

exchanges/bybit/bybit_websocket.go
```diff
@@ -313,14 +240,51 @@ func (by *Bybit) wsHandleData(assetType asset.Item, respRaw []byte) error {
...
 	case chanPositions:
 		return by.wsProcessPosition(&result)
 	case chanExecution:
-		return by.wsProcessExecution(asset.Spot, &result)
+		return by.wsProcessExecution(&result)
 	case chanOrder:
-		return by.wsProcessOrder(asset.Spot, &result)
+		// Below provides a way of matching an order change to a websocket request. There is no batch support for this
+		// so the first element will be used to match the order ID.
+		if id, err := jsonparser.GetString(respRaw, "data", "[0]", "orderId"); err == nil {
```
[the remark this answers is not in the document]
shazbert: Nice catch this was actually updated to that in the request PR built on this. 🚀 I will roll that update here.

exchanges/bybit/bybit_websocket.go
```diff
@@ -792,3 +740,162 @@ const subTplText = `
...
+// TODO: Remove this function when template expansion is across all assets
+func (by *Bybit) handleSubscriptionNonTemplate(ctx context.Context, conn websocket.Connection, a asset.Item, operation string, channelsToSubscribe subscription.List) error {
+	payloads, err := by.handleSubscriptionsNonTemplate(conn, a, operation, channelsToSubscribe)
+	if err != nil {
+		return err
+	}
+	for _, payload := range payloads {
+		if a == asset.Options {
+			// The options connection does not send the subscription request id back with the subscription notification payload
+			// therefore the code doesn't wait for the response to check whether the subscription is successful or not.
```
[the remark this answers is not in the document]
shazbert: Going to be RM'd anyway

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -1178,7 +1178,7 @@ Please test all `pair` commands to disable and enable different assets types to
 - `get` to ensure correct enabled and disabled pairs for a supported asset type.
 - `disableasset` to ensure disabling of entire asset class and associated unsubscriptions.
 - `enableasset` to ensure correct enabling of entire asset class and associated subscriptions.
-- `disable` to ensure correct disabling of pair(s) and and associated unsubscriptions.
+- `disable` to ensure correct disabling of pair(s) and associated unsubscriptions.
```
gbjk: 🐒 This monkey thinks `unsubscription` isn't a real word, and "disabling X and associated subscriptions" is the same as "disabling [...] subscriptions"
```suggestion
- `disable` to ensure correct disabling of pair(s) and associated subscriptions.
```

exchanges/gateio/gateio.go
```diff
@@ -187,8 +187,9 @@ func timeInForceFromString(tif string) (order.TimeInForce, error) {
 
 // Gateio is the overarching type across this package
 type Gateio struct {
-	Counter common.Counter // Must be first	due to alignment requirements
 	exchange.Base
+
+	messageIDSeq  common.Counter
```
gbjk, line 193: 🐒 For GateIO, can't we just use uuid.UUID V7. Particularly [google's implementation](https://pkg.go.dev/github.com/google/uuid#NewV7)

exchange/websocket/connection.go
```diff
@@ -111,23 +111,23 @@ type Response struct {
...
+	URL                  string
+	ProxyURL             string
+	Wg                   *sync.WaitGroup
+	Connection           *gws.Conn
+	shutdown             chan struct{}
+	Match                *Match
+	ResponseMaxLimit     time.Duration
+	Traffic              chan struct{}
+	readMessageErrors    chan error
+	requestIDGenerator   func() int64
```
gbjk, line 131: 🐒  I think if we're touching this, we should change to string.
Most sane APIs would support a uuid, and uuid v7 is sequential. `uuid.Must(uuid.NewV7)`
```suggestion
	requestIDGenerator   func() string
```
Optionally I'd guess to allow errors. I don't have a generator I'd want to use that'd return an error, though.

exchanges/bybit/bybit.go
```diff
@@ -27,7 +27,9 @@ import (
 // Bybit is the overarching type across this package
 type Bybit struct {
 	exchange.Base
-	account accountTypeHolder
+
+	messageIDSeq common.Counter
```
gbjk, line 33: 🐒 For Bybit, can't we just use uuid.UUID V7. Particularly [google's implementation (https://pkg.go.dev/github.com/google/uuid#NewV7)
Docs say it's just a string.

exchanges/bybit/bybit_websocket.go
```diff
@@ -799,3 +747,164 @@ const subTplText = `
 func hasPotentialDelimiter(a asset.Item) bool {
 	return a == asset.Options || a == asset.USDCMarginedFutures
 }
+
+// TODO: Remove this function when template expansion is across all assets
+func (by *Bybit) submitSubscriptionNonTemplate(ctx context.Context, conn websocket.Connection, a asset.Item, operation string, channelsToSubscribe subscription.List) error {
```
gbjk: 🐒 I'm not wild about needing to do this, but it'll get us further faster we just take this step first.
However even in the short term, I think `NonTemplate` is better as `NoTemplate` or `WithoutTemplate`.
But I'd argue for just `submitSimpleSubscription` or `submitDirectSubscription`

exchanges/bybit/bybit_websocket.go
```diff
@@ -799,3 +747,164 @@ const subTplText = `
...
+			}
+		}
+		if operation == "unsubscribe" {
+			err = by.Websocket.RemoveSubscriptions(conn, payload.associatedSubs...)
+		} else {
+			err = by.Websocket.AddSubscriptions(conn, payload.associatedSubs...)
+		}
+		if err != nil {
+			return err
+		}
```
gbjk: 🦍 Signatures are the same, so optionally could use this pattern.
`op` can move out of the for loop, obviously.
```suggestion
		op := by.Websocket.AddSubscriptions
		if operation == "unsubscribe" {
			op = by.Websocket.RemoveSubscriptions
		}
		if err := op(conn, payload.associatedSubs...); err != nil {
			return err
		}
```

exchanges/bybit/bybit_websocket.go
```diff
@@ -799,3 +747,164 @@ const subTplText = `
...
+	}
+	return nil
+}
+
+// TODO: Remove this function when template expansion is across all assets
+func (by *Bybit) handleSubscriptionsNonTemplate(conn websocket.Connection, assetType asset.Item, operation string, channelsToSubscribe subscription.List) ([]SubscriptionArgument, error) {
```
gbjk: 🐒 This isn't actually handling anything.
```suggestion
func (by *Bybit) directSubscriptionPayload(conn websocket.Connection, assetType asset.Item, operation string, channelsToSubscribe subscription.List) ([]SubscriptionArgument, error) {
```

exchanges/bybit/bybit_websocket.go
```diff
@@ -799,3 +747,164 @@ const subTplText = `
...
+		var pair currency.Pair
+		if len(s.Pairs) > 1 {
+			return nil, subscription.ErrNotSinglePair
+		}
+		if len(s.Pairs) == 1 {
+			pair = s.Pairs[0]
+		}
+		switch s.Channel {
+		case chanOrderbook:
+			arg.Arguments = append(arg.Arguments, fmt.Sprintf("%s.%d.%s", s.Channel, 50, pair.Format(pairFormat).String()))
```
gbjk: 🐒 pairFormat.Format appears to do this in one go.
Also think get can get away with pairFmt
🌴 This monkey lives in multiple trees below too.
```suggestion
			arg.Arguments = append(arg.Arguments, fmt.Sprintf("%s.%d.%s", s.Channel, 50, pairFmt.Format(pair)))
```

exchanges/bybit/bybit_test.go
```diff
@@ -3211,21 +3188,36 @@ var pushDataMap = map[string]string{
...
+	"Private Greek":    `{ "id": "592324fa945a30-2603-49a5-b865-21668c29f2a6", "topic": "greeks", "creationTime": 1672364262482, "data": [ { "baseCoin": "ETH", "totalDelta": "0.06999986", "totalGamma": "-0.00000001", "totalVega": "-0.00000024", "totalTheta": "0.00001314" } ] }`,
+	"Execution":        `{"id": "592324803b2785-26fa-4214-9963-bdd4727f07be", "topic": "execution", "creationTime": 1672364174455, "data": [ { "category": "linear", "symbol": "XRPUSDT", "execFee": "0.005061", "execId": "7e2ae69c-4edf-5800-a352-893d52b446aa", "execPrice": "0.3374", "execQty": "25", "execType": "Trade", "execValue": "8.435", "isMaker": false, "feeRate": "0.0006", "tradeIv": "", "markIv": "", "blockTradeId": "", "markPrice": "0.3391", "indexPrice": "", "underlyingPrice": "", "leavesQty": "0", "orderId": "f6e324ff-99c2-4e89-9739-3086e47f9381", "orderLinkId": "", "orderPrice": "0.3207", "orderQty":"25","orderType":"Market","stopOrderType":"UNKNOWN","side":"Sell","execTime":"1672364174443","isLeverage": "0","closedSize": "","seq":4688002127}]}`,
```
gbjk: 🐒   Can this not use the FixtureToDataHandler pattern?
We're already using it in bybit_test

exchanges/bybit/bybit_test.go
```diff
@@ -3788,46 +3780,47 @@ func TestGenerateSubscriptions(t *testing.T) {
...
-	b := new(Bybit)
-	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+
+	b := new(Bybit) //nolint:govet // Intentional shadow to avoid future copy/paste mistakes
+	require.NoError(t, testexch.Setup(b))
+	require.NoError(t, b.authSubscribe(t.Context(), &DummyConnection{}, subscription.List{}))
+
+	authsubs, err := b.generateAuthSubscriptions()
+	require.NoError(t, err)
+	require.Empty(t, authsubs)
```
gbjk: 🐒 🌴 
```suggestion
	require.NoError(t, err, "generateAuthSubscriptions must not error")
	require.Empty(t, authsubs, "generateAuthSubscriptions must not return subs")
```

exchanges/bybit/bybit_test.go
```diff
@@ -3788,46 +3780,47 @@ func TestGenerateSubscriptions(t *testing.T) {
...
-	b = testexch.MockWsInstance[Bybit](t, testws.CurryWsMockUpgrader(t, mock))
-	b.Websocket.AuthConn = b.Websocket.Conn
-	err = b.Subscribe(subs)
-	require.NoError(t, err, "Subscribe must not error")
-	success = false
-	err = b.Subscribe(subs)
-	assert.ErrorContains(t, err, "Mock Resp Error", "Subscribe should error containing the returned RetMsg")
+	authsubs, err = b.generateAuthSubscriptions()
+	require.NoError(t, err)
+	require.NotEmpty(t, authsubs)
```
gbjk: And throughout.
```suggestion
	require.NoError(t, err, "generateAuthSubscriptions must not error")
	require.Empty(t, authsubs, "generateAuthSubscriptions must not return subs")
```

exchanges/bybit/bybit_test.go
```diff
@@ -3893,3 +3886,89 @@ func TestTransformSymbol(t *testing.T) {
...
+	require.Len(t, o, 1)
+	require.Equal(t, "c1956690-b731-4191-97c0-94b00422231b", o[0].OrderID)
+	require.Equal(t, "BTC_USDT", o[0].Pair.String())
+	require.Equal(t, order.Sell, o[0].Side)
+	require.Equal(t, order.Filled, o[0].Status)
+	require.Equal(t, 1.7, o[0].Amount)
+	require.Equal(t, 4.033, o[0].Price)
+	require.Equal(t, 4.24, o[0].AverageExecutedPrice)
+	require.Equal(t, 0., o[0].RemainingAmount)
+	require.Equal(t, asset.USDTMarginedFutures, o[0].AssetType)
```
gbjk: 🐒  Also here; Can this not use the FixtureToDataHandler pattern?

gbjk (approved):
Good work 👍 

I'm generally unhappy with needing to do any of this, but I can see the constraints, and if we spend more time hacking out anything better right now we might as well spend that time on the already open branches.

I've downgraded everything to nitpicks from change requests because this is high priority, and you can fix them later. None of them introduce bugs or regressions.

exchange/websocket/connection.go
shazbert: Good suggestion, I think I might make a separate PR for this, because this will impact GenerateMessageID signature as well, I would probably opt for `defaultGenerateMessageID` to be v7 for searchability and `ClientOrderID` field setting. There is a tiny overhead associated with v7 but it's minute, so keeping it as a baseline would be perfect. The current default is just using `rand.Int` and has the potential to have collisions anyway. 

For GateIO/bybit the counter is used mainly for request response matching and the counters don't seem to be tracked or are reset on each new connection and they don't come back in fills/order history. So if possible would like to keep them as is and then refresh them if needed in that separate PR?

Thoughts? @thrasher- @gbjk

exchanges/bybit/bybit_test.go
shazbert: [removed and consolidated in](https://github.com/thrasher-corp/gocryptotrader/pull/1670/commits/49930d2190a74540225f1d89737ac66039fa6cc7)

exchanges/bybit/bybit_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1670/commits/49930d2190a74540225f1d89737ac66039fa6cc7

shazbert:
> I'm generally unhappy with needing to do any of this

image

exchange/websocket/connection.go
gbjk: Agreed to slide this to later.

exchanges/gateio/gateio.go
gbjk: Agreed to slide this to later following [this reply](https://github.com/thrasher-corp/gocryptotrader/pull/1670#discussion_r2149088220)

exchanges/bybit/bybit_test.go
```diff
@@ -3060,6 +3036,172 @@ func TestPushData(t *testing.T) {
...
+
+	b := new(Bybit) //nolint:govet // Intentional shadow to avoid future copy/paste mistakes
+	require.NoError(t, testexch.Setup(b), "Test instance Setup must not error")
+	b.API.AuthenticatedSupport = true
+	b.API.AuthenticatedWebsocketSupport = true
+	b.SetCredentials("test", "test", "", "", "", "")
+	testexch.FixtureToDataHandler(t, "testdata/wsAuth.json", func(ctx context.Context, r []byte) error {
+		if bytes.Contains(r, []byte("%s")) {
+			r = fmt.Appendf(nil, string(r), optionsTradablePair.String())
+		}
```
gbjk, line 3103: 🥃 🗻  This feels like a generic problem.
Wondering if we can text/template the content easily, and introduce a middleware:
```
FixtureToDataHandler(t, "foo.json", testexch.WrapFixture(b.wsHandleAuthenticatedData), testexch.FixtureVars{
    "optionsPair": optionsTradablePair,
}
```
and then in the fixture json:
```
{{ .vars.optionsPair.String }}
```

exchanges/bybit/bybit_test.go
```diff
@@ -3060,6 +3036,172 @@ func TestPushData(t *testing.T) {
...
+		return b.wsHandleAuthenticatedData(ctx, r)
+	})
+	close(b.Websocket.DataHandler)
+	require.Len(t, b.Websocket.DataHandler, 6, "Should see correct number of private messages")
```
gbjk: 🐒 
```suggestion
	require.Len(t, b.Websocket.DataHandler, 6, "Should see correct number of messages")
```

exchanges/bybit/bybit_test.go
```diff
@@ -3060,6 +3036,172 @@ func TestPushData(t *testing.T) {
...
+
+	i := 0
+	for data := range b.Websocket.DataHandler {
+		i++
+		switch v := data.(type) {
+		case WsPositions:
+			require.Len(t, v, 1, "must see 1 position")
+			assert.Zero(t, v[0].PositionIdx, "PositionIdx should be 0")
+			assert.Zero(t, v[0].TradeMode, "TradeMode should be 0")
+			assert.Equal(t, int64(41), v[0].RiskID, "RiskID should match")
```
gbjk: 🦍  The wording I've used so far is "X should be correct".
"match" feels overloaded, because of times we actually mean that some function or method called match returned true, etc.
Don't have to action, I'm just being pedantic/consistent.

exchanges/bybit/bybit_test.go
```diff
@@ -3060,6 +3036,172 @@ func TestPushData(t *testing.T) {
...
+			assert.Equal(t, asset.USDTMarginedFutures, v[0].AssetType, "Asset type should match")
+			assert.Equal(t, "XRP_USDT", v[0].CurrencyPair.String(), "Symbol should match")
+			assert.Equal(t, order.Sell, v[0].Side, "Side should match")
+			assert.Equal(t, "f6e324ff-99c2-4e89-9739-3086e47f9381", v[0].OrderID, "Order ID should match")
+			assert.Empty(t, v[0].ClientOrderID, "Client order ID should be empty")
+			assert.Empty(t, v[0].TradeID, "Trade ID should be empty")
+			assert.Equal(t, 0.3374, v[0].Price, "price should match")
+			assert.Equal(t, 25.0, v[0].Amount, "amount should match")
+		default:
+			t.Errorf("Unexpected data type received: %T", v)
```
gbjk: 🐒 
```suggestion
			t.Errorf("Unexpected data received: %v", v)
```

gbjk (approved):
All good.
Only wildlife to see. 🐒 🦍

exchanges/bybit/bybit_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1670/commits/e6f4a93426fa1da4a321b9342aa7f2e1d3de15c3

exchanges/bybit/bybit_test.go
shazbert: Yeah that would be cool to do, can do this in another PR.

gbjk (approved):

exchanges/bybit/bybit_wrapper.go
```diff
@@ -214,65 +209,137 @@ func (by *Bybit) SetDefaults() {
...
 		return nil
 	}
 
-	err = by.SetupDefaults(exch)
-	if err != nil {
+	if err := by.SetupDefaults(exch); err != nil {
 		return err
 	}
 
-	wsRunningEndpoint, err := by.API.Endpoints.GetURL(exchange.WebsocketSpot)
```
cranktakular: It wasn't fully utilised before, but wouldn't we still want to retrieve the URL this way, instead of directly assigning it to a constant? Especially since some users will need to use non-standard URLs ("bybit-tr.com" or "bybit.kz" instead of "bybit.com")

And if we don't want to retrieve it this way, would we want to remove that functionality across the entire exchange?

exchanges/bybit/bybit_wrapper.go
```diff
@@ -214,65 +209,137 @@ func (by *Bybit) SetDefaults() {
...
-		})
-	if err != nil {
+	// Spot
+	if err := by.Websocket.SetupNewConnection(&websocket.ConnectionSetup{
+		URL:                   spotPublic,
+		ResponseCheckTimeout:  exch.WebsocketResponseCheckTimeout,
+		ResponseMaxLimit:      exch.WebsocketResponseMaxLimit,
+		RateLimit:             request.NewWeightedRateLimitByDuration(time.Microsecond),
+		Connector:             by.WsConnect,
+		GenerateSubscriptions: func() (subscription.List, error) { return by.generateSubscriptions() },
```
cranktakular: Any reason why this can't just be `by.generateSubscriptions`?

exchanges/bybit/bybit_websocket.go
```diff
@@ -63,94 +65,56 @@ var defaultSubscriptions = subscription.List{
...
-	subscription.MyTradesChannel:  chanExecution,
 	subscription.MyWalletChannel:  chanWallet,
+	subscription.MyTradesChannel:  chanExecution,
 	subscription.CandlesChannel:   chanKline,
 }
 
 // WsConnect connects to a websocket feed
-func (by *Bybit) WsConnect() error {
-	ctx := context.TODO()
-	if !by.Websocket.IsEnabled() || !by.IsEnabled() || !by.IsAssetWebsocketSupported(asset.Spot) {
```
cranktakular: Is it truly wise to remove these enabled checks?

exchanges/bybit/bybit_websocket.go
```diff
@@ -176,60 +139,41 @@ func (by *Bybit) handleSubscriptions(operation string, subs subscription.List) (
...
-	ctx := context.TODO()
-	return by.handleSpotSubscription(ctx, "unsubscribe", channelsToUnsubscribe)
+func (by *Bybit) Unsubscribe(ctx context.Context, conn websocket.Connection, channelsToUnsubscribe subscription.List) error {
+	return by.handleSpotSubscription(ctx, conn, "unsubscribe", channelsToUnsubscribe)
 }
 
-func (by *Bybit) handleSpotSubscription(ctx context.Context, operation string, channelsToSubscribe subscription.List) error {
-	payloads, err := by.handleSubscriptions(operation, channelsToSubscribe)
+func (by *Bybit) handleSpotSubscription(ctx context.Context, conn websocket.Connection, operation string, channelsToSubscribe subscription.List) error {
+	payloads, err := by.handleSubscriptions(conn, operation, channelsToSubscribe)
```
cranktakular: Are these appropriate function names? If handleSubscriptions() is only called through handleSpotSubscription(), then wouldn't either both or neither operate exclusively on Spot endpoints?

exchanges/bybit/bybit_websocket.go
```diff
@@ -569,47 +531,35 @@ func (by *Bybit) wsProcessKline(assetType asset.Item, resp *WebsocketResponse, t
...
-	if resp.Type == "snapshot" {
-		tick = &ticker.Price{
-			Pair:         p,
-			ExchangeName: by.Name,
-			AssetType:    assetType,
-		}
-	} else {
-		// ticker updates may be partial, so we need to update the current ticker
-		tick, err = ticker.GetTicker(by.Name, p, assetType)
+	tick := &ticker.Price{Pair: p, ExchangeName: by.Name, AssetType: assetType}
```
cranktakular: I prefer the old control flow. Always retrieving and setting this data, even when it would be overwritten later, doesn't seem worthwhile.

But a benchmark could easily prove that hunch wrong.

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +743,165 @@ const subTplText = `
...
+	}
+	return subscriptions, nil
+}
+
+// LinearSubscribe sends a subscription message to linear public channels.
+func (by *Bybit) authSubscribe(ctx context.Context, conn websocket.Connection, channelSubscriptions subscription.List) error {
+	return by.submitDirectSubscription(ctx, conn, asset.Spot, "subscribe", channelSubscriptions)
+}
+
+// LinearUnsubscribe sends an unsubscription messages through linear public channels.
```
cranktakular: The function names in these two subscribe/unsubscribe comments are incorrect

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +743,165 @@ const subTplText = `
...
+			if err != nil {
+				return nil, err
+			}
+			arg.Arguments = append(arg.Arguments, s.Channel+"."+interval+"."+pairFmt.Format(pair))
+			arg.associatedSubs = append(arg.associatedSubs, s)
+		case chanPositions, chanExecution, chanOrder, chanWallet, chanGreeks, chanDCP:
+			if chanMap[s.Channel] {
+				continue
+			}
+			authArg.Arguments = append(authArg.Arguments, s.Channel)
```
cranktakular, line 750: Is DCP fine with a subscription being performed without the asset? [The documentation](https://bybit-exchange.github.io/docs/v5/websocket/private/dcp) kind of implies it won't, but I can't test it myself.

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +743,165 @@ const subTplText = `
...
+			// adding the channel to selected channels so that we will not visit it again.
+			chanMap[s.Channel] = true
+			authArg.associatedSubs = append(authArg.associatedSubs, s)
+		}
+
+		if len(arg.Arguments) >= 10 {
```
cranktakular, line 756: This limit [appears to be for spot only](https://bybit-exchange.github.io/docs/v5/ws/connect#public-channel---args-limits). You may not want to bother handling that right now since this will be supplanted by templating stuff anyway, but there will be some efficiency improvements for options/futures possible.

exchanges/bybit/bybit_inverse_websocket.go
```diff
@@ -2,84 +2,43 @@ package bybit
...
 
 // InverseSubscribe sends a subscription message to linear public channels.
-func (by *Bybit) InverseSubscribe(channelSubscriptions subscription.List) error {
-	ctx := context.TODO()
-	return by.handleInversePayloadSubscription(ctx, "subscribe", channelSubscriptions)
+func (by *Bybit) InverseSubscribe(ctx context.Context, conn websocket.Connection, channelSubscriptions subscription.List) error {
+	return by.submitDirectSubscription(ctx, conn, asset.CoinMarginedFutures, "subscribe", channelSubscriptions)
 }
 
 // InverseUnsubscribe sends an unsubscription messages through linear public channels.
```
cranktakular: It seems odd for these comments to say they're sending messages through *linear* public channels.

exchanges/bybit/bybit_linear_websocket.go
```diff
@@ -2,102 +2,43 @@ package bybit
...
-		// The options connection does not send the subscription request id back with the subscription notification payload
-		// therefore the code doesn't wait for the response to check whether the subscription is successful or not.
-		err = by.Websocket.Conn.SendJSONMessage(ctx, request.Unset, payloads[a])
-		if err != nil {
-			return err
-		}
-	}
-	return nil
+func (by *Bybit) LinearUnsubscribe(ctx context.Context, conn websocket.Connection, channelSubscriptions subscription.List) error {
+	return by.submitDirectSubscription(ctx, conn, asset.USDTMarginedFutures, "unsubscribe", channelSubscriptions)
```
cranktakular: This mandates the use of USDT, while Linear on this exchange could also be USDC. Could you find a way to properly accommodate both here?

exchanges/bybit/testdata/wsAuth.json
```diff
@@ -0,0 +1,6 @@
+{"id": "59232430b58efe-5fc5-4470-9337-4ce293b68edd", "topic": "position", "creationTime": 1672364174455, "data": [ { "positionIdx": 0, "tradeMode": 0, "riskId": 41, "riskLimitValue": "200000", "symbol": "XRPUSDT", "side": "Buy", "size": "75", "entryPrice": "0.3615", "leverage": "10", "positionValue": "27.1125", "positionBalance": "0", "markPrice": "0.3374", "positionIM": "2.72589075", "positionMM": "0.28576575", "takeProfit": "0", "stopLoss": "0", "trailingStop": "0", "unrealisedPnl": "-1.8075", "cumRealisedPnl": "0.64782276", "createdTime": "1672121182216", "updatedTime": "1672364174449", "tpslMode": "Full", "liqPrice": "", "bustPrice": "", "category": "linear","positionStatus":"Normal","adlRankIndicator":2}]}
```
cranktakular, line 1: OOS but these are missing some fields that the exchange now sends.

exchanges/bybit/bybit_websocket.go
```diff
@@ -254,30 +198,9 @@ func (by *Bybit) GetSubscriptionTemplate(_ *subscription.Subscription) (*templat
...
-			err := by.wsHandleData(ctx, assetType, resp.Raw)
-			if err != nil {
-				by.Websocket.DataHandler <- err
-			}
-		}
-	}
-}
-
-func (by *Bybit) wsHandleData(ctx context.Context, assetType asset.Item, respRaw []byte) error {
+func (by *Bybit) wsHandleData(_ context.Context, assetType asset.Item, respRaw []byte) error {
```
cranktakular: You can just remove the context.Context from the function signature. Everything still compiles and runs fine.

cranktakular (changes requested):
Got some changes I want.

On top of those comments about specific places, some more general things:
* When running all exchange tests, I got an error in okx I didn't get on master (line 333, TestGetBlockTrade, "0 is not positive%!"). Find out whether that was caused by your changes, or if it's just intermittent.
* Potentially OOS, but consider writing detailed test functions to get more coverage out of GenerateInverseDefaultSubscriptions, InverseUnsubscribe, GenerateLinearDefaultSubscriptions, LinearUnsubscribe, GenerateOptionsDefaultSubscriptions, OptionUnsubscribe, Unsubscribe, and Setup.
* Potentially OOS, but I saw some room for more coverage in handleSubscriptions, handleSpotSubscription, wsHandleData, wsProcessWalletPushData, wsProcessOrder, wsProcessExecution, wsProcessLeverageTokenTicker, wsProcessPublicTicker, GetWalletBalance, and UpdateAccountInfo.
* I also saw some coverage shortfalls that I think are in scope. Within the functions wsHandleAuthenticatedData, wsProcessOrder, wsProcessPublicTicker, submitDirectSubscription, directSubscriptionPayload, getPairFromCategory, WsConnect, and WebsocketAuthenticatedConnection.

exchanges/bybit/bybit_wrapper.go
shazbert: nice catch!

exchanges/bybit/bybit_websocket.go
shazbert: Can defer this to https://github.com/thrasher-corp/gocryptotrader/pull/1672 if there are any issues when trading on it. But yeah it should be removed entirely as soon as possible.

exchanges/bybit/bybit_websocket.go
shazbert: Doesn't look like we handle DCP at all in `generateAuthSubscriptions`, added a TODO for implementation maybe for https://github.com/thrasher-corp/gocryptotrader/pull/1672 or when full template pass comes in.

exchanges/bybit/bybit_websocket.go
shazbert: I was assuming it would be used across templates when that upgrade occurs. Suggest a better name I could implement.

exchanges/bybit/bybit_websocket.go
shazbert: Have you found it connects with the asset turned off? Most of this is handled by the websocket manager.

exchanges/bybit/bybit_wrapper.go
shazbert: Good catch will re-implement

exchanges/bybit/bybit_websocket.go
shazbert: I might defer this, I think that @thrasher- has an update incoming for this on one of his PR's

exchanges/bybit/testdata/wsAuth.json
shazbert: Defer to #1672 I can't get concrete values from the position data example.

shazbert:
@cranktakular most nits addressed thank you!

shazbert:
@cranktakular Still have more things to do but time constrained hold off on reviewing this just yet until I can get around to it.

exchanges/bybit/bybit_websocket.go
cranktakular: Ah, if it's pending that sort of thing then don't worry.

shazbert:
- [x]  websocket reconnection is broken, need to investigate. Something to do with ping handlers not being cleaned up. 🪦

shazbert:
> * [ ]   websocket reconnection is broken, need to investigate. Something to do with ping handlers not being cleaned up. 🪦

nevermind that was bad code from me. 
image

shazbert:
@cranktakular 

> * When running all exchange tests, I got an error in okx I didn't get on master (line 333, TestGetBlockTrade, "0 is not positive%!"). Find out whether that was caused by your changes, or if it's just intermittent.

This PR doesn't impact that. 

> * Potentially OOS, but consider writing detailed test functions to get more coverage out of GenerateInverseDefaultSubscriptions, InverseUnsubscribe, GenerateLinearDefaultSubscriptions, LinearUnsubscribe, GenerateOptionsDefaultSubscriptions, OptionUnsubscribe, Unsubscribe, and Setup.

Added minor coverage improvements

> * Potentially OOS, but I saw some room for more coverage in handleSubscriptions, handleSpotSubscription, wsHandleData, wsProcessWalletPushData, wsProcessOrder, wsProcessExecution, wsProcessLeverageTokenTicker, wsProcessPublicTicker, GetWalletBalance, and UpdateAccountInfo.

Added minor coverage to this

> * I also saw some coverage shortfalls that I think are in scope. Within the functions wsHandleAuthenticatedData, wsProcessOrder, wsProcessPublicTicker, submitDirectSubscription, directSubscriptionPayload, getPairFromCategory, WsConnect, and WebsocketAuthenticatedConnection.

Added minor coverage for these, there are some which are going to be removed in template pass, `getPairFromCategory` will only error on nil currency manager so not going to hit that

exchanges/bybit/bybit_test.go
```diff
@@ -3015,42 +3019,34 @@ func TestCancelBatchOrders(t *testing.T) {
...
-	if mockTests {
-		t.Skip(skippingWebsocketFunctionsForMockTesting)
-	}
-	err := e.WsOptionsConnect()
-	assert.Truef(t, errors.Is(err, websocket.ErrWebsocketNotEnabled) || err == nil, "WsOptionsConnect should not error: %s", err)
+	err := e.WsConnect(t.Context(), &DummyConnection{dialError: nil})
+	require.NoError(t, err)
+	exp := errors.New("dial error")
+	err = e.WsConnect(t.Context(), &DummyConnection{dialError: exp})
+	require.ErrorIs(t, err, exp)
```
samuael, line 3054: Could you use a mere string comparison instead?

exchanges/bybit/spot_websocket.go
```diff
@@ -0,0 +1,50 @@
...
+	if err != nil {
+		return err
+	}
+	for _, payload := range payloads {
+		response, err := conn.SendMessageReturnResponse(ctx, request.Unset, payload.RequestID, payload)
+		if err != nil {
+			return err
+		}
+		var resp SubscriptionResponse
+		if err := json.Unmarshal(response, &resp); err != nil {
```
samuael, line 24: You can remove this immediate assignment since there is a big scoped err variable above.

samuael (approved):
Just two optional code suggestion spotted

exchanges/bybit/bybit_websocket.go
```diff
@@ -318,36 +187,68 @@ func (e *Exchange) wsHandleData(ctx context.Context, assetType asset.Item, respR
...
+func (e *Exchange) wsHandleAuthenticatedData(ctx context.Context, respRaw []byte) error {
+	var result WebsocketResponse
+	if err := json.Unmarshal(respRaw, &result); err != nil {
+		return err
+	}
+	if result.Topic == "" {
+		return e.handleNoTopicWebsocketResponse(&result, respRaw)
+	}
+	topicSplit := strings.Split(result.Topic, ".")
+	switch topicSplit[0] {
```
cranktakular, line 205: You could pull the stuff before this switch case into its own function, since it's shared with the non-auth version. Although this would add four lines, so might not be worthwhile, despite it de-duplicating code.

Actually, you wouldn't need to split off handleNoTopicWebsocketResponse that way, so it should ultimately save lines.

exchanges/bybit/bybit_websocket.go
```diff
@@ -318,36 +187,68 @@ func (e *Exchange) wsHandleData(ctx context.Context, assetType asset.Item, respR
...
 		return e.wsProcessPosition(&result)
 	case chanExecution:
-		return e.wsProcessExecution(asset.Spot, &result)
+		return e.wsProcessExecution(&result)
 	case chanOrder:
-		return e.wsProcessOrder(asset.Spot, &result)
+		// Below provides a way of matching an order change to a websocket request. There is no batch support for this
+		// so the first element will be used to match the order link ID.
+		if id, err := jsonparser.GetString(respRaw, "data", "[0]", "orderLinkId"); err == nil {
+			if e.Websocket.Match.IncomingWithData(id, respRaw) {
```
cranktakular: Might be misunderstanding part of how this works, but it seems conceptually weird to pass through the entirety of respRaw here, if it's only attempting to match with the first order change in that data.

Seems like, at best, the extra data would be superfluous, and at worst, it could taint things by associating those other orders with the first one, if handled improperly.

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +680,164 @@ const subTplText = `
...
+	case "spot":
+		assets = append(assets, asset.Spot)
+	case "inverse":
+		assets = append(assets, asset.CoinMarginedFutures)
+	case "linear":
+		assets = append(assets, asset.USDTMarginedFutures, asset.USDCMarginedFutures)
+	case "option":
+		assets = append(assets, asset.Options)
+	default:
+		return currency.EMPTYPAIR, 0, fmt.Errorf("category %q not supported for incoming symbol %q", category, symbol)
```
cranktakular: Bit of a nit, but this currently says the category is unsupported *for that symbol*, while it's simply unsupported *in general*.

Maybe replace with something like "incoming symbol %q has unsupported category %q"?

exchanges/bybit/bybit_websocket.go
```diff
@@ -569,47 +469,35 @@ func (e *Exchange) wsProcessKline(assetType asset.Item, resp *WebsocketResponse,
...
 	tick.LastUpdated = resp.PushTimestamp.Time()
-
-	if err = ticker.ProcessTicker(tick); err == nil {
-		e.Websocket.DataHandler <- tick
+	if err := ticker.ProcessTicker(tick); err != nil {
+		return err
 	}
-
+	e.Websocket.DataHandler <- tick
 	return err
```
cranktakular: This will always be nil regardless, this err was assigned by the return value of MatchSymbolWithAvailablePairs, which was already handled if not nil.

exchanges/bybit/bybit_websocket.go
```diff
@@ -569,47 +469,35 @@ func (e *Exchange) wsProcessKline(assetType asset.Item, resp *WebsocketResponse,
...
-			Pair:         p,
-			ExchangeName: e.Name,
-			AssetType:    assetType,
-		}
-	} else {
-		// ticker updates may be partial, so we need to update the current ticker
-		tick, err = ticker.GetTicker(e.Name, p, assetType)
+	tick := &ticker.Price{Pair: p, ExchangeName: e.Name, AssetType: assetType}
+	if resp.Type != "snapshot" {
+		snapshot, err := e.GetCachedTicker(p, assetType)
```
cranktakular: An err is already in scope and handled, why not change this to ticker, err = e.GetCachedTicker(p, assetType) so you don't need to re-assign?

exchanges/bybit/bybit_wrapper.go
```diff
@@ -214,65 +214,168 @@ func (e *Exchange) SetDefaults() {
 
 // Setup takes in the supplied exchange configuration details and sets params
 func (e *Exchange) Setup(exch *config.Exchange) error {
-	err := exch.Validate()
-	if err != nil {
+	if err := exch.Validate(); err != nil {
 		return err
 	}
+
```
cranktakular: Unnecessary added newline.

exchanges/bybit/inverse_websocket_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+	testexch "github.com/thrasher-corp/gocryptotrader/internal/testing/exchange"
+)
+
+func TestGenerateInverseDefaultSubscriptions(t *testing.T) {
+	t.Parallel()
+	e := new(Exchange) //nolint:govet // Intentional shadow
```
cranktakular, line 14: Weird that the declarations in this file are the only ones which need this comment. Seems like a bug with the linter, honestly.

exchanges/bybit/spot_websocket_test.go
```diff
@@ -0,0 +1,30 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	testexch "github.com/thrasher-corp/gocryptotrader/internal/testing/exchange"
+)
+
+func TestSpotSubscribe(t *testing.T) {
+	t.Parallel()
+	e := new(Exchange)
+	require.NoError(t, testexch.Setup(e), "Test instance Setup must not error")
```
cranktakular, line 13: Consider placing these two lines in their own function, like `e := newExchangeWithSetup(t)`, would save some lines across the tests in this package.

exchanges/bybit/bybit_test.go
```diff
@@ -3062,22 +3058,200 @@ var pushDataMap = map[string]string{
...
+			assert.Equal(t, "5cf98598-39a7-459e-97bf-76ca765ee020", v[0].OrderID, "Order ID should be correct")
+			assert.Equal(t, order.Sell, v[0].Side, "Side should be correct")
+			assert.Equal(t, order.Market, v[0].Type, "Order type should be correct")
+			assert.Equal(t, 72.5, v[0].Price, "Price should be correct")
+			assert.Equal(t, 1.0, v[0].Amount, "Amount should be correct")
+			assert.Equal(t, order.ImmediateOrCancel, v[0].TimeInForce, "Time in force should be correct")
+			assert.Equal(t, order.Filled, v[0].Status, "Order status should be correct")
+			assert.Empty(t, v[0].ClientOrderID, "client order ID should be empty")
+			assert.False(t, v[0].ReduceOnly, "Reduce only should be false")
+			assert.Equal(t, 1.0, v[0].ExecutedAmount, "Cum exec qty should be correct")
```
cranktakular: Failure message doesn't match variable name

exchanges/bybit/bybit_test.go
```diff
@@ -3062,22 +3058,200 @@ var pushDataMap = map[string]string{
...
+			assert.Equal(t, 75.0, v[0].AverageExecutedPrice, "Avg price should be correct")
+			assert.Equal(t, 0.358635, v[0].Fee, "Cum exec fee should be correct")
```
cranktakular: Failure message doesn't match variable name

exchanges/bybit/bybit_test.go
```diff
@@ -3062,22 +3058,200 @@ var pushDataMap = map[string]string{
...
+		if bytes.Contains(r, []byte("%s")) {
+			r = fmt.Appendf(nil, string(r), optionsTradablePair.String())
+		}
+		return e.wsHandleAuthenticatedData(ctx, r)
+	})
+	close(e.Websocket.DataHandler)
+	require.Len(t, e.Websocket.DataHandler, 6, "Should see correct number of messages")
+
+	i := 0
+	for data := range e.Websocket.DataHandler {
```
cranktakular, line 3110: This data validation here feels like overkill, do we want to test that this many parameters are being set?

exchanges/bybit/bybit_test.go
```diff
@@ -3623,48 +3791,43 @@ func TestGenerateSubscriptions(t *testing.T) {
...
-	require.NoError(t, testexch.Setup(e), "Test instance Setup must not error")
-	subs, err := e.Features.Subscriptions.ExpandTemplates(e)
-	require.NoError(t, err, "ExpandTemplates must not error")
-	e.Features.Subscriptions = subscription.List{}
-	testexch.SetupWs(t, e)
-	err = e.Subscribe(subs)
-	require.NoError(t, err, "Subscribe must not error")
+
+	e := new(Exchange) //nolint:govet // Intentional shadow
+	require.NoError(t, testexch.Setup(e))
```
cranktakular: I can't tell how you decide whether or not to include comments explaining the error. They generally feel superfluous to me, so I don't mind them being removed, but it seems inconsistent for only this one to be.

exchanges/bybit/bybit_websocket.go
cranktakular: I have not looked into that, I just want to ensure the checks are done _somewhere_ if they're being removed here.

exchanges/bybit/bybit_websocket.go
cranktakular: Seems a bit missable as-is, due to this section having some handling for DCP already, even if we don't invoke it just yet.

cranktakular (changes requested):
Got some more changes I want.

In addition to those comments, I'm pretty sure these functions could get more test coverage: handleSpotSubscription, wsHandleAuthenticatedData, handleNoTopicWebsocketResponse, wsProcessOrder, getPairFromCategory, Setup.

And I think these functions have a slight chance of being able to get more test coverage: GenerateOptionsDefaultSubscriptions, GenerateLinearDefaultSubscriptions, GenerateInverseDefaultSubscriptions

exchanges/bybit/bybit_websocket.go
shazbert: Will pass on this due to time constraints. You can open a PR against this merged state if it bugs you.

exchanges/bybit/bybit_websocket.go
shazbert: Will be RM'd for the appropriate PR #1672

exchanges/bybit/inverse_websocket_test.go
shazbert: Yeah it will probably resolve itself when they update upstream. Until then.

exchanges/bybit/spot_websocket_test.go
shazbert: Pass

exchanges/bybit/bybit_test.go
shazbert: I agree but there can be regressions especially for order management account level stuff that are pretty critical.

exchanges/bybit/bybit_test.go
shazbert: Yeah I generally do not include them because it has a line failure number. But you make a good point.

exchanges/bybit/bybit_websocket.go
shazbert: Yeah they won't generate subs, and having no subs means that no connection will be spawned and this function won't be called.

exchanges/bybit/bybit_websocket.go
shazbert: What would you like me to do, that is in scope for this PR?

shazbert:
> Got some more changes I want.
> 
> In addition to those comments, I'm pretty sure these functions could get more test coverage: handleSpotSubscription, wsHandleAuthenticatedData, handleNoTopicWebsocketResponse, wsProcessOrder, getPairFromCategory, Setup.
> 
> And I think these functions have a slight chance of being able to get more test coverage: GenerateOptionsDefaultSubscriptions, GenerateLinearDefaultSubscriptions, GenerateInverseDefaultSubscriptions

Will pass on this due to time constraints. You can open a PR against this merged state if it bugs you. These generally have sufficient coverage and this need to be merged sooner rather than later.

exchanges/bybit/spot_websocket.go
shazbert: prefer if short, won't resolve so people can yell at me.

gloriousCode: @shazbert I PREFER HOW YOU'VE DONE IT

exchanges/bybit/bybit_websocket.go
cranktakular: Remove (or comment out obviously correct) invocations of DCP, since it's currently unsupported, and that would require future implementations to make sure it adheres to the documentation in all relevant places.

cranktakular (approved):
Just got one suggested change left, other than that I can tACK this.

exchanges/bybit/bybit_websocket.go
```diff
@@ -63,94 +64,58 @@ var defaultSubscriptions = subscription.List{
 	{Enabled: true, Asset: asset.Spot, Channel: subscription.OrderbookChannel, Levels: 50},
 	{Enabled: true, Asset: asset.Spot, Channel: subscription.AllTradesChannel},
 	{Enabled: true, Asset: asset.Spot, Channel: subscription.CandlesChannel, Interval: kline.OneHour},
-	{Enabled: true, Asset: asset.Spot, Authenticated: true, Channel: subscription.MyOrdersChannel},
-	{Enabled: true, Asset: asset.Spot, Authenticated: true, Channel: subscription.MyWalletChannel},
-	{Enabled: true, Asset: asset.Spot, Authenticated: true, Channel: subscription.MyTradesChannel},
+	// Authenticated channels are currently being managed by the `generateAuthSubscriptions` method for the private connection
```
gloriousCode, line 68: Just to be clear about the removal here. Is this a change where authenticated subscriptions that are currently in user config subscriptions will cease working? Or is it that they do not work already? 

If the former, I think the user deserves a warning that you are removing support. I understand that there is an issue about sub propagation, but I think my suggestion will suffice.

I've made a suggestion for you below

exchanges/bybit/bybit_websocket.go
shazbert: Removed https://github.com/thrasher-corp/gocryptotrader/pull/1670/commits/b5b082a27174cff28eaf1ef74b9728f1d0b51ea9

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +669,164 @@ const subTplText = `
...
+}
+
+// generateAuthSubscriptions generates default subscription for the dedicated auth websocket connection. These are
+// agnostic to the asset type and pair as all account level data will be routed through this connection.
+// TODO: Remove this function when template expansion is across all assets
+func (e *Exchange) generateAuthSubscriptions() (subscription.List, error) {
+	if !e.Websocket.CanUseAuthenticatedEndpoints() {
+		return nil, nil
+	}
+	var subscriptions subscription.List
```
gloriousCode, line 789:
```suggestion
        for _, configSub := range e.Config.Features.Subscriptions.Enabled() {
		if configSub.Authenticated {
			log.Warnf(log.WebsocketMgr, "%q has an authenticated subscription %q in config which is not supported. Please remove",
				e.Name, configSub.Channel)
		}
	}
	
	var subscriptions subscription.List
```

exchanges/exchange.go
```diff
@@ -1380,6 +1379,16 @@ func (u URL) String() string {
 		return restSwapURL
 	case WebsocketSpot:
 		return websocketSpotURL
+	case WebsocketCoinMargined:
```
gloriousCode, line 1382: Thank you for expanding these. Its one of those things I get annoyed at and then forget about a second later

exchanges/bybit/bybit_test.go
```diff
@@ -3730,3 +3893,74 @@ func TestTransformSymbol(t *testing.T) {
 		})
 	}
 }
+
+func TestGetPairFromCategory(t *testing.T) {
```
gloriousCode: `e.matchPairAssetFromResponse(...)` I think is a better name for this and would like the rename

exchanges/bybit/bybit_websocket.go
gloriousCode: This broke tests 🐝

exchanges/exchange.go
shazbert: You are welcome, but thank mr @cranktakular for kicking my butt on this. I originally just purged this completely.

exchanges/bybit/bybit_websocket.go
shazbert: They will silently cease working, I have applied your suggestion thank you ser. 🖖

shazbert:
```console
[WARN]  | WEBSOCKET | 06/08/2025 10:58:52 | "Bybit" has an authenticated subscription "myOrders" in config which is not supported. Please remove
[WARN]  | WEBSOCKET | 06/08/2025 10:58:52 | "Bybit" has an authenticated subscription "myWallet" in config which is not supported. Please remove
[WARN]  | WEBSOCKET | 06/08/2025 10:58:52 | "Bybit" has an authenticated subscription "myTrades" in config which is not supported. Please remove
```

exchanges/bybit/bybit_websocket.go
```diff
@@ -63,94 +65,61 @@ var defaultSubscriptions = subscription.List{
...
 }
 
 var subscriptionNames = map[string]string{
 	subscription.TickerChannel:    chanPublicTicker,
 	subscription.OrderbookChannel: chanOrderbook,
 	subscription.AllTradesChannel: chanPublicTrade,
 	subscription.MyOrdersChannel:  chanOrder,
-	subscription.MyTradesChannel:  chanExecution,
 	subscription.MyWalletChannel:  chanWallet,
+	subscription.MyTradesChannel:  chanExecution,
```
gloriousCode, line 78:
```
[DEBUG] | WEBSOCKET | 07/08/2025 07:51:42 | Bybit wss://stream.bybit.com/v5/public/spot: Sending message: {"req_id":"2","op":"subscribe","args":["execution.spot","order.spot","wallet"]}
...
[DEBUG] | WEBSOCKET | 07/08/2025 07:51:42 | Bybit wss://stream.bybit.com/v5/public/spot: Message received: {"success":false,"ret_msg":"Invalid topic :[execution.spot]","conn_id":"6afb9045-16cc-483a-ba15-5fb8cac66a2b","req_id":"2","op":"subscribe"}
...
[ERROR] | WEBSOCKET | 07/08/2025 07:41:11 | Bybit Error subscribing subscribe with request ID 2 msg: Invalid topic :[execution.spot]
```

I've been getting this using both subscriptions from master and generated ones on this branch

It does not happen on master

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +674,171 @@ const subTplText = `
...
+				return nil, err
+			}
+			arg.Arguments = append(arg.Arguments, s.Channel+"."+interval+"."+pairFmt.Format(pair))
+			arg.associatedSubs = append(arg.associatedSubs, s)
+		case chanPositions, chanExecution, chanOrder, chanWallet, chanGreeks:
+			if chanMap[s.Channel] {
+				continue
+			}
+			authArg.Arguments = append(authArg.Arguments, s.Channel)
+			// adding the channel to selected channels so that we will not visit it again.
```
gloriousCode:
```suggestion
			// add channel name to map so we only subscribe to channel once
```

gloriousCode (changes requested):
I think once the bug is addressed I'm happy with the work

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +674,171 @@ const subTplText = `
...
+// agnostic to the asset type and pair as all account level data will be routed through this connection.
+// TODO: Remove this function when template expansion is across all assets
+func (e *Exchange) generateAuthSubscriptions() (subscription.List, error) {
+	if !e.Websocket.CanUseAuthenticatedEndpoints() {
+		return nil, nil
+	}
+
+	for _, configSub := range e.Config.Features.Subscriptions.Enabled() {
+		if configSub.Authenticated {
+			log.Warnf(log.WebsocketMgr, "%q has an authenticated subscription %q in config which is not supported. Please remove", e.Name, configSub.Channel)
```
gloriousCode:
```suggestion
			log.Warnf(log.WebsocketMgr, "%q has an authenticated subscription %q in config which is not supported. Please remove", e.Name, configSub.Channel)
			configSub.Enabled = false
```

exchanges/bybit/bybit_websocket.go
```diff
@@ -755,7 +636,7 @@ func channelName(s *subscription.Subscription) string {
 // isSymbolChannel returns whether the channel accepts a symbol parameter
 func isSymbolChannel(name string) bool {
 	switch name {
-	case chanPositions, chanExecution, chanOrder, chanDCP, chanWallet:
+	case chanPositions, chanExecution, chanOrder, chanWallet:
 		return false
 	}
 	return true
```
gloriousCode, line 640: Update `subTplText` to not process spot auth subs (ie `isCategorisedChannel` subs):
```
{{ with $name := channelName $.S }}
	{{- range $asset, $pairs := $.AssetPairs }}
		{{- if isSymbolChannel $name }}
			{{- range $p := $pairs }}
				{{- $name -}} .
				{{- if eq $name "orderbook" -}} {{- $.S.Levels -}} . {{- end }}
				{{- if eq $name "kline" -}} {{- intervalToString $.S.Interval -}} . {{- end }}
				{{- $p }}
				{{- $.PairSeparator }}
			{{- end }}
		{{- end }}
	{{- end }}
	{{- $.AssetSeparator }}
{{- end }}
```

gloriousCode (changes requested):
This should fix your issues

exchanges/bybit/bybit_websocket.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1670/commits/80a5edb11e8b8d4093df7c3d133def5dc2b45b60

gloriousCode (approved):
i am pleased. tACK

exchanges/bybit/bybit_websocket.go
```diff
@@ -793,3 +671,172 @@ const subTplText = `
...
+// agnostic to the asset type and pair as all account level data will be routed through this connection.
+// TODO: Remove this function when template expansion is across all assets
+func (e *Exchange) generateAuthSubscriptions() (subscription.List, error) {
+	if !e.Websocket.CanUseAuthenticatedEndpoints() {
+		return nil, nil
+	}
+
+	for _, configSub := range e.Config.Features.Subscriptions.Enabled() {
+		if configSub.Authenticated {
+			log.Warnf(log.WebsocketMgr, "%q has an authenticated subscription %q in config which is not supported. Please remove", e.Name, configSub.Channel)
```
thrasher-:
```suggestion
			log.Warnf(log.WebsocketMgr, "%s has an authenticated subscription %q in config which is not supported. Please remove.", e.Name, configSub.Channel)
```

Just a consistency nit since we generally use the exchange name without quotes

common/common.go
```diff
@@ -669,15 +669,15 @@ func SortStrings[S ~[]E, E fmt.Stringer](x S) S {
 
 // Counter is a thread-safe counter.
 type Counter struct {
-	n int64 // privatised so you can't use counter as a value type
+	n atomic.Int64 // privatised so you can't use counter as a value type
```
thrasher-: A wise man once said:

"This realignment due to the requirement for assignment alignment brings no excitement, only chastisement that my review indictment brought defilement"

However this change rids the need for such things 😆

gbjk: 🦍 
```suggestion
	n atomic.Int64 // private so you can't use counter as a value type
```

common/common.go
```diff
@@ -669,15 +669,15 @@ func SortStrings[S ~[]E, E fmt.Stringer](x S) S {
...
 }
 
 // IncrementAndGet returns the next count after incrementing.
 func (c *Counter) IncrementAndGet() int64 {
-	newID := atomic.AddInt64(&c.n, 1)
+	newID := c.n.Add(1)
 	// Handle overflow by resetting the counter to 1 if it becomes negative
```
gbjk, line 678: If this is a straight up counter, why don't we just use `atomic.Uint64` ?

We'd still need to handle overflow it it's already `math.MaxUint64`, but we'd have more available positive values

common/common_test.go
```diff
@@ -644,7 +644,8 @@ func TestSortStrings(t *testing.T) {
 
 func TestCounter(t *testing.T) {
 	t.Parallel()
-	c := Counter{n: -5}
+	c := Counter{}
+	c.n.Store(-5)
 	require.Equal(t, int64(1), c.IncrementAndGet())
```
gbjk: Since this test is checking boundary behaviour, it should have a comment.
And if we change to Uint64 the same applies.
```suggestion
	require.Equal(t, int64(1), c.IncrementAndGet(), "Adding to a negative Counter must reset to zero and then increment")
```

exchanges/bybit/bybit_test.go
```diff
@@ -3018,20 +3019,39 @@ func TestCancelBatchOrders(t *testing.T) {
 	}
 }
 
-type DummyConnection struct{ websocket.Connection }
+type DummyConnection struct {
```
gbjk: 🦍 Just a preference for naming [test fixtures](https://en.wikipedia.org/wiki/Test_fixture).
```suggestion
type FixtureConnection struct {
```

exchanges/bybit/bybit_test.go
```diff
@@ -3051,31 +3073,43 @@ func TestPushDataPublic(t *testing.T) {
...
+		if k == "unhandled" {
+			return errUnhandledStreamData
+		}
+		return nil
+	}
+
 	for x := range keys {
-		err := e.wsHandleData(t.Context(), asset.Spot, []byte(pushDataMap[keys[x]]))
-		assert.NoError(t, err, "wsHandleData should not error")
+		assert.ErrorIsf(t, e.wsHandleData(nil, asset.Spot, []byte(pushDataMap[keys[x]])), expErrFn(keys[x]), "wsHandleData should not error for %s", keys[x])
```
gbjk: 🚧 The test message here is not correct.
🐒 Additionally this felt less clear and I think we should avoid ErrorIs `nil` where possible
```suggestion
	for x := range keys {
		err := e.wsHandleData(t.Context(), asset.Spot, []byte(pushDataMap[keys[x]]))
		if keys[x] == "unhandled" {
			assert.ErrorIs(t, err, errUnhandledStreamData, "wsHandleData should error correctly for unhandled topics")
		} else {
			assert.NoError(t, err, "wsHandleData should not error")
		}
	}
```

exchanges/bybit/bybit_websocket.go
```diff
@@ -238,71 +189,59 @@ func (by *Bybit) wsHandleData(_ context.Context, assetType asset.Item, respRaw [
...
+	return fmt.Errorf("%w %s", errUnhandledStreamData, string(respRaw))
+}
+
+func (e *Exchange) handleNoTopicWebsocketResponse(conn websocket.Connection, result *WebsocketResponse, respRaw []byte) error {
+	switch result.Operation {
+	case "subscribe", "unsubscribe", "auth":
+		if result.RequestID != "" {
+			return conn.RequireMatchWithData(result.RequestID, respRaw)
+		}
+
```
gbjk: 🦍 
```suggestion
```

exchanges/bybit/bybit_websocket.go
```diff
@@ -730,9 +661,6 @@ const subTplText = `
 				{{- $p }}
 				{{- $.PairSeparator }}
 			{{- end }}
-		{{- else }}
-			{{- $name }}
-			{{- if and (isCategorisedChannel $name) ($categoryName := getCategoryName $asset) -}} . {{- $categoryName -}} {{- end }}
```
gbjk, line 785: 🚧 I think this means we should be removing `isCategorisedChannel`.
I can't see it used anywhere. Sorry if I've missed it.

exchanges/bybit/bybit_websocket.go
```diff
@@ -891,10 +826,10 @@ func (by *Bybit) getPairFromCategory(category, symbol string) (currency.Pair, as
 	case "option":
 		assets = append(assets, asset.Options)
 	default:
-		return currency.EMPTYPAIR, 0, fmt.Errorf("category %q not supported for incoming symbol %q", category, symbol)
+		return currency.EMPTYPAIR, 0, fmt.Errorf("incoming symbol %q %w: %q", symbol, errUnsupportedCategory, category)
```
gbjk, line 819: 🐒 This error feels back to front
```suggestion
		return currency.EMPTYPAIR, 0, fmt.Errorf("%w %q for symbol %q %w: %q", errUnsupportedCategory, category, symbol)
```

gbjk (changes requested):
Thanks for the updates. 

Just a few change requests and nitpicks.

common/common.go
shazbert: That is a great point! 🔥

shazbert: Going to defer this as it impacts quite a few things in websocket manager. But your future idea of turning ID's to strings and going towards uuid v7 for default might be better as a complementary update.

exchanges/bybit/bybit_test.go
shazbert: much better

common/common.go
gbjk: Agreed

gbjk (approved):
Thanks for addressing those. I'm happy.

thrasher- (approved):
Changes ACK

PR #1671, opened 2024-10-10
Config: Add versioning

gbjk:
# Config

* Add config versioning manager. Motivation:
  * Huobi would like to move futures to coinmargined futures, but we don't want to lose the user's existing config
  * We have a number of places we have deprecated config clutter, and pointers to native types, purely so we can upgrade it. 
* Version1 handles the pairs manager format migration
* Version2 handles GDAX to Coinbase rename
* Version3 will handle AssetEnabled, but keeping this initial PR minimal
* Upgrade configtest to Version2
* Not upgrading config_example because it would jump over some null changes - I'll handle those in Version4 quickly
* Add -decrypt and -upgrade to cmd/config, no longer explicitly decrypts, or automatically reverses what was there
* Add -edit to edit a file in place
* -in instead of -infile and -out instead of outfile
* Do not echo input to user when entering encryption key
* 🐛 Fix encryption keys not allowing spaces
* This has some light touch improvements to encryption handling. I think I'm going to have to circle back to that separately later.

**Note Well:** The test failure in `TestConfigAllJsonResponse` is fixed by the dependency below

# Dependencies
* #1593 fixes `TestConfigAllJsonResponse`

## Type of change


config/versions/versions.go
```diff
@@ -0,0 +1,202 @@
...
+func (m *manager) registerVersion(v any) {
+	m.m.Lock()
+	defer m.m.Unlock()
+	ver, err := strconv.Atoi(strings.TrimSuffix(strings.TrimPrefix(fmt.Sprintf("%T", v), "*v"), ".Version"))
+	if err != nil {
+		m.errors = common.AppendError(m.errors, fmt.Errorf("%w '%T': %w", errRegisteringVersion, v, err))
+		return
+	}
+	switch v.(type) {
+	case ExchangeVersion, ConfigVersion:
```
shazbert, line 181: Any issues if v conforms to both at the same time? 🤯

config/versions/versions.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/buger/jsonparser"
+	"github.com/thrasher-corp/gocryptotrader/common"
+)
+
+var (
+	errRegisteringVersion  = errors.New("error registering config version")
+	errVersionIncompatible = errors.New("version does not implement ConfigVersion or ExchangeVersion")
+	errVersionSequence     = errors.New("version registered out of sequence")
+	errModifiyingExchange  = errors.New("error modifying exchange config")
```
shazbert, line 30:
```suggestion
	errModifyingExchange  = errors.New("error modifying exchange config")
```

config/versions/versions.go
```diff
@@ -0,0 +1,202 @@
...
+	errNoVersions          = errors.New("error retrieving latest config version: No config versions are registered")
+	errApplyingVersion     = errors.New("error applying version")
+)
+
+// DisabledVersion allows authors to rollback changes easily during development
+type DisabledVersion interface {
+	Disabled() bool
+}
```
shazbert, line 38: This needs handling or RM until needed

shazbert (changes requested):
Good stuff. Will give it a thorough testing next round. Thanks.

cmd/config/config.go
```diff
@@ -1,82 +1,96 @@
...
+	if outFile == "" {
+		outFile = inFile + ".out"
+	}
+
+	fmt.Println("GoCryptoTrader: config-helper tool.")
+
+	if !(encrypt || decrypt || upgrade) {
+		fatal("Must provide one of -encrypt, -decrypt or -upgrade")
+	}
+
```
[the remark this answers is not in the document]
gbjk: Undoubtedly more than you were asking for, but you should know not to encourage me by now.

Fixed 048bced6f

cmd/config/config.go
```diff
@@ -1,82 +1,96 @@
...
+	}
 	if err != nil {
-		log.Fatalf("Unable to write output file %s. Error: %s", outFile, err)
+		fatal(err.Error())
 	}
-	log.Printf(
-		"Successfully %s input file %s and wrote output to %s.\n",
-		EncryptOrDecrypt(encrypt), inFile, outFile,
-	)
+	fmt.Println("Success! File written to ", out)
```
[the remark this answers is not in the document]
gbjk: Fixed 048bced

cmd/config/config.go
```diff
@@ -1,82 +1,96 @@
...
-	var encrypt bool
+	var encrypt, decrypt, upgrade bool
 	defaultCfgFile := config.DefaultFilePath()
 	flag.StringVar(&inFile, "infile", defaultCfgFile, "The config input file to process.")
-	flag.StringVar(&outFile, "outfile", defaultCfgFile+".out", "The config output file.")
-	flag.BoolVar(&encrypt, "encrypt", true, "Whether to encrypt or decrypt.")
+	flag.StringVar(&outFile, "outfile", "", "The config output file.")
+	flag.BoolVar(&encrypt, "encrypt", false, "Encrypt the config file")
+	flag.BoolVar(&decrypt, "decrypt", false, "Decrypt the config file")
+	flag.BoolVar(&upgrade, "upgrade", false, "Upgrade the config file")
```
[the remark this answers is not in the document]
gbjk: I subscribe to minimal full stops, and in this case I'm following the example from go (`go help`)

Fixed 048bced

config/config_test.go
```diff
@@ -1321,7 +1317,6 @@ func TestCheckExchangeConfigValues(t *testing.T) {
 
 	// Test API settings migration
```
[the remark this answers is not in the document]
gbjk: Can you check if that's true?
Does another process have test.json open?
I can't imagine that this is go itself, because it wouldn't constitute "another process" itself.

config/config_test.go
```diff
@@ -1321,7 +1317,6 @@ func TestCheckExchangeConfigValues(t *testing.T) {
...
 	sptr := func(s string) *string { return &s }
-	int64ptr := func(i int64) *int64 { return &i }
 
 	cfg.Exchanges[0].APIKey = sptr("awesomeKey")
```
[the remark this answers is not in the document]
gbjk: See Dependencies.

I'll make it clearer above so nobody else has to ask.

config/config.go
```diff
@@ -840,263 +836,210 @@ func (c *Config) UpdateExchangeConfig(e *Exchange) error {
...
-		return errors.New("no exchange configs found")
+		return errNoEnabledExchanges
 	}
 
 	exchanges := 0
 	for i := range c.Exchanges {
-		if strings.EqualFold(c.Exchanges[i].Name, "GDAX") {
-			c.Exchanges[i].Name = "CoinbasePro"
+		e := &c.Exchanges[i]
+		if strings.EqualFold(e.Name, "GDAX") {
```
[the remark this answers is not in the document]
gbjk: 2029 ?
Though I can just make it a Version :)

config/versions/versions.go
gbjk: Oooh. Thanks for that catch. I lost some code here.

config/config.go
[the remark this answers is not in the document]
gbjk: Fixed 633f6b21

config/config.go
```diff
@@ -1530,115 +1468,103 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
+
+	if j, err = versions.Manager.Deploy(context.Background(), j); err != nil {
+		return err
+	}
+
+	if err := json.Unmarshal(j, c); err != nil {
+		return err
+	}
+
 	return nil
```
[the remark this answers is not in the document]
gbjk: Fixed cd6f7d06e

config/config.go
```diff
@@ -1530,115 +1468,103 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
-	if !ConfirmECS(pref) {
-		// Read unencrypted configuration
-		decoder := json.NewDecoder(reader)
-		c := &Config{}
-		err = decoder.Decode(c)
-		return c, false, err
+	c.EncryptConfig = fileEncryptionDisabled
+	if err := c.SaveConfigToFile(path); err != nil {
+		return err
 	}
```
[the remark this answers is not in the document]
gbjk: Fixed cd6f7d06e

config/config.go
```diff
@@ -1530,115 +1468,103 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
+	return nil
 }
 
-// readEncryptedConf reads encrypted configuration and requests key from provider
-func readEncryptedConfWithKey(reader *bufio.Reader, keyProvider func() ([]byte, error)) (*Config, error) {
-	fileData, err := io.ReadAll(reader)
-	if err != nil {
-		return nil, err
-	}
+type keyProvider func() ([]byte, error)
```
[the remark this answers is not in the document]
gbjk: I'd like to keep it next to usage, since it's only used in these 2 funcs.
It should be above `readConfig` though!

Fixed a70c38ac5

Discussion:
TL;DR: This type is not the same as the exported/shared types in config_types.go and can live here, and that's idiomatic.
I think I've mentioned before, but IMHO `_types.go` files are mostly idiosyncratic; There's cases where they make sense for a large collection of types related to an exchange that hasn't been broken down, but it's more idiomatic to have organisation where the types and their related functions live in separate files. That's how it is in Go source, unless I've missed any examples, and popular projects ( docker, kubernetes, hugo, gin, prometheus )
The bulk of config types are fine in `_types` because they really are shared types without function, and would overwhelm a standard module package.

config/config.go
```diff
@@ -1530,115 +1468,103 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
 			log.Errorf(log.ConfigMgr, "PromptForConfigKey err: %s", err)
 			continue
 		}
 
-		var c *Config
-		c, err = readEncryptedConf(bytes.NewReader(fileData), key)
+		d, err := c.decryptConfigData(j, key)
 		if err != nil {
-			log.Errorln(log.ConfigMgr, "Could not decrypt and deserialise data with given key. Invalid password?", err)
+			log.Errorln(log.ConfigMgr, "Could not decrypt and deserialise data with given key. Invalid password? ", err)
```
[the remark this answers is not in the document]
gbjk: Fixed aee2f5b20

config/versions/versions.go
gbjk: Fixed 8cf5f2d92 🙏

config/versions/versions.go
gbjk: Nope. A version can be both.

Fixed 727d1407f 📖

config/versions/versions.go
gbjk: Fixed a5bb916c7 🙏

gbjk:
@shazbert I'm done restructuring this.
All versions live in the same package, in separate files.
Stateful representations of types used for state-in-time upgrades live in vN/types.go
Not expecting types for each version. A type that changes a thing can introduce the new representation, and if necessary add the old representation to a previous version as well.

cmd/config/config.go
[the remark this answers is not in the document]
shazbert: 😆 nice

config/config.go
```diff
@@ -1510,13 +1450,8 @@ func GetFilePath(configFile string) (configPath string, isImplicitDefaultPath bo
 // config directory as `File` or `EncryptedFile` depending on whether the config
 // is encrypted
```
shazbert, line 1451: Out of scope: But while you are here can you change the log.Warn to fmt.Printlns for SetNTPCheck; as they don't appear if log levels are not set and the code block pauses execution. 
image

config/config_test.go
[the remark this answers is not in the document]
shazbert: No idea seems like its a windows issue, happening on another computer:

```
    c:\Users\graem\ryan\thrasher-corp\gocryptotrader\config\config_test.go:2105: migrateConfig() error = remove test.json: The process cannot access the file because it is being used by another process., wantErr false
=== RUN   TestMigrateConfig/source_same_as_target
=== RUN   TestMigrateConfig/source_and_target_present
config file already found in 'C:\Users\graem\AppData\Local\Temp\TestMigrateConfig1792913445\001\config.json'; not overwriting, defaulting to C:\Users\graem\AppData\Local\Temp\TestMigrateConfig1792913445\001\src\config.dat
    c:\Users\graem\ryan\thrasher-corp\gocryptotrader\config\testing.go:1232: TempDir RemoveAll cleanup: remove C:\Users\graem\AppData\Local\Temp\TestMigrateConfig1792913445\001\config.json: The process cannot access the file because it is being used by another process.
--- FAIL: TestMigrateConfig (1.87s)
    --- PASS: TestMigrateConfig/nonexisting (0.00s)
    --- FAIL: TestMigrateConfig/source_present,_no_target_dir (0.00s)
    --- PASS: TestMigrateConfig/source_same_as_target (0.00s)
    --- PASS: TestMigrateConfig/source_and_target_present (0.00s)
```

shazbert (changes requested):
Nice changes! Thanks for changing the requested. Also just mentioning config_example.json can we now shift that up in line with this versioning system, I see you have commited the configtest.json file.

gbjk:
> Nice changes! Thanks for changing the requested. Also just mentioning config_example.json can we now shift that up in line with this versioning system, I see you have commited the configtest.json file.

See PR description 🙂 
> Not upgrading config_example because it would jump over some null changes - I'll handle those in Version4 quickly
The test config I could upgrade without introducing any new nulls.

cmd/config/config.go
```diff
@@ -1,82 +1,146 @@
...
+
+	fmt.Println("Success! File written to " + out)
+}
+
+func upgradeFile(in, out string) {
+	if config.IsFileEncrypted(in) {
+		fatal("Cannot upgrade an encrypted file. Please decrypt first")
+	}
+	c := &config.Config{}
+	err := c.ReadConfigFromFile(in, true)
```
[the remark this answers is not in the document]
gbjk: Agreed 🙏 

Fixed 89155efbf

config/versions/versions.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+	if len(m.versions) != ver {
+		m.errors = common.AppendError(m.errors, fmt.Errorf("%w: %v", errVersionSequence, ver))
+		return
+	}
+	m.versions = append(m.versions, v)
+}
+
+// latest returns the highest version number
+// May return -1 if something has gone deeply wrong
```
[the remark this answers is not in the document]
gbjk: Fixed 54b7029fc

cmd/config/config.go
```diff
@@ -1,82 +1,146 @@
...
-			log.Fatalf("Unable to obtain encryption/decryption key: %s", err)
-		}
-		key = string(result)
+	if out == "[infile].out" {
+		out = in + ".out"
 	}
 
-	fileData, err := os.ReadFile(inFile)
+	switch parseCommand(fs) {
+	case "upgrade":
```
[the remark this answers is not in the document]
gbjk: If we were going to do this, I think we'd want to also take in the version to move to, and break out a bit of extra version handling for it.
I don't have a use case for that yet.
I think I'd like to live with this for a while before adding features.
Is that okay?

config/config.go
gbjk: Yeah, since this bug isn't too important if it's lost, I can driveby it now.
I'm unhappy with the theme  of user-interaction from the CLI, where we act more like a windows app hand-holding the user with pop-up prompts on edge cases instead of just bailing on the user and telling them to fix it.
We have a lot of code, maintenance, bugs, dedicated to user interaction for paths that are one-off, rare, and could be error and bail, and I think this is one of them. That said: Backing away...

Fixed 95cd6c21b

config/config_test.go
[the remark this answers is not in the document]
gbjk: There was an inconsistency on test.json being not in TempDir, and trying to remove it when it shouldn't have been there.
I'm not sure if I've fixed it or not, but let's see.
Since I haven't been able to reproduce, Could you please test this for me?

Fixed 9a59bae7f

cmd/config/config.go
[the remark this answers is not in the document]
shazbert: No thanks this is good for now. 👍

shazbert (changes requested):
Thanks sir, just a diff with changes that needs to be added then looks good to go after merge with master. Nice work!

config/config_test.go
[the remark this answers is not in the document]
shazbert: See [this](https://github.com/thrasher-corp/gocryptotrader/pull/1671#discussion_r1818248800)

config/config_encryption.go
```diff
@@ -8,29 +8,30 @@ import (
...
+	saltRandomLength = 12
+)
+
+var (
+	errAESBlockSize = errors.New("config file data is too small for the AES required block size")
+	errNoPrefix     = errors.New("data does not start with Encryption Prefix")
+
+	// encryptionPrefix is a prefix to tell us the file is encrypted
+	encryptionPrefix = []byte("THORS-HAMMER")
+	saltPrefix       = []byte("~GCT~SO~SALTY~")
```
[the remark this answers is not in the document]
gbjk: The change was literally to move it *from* being a const string 🙁 
If we make it a string constant it'll be untyped.
I changed it so we can use it directly in `bytes.` functions without runtime conversion constantly (heh).

config/config_test.go
```diff
@@ -2141,72 +2023,57 @@ func TestMigrateConfig(t *testing.T) {
...
-			cleanup: func(t *testing.T) {
-				t.Helper()
-				os.Remove("test.json")
+				test, err := os.Create(filepath.Join(dir, "test.json"))
+				require.NoError(t, err, "os.Create must not error")
+				require.NoError(t, test.Close(), "file Close must not error")
 			},
 			args: args{
-				configFile: "test.json",
+				configFile: filepath.Join(dir, "test.json"),
```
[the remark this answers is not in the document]
gbjk: Fixed 96d9033d4 and 86a00a7c3

config/config_encryption.go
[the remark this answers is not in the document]
shazbert: cmd/compile nor gccgo I didn't think allocates those strings for conversion, but either way your way is the highway and more concise. 👍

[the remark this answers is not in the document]
gbjk: Maybe I'm misunderstanding you:
```
M config/config_encryption.go
@@ -17,6 +17,7 @@ import (
 
 const (
 	saltRandomLength = 12
+	saltPrefix       = `~GCT~SO~SALTY~`
 )
 
 var (
@@ -25,7 +26,6 @@ var (
 
 	// encryptionPrefix is a prefix to tell us the file is encrypted
 	encryptionPrefix = []byte("THORS-HAMMER")
-	saltPrefix       = []byte("~GCT~SO~SALTY~")
```
👇 
`config/config_encryption.go|144 col 24| cannot use saltPrefix (untyped string constant "~GCT~SO~SALTY~") as []byte value in argument to bytes.HasPrefix`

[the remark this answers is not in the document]
shazbert: You understanded me well, but you have to add the []byte conversion which is janky which is not nice and a good point.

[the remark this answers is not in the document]
gbjk: For posterity:
Under most architectures, we'll end up with 2 extra instructions at runtime compared to if we use a pkg const.
There's no extra allocs, just the extra instructions.
Not worth caring about, but just out of curiosity.

shazbert (approved):
Thanks. Nothing else from me.

gbjk:
Rebased.

I've dropped out the "upgrade test config" change. I've realised that I should work through moving all of the things that trigger warnings to proper versions first, then it won't even matter. That's pretty trivial but I'll let this land as-is first.
Otherwise we'll get endless conflicts until this is done.

The method to upgrade it is to run config tool with upgrade, then remove anything that would be nulled out.
Which is precisely why I'm not doing it now.

cmd/config/config.go
```diff
@@ -1,82 +1,145 @@
...
-	flag.BoolVar(&encrypt, "encrypt", true, "Whether to encrypt or decrypt.")
-	flag.StringVar(&key, "key", "", "The key to use for AES encryption.")
-	flag.Parse()
 
-	log.Println("GoCryptoTrader: config-helper tool.")
+	var in, out, key string
+	fs := flag.NewFlagSet("config", flag.ExitOnError)
+	fs.Usage = func() { usage(fs) }
+	fs.StringVar(&in, "infile", defaultCfgFile, "The config input file to process")
+	fs.StringVar(&out, "outfile", "[infile].out", "The config output file")
```
gloriousCode, line 25: I will forgive it  because the existing application was also unfriendly. No way to upgrade in place with a `config.vX.json` previous version output
Hate having to fiddle with files after doing the task.

I'm just ranting. This is something I can add later

cmd/config/config.go
```diff
@@ -1,82 +1,145 @@
...
+	case "encrypt":
+		encryptWrapper(in, out, key, encryptFile)
+	}
 
+	fmt.Println("Success! File written to " + out)
+}
+
+func upgradeFile(in, out string) {
+	if config.IsFileEncrypted(in) {
+		fatal("Cannot upgrade an encrypted file. Please decrypt first")
```
gloriousCode, line 47: Why can't you prompt for key? Why can't there be decrypted config in memory instead of having to save decryption to disk to continue?

```
go run . --infile="/Users/scottg/.gocryptotrader/config.json.out" --key="1" upgrade
GoCryptoTrader: config-helper tool
Cannot upgrade an encrypted file. Please decrypt first
exit status 2
```

engine/engine.go
```diff
@@ -71,7 +71,7 @@ func New() (*Engine, error) {
 		return nil, fmt.Errorf("failed to load config. Err: %s", err)
 	}
 
-	return &b, nil
+	return &b, err
```
gloriousCode, line 74: ?

config/versions/v0.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// Version0 is a baseline version with no changes, so we can downgrade back to nothing
+// It does not implement any upgrade interfaces
+type Version0 struct {
+}
+
+func init() {
+	Manager.registerVersion(0, &Version0{})
+}
```
gloriousCode, line 14: `init` are read in filename order. You have limited this functionality to a total of 10 versions using your current naming convention

I created [10 upgrade version files ](https://github.com/user-attachments/assets/1d55d813-49d6-4445-a771-729bc5244b75)and structs for testing, in each I add a println to highlight this:

```go
func init() {
	fmt.Println("v0")
	Manager.registerVersion(0, &Version0{})
}
```

```
 go run . upgrade
v0
v1
v10
v2
v3
v4
v5
v6
v7
v8
v9
GoCryptoTrader: config-helper tool
version registered out of sequence: 10
exit status 2
```

Easiest solution while still wanting to rely on `init` is to rename files to have leading zeroes.

gloriousCode (changes requested):
I'm excited for config versioning. Just did some light prodding

cmd/config/config.go
gbjk: Yeah, I dislike this too.
I thought I'd get too much pushback if I changed it.
I'd like us to support unix centric as a starting point; Unless otherwise specified `-in` defaults to `-` meaning STDIN and `-out` defualts to `-` meaning STDOUT. If you specify `-in` but not `-out` you'd get output to STDOUT.
Add `-e | --edit` for edit-in-place.
I think that'd work for windows as well as unix.
I could do this now if it has support.

cmd/config/config.go
gbjk: We could.
I'm trying to decrease footprint.
There's a lot of places where we've added features that aren't necessary, and it's resulted in a very wide codebase.
This is one of the places where the number of times a user will run upgrade _manually_ is limited, and supporting decryption isn't free.
Note: I *really* want to break encryption out of this entirely as soon as hell.
Secrets belong in a secret store merged into config, not mixed in with config.
Config merging and secret mounting was one of the most painful areas of service-orienting gct for cloud deployment for me originally.

So: Do you need me to support decryption for upgrades? :)

engine/engine.go
gbjk: reducible complexity must have occurred.

Fixed [76067d3b8](https://github.com/gbjk/gocryptotrader/commit/76067d3b8)

config/versions/v0.go
gbjk: Pretty sure that I can revamp that sequence requirement, since I added version number in.
And I don't think it's strictly necessary.

gbjk: Fixed [d4eb81025](https://github.com/gbjk/gocryptotrader/commit/d4eb81025)

cmd/config/config.go
gloriousCode: >This is one of the places where the number of times a user will run upgrade manually is limited

? I can only review the code and the pull request text in front of me. Is the goal here to upgrade user's config in place on run in future or something?

I have to decrypt my config, upgrade my config with a new infile and outfile, encrypt it again with a new infile and outfile and replace my original config to upgrade my config then delete the old ones because I don't want my keys to be unencrypted at rest. Below is with an encrypted config
```
go run . --key="1" --infile="/Users/DOGE/.gocryptotrader/config.json.out2" upgrade
GoCryptoTrader: config-helper tool
2024/11/14 11:17:39 Running upgrade to config version 0
2024/11/14 11:17:39 Running upgrade to config version 1
2024/11/14 11:17:39 Running upgrade to config version 2
2024/11/14 11:17:39 Version management finished
Success! File written to /Users/scottg/.gocryptotrader/config.json.out2.out
```

I'll give you the benefit of the doubt and not make you do it then. Right now its an elaborate process prone to error due to the level of manual interventions and I was just suggesting ways to improve it.

gbjk: Yeah, okay, I see your point.
I'll make it so.

cmd/config/config.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/1b498f3bb

Mainly this gives you `-edit` to solve your problem.

This also fixes location of the command so it can be anywhere.
I didn't go so far as to switch to unixesque usage, because I don't think we have a use-case for input.

cmd/config/config.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/1b498f3bb

Upgrade now supports encryption.

gloriousCode: Excellent updates, thank you 🚀

config/versions/versions.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+// registerVersion takes instances of config versions and adds them to the registry
+func (m *manager) registerVersion(ver int, v any) {
+	m.m.Lock()
+	defer m.m.Unlock()
+	if ver >= len(m.versions) {
+		m.versions = slices.Grow(m.versions, ver+1)[:ver+1]
+	}
+	m.versions[ver] = v
```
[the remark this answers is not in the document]
gbjk: Agreed on both counts 😂 
Anyone who does that deserves what happens next.

config/versions/versions.go
```diff
@@ -0,0 +1,200 @@
...
+	"slices"
+	"strconv"
+	"sync"
+
+	"github.com/buger/jsonparser"
+	"github.com/thrasher-corp/gocryptotrader/common"
+)
+
+var (
+	errRegisteringVersion  = errors.New("error registering config version")
```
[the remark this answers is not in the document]
gbjk: Fixed [d6f3eb238](https://github.com/gbjk/gocryptotrader/commit/d6f3eb238)

config/versions/versions.go
```diff
@@ -0,0 +1,200 @@
...
+		return j, err
+	}
+
+	m.m.RLock()
+	defer m.m.RUnlock()
+
+	current64, err := jsonparser.GetInt(j, "version")
+	current := int(current64)
+	switch {
+	case err == jsonparser.KeyPathNotFoundError:
```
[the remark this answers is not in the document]
gbjk: No, this is just wrong. Don't know why I did it.

Fixed https://github.com/gbjk/gocryptotrader/commit/8dde6859d

config/versions/versions.go
```diff
@@ -0,0 +1,200 @@
+/*
```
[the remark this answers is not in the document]
gbjk: Thanks.

Fixed https://github.com/gbjk/gocryptotrader/commit/1786defba

cmd/config/config.go
```diff
@@ -1,82 +1,178 @@
...
+	if inplace {
+		out = in
+	} else if out == "[in].out" {
+		out = in + ".out"
+	}
+
+	key := []byte(keyStr)
+	var err error
+	switch cmd {
+	case "upgrade":
```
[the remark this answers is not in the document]
gbjk: I can't replicate :'(
The first failure was harmless. It errored, didn't rewrite, and everything should be good.
I can run encrypt, upgrade over and over, and it's always encrypted afterwards.

Stupid questions, because I don't know what else to check:
* Is there anything omitted from this shell transcript?
* Anything might have happened in another terminal?
* Can you replicate again?

gbjk:
Rebased on master and squashed ready for last review.

shazbert (changes requested):
Linter issues need to be rectified then this looks good to go! Nice work 🚀

gbjk:
@shazbert Linter fixed. Was trivial fixes so folded them down immediately.

config/config_encryption.go
```diff
@@ -40,52 +50,52 @@ func promptForConfigEncryption() (bool, error) {
...
-			break
+		if bytes.Equal(key, conf) {
+			return key, nil
+		}
+		fmt.Println("Keys did not match, please try again.")
+	}
+	return nil, fmt.Errorf("%w: %w", errUserInput, io.EOF)
+}
+
+// getSensitiveInput reads input from stdin, with echo off it stdin is a terminal
```
shazbert, line 82:
```suggestion
// getSensitiveInput reads input from stdin, with echo off if stdin is a terminal
```

cmd/config/config.go
```diff
@@ -1,82 +1,178 @@
...
+	}
+
+	if err := fs.Parse(args); err != nil {
+		fatal(err.Error())
+	}
+
+	if inplace {
+		out = in
+	} else if out == "[in].out" {
+		out = in + ".out"
```
shazbert, line 46: Not a change request, can consider `out = strings.TrimSuffix(in, filepath.Ext(in)) + ".out"` just to .json and remove it completely. I am happy with it stay as default though as well. Can resolve.

config/config.go
```diff
@@ -1530,115 +1464,96 @@ func migrateConfig(configFile, targetDir string) (string, error) {
...
+		}
 	}
 
-	if !ConfirmECS(pref) {
-		// Read unencrypted configuration
-		decoder := json.NewDecoder(reader)
-		c := &Config{}
-		err = decoder.Decode(c)
-		return c, false, err
+	if j, err = versions.Manager.Deploy(context.Background(), j); err != nil {
```
shazbert, line 1515: 🔥

config/config_encryption.go
```diff
@@ -40,52 +50,52 @@ func promptForConfigEncryption() (bool, error) {
...
+func getSensitiveInput(prompt string) (resp []byte, err error) {
+	fmt.Print(prompt)
+	defer fmt.Println()
+	if term.IsTerminal(int(os.Stdin.Fd())) {
+		return term.ReadPassword(int(os.Stdin.Fd()))
```
shazbert, line 87: This might need a bytes.TrimRight(resp, "\r\n") as well?

shazbert (changes requested):
One nit others are commentary. Nice work!

config/config_encryption.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/dbb9635ef

cmd/config/config.go
gbjk: Don't agree with removing the extension.
I'd be more tempted to do `out = prefix + ".out." + suffix`
I'm not enamoured at *all* with having an outfile that isn't STDOUT, or `-edit` not being the default.

If you all gang up on me and make `-edit` the default I'll smile while I make the change 😉

config/config_encryption.go
gbjk: [Docs say not](https://pkg.go.dev/golang.org/x/term#ReadPassword)
> The slice returned does not include the \n.

shazbert (approved):
Ch ACK! Thank you 🖖

cmd/config/config.go
```diff
@@ -1,82 +1,178 @@
...
+	}
+
+	key := []byte(keyStr)
+	var err error
+	switch cmd {
+	case "upgrade":
+		err = upgradeFile(in, out, key)
+	case "decrypt":
```
thrasher-, line 54: On an already upgraded and encrypted config, using the right password decrypts and saves the output as expected:

$ ./config decrypt --in=../../config_example.json
GoCryptoTrader: config-helper tool
Please enter encryption key:
Success! File written to ../../config_example.json.out

However, if I input a wrong password for this. It goes ahead and mangles the file:

./config decrypt --in=../../config_example.json
GoCryptoTrader: config-helper tool
Please enter encryption key: MEOW69420
Success! File written to ../../config_example.json.out

Expectation would be to check if the PW is correct

thrasher- (changes requested):
Tested everything and looks to all be good aside from a basic issue. Great work on this feature!

cmd/config/config.go
```diff
@@ -1,82 +1,178 @@
...
+
+func decryptFile(in string, key []byte) ([]byte, error) {
+	if !config.IsFileEncrypted(in) {
+		return nil, errors.New("file is already decrypted")
+	}
+	outData, err := config.DecryptConfigFile(readFile(in), key)
+	if err != nil {
+		return nil, fmt.Errorf("unable to decrypt config data. Error: %w", err)
+	}
+	if json, err := jsonparser.Set(outData, []byte("-1"), "encryptConfig"); err == nil {
```
thrasher-, line 122:
```suggestion
	if json, err := jsonparser.Set(outData, []byte("-1"), "encryptConfig"); err != nil {
		// err probably means it didn't decrypt, but we don't tell the user that for security
		return nil, err
	} else {
		outData = json
	}
	return outData, nil
```

Will prevent the mangling file issue

cmd/config/config.go
gbjk: I considered this and rejected it for security reasons.
Previously I thought that what we're doing is less prone to brute force decryption cycling.
But the reality is that anyone with the codebase, also with the encrypted file, isn't going to be using GCT directly to bruteforce it.
And showing how we'd decide if it's valid or not is also trivial.

Therefore I'll fix.

cmd/config/config.go
gbjk: Just realised I added a comment to explain my reasoning for not returning the error☝️

cmd/config/config.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/57dab4133

cmd/config/config.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/57dab4133

config/versions/versions_test.go
```diff
@@ -0,0 +1,113 @@
...
+	assert.NotEmpty(t, m.versions)
+
+	m.registerVersion(2, &TestVersion2{})
+	require.Equal(t, 3, len(m.versions), "Must allocate a space for missing version 1")
+	require.NotNil(t, m.versions[2], "Must put Version 2 in the correct slot")
+	require.Nil(t, m.versions[1], "Must leave Version 1 alone")
+
+	m.registerVersion(1, &TestVersion1{})
+	require.Equal(t, 3, len(m.versions), "Must leave len alone when registering out-of-sequence")
+	require.NotNil(t, m.versions[1], "Should put Version 1 in the correct slot")
```
thrasher-, line 94:
```suggestion
	require.NotNil(t, m.versions[1], "Must put Version 1 in the correct slot")
```

config/versions/versions_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	m.registerVersion(1, &TestVersion1{})
+	_, err = m.Deploy(context.Background(), []byte(``))
+	require.ErrorIs(t, err, errVersionIncompatible)
+
+	m = manager{}
+
+	m.registerVersion(0, &Version0{})
+	_, err = m.Deploy(context.Background(), []byte(`not an object`))
+	require.ErrorIs(t, err, jsonparser.KeyPathNotFoundError, "Should throw the correct error trying to add version to bad json")
```
thrasher-, line 27:
```suggestion
	require.ErrorIs(t, err, jsonparser.KeyPathNotFoundError, "Must throw the correct error trying to add version to bad json")
```

config/versions/versions_test.go
```diff
@@ -0,0 +1,113 @@
...
+	require.ErrorIs(t, err, common.ErrSettingField, "Should throw the correct error trying to add version to bad json")
```
thrasher-, line 28:
```suggestion
	require.ErrorIs(t, err, common.ErrSettingField, "Must throw the correct error trying to add version to bad json")
```

config/versions/versions_test.go
```diff
@@ -0,0 +1,113 @@
...
+	require.ErrorContains(t, err, "version", "Should throw the correct error trying to add version to bad json")
```
thrasher-, line 29:
```suggestion
	require.ErrorContains(t, err, "version", "Must throw the correct error trying to add version to bad json")
```

config/versions/versions_test.go
```diff
@@ -0,0 +1,113 @@
...
+
+	_, err = m.Deploy(context.Background(), []byte(`{"version":"not an int"}`))
+	require.ErrorIs(t, err, common.ErrGettingField, "Should throw the correct error trying to get version from bad json")
```
thrasher-, line 32:
```suggestion
	require.ErrorIs(t, err, common.ErrGettingField, "Must throw the correct error trying to get version from bad json")
```

thrasher- (changes requested):
Thanks for making those changes! Just some remaining minor consistency issues and then this looks good to go!

config/versions/versions_test.go
gbjk: Missed the case insensitive versions of this in my new check, soz.

Fixed https://github.com/gbjk/gocryptotrader/commit/ba1a21453

config/versions/versions_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/ba1a21453

config/versions/versions_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/6709cf767

config/versions/versions_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/ba1a21453

config/versions/versions_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/ba1a21453

gbjk:
Rebased and squashed following last fixes, which were cosmetic.

PR #1673, opened 2024-10-10
Config: Version2 - AssetEnabled

gbjk:
This PR moves the AssetEnabled upgrade into a Versioned upgrade, and as a result can remove all the bool pointer logic.
It also upgrades the example and test config to Version2.

## Dependencies
- [ ] #1671 

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


gbjk:
Rejigging naming. Will reopen

PR #1695, opened 2024-10-30
Tests: Fix fail on vendored modules

gbjk:
This fixes looking for LICENSE files in the vendor directory

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/internal/testing/utils/path_test.go
+++ b/internal/testing/utils/path_test.go
@@ -27,6 +27,10 @@ func TestRootPathFromCWD(t *testing.T) {
 		w, _ := filepath.Split(p)
 		w = filepath.Clean(w)
 		if w == r {
+			switch d.Name() {
+			case "vendor", "web":
+				return filepath.SkipDir
+			}
 			return nil
 		}
 		if d.Type().IsRegular() && d.Name() == "LICENSE" {


gloriousCode (approved):
tACK!
`go mod vendor`
before:
```
found an unexpected LICENSE file in a sub-directory: /Users/POPWIF/go/src/github.com/thrasher-corp/gocryptotrader/vendor/dario.cat/mergo/LICENSE
```
after:
```
ok  	github.com/thrasher-corp/gocryptotrader/internal/testing/utils	0.293s
```

thrasher- (changes requested):
Need to also include a fix for `web/`:

                                found an unexpected LICENSE file in a sub-directory: /src/github.com/thrasher-corp/gocryptotrader/web/node_modules/@adobe/css-tools/LICENSE
                Test:           TestRootPathFromCWD

gbjk:
> Need to also include a fix for web/:

Fixed 61f654b2

Considered just respecting .gitignore, but way too much effort, especially given this case would be cascading and ignoring all of web is "the right thing to do".

Will merge down immediately

thrasher- (approved):
tACK, thanks for making those changes!

PR #1604, opened 2024-08-06
Huobi: Add subscription configuration

gbjk:
Just adds configurable subs.

This PR does not fix auth WS usage, which appears broken due to using deprecated v1 paths.
Also doesn't seek to make subs synchronous.

## Stacked Dependencies

- [x] #1579 

## Type of change

- [x] New feature (non-breaking change which adds functionality)

Issue #434, opened 2020-02-01
No Huobi contractual transactions


idoall:
The trading volume of this exchange is also very high in the ranking. When can it be supported?

thrasher-:
We'll be adding support for it soon but as of now there are other priorities to address first

PR #1703, opened 2024-11-06
Huobi: V2 subscription support

gbjk:
* Removes deprecated and non-functional WS endpoints
* Adds V2 endpoints available as described [here](https://www.htx.com/en-us/opend/newApiPages/?id=664) ( Account updates, Order updates, Trades )
* Mock tests for Private subscriptions
* Improves code structure for wsHandleData
* Assertifies tests
* Quite a few test fixes as well
* Adds `testexch.FixtureToDataHandlerWithErrors`

## Stacked Dependencies

- [x] #1604

## Related

- [x] Progresses #434 ( But probably doesn't cover all requirements since this is WS only )

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


gbjk:
Linter fixed; Rebased to resolve conflict

exchanges/huobi/huobi_test.go
```diff
@@ -3003,3 +2015,20 @@ func TestBootstrap(t *testing.T) {
...
+	updatePairsMutex.Lock()
+	defer updatePairsMutex.Unlock()
+
+	testexch.UpdatePairsOnce(tb, h)
+
+	p, err := h.pairFromContractExpiryCode(btccwPair)
+	require.NoError(tb, err, "pairFromContractCode must not error")
+	err = h.CurrencyPairs.EnablePair(asset.Futures, p)
+	require.NoError(tb, common.ExcludeError(err, currency.ErrPairAlreadyEnabled))
+	btcFutureDatedPair = p
```
shazbert, line 2033: [RACE](https://github.com/thrasher-corp/gocryptotrader/actions/runs/12289907293/job/34296163066?pr=1703#step:10:765) 🚗 

Maybe have a global bool above with the mutex so we don't set this value again

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,240 @@ var subscriptionNames = map[string]string{
...
 		}
-		comms <- WsMessage{Raw: resp.Raw, URL: url}
+		ch <- msg.Raw
 	}
 }
 
-// wsReadData receives and passes on websocket messages for processing
-func (h *HUOBI) wsReadData() {
+// wsReadMsgs receives messages from a message funnel and processes them
+func (h *HUOBI) wsReadMsgs(ch chan []byte) {
```
shazbert, line 121: Can remove funnelling and channel

```diff
diff --git a/exchanges/huobi/huobi_websocket.go b/exchanges/huobi/huobi_websocket.go
index 269d72e46..0acd854c9 100644
--- a/exchanges/huobi/huobi_websocket.go
+++ b/exchanges/huobi/huobi_websocket.go
@@ -88,10 +88,8 @@ func (h *HUOBI) WsConnect() error {
        }
        ctx := context.Background()

-       ch := make(chan []byte)
-       h.Websocket.Wg.Add(2)
-       go h.wsReadMsgs(ch)
-       go h.wsFunnelMsgs(h.Websocket.Conn, ch)
+       h.Websocket.Wg.Add(1)
+       go h.wsReadMsgs(h.Websocket.Conn)

        h.Websocket.SetCanUseAuthenticatedEndpoints(false)
        if h.IsWebsocketAuthenticationSupported() {
@@ -99,38 +97,26 @@ func (h *HUOBI) WsConnect() error {
                        return fmt.Errorf("error authenticating websocket: %w", err)
                }
                h.Websocket.Wg.Add(1)
-               go h.wsFunnelMsgs(h.Websocket.AuthConn, ch)
+               go h.wsReadMsgs(h.Websocket.AuthConn)
        }

        return nil
 }

-// wsFunnelMsgs relays messages from a websocket to a channel for central processing
-func (h *HUOBI) wsFunnelMsgs(s stream.Connection, ch chan []byte) {
+// wsReadMsgs reads and processes data from a websocket connection
+func (h *HUOBI) wsReadMsgs(s stream.Connection) {
        defer h.Websocket.Wg.Done()
        for {
                msg := s.ReadMessage()
                if msg.Raw == nil {
                        return
                }
-               ch <- msg.Raw
-       }
-}
-
-// wsReadMsgs receives messages from a message funnel and processes them
-func (h *HUOBI) wsReadMsgs(ch chan []byte) {
-       defer h.Websocket.Wg.Done()
-       for {
-               select {
-               case <-h.Websocket.ShutdownC:
-                       return
-               case msg := <-ch:
-                       if err := h.wsHandleData(msg); err != nil {
-                               h.Websocket.DataHandler <- err
-                       }
+               if err := h.wsHandleData(msg.Raw); err != nil {
+                       h.Websocket.DataHandler <- err
                }
        }
 }
+
 func (h *HUOBI) wsHandleData(respRaw []byte) error {
        for _, op := range []string{wsSubOp, wsUnsubOp} {
                key := op + "bed" // subbed, unsubbed
```

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,240 @@ var subscriptionNames = map[string]string{
...
-			Type:            oType,
-			Side:            oSide,
-			Status:          oStatus,
-			AssetType:       a,
-			LastUpdated:     time.Unix(response.TS*1000, 0),
-			Pair:            p,
+	if ch, err := jsonparser.GetString(respRaw, "ch"); err == nil {
+		s := h.Websocket.GetSubscription(ch)
+		if s == nil {
+			return subscription.ErrNotFound
```
shazbert, line 173: Can we please add the ch var to this return?

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,240 @@ var subscriptionNames = map[string]string{
...
-		p, a, err = h.GetRequestFormattedPairAndAssetType(data[1])
-		if err != nil {
-			return err
-		}
+func (h *HUOBI) wsHandleV2ping(_ string, respRaw []byte) error {
+	ts, err := jsonparser.GetInt(respRaw, "data", "ts")
+	if err != nil {
+		return fmt.Errorf("error getting ts from auth ping: %w", err)
+	}
+	if err := h.Websocket.AuthConn.SendJSONMessage(context.Background(), request.Unset, json.RawMessage(`{"action":"pong","data":{"ts":`+strconv.Itoa(int(ts))+`}}`)); err != nil {
```
shazbert, line 190: This is auth only? Maybe inject the connection into the param if not and upgrade wsHandleData params.

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,240 @@ var subscriptionNames = map[string]string{
...
-	if err != nil {
+func (h *HUOBI) wsHandleCandleMsg(s *subscription.Subscription, respRaw []byte) error {
+	if len(s.Pairs) != 1 {
+		return subscription.ErrNotSinglePair
+	}
+	var c WsKline
+	if err := json.Unmarshal(respRaw, &c); err != nil {
 		return err
 	}
+	h.Websocket.DataHandler <- stream.KlineData{
```
shazbert, line 233: Any chance in a future PR purge stream.KlineData completely and use the kline package type?

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,240 @@ var subscriptionNames = map[string]string{
...
+		Timestamp:  time.UnixMilli(c.Timestamp),
```
shazbert, line 234: Suggestion: Might be able to change `c.Timestamp` type to `types.Time` and across these changes if it can. Will leave that up to you if you want to do that.

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,240 @@ var subscriptionNames = map[string]string{
...
+			AssetType:    s.Asset,
+			CurrencyPair: s.Pairs[0],
+			Timestamp:    time.UnixMilli(t.Tick.Data[i].Timestamp),
+			Amount:       t.Tick.Data[i].Amount,
+			Price:        t.Tick.Data[i].Price,
+			Side:         side,
+			TID:          strconv.FormatFloat(t.Tick.Data[i].TradeID, 'f', -1, 64),
+		})
+	}
+	return trade.AddTradesToBuffer(h.Name, trades...)
```
shazbert, line 276: Future suggestion: Maybe we can have this and above processing in engine websocket routine handler, and return this. There is a bit of toil going on here for just trades.

exchanges/huobi/huobi_websocket.go
```diff
@@ -548,284 +500,232 @@ func (h *HUOBI) generateSubscriptions() (subscription.List, error) {
...
-	values.Set("SignatureVersion", signatureVersion)
-	values.Set("Timestamp", timestamp)
-	host := "api.huobi.pro"
-	payload := fmt.Sprintf("%s\n%s\n%s\n%s",
-		http.MethodGet, host, endpoint, values.Encode())
+	values.Set("accessKey", creds.Key)
+	values.Set("signatureMethod", signatureMethod)
+	values.Set("signatureVersion", signatureVersion)
+	values.Set("timestamp", timestamp)
+	payload := fmt.Sprintf("%s\n%s\n%s\n%s", http.MethodGet, wsSpotHost, wsPrivatePath, values.Encode())
```
shazbert, line 574: concatenate please instead, while you are here 🙏

shazbert (changes requested):
Awesome update. 🍰 Just some minor nits.

exchanges/huobi/huobi_test.go
gbjk: We have a bool type already; if btcFutureDatedPair == currency.EMPTYPAIR.
However for some reason that's not enough.

gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/d176b4911

This turned out to be relatively complicated; I forgot that the shared test utility `updatePairsOnce` will nuke the `CurrencyPairs`. That means that not only was it unsafe to call the local `updatePairsOnce` in huobi_test without re-enabling the (already found) pair, but also anything else that Enables a pair should localise it's exchange instance `h` to avoid getting clobbered. In this case `TestGetLatestFundingRates`

exchanges/huobi/huobi_websocket.go
```diff
@@ -548,284 +500,232 @@ func (h *HUOBI) generateSubscriptions() (subscription.List, error) {
...
+	if op == wsSubOp {
+		s.SetKey(s.QualifiedChannel)
+		if err := h.Websocket.AddSubscriptions(c, s); err != nil {
+			return fmt.Errorf("%w: %s; error: %w", stream.ErrSubscriptionFailure, s, err)
 		}
-		if err == nil {
-			err = h.Websocket.RemoveSubscriptions(h.Websocket.Conn, s)
+	}
+	ctx := context.Background()
+	respRaw, err := c.SendMessageReturnResponse(ctx, request.Unset, wsSubOp+":"+s.QualifiedChannel, req)
```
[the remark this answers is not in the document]
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/ff199aafe

exchanges/huobi/huobi_websocket.go
gbjk: Agreed. Thanks for this! 🙏 

_Aside: I always think `data` is bad in nearly all contexts. In the context of this comment, we know it's a message.
At some point I want to go on a `respRaw` destroying crusade too. And `wsHandleData`. Is there a wsDoThings around as well?_

Fixed https://github.com/gbjk/gocryptotrader/commit/479095c99

exchanges/huobi/huobi_websocket.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/32d533c15

exchanges/huobi/huobi_websocket.go
gbjk: Yes. It's auth only.
Note the error messages if unsure.

Refactored for clarity.

Fixed https://github.com/gbjk/gocryptotrader/commit/a89f6f8f0

exchanges/huobi/huobi_websocket.go
gbjk: Sure.

exchanges/huobi/huobi_websocket.go
gbjk: Ouch. Yeah. Can do.

Fixed https://github.com/gbjk/gocryptotrader/commit/e6f030cb1

exchanges/huobi/huobi_websocket.go
gbjk: Response from Future:
Yeah. I wasn't happy with what I saw in trades in general and have an open item to review it both in Huobi and generally.

exchanges/huobi/huobi_websocket.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/68f571f28

exchanges/huobi/huobi_websocket.go
```diff
@@ -531,15 +335,156 @@ func (h *HUOBI) WsProcessOrderbook(update *WsDepth, symbol string) error {
...
+		Side:            o.Side,
+		AssetType:       s.Asset,
+		Pair:            p,
+	}
+	if o.OrderID != 0 {
+		d.OrderID = strconv.FormatInt(o.OrderID, 10)
+	}
+	switch o.EventType {
+	case "trigger", "deletion", "cancellation":
+		d.LastUpdated = time.Unix(o.LastActTime*1000, 0)
```
shazbert, line 373: This clips the time values I went through and converted the rest over to type.Time for you:

```diff 
diff --git a/exchanges/huobi/huobi_test.go b/exchanges/huobi/huobi_test.go
index e44088354..958452307 100644
--- a/exchanges/huobi/huobi_test.go
+++ b/exchanges/huobi/huobi_test.go
@@ -35,6 +35,7 @@ import (
        testsubs "github.com/thrasher-corp/gocryptotrader/internal/testing/subscriptions"
        mockws "github.com/thrasher-corp/gocryptotrader/internal/testing/websocket"
        "github.com/thrasher-corp/gocryptotrader/portfolio/withdraw"
+       "github.com/thrasher-corp/gocryptotrader/types"
 )

 // Please supply you own test keys here for due diligence testing.
@@ -1404,8 +1405,8 @@ func TestWsAccountUpdate(t *testing.T) {
        close(h.Websocket.DataHandler)
        require.Len(t, h.Websocket.DataHandler, 3, "Must see correct number of records")
        exp := []WsAccountUpdate{
-               {Currency: "btc", AccountID: 123456, Balance: 23.111, ChangeType: "transfer", AccountType: "trade", ChangeTime: 1568601800000, SeqNum: 1},
-               {Currency: "btc", AccountID: 33385, Available: 2028.69, ChangeType: "order.match", AccountType: "trade", ChangeTime: 1574393385167, SeqNum: 2},
+               {Currency: "btc", AccountID: 123456, Balance: 23.111, ChangeType: "transfer", AccountType: "trade", ChangeTime: types.Time(time.UnixMilli(1568601800000)), SeqNum: 1},
+               {Currency: "btc", AccountID: 33385, Available: 2028.69, ChangeType: "order.match", AccountType: "trade", ChangeTime: types.Time(time.UnixMilli(1574393385167)), SeqNum: 2},
                {Currency: "usdt", AccountID: 14884859, Available: 20.29388158, Balance: 20.29388158, AccountType: "trade", SeqNum: 3},
        }
        for _, e := range exp {
@@ -1436,7 +1437,7 @@ func TestWsOrderUpdate(t *testing.T) {
                        Status:        order.Rejected,
                        ClientOrderID: "test1",
                        AssetType:     asset.Spot,
-                       LastUpdated:   time.Unix(1583853365586000, 0),
+                       LastUpdated:   time.UnixMicro(1583853365586000),
                },
                {
                        Exchange:      h.Name,
@@ -1445,7 +1446,7 @@ func TestWsOrderUpdate(t *testing.T) {
                        Status:        order.Cancelled,
                        ClientOrderID: "test2",
                        AssetType:     asset.Spot,
-                       LastUpdated:   time.Unix(1583853365586000, 0),
+                       LastUpdated:   time.UnixMicro(1583853365586000),
                },
                {
                        Exchange:      h.Name,
@@ -1458,7 +1459,7 @@ func TestWsOrderUpdate(t *testing.T) {
                        Amount:        2,
                        Type:          order.Limit,
                        OrderID:       "27163533",
-                       LastUpdated:   time.Unix(1583853365586000, 0),
+                       LastUpdated:   time.UnixMicro(1583853365586000),
                },
                {
                        Exchange:    h.Name,
@@ -1470,7 +1471,7 @@ func TestWsOrderUpdate(t *testing.T) {
                        Amount:      0.000157,
                        Type:        order.Limit,
                        OrderID:     "1199329381585359",
-                       LastUpdated: time.Unix(1731039387696000, 0),
+                       LastUpdated: time.UnixMicro(1731039387696000),
                },
        }
        for _, e := range exp {
@@ -1500,8 +1501,8 @@ func TestWsMyTrades(t *testing.T) {
                ClientOrderID: "a001",
                OrderID:       "99998888",
                AssetType:     asset.Spot,
-               Date:          time.Unix(1583853365586000, 0),
-               LastUpdated:   time.Unix(1583853365996000, 0),
+               Date:          time.UnixMicro(1583853365586000),
+               LastUpdated:   time.UnixMicro(1583853365996000),
                Price:         10000,
                Amount:        1,
                Trades: []order.TradeHistory{
@@ -1513,7 +1514,7 @@ func TestWsMyTrades(t *testing.T) {
                                TID:       "919219323232",
                                Side:      order.Buy,
                                IsMaker:   false,
-                               Timestamp: time.Unix(1583853365996000, 0),
+                               Timestamp: time.UnixMicro(1583853365996000),
                        },
                },
        }
@@ -2037,7 +2038,7 @@ func updatePairsOnce(tb testing.TB, h *HUOBI) {

        testexch.UpdatePairsOnce(tb, h)

-       if btcFutureDatedPair == currency.EMPTYPAIR {
+       if btcFutureDatedPair.Equal(currency.EMPTYPAIR) {
                p, err := h.pairFromContractExpiryCode(btccwPair)
                require.NoError(tb, err, "pairFromContractCode must not error")
                btcFutureDatedPair = p
diff --git a/exchanges/huobi/huobi_types.go b/exchanges/huobi/huobi_types.go
index b460f912e..80a04aaa1 100644
--- a/exchanges/huobi/huobi_types.go
+++ b/exchanges/huobi/huobi_types.go
@@ -905,14 +905,14 @@ type wsAccountUpdateMsg struct {

 // WsAccountUpdate contains account updates to balances
 type WsAccountUpdate struct {
-       Currency    string  `json:"currency"`
-       AccountID   int64   `json:"accountId"`
-       Balance     float64 `json:"balance,string"`
-       Available   float64 `json:"available,string"`
-       ChangeType  string  `json:"changeType"`
-       AccountType string  `json:"accountType"`
-       ChangeTime  int64   `json:"changeTime"`
-       SeqNum      int64   `json:"seqNum"`
+       Currency    string     `json:"currency"`
+       AccountID   int64      `json:"accountId"`
+       Balance     float64    `json:"balance,string"`
+       Available   float64    `json:"available,string"`
+       ChangeType  string     `json:"changeType"`
+       AccountType string     `json:"accountType"`
+       ChangeTime  types.Time `json:"changeTime"`
+       SeqNum      int64      `json:"seqNum"`
 }

 type wsOrderUpdateMsg struct {
@@ -939,9 +939,9 @@ type WsOrderUpdate struct {
        IsTaker         bool       `json:"aggressor"`
        Side            order.Side `json:"orderSide"`
        OrderStatus     string     `json:"orderStatus"`
-       LastActTime     int64      `json:"lastActTime"`
-       CreateTime      int64      `json:"orderCreateTime"`
-       TradeTime       int64      `json:"tradeTime"`
+       LastActTime     types.Time `json:"lastActTime"`
+       CreateTime      types.Time `json:"orderCreateTime"`
+       TradeTime       types.Time `json:"tradeTime"`
        ErrCode         int64      `json:"errCode"`
        ErrMessage      string     `json:"errMessage"`
 }
@@ -961,7 +961,7 @@ type WsTradeUpdate struct {
        OrderType       string     `json:"orderType"`
        IsTaker         bool       `json:"aggressor"`
        TradeID         int64      `json:"tradeId"`
-       TradeTime       int64      `json:"tradeTime"`
+       TradeTime       types.Time `json:"tradeTime"`
        TransactFee     float64    `json:"transactFee,string"`
        FeeCurrency     string     `json:"feeCurrency"`
        FeeDeduct       string     `json:"feeDeduct"`
@@ -974,7 +974,7 @@ type WsTradeUpdate struct {
        ClientOrderID   string     `json:"clientOrderId"`
        StopPrice       string     `json:"stopPrice"`
        Operator        string     `json:"operator"`
-       OrderCreateTime int64      `json:"orderCreateTime"`
+       OrderCreateTime types.Time `json:"orderCreateTime"`
        OrderStatus     string     `json:"orderStatus"`
 }

@@ -1169,6 +1169,6 @@ type WithdrawalData struct {
        State           string        `json:"state"`
        ErrorCode       string        `json:"error-code"`
        ErrorMessage    string        `json:"error-message"`
-       CreatedAt       int64         `json:"created-at"`
-       UpdatedAt       int64         `json:"updated-at"`
+       CreatedAt       types.Time    `json:"created-at"`
+       UpdatedAt       types.Time    `json:"updated-at"`
 }
diff --git a/exchanges/huobi/huobi_websocket.go b/exchanges/huobi/huobi_websocket.go
index 0df94c208..44611144b 100644
--- a/exchanges/huobi/huobi_websocket.go
+++ b/exchanges/huobi/huobi_websocket.go
@@ -370,11 +370,11 @@ func (h *HUOBI) wsHandleMyOrdersMsg(s *subscription.Subscription, respRaw []byte
        }
        switch o.EventType {
        case "trigger", "deletion", "cancellation":
-               d.LastUpdated = time.Unix(o.LastActTime*1000, 0)
+               d.LastUpdated = o.LastActTime.Time()
        case "creation":
-               d.LastUpdated = time.Unix(o.CreateTime*1000, 0)
+               d.LastUpdated = o.CreateTime.Time()
        case "trade":
-               d.LastUpdated = time.Unix(o.TradeTime*1000, 0)
+               d.LastUpdated = o.TradeTime.Time()
        }
        if d.Status, err = order.StringToOrderStatus(o.OrderStatus); err != nil {
                return &order.ClassificationError{
@@ -428,8 +428,8 @@ func (h *HUOBI) wsHandleMyTradesMsg(s *subscription.Subscription, respRaw []byte
                Side:          t.Side,
                AssetType:     s.Asset,
                Pair:          p,
-               Date:          time.Unix(t.OrderCreateTime*1000, 0),
-               LastUpdated:   time.Unix(t.TradeTime*1000, 0),
+               Date:          t.OrderCreateTime.Time(),
+               LastUpdated:   t.TradeTime.Time(),
                OrderID:       strconv.FormatInt(t.OrderID, 10),
        }
        if d.Status, err = order.StringToOrderStatus(t.OrderStatus); err != nil {
@@ -469,7 +469,7 @@ func (h *HUOBI) wsHandleMyTradesMsg(s *subscription.Subscription, respRaw []byte
                        Type:      d.Type,
                        Side:      d.Side,
                        IsMaker:   !t.IsTaker,
-                       Timestamp: time.Unix(t.TradeTime*1000, 0),
+                       Timestamp: t.TradeTime.Time(),
                },
        }
        h.Websocket.DataHandler <- d
diff --git a/exchanges/huobi/huobi_wrapper.go b/exchanges/huobi/huobi_wrapper.go
index 8ff734c4c..a27a16e79 100644
--- a/exchanges/huobi/huobi_wrapper.go
+++ b/exchanges/huobi/huobi_wrapper.go
@@ -877,7 +877,7 @@ func (h *HUOBI) GetWithdrawalsHistory(ctx context.Context, c currency.Code, a as
                resp[i] = exchange.WithdrawalHistory{
                        Status:          withdrawals.Data[i].State,
                        TransferID:      withdrawals.Data[i].TransactionHash,
-                       Timestamp:       time.UnixMilli(withdrawals.Data[i].CreatedAt),
+                       Timestamp:       withdrawals.Data[i].CreatedAt.Time(),
                        Currency:        withdrawals.Data[i].Currency.String(),
                        Amount:          withdrawals.Data[i].Amount,
                        Fee:             withdrawals.Data[i].Fee,
```

shazbert (changes requested):
Nice work! Thanks for addressing my initial nits, just one left over issue then looks good to go, good stuff! 🚀

exchanges/huobi/huobi_websocket.go
shazbert: Just realised I included `if btcFutureDatedPair.Equal(currency.EMPTYPAIR) {` you don't have to change that

gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/4d0a56c0b

exchanges/huobi/huobi_wrapper.go
```diff
@@ -176,7 +176,7 @@ func (h *HUOBI) SetDefaults() {
 		exchange.RestSpot:         huobiAPIURL,
```
shazbert, line 176: Could you please add as it takes a while for initial sync to finish?

```diff
diff --git a/exchanges/huobi/huobi_wrapper.go b/exchanges/huobi/huobi_wrapper.go
index a27a16e79..9f0f8500e 100644
--- a/exchanges/huobi/huobi_wrapper.go
+++ b/exchanges/huobi/huobi_wrapper.go
@@ -80,6 +80,16 @@ func (h *HUOBI) SetDefaults() {
                log.Errorln(log.ExchangeSys, err)
        }

+       err = h.DisableAssetWebsocketSupport(asset.Futures)
+       if err != nil {
+               log.Errorln(log.ExchangeSys, err)
+       }
+
+       err = h.DisableAssetWebsocketSupport(asset.CoinMarginedFutures)
+       if err != nil {
+               log.Errorln(log.ExchangeSys, err)
+       }
+
        h.Features = exchange.Features{
                Supports: exchange.FeaturesSupported{
                        REST:      true,
```

shazbert (changes requested):
🕊️ Last thing from me, I hope 🙏

exchanges/huobi/huobi_wrapper.go
gbjk: Note: I corrected all of these somewhere recently in #1735 - That error would not have enough context :) 

Fixed https://github.com/gbjk/gocryptotrader/commit/fde0ad954

shazbert (commented):
unauth tACK unauth utACK. Thanks!

shazbert (approved):

shazbert (changes requested):
[Whoops tests broke](https://github.com/thrasher-corp/gocryptotrader/actions/runs/12366183410/job/34512443787?pr=1703#step:10:1086)

exchanges/huobi/huobi_websocket.go
```diff
@@ -531,15 +335,156 @@ func (h *HUOBI) WsProcessOrderbook(update *WsDepth, symbol string) error {
...
+			}
+		}
+	}
+	d.Trades = []order.TradeHistory{
+		{
+			Price:     t.TradePrice,
+			Amount:    t.TradeVolume,
+			Fee:       t.TransactFee,
+			Exchange:  h.Name,
+			TID:       strconv.Itoa(int(t.TradeID)),
```
gbjk: 🚨 This is not safe for 32-bit

gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/c942589e0

gbjk:
> [Whoops tests broke](https://github.com/thrasher-corp/gocryptotrader/actions/runs/12366183410/job/34512443787?pr=1703#step:10:1086)

This has been an transient test failure for ages, impervious to introspection.
Think I finally cracked it with:

Fixed https://github.com/gbjk/gocryptotrader/commit/52dfbfd8f

but running long `-count=100 -race` again to be (somewhat) sure.

gbjk:
CI failures all have 5 or 6 failures for just Coinbase and/or test server.
All clean for Huobi.

Also nothing locally for hundreds of runs.

shazbert (approved):
Changes ACK, Thanks 🫡

exchanges/huobi/huobi_test.go
```diff
@@ -1981,489 +1223,305 @@ func TestCancelAllExchangeOrders(t *testing.T) {
...
+	_, err := h.QueryWithdrawQuotas(context.Background(), currency.BTC.Lower().String())
+	if sharedtestvalues.AreAPICredentialsSet(h) {
+		require.NoError(t, err)
+	} else {
+		require.ErrorIs(t, err, exchange.ErrCredentialsAreEmpty)
 	}
 }
 
+func TestWsCandles(t *testing.T) {
+	h := new(HUOBI) //nolint:govet // Intentional shadow to avoid future copy/paste mistakes
```
gloriousCode, line 1306: Any reason these can't be parallel given that they are shadowed and use their own Json files?

exchanges/huobi/huobi_test.go
```diff
@@ -2631,27 +1621,23 @@ func TestGetAvailableTransferChains(t *testing.T) {
...
-	// Test for upper case 'BTC' not lower case 'btc', disregarded numerals
-	// as they not deterministic from this endpoint.
-	assert.Equal(t, "BTC", r[0:3])
-
-	r, err = h.formatFuturesPair(futuresTestPair, true)
+	r, err = h.formatFuturesPair(btccwPair, true)
 	require.NoError(t, err)
 	assert.Len(t, r, 9, "Should be an 9 character string")
-	assert.Equal(t, "BTC", r[0:3])
+	assert.Equal(t, "BTC2", r[0:4], "Should start with btc and a date this millennium")
```
gloriousCode, line 1640: You are responsible for changing this in the year 3000

exchanges/huobi/huobi_websocket.go
```diff
@@ -12,6 +12,7 @@ import (
 	"text/template"
```
gloriousCode, line 12: PSA for any other reviews with old fields like me. If you have a subscription called `account` its changed to `myAccount` and this will fail. Just delete all subs and they'll be re-added

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,233 @@ var subscriptionNames = map[string]string{
...
-func (h *HUOBI) wsAuthenticatedDial(dialer *websocket.Dialer) error {
-	if !h.IsWebsocketAuthenticationSupported() {
-		return fmt.Errorf("%v AuthenticatedWebsocketAPISupport not enabled",
-			h.Name)
-	}
-	err := h.Websocket.AuthConn.Dial(dialer, http.Header{})
-	if err != nil {
-		return err
+	h.Websocket.SetCanUseAuthenticatedEndpoints(false)
+	if h.IsWebsocketAuthenticationSupported() {
```
gloriousCode, line 95: ?
I don't see any other exchange behave in this manner

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,233 @@ var subscriptionNames = map[string]string{
...
+func (h *HUOBI) wsHandleData(respRaw []byte) error {
+	for _, op := range []string{wsSubOp, wsUnsubOp} {
+		key := op + "bed" // subbed, unsubbed
+		if ch, err := jsonparser.GetString(respRaw, key); err == nil {
+			if !h.Websocket.Match.IncomingWithData(op+":"+ch, respRaw) {
+				return fmt.Errorf("%w: %s:%s", stream.ErrNoMessageListener, op, ch)
 			}
+			return nil
 		}
 	}
```
gloriousCode, line 129: Perhaps there's something I've missed, but I don't see the point to this loop when you have a handler for it that performs the same function below. It also doesn't force every single message on the ws into a loop - costing about 90ns per message

```
case wsSubOp, wsUnsubOp:
			return h.wsHandleV2subResp(action, respRaw)
```

gloriousCode (changes requested):
Nicely done

exchanges/huobi/huobi_test.go
```diff
@@ -1981,489 +1223,305 @@ func TestCancelAllExchangeOrders(t *testing.T) {
...
-	sharedtestvalues.SkipTestIfCredentialsUnset(t, h)
-
-	_, err := h.UpdateAccountInfo(context.Background(), asset.Spot)
-	if err != nil {
-		// Spot and Futures have separate api keys. Please ensure that the correct keys are provided
-		t.Error(err)
+	sharedtestvalues.SkipTestIfCredentialsUnset(t, h, canManipulateRealOrders)
+	for _, a := range []asset.Item{asset.Spot, asset.CoinMarginedFutures, asset.Futures} {
+		_, err := h.UpdateAccountInfo(context.Background(), a)
+		require.NoErrorf(t, err, "UpdateAccountInfo must not error for asset %s; Spot and Futures have separate api keys. Please ensure that the correct keys are provided", a)
```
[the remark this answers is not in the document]
gbjk: Sure. That's something I didn't touch, so if you say it's no longer relevant I'll Make-It-So.
 
Fixed https://github.com/gbjk/gocryptotrader/commit/bff1ad3a5

exchanges/huobi/huobi_test.go
```diff
@@ -1981,489 +1223,305 @@ func TestCancelAllExchangeOrders(t *testing.T) {
...
+	t.Parallel()
+	_, err := h.QueryWithdrawQuotas(context.Background(), currency.BTC.Lower().String())
+	if sharedtestvalues.AreAPICredentialsSet(h) {
+		require.NoError(t, err)
+	} else {
+		require.ErrorIs(t, err, exchange.ErrCredentialsAreEmpty)
 	}
 }
 
+func TestWsCandles(t *testing.T) {
```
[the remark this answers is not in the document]
gbjk: Agreed. I only really internalised that recently.

Fixed https://github.com/gbjk/gocryptotrader/commit/c2984af50

exchanges/huobi/huobi_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/58ecd3c1b

exchanges/huobi/huobi_websocket.go
gbjk: In which sense?
Most of them use `SetCanUseAuthenticatedEndpoints` false on failure (Kraken, Okx)
Many of them use `IsWebsocketAuthenticationSupported` to choose to dial auth (Kraken, Okx)

This is using a different pattern, but I'm actually happier with it.
Set CanUseAuthEndpoints to false on a (re)connect, and then true if the auth succeeds.
It feels cleaner than having the branching logic we've seen elsewhere.
Definitely cleaner than using `CanUseAuthenticatedEndpoints` to decide to even dial, which I saw on one exchange, and feels definitely wrong.

Is there a downside to this that I'm missing?

If anything, I'd follow this pattern with others when I'm interfering with them.

exchanges/huobi/huobi_websocket.go
gbjk: It's the V1 sub support, so the section you highlighted isn't relevant.
However it's also entirely superfluous because the id is in there and it's working, so ... remove this section and everything is still fine.

Fixed https://github.com/gbjk/gocryptotrader/commit/6938c5573

exchanges/huobi/huobi_websocket.go
gloriousCode: Why set something to false when you don't even know if they have credentials set or if its going to be used? 
If you put the `h.Websocket.SetCanUseAuthenticatedEndpoints(false)` inside the if statement, it no longer looks out of place and that change will still work with all the reasoning you outlined

gbjk: > Why set something to false when you don't even know if they have credentials set or if its going to be used?
Because if we aren't about to go into the `if` statement, for any reason at all, then `false` is the correct state for the value. Not setting it to false seems wrong to me; Doing it inside an if statement that _could_ be false and not run seems wrong to me.

I don't know what the use-cases might be. Feels like the user would have to unset credentials for it to need to be set to false where I have it. But that's kind of irrelevant. My point is that it's the best representation of the state at the right time.

gbjk: Kinda afraid you're going to ask me to change all the others if I convince you 🤦

gloriousCode: After 30 days, I still don't like it, nor would I request you put it anywhere else. I didn't expect to talk this much about a bool placement which doesn't affect performance. Its just doing something before even knowing its necessary.

gloriousCode (changes requested):
Just needs a rebase, thank you! 🌈

gbjk:
Rebase required an additional fix: https://github.com/gbjk/gocryptotrader/commit/65e81e470

Immediately folded that down and pushed.

gloriousCode (approved):
tACK

exchanges/huobi/huobi_websocket.go
thrasher-: I too don't see the point of this, we should only SetCanUseAuthenticatedEndpoints(false) if auth fails. We'll also always default to off unless a user has provided auth credentials and enabled websocket, which we'll test upon engine starting up

gbjk: Okay; I'll change it.

I'm still sticking to my view that `WsConnect` is the point at which we have *lost* whatever possible authenticated state we had before, not on failed auth response.
( _Arguably we lost it on disconnect_ )
So if reconnect and we don't go into `h.IsWebsocketAuthenticationSupported` for literally any reason now, or in the future, we've left the bool in the wrong state. I know that's an edge case. I just don't see the point in not coding for it.

Fixup coming.

gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/53425ab42

Moved the set to true up into WsConnect because that matches other exchanges (Kraken), and means setting true and false is in the same place.

internal/testing/exchange/exchange.go
```diff
@@ -118,23 +118,45 @@ func MockWsInstance[T any, PT interface {
...
+}
+
+// FixtureToDataHandlerWithErrors squirts the contents of a file to a reader function (probably e.wsHandleData) and returns handler errors
+// Any errors setting up the fixture will fail tests
+func FixtureToDataHandlerWithErrors(tb testing.TB, fixturePath string, reader func([]byte) error) []FixtureError {
+	tb.Helper()
+
 	fixture, err := os.Open(fixturePath)
-	assert.NoError(tb, err, "Opening fixture '%s' should not error", fixturePath)
+	require.NoError(tb, err, "Opening fixture '%s' should not error", fixturePath)
```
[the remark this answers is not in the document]
gbjk: Fixed f3b305e6f

exchanges/huobi/huobi_websocket.go
```diff
@@ -75,427 +73,222 @@ var subscriptionNames = map[string]string{
...
 	if !h.Websocket.IsEnabled() || !h.IsEnabled() {
 		return stream.ErrWebsocketNotEnabled
 	}
-	var dialer websocket.Dialer
-	err := h.wsDial(&dialer)
-	if err != nil {
+	if err := h.Websocket.Conn.Dial(&websocket.Dialer{}, http.Header{}); err != nil {
 		return err
 	}
+	ctx := context.Background()
```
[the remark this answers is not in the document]
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/95a9f51d2

thrasher- (approved):

PR #1707, opened 2024-11-08
Huobi: Fix invalid contract type error when autoUpdatePairs is false

gbjk:
We were relying on autoUpdatePairs to fetch contract types before we need them. If the user disables it in config, we'll lazily fetch just the relevant Futures.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/huobi/huobi_wrapper.go
+++ b/exchanges/huobi/huobi_wrapper.go
@@ -187,6 +187,17 @@ func (h *HUOBI) SetDefaults() {
 	h.WebsocketOrderbookBufferLimit = exchange.DefaultWebsocketOrderbookBufferLimit
 }
 
+// Bootstrap ensures that future contract expiry codes are loaded if AutoPairUpdates is not enabled
+func (h *HUOBI) Bootstrap(_ context.Context) (continueBootstrap bool, err error) {
+	continueBootstrap = true
+
+	if !h.GetEnabledFeatures().AutoPairUpdates && h.SupportsAsset(asset.Futures) {
+		_, err = h.FetchTradablePairs(context.Background(), asset.Futures)
+	}
+
+	return
+}
+
 // Setup sets user configuration
 func (h *HUOBI) Setup(exch *config.Exchange) error {
 	err := exch.Validate()
--- a/exchanges/huobi/huobi_test.go
+++ b/exchanges/huobi/huobi_test.go
@@ -2991,3 +2991,19 @@ func TestChannelName(t *testing.T) {
 	assert.Panics(t, func() { channelName(&subscription.Subscription{Channel: wsOrderbookChannel}, p) })
 	assert.Panics(t, func() { channelName(&subscription.Subscription{Channel: subscription.MyAccountChannel}, p) }, "Should panic on V2 endpoints until implemented")
 }
+
+func TestBootstrap(t *testing.T) {
+	t.Parallel()
+	h := new(HUOBI)
+	require.NoError(t, testexch.Setup(h), "Test Instance Setup must not fail")
+
+	c, err := h.Bootstrap(context.Background())
+	require.NoError(t, err)
+	assert.True(t, c, "Bootstrap should return true to continue")
+
+	h.futureContractCodes = nil
+	h.Features.Enabled.AutoPairUpdates = false
+	_, err = h.Bootstrap(context.Background())
+	require.NoError(t, err)
+	require.NotNil(t, h.futureContractCodes)
+}


exchanges/huobi/huobi_futures.go
```diff
@@ -1249,6 +1249,11 @@ func (h *HUOBI) formatFuturesPair(p currency.Pair, convertQuoteToExpiry bool) (s
 // We need this because some apis, such as ticker, use BTC_CW, NW, CQ, NQ
 // Other apis, such as contract_info, use contract type of this_week, next_week, quarter (sic), and next_quater
 func (h *HUOBI) pairFromContractExpiryCode(p currency.Pair) (currency.Pair, error) {
+	if h.futureContractCodes == nil { // Worst race-case here is that we fetch twice. No avoiding it.
```
gloriousCode, line 1252: A nil check is not race proof
```
WARNING: DATA RACE
Read at 0x0001014ee5a0 by goroutine 238:
  github.com/thrasher-corp/gocryptotrader/exchanges/huobi.(*HUOBI).pairFromContractExpiryCode()
      /Users/GRASS/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/huobi/huobi_futures.go:1252 +0x64
  github.com/thrasher-corp/gocryptotrader/exchanges/huobi.TestButts.func1()
      /Users/GRASS/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/huobi/huobi_test.go:2830 +0x11c
  github.com/thrasher-corp/gocryptotrader/exchanges/huobi.TestButts.gowrap1()
      /Users/GRASS/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/huobi/huobi_test.go:2834 +0x44

Previous write at 0x0001014ee5a0 by goroutine 239:
  github.com/thrasher-corp/gocryptotrader/exchanges/huobi.(*HUOBI).FetchTradablePairs()
      /Users/GRASS/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/huobi/huobi_wrapper.go:292 +0x1ec
  github.com/thrasher-corp/gocryptotrader/exchanges/huobi.TestButts.func2()
      /Users/GRASS/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/huobi/huobi_test.go:2837 +0x8c
  github.com/thrasher-corp/gocryptotrader/exchanges/huobi.TestButts.gowrap2()
      /Users/GRASS/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/huobi/huobi_test.go:2838 +0x44
```
You can argue that me calling both this and `FetchTradablePairs` in go routines to prove this point isn't the best behaviour, but I'd say the same about not synchronising tradable pairs for futures with expiring contracts 😄

gloriousCode (changes requested):

exchanges/huobi/huobi_futures.go
gbjk: Nit pick: Race proof versus Race detected are not the same, though, right?
I think that a nil check is race proof and atomic.
The access isn't synchronised, but it is atomic.

Suppose we wrap a read lock around this nil check.
We then have to unlock the read lock to call fetchPairs. That'd be more of a race opportunity there, than just straight-lining from the nil check into the fetch call.

All of that said, if this has the potential to trigger the race detector I'll work a way around it somehow.

gbjk: Fixed 1ea8fd4df ( with extreme prejudice 😈 )

exchanges/huobi/huobi_wrapper.go
```diff
@@ -186,6 +186,17 @@ func (h *HUOBI) SetDefaults() {
 	h.WebsocketOrderbookBufferLimit = exchange.DefaultWebsocketOrderbookBufferLimit
 }
 
+// Bootstrap ensures that future contract expiry codes are loaded if AutoPairUpdates is not enabled
+func (h *HUOBI) Bootstrap(_ context.Context) (continueBootstrap bool, err error) {
```
gloriousCode, line 190: A much cleaner solution, thank you!
Would you consider adding this?
```go
func TestBootstrap(t *testing.T) {
	t.Parallel()
	h := new(HUOBI) //nolint:govet // Intentional shadow to avoid future copy/paste mistakes
	require.NoError(t, testexch.Setup(h), "Test Instance Setup must not fail")

	_, err := h.Bootstrap(context.Background())
	require.NoError(t, err)

	h.futureContractCodes = nil
	h.Features.Enabled.AutoPairUpdates = false
	_, err = h.Bootstrap(context.Background())
	require.NoError(t, err)
	require.NotNil(t, h.futureContractCodes)
}

```

gbjk: Fixed [f003ce6](https://github.com/gbjk/gocryptotrader/commit/f003ce690d0da9f05ae1f6ea591bfcafd199a878)

Immediately rebased down.

gbjk: Two lint niggles fixed immediately after this. Still no idea when and when govet doesn't like these shadows. It seems completely arbitrary.

gloriousCode (approved):
tACK!

shazbert (approved):
tACK please rebase. ⭐

gbjk:
> tACK please rebase. ⭐

Done.

thrasher- (approved):
Thanks for resolving this + @gloriousCode for test coverage C+P

PR #1717, opened 2024-11-18
stream: force subscription store check as stop gap for wrapper side implementation

shazbert:
# PR Description

Noticed in Bybit that subscriptions weren't being added when I was flushing the connection every hour, ~~so abstracted adding and removing to websocket package then I can merge this to the open [PR chain](https://github.com/thrasher-corp/gocryptotrader/pull/1670). This is so we don't need to toil in the wrappers as well but there might be some edge cases I am not considering.~~ 

~~This only impacts exchanges that are upgraded to the websocket multi connection.~~

Added in checks in stream package so that the outbound subscriptions are added or removed from store.

Fixes bybit and Deribit

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


gbjk:
I have concerns about building this out further:
* Some exchanges set PendingState and rely upon it
* Some exchanges add temporary subs until they have a better known sub key
* Fan out mechanics mean that sometimes the subs we add are't the ones we expected to

exchanges/stream/websocket.go
```diff
@@ -434,6 +434,13 @@ func (w *Websocket) connect() error {
 			break
 		}
 
+		if len(subs) != 0 {
+			if err = w.AddSuccessfulSubscriptions(conn, subs...); err != nil {
+				multiConnectFatalError = fmt.Errorf("%v Error adding successful subscriptions %w", w.exchangeName, err)
+				break
+			}
+		}
+
```
gbjk, line 443: I don't think we can do this all-or-nothing approach.
I'm not sure we should bail on connect in *either* situation (multi or not), because failing to subscribe does not mean failing to connect. We should probably log the error and continue on.

But in this instance, if we managed to subscribe to 4 subs out of 5, we don't want to have not recorded that we subscribed to them.

As Subscriber is right now, I don't think we can delay Adding until here. It has to be in the manageSubs with more granular visibility.

This feels like it's raised a new issue for me of "connect shouldn't error if some subs fail", but that's separate.

gbjk (changes requested):
Given comment, I don't see a way to move this forward right now.
I feel like the manageSubs is the only place we can properly manage sub state.
One edge case in the back of my mind is resubscribing, as well, particularly on orderbook fail.

exchanges/stream/websocket.go
shazbert: Yeah I differ in perspective on that, I regard a failure on 1 is a failure on the entire set. Especially when trading I want to ensure when I connect its subscribed to all the required subscriptions I want. 

If there is a failure on subscription its either a malformed outbound payload or the response from the connection is delayed in both cases the connection is not ideal for trading ops. A retry under a for loop is easier and less complicated after handling error from `connect` if it is a subscription delay error.

shazbert:
These are all fair points, an option I will investigate is check the connection subscription status after subscriptions to see if they have been added or removed. So that at least it complains that it needs to be added in the exchange packages themselves.

exchanges/stream/websocket.go
gbjk: Feel free to return error from connect.
I don't agree with it, but sure.

_Aside: The best solution to my problem with connect returning error is to not have connect do other things, and have `Run` call connect, then auth, then subscribe. Clearer responsibilities for methods.
I really can't get onboard with calling a method called connect, it connects, then does something else and returns an error about that other thing._

But for today, my problem is: Subs succeeded, but you haven't added any of them as successful.

If you want to return error from connect on sub failure, please ensure the successful ones are added first, then return error.

shazbert: > Aside: The best solution to my problem with connect returning error is to not have connect do other things, and have Run call connect, then auth, then subscribe. Clearer responsibilities for methods.
I really can't get onboard with calling a method called connect, it connects, then does something else and returns an error about that other thing.

Well no the current design differs slightly from  `connect -> auth -> subscribe` so we retrieve subscriptions -> only if subscriptions are available attempt connection/connect -> authenticate connection if available -> then subscribe (auth/unauth subs). We don't need to spawn connection/connect when there are no subscriptions, there is a caveat here for when this is allowed for a dedicated outbound connection for requests. This allows us to scale in the event of connection only allowing N subs per connection in future as well (OKX). Connect() means nothing without subscriptions. I guess we could rename to something different? 

> But for today, my problem is: Subs succeeded, but you haven't added any of them as successful.
If you want to return error from connect on sub failure, please ensure the successful ones are added first, then return error.

I don't know what you mean by this sorry? In regards to the current state of the PR I have added in `AddSuccessfulSubscriptions` but we are not going to do this any way? If this is directly related to `I will investigate is check the connection subscription status after subscriptions to see if they have been added or removed. So that at least it complains that it needs to be added in the exchange packages themselves.` If there are issues yeah I will address that for this. 

I just want to steer this into a less complicated scenario for exchange implementations, pulling everything apart and toiling on things like this, seems like a good to have instead of a need to have. Everything you have said on your concerns has been on point. I am just trying to give a different perspective.

gbjk: > Connect() means nothing without subscriptions.

We have non-subscription ws methods sometimes, right?
Or do you mean purely in the context of GateIO?

gbjk: > I don't know what you mean by this sorry?

If this PR merges as it currently is, it means that subs which succeeded are not added as successful, if any failed.

Previously they were added as they succeeded by the exchange, one by one.
Now, if one of them fails, then L434 `break` means none of them is added, and that's wrong.

I don't know how you can do this outside of the exchange implementation, because it's so varied when and how we decide it's successful.

gbjk:
@shazbert Just realised that I haven't mentioned:

I completely agree about abstracting from exchange implementations. I'm just not yet ready to see how to do it.
My intention was to finish merging all the exchange sub changes, then handle how multi-asset sub confs should work (which is GateIO + multi-sockets + assets ), and then review what hinge points stop us moving stuff up.

I've been looking to abstract this, but I don't think we're ready yet because too much is in flight and unmerged.
My hope is that they all get ParallelChanOp and GenerateSubs for free, at least, and then maybe some more.

shazbert:
> @shazbert Just realised that I haven't mentioned:
> 
> I completely agree about abstracting from exchange implementations. I'm just not yet ready to see how to do it. My intention was to finish merging all the exchange sub changes, then handle how multi-asset sub confs should work (which is GateIO + multi-sockets + assets ), and then review what hinge points stop us moving stuff up.
> 
> I've been looking to abstract this, but I don't think we're ready yet because too much is in flight and unmerged. My hope is that they all get ParallelChanOp and GenerateSubs for free, at least, and then maybe some more.

@gbjk I have added in new code now with fixes, when you have time could you please check them out.

exchanges/stream/websocket.go
```diff
@@ -683,6 +679,31 @@ func (w *Websocket) FlushChannels() error {
...
+// updateChannelSubscriptions subscribes or unsubscribes from channels and checks that the correct number of channels
+// have been subscribed to or unsubscribed from.
+func (w *Websocket) updateChannelSubscriptions(c Connection, store *subscription.Store, incoming subscription.List) error {
+	subs, unsubs := w.GetChannelDifference(c, incoming)
+	if len(unsubs) != 0 {
+		prevState := store.Len()
+		if err := w.UnsubscribeChannels(c, unsubs); err != nil {
+			return err
+		}
+		if diff := prevState - store.Len(); diff != len(unsubs) {
```
gbjk, line 691: This doesn't feel right ...
1) Shouldn't Unsubscribe be erroring if it wasn't successful ?
2) Shouldn't the state of all subs in unsubs be changing ?
That said, it's not too bad either.

gbjk (approved):
Looks good.
It's not too invasive, and doesn't change too much, so I'm happy.

I'm working on sub management and routing separately for gateio, but this doesn't tread on any toes.

exchanges/stream/websocket.go
shazbert: > Shouldn't Unsubscribe be erroring if it wasn't successful ?

I think it should, I just added it as a catch all in the event we forgot to remove it from the store when it was successful. Then it should complain. _should_ 😬 

> Shouldn't the state of all subs in unsubs be changing

Now you are making me think this is all completely wrong 😆. This specifically didn't catch any issues. Can you suggest a better way as a back up check? Cause I am drooling at my screen trying to figure it out 🤤.

exchanges/stream/websocket.go
```diff
@@ -683,6 +679,31 @@ func (w *Websocket) FlushChannels() error {
 	return nil
 }
 
+// updateChannelSubscriptions subscribes or unsubscribes from channels and checks that the correct number of channels
+// have been subscribed to or unsubscribed from.
+func (w *Websocket) updateChannelSubscriptions(c Connection, store *subscription.Store, incoming subscription.List) error {
```
gloriousCode, line 684: Please add a dedicated test to this function

gloriousCode (changes requested):
A strict rule, but it makes sense

exchanges/deribit/deribit_websocket.go
```diff
@@ -836,7 +836,14 @@ func (d *Deribit) handleSubscription(method string, subs subscription.List) erro
...
 	}
-	return err
+	if err != nil {
+		return err
+	}
+
+	if method == "unsubscribe" {
+		return d.Websocket.RemoveSubscriptions(d.Websocket.Conn, subs...)
+	}
+	return d.Websocket.AddSubscriptions(d.Websocket.Conn, subs...)
```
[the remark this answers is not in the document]
shazbert: Nice catch, didn't even notice that. None of those subs were being added to the subscription store and my changes complained. I will check it out.

exchanges/stream/websocket_test.go
```diff
@@ -63,31 +63,33 @@ type testSubKey struct {
 	Mood string
 }
 
-var defaultSetup = &WebsocketSetup{
-	ExchangeConfig: &config.Exchange{
-		Features: &config.FeaturesConfig{
-			Enabled: config.FeaturesEnabledConfig{Websocket: true},
+func newDefaultSetup() *WebsocketSetup {
```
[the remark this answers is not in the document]
shazbert: Sure, what do you want it called? It usually prepend `new` when its a pointer to a type but its not a zero valued allocation so it's not technically correct.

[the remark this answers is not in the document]
gloriousCode: Actually never mind. I'm going to say I was too tired when I looked at that

exchanges/stream/websocket.go
shazbert: [Done here](https://github.com/thrasher-corp/gocryptotrader/pull/1717/commits/759e225bba1e50cac9bcc6e592a218c336579d2d)

exchanges/deribit/deribit_websocket.go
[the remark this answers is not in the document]
shazbert: [Done here](https://github.com/thrasher-corp/gocryptotrader/pull/1717/commits/fcf1d23cea4a618ad686cdd909fd136230c4e9df)

exchanges/deribit/deribit_websocket.go
```diff
@@ -827,23 +831,23 @@ func (d *Deribit) handleSubscription(method string, subs subscription.List) erro
 		subAck[c] = true
 	}
 	if len(subAck) != len(subs) {
-		err = common.ErrUnknownError
+		err = errSubscriptionFailureDetected
 	}
 	for _, s := range subs {
 		if _, ok := subAck[s.QualifiedChannel]; ok {
-			err = common.AppendError(err, d.Websocket.AddSuccessfulSubscriptions(d.Websocket.Conn, s))
+			if strings.Contains(method, "subscribe") {
```
gloriousCode, line 838: I haven't tested yet, but "unsubscribe" contains the string "subscribe" and so both would likely fall in here

shazbert: [pushed](https://github.com/thrasher-corp/gocryptotrader/pull/1717/commits/7401285cad333b21b7295baf48c1c85b139c621a)

exchanges/bybit/bybit_websocket.go
```diff
@@ -167,28 +167,30 @@ func (by *Bybit) handleSubscriptions(operation string, subs subscription.List) (
 	if err != nil {
 		return
 	}
-	chans := []string{}
-	authChans := []string{}
+	var chans subscription.List
+	var authChans subscription.List
```
gbjk, line 171: Nit:
If `handleSubscriptions` accepts `auth bool` then you can call it twice with `subs.Public(), false` and `subs.Private(), true`.
Then all the code below becomes the same, and auth field uses the method param.
Side effect is that authed subs would get batched into groups of 10, but I think that's neither an issue or possible.

exchanges/deribit/deribit_websocket.go
```diff
@@ -123,6 +123,8 @@ var (
 	}
 )
 
+var errSubscriptionFailureDetected = errors.New("subscription failure detected")
```
gbjk, line 126: This feels like it's `stream.ErrSubscriptionFailure `in disguise 🤠

exchanges/deribit/deribit_websocket.go
```diff
@@ -779,6 +781,7 @@ func (d *Deribit) GetSubscriptionTemplate(_ *subscription.Subscription) (*templa
 		"channelName":     channelName,
 		"interval":        channelInterval,
 		"isSymbolChannel": isSymbolChannel,
+		"fmt":             formatPerpetualPairWithSettlement,
```
gbjk, line 784: Nit: Don't think we can call this `fmt` when it's specific to Futures.

exchanges/deribit/deribit_websocket.go
```diff
@@ -827,16 +831,23 @@ func (d *Deribit) handleSubscription(method string, subs subscription.List) erro
...
 		if _, ok := subAck[s.QualifiedChannel]; ok {
-			err = common.AppendError(err, d.Websocket.AddSuccessfulSubscriptions(d.Websocket.Conn, s))
+			if !strings.Contains(method, "unsubscribe") {
+				err = common.AppendError(err, d.Websocket.AddSuccessfulSubscriptions(d.Websocket.Conn, s))
+			} else {
+				err = common.AppendError(err, d.Websocket.RemoveSubscriptions(d.Websocket.Conn, s))
+			}
 		} else {
-			err = common.AppendError(err, errors.New(s.String()))
+			err = common.AppendError(err, errors.New(s.String()+"failed to"+method))
```
gbjk, line 844: Unless I'm wrong:
```suggestion
			err = common.AppendError(err, errors.New(s.String()+" failed to "+method))
```

exchanges/deribit/deribit_websocket.go
```diff
@@ -827,16 +831,23 @@ func (d *Deribit) handleSubscription(method string, subs subscription.List) erro
 		subAck[c] = true
 	}
 	if len(subAck) != len(subs) {
-		err = common.ErrUnknownError
+		err = errSubscriptionFailureDetected
```
gbjk, line 834: So, this would be covering ... any subs that come back that we didn't expect, which won't be reported in the for range below?
It feels like it'd be better to delete from subAck as we range subs, and then error to say "Unexpected channels in Result" or something.
Cos otherwise, this error will be impossible to diagnose and I'd probably say misleading.

exchanges/stream/websocket.go
```diff
@@ -455,6 +461,11 @@ func (w *Websocket) connect() error {
 			break
 		}
 
+		if len(subs) != 0 && w.connectionManager[i].Subscriptions.Len() != len(subs) {
+			multiConnectFatalError = fmt.Errorf("%v %w expecting %d subscribed %v", w.exchangeName, errSubscriptionsNotAdded, len(subs), subs)
+			break
```
gbjk, line 466: I still have this problem:
We call connect on an exchange, one pair on one asset fails to subscribe (like TRUMPEW the other day) and all connections get torn down for all asset types and subs.
I still don't agree with that.
The biggest problem I think I have is the strong coupling between connections and subs.
If a consumer wants to connect the websocket without subs, and then later call `Subscribe` manually, we both *allow* it, and *don't support it* at the same time.

I think this is a wider topic, because this change doesn't make 454/460 *worse*, but it doubles down on the concept.

I think we need to say "Connecting is connecting, and Subscribing is separate" especially when new subs or resubs could connect new subs.

I'll continue to work on this, so I'm just highlighting for alignment on direction.

In the meantime, I'd gently request again to log the error, and return it, but not tear down all of the subs and conns for all assets.

exchanges/stream/websocket.go
gbjk: Sorry for late reply on this.
I think I'd want see that ranging the subs and unsubs have changed State, and that store contains (or doesn't) each one. `ContainsAll` and `ContainsNone` or something.
I really don't like checking len as a catchall, because it could false positive.
We're not locking store when we do any of this, afterall.

exchanges/stream/websocket.go
```diff
@@ -683,6 +679,34 @@ func (w *Websocket) FlushChannels() error {
 	return nil
 }
 
+// updateChannelSubscriptions subscribes or unsubscribes from channels and checks that the correct number of channels
+// have been subscribed to or unsubscribed from.
+func (w *Websocket) updateChannelSubscriptions(c Connection, store *subscription.Store, incoming subscription.List) error {
+	if store == nil {
+		return fmt.Errorf("%w for %T", common.ErrNilPointer, store)
+	}
```
gbjk, line 687: Nit: Since this is a private method, I'd hope we wouldn't need this check, and associated tests.

gbjk (approved):
Good work.

Only nits.

I'm asking for a change to not tear everything down on single sub failure, but if you don't make that change, it's no worse than before, so I'm fine.

exchanges/deribit/deribit_websocket.go
shazbert: It calls it against every pair, are you asking me to only call it against futures?

exchanges/stream/websocket.go
shazbert: > We call connect on an exchange, one pair on one asset fails to subscribe (like TRUMPEW the other day) and all connections get torn down for all asset types and subs.

Do you remember what specifically caused this?

The coupling connect() with subs is not great, regarding `"Connecting is connecting, and Subscribing is separate"` though is I think `connect` function might be changed to handle connections that don't require subscriptions and spawn a middleware handler that performs a JIT connection based on current/incoming subscriptions which also drops connections when unsubscribing when subs is empty, that way it scales a bit better with respect to a connections max sub capacity. I don't currently have a design that could help you in that respect though. 

> In the meantime, I'd gently request again to log the error, and return it, but not tear down all of the subs and conns for all assets.

Sure, if it's easier for everyone else I will just log it and I can just modify my own trading branch to be throwing the baby out with the bathwater myself. 🤷

shazbert: @gloriousCode @thrasher- I pivot off these errors in my own trading branches that retries connect until it establish a clean slate or base line of full subscriptions across all trading pairs that I require. Happy to just log this out like GK suggests if we all agree and I will update tests.

exchanges/deribit/deribit_websocket.go
gbjk: Nope. But I think it's misleading that it's not really format for every pair, even if it's called for all of them.
So I guess I'd vote for: Rename the template func to `fmtFuturePairs` or rename the func called to formatChannelPair.
What's catching me is the disconnect between what's in the template and the actual func name.
Can also ignore.

exchanges/stream/websocket.go
gbjk: > Do you remember what specifically caused this?

A badly listed futures pair erroring on orderbook.

> I pivot off these errors in my own trading branches that retries connect until it establish a clean slate or base line of full subscriptions across all trading pairs that I require.

Okay. This might be the crux of our disagreement.
You should not need to disconnect and unsub everything just because one thing fails, if your goal is to ensure it eventually works.
connectionManager should be monitoring and reconnecting, and if it doesn't have requisite granularity we should improve that.
And I think subscriptionManager should manager subscriptions.

What I'm aiming for is smaller responsibilities for functions that have names (and meanings) like connect, so they're more versatile (or composable) and easier to test.

Can you throw me any other requirements you might have like "retry subs and conns until everything is working", if you have any ?

I see exactly the same situation, and also wanna know "When's everything looking good?".

exchanges/subscription/store.go
```diff
@@ -206,3 +206,21 @@ func (s *Store) Len() int {
 	defer s.mu.RUnlock()
 	return len(s.m)
 }
+
+// PartitionByPresence returns two lists; one with the subscriptions that are present in the store and one with the
+// subscriptions that are not
+func (s *Store) PartitionByPresence(compare List) (matched, unmatched List) {
```
gbjk, line 212: Thought about this for a while, and I think it's cute, but wrong to have a function where you're always using one or the other return.
I thought about giving it a bool for whether you want it there or not.
Took me a while to work out why "Presence" is wrong; It's a noun. `Present` is more correct, [having an adjective form](https://dictionary.cambridge.org/dictionary/english/present), but still sounds weird.

I'd vote for 2 funcs:
* Contained returns a List of subs in compare which are in the store already
  * alt. Filter, Matching, Match, Union, Contains (but sounds boolish)
* Missing returns a List of subs in compare which are not in the store already
  * alt. Absent, NotContained, Excluded

Alternative note: `FilterMatching` is probably the best name for keeping a single func, but adding a "contained or missing bool"

I'm feeling a tingle about how similar this is to Diff, and yet not.
Also around Venn and terminology. Spent a while looking at `slices` and at wiki on set theory and any precedence on this in stack exchange, but came up with nothing good.

gbjk (changes requested):
Great work.
Sorry for change request. Spent a while debating. 😢

exchanges/subscription/store.go
shazbert: Yeah good suggestion. I will introduce contained and missing as it has a better approach and single responsibility is more efficient even though it increases code base. If you wanted a set theory naming convention could have `Intersection` and `Difference` but it becomes less obvious for people with little to no experience with that terminology. (me)

exchanges/stream/websocket.go
shazbert: > What I'm aiming for is smaller responsibilities for functions that have names (and meanings) like connect, so they're more versatile (or composable) and easier to test.

Good direction to have, I agree.

> Can you throw me any other requirements you might have like "retry subs and conns until everything is working", if you have any ? && "When's everything looking good?"

Pretty much just this; Personally, I am still going to rely on ensuring a clean slate of full subscriptions before proceeding. If even one subscription fails, it indicates a potential issue with the connection or the subscription logic, and I prefer to retry until everything is established correctly, or bring the instance offline and fix it.

gloriousCode: - I see gk's issue with regards to killing all connections, but I don't see if as something huge because we already kill everything when `Subscribe` is called and return errors. The refactoring you speak of gk will change that section regardless, but I get not wanting to tangle further.
- Logging the issue isn't actionable via code and so I don't think that's the best way to go

I propose we move subscribing for multi-connections to the end of the `connect()` function. It is a small change that satisfies the following criteria:
- The default handling of websocket connections via `websocket_routinemanager.go` will log any errors returned
- It won't by default kill all connections when subscriptions have issue
- It allows users who call `Connect()` directly to handle any errors that are returned, and resubscribe/reconnect/`panic("help!")`

Gist: 
https://gist.github.com/gloriousCode/609d25ea1ee9b954ecbb95b505c76609

It comes with the caveat that I haven't tested things thoroughly. I tried things like separating subscribing, but that's way too large a refactor for this PR and comes with many extra considerations that I don't think are what is desired out of this PR - which is to ensure all subscriptions are subscribed to and that the caller knows it

Does this solution make people happy?

gbjk: Yep, that's a good compromise with minimal intrusion.

shazbert: [Check me please 🙏 ](https://github.com/thrasher-corp/gocryptotrader/pull/1717/commits/ba022ece8c778bdcfa27894bb785baaef190e441)

exchanges/deribit/deribit_websocket.go
shazbert: [fixed](https://github.com/thrasher-corp/gocryptotrader/pull/1717/commits/9d7f1169043d7216e6d1cd18f2612cf799967662)

exchanges/stream/websocket.go
```diff
@@ -56,6 +56,8 @@ var (
 	errReadMessageErrorsNil                 = errors.New("read message errors is nil")
 	errWebsocketSubscriptionsGeneratorUnset = errors.New("websocket subscriptions generator function needs to be set")
 	errSubscriptionsExceedsLimit            = errors.New("subscriptions exceeds limit")
+	errSubscriptionsNotAdded                = errors.New("subscriptions not added")
+	errSubscriptionsNotRemoved              = errors.New("subscriptions not removed")
```
gloriousCode, line 60: Now that these are being returned by `Connect()`, I think it would be nice to export these so you can help distinguish an error from connecting or an error for subscribing and act accordingly

shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1717/commits/aca87c35d233d709a78ffb60b8085ed8e5d7781d)

exchanges/subscription/store.go
```diff
@@ -207,20 +207,34 @@ func (s *Store) Len() int {
...
-func (s *Store) PartitionByPresence(compare List) (matched, unmatched List) {
+// Contained returns a list of subscriptions in `compare` that are already in the store.
+func (s *Store) Contained(compare List) List {
 	if s == nil || s.m == nil {
-		return nil, compare // All are unmatched
+		return nil
 	}
 	s.mu.RLock()
 	defer s.mu.RUnlock()
+	var matched List
```
gbjk, line 217: Can move this to the signature if you want, and/but still use explicit return with it.

exchanges/subscription/store.go
```diff
@@ -207,20 +207,34 @@ func (s *Store) Len() int {
...
+}
+
+// Missing returns a list of subscriptions in `compare` that are not in the store.
+func (s *Store) Missing(compare List) List {
+	if s == nil || s.m == nil {
+		return compare // All are unmatched
+	}
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+	var unmatched List
```
gbjk, line 233: Whilst considering maybe moving this to the signature, can also consider calling it `missing` for consistency.

gbjk (approved):
All good from me 👍 Great work.

Comments are entirely optional notes.

gloriousCode (approved):
tACK!

exchanges/deribit/deribit_websocket.go
```diff
@@ -827,16 +829,29 @@ func (d *Deribit) handleSubscription(method string, subs subscription.List) erro
...
+			err = common.AppendError(err, errors.New(s.String()+" failed to "+method))
 		}
 	}
-	return err
+
+	for key := range subAck {
+		err = common.AppendError(err, fmt.Errorf("unexpected channel `%s` in result", key))
+	}
+
+	if err != nil {
```
thrasher-, line 851:
```suggestion
	return err
```

exchanges/stream/websocket.go
```diff
@@ -24,8 +24,10 @@ const jobBuffer = 5000
 var (
 	ErrWebsocketNotEnabled      = errors.New("websocket not enabled")
 	ErrSubscriptionFailure      = errors.New("subscription failure")
-	ErrSubscriptionNotSupported = errors.New("subscription channel not supported ")
 	ErrUnsubscribeFailure       = errors.New("unsubscribe failure")
+	ErrSubscriptionNotSupported = errors.New("subscription channel not supported ")
```
thrasher-, line 28: Whilst noticing the space after this, also noticed that this is unused so can be removed

exchanges/stream/websocket_test.go
```diff
@@ -1111,16 +1105,37 @@ func TestFlushChannels(t *testing.T) {
 	// Disable pair and flush system
 	newgen.EnabledPairs = []currency.Pair{currency.NewPair(currency.BTC, currency.AUD)}
 	w.GenerateSubs = func() (subscription.List, error) { return subscription.List{{Channel: "test"}}, nil }
-	err = w.FlushChannels()
-	require.NoError(t, err, "Flush Channels must not error")
+
+	require.ErrorIs(t, w.FlushChannels(), ErrSubscriptionsNotAdded, "FlushChannels should error correctly on no subscriptions added")
```
thrasher-, line 1109:
```suggestion
	require.ErrorIs(t, w.FlushChannels(), ErrSubscriptionsNotAdded, "FlushChannels must error correctly on no subscriptions added")
```

exchanges/stream/websocket_test.go
```diff
@@ -1111,16 +1105,37 @@ func TestFlushChannels(t *testing.T) {
...
+	w.Subscriber = func(subs subscription.List) error {
+		for _, sub := range subs {
+			if err := w.subscriptions.Add(sub); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+
+	require.NoError(t, w.FlushChannels(), "Flush Channels must not error")
```
thrasher-, line 1120:
```suggestion
	require.NoError(t, w.FlushChannels(), "FlushChannels must not error")
```

To be consistent with the previous one

exchanges/stream/websocket_test.go
```diff
@@ -1111,16 +1105,37 @@ func TestFlushChannels(t *testing.T) {
...
+
+	w.Unsubscriber = func(subs subscription.List) error {
+		for _, sub := range subs {
+			if err := w.subscriptions.Remove(sub); err != nil {
+				return err
+			}
+		}
+		return nil
+	}
+	assert.NoError(t, w.FlushChannels(), "Flush Channels should not error")
```
thrasher-, line 1138:
```suggestion
	assert.NoError(t, w.FlushChannels(), "FlushChannels should not error")
```

thrasher- (changes requested):
Some minor nitterinos

exchanges/stream/websocket.go
shazbert: [thanks](https://github.com/thrasher-corp/gocryptotrader/pull/1717/commits/68461c546d5d3ec4fc82e338338091bfa1d66c71)

thrasher- (approved):
tACK Bybit and Deribit, nice work on resolving the pair formatting issue on Deribit!

exchanges/stream/websocket.go
thrasher-: > > Do you remember what specifically caused this?
> 
> A badly listed futures pair erroring on orderbook.
> 
> > I pivot off these errors in my own trading branches that retries connect until it establish a clean slate or base line of full subscriptions across all trading pairs that I require.
> 
> Okay. This might be the crux of our disagreement. You should not need to disconnect and unsub everything just because one thing fails, if your goal is to ensure it eventually works. connectionManager should be monitoring and reconnecting, and if it doesn't have requisite granularity we should improve that. And I think subscriptionManager should manager subscriptions.
> 
> What I'm aiming for is smaller responsibilities for functions that have names (and meanings) like connect, so they're more versatile (or composable) and easier to test.
> 
> Can you throw me any other requirements you might have like "retry subs and conns until everything is working", if you have any ?
> 
> I see exactly the same situation, and also wanna know "When's everything looking good?".

Agree that long term we shouldn't have to disconnect/reconnect on a single sub failure. Some long term goals: I'd like to be able to observe the complete subs state per asset, with a retry counter threshold if it fails and for it to disable a sub if that retry counter failure limit is reached. Ryan has already start implementing some metric counters per sub/event which is quite cool. Also, `gctcli` has the ability to enable/disable pairs on the fly, so our subs should scale up/down depending on that, with a viewable state status and without having to flush existing subs/conns to do so

PR #1722, opened 2024-11-19
gateio: Add concurrent subscription support

shazbert:
# PR Description

* Move ParallelChanOp functionality which is epic (shout out @gbjk  :fire:) and create generic common func `ProcessBatches`
* Create specific generic processing function with different signature for higher throughput `ProcessElementsByBatch`
* concurrently subscribe/unsubscribe support across asset types
* remove rate limiter
* requires #1733 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


common/common.go
```diff
@@ -645,3 +645,59 @@ func (c *Counter) IncrementAndGet() int64 {
...
+	return errs
+}
+
+// BatchProcessor defines the function signature for processing a batch of elements.
+type BatchProcessor[S ~[]E, E any] func(batch S) error
+
+// ProcessBatches takes a slice of elements and processes them in batches of `batchSize` concurrently.
+// For example, if batchSize = 10 and list has 100 elements, 10 goroutines will be created to process
+// 10 batches. Each batch is processed sequentially by the `process` function, and batches are processed
+// in parallel.
```
gbjk, line 684: I'm afraid I had to re-read this comment a few times.
I don't believe we should say:
```Each batch is processed sequentially by the `process` function```
Because I don't think you can guarantee the goros will execute in the order scheduled.

I'd suggest vastly simplified:
```suggestion
// ProcessBatches processes `S` concurrently in batches of `batchSize`
```
Note: I often remove types from comments, because the user sees the param names and types in the signature in godoc and all other circumstances. It doesn't help to repeat the types, and it makes it harder to read and internalise quickly.

common/common.go
```diff
@@ -645,3 +645,59 @@ func (c *Counter) IncrementAndGet() int64 {
...
 }
+
+// ElementProcessor defines the function signature for processing an individual element with its index.
+type ElementProcessor[E any] func(index int, element E) error
+
+// ProcessElementsByBatch takes a slice of elements and processes them in batches of `batchSize` concurrently.
+// For example, if batchSize = 10 and list has 100 elements, 10 goroutines will process 10 elements concurrently
+// in each batch. Each batch completes before the next batch begins.
+// `process` is a function called for each individual element with its index and value.
+func ProcessElementsByBatch[S ~[]E, E any](batchSize int, list S, process ElementProcessor[E]) (errs error) {
```
gbjk, line 656: This is more about throttling than it is batching.
I'll think today about naming / wording, but it wasn't until the third time round that I really clocked fully that this is a concurrent throughput throttler, not a batcher.

common/common.go
```diff
@@ -645,3 +645,59 @@ func (c *Counter) IncrementAndGet() int64 {
...
+
+// ProcessBatches takes a slice of elements and processes them in batches of `batchSize` concurrently.
+// For example, if batchSize = 10 and list has 100 elements, 10 goroutines will be created to process
+// 10 batches. Each batch is processed sequentially by the `process` function, and batches are processed
+// in parallel.
+func ProcessBatches[S ~[]E, E any](batchSize int, list S, process BatchProcessor[S, E]) (errs error) {
+	var wg sync.WaitGroup
+	errC := make(chan error, len(list))
+	for _, s := range Batch(list, batchSize) {
+		wg.Add(1)
```
gbjk, line 689:
```suggestion
	var wg sync.WaitGroup
	wg.Add(len(list))
	errC := make(chan error, len(list))
	for _, s := range Batch(list, batchSize) {
```

common/common.go
```diff
@@ -645,3 +645,59 @@ func (c *Counter) IncrementAndGet() int64 {
...
+// BatchProcessor defines the function signature for processing a batch of elements.
+type BatchProcessor[S ~[]E, E any] func(batch S) error
+
+// ProcessBatches takes a slice of elements and processes them in batches of `batchSize` concurrently.
+// For example, if batchSize = 10 and list has 100 elements, 10 goroutines will be created to process
+// 10 batches. Each batch is processed sequentially by the `process` function, and batches are processed
+// in parallel.
+func ProcessBatches[S ~[]E, E any](batchSize int, list S, process BatchProcessor[S, E]) (errs error) {
+	var wg sync.WaitGroup
+	errC := make(chan error, len(list))
```
gbjk, line 687: errs := CollectErrors(len(list))

gbjk, line 687:
```suggestion
    errs := CollectErrors(len(list))
```

common/common.go
```diff
@@ -645,3 +645,59 @@ func (c *Counter) IncrementAndGet() int64 {
...
+				errC <- err
+			}
+		}(s)
+	}
+	wg.Wait()
+	close(errC)
+	for err := range errC {
+		errs = AppendError(errs, err)
+	}
+	return errs
```
gbjk, line 702:
```suggestion
	return errs.Collect()
```

gbjk (changes requested):
Feedback on the common part only to start with.

More to come, but some of it is easy to action.

common/common.go
```diff
@@ -645,3 +645,59 @@ func (c *Counter) IncrementAndGet() int64 {
...
+func ProcessBatches[S ~[]E, E any](batchSize int, list S, process BatchProcessor[S, E]) (errs error) {
+	var wg sync.WaitGroup
+	errC := make(chan error, len(list))
+	for _, s := range Batch(list, batchSize) {
+		wg.Add(1)
+		go func(s S) {
+			defer wg.Done()
+			if err := process(s); err != nil {
+				errC <- err
+			}
```
gbjk, line 694: When using ErrorCollector you can just send the nils to errs as well :)
```suggestion
		        errs <- process(s)
```

common/common.go
gbjk: I also just realised:
Is this implementation mechanics truly optimal for your use-case?
Normally you'd see things processed with a concurrency of 10.
Your implementation here may end up with a concurrency of 1 or 2 at then end of each batch, instead of maintaining 10 throughout
Maybe this is going to be self-evident in the code that calls this.

exchanges/gateio/gateio_websocket.go
```diff
@@ -581,32 +584,34 @@ func (g *Gateio) manageSubs(ctx context.Context, event string, conn stream.Conne
...
-			errs = common.AppendError(errs, fmt.Errorf("%s %s %s: %w", s.Channel, s.Asset, s.Pairs, err))
+	return common.ProcessElementsByBatch(subscriptionBatchCount, subs, func(_ int, s *subscription.Subscription) error {
+		if err := g.manageTemplatePayload(ctx, conn, event, s); err != nil {
+			return fmt.Errorf("%s %s %s: %w", s.Channel, s.Asset, s.Pairs, err)
 		}
+		return nil
+	})
+}
+
+func (g *Gateio) manageTemplatePayload(ctx context.Context, conn stream.Connection, event string, s *subscription.Subscription) error {
```
gbjk, line 595: This isn't anything to do with templates, especially since a user might have called Subscribe manually.
I think this is:
```suggestion
func (g *Gateio) manageSubPayload(ctx context.Context, conn stream.Connection, event string, s *subscription.Subscription) error {
```

exchanges/gateio/gateio_websocket.go
```diff
@@ -581,32 +584,34 @@ func (g *Gateio) manageSubs(ctx context.Context, event string, conn stream.Conne
...
-			}
-			return g.Websocket.AddSuccessfulSubscriptions(conn, s)
-		}(); err != nil {
-			errs = common.AppendError(errs, fmt.Errorf("%s %s %s: %w", s.Channel, s.Asset, s.Pairs, err))
+	return common.ProcessElementsByBatch(subscriptionBatchCount, subs, func(_ int, s *subscription.Subscription) error {
+		if err := g.manageTemplatePayload(ctx, conn, event, s); err != nil {
+			return fmt.Errorf("%s %s %s: %w", s.Channel, s.Asset, s.Pairs, err)
 		}
+		return nil
+	})
```
gbjk, line 592: Can you add a comment to explain *why* this should be done in batches *rather* than just throttled ?

exchanges/gateio/gateio_websocket.go
```diff
@@ -696,27 +701,29 @@ func (g *Gateio) handleSubscription(ctx context.Context, conn stream.Connection,
...
+
+	return common.ProcessElementsByBatch(subscriptionBatchCount, payloads, func(index int, payload WsInput) error {
+		if err := g.managePayload(ctx, conn, event, channelsToSubscribe[index], &payload); err != nil {
+			return fmt.Errorf("%s %s %s: %w", channelsToSubscribe[index].Channel, channelsToSubscribe[index].Asset, channelsToSubscribe[index].Pairs, err)
 		}
+		return nil
+	})
+}
+
+func (g *Gateio) managePayload(ctx context.Context, conn stream.Connection, event string, s *subscription.Subscription, payload *WsInput) error {
```
gbjk, line 713: This looks like: 
```suggestion
func (g *Gateio) sendSubPayload(ctx context.Context, conn stream.Connection, event string, s *subscription.Subscription, payload *WsInput) error {
```

exchanges/gateio/gateio_websocket.go
gbjk: Think you've made Subscribe into manageSubs, and manageSubs into manageTemplatePayload :)

common/common.go
shazbert: > Because I don't think you can guarantee the goros will execute in the order scheduled.

They are executed sequentially in that individual go routine.

Here's a better example:

```md
List of elements: [E1, E2, E3, ..., E100]

Batch size: 10

Step-by-step process:
1. Divide the list into batches of 10 elements each:
   Batch 1: [E1, E2, E3, ..., E10]
   Batch 2: [E11, E12, E13, ..., E20]
   ...
   Batch 10: [E91, E92, E93, ..., E100]

2. Spawn 10 goroutines to process the batches concurrently:
   Goroutine 1 -> Processes Batch 1
   Goroutine 2 -> Processes Batch 2
   ...
   Goroutine 10 -> Processes Batch 10

3. Within each goroutine:
   - Process elements in the batch sequentially using the `process` function.
   Example for Batch 1:
       Process E1 -> Process E2 -> Process E3 -> ... -> Process E10
```

common/common.go
shazbert: Have you tested this as happening? Here's a better example like this one above:

```md
List of elements: [E1, E2, E3, ..., E100]

Batch size: 10

Step-by-step process:
1. Divide the list into batches of 10 elements each:
   Batch 1: [E1, E2, E3, ..., E10]
   Batch 2: [E11, E12, E13, ..., E20]
   ...
   Batch 10: [E91, E92, E93, ..., E100]

2. Process the batches sequentially:
   Batch 1 -> Batch 2 -> Batch 3 -> ... -> Batch 10
   (Each batch completes fully before moving to the next batch.)

3. Within each batch:
   - Spawn 10 goroutines (one for each element in the batch).
   - Each goroutine processes a single element by calling the `process` function with its index and value.
   
   Example for Batch 1:
       Goroutine 1 -> `process(0, E1)`
       Goroutine 2 -> `process(1, E2)`
       Goroutine 3 -> `process(2, E3)`
       ...
       Goroutine 10 -> `process(9, E10)`
```

So concurrency with this one scales linearly to how big the size is. `100 batch size` in this case will spawn `100 go routines` and process them all at once. Whereas the original I think (?) had an inverse relationship with the batch size, the higher the batch size the lower the concurrency. `100 batch size`  will process the entire list sequentially but a batch size of 1 would process instantly.

common/common.go
shazbert: 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥 🔥

common/common.go
gloriousCode: You can't guarantee `process` does it sequentially. You can't even guarantee that `process` does anything from here. Its just that you're doing it sequentially for ws subs for the moment. I agree that comment about sequentially processed should be removed.

common/common.go
gbjk: Yeah, I did understand from the start.
I just don't know why you'd want to separate things into groups of 10 and then process all 10 before processing the next batch.
If 9 subscriptions complete in 0.2s each concurrently, but gateio decides to take 4 seconds on just one in that batch, because it decided now's a good time to cache invalidate, or whatever, then you're sitting there twiddling your thumbs.

I *can* imagine situations where this pattern is necessary, where there's some inter-relation that 10 must happen fully before the next 10. I just don't think that's the case for you here.

What would be the downside of a channel throttle that meant that you're *always* processing no more than 10 subscriptions?
That would make this a standard mini-worker and queue pattern.
I'm sure you know, but: _Fill a channel with slice, spawn N goros which read from the channel and exit when they don't get anything, waitgroup them and it's done._

Happy to throw up a PR into your branch if it helps.

But maybe I'm being really stupid and missing something obvious about why we'd need fully completed batches.

common/common.go
```diff
@@ -647,3 +647,38 @@ func (c *Counter) IncrementAndGet() int64 {
...
+	var errs error
+	for i, s := range Batch(list, batchSize) {
+		err := CollectErrors(len(s))
+		for j, e := range s {
+			go func(index int, element E) { err.C <- process(index, element); err.Wg.Done() }((i*batchSize)+j, e)
+		}
+		errs = AppendError(errs, err.Collect())
+	}
+	return errs
+}
```
gbjk, line 668: Okay, so this is what I *think* is appropriate for your use case.
Note: Completely untested. It compiles and it's (probably) sane)
I've omitted i because I can't see that you need it, and can't imagine how it'd make sense given it's not guaranteed to be in sequence at all. If there was some use-case for it that I'm missing, then you could use a sequence generator to get it, or something. You could also maybe do this less elegantly by using mutexes and reducing the slice, or using a producer goro.
Anyway. Enough babbling. More code:
```suggestion
// PipelineProcessor defines the function signature for processing an individual elements
type PipelineProcessor[E any] func(element E) error

// ThrottledPipeline processes a slice concurrently with a throttled number of workers
func ThrottledPipeline[S ~[]E, E any](workers int, list S, process PipelineProcessor[E]) error {
	q := make(chan E, len(list))
	for _, s := range list {
		q <- s
	}
	err := CollectErrors(len(list))
	for range workers {
		go func() {
			defer err.Wg.Done()
			for {
				select {
				case s := <-q:
					err.C <- process(s)
				default:
					return
				}
			}
		}()
	}
	return err.Collect()
}
```

exchanges/gateio/gateio_websocket.go
gbjk: Thanks for the comment.
I think your explanation doesn't justify staggered batch processing over outright throughput workers.
I've provided an example implementation covering what I think might be more suitable.

common/common.go
shazbert: I just remembered the main reasoning for Gateio snapshots specifically with 100 len books starts to interfere with gorilla websocket buffer and clips data, now that rate limiting subscription requests have been removed, so its been limited to 200 sub blocks which spawns 200 go routines. I will add that into the commentary 😆. 

> If 9 subscriptions complete in 0.2s each concurrently, but gateio decides to take 4 seconds on just one in that batch, because it decided now's a good time to cache invalidate, or whatever, then you're sitting there twiddling your thumbs.

Has this happened in testing? We can always mitigate it with a context deadline? But this seems like a bottom quintile issue only on startup and not really an issue with a long running program as the changes with subs are going to be limited to scaling up and down on exchange pair additions, removals and strategy requirements.  

Will check out your example now thanks.

common/common.go
shazbert: There's definitely a use case for this if it works 🔥 `Note: Completely untested` 😆 Will Save this.

common/common.go
shazbert: How about this as an option for the commentary.?
// ProcessBatches splits `S` into batches of `batchSize` and processes each batch concurrently.

There is also an issue when batchSize >= len(S) then there is no concurrency. Should I handle that case?

common/common.go
shazbert: @gbjk renamed, opted for original, added commentary as its used externally and battle tested. Can circle back to this if required. [f1f2299](https://github.com/thrasher-corp/gocryptotrader/pull/1722/commits/f1f2299f29e58fed83fc7848c4e1d497d2a3715e)

common/common.go
```diff
@@ -679,3 +679,39 @@ func (c *Counter) IncrementAndGet() int64 {
...
+
+// ElementProcessor defines the function signature for processing an individual element with its index.
+// Index required in old subscription matching // TODO: Remove this when gateio templates are updated to all assets.
+type ElementProcessor[E any] func(index int, element E) error
+
+// ThrottledBatch takes a slice of elements and processes them in batches of `batchSize` concurrently.
+// For example, if batchSize = 10 and list has 100 elements, 10 goroutines will process 10 elements concurrently
+// in each batch. Each batch completes before the next batch begins.
+// `process` is a function called for each individual element with its index and value.
+func ThrottledBatch[S ~[]E, E any](batchSize int, list S, process ElementProcessor[E]) error {
```
gloriousCode, line 691: I feel like these names is something you've already discussed with gk. I don't think `ThrottledBatch` works here. You may wish to have a larger name for both elements to make it clearer.

My understanding of what you're saying is that this is `SequentialBatchConcurrentProcessing` Whereas the other function is `ConcurrentBatchSequentialProcessing`
But even with all the commentary, I'm not sure if I have the right of it.

gloriousCode, line 691: No throttling occurs here. You are not analysing and/or limiting throughput based on anything. I don't think ThrottledBatch works

exchanges/gateio/gateio_websocket.go
```diff
@@ -52,12 +52,15 @@ const (
...
+
+	// subscriptionBatchCount is the number of subscriptions to send in a single batch
+	subscriptionBatchCount = 200
 )
 
 var defaultSubscriptions = subscription.List{
 	{Enabled: true, Channel: subscription.TickerChannel, Asset: asset.Spot},
 	{Enabled: true, Channel: subscription.CandlesChannel, Asset: asset.Spot, Interval: kline.FiveMin},
-	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds},
+	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds, Levels: 100},
```
gloriousCode, line 63: This change like what I talked about in your other PR, has issues using `master` based subscriptions
```
[ERROR] | WEBSOCKET | 29/04/2025 10:46:15 | GateIO Error subscribing orderbook spot [ALEPH_USDT]: (2) provided level not supported: 0, orderbook spot [OGN_USDT]: (2) provided level not supported: 0, orderbook spot [ALA_USDT]: (2) provided level not supported: 0, orderbook spot [BTC_USDT]: (2) provided level not supported: 0
```
I don't mind where it gets fixed, but this PR is blocked until its reverted/fixed here/fixed elsewhere

common/common.go
gloriousCode: But even then, much like `ThrottleBatch` doesn't throttle, `BatchProcessor` doesn't actually mean it does anything sequentially. Its passed to a function via goroutines which may also do it concurrently.

Since I'm having such issue understanding what you're doing, I'm going to ask you to provide different names because I can't help

gloriousCode (changes requested):
This PR makes me think I need to pass over a 300 coffee threshold to get it sorry. The only thing I know for sure is that `master` branch will receive errors due to the `Levels` subscription change

common/common.go
shazbert: Don't look at me for naming 😆. The only issue it rectifies is for an orderbook you need to suppress (throttle) the outbound requests because the websocket frame gets clipped in an orderbook snapshot context. So processing one block of 100 subscriptions and wait until that is done before going on to the next. Just note this is not a 429 issue as outbound subscriptions can occur without any rate limiting its only on the return leg and the amount of data is smashing against gorillas websocket frame. `SequentialBatchConcurrentProcessing` and `ConcurrentBatchSequentialProcessing` works for me. But I will get @gbjk to comment as you guys can whipsaw back and forth on this.

exchanges/gateio/gateio_websocket.go
shazbert: Yeah I will revert this 👍

common/common.go
gbjk: Yep. I think I'm still [here](https://github.com/thrasher-corp/gocryptotrader/pull/1722#discussion_r1893400260) but I need to parse this again.
I'll clear my stack and do that on Sunday.

common/common.go
gbjk: Yep, I think I'm sticking to a core concept here:
A set of batch workers which drains all work items from each batch before starting the next one does not make sense.
The amount of work being done over time is uneven and not time-efficient.
The pattern I've provided is a well trodden concept that's been used a lot, whereas this one is counter to what people would expect.

Can you reconsider again please? 🥺 

I'll give it a test this morning myself, as well.

exchanges/exchange.go
```diff
@@ -1772,28 +1772,7 @@ func (b *Base) GetOpenInterest(context.Context, ...key.PairAsset) ([]futures.Ope
...
-	wg.Wait()
-	close(errC)
-
-	var errs error
-	for err := range errC {
-		errs = common.AppendError(errs, err)
-	}
-
-	return errs
+	return common.ProcessBatches(batchSize, channels, m)
```
gbjk: Think this is a no-op now 🤷

exchanges/gateio/gateio_websocket.go
```diff
@@ -668,38 +673,41 @@ func (g *Gateio) GetSubscriptionTemplate(_ *subscription.Subscription) (*templat
...
-			}
-			return g.Websocket.AddSuccessfulSubscriptions(conn, s)
-		}(); err != nil {
-			errs = common.AppendError(errs, fmt.Errorf("%s %s %s: %w", s.Channel, s.Asset, s.Pairs, err))
+	// ThrottledBatch will batch the subscriptions into groups of subscriptionBatchCount then concurrently subscribe to them.
+	// Need to throttle the requests to allow gct to process the incoming responses or else the websocket frame will be
+	// clipped.
+	return common.ThrottledBatch(subscriptionBatchCount, exp, func(_ int, s *subscription.Subscription) error {
+		if err := g.manageSubPayload(ctx, conn, event, s); err != nil {
+			return fmt.Errorf("%s %s %s: %w", s.Channel, s.Asset, s.Pairs, err)
```
gbjk, line 673: 🥃 I worry we should not need to do this generic error wrapping at this level; If we do, every exchange will have to do it.

exchanges/gateio/gateio_websocket.go
```diff
@@ -53,12 +53,15 @@ const (
 
 	subscribeEvent   = "subscribe"
 	unsubscribeEvent = "unsubscribe"
+
+	// subscriptionBatchCount is the number of subscriptions to send in a single batch
+	subscriptionBatchCount = 200
```
gbjk, line 61: I think this is misleading, because it's actually about subscription concurrency, and batching happens inside ExpandTemptlates and manageSubReq where we expand `QualifiedChannel` into a comma separated list.
Other than that pair grouping, we're not batching at all.

exchanges/gateio/gateio_websocket.go
```diff
@@ -668,38 +673,41 @@ func (g *Gateio) GetSubscriptionTemplate(_ *subscription.Subscription) (*templat
...
+	// clipped.
+	return common.ThrottledBatch(subscriptionBatchCount, exp, func(_ int, s *subscription.Subscription) error {
+		if err := g.manageSubPayload(ctx, conn, event, s); err != nil {
+			return fmt.Errorf("%s %s %s: %w", s.Channel, s.Asset, s.Pairs, err)
 		}
+		return nil
+	})
+}
+
+func (g *Gateio) manageSubPayload(ctx context.Context, conn websocket.Connection, event string, s *subscription.Subscription) error {
```
gbjk: 🐒  `xPayload` suggests it's going to return a payload or something.
I think this is actually just `manageSub`
```suggestion
func (g *Gateio) manageSubPayload(ctx context.Context, conn websocket.Connection, event string, s *subscription.Subscription) error {
```

exchanges/gateio/gateio_websocket.go
```diff
@@ -789,29 +797,33 @@ func (g *Gateio) handleSubscription(ctx context.Context, conn websocket.Connecti
 	if err != nil {
```
gbjk, line 931: 🐒 `handleSubscription` is actually `manageOtherSub` vs `manageSpotSub`
Takes me a while each time to understand that.

shazbert:
Closing, not needed now due to gateio library updates and too hard to merge

PR #1734, opened 2024-11-28
stream/multiconnection: Default process reporter 

shazbert:
# PR Description

* Adds optional reader process reporter for metrics aggregation
* In the exchange config set `WebsocketMetricsLogging` to true (only for Gateio) functionality will expand when exchanges get converted over the `multi-connection` usage.
```
[DEBUG] | WEBSOCKET | 28/11/2024 13:21:50 | wss://api.gateio.ws/ws/v4/: Operations/Second: 10, Avg Processing/Operation: 323.81µs, Peak: 168.6498ms
```

Reasoning:
* Can provide detailed websocket operation metrics to help identify performance bottlenecks and areas for improvement. As above clearly shows there can be improvements as per `peak processing timing` which needs to be investigated. 
* This can scale across all exchanges when the multi connection updates has occurred.
* This can be defined outside of GCT

~~Opened as draft if this could be a potential update. Once look over is done I will write tests and cleanup.~~

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/stream/websocket.go
```diff
@@ -1075,14 +1075,27 @@ func (w *Websocket) checkSubscriptions(conn Connection, subs subscription.List)
 // Reader reads and handles data from a specific connection
 func (w *Websocket) Reader(ctx context.Context, conn Connection, handler func(ctx context.Context, message []byte) error) {
 	defer w.Wg.Done()
+	var reporter ProcessReporter
+	if w.processReporter != nil {
+		reporter = w.processReporter.New(conn)
+	}
 	for {
 		resp := conn.ReadMessage()
+		readAt := time.Now()
```
gloriousCode, line 1084:
```suggestion
var readAt time.Time
if reporter != nil {
	readAt := time.Now()
}
```
`40.01 ns/op` vs `0.3179 ns/op` for those without a processReporter on a hot path

It is a shame gorilla doesn't have time received per message so we wouldn't have to call it.

engine/websocketroutine_manager.go
```diff
@@ -134,6 +134,10 @@ func (m *WebsocketRoutineManager) websocketRoutine() {
 			continue
 		}
 
+		if m.verbose {
```
gloriousCode, line 137: I'd be cool with this being separate from `verbose` along the lines of a config field ala `WebsocketResponseMaxLimit` per exchange, that way you don't have to log everything to get some stats
Plus it also then doesn't have the weird dependency on websocketroutine_manager. 
Plus Plus you only would need to actually add the field to gateio rn

exchanges/stream/reporting.go
```diff
@@ -0,0 +1,104 @@
...
+			return
+		case <-ticker.C:
+			r.m.Lock()
+			if r.operations > 0 {
+				avgOperationsPerSecond := r.operations / 60
+				avgProcessingTime := r.totalProcessingTime / time.Duration(r.operations)
+				peakTime := r.peakProcessingTime
+
+				// Reset metrics for the next interval.
+				r.operations, r.totalProcessingTime, r.peakProcessingTime = 0, 0, 0
```
gloriousCode, line 93: I think there'd be value in not resetting some so you can know the slowest ever etc. Trudging through logs looking for the highest number doesn't sound so good

gloriousCode (changes requested):
Looks good!
```
[DEBUG] | WEBSOCKET | 04/12/2024 09:41:05 | wss://op-ws.gateio.live/v4/ws: Operations/Second: 0, Avg Processing/Operation: 124.162µs, Peak: 786.542µs
...
[DEBUG] | WEBSOCKET | 04/12/2024 09:42:05 | wss://op-ws.gateio.live/v4/ws: Operations/Second: 1, Avg Processing/Operation: 127.424µs, Peak: 317.791µs
```

exchanges/stream/reporting.go
```diff
@@ -0,0 +1,104 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-r.ch:
+			return
+		case <-ticker.C:
+			r.m.Lock()
+			if r.operations > 0 {
+				avgOperationsPerSecond := r.operations / 60
```
gloriousCode, line 88: Probably make that a float, now looking at my output above with 0 operations per second

gloriousCode (changes requested):
I like this, thank you!

exchanges/stream/websocket.go
```diff
@@ -203,6 +203,10 @@ func (w *Websocket) Setup(s *WebsocketSetup) error {
 	w.setState(disconnectedState)
 
 	w.rateLimitDefinitions = s.RateLimitDefinitions
+
+	if s.ExchangeConfig.WebsocketMetricsLogging {
+		w.processReporter = NewDefaultProcessReporterManager()
```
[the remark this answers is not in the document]
shazbert: epic 🔥

exchanges/stream/reporting.go
```diff
@@ -0,0 +1,119 @@
...
+}
+
+// ProcessReporter defines an interface for reporting processed data from a connection
+type ProcessReporter interface {
+	// Report logs the processing time for a received data packet and updates metrics.
+	// read is the time the data was read from the connection.
+	// data is the raw data received from the connection.
+	// err is any error that occurred while processing the data.
+	Report(read time.Time, data []byte, err error)
+	// close closes the process reporter and handles any cleanup.
```
gloriousCode, line 23:
```suggestion
	// Close closes the process reporter and handles any cleanup.
```

exchanges/stream/reporting.go
```diff
@@ -0,0 +1,119 @@
...
+				peakCause := r.peakCause
+				errors := r.errors
+				// Reset metrics for the next interval.
+				r.operations, r.totalProcessingTime, r.peakProcessingTime, r.peakCause, r.errors = 0, 0, 0, nil, 0
+				r.m.Unlock()
+				if len(peakCause) > 100 {
+					peakCause = append(peakCause[:100], []byte("...")...)
+				}
+				// Log metrics outside of the critical section to avoid blocking other threads.
+				log.Debugf(log.WebsocketMgr, "Connection: %v Operations/Second: %.2f, Avg Processing/Operation: %v, Errors: %v Peak: %v Cause: %v...", conn.GetURL(), avgOperationsPerSecond, avgProcessingTime, errors, peakTime, string(peakCause))
```
gloriousCode, line 113:
```suggestion
					peakCause = append(peakCause[:100], []byte("...")...)
				}
				// Log metrics outside of the critical section to avoid blocking other threads.
				log.Debugf(log.WebsocketMgr, "Connection: %v Operations/Second: %.2f, Avg Processing/Operation: %v, Errors: %v Peak: %v Cause: %v", conn.GetURL(), avgOperationsPerSecond, avgProcessingTime, errors, peakTime, string(peakCause))
```
I removed the ellipses in the main output because you add them if its > 100 and doesn't really mean much if < 100

gloriousCode (approved):
Nice! tACK! Optional minor changes that don't affect tackiness

exchanges/stream/reporting.go
shazbert: done

config/versions/v4.go
```diff
@@ -0,0 +1,39 @@
+package versions
+
+import (
+	"context"
+	"encoding/json"
```
gloriousCode, line 5: killed by your own linter

exchanges/stream/reporting.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+	timer := time.NewTimer(time.Until(time.Now().Truncate(period).Add(period)))
+	defer timer.Stop()
+
+	for {
+		select {
+		case <-r.ch:
+			return
+		case <-timer.C:
+			timer.Reset(time.Until(time.Now().Truncate(period).Add(period)))
```
gloriousCode, line 98: I think it would be better practice to reset the timer after the operation is done. Not that it should really impact with this kind of code especially the default time wait of 1 minute.

gloriousCode (changes requested):
ty4changes

config/versions/v4.go
shazbert: 💀

exchanges/stream/reporting.go
shazbert: [Done](https://github.com/thrasher-corp/gocryptotrader/pull/1734/commits/e9cbcd334557ae216ea399207be5ec7f185fadef)

gloriousCode (approved):
tACK!

gbjk (commented):
Good work 👍 

Initial bigger architecture comments:
* reporter doesn't feel like it should be in stream.
We're create "websocket metrics" but I think we should have "metrics, of which websocket is one contributor"

Can we consider moving the reporter to `internal/metrics` ?

`ProcessReporter` is a bit weird, because nobody's going to find that and understand what it is without reading docs, `Process` is everything.
I'll provide a better suggestion inside a real review, but I'm thinking that:
* overall collector: `metrics.NewCollector()`
* Reporting things: `span := mc.Start()` and `span.End()` should JDTRT
* A single duration logged with `mc.Collect(mc.NewSpan(event: "db.compact", Duration: took))`

These names are alligning our naming with what other metric systems generally use, so it's easier to understand, and also so we don't re-invent conceptual wheels.

* Plugging into reader:
Given that I'm suggesting the bulk of metrics moves outside of websocket/stream, obviously we still need to plug it in/adapt it.
We could keep it exactly as-is. It's very light weight when turned off, and that's good.

Alternative suggestion:
Have an interface for generic stream handlers.
The purpose of this is to avoid adding bespoke logic for every type of stream processor.
The normal "send it to the exchange" _could_ be moved to be one of them, or we leave it as a special case.
They all get `PreProcess`, `Process` and `PostProcess`.
A metrics collector just gets added to them, and it runs in PreProcess and PostProcess.

* I considered suggesting we should be using opentelemetry instead of rolling our own. However that would create a big dependency impact, and I'm pretty sure it's unacceptable exposure. ~I'll check.~
Whoa. [Surprisingly small dependency footprint](https://pkg.go.dev/go.opentelemetry.io/otel)
5k imports in it, and 10k import it!

* Assuming we don't use otel, then we should probably use otel 😭 What I mean is: We should be able to send metrics out separately from logging, to a receiver in a common format, so that we can run one of Grafana, Prometheus, Influx, Google Cloud Metrics or whatever alongside it.

* I don't think this is realistic yet, but I think we need to categorise the requests for meaningful metrics. Specifically, that'd mean that we start a processing run, and then things way down the line can add information to the metric span being reported.
Our exchange wsHandleData processes could then add what event this is once they've worked it out, and we could aggregate by that.

~I'll get more into it shortly.~
I'll wait for your comment before reviewing more

shazbert:
> Initial bigger architecture comments:
> 
> * reporter doesn't feel like it should be in stream.
>   We're create "websocket metrics" but I think we should have "metrics, of which websocket is one contributor"
> 
> Can we consider moving the reporter to `internal/metrics` ?

I don't mind doing that there is bound to be more metrics that are needed across everything.

> `ProcessReporter` is a bit weird, because nobody's going to find that and understand what it is without reading docs, `Process` is everything. I'll provide a better suggestion inside a real review, but I'm thinking that:
> 
> * overall collector: `metrics.NewCollector()`
> * Reporting things: `span := mc.Start()` and `span.End()` should JDTRT
> * A single duration logged with `mc.Collect(mc.NewSpan(event: "db.compact", Duration: took))`
> 
> These names are alligning our naming with what other metric systems generally use, so it's easier to understand, and also so we don't re-invent conceptual wheels.

Happy with this as well. 

> * Plugging into reader:
>   Given that I'm suggesting the bulk of metrics moves outside of websocket/stream, obviously we still need to plug it in/adapt it.
>   We could keep it exactly as-is. It's very light weight when turned off, and that's good.

When you are doing a review of this direction can you give me a design for this on how we can drop it in?

> Alternative suggestion: Have an interface for generic stream handlers. The purpose of this is to avoid adding bespoke logic for every type of stream processor. The normal "send it to the exchange" _could_ be moved to be one of them, or we leave it as a special case. They all get `PreProcess`, `Process` and `PostProcess`. A metrics collector just gets added to them, and it runs in PreProcess and PostProcess.

sounds good.

> * I considered suggesting we should be using opentelemetry instead of rolling our own. However that would create a big dependency impact, and I'm pretty sure it's unacceptable exposure. ~I'll check.~
>   Whoa. [Surprisingly small dependency footprint](https://pkg.go.dev/go.opentelemetry.io/otel)
>   5k imports in it, and 10k import it!
> * Assuming we don't use otel, then we should probably use otel 😭 What I mean is: We should be able to send metrics out separately from logging, to a receiver in a common format, so that we can run one of Grafana, Prometheus, Influx, Google Cloud Metrics or whatever alongside it.
> * I don't think this is realistic yet, but I think we need to categorise the requests for meaningful metrics. Specifically, that'd mean that we start a processing run, and then things way down the line can add information to the metric span being reported.
>   Our exchange wsHandleData processes could then add what event this is once they've worked it out, and we could aggregate by that.
> 
> ~I'll get more into it shortly.~ I'll wait for your comment before reviewing more


Yeah sounds good on this level as well, depending on other thoughts. I use Grafana and Prometheus a lot and it's very helpful; even just shadowing `otel` as an initial concept might be a direction? But I have never used `Otel` so can't reaaaly comment on design. At the moment I am not using a bespoke only the default implementation for logs, but I was going to register my own with method `SetProcessReportManager` and using that to hook into Prometheus counters. If it has one place where I can register or set that, that would be sweet. If I am understanding correctly. 

Also should we open a fresh PR with that instead and I will close this as I am using this externally?

gbjk:
Nice; Sounds like we're on the same page!

> Also should we open a fresh PR with that instead and I will close this as I am using this externally?

Up to you; I'd probably close it and open a new one 😄

shazbert:
Closed but will keep this updated with trading branches until #2026

PR #1735, opened 2024-11-29
Config: AssetEnabled upgrade

gbjk:
Config:
* Moves AssetEnabled from being a bool pointer to just being a bool

Exchanges: 
* Rename `StoreAssetPairFormat` to `StoreAssetPairStore`
* StorePairs(enabled) set assetEnabled if pairs list is not empty
* StorePairs(enabled) add pairs to Available as well
  * This is a convenience for all the test that have to do both, as well as users. Can be dropped if contentious
* Fix error formatting for DisableAssetWebsocketSupport logging
* Various assertifications in tests
* Simplify exchange's defaultPairFormatting handling and make it somewhat consistent.
  * Could do more, but there's a larger overhaul of this coming
* `SetAssetEnabled` will not error on no-op (already disabled or enabled)

Bitfinex:
* Fix tests for MarginFunding subs

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


shazbert (changes requested):
Quick test on gateio, when I set everything but spot assets to false for asset enabled it still fetches and subscribes all assets. Doesn't seem to be happening on master, once resolved I will do a more in-depth check.

gbjk:
> Quick test on gateio, when I set everything but spot assets to false for asset enabled it still fetches and subscribes all assets. Doesn't seem to be happening on master, once resolved I will do a more in-depth check.

Fixed https://github.com/gbjk/gocryptotrader/commit/65c05a063

currency/manager_test.go
```diff
@@ -463,72 +462,54 @@ func TestEnablePair(t *testing.T) {
...
-	if err != nil {
-		t.Fatal(err)
-	}
+	assert.NoError(t, err)
 
 	err = p.SetAssetEnabled(asset.Spot, false)
-	if err == nil {
-		t.Fatal("unexpected result")
-	}
+	assert.NoError(t, err, "Setting to disabled twice should not error")
```
shazbert, line 493: Suggestion: I usually return an error on no-op because I find it easier to track my own whoopsies, I also like the fact if I change it by CLI it then returns that it failed. I do see the idempotency aspect of the new behaviour though. Keep this unresolved for other peoples input if you want to pass on this.

engine/rpcserver_test.go
```diff
@@ -1534,97 +1533,53 @@ func TestCheckVars(t *testing.T) {
...
-	if err != nil {
-		t.Error(err)
-	}
-	err = e.GetBase().StoreAssetPairFormat(asset.USDTMarginedFutures, usdtFutures)
-	if err != nil {
-		t.Error(err)
+	b := e.GetBase()
+
+	for _, a := range []asset.Item{asset.Spot, asset.Margin, asset.CoinMarginedFutures, asset.USDTMarginedFutures} {
+		fmt := currency.PairStore{
```
shazbert, line 1550: `fmt` conflicts with a package name maybe just `ps`?

exchanges/bitmex/bitmex_wrapper.go
```diff
@@ -42,37 +42,22 @@ func (b *Bitmex) SetDefaults() {
...
-	if err != nil {
-		log.Errorln(log.ExchangeSys, err)
+	for _, a := range []asset.Item{asset.Spot, asset.PerpetualContract, asset.Futures, asset.Index} {
+		ps := currency.PairStore{
+			AssetEnabled:  true,
+			RequestFormat: &currency.PairFormat{Uppercase: true},
+			ConfigFormat:  &currency.PairFormat{Uppercase: true, Delimiter: currency.DashDelimiter},
+		}
+		if a == asset.Spot {
+			ps.RequestFormat = &currency.PairFormat{Uppercase: true, Delimiter: currency.UnderscoreDelimiter}
```
shazbert, line 52:
```suggestion
			ps.RequestFormat.Delimiter = currency.UnderscoreDelimiter
```

shazbert (changes requested):
great cleanup! tested and runs great. Just some minor nits.

cmd/exchange_template/wrapper_file.tmpl
```diff
@@ -55,22 +55,24 @@ func ({{.Variable}} *{{.CapitalName}}) SetDefaults() {
...
+		AssetEnabled:  true,
 		RequestFormat: &currency.PairFormat{Uppercase: true},
 		ConfigFormat:  &currency.PairFormat{Uppercase: true, Delimiter: ":"},
 	}
 
-	err = {{.Variable}}.StoreAssetPairFormat(asset.Spot, fmt1)
+	err = {{.Variable}}.StoreAssetPairStore(asset.Spot, fmt1)
 	if err != nil {
-		log.Errorln(log.ExchangeSys, err)
+		log.Errorf(log.ExchangeSys, "%s error storing `%s` default asset formats: %s", {{.Variable}}.Name, asset.Spot, err)
```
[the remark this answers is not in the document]
gbjk: The whole direction here was the opposite of this change request.
The ideology I'm following is that functions return the error they encounter, without their own context.
If you call a go standard library function, you'll get the same thing.
It's down to the caller to *decide* if it wants to give context to that error, and what that context looks like from it's perspective.
From the perspective of the function caller, they know what they called, and so the error returned should be undecorated.

Case in point: `StoreAssetPairStore` is not `default asset formats` so these callers all needed to clarify the context they were calling that function in.

This is why you'll often see me decorate errors one level *above* where GCT has previously been doing it.
Sometimes it can feel slightly weighty, in this example where we copy the same context everywhere.

We could definitely add a common error for `errStoringDefaultAssetFormats` though, and I think that's warranted.
I'll circle back to this after the others and see if you've replied :)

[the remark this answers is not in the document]
gbjk: Adding one more bit of context:
This implies that functions must also add their own context.
e.g. If StoreAssetPairStore encountered some error interacting with a database, like "Record not found", We'd expect that error to come *without* context.
So `StoreAssetPairStore` should add the context `error searching for glove size in database: %w`.

config/config_test.go
```diff
@@ -754,123 +745,51 @@ func TestCheckPairConsistency(t *testing.T) {
...
+	assert.Empty(t, pm.Pairs[asset.Spot].Enabled, "Unavailable pairs should be removed from enabled")
 
 	// Test that enabled pair is not found in the available pairs
-	c.Exchanges[0].CurrencyPairs.Pairs[asset.Spot].Available = currency.Pairs{
-		p1,
-	}
+	pm.Pairs[asset.Spot].Available = currency.Pairs{p1}
 
 	// LTC_USD is only found in the available pairs list and should therefore
 	// be added to the enabled pairs list due to the atLestOneEnabled code
```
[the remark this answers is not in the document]
gbjk: Imma fixup into 4d32d83882 cos I obviously can't fixup into 14c72c9c6b

Fixed https://github.com/gbjk/gocryptotrader/commit/36e3dffbc

currency/manager_test.go
gbjk: I believe if something wants to error on it already being disabled/enabled, that's *it's* problem.
The responsibility of this function is to do what it's told 😃 
So I'm happy to add a check in api server or CLI for you to see if it's already enabled/disabled and error *there*.
Would you like that?

engine/rpcserver_test.go
gbjk: Agreed, Thanks! 🙏 

Fixed https://github.com/gbjk/gocryptotrader/commit/d6c896634

exchanges/bitmex/bitmex_wrapper.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/bfa72e20a

currency/manager_test.go
shazbert: No thanks, if it's going to stay like this, having that check would need to go to all potential API calls cli, scripting etc. State change responsibility is not a major concern 🤷 just easier here but it's such a minor thing.

cmd/exchange_template/wrapper_file.tmpl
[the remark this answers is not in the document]
shazbert: From an implementation perspective, maintaining this standard can introduce toil during implementation and reviews, especially since any failure in SetDefaults would inherently cause the exchange to malfunction. Differentiating between errors in this context seems redundant. Still suggest wrapping asset and exchange for the exchange method. 
But I will leave this open for other peoples perspective. @thrasher- @gloriousCode

shazbert (approved):
Current changes tACK. 🚀

config/versions/v3.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+	if err = jsonparser.ObjectEach(bytes.Clone(e), assetEnabledFn, "currencyPairs", "pairs"); err != nil {
+		return e, fmt.Errorf("%w: %w", errUpgradingCurrencyPairs, err)
+	}
+	e = jsonparser.Delete(e, "currencyPairs", "assetTypes")
+	return e, err
+}
+
+// DowngradeExchange moves AssetEnabled assets into AssetType field
+func (v *Version3) DowngradeExchange(_ context.Context, e []byte) ([]byte, error) {
```
gloriousCode, line 63: This is not part of a full review yet. So I was accidentally on your other branch which upgrades config to v4. On switching to this branch, it runs downgrade. This downgrade here is about removing the effects of v3. GCT recovers from this:
```
Okx assets disabled, turning on asset perpetualswap
Binanceus assets disabled, turning on asset spot
Binance assets disabled, turning on asset spot
etc etc
```
I don't think I understand how `Downgrade` works at present. It is logical that it can unwind its own changes. I do not understand why the v3 downgrade is executed going from v4->v3. I get that by changing the branch, it is impossible for v4 downgrade to run, but why is v3 downgrade executing? After that is run, it would theoretically be considered v2, however the config regards it as v3 and no v3 upgrade has run. I also understand that the switching branches and removing versions is outside of normal intended operations.

Lot's of text sorry. Am I correct in thinking that v3 downgrade should not execute on the downgrade from v4-v3?

gbjk: You're completely correct.

Will be addressed in #1769 and associated PR

gbjk:
I'm going to rebase this onto #1770, because I think that's kinda necessary for `Downgrade` to be testable.

gbjk:
Clarification: 1770 makes this *manually* testable for Downgrade, which is what caught @gloriousCode out.
The unit tests work before and after.

exchanges/exchange.go
```diff
@@ -309,68 +309,56 @@ func (b *Base) SetCurrencyPairFormat() error {
...
 		}
+	}
+	return nil
+}
 
-		err = b.CurrencyPairs.StorePairs(assetTypes[x], cfgPS.Available, false)
-		if err != nil {
+// setConfigPairFormatFromExchange sets the config formats from the exchange's format
+// This deprecated behaviour will be removed, because config should not be backloaded from runtime
+func (b *Base) setConfigPairFormatFromExchange(a asset.Item) error {
```
gloriousCode, line 350: Add test please 🌈 🌞

exchanges/exchange.go
```diff
@@ -309,68 +309,56 @@ func (b *Base) SetCurrencyPairFormat() error {
...
 		}
-
-		err := b.CurrencyPairs.SetAssetEnabled(assetTypes[x], enabledAsset)
-		// Suppress error when assets are enabled by default and they are being
-		// enabled by config. A check for the inverse
-		// e.g. currency.ErrAssetAlreadyDisabled is not needed.
-		if err != nil && !errors.Is(err, currency.ErrAssetAlreadyEnabled) {
-			return err
+		if !b.Config.CurrencyPairs.UseGlobalFormat {
+			// TODO: Should probably be in SetCurrencyPairFormat. See #1748
```
gloriousCode, line 318: _should_? _probably_? No confidence in this direction? Remove that double-weak wording 💪

config/versions/v4_test.go
```diff
@@ -0,0 +1,77 @@
...
+func TestVersion4Upgrade(t *testing.T) {
+	t.Parallel()
+
+	_, err := new(Version4).UpgradeExchange(context.Background(), []byte{})
+	require.ErrorIs(t, err, errUpgradingAssetTypes)
+
+	_, err = new(Version4).UpgradeExchange(context.Background(), []byte(`{}`))
+	require.ErrorIs(t, err, errUpgradingCurrencyPairs)
+
+	in := []byte(`{"name":"Cracken","currencyPairs":{"assetTypes":["spot"],"pairs":{"spot":{"enabled":"BTC-AUD","available":"BTC-AUD"},"futures":{"assetEnabled":true},"options":{}}}}`)
```
gloriousCode, line 32: Given that the value in the config can be `"assetEnabled":null` and not just not-present, I think it should part of the test
```
error applying version upgrade to 4: error modifying exchange config: error upgrading currencyPairs.pairs: Value is not a boolean: null
```

gloriousCode (changes requested):
Looks good for the most part. Looking forward to it being rebased so that I can verify with the downgrade bugfix

gbjk:
> Looks good for the most part. Looking forward to it being rebased so that I can verify with the downgrade bugfix

It was already rebased for you, when I removed restructuring 😄  At least I hope.
But I'll rebase once more to resolve conflicts. And sniff it's still working correctly.

cmd/exchange_template/wrapper_file.tmpl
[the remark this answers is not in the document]
gbjk: That's an interesting point about how GCT is a system.
I'd be taking a completely different perspective if GCT was more closed, and the paths to these functions were all private, and callers knew to log it without decoration.

exchanges/exchange.go
```diff
@@ -1004,37 +985,26 @@ func (b *Base) EnableRateLimiter() error {
 	return b.Requester.EnableRateLimiter()
 }
 
-// StoreAssetPairFormat initialises and stores a defined asset format
-func (b *Base) StoreAssetPairFormat(a asset.Item, f currency.PairStore) error {
+// StoreAssetPairStore initialises and stores a defined asset format
+func (b *Base) StoreAssetPairStore(a asset.Item, f currency.PairStore) error {
```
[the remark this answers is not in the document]
gbjk: Yeah. I saw this recently and wondered what the hell I was thinking.

Fixed https://github.com/gbjk/gocryptotrader/commit/be22c6e30f

Sorry that it's a rebase, but the fastest way to do this was with a `git rebase` and a patch regex.

exchanges/exchange.go
gbjk: Impossible to get to those nested err situations, both in tests, and in runtime, because we're detecting the nils already, and bad asset or nil-guarding is all that StoreFormat would error on.
Nevermind. I've done the best I can.
Definitely don't want to re-inline this.

Fixed https://github.com/gbjk/gocryptotrader/commit/3bf8af59f3

exchanges/exchange.go
gbjk: 🫡 

Fixed https://github.com/gbjk/gocryptotrader/commit/d01aa79f92

config/versions/v4_test.go
gbjk: > I think it should part of the test

Defaults to false, since that's what null would mean.

Fixed https://github.com/gbjk/gocryptotrader/commit/26cb329223

> Speaking of unlabelled errors, which exchange? which asset?

Fixed https://github.com/gbjk/gocryptotrader/commit/0a864c2963
Fixed https://github.com/gbjk/gocryptotrader/commit/65b768612f

Specifically dealt with other data types as well, since that would have silently passed.

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -156,6 +156,7 @@ Similar to the configs, spot support is inbuilt but other asset types will need
 
 ```go
 	spot := currency.PairStore{
+        AssetEnabled:  true,
 		RequestFormat: &currency.PairFormat{
```
shazbert, line 160:
```suggestion
                AssetEnabled:  true,
		RequestFormat: &currency.PairFormat{
```

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -166,6 +167,7 @@ Similar to the configs, spot support is inbuilt but other asset types will need
 		},
 	}
 	futures := currency.PairStore{
+        AssetEnabled:  true,
 		RequestFormat: &currency.PairFormat{
```
shazbert, line 171:
```suggestion
                AssetEnabled:  true,
		RequestFormat: &currency.PairFormat{
```

shazbert (changes requested):

docs/ADD_NEW_EXCHANGE.md
gbjk: Thanks! Deja-vu

Fixed https://github.com/gbjk/gocryptotrader/commit/30c73ee08e

docs/ADD_NEW_EXCHANGE.md
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/30c73ee08e

shazbert (approved):
Nothing more from me!

exchanges/exchange.go
gloriousCode: Thank you! 🌞

config/versions/v4.go
```diff
@@ -0,0 +1,87 @@
...
+// DowngradeExchange moves AssetEnabled assets into AssetType field
+func (v *Version4) DowngradeExchange(_ context.Context, e []byte) ([]byte, error) {
+	assetTypes := []string{}
+
+	assetEnabledFn := func(asset []byte, v []byte, _ jsonparser.ValueType, _ int) error {
+		if b, err := jsonparser.GetBoolean(v, "assetEnabled"); err == nil {
+			if b {
+				assetTypes = append(assetTypes, fmt.Sprintf("%q", asset))
+			}
+			e = jsonparser.Delete(e, "currencyPairs", "pairs", string(asset), "assetEnabled")
```
gloriousCode, line 78: I don't quite understand why this removes the field for every exchange. Why not keep the value?

gloriousCode (changes requested):
Thank you for making those changes, just one question

config/versions/v4.go
gbjk: Version consistency.
If you downgrade, and it's left, then user would be left not knowing which was correct at the current version. They'd end up changing both, at best, or the wrong one at worst.
Do you disagree?

gloriousCode: If `null` becomes `false` and `true` remains `true` during upgrade, couldn't only `true` remain during downgrade?

gbjk: We discussed, and I think we're on the same page.
Partly, this is about idempotent upgrades and downgrades; If we upgrade twice, assetEnabled should remain there.
@gloriousCode 's concern was that upgrades takes the existing value of assetEnabled and translate null to false.
The reason for that is that this version is retrospective, and that we want defined and consistent behaviour on re-upgrading. So we need to deal with nulls and convert them to false.

So I think we're good.
TL;DR:
* Upgrade and downgrade must be Idempotent; No effect re-running
* Retrospective nature of this means it's possible it will encounter nulls, and it's good that it converts them
* Downgrade moving to a state with the old field removed does not create an unexpected situation

currency/manager_test.go
thrasher-: I generally prefer the first approach, since I'd consider something calling something already enabled or disabled an issue on the caller and that they should fix their app. Why are you enabling something which is already enabled? This change is less precise but works. With that said, I'm not too fussed about this anyway to request a change back but just thought I'd mention this

gloriousCode: I'd normally prefer with error, then it can then be up to the caller to ignore it, rather than your suggestion to add additional code everywhere else. I see it as a way of highlighting poor code as well as minimising all the additional checks of `if isEnabled() { return err } else { enable() }` . 

I've checked all instances of how this is called and I would not say that this specific instance has negative impacts so I'm fine with this change

config/versions/v4.go
gloriousCode: My concern was in the other direction. When downgrading, this PR deletes existing fields rather than translating the value.
However, even if we were to only parse the `true` values when downgrading, GCT on running will still come in and add `true` fields to everything else on startup on v3. So I feel this instance is okay. I think `downgrade` should be able to transition between the versions just like `upgrade` and not just delete entire fields

config/versions/v4.go
```diff
@@ -0,0 +1,87 @@
...
+		}
+		return nil
+	}
+	err := jsonparser.ObjectEach(bytes.Clone(e), assetEnabledFn, "currencyPairs", "pairs")
+	if err == nil {
+		e, err = jsonparser.Set(e, []byte(`[`+strings.Join(assetTypes, ",")+`]`), "currencyPairs", "assetTypes")
+	}
+	return e, err
```
gloriousCode, line 86:
```suggestion
	if err := jsonparser.ObjectEach(bytes.Clone(e), assetEnabledFn, "currencyPairs", "pairs"); err != nil {
		return nil, err
	}
	return jsonparser.Set(e, []byte(`[`+strings.Join(assetTypes, ",")+`]`), "currencyPairs", "assetTypes")
```
# 🙏 🙏 🙏

gloriousCode (approved):
tACK! Thanks for making those changes. One strong suggestion to get away from that practice 😄

config/versions/v4.go
gbjk: Yeah. That's just flat out better 👍 

Fixed https://github.com/gbjk/gocryptotrader/commit/87517e107c

thrasher- (approved):
Nice work @gbjk and thanks for the drive-by cleanups too 👍

PR #1752, opened 2024-12-18
Coinbase: Skip deprecated APIs

gbjk:
Skip erroring deprecated APIs.
Bare minimum API. Skip all or remove exchange are options, but this too seconds.

DriveBy: error assertion for GetHistoricTrades. Accidentally mowed it down at a pedestrian crossing.

## Type of change


gbjk:
Closing in favour of an upcoming PR which also fixes GateIO

Issue #1323, opened 2023-08-16
Bitfinex: Switch WS type assertions 


gbjk:
See [this request](https://github.com/thrasher-corp/gocryptotrader/pull/1317/files#r1295288979) during PR #1317 1317

I think we'd use one of:

[array to struct](https://eagain.net/articles/go-json-array-to-struct/)
And there's a note from @thrasher- to consider https://github.com/thrasher-corp/gocryptotrader/commit/b2dfef87989d1ecbe35822e0cd973bb47ada150a

PR #1754, opened 2024-12-18
Bitfinex: Fix WS trade processing

### LINKED ISSUE #1746: Bitfinex: Fix WS Trade stream erroring
The public trade stream in bitfinex websocket now contains a new final field in the snapshot update:
Index | Field | Type | Description
-- | -- | -- | --
[0] | CHANNEL_ID | Int | Identification number assigned to the channel for the duration of this connection.
[1] | SNAPSHOT | Array | Array with an array of recent trades (Indices [0...n] will be trades)
[1][0...n] | TRADE | Array | Trade array or funding trade array

Note that the documentation is, of course, wrong, and that final field is actually just an int:
```
[8,[[1688909434,1733893263491,0.0000858,97680],...],1]
```

This triggers an error:
`exchange Bitfinex websocket error - unhandled WS trade update event: `
( which goes to the DataHandler )

because [this code](https://github.com/thrasher-corp/gocryptotrader/blob/ca62b0a4c/exchanges/bitfinex/bitfinex_websocket.go#L935-L937) expects to use the event_type from the other format.

Our test avoids seeing this by mocking the input to the trade stream in the old state.

Probably consider #1323 at the same time

gbjk:
* Bitfinex
  * Fixes `exchange Bitfinex websocket error - unhandled WS trade update event:`
  * Add tests for TestWSAllTrades
  * Add handling for public marginFunding trades
  * Fix idiosyncratic naming using `TestWs` instead of `TestWS`
  * Fix naming for wsTradeUpdated ( was wsTradeExecutionUpdate )
  * Fix trades not going to DataHandler unless SaveTrade was enabled
  * Enable tradeFeed for tests
* Linter
  * Disables shadow checks for err and ctx. Advocated this before, but had my fill of stylecheck (correctly, IMO) fighting with shadow over this rule. This isn't strictly necessary, but it's nearly christmas again.
  * Remove coploopvar since we're pinned > 1.22 and it's unnecessary

Fixes #1746 and #1323 

## Type of change


exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -869,139 +872,99 @@ func (b *Bitfinex) handleWSTickerUpdate(c *subscription.Subscription, d []interf
...
+func (b *Bitfinex) handleWSPublicTradesSnapshot(respRaw []byte) (trades []*wsTrade, errs error) {
+	handleTrade := func(v []byte, valueType jsonparser.ValueType, _ int, _ error) {
+		if valueType != jsonparser.Array {
+			errs = common.AppendError(errs, fmt.Errorf("%w `tradesSnapshot[1][*]`: %w `%s`", errParsingWSField, jsonparser.UnknownValueTypeError, valueType))
 		} else {
-			price, ok := data[3].(float64)
-			if !ok {
-				return errors.New("unable to type assert trade price")
+			t := &wsTrade{}
+			if err := json.Unmarshal(v, &[]any{&t.ID, &t.Timestamp, &t.Amount, &t.Price, &t.Period}); err != nil {
```
shazbert, line 941: Suggestion: can define an UnmarshalJSON method with this and use the type `wsTrade` here and below in `handleWSPublicTradeUpdate`.

shazbert (changes requested):
nice stuff, not much from me.

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -33,6 +33,10 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/log"
 )
 
+var (
+	errParsingWSField = errors.New("error parsing WS field")
+)
```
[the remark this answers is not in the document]
gbjk: Not so convinced, oddly, since it feels like this is a block of error vars, that happens to only contain one.
I want devs to add to it, essentially.

Did linter or editor highlight this?

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -869,139 +872,99 @@ func (b *Bitfinex) handleWSTickerUpdate(c *subscription.Subscription, d []interf
...
-			return errors.New("unable to type assert trade timestamp")
+		if w.Period != 0 {
+			t.AssetType = asset.MarginFunding
 		}
-		amount, ok := data[2].(float64)
-		if !ok {
-			return errors.New("unable to type assert trade amount")
+		if t.Amount < 0 {
+			t.Side = order.Sell
+			t.Amount *= -1
```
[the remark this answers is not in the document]
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/8b8da5cb0

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Oh, yeah, that's really cute, thank you 🙏 
I've put it in `_types` since it's a micro-method and there's nowhere moar organic to put it, other than maybe in `_websocket` next to the funcs that'll use it.

Fixed https://github.com/gbjk/gocryptotrader/commit/4d3a7e0a4

exchanges/bitfinex/bitfinex_websocket.go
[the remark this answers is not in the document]
shazbert: I am just reducing lines. Something like error blocks is usually the responsibility of the reviewers to maintain/enforce. But will leave it up to you if you want to commit that the suggestion.

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -869,139 +873,99 @@ func (b *Bitfinex) handleWSTickerUpdate(c *subscription.Subscription, d []interf
...
+		if t.Amount < 0 {
+			t.Side = order.Sell
+			t.Amount = math.Abs(t.Amount)
 		}
-		wsTrade := WebsocketTrade{
-			ID:        int64(tradeID),
-			Timestamp: int64(timestamp),
-			Amount:    amount,
+		if feedEnabled {
+			b.Websocket.DataHandler <- &t
```
shazbert, line 926: With verbose on; also we should send it over as a slice instead of individually. 
```
[WARN]  | WEBSOCKET | 29/12/2024 09:21:31 | Bitfinex websocket Unknown type: &{ID:00000000-0000-0000-0000-000000000000 TID:1695543964 Exchange:Bitfinex CurrencyPair:ETHUSD AssetType:spot Side:SELL Price:3414.9 Amount:1.01092617 Timestamp:2024-12-28 22:21:31.398 +0000 UTC}
```

shazbert (changes requested):
Thanks for addressing my last nits.

exchanges/bitfinex/bitfinex_types.go
```diff
@@ -488,16 +489,18 @@ type WebsocketBook struct {
...
-	Rate float64
-	// Funding offer period in days
-	Period int64
+	Price     float64
+	Period    int64 // Funding offer period in days
+}
+
+// UnmarshalJSON unmarshals json bytes into a wsTrade
+func (t *wsTrade) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &[]any{&t.ID, &t.Timestamp, &t.Amount, &t.Price, &t.Period})
```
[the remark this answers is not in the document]
gbjk: Yeah. That works.
Kinda expected it to panic, but goplay says not, and I guess you already tested.

Fixed https://github.com/gbjk/gocryptotrader/commit/d3406a654

.golangci.yml
```diff
@@ -150,6 +150,9 @@ issues:
     - text: "Expect WriteFile permissions to be 0600 or less"
       linters:
         - gosec
+    - text: 'shadow: declaration of "(err|ctx)" shadows declaration at'
```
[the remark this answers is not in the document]
gbjk: No reason specifically.
But the idea of erroring on shadowing context when you nearly *always* are going to shadow a context is insane, right?

exchanges/bitfinex/bitfinex_websocket.go
gbjk: This was  a very deliberate change from me.
Consumers of trade updates want to get trades, not slices of varying lengths of trades.
It makes a consumer's life unnecessarily harder to say: _Oh, I read something from a channel. Is it a thing, or a slice of those things?_
Unless the slice is integral, like a related tuple or an orderbook, I definitely believe we should send things, not slices of things.

I could go change engine to accept a pointer to a trade instead of a slice.
But at the same time, I don't want to introduce inconsistency across the exchanges.
So I'll revert my change, and then switch over to individual trades in a separate PR later.

Fixed https://github.com/gbjk/gocryptotrader/commit/48ebaf7be

.golangci.yml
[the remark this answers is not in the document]
shazbert: Well I am worried about issues for only context specifically and future dev implementation sillyness (me), say we put in a deadline or other val and I accidently shadow and it needs to propagate outside of a nested scope. I still prefer to explicit renaming of ctx -> ctxDiffName when nested. I will keep this unresolved for other peoples point of view.

exchanges/bitfinex/bitfinex_websocket.go
shazbert: Right that does make consumer logic more simple. I was more coming from the point of view reducing the channel contention which adds a little bit of latency between handovers. Your way is more maintainable maybe in future we can have a configurable layer which you can process both but for such an edge case, it's doubtful.

shazbert (approved):
tACK.

.golangci.yml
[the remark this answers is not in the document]
gbjk: It's normal to shadow inside a scope WithValue, and then pass that down to the things below.
I do this a lot normally, to add log fields to context for each iteration of a for loop across exchanges, for example.

Giving it a different name is _as_ likely to go wrong and have someone copy/paste/correct a line to just use `ctx`.
That's my biggest issue with variant naming of something as singular as context. You have a scope, and you have a context that's current. If you suddenly have `subCtx` and `ctx` it seems more likely that someone will mis-use `ctx` by copying a line across.

Also: shadowing linter is very inconsistent, and I linked to why, related to tags in scope when the shadow is seen. If we rely upon it to save us, things are going to fall through the cracks.

Note: Happy to remove ctx for now until I have a use-case that's pissing me off. I don't yet.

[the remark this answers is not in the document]
gbjk: Also raised by @gloriousCode in #1766 so removing ctx.

Fixed https://github.com/gbjk/gocryptotrader/commit/f86fff443

Had to rebase down to reword the commit

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -869,139 +873,99 @@ func (b *Bitfinex) handleWSTickerUpdate(c *subscription.Subscription, d []interf
...
-			tradeHolder = append(tradeHolder, wsTrade)
-		}
-	case 3:
-		if eventType != wsFundingTradeUpdate && eventType != wsTradeExecutionUpdate {
-			return fmt.Errorf("unhandled WS trade update event: %s", eventType)
-		}
-		data, ok := d[2].([]interface{})
-		if !ok {
-			return errors.New("trade data type assertion error")
+	var wsTrades []*wsTrade
```
gloriousCode, line 891: Please change this shadowed var name/the original string var name

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -869,139 +873,99 @@ func (b *Bitfinex) handleWSTickerUpdate(c *subscription.Subscription, d []interf
...
+	if b.IsSaveTradeDataEnabled() {
+		err = trade.AddTradesToBuffer(b.GetName(), trades...)
+	}
+	return err
+}
+
+func (b *Bitfinex) handleWSPublicTradesSnapshot(respRaw []byte) (trades []*wsTrade, errs error) {
+	handleTrade := func(v []byte, valueType jsonparser.ValueType, _ int, _ error) {
+		if valueType != jsonparser.Array {
+			errs = common.AppendError(errs, fmt.Errorf("%w `tradesSnapshot[1][*]`: %w `%s`", errParsingWSField, jsonparser.UnknownValueTypeError, valueType))
```
gloriousCode, line 939: This is a little bit extra given that there is only one caller to this and its behind a gate checking if its an array. I am normally one for all checks should be within the function, but its that you just checked to get to this function

gloriousCode: How about this simplification?
```go
func (b *Bitfinex) handleWSPublicTradesSnapshot(respRaw []byte) ([]*wsTrade, error) {
	var trades []*wsTrade
	err := json.Unmarshal(respRaw, &trades)
	return trades, err
}
```
Where `handleWSAllTrades` changes with:
```go
k, valueType, _, err := jsonparser.Get(respRaw, "[1]")
...
if wsTrades, err = b.handleWSPublicTradesSnapshot(k); err != nil {
```
A lot less tangling of additional jsonparser functions.

gloriousCode (changes requested):

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/3a52f3fd1 and https://github.com/gbjk/gocryptotrader/commit/0b3fdd6fe

exchanges/bitfinex/bitfinex_websocket.go
gbjk: > This is a little bit extra given that there is only one caller to this and its behind a gate checking if its an array. 

I think you might have missed that this is checking that every element inside the array is also an array.
That's the answer to why we're doing the second check.

And yeah. I think I didn't re-sniff simplifying this after applying shazbert's suggestion of the UnmarshalJSON on wsTrade. I think we can probably do exactly as you suggest. Will fix the test and confirm.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: I'm gonna uturn on this.
Fixing the test made me realise how really horrible this is.
Switching to sending trades as trades and accepting either in the engine.

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Yep. Exactly as you suggested. Extended to the handleWSPublicTradeUpdate as well.

Fixed https://github.com/gbjk/gocryptotrader/commit/82eb2c02d

exchanges/bitfinex/bitfinex_test.go
```diff
@@ -1382,17 +1372,44 @@ func TestWsOrderBook(t *testing.T) {
 	assert.ErrorIs(t, err, errNoSeqNo, "handleWSBookUpdate should send correct error")
 }
 
-func TestWsTradeResponse(t *testing.T) {
+func TestWSAllTrades(t *testing.T) {
```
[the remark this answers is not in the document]
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/e28ee04fb

exchanges/bitfinex/bitfinex_websocket.go
gbjk: Related https://github.com/gbjk/gocryptotrader/commit/e28ee04fb
This adds acceptance of individual trades to the websocket data ~manager~ repeater

gloriousCode (approved):
tACK! Thanks for the fixes, sorry for making it stale 🥖💥

shazbert (approved):
Changes ACK!

exchanges/bitfinex/bitfinex_test.go
```diff
@@ -1382,17 +1372,44 @@ func TestWsOrderBook(t *testing.T) {
...
+	for resp := range b.Websocket.DataHandler {
+		switch v := resp.(type) {
+		case trade.Data:
+			i := 6 - len(b.Websocket.DataHandler)
+			exp := trade.Data{
+				Exchange:     b.Name,
+				CurrencyPair: btcusdPair,
+			}
+			require.NoErrorf(t, json.Unmarshal([]byte(expJSON[i]), &exp), "Must not error unmarshalling json %d: %s", i, expJSON[i])
+			require.Equalf(t, exp, v, "Trade [%d] should be correct", i)
```
thrasher-, line 1403:
```suggestion
			require.Equalf(t, exp, v, "Trade [%d] must be correct", i)
```

exchanges/bitfinex/bitfinex_test.go
```diff
@@ -1258,67 +1260,58 @@ func TestSubToMap(t *testing.T) {
 	assert.Equal(t, "tBTCLTC", r["symbol"], "symbol should not use colon delimiter if both currencies < 3 chars")
 }
 
-// TestWsPlaceOrder dials websocket, sends order request.
-func TestWsPlaceOrder(t *testing.T) {
+// TestWSNewOrder exercises WsNewOrder
```
thrasher-, line 1263:
```suggestion
// WSNewOrder exercises WSNewOrder
```
For this one and the others, though we can rid these comments as they're self explanatory and the linter only complains on non-exported test code comments

thrasher- (changes requested):
Only minor things and then this looks good to go

exchanges/bitfinex/bitfinex_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/b7ac01c8a

exchanges/bitfinex/bitfinex_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/92686e35d

thrasher- (approved):

shazbert (approved):
Changes ACK!

PR #1765, opened 2024-12-31
Subscriptions: Fix IBotExchange not implementing sub.IExchange

gbjk:
Adds GetPairFormat to IBotExchange

Thusfar ExpandTemplates had only been called internally. All methods consumers might want to get, in this case GetPairFormat, need to be in the interface we're passing out, otherwise users can't call them (or things that use them, like ExpandTemplates)

## Type of change


shazbert (approved):
ch ACK.

gloriousCode (approved):
tACK!

thrasher- (approved):
ACK, thanks also for adding the test to ensure it conforms from this day onwards

PR #1623, opened 2024-08-20
encoding/json: Add custom JSON package with build tag support for Sonic

shazbert:
# PR Description

This pull request introduces a custom JSON package located in encoding/json/ to facilitate swapping between the default Go encoding/json package and the sonic library based on build tags. 

Adds:
* json.go (implementation using encoding/json).
* sonic.go (default implementation using sonic library).

Import Updates:

All imports to the standard encoding/json package have been updated to the new custom package github.com/thrasher-corp/gocryptotrader/encoding/json.

Added basic benchmarks for both the default encoding/json and the sonic implementation to evaluate *hot-swap* performance.

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1766, opened 2025-01-04
GateIO: Fix GetFuturesContractDetails for Deliveries and minor other fixes

gbjk:
# GateIO
* 🐛 🚨 Fix GetFuturesContractDetails for Deliveries
Was returning the product of all the contracts, so 1444 instead of 38 contracts.
Doesn't touch anything else around this; Futures asset split coming separately
* Rename Get*PersonalTradingHistory to GetMy*TradingHistory
* Remove duplicate DeliveryTradingHistory
* Rename GetMyPersonalTradingHistory to GetMyFuturesTradingHistory
* Rename GateIOGetPersonalTradingHistory to GetMySpotTradingHistory
* Rename GetSingleContract and GetSingleDeliveryContracts
* Fix GetOpenInterest returning asset.ErrNotEnabled
# Linter
* Disables shadow checks for err  ( Cherry-pick from #1754 )

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] GetFuturesContractDetails

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -150,6 +150,9 @@ issues:
     - text: "Expect WriteFile permissions to be 0600 or less"
       linters:
         - gosec
+    - text: 'shadow: declaration of "err" shadows declaration at'
+      linters: [ govet ]
+
 
   exclude-dirs:
     - vendor
--- a/exchanges/gateio/gateio_types.go
+++ b/exchanges/gateio/gateio_types.go
@@ -585,18 +585,17 @@ type Orderbook struct {
 
 // Trade represents market trade.
 type Trade struct {
-	ID           int64        `json:"id,string"`
-	TradingTime  types.Time   `json:"create_time"`
-	CreateTimeMs types.Time   `json:"create_time_ms"`
-	OrderID      string       `json:"order_id"`
-	Side         string       `json:"side"`
-	Role         string       `json:"role"`
-	Amount       types.Number `json:"amount"`
-	Price        types.Number `json:"price"`
-	Fee          types.Number `json:"fee"`
-	FeeCurrency  string       `json:"fee_currency"`
-	PointFee     string       `json:"point_fee"`
-	GtFee        string       `json:"gt_fee"`
+	ID          int64        `json:"id,string"`
+	CreateTime  types.Time   `json:"create_time_ms"`
+	OrderID     string       `json:"order_id"`
+	Side        string       `json:"side"`
+	Role        string       `json:"role"`
+	Amount      types.Number `json:"amount"`
+	Price       types.Number `json:"price"`
+	Fee         types.Number `json:"fee"`
+	FeeCurrency string       `json:"fee_currency"`
+	PointFee    string       `json:"point_fee"`
+	GtFee       string       `json:"gt_fee"`
 }
 
 // Candlestick represents candlestick data point detail.
@@ -688,13 +687,12 @@ type FuturesContract struct {
 
 // TradingHistoryItem represents futures trading history item.
 type TradingHistoryItem struct {
-	ID           int64        `json:"id"`
-	CreateTime   types.Time   `json:"create_time"`
-	CreateTimeMs types.Time   `json:"create_time_ms"`
-	Contract     string       `json:"contract"`
-	Text         string       `json:"text"`
-	Size         float64      `json:"size"`
-	Price        types.Number `json:"price"`
+	ID         int64        `json:"id"`
+	CreateTime types.Time   `json:"create_time_ms"`
+	Contract   string       `json:"contract"`
+	Text       string       `json:"text"`
+	Size       float64      `json:"size"`
+	Price      types.Number `json:"price"`
 	// Added for Derived market trade history data
 	Fee      types.Number `json:"fee"`
 	PointFee types.Number `json:"point_fee"`
@@ -1218,8 +1216,7 @@ type AccountBalanceInformation struct {
 // MarginAccountBalanceChangeInfo represents margin account balance
 type MarginAccountBalanceChangeInfo struct {
 	ID            string     `json:"id"`
-	Time          types.Time `json:"time"`
-	TimeMs        types.Time `json:"time_ms"`
+	Time          types.Time `json:"time_ms"`
 	Currency      string     `json:"currency"`
 	CurrencyPair  string     `json:"currency_pair"`
 	AmountChanged string     `json:"change"`
@@ -1388,10 +1385,8 @@ type SpotOrder struct {
 	Succeeded          bool         `json:"succeeded"`
 	ErrorLabel         string       `json:"label,omitempty"`
 	Message            string       `json:"message,omitempty"`
-	CreateTime         types.Time   `json:"create_time,omitempty"`
-	CreateTimeMs       types.Time   `json:"create_time_ms,omitempty"`
-	UpdateTime         types.Time   `json:"update_time,omitempty"`
-	UpdateTimeMs       types.Time   `json:"update_time_ms,omitempty"`
+	CreateTime         types.Time   `json:"create_time_ms,omitempty"`
+	UpdateTime         types.Time   `json:"update_time_ms,omitempty"`
 	CurrencyPair       string       `json:"currency_pair,omitempty"`
 	Status             string       `json:"status,omitempty"`
 	Type               string       `json:"type,omitempty"`
@@ -1452,8 +1447,7 @@ type CancelOrderByIDResponse struct {
 // SpotPersonalTradeHistory represents personal trading history.
 type SpotPersonalTradeHistory struct {
 	TradeID      string       `json:"id"`
-	CreateTime   types.Time   `json:"create_time"`
-	CreateTimeMs types.Time   `json:"create_time_ms"`
+	CreateTime   types.Time   `json:"create_time_ms"`
 	CurrencyPair string       `json:"currency_pair"`
 	OrderID      string       `json:"order_id"`
 	Side         string       `json:"side"`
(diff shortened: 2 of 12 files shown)


exchanges/gateio/gateio_wrapper.go
```diff
@@ -2161,25 +2151,22 @@ func (g *Gateio) GetFuturesContractDetails(ctx context.Context, item asset.Item)
...
+				IsActive:             !contracts[i].InDelisting,
 				Type:                 ct,
 				SettlementCurrencies: currency.Currencies{currency.USDT},
 				MarginCurrency:       currency.Code{},
-				Multiplier:           contracts[j].QuantoMultiplier.Float64(),
-				MaxLeverage:          contracts[j].LeverageMax.Float64(),
+				Multiplier:           contracts[i].QuantoMultiplier.Float64(),
+				MaxLeverage:          contracts[i].LeverageMax.Float64(),
 			}
-			resp = append(resp, contractsToAdd...)
```
gloriousCode, line 2182: https://www.youtube.com/watch?v=3w-2gUSus34
Use of `j` and this shows some missed copy paste action

.golangci.yml
```diff
@@ -150,6 +150,9 @@ issues:
     - text: "Expect WriteFile permissions to be 0600 or less"
       linters:
         - gosec
+    - text: 'shadow: declaration of "(err|ctx)" shadows declaration at'
```
gloriousCode, line 153: The hardest part for me is that my IDE does not have the nuance to not care about `err` shadow. I will have to get used to the green colour. Otherwise, godbolt shows it has no impacts 👍 
Second generalised PSA is that `ineffectual assignment to err (ineffassign)`/golang compiler will prevent any other bad usage of shadow err

gloriousCode, line 153: Could you please show me some context for the ctx usage within our codebase? If not, I argue its premature to introduce it. Wait for the PR where you use it and we can judge it on its merits then

gloriousCode (changes requested):
Thanks!

.golangci.yml
gbjk: I'll remove.
Duplicate discussion and reply [here](https://github.com/thrasher-corp/gocryptotrader/pull/1754/files#r1899248775)

Fixed https://github.com/gbjk/gocryptotrader/commit/c19f66c5a

Folded down immediately into https://github.com/gbjk/gocr yptotrader/commit/f21d44717 so I can rename the commit.

gloriousCode (approved):
tACK!

shazbert (changes requested):
Thanks for the update, Just some fields that have been left out:

```diff
diff --git a/exchanges/gateio/gateio_types.go b/exchanges/gateio/gateio_types.go
index 3bca9c0cc..1a91b5e24 100644
--- a/exchanges/gateio/gateio_types.go
+++ b/exchanges/gateio/gateio_types.go
@@ -679,16 +679,22 @@ type FuturesContract struct {
        OrdersLimit           int64        `json:"orders_limit"`
        TradeID               int64        `json:"trade_id"`
        OrderbookID           int64        `json:"orderbook_id"`
+       EnableBonus           bool         `json:"enable_bonus"`
+       EnableCredit          bool         `json:"enable_credit"`
+       CreateTime            types.Time   `json:"create_time"`
+       FundingCapRatio       types.Number `json:"funding_cap_ratio"`
+       VoucherLeverage       types.Number `json:"voucher_leverage"`
 }

 // TradingHistoryItem represents futures trading history item.
 type TradingHistoryItem struct {
-       ID         int64        `json:"id"`
-       CreateTime types.Time   `json:"create_time"`
-       Contract   string       `json:"contract"`
-       Text       string       `json:"text"`
-       Size       float64      `json:"size"`
-       Price      types.Number `json:"price"`
+       ID           int64        `json:"id"`
+       CreateTime   types.Time   `json:"create_time"`
+       CreateTimeMs types.Time   `json:"create_time_ms"`
+       Contract     string       `json:"contract"`
+       Text         string       `json:"text"`
+       Size         float64      `json:"size"`
+       Price        types.Number `json:"price"`
        // Added for Derived market trade history datas.
        Fee      types.Number `json:"fee"`
        PointFee types.Number `json:"point_fee"`
```

gbjk:
> Thanks for the update, Just some fields that have been left out:

Fixed https://github.com/gbjk/gocryptotrader/commit/1565c773b

shazbert (approved):
Clickity Clack Changes ACK.

exchanges/gateio/gateio_types.go
```diff
@@ -679,17 +679,23 @@ type FuturesContract struct {
...
-	ID         int64        `json:"id"`
-	CreateTime types.Time   `json:"create_time"`
-	Contract   string       `json:"contract"`
-	Text       string       `json:"text"`
-	Size       float64      `json:"size"`
-	Price      types.Number `json:"price"`
-	// Added for Derived market trade history datas.
+	ID           int64        `json:"id"`
+	CreateTime   types.Time   `json:"create_time"`
+	CreateTimeMs types.Time   `json:"create_time_ms"`
```
thrasher-, line 693:
```suggestion
	CreateTimeMS types.Time   `json:"create_time_ms"`
```

Or CreateTimeMilliseconds; I generally prefer the latter to be more human friendly with fields even though someone can infer what it means

thrasher- (changes requested):
One minor nit and then LGTG

exchanges/gateio/gateio_types.go
gbjk: Can't fix this one alone, without fixing all the others. :)
And if we're going to touch all of them, I think we should just ditch the non _ms field and rename the _ms field:
```
	CreateTime types.Time   `json:"create_time_ms"`
```
Unless you object real quick, that's coming in on runway 2 🛩️

thrasher-: If there's two fields doing the same thing and one with less precision than the other, makes sense. We should always default to whatever gives us the most precision (milliseconds field in this case) 👍 Assuming that both fields are always populated as well

gbjk: There's one edge case I'm chasing down now - The V3 implementation under the V4 api url for the options websocket, and the response header not having time_ms.
Anyway, yeah. I'm on it.

gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/7e730dda2 and  https://github.com/gbjk/gocryptotrader/commit/12122726d

gbjk: Benchmark referenced in commit message. Just realised that won't persist.
[Here it is for posterity](https://gist.github.com/gbjk/7cacb63b9a256e745534bb05ca853c48)

exchanges/gateio/gateio_websocket.go
```diff
@@ -216,6 +217,45 @@ func (g *Gateio) WsHandleSpotData(_ context.Context, respRaw []byte) error {
 	return nil
 }
 
+func parseWSHeader(msg []byte) (r *WSResponse, errs error) {
```
shazbert, line 220: This is definitely more performant than json I did my own benchmarks to validate

jsonparser:
```go
// 1646160	       666.3 ns/op	     312 B/op	       7 allocs/op
func BenchmarkParseHeader(b *testing.B) {
	in := `{"time":1726121320,"time_ms":1726121320745,"id":1,"channel":"spot.tickers","event":"subscribe","result":{"status":"success"},"request_id":"a4"}`
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		_, err := parseWSHeader([]byte(in))
		if err != nil {
			b.Fatal(err)
		}
	}
}
```

normal json:

```go
// 654258	      1810 ns/op	     408 B/op	      10 allocs/op
func BenchmarkJson(b *testing.B) {
	in := []byte(`{"time":1726121320,"time_ms":1726121320745,"id":1,"channel":"spot.tickers","event":"subscribe","result":{"status":"success"},"request_id":"a4"}`)
	b.ReportAllocs()
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		var resp WsResponse
		err := json.Unmarshal(in, &resp)
		if err != nil {
			b.Fatal(err)
		}
	}
}
```

Using the bench mark above I then switched branch #1623 to use sonic and that has comparative performance using `go test --tags=sonic -bench=BenchmarkJson -v` but has way less implementation complexity. I would suggest to reverse course just to reduce or limit that complexity and the ability for all websocket asset types pathways to achieve that relative boost without that toil.

```
BenchmarkJson-12         1775338               643.5 ns/op           361 B/op          7 allocs/op
BenchmarkJson-12         1857788               742.5 ns/op           362 B/op          7 allocs/op
BenchmarkJson-12         1890859               640.5 ns/op           361 B/op          7 allocs/op
BenchmarkJson-12         1871022               686.6 ns/op           362 B/op          7 allocs/op
```

shazbert (changes requested):
Changes look great, just 2 suggestions. Thanks heaps!

exchanges/gateio/gateio_types.go
```diff
@@ -2028,8 +2021,7 @@ type WsTicker struct {
 // WsTrade represents a websocket push data response for a trade
 type WsTrade struct {
```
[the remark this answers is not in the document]
gbjk: It felt a bit too 
image
to go on that rampage during a bugfix.
I have 2 more GateIO feature branches. I'll do it in one of those.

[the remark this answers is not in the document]
shazbert: 👍

exchanges/gateio/gateio_websocket.go
gbjk: Hard pass: I was already on the very very edge of acceptability when dealing with a bugfix.
This implementation is more performant than currently and not more complex for the callers of `parseWSHeader`.

I'm very excited to adopt sonic, but:
1) Not in a bug fix
2) I'm wary because of the hit we take on Silicon chips if we go sonic over jsonparser.
That's why I haven't yet gone gungho on it.
In the example here, we don't gain any performance by using sonic, but we lose performance on M* chips.
That's not a debate for here, admittedly.

thrasher- (approved):
Changes ACK, thanks for making the WSHeader improvements and benching it!

shazbert (approved):
tACK! Thanks

PR #1778, opened 2025-01-17
exchanges/gateio: Add websocket order batch [SPOT] support with context mocking

shazbert:
# PR Description

* Add wrapper interface function `WebsocketSubmitBatchOrders`
* Add gateio wrapper function `WebsocketSubmitBatchOrders` for spot only
* Adds context mock support for websocket and REST. This is just a basic POC implementation, in future we could take in an interface to our mock server but for now what you put in is what it is going to return. The websocket specifically will generate a mock connection when you get the connection for the request itself, so no need setup or connect, but will allow the full return functionality to be tested for an exchange wrapper. This can be run during live testing as well complimentary for both rest and WS and full coverage can be obtained. 

Futures not done just to keep this as small as possible and I haven't done any trading batch wise to verify it yet. 

Requires #1603 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/request/mock.go
```diff
@@ -0,0 +1,38 @@
...
+	"bytes"
+	"context"
+	"io"
+	"net/http"
+)
+
+var mockResponseFlag = struct{ name string }{name: "mockResponse"}
+
+// IsMockResponse returns true if the request has a mock response set
+func IsMockResponse(ctx context.Context) bool {
```
gloriousCode, line 13:
```suggestion
func HasMockResponse(ctx context.Context) bool {
```

exchanges/request/mock.go
```diff
@@ -0,0 +1,38 @@
...
+// GetMockResponse returns the mock response for a request
+func GetMockResponse(ctx context.Context) [][]byte {
+	mockResponse, _ := ctx.Value(mockResponseFlag).([][]byte)
+	return mockResponse
+}
+
+func getRESTResponseFromMock(ctx context.Context) *http.Response {
+	mockResp := GetMockResponse(ctx)
+	if len(mockResp) != 1 {
+		panic("mock REST response invalid, requires exactly one response")
```
gloriousCode, line 32: Let's not panic. This can be called from anywhere given `WithMockResponse` is exported. The function where this is used already has error handling after this function is called

exchanges/stream/websocket.go
```diff
@@ -1274,6 +1275,10 @@ func (w *Websocket) GetConnection(messageFilter any) (Connection, error) {
 		return nil, errMessageFilterNotSet
 	}
 
+	if request.IsMockResponse(ctx) {
+		return newMockWebsocketConnection(), nil
```
gloriousCode, line 1279: This placement does not appear the most appropriate. GetConnection creating a whole new one while existing functionality does not? 
Why not have a new function instead of this? Reduces the need for inspecting context on every call as well. Even if you don't make a new function, I think it should be removed from this one
I don't see any benefit to this over just assigning a websocket with `stream.MockWebsocketConnection` inside test code

exchanges/request/mock.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+	return &http.Response{
+		StatusCode: http.StatusOK,
+		Body:       io.NopCloser(io.Reader(io.LimitReader(bytes.NewBuffer(mockResp[0]), drainBodyLimit))),
```
gloriousCode, line 36:
```suggestion
		Body:       io.NopCloser(io.LimitReader(bytes.NewBuffer(mockResp[0]), drainBodyLimit)),
```

exchanges/request/mock_test.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"context"
+	"io"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestMockResponse(t *testing.T) {
```
gloriousCode, line 11: I will strongly encourage you write tests at a function level and not just a coverage level in a OneBigTest. Helps enforce rules and expectations especially when tests/functions are updated

exchanges/stream/mock_test.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/request"
+)
+
+func TestNewMockWebsocketConnection(t *testing.T) {
```
gloriousCode, line 11: Same here, I really encourage you to split out tests into their functions

exchanges/stream/mock_test.go
```diff
@@ -0,0 +1,25 @@
...
+	require.Panics(t, func() { got.SendMessageReturnResponse(context.Background(), 0, nil, nil) })
+	resp, err := got.SendMessageReturnResponsesWithInspector(context.Background(), 0, nil, nil, 0, nil)
+	require.NoError(t, err)
+	require.Nil(t, resp)
+	singleResp, err := got.SendMessageReturnResponse(request.WithMockResponse(context.Background(), []byte("test")), 0, nil, nil)
+	require.NoError(t, err)
+	require.NotNil(t, singleResp)
+	resp, err = got.SendMessageReturnResponsesWithInspector(request.WithMockResponse(context.Background(), []byte("test")), 0, nil, nil, 0, nil)
+	require.NoError(t, err)
+	require.NotNil(t, resp)
```
gloriousCode, line 24: I think the functionality is important enough to actually check that the result is what you expect. As well as including multi-byte-slice scenario too

gloriousCode (changes requested):
Ignoring gateio completely and just focussing on the mock functionality rn, which could probably have been its own pr, but that's okay we are here now

shazbert:
in favour of #1924

PR #1782, opened 2025-01-20
Binance: Add order side

Beadko:
# PR Description

Currently, the Binance side is marked as unknown. The raw data includes the isBuyerMaker: bool flag, which indicates the initiating side. Based on that, add a side.
Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [ ] golangci-lint run
- [x] TestGetRecentTrades
- [x] TestGetMostRecentTrades 
- [x] TestWsTradeUpdate 

## Checklist

### DIFF
--- a/exchanges/binance/binance_websocket.go
+++ b/exchanges/binance/binance_websocket.go
@@ -344,7 +344,6 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 			Amount:       t.Quantity.Float64(),
 			Exchange:     b.Name,
 			AssetType:    asset.Spot,
-			Side:         order.Sell,
 			TID:          strconv.FormatInt(t.TradeID, 10)}
 
 		if t.IsBuyerMaker { // Seller is Taker
--- a/exchanges/binance/binance_websocket.go
+++ b/exchanges/binance/binance_websocket.go
@@ -337,6 +337,11 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 				b.Name,
 				err)
 		}
+		// maker placed buy order, so taker market sell
+		side := order.Buy
+		if t.Maker {
+			side = order.Sell
+		}
 		return b.Websocket.Trade.Update(saveTradeData,
 			trade.Data{
 				CurrencyPair: pair,
@@ -345,6 +350,7 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 				Amount:       t.Quantity.Float64(),
 				Exchange:     b.Name,
 				AssetType:    asset.Spot,
+				Side:         side,
 				TID:          strconv.FormatInt(t.TradeID, 10),
 			})
 	case "ticker":
--- a/exchanges/binance/binance_websocket.go
+++ b/exchanges/binance/binance_websocket.go
@@ -337,22 +337,22 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 				b.Name,
 				err)
 		}
-		// maker placed buy order, so taker market sell
-		side := order.Buy
-		if t.Maker {
-			side = order.Sell
+		td := trade.Data{
+			CurrencyPair: pair,
+			Timestamp:    t.TimeStamp,
+			Price:        t.Price.Float64(),
+			Amount:       t.Quantity.Float64(),
+			Exchange:     b.Name,
+			AssetType:    asset.Spot,
+			Side:         order.Sell,
+			TID:          strconv.FormatInt(t.TradeID, 10)}
+
+		if t.IsBuyerMaker { // Seller is Taker
+			td.Side = order.Sell
+		} else { // Buyer is Taker
+			td.Side = order.Buy
 		}
-		return b.Websocket.Trade.Update(saveTradeData,
-			trade.Data{
-				CurrencyPair: pair,
-				Timestamp:    t.TimeStamp,
-				Price:        t.Price.Float64(),
-				Amount:       t.Quantity.Float64(),
-				Exchange:     b.Name,
-				AssetType:    asset.Spot,
-				Side:         side,
-				TID:          strconv.FormatInt(t.TradeID, 10),
-			})
+		return b.Websocket.Trade.Update(saveTradeData, td)
 	case "ticker":
 		var t TickerStream
 		err = json.Unmarshal(jsonData, &t)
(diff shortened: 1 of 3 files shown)


exchanges/binance/binance_websocket.go
```diff
@@ -337,6 +337,11 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
 				b.Name,
 				err)
 		}
+		// maker placed buy order, so taker market sell
+		side := order.Buy
+		if t.Maker {
+			side = order.Sell
+		}
```
shazbert, line 344: Being a maker or taker doesn't necessarily give you direction of order for this trade; because your maker can be on the bid liquidity (buy limit order) and the taker will hit the bids (sell market order) with their base amount whereas a maker can be on the ask liquidity (sell limit order) then your taker will lift the asks (buy market order) with their quote amount. I think a better approach might be to update trade.Data to have a field IsTaker or IsMaker.

shazbert (changes requested):
Thanks for opening this PR @Beadko and welcome back 🚀. Just one nit.

exchanges/binance/binance_websocket.go
gbjk: I had the same response at first.
But that's because we've misnamed the field.
[As per API docs](https://developers.binance.com/docs/binance-spot-api-docs/web-socket-api/market-data-requests#recent-trades) the field `m` is actually `isBuyerMaker`.

Since this is a trade, a market event, we know that if the buyer is the maker, then the seller is the taker, and this is a market SELL. And visa versa.

I suggest we make this clearer:
```suggestion
		if t.IsBuyerMaker { // Seller is Taker
		    t.Side = order.Sell
		} else { // Buyer is Taker
		    t.Side = order.Buy
		}
```
Note: I'm also implicitly suggesting moving declaration of td := trade.Data{} higher up so that we don't need to allocate a `side` var just to put it into another var.

Aside: This must be one of the only usages of `Trade.Update`, which I'm hoping to eradicate.

shazbert: `buyer` is the maker, then the seller is the `taker` and vice versa. That's sound. But direction can still be ambiguous, I do see where you are coming from but if you are a buyer (BTC USDT e.g.) and you are a maker you have put a limit order on the bid side whereas as a taker you are taking liquidity from the asks, but from the buyers perspective you are still going long, you are still picking up BTC. Or maybe I am just really confused which could be the case 😆

gbjk: > if you are a buyer (BTC USDT e.g.) and you are a maker you have put a limit order on the bid side whereas as a taker you are taking liquidity from the asks, but from the buyers perspective you are still going long, you are still picking up BTC.

I realised I shouldn't be categorical about this, because I'm not 100% sure.
When talking about the market trade itself, we're asking which "side" it should be visualised as.
The reason I think of it as Taker is direction is because if the orderbook gets filled in one direction, e.g. Maker's Limit Bid/Buy orders are getting filled, then we'd expect to see "Last trades" all in Red as SELLs, so that the Last trade fee matches the chart movement.
Example here, where the green Taker Buys on Binance match the green candles:
Screenshot 2025-01-21 at 10 28 15

shazbert: Yeah that last trade makes what you suggest correct for visualisation to the exchange. 👍

shazbert: @Beadko GK's suggestion please 🙏

Beadko: Good to be back too! :)

I've moved the trades declaration up and updated the TradeStream to match Binance api

exchanges/binance/binance_wrapper.go
```diff
@@ -864,6 +882,11 @@ func (b *Binance) GetHistoricTrades(ctx context.Context, p currency.Pair, a asse
 	}
 	result := make([]trade.Data, len(trades))
 	for i := range trades {
+		// maker placed buy order, so taker market sell
+		side := order.Buy
+		if trades[i].Maker {
```
shazbert, line 887: Change field to IsBuyerMaker

exchanges/binance/binance_websocket.go
```diff
@@ -337,16 +337,22 @@ func (b *Binance) wsHandleData(respRaw []byte) error {
...
+			Exchange:     b.Name,
+			AssetType:    asset.Spot,
+			Side:         order.Sell,
+			TID:          strconv.FormatInt(t.TradeID, 10)}
+
+		if t.IsBuyerMaker { // Seller is Taker
+			td.Side = order.Sell
+		} else { // Buyer is Taker
+			td.Side = order.Buy
+		}
```
shazbert, line 354: just for consistency we can have this style across all your other changes or have the other changes to this, but with the updated commentary.

shazbert (changes requested):
Thanks Bea, minor things left over then this looks good to go.

exchanges/binance/binance_wrapper.go
Beadko: amended

exchanges/binance/binance_websocket.go
Beadko: Added with the commentary as above, unless you wanted me to change it?

Beadko:
codespell seems to be erroring in the code that is not related to the PR

shazbert (approved):
tested ACK. Thanks Bea!

exchanges/binance/binance_wrapper.go
```diff
@@ -800,15 +806,21 @@ func (b *Binance) GetRecentTrades(ctx context.Context, p currency.Pair, a asset.
...
 				TID:          strconv.FormatInt(tradeData[i].ID, 10),
 				Exchange:     b.Name,
 				CurrencyPair: p,
 				AssetType:    a,
 				Price:        tradeData[i].Price,
 				Amount:       tradeData[i].Qty,
 				Timestamp:    tradeData[i].Time.Time(),
-			})
+			}
+			if tradeData[i].IsBuyerMaker { // Seller is Taker
```
gloriousCode, line 818: Makes me wish for c# `?` operator

gloriousCode (approved):
tACK!

thrasher- (approved):
Thanks @Beadko for this PR and @gbjk for the field clarification!

PR #1792, opened 2025-01-30
Subscriptions: Fix no enabled pairs for one asset

gbjk:
If one asset is enabled but has no enabled pairs, we should ignore that asset, even for non-pair related subscriptions.
That matches the existing code, but wasn't happening in the context of asset.All subscriptions with just one asset in this state. If a user wanted to have non-pair subscriptions, they should use asset.Empty. Would expect other breakages with no pairs enabled, too.

Note: No enabled pairs for an enabled asset is a transient issue which can occur due to enableOnePair racing against no available pairs. The second run, available pairs would be populated and enableOnePair would work. This situation could persist due to running with --dry or containerised, though.

Fixes:
`
Okx websocket: subscription failure, myOrders all : subscription template did not generate the expected number of pair records for spread: Got 1; Expected 0
`

Relates to #1420

## Type of change

### DIFF
--- a/exchanges/subscription/list.go
+++ b/exchanges/subscription/list.go
@@ -96,7 +96,7 @@ func (l List) SetStates(state State) error {
 
 func fillAssetPairs(ap assetPairs, a asset.Item, e IExchange) error {
 	p, err := e.GetEnabledPairs(a)
-	if err != nil {
+	if err != nil || len(p) == 0 {
 		return err
 	}
 	f, err := e.GetPairFormat(a, true)
--- a/exchanges/subscription/list_test.go
+++ b/exchanges/subscription/list_test.go
@@ -114,6 +114,17 @@ func TestAssetPairs(t *testing.T) {
 	}
 }
 
+// TestFillAssetPairs exercises AssetPairs handling with no enabled pairs
+// All other code is covered under TestAssetPairs and TestExpandTemplates
+func TestFillAssetPairs(t *testing.T) {
+	e := newMockEx()
+	e.pairs = currency.Pairs{}
+	ap := assetPairs{}
+	err := fillAssetPairs(ap, asset.Spot, e)
+	require.NoError(t, err, "fillAssetPairs must not error with no pairs enabled")
+	assert.Empty(t, ap, "fillAssetPairs should return an empty map with no pairs enabled")
+}
+
 func TestListClone(t *testing.T) {
 	t.Parallel()
 	l := List{{Channel: TickerChannel}, {Channel: OrderbookChannel}}


exchanges/subscription/list_test.go
```diff
@@ -114,6 +114,17 @@ func TestAssetPairs(t *testing.T) {
 	}
 }
 
+// TestFillAssetPairs exercises AssetPairs handling with no enabled pairs
+// All other code is covered under TestAssetPairs and TestExpandTemplates
+func TestFillAssetPairs(t *testing.T) {
+	e := newMockEx()
+	e.pairs = currency.Pairs{}
+	ap := assetPairs{}
+	err := fillAssetPairs(ap, asset.Spot, e)
```
shazbert, line 123: suggestion: any chance we can have fillAssetPairs a method to assetPairs, can resolve if its too much drama.

shazbert (approved):
Ack with one suggestion.

exchanges/subscription/list_test.go
```diff
@@ -114,6 +114,17 @@ func TestAssetPairs(t *testing.T) {
 	}
 }
 
+// TestFillAssetPairs exercises AssetPairs handling with no enabled pairs
+// All other code is covered under TestAssetPairs and TestExpandTemplates
+func TestFillAssetPairs(t *testing.T) {
```
gloriousCode, line 119: Don't really like this
- A test dedicated to testing a function that doesn't test its function
- Relying on other functions to not change to ensure that parts of this code is tested
  - Relying on commentary, that also won't go stale when any of the three function change
- Its not really that much code to just test the function logic explicitly here:
```go
func TestFillAssetPairs(t *testing.T) {
	t.Parallel()
	e := newMockEx()
	e.pairs = currency.Pairs{}
	ap := assetPairs{}
	err := fillAssetPairs(ap, asset.Spot, e)
	require.NoError(t, err, "fillAssetPairs must not error with no pairs enabled")
	assert.Empty(t, ap, "fillAssetPairs should return an empty map with no pairs enabled")

	pairs := currency.Pairs{ethusdcPair, btcusdtPair}
	ap = assetPairs{
		asset.Spot: pairs,
	}
	e.pairs = pairs
	err = fillAssetPairs(ap, asset.Spot, e)
	require.NoError(t, err, "fillAssetPairs must not error with pairs enabled")
	assert.Len(t, ap, 1, "fillAssetPairs should return one set of pairs")
	sortedPairs := ap[asset.Spot]
	assert.Equal(t, pairs[1], sortedPairs[0], "fillAssetPairs should return sorted pairs")

	errHi := errors.New("hi")
	e.errFormat = errHi
	err = fillAssetPairs(ap, asset.Spot, e)
	require.ErrorIs(t, err, errHi, "fillAssetPairs must error with GetPairFormat error")
}
```

gloriousCode (changes requested):

exchanges/subscription/list_test.go
gbjk: I had the thought niggle at the back of my head "Writing this comment means I've stuffed up".
Combined this with the move to being a method.
I wanted to ensure we prove ordering, and a few other things, so there's some changes to the fixture to add granularity for asset pairs.

Fixed https://github.com/gbjk/gocryptotrader/commit/1aba0ee4a

exchanges/subscription/list_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/1aba0ee4a

exchanges/subscription/fixtures_test.go
```diff
@@ -65,7 +62,14 @@ func (m *mockEx) GetSubscriptionTemplate(s *Subscription) (*template.Template, e
 				ap[asset.Spot] = ap[asset.Spot][0:1]
 				return ""
 			},
-			"batch": common.Batch[currency.Pairs],
+			"batch": func(pairs currency.Pairs, size int) []string {
+				s := []string{}
+				for _, p := range common.Batch(pairs, size) {
+					p = p.Format(currency.PairFormat{Uppercase: true})
```
shazbert, line 68: Question: Wouldn't this purge incoming formats? Like maybe they require delimiter in batches as an edge case?

exchanges/subscription/list.go
```diff
@@ -169,3 +156,17 @@ func (l List) Public() List {
 	}
 	return n
 }
+
+// populate adds all enabled pairs for an asset to the assetPair map
+func (ap assetPairs) populate(a asset.Item, e IExchange) error {
```
shazbert, line 161: suggestion: can swap the params around for alignment to calls.

shazbert (changes requested):
🔥 One question on recent change and a suggestion. Thanks.

exchanges/subscription/fixtures_test.go
shazbert: nevermind just noticed this as being in a test file, silly billy.

exchanges/subscription/list.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/4bf3d04bc

shazbert (changes requested):
tests need to be fixed then looks good to go. thanks!

gbjk:
Really sorry about that; Stupid mistake.

Fixed https://github.com/gbjk/gocryptotrader/commit/7880d163a

gloriousCode (approved):
tACK!

shazbert (approved):
tested approved

thrasher- (approved):

PR #1804, opened 2025-02-14
build/ci: Update Go to v1.24, golangci-lint to v1.64.5 and fix issues

thrasher-:
# PR Description

Bumps Go to v1.24, golangci-lint to v1.64.5 and fixes the issues it found.

Summary of changes:

- Linter complained for the benchmark `for i := 0; i < b.N; i++` loops and suggested to use a range as is standard now, however I opted instead to use b.Loop() which is new in Go1.24. 
- assertify any things which were touched.
- Resolved several integer overflow issues, with an attempt to unify all var usage across the codebase and reduce type casting where possible. 
- `kline.CalculateCandleDateRanges` had a potential divide by 0 panic bug if limit was set to 0.
- Config AES CFB has been switched to AES GCM as the prior one has been deprecated.
- Removed HitBTC v1 authenticated endpoints (API on version 3)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


shazbert (changes requested):
Good stuff! nothing major I could see.

exchanges/hitbtc/hitbtc.go
```diff
@@ -486,7 +486,16 @@ func (h *HitBTC) GetTradableBalances(ctx context.Context) (map[string]map[string
 	for x, y := range result.Data {
 		balances[x] = make(map[string]float64)
 		for z, w := range y {
-			balances[x][z], _ = strconv.ParseFloat(w.(string), 64)
+			bal, ok := w.(string)
```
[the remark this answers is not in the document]
thrasher-: I was going to do this but then it turns out all these v1 endpoints are dead, so I've removed them instead 😅

backtester/btrpc/btrpc.proto
```diff
@@ -151,7 +151,7 @@ message ExchangeCredentials {
 }
 
 message DataSettings {
-  uint64 interval = 1;
+  int64 interval = 1;
```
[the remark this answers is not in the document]
thrasher-: image

shazbert (approved):
tACK tested old config encryption to this from master 🚀

exchanges/kline/kline.go
```diff
@@ -471,43 +476,46 @@ func CalculateCandleDateRanges(start, end time.Time, interval Interval, limit ui
...
-		potentialRequests[x].Start = CreateIntervalTime(requestStart)
-
-		count -= int64(limit)
-		if count < 0 {
-			potentialRequests[x].Intervals = make([]IntervalData, count+int64(limit))
-		} else {
-			potentialRequests[x].Intervals = make([]IntervalData, limit)
+	remaining := count
+
+	for x := range ranges {
```
gbjk, line 495: Hi there! 👋  It looks like you're creating batches of things.
image

```
// CalculateCandleDateRanges will calculate the expected candle data in intervals in a date range
// If an API is limited in the amount of candles it can make in a request, it will automatically separate
// ranges into the limit
func CalculateCandleDateRanges(start, end time.Time, interval Interval, limit uint64) (*IntervalRangeHolder, error) {
	if err := common.StartEndTimeCheck(start, end); err != nil && !errors.Is(err, common.ErrStartAfterTimeNow) {
		return nil, err
	}
	if interval <= 0 {
		return nil, ErrInvalidInterval
	}

	start, end = start.Round(interval.Duration()), end.Round(interval.Duration())

	count := uint64(end.Sub(start) / interval.Duration()) //nolint:gosec // No overflow risk
	if count == 0 {
		return nil, common.ErrStartEqualsEnd
	}

	intervals := make([]IntervalData, 0, count)
	for iStart := start; iStart.Before(end); iStart = iStart.Add(interval.Duration()) {
		intervals = append(intervals, IntervalData{Start: CreateIntervalTime(iStart), End: CreateIntervalTime(iStart.Add(interval.Duration()))})
	}

	if limit == 0 {
		limit = count
	}

	h := &IntervalRangeHolder{
		Start:  CreateIntervalTime(start),
		End:    CreateIntervalTime(end),
		Ranges: []IntervalRange{},
		Limit:  limit,
	}

	for _, b := range common.Batch(intervals, int(limit)) {
		h.Ranges = append(h.Ranges, IntervalRange{
			Start:     b[0].Start,
			End:       b[len(b)-1].End,
			Intervals: b,
		})
	}

	return h, nil
}
```

gbjk (approved):
Looks good 👍 

Only nits, and a few pedantic suggestions.

gbjk (approved):
Looks good to me 👍

exchanges/kline/kline.go
```diff
@@ -482,41 +482,33 @@ func CalculateCandleDateRanges(start, end time.Time, interval Interval, limit ui
...
-		}
+	h := &IntervalRangeHolder{
+		Start: CreateIntervalTime(start),
+		End:   CreateIntervalTime(end),
+		Limit: limit,
+	}
 
-		ranges[x].End = CreateIntervalTime(requestStart)
-		remaining -= current
+	for _, b := range common.Batch(intervals, int(limit)) { //nolint:gosec // Ignore this warning as Batch requires int
```
shazbert, line 503:
```diff
diff --git a/exchanges/kline/kline.go b/exchanges/kline/kline.go
index 7b4450293..2c9a9963f 100644
--- a/exchanges/kline/kline.go
+++ b/exchanges/kline/kline.go
@@ -494,21 +494,17 @@ func CalculateCandleDateRanges(start, end time.Time, interval Interval, limit ui
                limit = count
        }

-       h := &IntervalRangeHolder{
-               Start: CreateIntervalTime(start),
-               End:   CreateIntervalTime(end),
-               Limit: limit,
-       }
-
+       ranges := make([]IntervalRange, 0, (len(intervals)/int(limit))+1)
        for _, b := range common.Batch(intervals, int(limit)) { //nolint:gosec // Ignore this warning as Batch requires int
-               h.Ranges = append(h.Ranges, IntervalRange{
-                       Start:     b[0].Start,
-                       End:       b[len(b)-1].End,
-                       Intervals: b,
-               })
+               ranges = append(ranges, IntervalRange{Start: b[0].Start, End: b[len(b)-1].End, Intervals: b})
        }

-       return h, nil
+       return &IntervalRangeHolder{
+               Start:  CreateIntervalTime(ranges[0].Start.Time),
+               End:    CreateIntervalTime(ranges[len(ranges)-1].End.Time),
+               Limit:  limit,
+               Ranges: ranges,
+       }, nil
 }

 // HasDataAtDate determines whether a there is any data at a set
```

Suggestion: Preallocate ranges and I have used the values in range for time start and end so that we can ensure through processing it hasn't deviated.

shazbert (changes requested):
One suggestion changes look good.

exchanges/kline/kline.go
gbjk: Occurred to me that maybe batch shouldn't require int.
I flip flopped on uint64 at the time.
Would that not align more with this overall change?

thrasher-: > Occurred to me that maybe batch shouldn't require int. I flip flopped on uint64 at the time. Would that not align more with this overall change?

Yep I was thinking about this too. Sadly this highlights one of my dislikes with Go's int usage across many builtin funcs (make, len etc). Which also means different values on 32bit vs 64 bit. Why make it an `int` when len can only ever be non-zero? The direction is heading for uint64 usage where it makes sense but in some cases we have to use int (for len etc). Alternatively, we can have our object slices have their own Len method which returns uint64 which I've adjusted in other parts on this PR. Our config version should also be an int64 (since it go negative) for the same reasons of different arch values.

gbjk: We agree:
```
LatestVersion = MaxUint32
```
i.e.
```
LatestVersion = 1<<32 - 1
```

exchanges/hitbtc/hitbtc.go
```diff
@@ -253,18 +249,6 @@ func (h *HitBTC) GenerateNewAddress(ctx context.Context, currency string) (Depos
 	return resp, err
 }
 
-// GetActiveorders returns all your active orders
```
gloriousCode, line 256: I have added that you've removed these to your PR since you didn't mention it 💥🗞️

exchanges/poloniex/poloniex.go
```diff
@@ -112,7 +112,11 @@ func (p *Poloniex) GetOrderbook(ctx context.Context, currencyPair string, depth
 			Asks: make([]OrderbookItem, len(resp.Asks)),
 		}
 		for x := range resp.Asks {
-			price, err := strconv.ParseFloat(resp.Asks[x][0].(string), 64)
+			askPrice, ok := resp.Asks[x][0].(string)
```
gloriousCode, line 115: If resp.Asks becomes `[][2]types.Number`
you can do
```go
	for x := range resp.Asks {
			ob.Asks[x] = OrderbookItem{
				Price:  resp.Asks[x][0].Float64(),
				Amount: resp.Asks[x][1].Float64(),
			}
		}
```
Just stood out

gloriousCode (changes requested):
That was a slog. Only thing that really stood out was `usetesting` being commented out

shazbert (approved):
tACK new changes. 🚀

gloriousCode (approved):
tACK!

config/config_encryption.go
```diff
@@ -130,12 +131,18 @@ func (c *Config) encryptConfigData(configData []byte) ([]byte, error) {
...
 	ciphertext := aead.Seal(nil, nil, configData, nil)
 
-	appendedFile := append(bytes.Clone(encryptionPrefix), c.storedSalt...)
-	appendedFile = append(appendedFile, encryptionVersionPrefix...)
-	versionOffset := len(encryptionPrefix) + len(c.storedSalt) + len(encryptionVersionPrefix)
-	appendedFile = append(appendedFile, make([]byte, 2)...)
-	binary.BigEndian.PutUint16(appendedFile[versionOffset:], encryptionVersion)
-	appendedFile = append(appendedFile, ciphertext...)
+	totalLen := len(encryptionPrefix) + len(c.storedSalt) + len(encryptionVersionPrefix) + versionSize + len(ciphertext)
+	appendedFile := make([]byte, totalLen)
```
gbjk, line 135: I'm easy, but arguably:
```suggestion
	appendedFile := make([]byte, len(encryptionPrefix) + len(c.storedSalt) + len(encryptionVersionPrefix) + versionSize + len(ciphertext))
```
For me, it's very implicit that the maths there amounts to "total length"

engine/engine_test.go
```diff
@@ -382,40 +380,18 @@ func TestGetDefaultConfigurations(t *testing.T) {
...
-				t.Error("name unset SetDefaults() not called")
-			}
-
-			if !defaultCfg.Enabled {
-				t.Error("expected enabled", defaultCfg.Name)
-			}
+			require.NoError(t, err, "GetDefaultConfig must not error")
+			require.NotNil(t, defaultCfg)
+			assert.NotEmpty(t, defaultCfg.Name)
+			assert.True(t, defaultCfg.Enabled)
```
gbjk, line 386: I think that when you're not testing the overall thing, you have to add context for the error.
That is: `require.NotEmpty(t, defaultCfg)` is fine, because it's about the unit test's target func main return value.
```suggestion
			assert.NotEmpty(t, defaultCfg.Name, "Name should not be empty")
			assert.True(t, defaultCfg.Enabled, "Enabled should have correct value")
```

gbjk (approved):
LGTM 🎉

backtester/eventhandlers/strategies/strategies.go
```diff
@@ -31,39 +31,29 @@ func LoadStrategyByName(name string, useSimultaneousProcessing bool) (Handler, e
...
+		return nil, fmt.Errorf("cannot load strategy %q: handler must be a non-nil pointer, got %T", name, h)
 	}
-	strategyInterface := newStrategy.Interface()
-	if strategyInterface == nil {
-		return nil, fmt.Errorf("cannot load %v new instance of strategy is not an interface. %w", name, common.ErrTypeAssertFailure)
-	}
-	strategy, ok := strategyInterface.(Handler)
+
+	// create new instance so strategy is not shared across all tasks
+	strategy, ok := reflect.New(strategyValue.Elem().Type()).Interface().(Handler)
```
gbjk, line 47: 🥃  _Aside:_ Feels like h should have been a func returning a Handler 🙂 
`NewSellEverything() Handler`. That's a bigger change and OOS.

backtester/eventhandlers/strategies/strategies_test.go
```diff
@@ -83,34 +84,33 @@ func TestCreateNewStrategy(t *testing.T) {
...
 	// invalid Handler
 	resp, err := createNewStrategy(dollarcostaverage.Name, false, nil)
-	if !errors.Is(err, common.ErrNilPointer) {
-		t.Errorf("received '%v' expected '%v'", err, common.ErrNilPointer)
-	}
-	if resp != nil {
-		t.Errorf("received '%v' expected '%v'", resp, nil)
-	}
+	assert.ErrorIs(t, err, common.ErrNilPointer)
+	assert.Nil(t, resp)
```
gbjk, line 88: 🦟 Don't think we should need to check what resp is when err is not nil 🙂 
```suggestion
```

backtester/eventhandlers/strategies/strategies_test.go
```diff
@@ -83,34 +84,33 @@ func TestCreateNewStrategy(t *testing.T) {
...
-	if !errors.Is(err, nil) {
-		t.Errorf("received '%v' expected '%v'", err, nil)
-	}
-	if resp == nil {
-		t.Errorf("received '%v' expected '%v'", resp, h)
-	}
+	h = new(dollarcostaverage.Strategy)
+	resp, err = createNewStrategy(dollarcostaverage.Name, true, h)
+	assert.NoError(t, err, "createNewStrategy should not error")
+	assert.NotNil(t, resp)
```
gbjk, line 104: 🦟 Nit:
```suggestion
assert.NotSame(t, h, resp, "createNewStrategy must return a new pointer")
```

gbjk (approved):
LGTM.

Only Nits.

backtester/eventhandlers/strategies/strategies.go
```diff
@@ -31,39 +31,29 @@ func LoadStrategyByName(name string, useSimultaneousProcessing bool) (Handler, e
...
-	}
-	strategyType := strategyElement.Type()
-	if strategyType == nil {
-		return nil, fmt.Errorf("cannot load %v strategy type is a %w", name, common.ErrNilPointer)
-	}
-	newStrategy := reflect.New(strategyType)
-	if newStrategy.IsNil() {
-		return nil, fmt.Errorf("cannot load %v new instance of strategy is a %w", name, common.ErrNilPointer)
+	if strategyValue.Kind() != reflect.Ptr || strategyValue.IsNil() {
+		return nil, fmt.Errorf("cannot load strategy %q: handler must be a non-nil pointer, got %T", name, h)
```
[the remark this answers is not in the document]
thrasher-: A non-ptr strat which implements Handler will panic on the call to strategyValue.Elem() so this is required

engine/engine_test.go
```diff
@@ -382,40 +380,18 @@ func TestGetDefaultConfigurations(t *testing.T) {
...
-				if defaultCfg.WebsocketResponseMaxLimit <= 0 {
-					t.Error("expected WebsocketResponseMaxLimit to be greater than 0", defaultCfg.Name)
-				}
-
-				if defaultCfg.WebsocketTrafficTimeout <= 0 {
-					t.Error("expected WebsocketTrafficTimeout to be greater than 0", defaultCfg.Name)
-				}
+				assert.Positive(t, defaultCfg.WebsocketResponseCheckTimeout, 0)
+				assert.Positive(t, defaultCfg.WebsocketResponseMaxLimit, 0)
+				assert.Positive(t, defaultCfg.WebsocketTrafficTimeout, 0)
```
shazbert, line 391:
```suggestion
				assert.Positive(t, defaultCfg.WebsocketResponseCheckTimeout)
				assert.Positive(t, defaultCfg.WebsocketResponseMaxLimit)
				assert.Positive(t, defaultCfg.WebsocketTrafficTimeout)
```

shazbert (changes requested):
Just something I noticed on new changes. Thanks!

backtester/eventhandlers/strategies/strategies.go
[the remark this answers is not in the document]
gbjk: Okay, understood.
But / so: he second past of the test is still redundant though, due to L33 nil checking, right?
```suggestion
	if strategyValue.Kind() != reflect.Ptr {
		return nil, fmt.Errorf("cannot load strategy %q: handler must be a non-nil pointer, got %T", name, h)
```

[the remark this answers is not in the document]
thrasher-: Nope we have to check for this too, since `var h Handler = (*customStrategy)(nil)` would pass the == nil check but fail on the IsNil() one. Without it, the below Type() call will panic 😆

gbjk (approved):
LGTM 🎉

gbjk (approved):

shazbert (approved):
Lovely, thanks!

PR #1809, opened 2025-02-22
Huobi: Fix false failure on NW date

gbjk:
Looks like "Within 2 weeks" fails on 22nd November during leap years. On 2025-02-21 we saw NW come up with 2025-03-07 for a few hours, and I'm guessing it's because they use local time. Added 1 day leeway to account for that timezone difference.

## Type of change

### DIFF
--- a/exchanges/huobi/huobi_test.go
+++ b/exchanges/huobi/huobi_test.go
@@ -1816,9 +1816,9 @@ func TestPairFromContractExpiryCode(t *testing.T) {
 		require.Falsef(t, d.Before(n), "%s expiry must be today or after", cType)
 		switch cType {
 		case "CW", "NW":
-			require.True(t, d.Before(n.Add(24*time.Hour*14)), "%s expiry must be within 2 weeks", cType)
+			require.Truef(t, d.Before(n.Add(24*time.Hour*15)), "%s expiry must be within 15 days; Got: `%s`", cType, d)
 		case "CQ", "NQ":
-			require.True(t, d.Before(n.Add(24*time.Hour*90*2)), "%s expiry must be within 2 quarters", cType)
+			require.Truef(t, d.Before(n.Add(24*time.Hour*181)), "%s expiry must be within 181 days; Got: `%s`", cType, d)
 		}
 	}
 }


shazbert (changes requested):
question/suggestion: I don't know times very well (so this is probably completely wrong) but shouldn't we load the expiration as cst then use add date because it accounts for leaps? We might need a mock for this as well. 

```diff
diff --git a/exchanges/huobi/huobi_test.go b/exchanges/huobi/huobi_test.go
index d3053550c..af7c2bf33 100644
--- a/exchanges/huobi/huobi_test.go
+++ b/exchanges/huobi/huobi_test.go
@@ -1811,14 +1811,16 @@ func TestPairFromContractExpiryCode(t *testing.T) {
                require.True(t, ok, "%s type must be in contractExpiryNames", cType)
                assert.Equal(t, currency.BTC, p.Base, "pair Base should be the same")
                assert.Equal(t, exp, p.Quote, "pair Quote should be the same")
-               d, err := time.Parse("060102", p.Quote.String())
+               cst, err := time.LoadLocation("Asia/Shanghai")
+               require.NoError(t, err)
+               d, err := time.ParseInLocation("060102", p.Quote.String(), cst)
                require.NoError(t, err, "currency code must be a parsable date")
                require.Falsef(t, d.Before(n), "%s expiry must be today or after", cType)
                switch cType {
                case "CW", "NW":
-                       require.Truef(t, d.Before(n.Add(24*time.Hour*15)), "%s expiry must be within 15 days; Got: `%s`", cType, d)
+                       require.Truef(t, d.Before(n.AddDate(0, 0, 14)), "%s expiry must be within 15 days; Got: `%s`", cType, d)
                case "CQ", "NQ":
-                       require.Truef(t, d.Before(n.Add(24*time.Hour*181)), "%s expiry must be within 181 days; Got: `%s`", cType, d)
+                       require.Truef(t, d.Before(n.AddDate(0, 3, 0)), "%s expiry must be within 181 days; Got: `%s`", cType, d)
                }
        }
 }
```

gbjk:
> question/suggestion: I don't know times very well (so this is probably completely wrong) but shouldn't we load the expiration as cst then use add date because it accounts for leaps? We might need a mock for this as well.

I mean ... *technically* they're based in the Seychelles 😂 
But actually I'm pretty sure their main base is Singapore now.

Yeah. I like your solution.
Hard pass to mocking this, though 😛 
If it fails again, we'll re-evaluate. RoI on mocking is super low.

Fixed https://github.com/gbjk/gocryptotrader/commit/ef5e85856b

exchanges/huobi/huobi_test.go
```diff
@@ -1811,14 +1811,16 @@ func TestPairFromContractExpiryCode(t *testing.T) {
 		require.True(t, ok, "%s type must be in contractExpiryNames", cType)
 		assert.Equal(t, currency.BTC, p.Base, "pair Base should be the same")
 		assert.Equal(t, exp, p.Quote, "pair Quote should be the same")
-		d, err := time.Parse("060102", p.Quote.String())
+		cst, err := time.LoadLocation("Asia/Singapore") // Huobi HQ and apparent local time for when codes become effective
+		require.NoError(t, err, "LoadLocation must not error")
+		d, err := time.ParseInLocation("060102", p.Quote.String(), cst)
```
shazbert, line 1816: Going from [this](https://www.timeanddate.com/time/zones/sgt)
```suggestion
		sgt, err := time.LoadLocation("Asia/Singapore") // Huobi HQ and apparent local time for when codes become effective
		require.NoError(t, err, "LoadLocation must not error")
		d, err := time.ParseInLocation("060102", p.Quote.String(), sgt)
```

shazbert (approved):
Changes ACK. one suggestion.

exchanges/huobi/huobi_test.go
gbjk: I think when we see a mistake happened because of unnecessary naming, best not to double down on it.

Fixed https://github.com/gbjk/gocryptotrader/commit/0f841230be

shazbert (approved):
Thanks for the changes!

exchanges/huobi/huobi_test.go
```diff
@@ -1811,14 +1811,16 @@ func TestPairFromContractExpiryCode(t *testing.T) {
...
+		d, err := time.ParseInLocation("060102", p.Quote.String(), tz)
 		require.NoError(t, err, "currency code must be a parsable date")
 		require.Falsef(t, d.Before(n), "%s expiry must be today or after", cType)
 		switch cType {
 		case "CW", "NW":
-			require.True(t, d.Before(n.Add(24*time.Hour*14)), "%s expiry must be within 2 weeks", cType)
+			require.Truef(t, d.Before(n.AddDate(0, 0, 14)), "%s expiry must be within 14 days; Got: `%s`", cType, d)
 		case "CQ", "NQ":
-			require.True(t, d.Before(n.Add(24*time.Hour*90*2)), "%s expiry must be within 2 quarters", cType)
+			require.Truef(t, d.Before(n.AddDate(0, 3, 14)), "%s expiry must be within 3 months; Got: `%s`", cType, d)
```
gloriousCode, line 1823: 180 days vs 104days. How come the change for NextQuarter? I know the contract rarely shows, but this looks like it will die when it does

gloriousCode (changes requested):
Thanks for the fix!

gloriousCode (approved):
tACK!

thrasher- (approved):
Thanks @gbjk and @shazbert for the improved solution!

PR #1825, opened 2025-03-03
Config: Tighten config version handling as uint16

gbjk:
This constrains the versions to uint16 and improves error handling. LatestVersion becomes literally that.
Fixes handling for negative or overflowing versions in config

## Type of change

### DIFF
--- a/cmd/config/config.go
+++ b/cmd/config/config.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"math"
 	"os"
 	"slices"
 	"strings"
@@ -23,7 +24,7 @@ func main() {
 
 	var in, out, keyStr string
 	var inplace bool
-	var version int
+	var version uint
 
 	fs := flag.NewFlagSet("config", flag.ExitOnError)
 	fs.Usage = func() { usage(fs) }
@@ -31,7 +32,7 @@ func main() {
 	fs.StringVar(&out, "out", "[in].out", "The config output file")
 	fs.BoolVar(&inplace, "edit", false, "Edit; Save result to the original file")
 	fs.StringVar(&keyStr, "key", "", "The key to use for AES encryption")
-	fs.IntVar(&version, "version", 0, "The version to downgrade to")
+	fs.UintVar(&version, "version", 0, "The version to downgrade to")
 
 	cmd, args := parseCommand(os.Args[1:])
 	if cmd == "" {
@@ -86,12 +87,12 @@ func main() {
 				os.Exit(3)
 			}
 			version = versions.LatestVersion
-		} else if version < 0 {
-			fmt.Fprintln(os.Stderr, "Error: version must be positive")
+		} else if version >= math.MaxUint16 {
+			fmt.Fprintln(os.Stderr, "Error: version must be less than 65535")
 			usage(fs)
 			os.Exit(3)
 		}
-		if data, err = versions.Manager.Deploy(context.Background(), data, version); err != nil {
+		if data, err = versions.Manager.Deploy(context.Background(), data, uint16(version)); err != nil {
 			fatal("Unable to " + cmd + " config; Error: " + err.Error())
 		}
 		if !isEncrypted {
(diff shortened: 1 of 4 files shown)


config/versions/versions.go
```diff
@@ -27,16 +28,19 @@ import (
 )
 
 // LatestVersion used as version param to Deploy to automatically use the latest version
-const LatestVersion = -1
+const LatestVersion = math.MaxUint16
```
gloriousCode, line 31: Given is not written to with the latest version, is not used to store the latest version, is a const, I think it would be more appropriately named as `MaxVersion`

cmd/config/config.go
```diff
@@ -23,15 +23,15 @@ func main() {
...
+	var version uint
 
 	fs := flag.NewFlagSet("config", flag.ExitOnError)
 	fs.Usage = func() { usage(fs) }
 	fs.StringVar(&in, "in", defaultCfgFile, "The config input file to process")
 	fs.StringVar(&out, "out", "[in].out", "The config output file")
 	fs.BoolVar(&inplace, "edit", false, "Edit; Save result to the original file")
 	fs.StringVar(&keyStr, "key", "", "The key to use for AES encryption")
-	fs.IntVar(&version, "version", 0, "The version to downgrade to")
+	fs.UintVar(&version, "version", 0, "The version to downgrade to")
```
gloriousCode, line 34: A truly petty nit, this version is limited to `16,383` versions and `LatestVersion` is limited to `65,535`  versions. Claustrophobic!

I suppose we have 16,380 versions' time to update this to uin64Var is needed 😄

gloriousCode (approved):
Would prefer the `LatestVersion` rename, but I'm not blocking coverage based on a pre-existing condition 🍝 
tACK

cmd/config/config.go
thrasher-: I was thinking WTF is arm/Apple silicon really this cooked? But looks like you got that value because the version you inputted was > MaxUint16 and we just type cast it without checking. Without that, should fit 65355 fine without any issues.

cmd/config/config.go
```diff
@@ -91,7 +91,7 @@ func main() {
 			usage(fs)
 			os.Exit(3)
 		}
-		if data, err = versions.Manager.Deploy(context.Background(), data, version); err != nil {
+		if data, err = versions.Manager.Deploy(context.Background(), data, uint16(version)); err != nil {
```
thrasher-, line 94: Not in this diff, but you can rid the `else if version < 0` now since it's not possible

thrasher- (changes requested):
Thanks for hopping onto this so quickly

config/versions/versions.go
gbjk: `MaxVersion` for me would be the highest possible version.
Which is `math.MaxUint16 -1`
Since `LatestVersion` is used to say "automatically use the latest version", `UseLatestVersion` might be the right compromise.

Fixed https://github.com/gbjk/gocryptotrader/commit/c4d846e059

cmd/config/config.go
gbjk: @thrasher- is correct: The issue is that that we're not checking for overflow on uint16.
That's actually relatively bad because a fat finger error could cause a downgrade to an actual existing version.

So: We have 65534 available versions.

Fixed https://github.com/gbjk/gocryptotrader/commit/e23a2f1173

Slight niggle in that the version check for overflow happens inside the cmd wrapper so has a different format.
If someone puts in 65536 and then corrects to 65530 they'll get a second error about what the highest version  available *actually* is.
However I'm pretty comfortable with limited convenience in don't-shoot-your-foot-off circumstances.

cmd/config/config.go
gbjk: Yep, thanks. Actually makes the earlier thing not so crazy.

Fixed https://github.com/gbjk/gocryptotrader/commit/cf86eeb468

shazbert (changes requested):
tACK, one linter issue then looks good to go.

cmd/config/config.go
gloriousCode: Can't wait for v65420 now that this is resolved

shazbert (approved):
Thanks sir @gbjk !

thrasher- (approved):

PR #1826, opened 2025-03-05
modernise: Run new gopls modernise tool against the codebase and fix minor issues

thrasher-:
# PR Description

Runs the modernise tool against GCT correcting any issues and fixing additional issues along the way. You can find info about the tool here: https://pkg.go.dev/golang.org/x/tools/gopls/internal/analysis/modernize

This is now enabled by default by gopls in VS Code but this PR does not enforce it as this will likely be done as a linter upgrade for golangci-lint.

Depends on https://github.com/thrasher-corp/gocryptotrader/pull/1804

## Type of change

- [x] Gopls modernise fixes

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


shazbert (changes requested):
Good stuff!

Some references left over in docs
image

sonic 1.13.1 seems to be giving me a build error 😆 

```
go version go1.24.0 windows/amd64
go build
# github.com/thrasher-corp/gocryptotrader
link: github.com/bytedance/sonic/loader: invalid reference to runtime.lastmoduledatap
```

shazbert (approved):
Looks good to me tACK! so modern!

cmd/gctcli/orderbook.go
```diff
@@ -453,20 +453,14 @@ func getOrderbook(c *cli.Context) error {
...
-			maxLen = bidLen
-		} else {
-			maxLen = askLen
-		}
+		bidLen := int64(len(result.Bids) - 1)
+		askLen := int64(len(result.Asks) - 1)
+		maxLen := max(bidLen, askLen)
 		if depthLimit > 0 && depthLimit < maxLen {
 			maxLen = depthLimit
 		}
```
gbjk, line 461:
```suggestion
		if depthLimit > 0  {
			maxLen = min(maxLen, depthLimit)
		}
```

cmd/gctcli/orderbook.go
```diff
@@ -453,20 +453,14 @@ func getOrderbook(c *cli.Context) error {
...
-		if maxLen > 100 {
-			maxLen = 100
-		}
+
+		maxLen = min(maxLen, 100)
```
gbjk, line 463: Is this a depthLimit default in disguise?
I think:
```
diff --git a/cmd/gctcli/orderbook.go b/cmd/gctcli/orderbook.go
index 07dd180a9c..9253a237a5 100644
--- a/cmd/gctcli/orderbook.go
+++ b/cmd/gctcli/orderbook.go
@@ -374,7 +374,6 @@ func getOrderbook(c *cli.Context) error {
 
 	var (
 		exchangeName, pair, assetType string
-		depthLimit                    int64
 		exchangeStyle                 bool
 		err                           error
 	)
@@ -410,12 +409,15 @@ func getOrderbook(c *cli.Context) error {
 		}
 	}
 
-	if c.IsSet("depthlimit") {
-		depthLimit = c.Int64("depthlimit")
-	} else if c.Args().Get(4) != "" {
-		depthLimit, err = strconv.ParseInt(c.Args().Get(4), 10, 64)
-		if err != nil {
+	const depthCeiling = 100 // The maximum the depth can be regardless of user entry
+	depthLimit := uint64(depthCeiling)
+	if d := c.Uint64("depthlimit"); d > 0 && d < depthCeiling {
+		depthLimit = d
+	} else if d := c.Args().Get(4); d != "" {
+		if dU, err := strconv.ParseUint(d, 10, 64); err != nil {
 			return err
+		} else if dU > 0 && dU < depthCeiling {
+			depthLimit = dU
 		}
 	}
 
@@ -455,12 +457,7 @@ func getOrderbook(c *cli.Context) error {
 	if exchangeStyle {
 		bidLen := int64(len(result.Bids) - 1)
 		askLen := int64(len(result.Asks) - 1)
-		maxLen := max(bidLen, askLen)
-		if depthLimit > 0 && depthLimit < maxLen {
-			maxLen = depthLimit
-		}
-
-		maxLen = min(maxLen, 100)
+		maxLen := min(max(bidLen, askLen), depthLimit)
 		renderOrderbookExchangeStyle(result, exchangeName, assetType, maxLen, askLen, bidLen)
 	} else {
 		jsonOutput(result)
```

cmd/gctcli/orderbook.go
```diff
@@ -617,13 +611,7 @@ func getOrderbookStream(c *cli.Context) error {
...
 		bidLen := int64(len(resp.Bids) - 1)
 		askLen := int64(len(resp.Asks) - 1)
-
-		var maxLen int64
-		if bidLen >= askLen {
-			maxLen = bidLen
-		} else {
-			maxLen = askLen
-		}
+		maxLen := max(bidLen, askLen)
```
gbjk, line 614: If above applies, then here too.
Can provide patch if desired.

gbjk (commented):
Partial feedback.

Only Nits 👍

cmd/gctcli/orderbook.go
thrasher-: I agree and thanks for the offer but I can do 👍

thrasher-: Pushed: https://github.com/thrasher-corp/gocryptotrader/pull/1826/commits/c63c2576c7173b3e7fda0d7cb1bd10ea85b06713

backtester/funding/trackingcurrencies/trackingcurrencies.go
```diff
@@ -101,12 +102,9 @@ func CreateUSDTrackingPairs(tp []TrackingPair, em *engine.ExchangeManager) ([]Tr
...
 // tracks against USD value, ie is in rankedUSDs
 func CurrencyIsUSDTracked(code currency.Code) bool {
-	for i := range rankedUSDs {
-		if code.Equal(rankedUSDs[i]) {
-			return true
-		}
-	}
-	return false
+	return slices.ContainsFunc(rankedUSDs, func(c currency.Code) bool {
+		return c.Equal(code)
```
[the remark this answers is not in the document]
thrasher-: Unfortunately not, since slices.Contains wouldn't pickup case insensitive codes if a person were to `c := currency.NewCode("usd")` instead of `currency.USD`. Wheras using `Equal` will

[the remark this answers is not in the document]
gbjk: 🥃 Think that's something we might want to address, such that codes are case insensitive *after* the first one is registered.
🗻 Out of scope, resolving.

cmd/apichecker/apicheck.go
```diff
@@ -738,22 +738,13 @@ func htmlScrapeHitBTC(htmlData *HTMLScrapingData) ([]string, error) {
 	if err != nil {
 		return nil, err
 	}
-	aBody := string(a)
+
 	r, err := regexp.Compile(htmlData.RegExp)
```
gbjk, line 742: 🥃 This isn't coming from outside, so we could `MustCompile` safely.
Could also make htmlData.RegExp into a regexp.Regexp

engine/sync_manager.go
```diff
@@ -451,10 +451,7 @@ func (m *SyncManager) worker() {
 	}
 	defer cleanup()
 
-	interval := greatestCommonDivisor(m.config.TimeoutWebsocket, m.config.TimeoutREST)
-	if interval > time.Second {
-		interval = time.Second
-	}
+	interval := min(greatestCommonDivisor(m.config.TimeoutWebsocket, m.config.TimeoutREST), time.Second)
```
gbjk, line 454: 🐒 Nitpick: time.Second should be a const
```suggestion
	interval := min(greatestCommonDivisor(m.config.TimeoutWebsocket, m.config.TimeoutREST), minSyncInterval)
```

exchanges/asset/asset.go
```diff
@@ -146,14 +147,7 @@ func (a Items) Strings() []string {
 // Contains returns whether or not the supplied asset exists
 // in the list of Items
 func (a Items) Contains(i Item) bool {
```
gbjk, line 149: 🥃 Kinda getting shaky if this method should exist

exchanges/binance/binance.go
```diff
@@ -911,10 +911,8 @@ func (b *Binance) SendAuthHTTPRequest(ctx context.Context, ePath exchange.URL, m
...
 // CheckLimit checks value against a variable list
 func (b *Binance) CheckLimit(limit int) error {
-	for x := range b.validLimits {
-		if b.validLimits[x] == limit {
-			return nil
-		}
+	if slices.Contains(b.validLimits, limit) {
+		return nil
 	}
 	return errors.New("incorrect limit values - valid values are 5, 10, 20, 50, 100, 500, 1000")
```
gbjk, line 917: 🐒 Nitpick:
* If we're going to quote anything, it should be what the supplied value was
* I'm against prompting what valid values are in errors, but if we must, we should `strings.Join()` on `b.validLimits`
```suggestion
	return fmt.Errorf("%w `%v`, errInvalidLimit, limit)
```

exchanges/binanceus/binanceus.go
```diff
@@ -395,10 +396,8 @@ func (bi *Binanceus) GetOrderBookDepth(ctx context.Context, arg *OrderBookDataRe
...
 // CheckLimit checks value against a variable list
 func (bi *Binanceus) CheckLimit(limit int64) error {
-	for x := range bi.validLimits {
-		if bi.validLimits[x] == limit {
-			return nil
-		}
+	if slices.Contains(bi.validLimits, limit) {
+		return nil
 	}
 	return errIncorrectLimitValues
```
gbjk, line 402: 🐒 Nitpick: Spotting inconsistency c.f. `binance.go` CheckLimits return error.

gbjk (commented):
Next 100 files reviewed; Only minor nits raised.

exchanges/binance/binance.go
thrasher-: It turns out that Binance allows any limit value between 1-5000 so I've removed it entirely for it and BinanceUS https://github.com/thrasher-corp/gocryptotrader/pull/1826/commits/bbbe32bcb856851d2819794b2d652c1b6278432c

exchanges/btcmarkets/btcmarkets_websocket.go
```diff
@@ -468,10 +468,7 @@ func checksum(ob *orderbook.Base, checksum uint32) error {
 
 // concat concatenates price and amounts together for checksum processing
 func concat(liquidity orderbook.Tranches) string {
-	length := 10
-	if len(liquidity) < 10 {
-		length = len(liquidity)
-	}
+	length := min(10, len(liquidity))
 	var c string
 	for x := range length {
```
gbjk, line 473: 🐒 
* Since we're not re-assigning to it anymore, don't think we necessarily need a var:
```suggestion
	var c string
	for x := range 	min(10, len(liquidity)) {
```
* c can move to sig if you want:
```suggestion
func concat(liquidity orderbook.Tranches) (c string) {
	for i := range min(10, len(liquidity)) {
		c += trim(liquidity[i].Price) + trim(liquidity[i].Amount)
	}
	return c
```
* 🥃 And since we've come this far, `concat` is a bit of a loose name given it's in the whole of the exchange package namespace. `concatOrderbookLiquidity` or something

exchanges/coinut/coinut.go
```diff
@@ -197,16 +197,16 @@ func (c *COINUT) GetTradeHistory(ctx context.Context, instrumentID, start, limit
...
 
 	return result, c.SendHTTPRequest(ctx, exchange.RestSpot, coinutIndexTicker, params, false, &result)
 }
 
 // GetDerivativeInstruments returns a list of derivative instruments
-func (c *COINUT) GetDerivativeInstruments(ctx context.Context, secType string) (interface{}, error) {
-	var result interface{} // to-do
-	params := make(map[string]interface{})
+func (c *COINUT) GetDerivativeInstruments(ctx context.Context, secType string) (any, error) {
+	var result any // to-do
```
gbjk, line 208: 🐒 Unusual spelling of TODO won't get highlighted (at least for me)
```suggestion
	var result any // TODO: Provide a concrete type for the result
```

exchanges/coinut/coinut.go
```diff
@@ -380,50 +380,27 @@ func (c *COINUT) calculateTradingFee(base, quote currency.Code, purchasePrice, a
...
-			fee = amount * 0.005
-		}
+		return max(amount*0.005, 2.0)
 	case currency.SGD:
-		if amount*0.001 < 10 {
-			fee = 10
-		} else {
-			fee = amount * 0.001
-		}
+		return max(amount*0.001, 10.0)
```
gbjk, line 389: 🚧 I believe this fee is [incorrect](https://coinut.zendesk.com/hc/en-us/articles/19088967379609-4c-XSGD-Withdrawals-via-Bank)
```suggestion
		return 2.0
```

exchanges/order/orders.go
```diff
@@ -971,12 +972,11 @@ func FilterOrdersByPairs(orders *[]Detail, pairs []currency.Pair) {
 			continue
 		}
```
gbjk, line 974: 🥃 Note for me to sniff this func's usage.
* Could/Should be a func on a List type
* And that's if we accept it should be mutable. Idiomatically only Sort may be mutable (given slices.Sort). Otherwise we should be functional programming and returning a new slice
I'll resolve and solo this at some point. Or raise an issue

exchanges/orderbook/tranches.go
```diff
@@ -88,7 +89,7 @@ updates:
 				copy((*ts)[y:], (*ts)[y+1:])
 				*ts = (*ts)[:len(*ts)-1]
 			} else {
-				*ts = append((*ts)[:y], (*ts)[y+1:]...)
+				*ts = slices.Delete(*ts, y, y+1)
 			}
```
gbjk, line 93: 🚧  These are the same thing in disguise.
Can't make a suggestion for line 87, but this should just be:
```suggestion
*ts = slices.Clip(slices.Delete(*ts, y, y+1))
```

portfolio/portfolio.go
```diff
@@ -178,7 +179,7 @@ func (b *Base) UpdateAddressBalance(address string, amount float64) {
 func (b *Base) RemoveExchangeAddress(exchangeName string, coinType currency.Code) {
 	for x := range b.Addresses {
```
gbjk, line 180: 🚧 Hidden DeleteFunc here.
```suggestion
	b.Addresses = slices.DeleteFunc(b.Addresses, func(a Address) bool {
		return a.Address == exchangeName && a.CoinType.Equal(coinType)
	})
```

portfolio/portfolio.go
```diff
@@ -236,16 +237,14 @@ func (b *Base) RemoveAddress(address, description string, coinType currency.Code
...
-		if b.Addresses[x].Address == address &&
-			b.Addresses[x].CoinType.Equal(coinType) &&
-			b.Addresses[x].Description == description {
-			b.Addresses = append(b.Addresses[:x], b.Addresses[x+1:]...)
-			return nil
-		}
+	idx := slices.IndexFunc(b.Addresses, func(a Address) bool {
+		return a.Address == address && a.CoinType.Equal(coinType) && a.Description == description
+	})
+	if idx == -1 {
```
gbjk, line 243: 🐒 Think this pattern is probably cleaner and more efficient, and only mentioning it so we don't cargo-cult whichever we don't choose.
```suggestion
	oldSize := len(b.Addresses)
	b.Addresses = slices.Clip(slices.DeleteFunc(b.Addresses, func(a Address) bool {
		return a.Address == address && a.CoinType.Equal(coinType) && a.Description == description
	}))
	if len(b.Addresses) == oldSize {
```

portfolio/portfolio.go
```diff
@@ -236,16 +237,14 @@ func (b *Base) RemoveAddress(address, description string, coinType currency.Code
...
+		return errors.New("portfolio item does not exist")
 	}
-
-	return errors.New("portfolio item does not exist")
+	b.Addresses = slices.Delete(b.Addresses, idx, idx+1)
```
gbjk, line 246: 🐒 Otherwise:
```suggestion
	b.Addresses = slices.Clip(slices.Delete(b.Addresses, idx, idx+1))
```

exchanges/orderbook/tranches.go
gbjk: Though, I think actually this whole section might be better as a single `*ts = slices.DeleteFunc`

gbjk (changes requested):
Review finished.

Arguably anything I've raised is OOS because they're all existing issues, or optimisations.

I'll put it as a Request Change though, and happy to switch to Approval if we just wanna get this across the line and raise everything else as issues.

gbjk (approved):
Thanks for those changes 👍

exchanges/orderbook/tranches.go
```diff
@@ -78,27 +78,23 @@ updates:
...
-			return fmt.Errorf("delete error: %w %d not found",
-				errIDCannotBeMatched,
-				updts[x].ID)
+			continue
 		}
+
+		*ts = slices.Delete(*ts, idx, idx+1)
 	}
+
+	*ts = slices.Clip(*ts)
```
shazbert, line 96: We don't need to clip as the underlying array is fine to stay big, every time a new update comes in it will resize the underlying and hurt performance.

exchanges/orderbook/tranches.go
```diff
@@ -78,27 +78,23 @@ updates:
 
 // deleteByID deletes reference by ID
 func (ts *Tranches) deleteByID(updts Tranches, bypassErr bool) error {
-updates:
 	for x := range updts {
-		for y := range *ts {
-			if updts[x].ID != (*ts)[y].ID {
-				continue
-			}
+		idx := slices.IndexFunc(*ts, func(t Tranche) bool {
```
shazbert, line 82: The changes here are a slight decrease in performance

```
master
 6487341	       197.3 ns/op	     480 B/op	       1 allocs/op

this branch:
 4987383	       236.9 ns/op	     480 B/op	       1 allocs/op
```

shazbert (changes requested):
One minor thing on new changes.

gloriousCode (commented):
Very modern and fancy. Pretty much all my comments are the same thing, but I'm seeking clarity rather than changes
image

backtester/eventhandlers/statistics/statistics_types.go
```diff
@@ -173,7 +173,7 @@ type CurrencyPairStatistic struct {
 
 	Events []DataAtOffset `json:"-"`
 
-	MaxDrawdown           Swing               `json:"max-drawdown,omitempty"`
+	MaxDrawdown           Swing               `json:"max-drawdown"`
```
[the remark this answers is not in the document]
thrasher-: Yep that's correct, currently this doesn't work as what was intended. https://go.dev/play/p/T5raI0q16iU

exchanges/binance/binance_test.go
```diff
@@ -1969,8 +1969,7 @@ func BenchmarkWsHandleData(bb *testing.B) {
 			<-b.Websocket.DataHandler
 		}
 	}()
-	bb.ResetTimer()
-	for range bb.N {
+	for bb.Loop() {
```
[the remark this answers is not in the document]
thrasher-: Taken from: https://pkg.go.dev/testing@master#B.Loop:

`Loop resets the benchmark timer the first time it is called in a benchmark, so any setup performed prior to starting the benchmark loop does not count toward the benchmark measurement. Likewise, when it returns false, it stops the timer so cleanup code is not measured.`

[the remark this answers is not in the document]
gloriousCode: I've been wrong all this time 😭

gloriousCode (approved):
I am happy, but I will relook post master merge/rebase

shazbert (approved):
Changes tACK. 🙏

gloriousCode (approved):
### 🎩 
tACK

exchanges/orderbook/tranches.go
```diff
@@ -78,23 +78,25 @@ updates:
...
 
-		if idx == -1 {
-			if !bypassErr {
-				return fmt.Errorf("delete error: %w %d not found", errIDCannotBeMatched, updts[x].ID)
+			if y < len(*ts) {
+				copy((*ts)[y:], (*ts)[y+1:])
+				*ts = (*ts)[:len(*ts)-1]
+			} else {
+				*ts = slices.Delete(*ts, y, y+1)
 			}
```
gbjk: 🐒 Even if this is slightly faster, I think it's not worth doing instead of unifying.
```suggestion
			*ts = slices.Delete(*ts, y, y+1)
```

gbjk: 🚧 I do not understand this code. I feel that I must be being really stupid, except ... I don't think I am.

* y is an index range of *ts
* So `y < len(*ts)` is always true? The last value for y will be `len(*ts) - 1`

```suggestion
			copy((*ts)[y:], (*ts)[y+1:])
			*ts = (*ts)[:len(*ts)-1]
```

gbjk (changes requested):

exchanges/orderbook/tranches.go
gbjk: I'm wary about this rationale; Doesn't it pin us to high-watermark memory usage?
If there's an absurd update, and the orderbook shrinks by 75% ... would we reclaim that?

exchanges/orderbook/tranches.go
gbjk: I don't agree with the source of the delay being `IndexFunc`. I'm confident it's the Delete.

```
BenchmarkDeleteByID_Orig-8                       8801080               133.4 ns/op           480 B/op          1 allocs/op
BenchmarkDeleteByID_Delete-8                     7127386               167.2 ns/op           480 B/op          1 allocs/op
BenchmarkDeleteByID_IndexFunc-8                  8379580               143.3 ns/op           480 B/op          1 allocs/op
BenchmarkDeleteByID_IndexFuncDelete-8            6670990               179.2 ns/op           480 B/op          1 allocs/op
```

gbjk (changes requested):

exchanges/orderbook/tranches.go
thrasher-: Yeah in all my testing DeleteFunc was slower than an IndexFunc like I used in the other funcs which was quite hilarious. Whereas the origin wins in all circumstances (Ryan the 🏎️ driver)

exchanges/orderbook/tranches.go
thrasher-: On very liquid pairs we'd likely not see an orderbook shrink by 75%, esp since most exchanges will have a limit of N for ob depth on websocket which continually refresh at that limit but could potentially happen on illiquid pairs and they generally update way less, so in that case perf won't be as critical.

Perhaps a conditional thing? 

```
if cap(*ts) > 2*len(*ts) {
	*ts = slices.Clip(*ts)
}
```

But this adds complexity and we'd need hard data to prove either case

gbjk (approved):
Thanks 👍

Issue #1831, opened 2025-03-08
bug: UpdateAccountInfo does not update balances to zero due to account.Process behavior


junnplus:
`UpdateAccountInfo` retrieves the full list of balances for an asset type. However, when updating the balances, it uses `account.Process`, which only processes the currencies present in the response. If an exchange does not return currencies with a zero balance, `account.Process` does not update them, leaving outdated balance data instead of setting them to zero.

PR #1846, opened 2025-03-13
account: refactor for save holdings

junnplus:
# PR Description

Fixes https://github.com/thrasher-corp/gocryptotrader/issues/1831
## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/rpcserver.go
+++ b/engine/rpcserver.go
@@ -653,36 +653,6 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
 		return err
 	}
 
-	initAcc, err := exch.GetCachedAccountInfo(stream.Context(), assetType)
-	if err != nil {
-		return err
-	}
-
-	accounts := make([]*gctrpc.Account, len(initAcc.Accounts))
-	for x := range initAcc.Accounts {
-		subAccounts := make([]*gctrpc.AccountCurrencyInfo, len(initAcc.Accounts[x].Currencies))
-		for y := range initAcc.Accounts[x].Currencies {
-			subAccounts[y] = &gctrpc.AccountCurrencyInfo{
-				Currency:   initAcc.Accounts[x].Currencies[y].Currency.String(),
-				TotalValue: initAcc.Accounts[x].Currencies[y].Total,
-				Hold:       initAcc.Accounts[x].Currencies[y].Hold,
-				UpdatedAt:  timestamppb.New(initAcc.Accounts[x].Currencies[y].UpdatedAt),
-			}
-		}
-		accounts[x] = &gctrpc.Account{
-			Id:         initAcc.Accounts[x].ID,
-			Currencies: subAccounts,
-		}
-	}
-
-	err = stream.Send(&gctrpc.GetAccountInfoResponse{
-		Exchange: initAcc.Exchange,
-		Accounts: accounts,
-	})
-	if err != nil {
-		return err
-	}
-
 	pipe, err := account.SubscribeToExchangeAccount(r.Exchange)
 	if err != nil {
 		return err
@@ -694,16 +664,23 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
 			log.Errorln(log.DispatchMgr, pipeErr)
 		}
 	}()
+	init := make(chan struct{}, 1)
+	init <- struct{}{}
 
 	for {
-		data, ok := <-pipe.Channel()
-		if !ok {
-			return errDispatchSystem
+		select {
+		case <-stream.Context().Done():
+			return stream.Context().Err()
+		case _, ok := <-pipe.Channel():
+			if !ok {
+				return errDispatchSystem
+			}
+		case <-init:
 		}
 
-		holdings, ok := data.(*account.Holdings)
-		if !ok {
-			return common.GetTypeAssertError("*account.Holdings", data)
+		holdings, err := exch.GetCachedAccountInfo(stream.Context(), assetType)
+		if err != nil {
+			return err
 		}
 
 		accounts := make([]*gctrpc.Account, len(holdings.Accounts))
--- a/exchanges/account/account.go
+++ b/exchanges/account/account.go
@@ -171,8 +171,7 @@ func GetBalance(exch, subAccount string, creds *Credentials, a asset.Item, c cur
 
 	subAccounts, ok := accounts.subAccounts[*creds]
 	if !ok {
-		return nil, fmt.Errorf("%w for %s %s",
-			errNoCredentialBalances, exch, creds)
+		return nil, fmt.Errorf("%w for %s %s", errNoCredentialBalances, exch, creds)
 	}
 
 	assets, ok := subAccounts[key.SubAccountAsset{
@@ -180,11 +179,11 @@ func GetBalance(exch, subAccount string, creds *Credentials, a asset.Item, c cur
 		Asset:      a,
 	}]
 	if !ok {
-		return nil, fmt.Errorf("%w for %s SubAccount `%s` %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
+		return nil, fmt.Errorf("%w for %s SubAccount %q %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
 	}
 	bal, ok := assets[c.Item]
 	if !ok {
-		return nil, fmt.Errorf("%w for %s SubAccount `%s` %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
+		return nil, fmt.Errorf("%w for %s SubAccount %q %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
 	}
 	return bal, nil
 }
--- a/exchanges/account/account.go
+++ b/exchanges/account/account.go
@@ -73,7 +73,7 @@ func SubscribeToExchangeAccount(exchange string) (dispatch.Pipe, error) {
 
 // Process processes new account holdings updates
 func Process(h *Holdings, c *Credentials) error {
-	return service.Update(h, c)
+	return service.Save(h, c)
 }
 
 // GetHoldings returns full holdings for an exchange.
@@ -101,28 +101,30 @@ func GetHoldings(exch string, creds *Credentials, assetType asset.Item) (Holding
 		return Holdings{}, fmt.Errorf("%s %w: `%s`", exch, ErrExchangeHoldingsNotFound, assetType)
 	}
 
-	subAccountHoldings, ok := accounts.SubAccounts[*creds]
+	subAccountHoldings, ok := accounts.subAccounts[*creds]
 	if !ok {
 		return Holdings{}, fmt.Errorf("%s %s %s %w %w", exch, creds, assetType, errNoCredentialBalances, ErrExchangeHoldingsNotFound)
 	}
 
 	currencyBalances := make([]Balance, 0, len(subAccountHoldings))
 	cpy := *creds
-	for mapKey, assetHoldings := range subAccountHoldings {
+	for mapKey, assets := range subAccountHoldings {
 		if mapKey.Asset != assetType {
 			continue
 		}
-		assetHoldings.m.Lock()
-		currencyBalances = append(currencyBalances, Balance{
-			Currency:               mapKey.Currency.Currency().Upper(),
-			Total:                  assetHoldings.total,
-			Hold:                   assetHoldings.hold,
-			Free:                   assetHoldings.free,
-			AvailableWithoutBorrow: assetHoldings.availableWithoutBorrow,
-			Borrowed:               assetHoldings.borrowed,
-			UpdatedAt:              assetHoldings.updatedAt,
-		})
-		assetHoldings.m.Unlock()
+		for currItem, bal := range assets {
+			bal.m.Lock()
+			currencyBalances = append(currencyBalances, Balance{
+				Currency:               currItem.Currency().Upper(),
+				Total:                  bal.total,
+				Hold:                   bal.hold,
+				Free:                   bal.free,
+				AvailableWithoutBorrow: bal.availableWithoutBorrow,
+				Borrowed:               bal.borrowed,
+				UpdatedAt:              bal.updatedAt,
+			})
+			bal.m.Unlock()
+		}
 		if cpy.SubAccount == "" && mapKey.SubAccount != "" {
 			// TODO: fix this backwards population
 			// the subAccount here may not be associated with the balance across all subAccountHoldings
@@ -141,13 +143,13 @@ func GetHoldings(exch string, creds *Credentials, assetType asset.Item) (Holding
 }
 
 // GetBalance returns the internal balance for that asset item.
-func GetBalance(exch, subAccount string, creds *Credentials, ai asset.Item, c currency.Code) (*ProtectedBalance, error) {
+func GetBalance(exch, subAccount string, creds *Credentials, a asset.Item, c currency.Code) (*ProtectedBalance, error) {
 	if exch == "" {
 		return nil, fmt.Errorf("cannot get balance: %w", errExchangeNameUnset)
 	}
 
-	if !ai.IsValid() {
-		return nil, fmt.Errorf("cannot get balance: %s %w", ai, asset.ErrNotSupported)
+	if !a.IsValid() {
+		return nil, fmt.Errorf("cannot get balance: %s %w", a, asset.ErrNotSupported)
 	}
 
 	if creds.IsEmpty() {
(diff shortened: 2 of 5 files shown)


exchanges/account/account_types.go
```diff
@@ -34,7 +34,11 @@ type Accounts struct {
 	// TODO: Credential tracker to match to keys that are managed and return
 	// pointer.
 	// TODO: Have different cred struct for centralized verse DEFI exchanges.
-	SubAccounts map[Credentials]map[key.SubAccountCurrencyAsset]*ProtectedBalance
+	subAccounts map[Credentials]map[key.SubAccountAsset]subAccountAsset
+}
+
+type subAccountAsset struct {
```
gloriousCode, line 40: I don't really understand the need for change to the `key` and `subAccountAsset`. I have a [diff here](https://gist.github.com/gloriousCode/9c45dadaeb7cdc9dda728aba7c5206eb) which simply builds a new map during the run to know which assets to clear after an `Update`

gloriousCode (changes requested):
Thanks for your pr!

exchanges/account/account_types.go
junnplus: While it's possible to achieve the goal without changing the `key` and `subAccountAsset`, the original `SubAccountCurrencyAsset` structure flattens all assets and currencies. This requires processing all asset types during each update, which introduces unnecessary overhead. Since both `Update` and `GetHoldings` operations are performed on individual asset types, I believe the flattened structure is not the best approach.

engine/rpcserver.go
```diff
@@ -696,14 +696,14 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
...
 			return errDispatchSystem
 		}
 
-		holdings, ok := data.(*account.Holdings)
-		if !ok {
-			return common.GetTypeAssertError("*account.Holdings", data)
+		holdings, err := exch.GetCachedAccountInfo(stream.Context(), assetType)
+		if err != nil {
+			return err
 		}
```
shazbert, line 707: Suggestion: If we move this call and the accounts info send above the pipe we can remove the initial code outside of this for loop for fetching the initial holdings.

engine/rpcserver.go
```diff
@@ -696,14 +696,14 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
 	}()
 
 	for {
-		data, ok := <-pipe.Channel()
+		_, ok := <-pipe.Channel()
 		if !ok {
 			return errDispatchSystem
 		}
```
shazbert, line 702: Potential leak on channel when there is no data incoming. 
```suggestion
				select {
		case <-stream.Context().Done():
			return stream.Context().Err()
		case _, ok := <-pipe.Channel():
			if !ok {
				return errDispatchSystem
			}
		}
```

exchanges/account/account.go
```diff
@@ -235,46 +248,49 @@ func (s *Service) Update(incoming *Holdings, creds *Credentials) error {
...
+		}
+		assets, exist := subAccounts[accAsset]
+		if !exist {
+			assets = subAccountAsset{
+				balances: make(map[*currency.Item]*ProtectedBalance),
+			}
+			accounts.subAccounts[*creds][accAsset] = assets
 		}
 
+		updated := make(map[*currency.Item]any)
```
shazbert, line 263:
```suggestion
		updated := make(map[*currency.Item]bool)
```

exchanges/account/account.go
```diff
@@ -235,46 +248,49 @@ func (s *Service) Update(incoming *Holdings, creds *Credentials) error {
...
 				errs = common.AppendError(errs, fmt.Errorf("%w for account ID `%s` [%s %s]: %w",
 					errLoadingBalance,
 					incoming.Accounts[x].ID,
 					incoming.Accounts[x].AssetType,
 					incoming.Accounts[x].Currencies[y].Currency,
 					err))
+				continue
+			}
+			assets.balances[accBal.Currency.Item] = bal
+			updated[accBal.Currency.Item] = new(any)
```
shazbert, line 283:
```suggestion
			updated[accBal.Currency.Item] = true
```

exchanges/account/account.go
```diff
@@ -235,46 +248,49 @@ func (s *Service) Update(incoming *Holdings, creds *Credentials) error {
...
+		}
+		for cur, bal := range assets.balances {
+			if _, ok := updated[cur]; !ok {
```
shazbert, line 286:
```suggestion
			if !updated[cur] {
```

exchanges/account/account_test.go
```diff
@@ -378,45 +362,51 @@ func TestUpdate(t *testing.T) {
...
-	}
+	require.NoError(t, err)
 
-	assert.Equal(t, 100.0, b.total)
-	assert.Equal(t, 20.0, b.hold)
+	b, ok = assets.balances[currency.BTC.Item]
+	require.True(t, ok)
 	assert.NotEmpty(t, b.updatedAt)
+	assert.Equal(t, 0.0, b.total)
+	assert.Equal(t, 0.0, b.hold)
```
shazbert, line 405:
```suggestion
	assert.Zero(t, b.total)
	assert.Zero(t, b.hold)
```

shazbert (changes requested):
Thanks Junn, not much from me.

engine/rpcserver.go
junnplus: I’ll use a `init` flag inside the loop to handle the first fetch, removing the initial fetch outside the loop.

engine/rpcserver.go
```diff
@@ -694,16 +664,26 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
...
-		if !ok {
-			return errDispatchSystem
+		select {
+		case <-stream.Context().Done():
+			return stream.Context().Err()
+		case _, ok := <-pipe.Channel():
+			if !ok {
+				return errDispatchSystem
+			}
+		default:
```
shazbert, line 677: Suggestion: Reduce CPU pressure

```diff
diff --git a/engine/rpcserver.go b/engine/rpcserver.go
index 1652d3208..d3caba1a1 100644
--- a/engine/rpcserver.go
+++ b/engine/rpcserver.go
@@ -664,7 +664,9 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
                        log.Errorln(log.DispatchMgr, pipeErr)
                }
        }()
-       init := false
+
+       init := make(chan struct{}, 1)
+       init <- struct{}{}

        for {
                select {
@@ -674,11 +676,7 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
                        if !ok {
                                return errDispatchSystem
                        }
-               default:
-                       if init {
-                               continue
-                       }
-                       init = true
+               case <-init:
                }

                holdings, err := exch.GetCachedAccountInfo(stream.Context(), assetType)
```

shazbert (changes requested):
Thanks @junnplus ! One minor thing left over on new changes.

shazbert (approved):
Changes ACK. 🌞 🌻

engine/rpcserver.go
```diff
@@ -694,16 +664,23 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
 			log.Errorln(log.DispatchMgr, pipeErr)
 		}
 	}()
+	init := make(chan struct{}, 1)
+	init <- struct{}{}
```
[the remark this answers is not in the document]
gbjk: Oh, but ... 😈

engine/rpcserver.go
```diff
@@ -694,16 +664,23 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
...
 
 	for {
```
gbjk, line 670: 🐒 🥃  Cute aside, but can be ignored.
I kept wondering how to localise this into the for scope.
[This works](https://go.dev/play/p/TsAraTqUm-Y).
Shame we can't close the channel.
Don't think it's any better or worse to try to send to a full channel than to try to read from an empty one.
```suggestion
	for init := make(chan struct{}, 1); ; {
```

engine/rpcserver.go
```diff
@@ -694,16 +664,23 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
...
-		if !ok {
-			return errDispatchSystem
+		select {
+		case <-stream.Context().Done():
+			return stream.Context().Err()
+		case _, ok := <-pipe.Channel():
+			if !ok {
+				return errDispatchSystem
+			}
+		case <-init:
```
gbjk, line 678:
```suggestion
		case init <- struct{}{}:
```

exchanges/account/account.go
```diff
@@ -167,26 +169,30 @@ func GetBalance(exch, subAccount string, creds *Credentials, ai asset.Item, c cu
...
 	}]
 	if !ok {
 		return nil, fmt.Errorf("%s %s %s %s %w",
 			exch, subAccount, ai, c, errNoExchangeSubAccountBalances)
 	}
+	bal, ok := assets.balances[c.Item]
+	if !ok {
+		return nil, fmt.Errorf("%s %s %s %s %w",
+			exch, subAccount, ai, c, errNoExchangeSubAccountBalances)
+	}
```
gbjk, line 190: 🐒 Nitpick: `ai` is a idiosyncratic var name and if you DriveBy replace it with `a`
🐒 Other errs in this func suffer the same problems, so fixing them all would be nice
* The error is coming at the end seems like it's going to be weird
* We never checked if subAccount is empty. I'd vote we did that earlier around L149. In Lieu of checking it, we need to quote it in errors as "\`%s\`" in case it's empty
* Don't think we need to split to 2 lines, but you tell me how it looks on your screen with the longer message below
So overall I'd suggest:
```suggestion
		return nil, fmt.Errorf("%w for %s SubAccount `%s` currency %s %s %s %w",			errNoExchangeSubAccountBalances, exch, subAccount, a, c)
	}
```

exchanges/account/account.go
```diff
@@ -167,26 +169,30 @@ func GetBalance(exch, subAccount string, creds *Credentials, ai asset.Item, c cu
...
 	return bal, nil
 }
 
-// Update updates holdings with new account info
-func (s *Service) Update(incoming *Holdings, creds *Credentials) error {
+// Load loads the holdings with new account info
+func (s *Service) Load(incoming *Holdings, creds *Credentials) error {
```
gbjk, line 195: 🥃 I don't think this naming improves things.
Maybe you could share your thinking about the distinction?

I think in my head, `Save` is when you give it everything to store, `Load` is when you want it to go get something, and `Update` is when you give it a partial update.

So in this case, I'm thinking you renamed `Update` to `Load` because it's a full dataset, and not a partial?

If so, need to make that clearer in the comment, because not everyone will assume `Load` means clobber.

I'd vote for `Save` 😄 

Can ignore.

exchanges/account/account.go
```diff
@@ -342,3 +358,16 @@ func (b *ProtectedBalance) GetFree() float64 {
 	defer b.m.Unlock()
 	return b.free
 }
+
+func (b *ProtectedBalance) clean() {
```
gbjk, line 362: 🐒  `clear`, `empty`, `reset`, or `zero` (as a verb)
```suggestion
func (b *ProtectedBalance) reset() {
```

exchanges/account/account_types.go
gbjk: That explanation makes sense to me too 🙂

exchanges/account/account_types.go
```diff
@@ -34,7 +34,11 @@ type Accounts struct {
...
+	balances map[*currency.Item]*ProtectedBalance
 }
```
gbjk, line 42: 🚧  
* This structure actually has nothing to do with subAccounts anymore. And doesn't actually imply anything about assets either. So `currencyBalances` is a better name
* Might be missing something, but why do we need it to be a struct 🤷 
```suggestion
type currencyBalances = map[*currency.Item]*ProtectedBalance
```

gbjk (changes requested):
Thanks for this; Good work 👍 

A few 🐒 nitpicks you can ignore or action as you please.
One 🚧  change request, for the naming and structure of the currency balances.

exchanges/account/account_types.go
junnplus: In the current context, currencyBalances seems like a good name. I'm considering introducing account position data in the future, so using a struct allows for better extensibility.

exchanges/account/account.go
junnplus: For me, both "Save" and "Load" work. If there are more votes for one, I'll update it accordingly.

engine/rpcserver.go
junnplus: Localizing it within the for scope is a good idea, but the current approach is simpler and easier to understand.
So, I'll ignore it.

exchanges/account/account.go
thrasher-: I'm a fan of the original `Update` since we use that terminology with other packages though I'm curious as to your thoughts on preferring Load @junnplus

gbjk (changes requested):

exchanges/account/account.go
junnplus: @thrasher- Considering that asset updates will be incrementally handled via WebSocket in the future, the name `Update` better aligns with the incremental update approach.

exchanges/account/account_types.go
junnplus: Got it! We'll keep it as a map type for now. We can revisit and change it to a struct later when handling the position update.

exchanges/account/account.go
junnplus: @gbjk Okay, I just felt that Update wasn't the best fit here. I'll change it to Save.

shazbert (approved):
tACK. Thanks!

exchanges/account/account_types.go
gbjk: Happy with this 👍 Thanks

exchanges/account/account.go
```diff
@@ -164,29 +166,31 @@ func GetBalance(exch, subAccount string, creds *Credentials, ai asset.Item, c cu
...
+	bal, ok := assets[c.Item]
+	if !ok {
+		return nil, fmt.Errorf("%w for %s SubAccount `%s` %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
 	}
 	return bal, nil
 }
 
-// Update updates holdings with new account info
-func (s *Service) Update(incoming *Holdings, creds *Credentials) error {
+// Save saves the holdings with new account info
```
gbjk, line 192: 🚧 Thanks. And sorry, really sorry, to have one more thing still.
I think you missed my point though:
Whatever we're trying to say about this function with it's name, needs to be explicitly documented.
Such as:
```suggestion
// Save saves the holdings with new account info
// incoming should be a full update, and any missing currencies will be zeroed
```

gbjk (changes requested):
Thanks and 🙈 sorrrrrry

shazbert (approved):
ch ACK!

gbjk (approved):
Good work 🎉

gloriousCode (approved):
tACK! I'm approving based on the overall improvements and that my issue with the chan is minor on a more niche endpoint

engine/rpcserver.go
```diff
@@ -694,16 +664,23 @@ func (s *RPCServer) GetAccountInfoStream(r *gctrpc.GetAccountInfoRequest, stream
...
 		}
 
-		holdings, ok := data.(*account.Holdings)
-		if !ok {
-			return common.GetTypeAssertError("*account.Holdings", data)
+		holdings, err := exch.GetCachedAccountInfo(stream.Context(), assetType)
```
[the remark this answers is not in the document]
junnplus: I think you might have missed this [change](https://github.com/thrasher-corp/gocryptotrader/pull/1846/files#diff-90120f8a75e009e8e9578402e276181f7bf25c5a48df7c3e95ed372185040f9cL273-L277)

The `Publish` call should be made for the account that has undergone asset changes.
So, the notification received from the pipe still needs to fetch the complete information again.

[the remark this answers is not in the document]
gloriousCode: Of course the exchange I test it on only has one asset type 😆 Yes that's fair

exchanges/account/account.go
```diff
@@ -164,29 +166,32 @@ func GetBalance(exch, subAccount string, creds *Credentials, ai asset.Item, c cu
...
+	assets, ok := subAccounts[key.SubAccountAsset{
 		SubAccount: subAccount,
-		Currency:   c.Item,
-		Asset:      ai,
+		Asset:      a,
 	}]
 	if !ok {
-		return nil, fmt.Errorf("%s %s %s %s %w",
-			exch, subAccount, ai, c, errNoExchangeSubAccountBalances)
+		return nil, fmt.Errorf("%w for %s SubAccount `%s` %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
```
thrasher-, line 183:
```suggestion
		return nil, fmt.Errorf("%w for %s SubAccount %q %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
```

exchanges/account/account.go
```diff
@@ -164,29 +166,32 @@ func GetBalance(exch, subAccount string, creds *Credentials, ai asset.Item, c cu
...
+	}
+	bal, ok := assets[c.Item]
+	if !ok {
+		return nil, fmt.Errorf("%w for %s SubAccount `%s` %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
```
thrasher-, line 187:
```suggestion
		return nil, fmt.Errorf("%w for %s SubAccount %q %s %s", errNoExchangeSubAccountBalances, exch, subAccount, a, c)
```

exchanges/account/account_test.go
```diff
@@ -337,20 +318,16 @@ func TestGetFree(t *testing.T) {
 	}
 }
 
-func TestUpdate(t *testing.T) {
+func TestLoad(t *testing.T) {
```
thrasher-, line 321:
```suggestion
func TestSave(t *testing.T) {
```

Since its name has been updated

thrasher- (changes requested):
Thanks @junnplus for this refactor and thanks for addressing all the comments! Just some minor nits and have discovered this issue when testing via `go test -count=10`:


```
--- FAIL: TestGetHoldings (0.00s)
    account_test.go:69: 
                Error Trace:    /home/thrasher/dev/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/account/account_test.go:69
                                dispatcher already running
                Test:           TestGetHoldings
--- FAIL: TestGetBalance (0.00s)
    account_test.go:200: 
                Error Trace:    /home/thrasher/dev/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/account/account_test.go:200
                Error:          Target error should be in err chain:
                                expected: "exchange holdings not found"
                                in chain: "no exchange sub account balances for bruh SubAccount `` spot BTC"
                                        "no exchange sub account balances"
                Test:           TestGetBalance
    account_test.go:238: 
                Error Trace:    /home/thrasher/dev/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/account/account_test.go:238
                                error loading balance for account ID `1337` [spot BTC]: out of sequence
                Test:           TestGetBalance
                Messages:       process must not error
```

junnplus:
@thrasher- The `service` in the `account` package is a singleton, so it cannot be repeatedly tested unless it is explicitly reset before each test.

thrasher- (approved):
Thanks for making those changes and addressing all the feedback @junnplus! In my testing it appears as though count=1 is working consistently compared to master, can address the multi count issue separately

PR #1850, opened 2025-03-17
exchanges: Order Types update

samuael:
This PR updates the order type declarations by introducing hybrid order types, adding time-in-force values, and providing a method to check the order type value, composition, and time-in-force value within an order type variable.
- This PR is implemented on-top of the TimeInForce Update PR.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test X

## Checklist

### DIFF
--- a/exchanges/okx/okx_test.go
+++ b/exchanges/okx/okx_test.go
@@ -421,14 +421,14 @@ func (ok *Okx) underlyingFromInstID(instrumentType, instID string) (string, erro
 			return "", errInvalidInstrumentType
 		}
 		for a := range insts {
-			if insts[a].InstrumentID == instID {
+			if insts[a].InstrumentID.String() == instID {
 				return insts[a].Underlying, nil
 			}
 		}
 	} else {
 		for _, insts := range ok.instrumentsInfoMap {
 			for a := range insts {
-				if insts[a].InstrumentID == instID {
+				if insts[a].InstrumentID.String() == instID {
 					return insts[a].Underlying, nil
 				}
 			}
@@ -4209,7 +4209,7 @@ func TestInstrument(t *testing.T) {
 	assert.Equal(t, currency.BTC.String(), i.ContractValueCurrency, "expected BTC contract value currency")
 	assert.True(t, i.ExpTime.Time().IsZero(), "expected empty expiry time")
 	assert.Equal(t, "BTC-USDC", i.InstrumentFamily, "expected BTC-USDC instrument family")
-	assert.Equal(t, "BTC-USDC-SWAP", i.InstrumentID, "expected BTC-USDC-SWAP instrument ID")
+	assert.Equal(t, "BTC-USDC-SWAP", i.InstrumentID.String(), "expected BTC-USDC-SWAP instrument ID")
 
 	swap := GetInstrumentTypeFromAssetItem(asset.PerpetualSwap)
 	assert.Equal(t, swap, i.InstrumentType, "expected SWAP instrument type")
@@ -6023,14 +6023,14 @@ func (ok *Okx) instrumentFamilyFromInstID(instrumentType, instID string) (string
 			return "", errInvalidInstrumentType
 		}
 		for a := range insts {
-			if insts[a].InstrumentID == instID {
+			if insts[a].InstrumentID.String() == instID {
 				return insts[a].InstrumentFamily, nil
 			}
 		}
 	} else {
 		for _, insts := range ok.instrumentsInfoMap {
 			for a := range insts {
-				if insts[a].InstrumentID == instID {
+				if insts[a].InstrumentID.String() == instID {
 					return insts[a].InstrumentFamily, nil
 				}
 			}
--- a/exchanges/order/order_test.go
+++ b/exchanges/order/order_test.go
@@ -943,9 +943,6 @@ func TestUpdateOrderFromModifyResponse(t *testing.T) {
 	od := Detail{OrderID: "1"}
 	updated := time.Now()
 
-	pair, err := currency.NewPairFromString("BTCUSD")
-	require.NoError(t, err)
-
 	om := ModifyResponse{
 		TimeInForce:     PostOnly | GoodTillTime,
 		Price:           1,
@@ -958,7 +955,7 @@ func TestUpdateOrderFromModifyResponse(t *testing.T) {
 		Status:          1,
 		AssetType:       1,
 		LastUpdated:     updated,
-		Pair:            pair,
+		Pair:            currency.NewBTCUSD(),
 	}
 
 	od.UpdateOrderFromModifyResponse(&om)
@@ -981,19 +978,16 @@ func TestUpdateOrderFromModifyResponse(t *testing.T) {
 }
 
 func TestUpdateOrderFromDetail(t *testing.T) {
-	leet := "1337"
+	t.Parallel()
 
-	updated := time.Now()
-
-	pair, err := currency.NewPairFromString("BTCUSD")
-	require.NoError(t, err)
+	var od *Detail
+	err := od.UpdateOrderFromDetail(nil)
+	require.ErrorIs(t, err, ErrOrderDetailIsNil)
 
 	id, err := uuid.NewV4()
 	require.NoError(t, err)
-
-	var od *Detail
-	err = od.UpdateOrderFromDetail(nil)
-	require.ErrorIs(t, err, ErrOrderDetailIsNil)
+	const leet = "1337"
+	updated := time.Now()
 
 	om := &Detail{
 		TimeInForce:     GoodTillCancel | PostOnly,
@@ -1019,7 +1013,7 @@ func TestUpdateOrderFromDetail(t *testing.T) {
 		Status:          1,
 		AssetType:       1,
 		LastUpdated:     updated,
-		Pair:            pair,
+		Pair:            currency.NewBTCUSD(),
 		Trades:          []TradeHistory{},
 	}
 
@@ -1161,10 +1155,7 @@ func TestValidationOnOrderTypes(t *testing.T) {
 	modifyOrder = new(Modify)
 	require.ErrorIs(t, modifyOrder.Validate(), ErrPairIsEmpty)
 
-	p, err := currency.NewPairFromString("BTC-USD")
-	require.NoError(t, err)
-
-	modifyOrder.Pair = p
+	modifyOrder.Pair = currency.NewBTCUSD()
 	require.ErrorIs(t, modifyOrder.Validate(), ErrAssetNotSet)
 
 	modifyOrder.AssetType = asset.Spot
--- a/backtester/common/common_test.go
+++ b/backtester/common/common_test.go
@@ -142,7 +142,7 @@ func TestDataTypeConversion(t *testing.T) {
 
 func TestFitStringToLimit(t *testing.T) {
 	t.Parallel()
-	for _, ti := range []struct {
+	for _, tc := range []struct {
 		str      string
 		sep      string
 		limit    int
@@ -187,13 +187,9 @@ func TestFitStringToLimit(t *testing.T) {
 			expected: "h",
 		},
 	} {
-		test := ti
-		t.Run(test.str, func(t *testing.T) {
+		t.Run(tc.str, func(t *testing.T) {
 			t.Parallel()
-			result := FitStringToLimit(test.str, test.sep, test.limit, test.upper)
-			if result != test.expected {
-				t.Errorf("received '%v' expected '%v'", result, test.expected)
-			}
+			assert.Equal(t, tc.expected, FitStringToLimit(tc.str, tc.sep, tc.limit, tc.upper))
 		})
 	}
 }
--- a/backtester/engine/grpcserver.go
+++ b/backtester/engine/grpcserver.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"context"
+	"encoding/base64"
 	"errors"
 	"fmt"
 	"math"
@@ -18,7 +19,6 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/backtester/btrpc"
 	"github.com/thrasher-corp/gocryptotrader/backtester/config"
 	gctcommon "github.com/thrasher-corp/gocryptotrader/common"
-	"github.com/thrasher-corp/gocryptotrader/common/crypto"
 	"github.com/thrasher-corp/gocryptotrader/currency"
 	"github.com/thrasher-corp/gocryptotrader/database"
 	"github.com/thrasher-corp/gocryptotrader/database/drivers"
@@ -154,7 +154,7 @@ func (s *GRPCServer) authenticateClient(ctx context.Context) (context.Context, e
 		return ctx, errors.New("basic not found in authorization header")
 	}
 
-	decoded, err := crypto.Base64Decode(strings.Split(authStr[0], " ")[1])
+	decoded, err := base64.StdEncoding.DecodeString(strings.Split(authStr[0], " ")[1])
 	if err != nil {
 		return ctx, errors.New("unable to base64 decode authorization header")
 	}
@@ -235,9 +235,6 @@ func (s *GRPCServer) ExecuteStrategyFromFile(_ context.Context, request *btrpc.E
 	}
 	if err := cfg.Validate(); err != nil {
 		return nil, err
-	}
-	if cfg == nil {
-		return nil, fmt.Errorf("%w backtester config", gctcommon.ErrNilPointer)
 	}
 
 	if !s.config.Report.GenerateReport {
--- a/cmd/websocket_client/main.go
+++ b/cmd/websocket_client/main.go
@@ -1,6 +1,8 @@
 package main
 
 import (
+	"crypto/sha256"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"log"
@@ -10,7 +12,6 @@ import (
 
 	gws "github.com/gorilla/websocket"
 	"github.com/thrasher-corp/gocryptotrader/common"
-	"github.com/thrasher-corp/gocryptotrader/common/crypto"
 	"github.com/thrasher-corp/gocryptotrader/config"
 	"github.com/thrasher-corp/gocryptotrader/encoding/json"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
@@ -99,18 +100,13 @@ func main() {
 	resp.Body.Close()
 	log.Println("Connected to websocket!")
 
-	hash, err := crypto.GetSHA256([]byte(cfg.RemoteControl.Password))
-	if err != nil {
-		log.Println("Unable to generate SHA256 hash from remote control password:", err)
-		return
-	}
-
 	log.Println("Authenticating..")
-	var wsResp WebsocketEventResponse
+	shasum := sha256.Sum256([]byte(cfg.RemoteControl.Password))
 	reqData := WebsocketAuth{
 		Username: cfg.RemoteControl.Username,
-		Password: crypto.HexEncodeToString(hash),
+		Password: hex.EncodeToString(shasum[:]),
 	}
+	var wsResp WebsocketEventResponse
 	err = SendWebsocketEvent("auth", reqData, &wsResp)
 	if err != nil {
 		log.Fatal(err)
--- a/common/crypto/crypto.go
+++ b/common/crypto/crypto.go
@@ -7,8 +7,6 @@ import (
 	"crypto/sha1" //nolint:gosec // Used for exchanges
 	"crypto/sha256"
 	"crypto/sha512"
-	"encoding/base64"
-	"encoding/hex"
 	"errors"
 	"hash"
 	"io"
@@ -22,25 +20,6 @@ const (
 	HashSHA512_384
 	HashMD5
 )
-
-// HexEncodeToString takes in a hexadecimal byte array and returns a string
-func HexEncodeToString(input []byte) string {
-	return hex.EncodeToString(input)
-}
-
-// Base64Decode takes in a Base64 string and returns a byte array and an error
-func Base64Decode(input string) ([]byte, error) {
-	result, err := base64.StdEncoding.DecodeString(input)
-	if err != nil {
-		return nil, err
-	}
-	return result, nil
-}
-
-// Base64Encode takes in a byte array then returns an encoded base64 string
-func Base64Encode(input []byte) string {
-	return base64.StdEncoding.EncodeToString(input)
-}
 
 // GetRandomSalt returns a random salt
 func GetRandomSalt(input []byte, saltLen int) ([]byte, error) {
@@ -58,27 +37,6 @@ func GetRandomSalt(input []byte, saltLen int) ([]byte, error) {
 	}
 	result = append(result, salt...)
 	return result, nil
-}
-
-// GetMD5 returns a MD5 hash of a byte array
-func GetMD5(input []byte) ([]byte, error) {
-	m := md5.New() //nolint:gosec // hash function used by some exchanges
-	_, err := m.Write(input)
-	return m.Sum(nil), err
-}
-
-// GetSHA512 returns a SHA512 hash of a byte array
-func GetSHA512(input []byte) ([]byte, error) {
-	sha := sha512.New()
-	_, err := sha.Write(input)
-	return sha.Sum(nil), err
-}
-
-// GetSHA256 returns a SHA256 hash of a byte array
-func GetSHA256(input []byte) ([]byte, error) {
-	sha := sha256.New()
-	_, err := sha.Write(input)
-	return sha.Sum(nil), err
 }
 
 // GetHMAC returns a keyed-hash message authentication code using the desired
--- a/common/crypto/crypto_test.go
+++ b/common/crypto/crypto_test.go
@@ -2,35 +2,11 @@ package crypto
 
 import (
 	"bytes"
-	"encoding/hex"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
-
-func TestHexEncodeToString(t *testing.T) {
-	t.Parallel()
-	assert.Equal(t, "737472696e67", HexEncodeToString([]byte("string")))
-}
-
-func TestBase64Decode(t *testing.T) {
-	t.Parallel()
-
-	r, err := Base64Decode("aGVsbG8=")
-	require.NoError(t, err, "Base64Decode must not error")
-	assert.Equal(t, []byte("hello"), r, "Base64Decode should return the correct byte slice")
-
-	_, err = Base64Decode("-")
-	assert.Error(t, err, "Base64Decode should error on invalid input")
-}
-
-func TestBase64Encode(t *testing.T) {
-	t.Parallel()
-
-	assert.Equal(t, "aGVsbG8=", Base64Encode([]byte("hello")),
-		"Base64Encode should return the correct base64 string")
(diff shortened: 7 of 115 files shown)


exchanges/order/order_types.go
```diff
@@ -354,33 +348,42 @@ const (
...
+	TWAP            // time-weighted average price.
+	Chase           // chase order. See https://www.okx.com/docs-v5/en/#order-book-trading-algo-trading-post-place-algo-order
+
+	// Hybrid order types
+	OptimalLimitIOC                  = 1<<iota | Type(ImmediateOrCancel)
+	MarketMakerProtection                                                     // market-maker-protection used with portfolio margin mode. See https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-place-order
+	MarketMakerProtectionAndPostOnly = MarketMakerProtection | Type(PostOnly) // market-maker-protection and post-only mode. Used in Okx exchange orders.
+	StopLimit                        = Stop | Limit
+	StopMarket                       = Stop | Market
+	TakeProfitMarket                 = TakeProfit | Market
```
shazbert, line 386: I would suggest to keep TIF values as TIF and have it processed on the exchange side e.g.

OptimalLimitIOC -> OptimalLimit then exchange search for Is(order.ImmediateOrCancel)
MarketMakerProtectionAndPostOnly I would remove completely and do Is(order.PostOnly)

That way there is no chance of collision. The others can stay as is and it's less confusing.

Would be good to group these types in future to:

Type: Market, Limit
Conditional: Stop, TakeProfit, TrailingStop etc
ExchangeAlgo(or whatever name is best): TWAP, Chase, MarketMakerProtection
TimeInForce: What you have done so far.  

The above has some caveats but is more explicit and can be done step by step, the flagging and mixing could be a massive complex overhaul.

shazbert (commented):
Thanks for the PR @samuael! Just a comment on current design that might be applicable. I am cautious on the added potential complexity and ongoing maintainability of this change. Added a suggestion for us to discuss and get your inputs.

shazbert (changes requested):
There are build tests that are failure could you please update them 🙏 🕊️ thank you boss.

exchanges/order/order_types.go
```diff
@@ -350,28 +350,34 @@ const (
...
 	Trigger
-	OptimalLimitIOC
-	OCO                              // One-cancels-the-other order
-	ConditionalStop                  // One-way stop order
-	MarketMakerProtection            // market-maker-protection used with portfolio margin mode. See https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-place-order
-	MarketMakerProtectionAndPostOnly // market-maker-protection and post-only mode. Used in Okx exchange orders.
-	TWAP                             // time-weighted average price.
-	Chase                            // chase order. See https://www.okx.com/docs-v5/en/#order-book-trading-algo-trading-post-place-algo-order
+	OCO             // One-cancels-the-other order
+	ConditionalStop // One-way stop order
```
shazbert, line 371: Suggestion: This is a bit more work but while you are here, In our order types we can start splitting all these out into specific fields for clarity and trading experience. 

```go
type Submit struct {
    Type        OrderType     // Core order type (e.g., Limit, Market)
    TimeInForce TimeInForce   // Time-in-force (e.g., GTC, IOC, FOK, PostOnly)
    Condition   ConditionType // Conditional trigger (e.g., Stop, TakeProfit)
    SpecialMode SpecialMode   // Exchange-specific modes (e.g., TWAP, MarketMakerProtection)
    // Existing fields (e.g., Pair, Amount, Side, etc.) remain unchanged
}
```

Then you don't need to define hybrid order types.

exchanges/order/order_test.go
```diff
@@ -785,6 +784,7 @@ func TestStringToOrderType(t *testing.T) {
 	for i := range cases {
 		testData := &cases[i]
 		t.Run(testData.in, func(t *testing.T) {
+			t.Parallel()
```
shazbert, line 851: suggestion: Seeing as there is not much needed to be done and t.Run doesn't need to be called but can stay as is.

shazbert (approved):
State approved. But with a couple of optional change requests. This will blow out a few things. Thanks Boss.

exchanges/order/order_types.go
samuael: Why do you think separating these types should be done since they all represent a single parameter type "orderType", for all/most exchanges. Don't you think doing so will cause a confusion and the need to check the three values just to know what order type is passed?

shazbert: Yeah that is a valid point on causing confusion in the wrapper implementations and increased toil there, I was more coming from a trading experience and clarity on the user side. It's still optional we can circle back to it if we need to change happy to keep as is.

exchanges/okx/helpers.go
```diff
@@ -24,15 +24,15 @@ func orderTypeFromString(orderType string) (order.Type, order.TimeInForce, error
...
-		return order.MarketMakerProtectionAndPostOnly, order.PostOnly, nil
+		return order.MarketMakerProtection, order.PostOnly, nil
 	case "twap":
 		return order.TWAP, order.UnknownTIF, nil
 	case "move_order_stop":
-		return order.TrailingStop, order.UnknownTIF, nil
+		return order.TrailingStop, 0, nil
 	case "chase":
 		return order.Chase, order.UnknownTIF, nil
 	default:
```
gbjk, line 38: 🗻 OOS DriveBy: Add quotes around the erroring value in the default.
Can't _suggest_ on this line, so I've left his comment on the default case. Soz 🤷 
```suggestion
	default:
		return order.UnknownType, order.UnknownTIF, fmt.Errorf("%w %q", order.ErrTypeIsInvalid, orderType)
```

exchanges/okx/helpers.go
```diff
@@ -66,6 +73,14 @@ func orderTypeString(orderType order.Type, tif order.TimeInForce) (string, error
...
+		switch tif {
+		case order.PostOnly:
+			return orderPostOnly, nil
+		case order.FillOrKill:
+			return orderFOK, nil
+		case order.ImmediateOrCancel:
+			return orderIOC, nil
+		}
 		return "", fmt.Errorf("%w: `%v`", order.ErrUnsupportedOrderType, orderType)
```
gbjk: 🗻 OOS DriveBy
```suggestion
		return "", fmt.Errorf("%w: %q", order.ErrUnsupportedOrderType, orderType)
```

exchanges/okx/helpers.go
```diff
@@ -42,21 +42,28 @@ func orderTypeFromString(orderType string) (order.Type, order.TimeInForce, error
...
+			return "mmp_and_post_only", nil
+		}
+		return "mmp", nil
+	case order.OptimalLimit:
+		return "optimal_limit_ioc", nil
+	case order.Limit:
+		if tif == order.PostOnly {
+			return orderPostOnly, nil
+		}
+		return orderType.Lower(), nil
```
gbjk: 🚧  Safer to use either inline or const. This is an exchange API, and we're tying the value we'll send to the value we use internally for orderType. Especially since we inline the others anyway.
Previously I was marking this type of issue as a nitpick, but I'm now more definite about the nature of the risk with this pattern.
```suggestion
		return "lower", nil
```

exchanges/okx/helpers.go
```diff
@@ -42,21 +42,28 @@ func orderTypeFromString(orderType string) (order.Type, order.TimeInForce, error
...
+	case order.Market:
+		switch tif {
+		case order.FillOrKill:
+			return orderFOK, nil
+		case order.ImmediateOrCancel:
+			return orderIOC, nil
+		}
+		return orderType.Lower(), nil
```
gbjk: 🚧 Same as above
```suggestion
		return "market", nil
```

exchanges/okx/okx_test.go
```diff
@@ -5846,21 +5846,21 @@ func TestOrderTypeString(t *testing.T) {
...
+		{OrderType: order.MarketMakerProtection, TIF: order.PostOnly}:   {Expected: "mmp_and_post_only"},
+		{OrderType: order.Liquidation, TIF: order.UnknownTIF}:           {Error: order.ErrUnsupportedOrderType},
+		{OrderType: order.OCO, TIF: order.UnknownTIF}:                   {Expected: "oco"},
+		{OrderType: order.TrailingStop, TIF: order.UnknownTIF}:          {Expected: "move_order_stop"},
+		{OrderType: order.Chase, TIF: order.UnknownTIF}:                 {Expected: "chase"},
+		{OrderType: order.TWAP, TIF: order.UnknownTIF}:                  {Expected: "twap"},
+		{OrderType: order.ConditionalStop, TIF: order.UnknownTIF}:       {Expected: "conditional"},
+		{OrderType: order.Trigger, TIF: order.UnknownTIF}:               {Expected: "trigger"},
 	}
 	for tc, val := range orderTypesToStringMap {
```
gbjk, line 5868: 🚧 Could you use`t.Run` or `ErrorIsf` please?

exchanges/okx/okx_test.go
```diff
@@ -6142,9 +6142,9 @@ func TestOrderTypeFromString(t *testing.T) {
...
 		"fok":               {OType: order.Limit, TIF: order.FillOrKill},
 		"ioc":               {OType: order.Limit, TIF: order.ImmediateOrCancel},
-		"optimal_limit_ioc": {OType: order.OptimalLimitIOC, TIF: order.ImmediateOrCancel},
+		"optimal_limit_ioc": {OType: order.OptimalLimit, TIF: order.ImmediateOrCancel},
 		"mmp":               {OType: order.MarketMakerProtection},
-		"mmp_and_post_only": {OType: order.MarketMakerProtectionAndPostOnly, TIF: order.PostOnly},
+		"mmp_and_post_only": {OType: order.MarketMakerProtection, TIF: order.PostOnly},
 		"trigger":           {OType: order.UnknownType, Error: order.ErrTypeIsInvalid},
 		"chase":             {OType: order.Chase},
 		"move_order_stop":   {OType: order.TrailingStop},
```
gbjk, line 6155: 🐒 Could you DriveBy add `t.Run` or `assquire.*f` (assert or require . ErrorIsf, etc) below?

exchanges/order/order_types.go
```diff
@@ -350,28 +350,34 @@ const (
 // Type enforces a standard for order types across the code base
 type Type uint32
 
+// Is checks to see if the enum contains the flag
+func (t Type) Is(in Type) bool {
+	return in != 0 && t&in == in
```
gbjk: 🐒 The existing comment didn't really help, because the terms `enum` and `flag` aren't part of the method.
In fixing that, I found it necessary to rename `in`.
Note:
* `t` is not an enum 🙂 
* `cmp` probably/hopefully is, though, but I'm not suggesting we enforce that `cmp` is only one value.
```suggestion
// Is checks to see if the Type contains the Type cmp
func (t Type) Is(cmp Type) bool {
	return cmp != 0 && t&cmp == cmp
```

exchanges/order/order_types.go
```diff
@@ -350,28 +350,34 @@ const (
...
-	OptimalLimitIOC
-	OCO                              // One-cancels-the-other order
-	ConditionalStop                  // One-way stop order
-	MarketMakerProtection            // market-maker-protection used with portfolio margin mode. See https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-place-order
-	MarketMakerProtectionAndPostOnly // market-maker-protection and post-only mode. Used in Okx exchange orders.
-	TWAP                             // time-weighted average price.
-	Chase                            // chase order. See https://www.okx.com/docs-v5/en/#order-book-trading-algo-trading-post-place-algo-order
+	OCO             // One-cancels-the-other order
+	ConditionalStop // One-way stop order
+	TWAP            // time-weighted average price.
```
gbjk: 🦍 
```suggestion
	TWAP            // time-weighted average price
```

exchanges/order/order_types.go
```diff
@@ -350,28 +350,34 @@ const (
...
+	Chase           // chase order. See https://www.okx.com/docs-v5/en/#order-book-trading-algo-trading-post-place-algo-order
+	OptimalLimit
+	MarketMakerProtection // market-maker-protection used with portfolio margin mode. See https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-place-order
```
gbjk: 🐒 I don't think we can talk about portfolio margin mode or okx here, since [MMP isn't an okx only thing](https://tinyurl.com/24s67a7h).
```suggestion
	MarketMakerProtection
```

exchanges/order/order_types.go
```diff
@@ -350,28 +350,34 @@ const (
...
-	OCO                              // One-cancels-the-other order
-	ConditionalStop                  // One-way stop order
-	MarketMakerProtection            // market-maker-protection used with portfolio margin mode. See https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-place-order
-	MarketMakerProtectionAndPostOnly // market-maker-protection and post-only mode. Used in Okx exchange orders.
-	TWAP                             // time-weighted average price.
-	Chase                            // chase order. See https://www.okx.com/docs-v5/en/#order-book-trading-algo-trading-post-place-algo-order
+	OCO             // One-cancels-the-other order
+	ConditionalStop // One-way stop order
+	TWAP            // time-weighted average price.
+	Chase           // chase order. See https://www.okx.com/docs-v5/en/#order-book-trading-algo-trading-post-place-algo-order
```
gbjk: 🦍 This nitpick is more superficial than the following one 🙂 
```suggestion
	Chase           // chase limit order
```

exchanges/order/orders.go
```diff
@@ -701,11 +691,21 @@ func (t Type) String() string {
...
-		return "OPTIMAL_LIMIT_IOC"
 	case OCO:
 		return "OCO"
+	case OptimalLimit:
+		return "OPTIMAL_LIMIT"
+	case MarketMakerProtection:
+		return "MMP"
 	default:
+		switch {
+		case t == AnyType:
```
cranktakular: Consider moving to the outer switch. It should function the same both ways, simply saving on a few characters.

exchanges/order/orders.go
```diff
@@ -701,11 +691,21 @@ func (t Type) String() string {
...
+			return "ANY"
+		case t.Is(StopLimit):
+			return "STOP LIMIT"
+		case t.Is(StopMarket):
```
cranktakular: Why are Stop Limit and Stop Market handled in this way? They're defined in a similar way to TakeProfitMarket, which seems to operate fine being grouped with the rest.

Are they uniquely provided alongside other flags, such that only they should return even if other flags are set as well?

exchanges/order/order_test.go
```diff
@@ -766,14 +767,12 @@ func TestStringToOrderType(t *testing.T) {
...
 		{"oCo", OCO, nil},
 		{"mMp", MarketMakerProtection, nil},
-		{"Mmp_And_Post_oNly", MarketMakerProtectionAndPostOnly, nil},
 		{"tWaP", TWAP, nil},
 		{"TWAP", TWAP, nil},
 		{"woahMan", UnknownType, errUnrecognisedOrderType},
 		{"chase", Chase, nil},
 		{"MOVE_ORDER_STOP", TrailingStop, nil},
 		{"mOVe_OrdeR_StoP", TrailingStop, nil},
-		{"optimal_limit_IoC", OptimalLimitIOC, nil},
```
cranktakular, line 776: A test was not re-added for OptimalLimit, leaving it without coverage here.

exchanges/okx/helpers.go
```diff
@@ -24,15 +24,15 @@ func orderTypeFromString(orderType string) (order.Type, order.TimeInForce, error
...
 	case "mmp":
 		return order.MarketMakerProtection, order.UnknownTIF, nil
 	case "mmp_and_post_only":
-		return order.MarketMakerProtectionAndPostOnly, order.PostOnly, nil
+		return order.MarketMakerProtection, order.PostOnly, nil
 	case "twap":
 		return order.TWAP, order.UnknownTIF, nil
 	case "move_order_stop":
-		return order.TrailingStop, order.UnknownTIF, nil
+		return order.TrailingStop, 0, nil
```
cranktakular: A bit strange to be returning 0 instead of order.UnknownTIF on only this line.

It's functionally the same, just aesthetically odd.

exchanges/okx/okx_test.go
```diff
@@ -5846,21 +5846,21 @@ func TestOrderTypeString(t *testing.T) {
...
+		{OrderType: order.MarketMakerProtection, TIF: order.UnknownTIF}: {Expected: "mmp"},
+		{OrderType: order.MarketMakerProtection, TIF: order.PostOnly}:   {Expected: "mmp_and_post_only"},
+		{OrderType: order.Liquidation, TIF: order.UnknownTIF}:           {Error: order.ErrUnsupportedOrderType},
+		{OrderType: order.OCO, TIF: order.UnknownTIF}:                   {Expected: "oco"},
+		{OrderType: order.TrailingStop, TIF: order.UnknownTIF}:          {Expected: "move_order_stop"},
+		{OrderType: order.Chase, TIF: order.UnknownTIF}:                 {Expected: "chase"},
+		{OrderType: order.TWAP, TIF: order.UnknownTIF}:                  {Expected: "twap"},
+		{OrderType: order.ConditionalStop, TIF: order.UnknownTIF}:       {Expected: "conditional"},
+		{OrderType: order.Trigger, TIF: order.UnknownTIF}:               {Expected: "trigger"},
 	}
```
cranktakular, line 5867: This no longer gets coverage on the tif switch; add three cases with unknown orderTypes, and the appropriate TIFs.

cranktakular (changes requested):
Got a few changes I'd like to see implemented.

exchanges/order/orders.go
samuael: Change applied and Unit test added.

exchanges/order/orders.go
samuael: I assumed I did add all these derived instances together. Anyways, the new approach works good now.

exchanges/okx/okx_test.go
```diff
@@ -5852,19 +5852,22 @@ func TestOrderTypeString(t *testing.T) {
...
+		{OrderType: order.Chase, TIF: order.GoodTillCancel}:             {Expected: orderChase},
+		{OrderType: order.TWAP, TIF: order.ImmediateOrCancel}:           {Expected: orderTWAP},
+		{OrderType: order.ConditionalStop, TIF: order.GoodTillDay}:      {Expected: orderConditional},
+		{OrderType: order.Trigger, TIF: order.UnknownTIF}:               {Expected: orderTrigger},
 	}
 	for tc, val := range orderTypesToStringMap {
 		orderTypeString, err := orderTypeString(tc.OrderType, tc.TIF)
-		require.ErrorIs(t, err, val.Error)
+		require.ErrorIsf(t, err, val.Error, "expected %v, got %v", val.Error, err)
 		assert.Equal(t, val.Expected, orderTypeString)
```
gbjk: 🚧 We never want to use `require.*f` to reprint the same params it's going to print anyway.
What we needed was the context of the error, which we still don't have, in either of these.
Note: I'm assuming these stringify okay. Untested.
```suggestion
		t.Run(tc.OrderType.String()+"/"+tc.TIF.String(), func(t *testing.T){
			t.Parallel()
			orderTypeString, err := orderTypeString(tc.OrderType, tc.TIF)
			require.ErrorIs(t, err, val.Error)
			assert.Equal(t, val.Expected, orderTypeString)
		})
```

exchanges/okx/okx_test.go
```diff
@@ -6152,10 +6154,14 @@ func TestOrderTypeFromString(t *testing.T) {
 		"abcd":              {OType: order.UnknownType, Error: order.ErrTypeIsInvalid},
 	}
 	for s, exp := range orderTypeStrings {
-		oType, tif, err := orderTypeFromString(s)
-		require.ErrorIs(t, err, exp.Error)
-		assert.Equal(t, exp.OType, oType)
-		assert.Equal(t, exp.TIF.String(), tif.String(), s)
+		t.Run(s, func(t *testing.T) {
+			t.Parallel()
+			s := s
```
gbjk: 🚧 I don't understand why this line is here 🙂 
```suggestion
```

exchanges/okx/okx_test.go
```diff
@@ -6152,10 +6154,14 @@ func TestOrderTypeFromString(t *testing.T) {
...
+			oType, tif, err := orderTypeFromString(s)
+			require.ErrorIs(t, err, exp.Error)
+			assert.Equal(t, exp.OType, oType)
+			assert.Equal(t, exp.TIF.String(), tif.String(), s)
```
gbjk:
```suggestion
			assert.Equal(t, exp.TIF.String(), tif.String())
```

exchanges/order/order_test.go
```diff
@@ -305,6 +305,39 @@ func TestOrderTypes(t *testing.T) {
...
+	{MarketMakerProtection, "MMP"},
+	{AnyType, "ANY"},
+	{UnknownType | Limit, "LIMIT"},
+	{StopMarket | ConditionalStop, "UNKNOWN"},
+}
+
+func TestOrderTypeToString(t *testing.T) {
+	t.Parallel()
+	for x := range orderToToStringsList {
+		assert.Equal(t, orderToToStringsList[x].String, orderToToStringsList[x].OrderType.String())
```
gbjk: 🦍  Just a nitpick
```suggestion
	for _, tt := range orderToToStringsList {
		assert.Equal(t, tt.String, tt.OrderType.String())
```

exchanges/order/order_test.go
```diff
@@ -305,6 +305,39 @@ func TestOrderTypes(t *testing.T) {
...
+	{IOS, "IOS"},
+	{Liquidation, "LIQUIDATION"},
+	{Trigger, "TRIGGER"},
+	{OCO, "OCO"},
+	{OptimalLimit, "OPTIMAL_LIMIT"},
+	{MarketMakerProtection, "MMP"},
+	{AnyType, "ANY"},
+	{UnknownType | Limit, "LIMIT"},
+	{StopMarket | ConditionalStop, "UNKNOWN"},
+}
```
gbjk: 🦍  Couldn't this be a `map[OrderType]string` to simplify simply

exchanges/order/order_types.go
```diff
@@ -350,9 +350,9 @@ const (
 // Type enforces a standard for order types across the code base
 type Type uint32
 
-// Is checks to see if the enum contains the flag
-func (t Type) Is(in Type) bool {
-	return in != 0 && t&in == in
+// Is checks to see if the Type contains the Type cmp
+func (t Type) Is(cmp Type) bool {
+	return cmp != 0 && t&cmp == cmp
 }
```
gbjk: 🚧 With the change to not using "Type.Is()` in `Type.String()`, I don't think this is used anymore.
```suggestion
```

exchanges/okx/okx_test.go
samuael: I think we should start declaring a local scope for such loop variables just to prevent from using an updated value of the loop variable.

exchanges/order/order_test.go
samuael: Looping over the structs is much faster than using the map. That's why.
Screenshot 2025-06-07 at 12 42 39 in the afternoon

exchanges/order/order_types.go
samuael: I added a unit test for this and I believe it is better to keep it than to remove as it has relevance.

exchanges/okx/okx_test.go
samuael: Change applied!

gbjk: Yep, I'm guessing you remembered that loopvars are not an issue since 1.22.
Thanks.

exchanges/order/order_test.go
gbjk: But ... this is a test?
I don't think we need to save nanoseconds in our test code.

exchanges/order/order_types.go
gbjk: ~I think it only has relevance if we have hybrid order types, which leads me to my only outstanding issue now.~

Okay, nevermind. If the hybrid order types are added, I'm happy with this function, even if it's unused.
Just noting that we try not to have unused code.

gbjk (approved):
Thanks for making those changes 🙏 

Overall, I'm concerned that we don't have a use case for changing to hybrid order types right now, so those specific changes are adding risk[1] without immediate benefit.

But I'll approve with that note 👍 

---
[1] _All changes introduce risk_

exchanges/order/order_test.go
samuael: 🤔  just to apply efficiency to wherever in the code. If the test samples were just 2 or 4, I wouldn't use such approach.

exchanges/order/orders.go
```diff
@@ -1120,16 +1120,14 @@ func StringToOrderType(oType string) (Type, error) {
 		return AnyType, nil
 	case Trigger.String():
 		return Trigger, nil
-	case OptimalLimitIOC.String():
-		return OptimalLimitIOC, nil
+	case OptimalLimit.String():
```
cranktakular: Tangential to the main work you're doing, but switches run faster when their cases are constants (or can be optimised into them, like the hardcoded strings in other branches here). Replacing these calls to .String() with constants that are shared between the String() method and this one would speed this up quite a bit.

Still, that's something we have to clean up in a lot of places, so I'm not too fussed about it being fixed here.

exchanges/okx/okx_test.go
cranktakular: This doesn't appear to have been fixed.

cranktakular (approved):
Hit an error in the exchange wrapper standards for Bithumb (UpdateOrderbook, cannot load book for exchange amount cannot be less or equal to zero) while testing this, that I didn't hit on the master branch. If that's due to the master branch being slightly more updated, or due to the failure being intermittent, I can tACK this. Otherwise that should be addressed.

samuael:
> amount cannot be less or equal to zero

This instrument has a zero amount value when the data is loaded from the source and that causing the `" Bids: amount cannot be less or equal to zero"` error.

Screenshot 2025-06-11 at 6 02 31 in the evening

So, this is an issue created by this PR. ✅

exchanges/order/order_test.go
```diff
@@ -297,6 +297,32 @@ func TestTitle(t *testing.T) {
 	require.Equal(t, "Limit", orderType.Title())
 }
 
+var orderComparisonList = []struct {
```
thrasher-: This can be merged into TestOrderIs, enhanced with t.Run + t.Parallel with subtests since it isn't used anywhere else

exchanges/order/order_test.go
```diff
@@ -305,6 +331,39 @@ func TestOrderTypes(t *testing.T) {
...
+	{Trigger, "TRIGGER"},
+	{OCO, "OCO"},
+	{OptimalLimit, "OPTIMAL_LIMIT"},
+	{MarketMakerProtection, "MMP"},
+	{AnyType, "ANY"},
+	{UnknownType | Limit, "LIMIT"},
+	{StopMarket | ConditionalStop, "UNKNOWN"},
+}
+
+func TestOrderTypeToString(t *testing.T) {
```
thrasher-, line 335: Same as my other comment

thrasher- (changes requested):
Just two minor things, then LGTG. Nice work! The Bithumb issue occurs on master randomly too, so it's not to do with this PR

thrasher- (approved):
Happy bings all round. Changes ACK

PR #1851, opened 2025-03-18
okx: Remove WsResponseMultiplexer and various refactors

shazbert:
# PR Description

* Remove `WsResponseMultiplexer` and assortment of methods, types and helper functions
* Use `common.Counter` for request ID response routing
* Uses `SendAuthenticatedWebsocketRequest` now for request to response matching
* Changes string status code to integer
* change `validatePlaceOrderParams` method to PlaceOrderRequestParam `Validate` method add test
* change `validateSpreadOrderParam` method to SpreadOrderParam `Validate` method add test
* shifts unused func `underlyingFromInstID` to test file
* removes `Public Channel Subscriptions` methods and tests as they were bypassing subscription storage and causing inconsistency 
* Shifts all websocket trading functionality to `ws_request.go` with a test file and testing has been carried out.

* Sets global rate limiter to be shared between rest and websocket instances allowing correct request rate limited between them

NOTE: Spread/okx_business_websocket is not tested and cannot be appropriately integrated without multiconnection handling

Fixes # (issue)

* Fixes websocket submit bug due to `ExpiryTime` field not supposed to be in outbound payload

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/okx/ws_requests_test.go
```diff
@@ -0,0 +1,227 @@
...
+
+	sharedtestvalues.SkipTestIfCredentialsUnset(t, ok, canManipulateRealOrders)
+
+	out := PlaceOrderRequestParam{
+		InstrumentID: btcusdt,
+		TradeMode:    TradeModeCash,
+		Side:         "Buy",
+		OrderType:    "limit",
+		Amount:       0.0001,
+		Price:        -1, // Intentional fail
```
gbjk, line 54: 🤷 Erm. Okay. But ... L59 and 60 suggest otherwise?

gbjk (changes requested):
Great work 👍 

Some change requests regarding the error handling for batches.

Otherwise looks good.

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,393 @@
...
+
+	if err := json.Unmarshal(incoming, &intermediary); err != nil {
+		return err
+	}
+
+	// Codes 0-2 have responses and can be handled individually by the caller, while codes > 2 are endpoint errors with
+	// messages and no responses, checked here.
+	if intermediary.Code > 2 {
+		return fmt.Errorf("status code: %d message: %s", intermediary.Code, intermediary.Message)
+	}
```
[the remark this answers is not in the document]
gloriousCode: Yeah I brought up the same last time. I think if you were truly wanting to go with a "the caller can handle it" as an approach shazbert, you may wish to return more than just the `Data` part of the response and you could instead be returning a struct of the full response. That way you can properly handle all scenarios. The caller of the ws request funcs can ultimately choose to ignore certain error types in the response

exchanges/okx/okx.go
```diff
@@ -5865,7 +5853,7 @@ func (ok *Okx) SendHTTPRequest(ctx context.Context, ep exchange.URL, f request.E
 		path := endpoint + requestPath
 		headers := make(map[string]string)
 		headers["Content-Type"] = "application/json"
-		if _, okay := ctx.Value(testNetVal).(bool); okay {
+		if simulate, okay := ctx.Value(testNetVal).(bool); okay && simulate {
```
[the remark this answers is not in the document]
shazbert: Conflicts with method okx variable

exchanges/okx/ws_requests.go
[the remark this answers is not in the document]
shazbert: > The caller's aren't checking this code either, so you'll have a call that partially or completely fails, and the caller's err var will be nil.

Sorry meaning caller as in library caller not direct function invocation. Partial failure or Total Failure can be determined by ranging over the returned orders (this will be done in a wrapper function and a map returned). I think we may be over complicating this; because 0;1;2 codes are not request level errors so return nil to me, is justifiable. Why do err != nil then range over orders to get specific level error, then append those errors together into a single error and return a slice which contains the errors associated as well. Just leaving it as is for batches seems easier to consume? I may be wrong. 🤷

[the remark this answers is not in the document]
gbjk: > because 0;1;2 codes are not request level errors

I don't understand this sentence 🙁 
It's [documented](https://www.okx.com/docs-v5/en/#error-code) as 2 = `Bulk operation partially succeeded` and 1 = `Operation failed`.

I'm going to stick to saying that it's the job of the API library function to faithfully take errors out of the response and put them into `err`. It's also it's job to add associative data to the errors, e.g. which order number failed, etc.

gbjk (changes requested):
Happy with the changes made so far. Thank you 🙏 

Outstanding issue for me is:
* Batch orders need to represent their errors in err
* Response error codes 1 & 2 are documented and should be returned with error messages matching the [documentation](https://www.okx.com/docs-v5/en/#error-code)

As a special case, I think that batch funcs will need to combine err returned from `SendHTTPRequest` with errors in the returned structure, probably regardless of whether it's 1 or 2.

exchanges/okx/ws_requests.go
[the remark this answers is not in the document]
shazbert: The `1` (total failure) and `2` (partial failure) codes in the ws is a direct status of a batch operation, but the actual error details (e.g., which order failed and why) are always tied to the individual order responses in the `Data` part of the response. For codes greater than `2`, the WebSocket response doesn’t include any order data, so we can return a non-nil `err` to indicate a complete failure and propagate that up the chain.

In the current implementation, the batch response returns as-is, and a future wrapper function or consumer can range over the returned orders to build a map of order IDs/ submit.Responses/whatever to their specific errors, if any. This keeps it simple and lets the caller decide how to handle partial failures, which is determined by iteration. 

Maybe i'm not really understanding the request, maybe send me a patch and I can apply it.

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,393 @@
...
+		// Request effective deadline. Unix timestamp format in milliseconds, e.g. 1597026383085
+	}{
+		ID:        id,
+		Operation: operation,
+		Arguments: payload,
+	}
+
+	incoming, err := ok.Websocket.AuthConn.SendMessageReturnResponse(ctx, epl, id, outbound)
+	if err != nil {
+		fmt.Println("authconn problems bruh", err)
```
gloriousCode, line 366: Get this 💩 outta here!

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,393 @@
...
+		Data: result,
+	}
+
+	if err := json.Unmarshal(incoming, &intermediary); err != nil {
+		return err
+	}
+
+	// Codes 0-2 have responses and can be handled individually by the caller, while codes > 2 are endpoint errors with
+	// messages and no responses, checked here.
+	if intermediary.Code > 2 {
```
gloriousCode, line 388: I don't believe enough is being done. I am partial to what you are saying, but I can't bring myself to approve the pr as is.
When we get to what GCT does as a public repo for exchanges, it is to implement the api and handle the response. We do this across all exchanges. You have implemented the interim json handling across many exchanges to handle such scenarios. Take okx batch cancel for example:
```
  okx_test.go:3682: 
        	Error Trace:	/Users/wif/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/okx/okx_test.go:3682
        	            	authenticated request failed, status code: `1` status message: `All operations failed`
        	Test:       	TestCancelBatchOrders
```

I don't believe this test should pass:
```
    ws_requests_test.go:61: [{OrderID: RequestID: ClientOrderID: Tag: StatusCode:51008 StatusMessage:Order failed. Insufficient USDT balance in account. Timestamp:1743383281273}]
--- PASS: TestWSPlaceMultipleOrder (0.14s)
```
I feel some of the resistance to change it stems from the difficulty of the solution.
I am still interested in us meeting somewhere in the middle. So I propose you add `Error` functions for the response types.

I propose that status code 1 should return error and all response functions should implement an interface that has an `Error()` function. Please see diff for a rough draft
https://gist.github.com/gloriousCode/27ad282a2e9b7b9946481df7319a8248

For code 2, I am more willing for you to not return an error there. By all responses having the function `Error()` it at least allows for users to get the error directly from the response themselves too. I would expect you to add commentary that emphasises this functionality

Result:
```
    ws_requests_test.go:59: 
        	Error Trace:	/Users/wif/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/okx/ws_requests_test.go:59
        	            	: invalid websocket request status code: `51008` status message: `Order failed. Insufficient USDT balance in account.`
        	Test:       	TestWSPlaceMultipleOrder
```

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,393 @@
...
+// WSCancelAllSpreadOrders cancels all spread orders and return success message through the websocket channel.
+func (ok *Okx) WSCancelAllSpreadOrders(ctx context.Context, spreadID string) (bool, error) {
+	arg := make(map[string]string, 1)
+	if spreadID != "" {
+		arg["sprdId"] = spreadID
+	}
+
+	id := strconv.FormatInt(ok.Websocket.AuthConn.GenerateMessageID(false), 10)
+
+	var resp []ResponseSuccess
```
gloriousCode, line 325: Rename this silly var. What if it errors?

gloriousCode (changes requested):

exchanges/okx/ws_requests.go
shazbert: Look if total failure is implemented then partial should as well so if we are adamant around this direction I will do all. I will change things around a little bit:

```
+	switch {
+	case intermediary.Code == 0:
+		return nil
+	case intermediary.Code == 1:
+		return fmt.Errorf("%s: %w %w", intermediary.Message, errInvalidWebsocketRequest, intermediary.Data.Error())
+	case intermediary.Code == 2:
+	}
+	return getStatusError(intermediary.Code, intermediary.Message)
+}
```

For case 1 and 2 there is no intermediary message available its just a code.

gloriousCode: Yeah I definitely didn't go deep into implementation for the draft there. The main thing I wanted was to be able to extract an error from within SendMessage function without going crazy.
Thanks for considering my changes 🚀

gbjk: Consolidating my feedback onto this thread as well.
I agree with Scott's direction. It's much clevererer than what I would have done.
Only things I think I want to add here is the concept:
Code 1 and Code 2 should be translated by us into error vars locally. We know what they are and mean.
We should then wrap the errors out of the data set, and the way Scott has done that is good.

I'm kinda amazed you can Unmarshal into an non eface interface like that and have it DTRT, but whoop if it works 🎉  
If it didn't work, I was expecting we had to see if we cast result into an `batchWithErrors` interface.
Which leads to the absolutely wicked observation 👿 ... `WsDataResponse` is actually the same interface as `error`, right .. soo ...
```
func (ok *Okx) SendAuthWSReq(..., payload any, result error) error {
```
I'm so not suggesting that... but just thought it was funny enough to worth commenting on.

So, I'd suggest we add to getStatusError that if code is 1 or 2, we return our local messages for them:
```
func getStatusError(statusCode int64, statusMessage string) error {
	switch statusCode {
        case 0:
		return nil
        case 1:
                 statusMessage = errOperationFailed // "Operation failed"
        case 2:
                 statusMessage = errPartialSuccess // "Bulk operation partially succeeded"
	}
	return fmt.Errorf("status code: `%d` status message: `%s`", statusCode, statusMessage)
}
```
And then we can say in `SendHTTPRequestViaHomingWeasels`:
```
errs := getStatusError(intermediary.Code, intermediary.Message)
if errs == nil {
    return errs
}
// Any batch errors should/must have set status to 1, so err is not nil, and here we are
if err := result.Error(); err != nil {err = common.AppendError(err,

File: exchanges/okx/ws_requests.go [hunk 7]
```diff
@@ -383,11 +325,40 @@ func (ok *Okx) SendAuthenticatedWebsocketRequest(ctx context.Context, epl reques
...
+		return errOperationFailed
+	case 2:
+		return errPartialSuccess
+	default:
 		return getStatusError(intermediary.Code, intermediary.Message)
 	}
+}
 
-	return nil
+func mergeErrorDetails[T interface{ Error() error }](err error, resp []T) error {
```
gbjk, line 340: 🐒 
* Firstly, this is super cute, nice work 💪 
* _cough_ 🙁  I think maybe it's not necessary
Because you're saying that T is something that implments a new interface which is actuallyt identical to the built in `error` [interface itself](https://go.dev/ref/spec#Errors). Inception time, right... T is an interface that returns an interface that matches itself.
* And I don't think we're really merging errors; More like collecting them. Specifically, we're not putting errors inside each other. And if we were, we'd probably use the idiomatic wrap terminology.
So all told, I'd like, but not insist upon, something that dances like:
```suggestion
func collectErrors(errs error) error {
...
```
And then use common.AppendErrors above and it's nil ignoring to shortcut this.
But then ...

gbjk: Okay, so ... if everything is an error implementing thing... isn't this just the same as calling:
```
common.AppendError(errs, err)
```
over range.

So ... I'm thinking that:
```
common.CollectErrors(err, resps...)
```
could be a thing. Naming clashes with the existing method on ErrorCollector, but this is about turning a slice of errors into a multiError via AppendErrors, so I think it's okay.

</ stream of consciousness >

gbjk (approved):
Good work 🎉 

Whilst I think the implementation could be simpler, this works, so I'm approving.

I'd like you to sniff removing the generics and even entire idea of doing any of this.

Happy to prototype and provide a diff if you want 🙂

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,364 @@
...
+	if errs == nil {
+		return err
+	}
+	if err == nil {
+		return errs
+	}
+	return fmt.Errorf("%w: %w", err, errs)
+}
+
+func extractSingleItem[T any](resp []*T) (*T, error) {
```
gbjk, line 356: 🐒 Sorry I missed this before.
* Slightly strange naming with `extract` in this. `get` is better. I'm wary that docs recently recommended not having `get`, though
* `Element` over `Item` in Go docs too.
* sometimes we see `must` prefix in naming like this
* 🥃 In this context I'm fond of the adjective `Singleton`
So overall this is what feels most idiomatic in naming:
```suggestion
func singleElement[T any](resp []*T) (*T, error) {
```

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,364 @@
...
+	arg := make(map[string]string, 1)
+	if spreadID != "" {
+		arg["sprdId"] = spreadID
+	}
+
+	id := strconv.FormatInt(ok.Websocket.AuthConn.GenerateMessageID(false), 10)
+
+	var resp []*ResponseResult
+	if err := ok.SendAuthenticatedWebsocketRequest(ctx, cancelAllSpreadOrderEPL, id, "sprd-mass-cancel", []map[string]string{arg}, &resp); err != nil {
+		return mergeErrorDetails(err, resp)
```
gbjk, line 273: 🚧 sprd-mass-cancel can't return errors to merge
If mergeErrorDetalis was just happening to all of them then I think this would be okay

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,364 @@
...
+// WSCancelAllSpreadOrders cancels all spread orders and return success message through the websocket channel.
+func (ok *Okx) WSCancelAllSpreadOrders(ctx context.Context, spreadID string) error {
+	arg := make(map[string]string, 1)
+	if spreadID != "" {
+		arg["sprdId"] = spreadID
+	}
+
+	id := strconv.FormatInt(ok.Websocket.AuthConn.GenerateMessageID(false), 10)
+
+	var resp []*ResponseResult
```
gbjk, line 271:
```suggestion
	var resps []*ResponseResult
```

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,364 @@
...
+	if err := ok.SendAuthenticatedWebsocketRequest(ctx, cancelAllSpreadOrderEPL, id, "sprd-mass-cancel", []map[string]string{arg}, &resp); err != nil {
+		return mergeErrorDetails(err, resp)
+	}
+
+	single, err := extractSingleItem(resp)
```
gbjk, line 276:
```suggestion
	resp, err := extractSingleItem(resp)
```

gbjk (changes requested):
Couple of things I noticed whilst looking at refactoring mergeErrorDetails

gbjk (changes requested):
That's great work, thanks 🙏 

Just a couple of minor change requests.

Here's a prototype of removing the need for mergeErrorDetails: gbjk@bc6c504fae
```
curl https://gist.githubusercontent.com/gbjk/699b32296c40d0c22f4b70fda5b792ba/raw/6b0222dc7c98f4fbacd05eb70ac8621ed3a05106/changes.patch | git apply
```

shazbert:
> That's great work, thanks 🙏
> 
> Just a couple of minor change requests.
> 
> Here's a prototype of removing the need for mergeErrorDetails: [gbjk/gocryptotrader@bc6c504fae](https://github.com/gbjk/gocryptotrader/commit/bc6c504fae)
> 
> ```
> curl https://gist.githubusercontent.com/gbjk/699b32296c40d0c22f4b70fda5b792ba/raw/6b0222dc7c98f4fbacd05eb70ac8621ed3a05106/changes.patch | git apply
> ```

Looks great on paper, I checked it out and tried to implement, I cant seem to get it to work. One thing I have noticed is that result is a pointer, if you have tested out and the returns are coming back, would prefer this, I must have done something wrong here.
```
if errGroup, ok := result.([]hasError); ok {
		for i := range errGroup {
			if subErr := errGroup[i].Error(); err != nil {
				err = common.AppendError(err, fmt.Errorf("%T[%d]: %w", errGroup[i], i+1, subErr))
			}
 		}
 	}
```

exchanges/okx/okx_types.go
```diff
@@ -768,46 +756,93 @@ type CurrencyTakerFlow struct {
...
+	TargetCurrency          string `json:"tgtCcy,omitempty"`  // values base_ccy and quote_ccy for spot market orders
+	SelfTradePreventionMode string `json:"stpMode,omitempty"` // Default to cancel maker, `cancel_maker`,`cancel_taker`, `cancel_both``
 	// Added in the websocket requests
-	BanAmend   bool       `json:"banAmend,omitempty"` // Whether the SPOT Market Order size can be amended by the system.
-	ExpiryTime types.Time `json:"expTime,omitzero"`
+	BanAmend bool `json:"banAmend,omitempty"` // Whether the SPOT Market Order size can be amended by the system.
+}
+
+// Validate validates the PlaceOrderRequestParam
+func (arg *PlaceOrderRequestParam) Validate() error {
```
gloriousCode, line 787: I feel like this can be moved to okx.go or okx_wrapper instead of important code being inside _types. But I know we do this around the place, so you can click resolve if you want

gloriousCode (changes requested):
I have checked out the rest of the PR instead of just zeroing in on the error handling. The rest of it looks good!

I'll wait on gk's/your comments to be resolved (_even though both of you are leaving unreplyable & unresolvable comments!_) before I take another look at the ws request/response handling, but I thank you for taking both mine and gk's feedback on the err handling

exchanges/okx/okx.go
```diff
@@ -5902,13 +5896,16 @@ func (ok *Okx) SendHTTPRequest(ctx context.Context, ep exchange.URL, f request.E
 		return err
 	}
 	if err == nil && resp.Code.Int64() != 0 {
+		if authenticated == request.AuthenticatedRequest {
+			err = request.ErrAuthRequestFailed
+		}
 		if resp.Msg != "" {
-			return fmt.Errorf("%w error code: %d message: %s", request.ErrAuthRequestFailed, resp.Code.Int64(), resp.Msg)
+			return common.AppendError(err, fmt.Errorf("error code: `%d`; message: `%s`", resp.Code.Int64(), resp.Msg))
```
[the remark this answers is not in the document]
shazbert: [hahaha](https://github.com/thrasher-corp/gocryptotrader/pull/1851#discussion_r2029712041) 🤣

exchanges/okx/okx_types.go
shazbert: Will leave this unresolved for other people to vote; I don't mind either way

exchanges/okx/ws_requests.go
shazbert: Nice spotting!

gbjk:
> Looks great on paper, I checked it out and tried to implement, I cant seem to get it to work

Nope, you're right, I lost the change I actually needed. Sorry about that!

In order to make it dance correctly, I can't seem to avoid using reflection.
We have to get down to the actual interface before we can cast.

Fixed https://github.com/gbjk/gocryptotrader/commit/7649892845

```
    ws_requests_test.go:34:
                Error Trace:    /Users/gbjk/go/src/github.com/gbjk/gocryptotrader/exchanges/okx/ws_requests_test.go:34
                                operation failed, ptr[1]: status code: `51008` status message: `Order failed. Insufficient USDT balance in account.`
                Test:           TestWSPlaceOrder
```

exchanges/okx/ws_requests_test.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/thrasher-corp/gocryptotrader/common"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/request"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/sharedtestvalues"
+)
+
+func TestWSPlaceOrder(t *testing.T) {
+	t.Parallel()
+
+	_, err := ok.WSPlaceOrder(t.Context(), nil)
```
gloriousCode, line 18: ```Error connecting auth socket: Okx websocket timeout waiting for response with signature login-response```
I will assume this is temporary for now, but please check that these tests still function

exchanges/okx/ws_requests_test.go
```diff
@@ -0,0 +1,258 @@
...
+	t.Parallel()
+
+	require.NoError(t, mergeErrors(nil, []interface{ Error() error }(nil)))
+	sillyErr := errors.New("silly error")
+	require.ErrorIs(t, mergeErrors(sillyErr, []interface{ Error() error }(nil)), sillyErr)
+	require.ErrorIs(t, mergeErrors(nil, []*SillyElement{{}}), errElement)
+	require.ErrorIs(t, mergeErrors(sillyErr, []*SillyElement{{}}), errElement)
+	require.ErrorIs(t, mergeErrors(sillyErr, []*SillyElement{{}}), sillyErr)
+}
+
```
gloriousCode, line 246: I am happy with gk's changes. I do have a test I'd like to add though. It is AI generated, so could be improved:
```go
func TestParseErrors(t *testing.T) {
	t.Parallel()

	// Case 1: `result` is not a pointer or not a slice
	err := parseErrors(123, errors.New("initial error"))
	assert.EqualError(t, err, "initial error")

	// Case 2: `result` is a pointer to a slice, but the slice is empty
	var emptySlice []*mockHasError
	err = parseErrors(&emptySlice, errors.New("initial error"))
	assert.EqualError(t, err, "initial error")

	// Case 3: `result` is a pointer to a slice with elements implementing `hasError`
	mockSlice := []*mockHasError{
		{err: nil},                   // No error
		{err: errors.New("error 1")}, // Error in the second element
		{err: errors.New("error 2")}, // Error in the third element
	}
	err = parseErrors(&mockSlice, errors.New("initial error"))
	assert.Error(t, err)
	assert.Contains(t, err.Error(), "initial error")
	assert.Contains(t, err.Error(), "mockHasError[2]: error 1")
	assert.Contains(t, err.Error(), "mockHasError[3]: error 2")
}
```

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,357 @@
...
+
+	switch intermediary.Code {
+	case 0:
+		return nil
+	case 1:
+		return errOperationFailed
+	case 2:
+		return errPartialSuccess
+	default:
+		return getStatusError(intermediary.Code, intermediary.Message)
```
gloriousCode, line 335: I think the main thing that I'd like to change from gk's request is the following:
```go

func parseErrors(result any, err error) error {
	rv := reflect.ValueOf(result)
	if rv.Kind() == reflect.Pointer && rv.Elem().Kind() == reflect.Slice {
		s := rv.Elem()
		for i := range s.Len() {
			v := s.Index(i)
			if v.Kind() == reflect.Interface {
				v = v.Elem()

			}
			if subErr, ok := v.Interface().(hasError); ok && subErr.Error() != nil {
				err = common.AppendError(err, fmt.Errorf("%s[%d]: %w", v.Type(), i+1, subErr.Error()))
			}
		}
	}
	return err
}
```
extracting it into its own function allows testing of the logic there instead of requiring an auth connection and request. See test comment

gloriousCode (changes requested):
I'm happy with the work you and GK have done! I'll do another review once you've merged his changes into this PR. I expect to be able to merge this when you are free enough to update things

exchanges/okx/ws_requests_test.go
shazbert: That shouldn't be happening, I am getting `Okx websocket timeout waiting for response with signature 1` because my key 💩 itself. I can't seem to replicate this issue just yet.

exchanges/okx/okx.go
```diff
@@ -30,15 +30,7 @@ import (
...
-	WsResponseMultiplexer wsRequestDataChannelsMultiplexer
-
-	// WsRequestSemaphore channel is used to block write operation on the websocket connection to reduce contention; a kind of bounded parallelism.
-	// it is made to hold up to 20 integers so that up to 20 write operations can be called over the websocket connection at a time.
-	// and when the operation is completed the thread releases (consumes) one value from the channel so that the other waiting operation can enter.
-	// ok.WsRequestSemaphore <- 1
-	// defer func() { <-ok.WsRequestSemaphore }()
-	WsRequestSemaphore chan int
-
+	counter                common.Counter
```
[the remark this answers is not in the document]
shazbert: [3b3da3a](https://github.com/thrasher-corp/gocryptotrader/pull/1851/commits/3b3da3a8b291c665cf26e2139e13f133e04486fa)

exchanges/okx/ws_requests_test.go
shazbert: So its due to a code not in the list, it just doesn't get handled:
```
{"event":"error","msg":"Your IP is dumb and  is not in linking trusted IP addresses.","code":"50110","connId":"889ef657"}
```

gbjk (approved):
LGTM for current changes 👍 

Happy to review again if there's more to come

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,344 @@
...
+
+	var resp []*OrderData
+	if err := ok.SendAuthenticatedWebsocketRequest(ctx, placeOrderEPL, id, "order", []PlaceOrderRequestParam{*arg}, &resp); err != nil {
+		return nil, err
+	}
+	return singleItem(resp)
+}
+
+// WSPlaceMultipleOrder submits multiple orders
+func (ok *Okx) WSPlaceMultipleOrder(ctx context.Context, args []PlaceOrderRequestParam) ([]*OrderData, error) {
```
thrasher-, line 41:
```suggestion
func (ok *Okx) WSPlaceMultipleOrders(ctx context.Context, args []PlaceOrderRequestParam) ([]*OrderData, error) {
```

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,344 @@
...
+	var resp []*OrderData
+	if err := ok.SendAuthenticatedWebsocketRequest(ctx, cancelOrderEPL, id, "cancel-order", []CancelOrderRequestParam{*arg}, &resp); err != nil {
+		return nil, err
+	}
+
+	return singleItem(resp)
+}
+
+// WSCancelMultipleOrder cancels multiple orders
+func (ok *Okx) WSCancelMultipleOrder(ctx context.Context, args []CancelOrderRequestParam) ([]*OrderData, error) {
```
thrasher-, line 81:
```suggestion
func (ok *Okx) WSCancelMultipleOrders(ctx context.Context, args []CancelOrderRequestParam) ([]*OrderData, error) {
```

exchanges/okx/ws_requests.go
```diff
@@ -0,0 +1,344 @@
...
+	var resp []*SpreadOrderResponse
+	if err := ok.SendAuthenticatedWebsocketRequest(ctx, amendSpreadOrderEPL, id, "sprd-amend-order", []AmendSpreadOrderParam{*arg}, &resp); err != nil {
+		return nil, err
+	}
+
+	return singleItem(resp)
+}
+
+// WsCancelSpreadOrder cancels an incomplete spread order through the websocket connection.
+func (ok *Okx) WsCancelSpreadOrder(ctx context.Context, orderID, clientOrderID string) (*SpreadOrderResponse, error) {
```
thrasher-, line 224:
```suggestion
func (ok *Okx) WSCancelSpreadOrder(ctx context.Context, orderID, clientOrderID string) (*SpreadOrderResponse, error) {
```
To be consistent with our naming scheme

exchanges/okx/okx.go
```diff
@@ -5905,13 +5900,16 @@ func (ok *Okx) SendHTTPRequest(ctx context.Context, ep exchange.URL, f request.E
 		return err
 	}
 	if err == nil && resp.Code.Int64() != 0 {
+		if authenticated == request.AuthenticatedRequest {
+			err = request.ErrAuthRequestFailed
+		}
 		if resp.Msg != "" {
-			return fmt.Errorf("%w error code: %d message: %s", request.ErrAuthRequestFailed, resp.Code.Int64(), resp.Msg)
+			return common.AppendError(err, fmt.Errorf("error code: `%d`; message: `%s`", resp.Code.Int64(), resp.Msg))
```
thrasher-, line 5907:
```suggestion
			return common.AppendError(err, fmt.Errorf("error code: `%d`; message: %q", resp.Code.Int64(), resp.Msg))
```

exchanges/okx/okx.go
```diff
@@ -5924,3 +5922,10 @@ func (ok *Okx) SendHTTPRequest(ctx context.Context, ep exchange.URL, f request.E
 
 	return nil
 }
+
+func getStatusError(statusCode int64, statusMessage string) error {
+	if statusCode == 0 {
+		return nil
+	}
+	return fmt.Errorf("status code: `%d` status message: `%s`", statusCode, statusMessage)
```
thrasher-, line 5930:
```suggestion
	return fmt.Errorf("status code: `%d` status message: %q", statusCode, statusMessage)
```

exchanges/okx/okx_wrapper.go
```diff
@@ -1278,8 +1252,8 @@ func (ok *Okx) CancelOrder(ctx context.Context, ord *order.Cancel) error {
 		if err != nil {
 			return err
 		}
-		if response.StatusCode != "0" {
-			return fmt.Errorf("sCode: %s sMessage: %s", response.StatusCode, response.StatusMessage)
+		if response.StatusCode != 0 {
```
thrasher-, line 1255:
```suggestion
		return getStatusError(response.StatusCode, response.StatusMessage)
```

Since the func does this check anyway and will make a neglible impact in terms of perf since the compiler will inline getStatusError 

image

thrasher- (changes requested):
Websocket appears to be stable, test coverage looks good, great work with this shedding! Only minor issues

exchanges/okx/okx_wrapper.go
shazbert: 🔥 [1333b42](https://github.com/thrasher-corp/gocryptotrader/pull/1851/commits/1333b42b7c47a5e8e03dc1ae59f4f28c967d6d9a)

thrasher- (approved):

PR #1423, opened 2023-12-15
Bitstamp/Kraken: Enhance test coverage

Beadko:
# PR Description

Fix Bitstamp tests that are failing, improve the test coverage

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Bitstamp_tests

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions/AppVeyor with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1860, opened 2025-03-25
Execution limits: move to service

gloriousCode:
# PR Description
This looks like a big PR, but its not really. This is a change being brought over from private work. I thought perhaps it could be helpful for others, but happy to keep it private it people are unhappy with it.

- This PR moves execution limits from something held by an exchange, into a storage service
- Moves all order limits to `internal/order/limits`, also obviously in a bid to appeal to gk
- It introduces a `key` package helper to minimise code lines
- It improves test coverage of `limits` package

### Use case
As a well-regarded trader, using multiple instances of an exchange, I wish to share the order execution limits between them.

### Potential future(s) plans
Across many exchanges, currency details such as expiry, contract decimals are sent in the same response as limits. It would be useful to store these details along with these limits. I currently store them seperately, but `UpdateTradablePairs`, `UpdateOrderExecutionLimits` and `GetFuturesContractDetails` are all the same endpoint across many exchanges, but could still push changes down to the one set of local storage.

## Type of change

## How has this been tested
- Improved coverage of limits
- Golangci-lint pass

## Checklist


exchanges/gateio/gateio_wrapper.go
```diff
@@ -2172,21 +2173,110 @@ func (g *Gateio) UpdateOrderExecutionLimits(ctx context.Context, a asset.Item) e
...
+		usdtContracts, err := g.GetAllFutureContracts(ctx, currency.USDT)
+		if err != nil {
+			return err
+		}
+		btcContracts = append(btcContracts, usdtContracts...)
+		l = make([]limits.MinMaxLevel, 0, len(btcContracts))
+		for x := range btcContracts {
+			if btcContracts[x].InDelisting {
+				continue
+			}
```
shazbert, line 2198: Probably don't need this and below, if its loaded in memory anyway its going to persist. If you have long or short contract open on startup you can at least reduce only it with the correct trading details.

exchanges/gateio/gateio_wrapper.go
```diff
@@ -2172,21 +2173,110 @@ func (g *Gateio) UpdateOrderExecutionLimits(ctx context.Context, a asset.Item) e
...
+			p := strings.ToUpper(btcContracts[x].Name)
+			if !g.IsValidPairString(p) {
+				continue
+			}
```
shazbert, line 2202: Can you please have a look at this function, it seems like it can be removed completely?

exchanges/gateio/gateio_wrapper.go
```diff
@@ -2172,21 +2173,110 @@ func (g *Gateio) UpdateOrderExecutionLimits(ctx context.Context, a asset.Item) e
...
+			cp, err := currency.NewPairFromString(p)
+			if err != nil {
+				return err
+			}
+			l = append(l, limits.MinMaxLevel{
+				Key:                     key.NewExchangePairAssetKey(g.Name, a, cp),
+				MinimumBaseAmount:       float64(btcContracts[x].OrderSizeMin),
+				MaximumBaseAmount:       float64(btcContracts[x].OrderSizeMax),
+				MarketStepIncrementSize: btcContracts[x].OrderPriceRound.Float64(),
+				AmountStepIncrementSize: btcContracts[x].OrderPriceRound.Float64(),
```
shazbert, line 2212: This looks like it should be `PriceStepIncrementSize` and below

internal/order/limits/storage.go
```diff
@@ -0,0 +1,124 @@
...
+	return nil
+}
+
+func (e *executionLimits) getOrderExecutionLimits(k key.ExchangePairAsset) (MinMaxLevel, error) {
+	e.mtx.RLock()
+	defer e.mtx.RUnlock()
+	if e.epaLimits == nil {
+		return MinMaxLevel{}, ErrExchangeLimitNotLoaded
+	}
+	k.Exchange = strings.ToLower(k.Exchange)
```
shazbert, line 96: I don't think this is necessary and below, we should have access to exchanges method for name.

shazbert (changes requested):
Not bad, just some things I noticed.

exchanges/gateio/gateio_wrapper.go
gloriousCode: I previously had them set to 1 (as in 1 contract) but thought I'd try and link it to something better in anticipation of your comments 😭

internal/order/limits/storage.go
gloriousCode: I achieved too many mistakes without it. I will remove it for you and add it back for me

exchanges/gateio/gateio_wrapper.go
shazbert: Yeah I set them as 1 in my own stuff and the step as 1 if its for contracts. Can't do this for every exchange though, because some (bybit I think?) uses the actual base amounts instead of contract and multi. If you do go down that route, maybe have a futures package const ContractStep = 1 maybe as an idea.

common/key/key.go
```diff
@@ -76,3 +58,31 @@ func (k *ExchangePairAsset) MatchesExchange(exch string) bool {
...
+// SubAccountCurrencyAsset is a unique map key signature for subaccount, currency code and asset
+type SubAccountCurrencyAsset struct {
+	SubAccount string
+	Currency   *currency.Item
+	Asset      asset.Item
+}
+
+// Pair combines the base and quote into a pair
+func (k *PairAsset) Pair() currency.Pair {
+	if k == nil || (k.Base == nil && k.Quote == nil) {
```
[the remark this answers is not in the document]
gloriousCode: No thank you. Though its not a huge deal, just may as well avoid mixing pointer and value receivers as per go guidelines (even though we do it all over the place and I think its silly). Please think of the underlines in my IDE.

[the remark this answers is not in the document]
gloriousCode: Or just maybe, I'll make the rest value receivers 
# 🤯

exchanges/gateio/gateio_wrapper.go
gloriousCode: Do you find that the order execution limits change when the currency is changed to delisting?

shazbert: No, I haven't experienced that when it goes into delisting for the old g i o

gloriousCode: I still rm'd for you 🌈

common/key/key.go
```diff
@@ -26,36 +26,27 @@ func NewExchangePairAssetKey(exch string, a asset.Item, cp currency.Pair) Exchan
 }
 
 // Pair combines the base and quote into a pair
-func (k *ExchangePairAsset) Pair() currency.Pair {
-	if k == nil || (k.Base == nil && k.Quote == nil) {
+func (k ExchangePairAsset) Pair() currency.Pair {
+	if k.Base == nil && k.Quote == nil {
 		return currency.EMPTYPAIR
 	}
```
shazbert, line 32: You can RM this

common/key/key.go
```diff
@@ -72,17 +63,16 @@ type PairAsset struct {
...
-	Asset      asset.Item
-}
-
 // Pair combines the base and quote into a pair
-func (k *PairAsset) Pair() currency.Pair {
-	if k == nil || (k.Base == nil && k.Quote == nil) {
+func (k PairAsset) Pair() currency.Pair {
+	if k.Base == nil && k.Quote == nil {
 		return currency.EMPTYPAIR
 	}
```
shazbert, line 70: You can RM this as well

exchanges/gateio/gateio_test.go
```diff
@@ -3155,45 +3154,21 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
-		if mm.QuoteStepIncrementSize <= 0 {
-			t.Fatalf("QuoteStepIncrementSize expected 0 but received %v for %v", mm.QuoteStepIncrementSize, avail[i])
-		}
-
-		if mm.AmountStepIncrementSize <= 0 {
-			t.Fatalf("AmountStepIncrementSize expected 0 but received %v for %v", mm.AmountStepIncrementSize, avail[i])
-		}
+		require.NoError(t, err, "GetOrderExecutionLimits must not error")
+		require.NotEmpty(t, mm, "GetOrderExecutionLimits must not return empty value")
+		assert.Positive(t, mm.MinimumBaseAmount)
```
shazbert, line 3169: I love this positivity 🌈

exchanges/gateio/gateio.go
```diff
@@ -3606,22 +3606,6 @@ func (g *Gateio) InitiateFlashSwapOrderReview(ctx context.Context, arg FlashSwap
 	return response, g.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, flashOrderReviewEPL, http.MethodPost, gateioFlashSwapOrdersPreview, nil, &arg, &response)
 }
 
-// IsValidPairString returns true if the string represents a valid currency pair
-func (g *Gateio) IsValidPairString(currencyPair string) bool {
```
shazbert, line 3610: cya mate

shazbert (approved):
Approved, only some suggestions that won't hold up anything.

common/key/key.go
```diff
@@ -15,63 +15,58 @@ type ExchangePairAsset struct {
...
-}
-
-// Pair combines the base and quote into a pair
-func (k *PairAsset) Pair() currency.Pair {
-	if k == nil || (k.Base == nil && k.Quote == nil) {
-		return currency.EMPTYPAIR
+// NewExchangePairAssetKey is a helper to reduce the amount of code needed to create a key
+func NewExchangePairAssetKey(exch string, a asset.Item, cp currency.Pair) ExchangePairAsset {
+	return ExchangePairAsset{
+		Exchange: strings.ToLower(exch),
```
shazbert, line 21: suggestion: I would leave this out and just pass it directly, because this adds in overhead and can be adjusted by the caller.

shazbert (approved):
tACK with one suggestion 🥇

common/key/key.go
gloriousCode: I meat to get rid of it everywhere for you sorry, will rm

shazbert (approved):
thanks m8 changes ACK.

samuael (commented):
This is simple and cool PR, man. I couldn't issues so I have left some review comment on minor nits that I felt should be changed.

exchange/order/limits/storage.go
```diff
@@ -0,0 +1,118 @@
...
+		if !levels[x].Key.Asset.IsValid() {
+			return fmt.Errorf("cannot load levels for '%s' '%s': %w", levels[x].Key.Exchange, levels[x].Key.Pair(), errAssetInvalid)
+		}
+		if levels[x].Key.Pair().IsEmpty() {
+			return fmt.Errorf("cannot load levels for '%s' '%s': %w", levels[x].Key.Exchange, levels[x].Key.Asset, errPairNotSet)
+		}
+		if levels[x].MinPrice > 0 &&
+			levels[x].MaxPrice > 0 &&
+			levels[x].MinPrice > levels[x].MaxPrice {
+			return fmt.Errorf("%w for %s %s %s supplied min: %f max: %f",
```
thrasher-, line 50: Just noticed that we're not being consistent with quote usage like the other nits where we do use quotes for here and the below errors

exchanges/binance/binance_test.go
```diff
@@ -2517,46 +2518,28 @@ func TestUFuturesHistoricalTrades(t *testing.T) {
...
 	err = limit.Conforms(0.000001, 0.1, order.Limit)
-	if !errors.Is(err, order.ErrAmountBelowMin) {
-		t.Fatalf("expected %v, but received %v", order.ErrAmountBelowMin, err)
-	}
+	require.ErrorIs(t, err, limits.ErrAmountBelowMin, "expected amount to be below min")
 
 	err = limit.Conforms(0.01, 1, order.Limit)
-	if !errors.Is(err, order.ErrPriceBelowMin) {
-		t.Fatalf("expected %v, but received %v", order.ErrPriceBelowMin, err)
+	if !errors.Is(err, limits.ErrPriceBelowMin) {
```
thrasher-, line 2541:
```suggestion
	require.ErrorIs(t, err, limits.ErrPriceBelowMin)
```

exchanges/binance/binance_test.go
```diff
@@ -2761,11 +2744,11 @@ func TestFetchExchangeLimits(t *testing.T) {
 	t.Parallel()
 	limits, err := b.FetchExchangeLimits(t.Context(), asset.Spot)
```
thrasher-, line 2745:
```suggestion
	l, err := b.FetchExchangeLimits(t.Context(), asset.Spot)
```
Looks like you meant this here, since it's used below and the build is currently broken

exchanges/binance/binance_test.go
```diff
@@ -2761,11 +2744,11 @@ func TestFetchExchangeLimits(t *testing.T) {
...
 	assert.NoError(t, err, "FetchExchangeLimits should not error")
-	assert.NotEmpty(t, limits, "Should get some limits back")
+	assert.NotEmpty(t, l, "Should get some limits back")
 
 	limits, err = b.FetchExchangeLimits(t.Context(), asset.Margin)
```
thrasher-, line 2749:
```suggestion
	l, err = b.FetchExchangeLimits(t.Context(), asset.Margin)
```
Same issue with the intended var usage

exchanges/bithumb/bithumb_test.go
```diff
@@ -539,7 +540,7 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 	require.NoError(t, err, "GetOrderExecutionLimits must not error")
 
 	err = limit.Conforms(46241000, 0.00001, order.Limit)
-	assert.ErrorIs(t, err, order.ErrAmountBelowMin, "expected error: %v, but got: %v", order.ErrAmountBelowMin, err)
+	assert.ErrorIs(t, err, limits.ErrAmountBelowMin, "expected error: %v, but got: %v", limits.ErrAmountBelowMin, err)
```
thrasher-, line 543:
```suggestion
	assert.ErrorIs(t, err, limits.ErrAmountBelowMin)
```
Redundant error decoration since we'll get that output anyway if the test were to fail:

image

exchanges/bybit/bybit_test.go
```diff
@@ -838,8 +835,8 @@ func TestPlaceOrder(t *testing.T) {
 		Side:      "buy",
 		OrderType: "limit",
 	})
-	if !errors.Is(err, order.ErrAmountBelowMin) {
-		t.Fatalf("expected %v, got %v", order.ErrAmountBelowMin, err)
+	if !errors.Is(err, limits.ErrAmountBelowMin) {
```
thrasher-, line 838:
```suggestion
	require.ErrorIs(t, err, limits.ErrAmountBelowMin)
```
And same for the below ones

exchanges/okx/okx_test.go
```diff
@@ -3349,10 +3350,10 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 		}
 
 		for _, p := range tests[a] {
-			limits, err := ok.GetOrderExecutionLimits(a, p)
+			l, err := ok.GetOrderExecutionLimits(a, p)
 			if assert.NoError(t, err, "GetOrderExecutionLimits should not error") {
-				require.Positivef(t, limits.PriceStepIncrementSize, "PriceStepIncrementSize should be positive for %s", p)
-				require.Positivef(t, limits.MinimumBaseAmount, "PriceStepIncrementSize should be positive for %s", p)
+				require.Positivef(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should be positive for %s", p)
+				require.Positivef(t, l.MinimumBaseAmount, "PriceStepIncrementSize should be positive for %s", p)
```
thrasher-, line 3356:
```suggestion
				require.Positivef(t, l.MinimumBaseAmount, "MinimumBaseAmount must be positive for %s", p)
```
Wasn't you, but mismatched var name and should/must violation 👀 Only noticed it due to changed coderinos

exchanges/okx/okx_test.go
```diff
@@ -3349,10 +3350,10 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 		}
 
 		for _, p := range tests[a] {
-			limits, err := ok.GetOrderExecutionLimits(a, p)
+			l, err := ok.GetOrderExecutionLimits(a, p)
 			if assert.NoError(t, err, "GetOrderExecutionLimits should not error") {
-				require.Positivef(t, limits.PriceStepIncrementSize, "PriceStepIncrementSize should be positive for %s", p)
-				require.Positivef(t, limits.MinimumBaseAmount, "PriceStepIncrementSize should be positive for %s", p)
+				require.Positivef(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should be positive for %s", p)
```
thrasher-, line 3355:
```suggestion
				require.Positivef(t, l.PriceStepIncrementSize, "PriceStepIncrementSize must be positive for %s", p)
```

exchanges/gateio/gateio_test.go
```diff
@@ -2533,49 +2533,33 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
-
-	err = g.UpdateOrderExecutionLimits(t.Context(), asset.Options)
-	if !errors.Is(err, common.ErrNotYetImplemented) {
-		t.Fatalf("received %v, expected %v", err, common.ErrNotYetImplemented)
+	scenarios := []struct {
+		assetType     asset.Item
+		expectedError error
+	}{
+		{assetType: asset.Spot, expectedError: nil},
+		{assetType: asset.Futures, expectedError: nil},
```
thrasher-, line 2541: Need to add in checks for `asset.CoinMarginedFutures` and `asset.USDTMarginedFutures` since they've been split from `asset.Futures` now as this test is currently failing with unsupported asset type.

This also highlights that the `UpdateOrderExecutionLimits` still has a switch case check for `asset.Futures` but will never get hit because the very first `g.SupportsAsset(a)` check will fail with the above

thrasher- (changes requested):
Lovely work on this! Some nits/issues for when you return 🚀

common/key/key.go
```diff
@@ -15,63 +13,58 @@ type ExchangePairAsset struct {
...
-type SubAccountAsset struct {
-	SubAccount string
-	Asset      asset.Item
-}
-
-// Pair combines the base and quote into a pair
-func (k *PairAsset) Pair() currency.Pair {
-	if k == nil || (k.Base == nil && k.Quote == nil) {
-		return currency.EMPTYPAIR
+// NewExchangePairAssetKey is a helper to reduce the amount of code needed to create a key
```
gbjk: 🐒 This comment feels like its a justification for the PR change, not a function description 🙂 
Something like:
```suggestion
// NewExchangePairAssetKey is a helper function to expand a Pair into an ExchangePairAsset
```

common/key/key.go
```diff
@@ -15,63 +13,58 @@ type ExchangePairAsset struct {
...
+func NewExchangePairAssetKey(exch string, a asset.Item, cp currency.Pair) ExchangePairAsset {
```
gbjk: 🤷‍♂️ 🦍 If we return a pointer, then all the methods below could be pointer methods.
```suggestion
func NewExchangePairAssetKey(exch string, a asset.Item, cp currency.Pair) *ExchangePairAsset {
```
I think that'd reduce memory usage everywhere, avoiding pass-by-value on keys for all these calls, right?

common/key/key.go
```diff
@@ -15,63 +13,58 @@ type ExchangePairAsset struct {
...
+	return ExchangePairAsset{
+		Exchange: exch,
```
gbjk, line 19: 🤷‍♂️ In recent work we've been discussing using pointers over strings here.
I think possibly import cycle avoidance played a role in some design choices to use strings in other places.
If we use instances, it'll reduce memory, separate out exchange instances for concurrent use, and it's just saner IMO.
For this, I'd use an `exchange` interface which does `GetName() string` and just that, since it satisfies identification, and any error handling, but can recast easily.
A pattern I've used for avoiding import cycles when doing this a lot is to have a bare interfaces package, such that `gct/interfaces` has `interfaces.Exchange` and that interface is native types and interfaces defined in that pkg only, and may not include other types, avoiding import cycles.

TL:DR: Can be an interface please?

common/key/key.go
```diff
@@ -15,63 +13,58 @@ type ExchangePairAsset struct {
 	Asset    asset.Item
```
gbjk: 🤷‍♂️ Historical question, but Why don't we use an actual Pair here?
We're now using a `New` method only anyway.
We could probably make the internal values private and expose only functions.

🤔 💭  Makes me consider: If we have all values as pointers, then the method for `Match` can just be one method.
A nil value would mean "I don't care" and an empty value would mean "must be empty".
Having done that, Key could be an interface of `Matches()` and all keys become the same.
Not suggesting we do this; I don't yet see a benefit except that it'd be uniform, and cute.

common/key/key_test.go
```diff
@@ -100,12 +71,29 @@ func TestPairAsset_Pair(t *testing.T) {
...
-	assert.Equal(t, cp, pa.Pair())
+	// demonstration of rules that it is simply a helper and not a validator
+	e = ""
+	a = 0
+	p = currency.EMPTYPAIR
+	k = NewExchangePairAssetKey(e, a, p)
+	assert.Equal(t, e, k.Exchange)
+	assert.Equal(t, p.Base.Item, k.Base)
+	assert.Equal(t, p.Quote.Item, k.Quote)
+	assert.Equal(t, a, k.Asset)
```
gbjk: 🐒 Since it can't return an error, and it can't return nil (currently), I don't see how it *could* be a validator, ot what other values those fields could have than the ones you've passed.
That is: I don't think this test is helpful.
🐒 Where we have a comment before some tests, most often it's better to become the test message.
```
	assert.Equal(t, a, k.Asset, "NewExchangePairAssetKey should not alter an invalid asset")
```

exchanges/okx/okx_test.go
gbjk: 🤚 Pretty sure that's my fault

gbjk: 🚧 Instead of switching the require/assert, and using `assquire.*f` can we use `t.Run` please?
```suggestion
			t.Run(a.String()+"/"+p.String(), func(t *testing.T){
				l, err := ok.GetOrderExecutionLimits(a, p)
				require.NoError(t, err, "GetOrderExecutionLimits must not error")
				assert.Positive(t, limits.PriceStepIncrementSize, "PriceStepIncrementSize should be positive")
				assert.Positivef(t, limits.MinimumBaseAmount, "PriceStepIncrementSize should be positive")
				assert.Positivef(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should be positive")
			})
```

exchanges/gateio/gateio_wrapper.go
```diff
@@ -1947,21 +1933,95 @@ func (g *Gateio) UpdateOrderExecutionLimits(ctx context.Context, a asset.Item) e
...
+					MaximumBaseAmount:       float64(contracts[c].OrderSizeMax),
+					PriceStepIncrementSize:  contracts[c].OrderPriceRound.Float64(),
+					AmountStepIncrementSize: 1,
+				})
+			}
+		}
 	default:
-		// TODO: Add in other assets
-		return fmt.Errorf("%s %w", a, common.ErrNotYetImplemented)
+		return fmt.Errorf("%w %v", asset.ErrNotSupported, a)
```
gbjk: 🐒 
```suggestion
		return fmt.Errorf("%w: %q", asset.ErrNotSupported, a)
```

exchanges/gateio/gateio_test.go
```diff
@@ -2533,49 +2533,33 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
+		{assetType: asset.CrossMargin, expectedError: asset.ErrNotSupported},
+		{assetType: asset.Margin, expectedError: asset.ErrNotSupported},
 	}
 
-	err = g.UpdateOrderExecutionLimits(t.Context(), asset.Spot)
-	if err != nil {
-		t.Fatal(err)
+	for _, s := range scenarios {
+		err := g.UpdateOrderExecutionLimits(t.Context(), s.assetType)
+		assert.ErrorIs(t, err, s.expectedError)
```
gbjk: 🐒 
```suggestion
		if s.expectedError == nil {
			assert.NoError(t, err)
		} else {
			assert.ErrorIs(t, err, s.expectedError)
		}
```

exchanges/exchange_test.go
```diff
@@ -475,11 +476,11 @@ func TestSetCurrencyPairFormat(t *testing.T) {
...
 	}
-	futures, err := b.GetPairFormat(asset.Futures, false)
+	f, err := b.GetPairFormat(asset.Futures, false)
 	if err != nil {
 		t.Fatal(err)
 	}
-	if futures.Delimiter != ":)" {
+	if f.Delimiter != ":)" {
 		t.Error("incorrect pair format delimiter")
 	}
```
gbjk: 🚧 
```suggestion
	f, err := b.GetPairFormat(asset.Futures, false)
	require.NoError(t, err, "GetPairFormat must not error")
	assert.Equal(t, ":", f.Delimiter, "GetPairFormat must return a format with correct delimiter")

File: exchanges/exchange_test.go [hunk 28]
```diff
@@ -3145,6 +3141,44 @@ func TestSetConfigPairFormatFromExchange(t *testing.T) {
...
+		Name: "TESTNAME",
+	}
+	cp := currency.NewBTCUSDT()
+	k := key.NewExchangePairAssetKey("TESTNAME", asset.Spread, cp)
+	l := limits.MinMaxLevel{
+		Key:      k,
+		MaxPrice: 1337,
+	}
+	err := limits.LoadLimits([]limits.MinMaxLevel{l})
+	require.NoError(t, err)
```
gbjk: 🐒 
```suggestion
	require.NoError(t, err, "LoadLimits must not error")
```

exchanges/exchange_test.go
```diff
@@ -3145,6 +3141,44 @@ func TestSetConfigPairFormatFromExchange(t *testing.T) {
...
+	cp := currency.NewBTCUSDT()
+	k := key.NewExchangePairAssetKey("TESTNAME", asset.Spread, cp)
+	l := limits.MinMaxLevel{
+		Key:      k,
+		MaxPrice: 1337,
+	}
+	err := limits.LoadLimits([]limits.MinMaxLevel{
+		l,
+	})
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "LoadLimits must not error")
```

exchanges/deribit/deribit_test.go
```diff
@@ -3865,8 +3865,8 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 		}
 	}
 	require.NotNil(t, instrumentDetail, "instrument required to be found")
-	require.Equalf(t, instrumentDetail.TickSize, limits.PriceStepIncrementSize, "Asset: %s Pair: %s Expected: %f Got: %f", asset.Spot, spotTradablePair, instrumentDetail.TickSize, limits.MinimumBaseAmount)
-	assert.Equalf(t, instrumentDetail.MinimumTradeAmount, limits.MinimumBaseAmount, "Pair: %s Expected: %f Got: %f", spotTradablePair, instrumentDetail.MinimumTradeAmount, limits.MinimumBaseAmount)
+	require.Equalf(t, instrumentDetail.TickSize, l.PriceStepIncrementSize, "Asset: %s Pair: %s Expected: %f Got: %f", asset.Spot, spotTradablePair, instrumentDetail.TickSize, l.MinimumBaseAmount)
+	assert.Equalf(t, instrumentDetail.MinimumTradeAmount, l.MinimumBaseAmount, "Pair: %s Expected: %f Got: %f", spotTradablePair, instrumentDetail.MinimumTradeAmount, l.MinimumBaseAmount)
```
gbjk: 🚧 Mixing ~my toasties~ "expected/got" syntax with `assquire.*f`
🐒 Also Seems weird that we have `Equalf` here for effectively constants.
```suggestion
	require.Equal(t, instrumentDetail.TickSize, l.PriceStepIncrementSize, "PriceStepIncrementSize should be correct")
	assert.Equal(t, instrumentDetail.MinimumTradeAmount, l.MinimumBaseAmount, "MinimumBaseAmount should be correct")
```

exchange/order/limits/levels.go
```diff
@@ -0,0 +1,180 @@
...
+
+	"github.com/shopspring/decimal"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
+)
+
+// Conforms checks outbound parameters
+func (m *MinMaxLevel) Conforms(price, amount float64, orderType order.Type) error {
+	// TODO: Update to take in account Quote amounts as well as Base amounts.
+	if m == nil {
+		return nil
```
gbjk: 🐒  I think this naming might be idiosyncratic, prefer `Validate`
🥃 TIL I realised that Validate is correct, but Verify is wrong.
🐒 This monkey also want to improve "outbound parameter" 
🦍 👑  KingKong-Super-nit, but all todos are "update to"
🤷‍♂️ Is a nil-pointer really a non-error condition? If not, `common.NilGuard`
```suggestion
// Validate ensures MinMaxLevel fields are valid
func (m *MinMaxLevel) Validate(price, amount float64, orderType order.Type) error {
	// TODO: Verify Quote as well as Base amounts
	if m == nil {
		return nil
```

exchange/order/limits/levels.go
```diff
@@ -0,0 +1,180 @@
...
+	}
+
+	if m.MinimumBaseAmount != 0 && amount < m.MinimumBaseAmount {
+		return fmt.Errorf("%w min: %.8f supplied %.8f",
+			ErrAmountBelowMin,
+			m.MinimumBaseAmount,
+			amount)
```
gbjk: 🦍 This gorilla likes one-liner errors and logs where possible and sane.
If accepted, Applies to several cases below as well.
```suggestion
		return fmt.Errorf("%w min: %.8f supplied %.8f", ErrAmountBelowMin, m.MinimumBaseAmount, amount)
```

exchange/order/limits/levels.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Max iceberg parts checking not done as we do not have that
+	// functionality yet (TODO)
+	// m.maxIcebergParts // How many components in an iceberg order
+
+	// Max total orders not done due to order manager limitations (TODO)
+	// m.maxTotalOrders
+
+	// Max algo orders not done due to order manager limitations (TODO)
+	// m.maxAlgoOrders
```
gbjk: 🦍 Could we use a `/* */` block here?

exchange/order/limits/levels.go
```diff
@@ -0,0 +1,180 @@
...
+					m.PriceStepIncrementSize,
+					price)
+			}
+		}
+		return nil
+	}
+
+	if m.MarketMinQty != 0 &&
+		m.MinimumBaseAmount < m.MarketMinQty &&
+		amount < m.MarketMinQty {
```
gbjk: 🦍 Found this hard to parse always due to same line indentation as logic. Unlike the gofmfmpt recommendation on bare-lines for `) {` I'm not aware of a similar solution for if statements
If possible, I'd prefer onelining this
If accepted applies below too
```suggestion
	if m.MarketMinQty != 0 && m.MinimumBaseAmount < m.MarketMinQty && amount < m.MarketMinQty {
```

exchange/order/limits/levels.go
```diff
@@ -0,0 +1,180 @@
...
+		return fmt.Errorf("%w min: %.8f supplied %.8f",
```
gbjk: 🦍 Adding `;` and maybe making it more succinct
```suggestion
		return fmt.Errorf("%w; %.8f < %.8f",
```

exchange/order/limits/levels.go
```diff
@@ -0,0 +1,180 @@
...
+				ErrMarketAmountExceedsStep,
+				m.MarketStepIncrementSize,
+				amount)
+		}
+	}
+	return nil
+}
+
+// ConformToDecimalAmount (POC) conforms amount to its amount interval
+func (m *MinMaxLevel) ConformToDecimalAmount(amount decimal.Decimal) decimal.Decimal {
```
gbjk: 🚧   I think Conform is actually `Floor` ( as opposed to `Round` )
🦍 It sucks, but I don't think `ToDecimalAmount` is verbose enough
🤷‍♂️ Is POC .. Proof of Concept?
🔁 Also applies to others below.
```suggestion
// FloorAmountToDecimalStepInterval rounds the amount to the AmountStepIncrementSize
func (m *MinMaxLevel) FloorToDecimalStepInterval(amount decimal.Decimal) decimal.Decimal {
```

exchange/order/limits/levels_test.go
```diff
@@ -0,0 +1,166 @@
...
+	tt = nil
+	err = tt.Conforms(0, 0, order.Limit)
+	assert.NoError(t, err)
+}
+
+func TestConformToDecimalAmount(t *testing.T) {
+	t.Parallel()
+	tt := &MinMaxLevel{}
+	val := tt.ConformToDecimalAmount(decimal.NewFromFloat(1.001))
+	assert.True(t, val.Equal(decimal.NewFromFloat(1.001)))
```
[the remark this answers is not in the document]
gbjk: I see suggestions that you can't use `assert.Equal` on decimals, and the above is the solution to raised issues about it.

However I'd prefer to use `Equal` with `.String()` and just have the value we expect. It looks like that should work.
```suggestion
	assert.Equal(t, "1.001", val.String())
```

exchange/order/limits/limits_types.go
```diff
@@ -0,0 +1,74 @@
...
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/thrasher-corp/gocryptotrader/common/key"
+)
+
+// Public errors for order limits
+var (
+	ErrCannotLoadLimit         = errors.New("cannot load limit, levels not supplied")
```
gbjk: 🚧 error name does really match the error context
Either of (but I prefer the former):
```suggestion
	ErrNoLevels         = errors.New("no levels supplied")
```
or
```suggestion
	ErrNoLevels         = errors.New("cannot load limits, no levels supplied")
```

exchange/order/limits/limits_types.go
```diff
@@ -0,0 +1,74 @@
...
+	ErrOrderLimitNotFound      = errors.New("order limit not found")
+	ErrExchangeLimitNotLoaded  = errors.New("exchange limits not loaded")
+	ErrPriceBelowMin           = errors.New("price below minimum limit")
+	ErrPriceExceedsMax         = errors.New("price exceeds maximum limit")
+	ErrPriceExceedsStep        = errors.New("price exceeds step limit") // price is not divisible by its step
```
gbjk: 🐒   Seems comment is more parsable than the error, so maybe better
And probably rename error too. I'd favour `Mod` since we all know it.
Alternative `HasStepRem`
```suggestion
	ErrPriceHasStepMod        = errors.New("price not multiple of step limit")
```

exchange/order/limits/limits_types.go
```diff
@@ -0,0 +1,74 @@
...
+	ErrAmountBelowMin          = errors.New("amount below minimum limit")
+	ErrAmountExceedsMax        = errors.New("amount exceeds maximum limit")
+	ErrAmountExceedsStep       = errors.New("amount exceeds step limit") // amount is not divisible by its step
```
gbjk: 🐒 As above
```suggestion
	ErrAmountHasStepMod       = errors.New("amount not multiple of step limit")
```

exchange/order/limits/limits_types.go
```diff
@@ -0,0 +1,74 @@
...
+	ErrNotionalValue           = errors.New("total notional value is under minimum limit")
+	ErrMarketAmountBelowMin    = errors.New("market order amount below minimum limit")
+	ErrMarketAmountExceedsMax  = errors.New("market order amount exceeds maximum limit")
+	ErrMarketAmountExceedsStep = errors.New("market order amount exceeds step limit") // amount is not divisible by its step for a market order
```
gbjk: 🐒  As above
🤷‍♂️   Wondering if we need to duplicate errors for `ErrMarket*` or could add `market` into thrown errors
```suggestion
	ErrMarketAmountHasStepMod = errors.New("market order amount not dividable by step")
```

exchange/order/limits/limits_types.go
```diff
@@ -0,0 +1,74 @@
...
+	errAssetInvalid        = errors.New("asset is invalid")
+	errPairNotSet          = errors.New("currency pair is not set")
+	errInvalidPriceLevels  = errors.New("invalid price levels, cannot load limits")
+	errInvalidAmountLevels = errors.New("invalid amount levels, cannot load limits")
+	errInvalidQuoteLevels  = errors.New("invalid quote levels, cannot load limits")
+)
+
+// executionLimits defines minimum and maximum values in relation to
+// order size, order pricing, total notional values, total maximum orders etc
+// for execution on an exchange.
```
gbjk: 🦍 Think we can make this briefer
```suggestion
// executionLimits defines minimum and maximum values for order size, pricing, max orders, etc
```

exchange/order/limits/storage.go
```diff
@@ -0,0 +1,118 @@
+package limits
```
gbjk: 🐒  can storage become store?

exchange/order/limits/limits_types.go
```diff
@@ -0,0 +1,74 @@
...
+type executionLimits struct {
+	epaLimits map[key.ExchangePairAsset]*MinMaxLevel
+	mtx       sync.RWMutex
+}
+
+var executionLimitsManager = executionLimits{
+	epaLimits: make(map[key.ExchangePairAsset]*MinMaxLevel),
+}
```
gbjk: 🐒 In the context of this package, can we just replace with `store` or `manager`?

exchange/order/limits/storage.go
```diff
@@ -0,0 +1,118 @@
...
+	return executionLimitsManager.getOrderExecutionLimits(k)
+}
+
+// CheckOrderExecutionLimits is a convenience method to check if the price and amount conforms
+// to the exchange order limits
+func CheckOrderExecutionLimits(k key.ExchangePairAsset, price, amount float64, orderType order.Type) error {
+	return executionLimitsManager.checkOrderExecutionLimits(k, price, amount, orderType)
+}
+
+func (e *executionLimits) loadLimits(levels []MinMaxLevel) error {
```
gbjk: 🐒 
```suggestion
func (e *executionLimits) load(levels []MinMaxLevel) error {
```

exchanges/binance/binance_test.go
```diff
@@ -2517,46 +2518,28 @@ func TestUFuturesHistoricalTrades(t *testing.T) {
 func TestSetExchangeOrderExecutionLimits(t *testing.T) {
```
gbjk: 🚧 
```suggestion
func TestUpdateOrderExecutionLimits(t *testing.T) {
```

exchanges/binance/binance_test.go
```diff
@@ -2790,28 +2773,28 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 		require.NoError(t, err, "UpdateOrderExecutionLimits should not error")
 
 		p := tests[a]
```
gbjk: 🚧  This feels like a `t.Run` with a.String+"/"+p.String and removing all the f-ing

exchanges/bitfinex/bitfinex_test.go
```diff
@@ -157,18 +157,13 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 		},
 	}
 	for assetItem, pairs := range tests {
-		if err := b.UpdateOrderExecutionLimits(t.Context(), assetItem); err != nil {
-			t.Errorf("Error fetching %s pairs for test: %v", assetItem, err)
+		if assert.NoError(t, b.UpdateOrderExecutionLimits(t.Context(), assetItem), "UpdateOrderExecutionLimits should not error") {
 			continue
 		}
```
gbjk: 🚧 This is a require in disguise, isn't it?
🐒 Also looks like this is a `t.Run`
```suggestion
		require.NoError(t, b.UpdateOrderExecutionLimits(t.Context(), assetItem))
```

exchanges/bitfinex/bitfinex_test.go
```diff
@@ -157,18 +157,13 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
-			limits, err := b.GetOrderExecutionLimits(assetItem, pair)
-			if err != nil {
-				t.Errorf("GetOrderExecutionLimits() error during TestExecutionLimits; Asset: %s Pair: %s Err: %v", assetItem, pair, err)
-				continue
-			}
-			if limits.MinimumBaseAmount == 0 {
-				t.Errorf("UpdateOrderExecutionLimits empty minimum base amount; Pair: %s Expected Limit: %v", pair, limits.MinimumBaseAmount)
+			l, err := b.GetOrderExecutionLimits(assetItem, pair)
+			if !assert.Error(t, err, "GetOrderExecutionLimits should not error") {
+				assert.NotZero(t, l.MinimumBaseAmount, "GetOrderExecutionLimits should not return empty minimum base amount")
```
gbjk: 🚧 I think this is incorrect, in that it's using `!assert.Error` when we're wanting no error. That's asserting it *will* error, and then running a check if that passes
🐒 Also, I think we're using `Get...Limits` to verify that `Update` worked, so the failure wording should be around setting something, not around the method we use to check it.
```suggestion
			require.Error(t, err, "GetOrderExecutionLimits must not error") {
			assert.NotZero(t, l.MinimumBaseAmount, "MinimumBaseAmount should be set")
```

common/key/key.go
gloriousCode: `*currency.Item` are directly comparable in a way that `currency.Pair` isn't without invoking functions - unless it has been fundamentally changed while I was away. A refinement of keys like that sounds good to make it an interface, but it is quite a tangent for this PR and think it can get its own run

common/key/key_test.go
gloriousCode: I feel like I made this to get infront of you or shazbert asking for it to be able to return nil or error 😄

exchange/order/limits/levels.go
gloriousCode: I'll have a look for you, my brain prefers new lines between each `if` check

exchanges/bitfinex/bitfinex_test.go
gloriousCode: I like to see all the errors rather than cancel on the first instance of one

common/key/key.go
gloriousCode: It definitely wasn't about import cycles for this decision. I wouldn't think to pass in an entire exchange when wanting to make a key, but happy to give it a go

gloriousCode: However it does lead to import cycle issues (😆) which means bigger things like moving the exchange interface elsewhere like you are suggesting.

Given you want other refactorings on the `key` package, I think it should go into its own `key` refactor PR instead

common/key/key.go
gloriousCode: The amount of packages this impacts is large. If we are using keys as pointers, then many of the stores are being changed. I feel like this too is way outside of the scope of just being able to make a simple key and is part of a key refactor PR. If I want to minimise the impact I can just point to the underlying, but then whats the point of this change if I do that?

If pressed, I can continue, but this also adds the burden for everyone else to test all over the place with such a change.

shazbert: I would opt not to do it; 
It would be heap allocated and then we would have to `deref` the ptr value on all map lookups?

common/key/key.go
shazbert: If you define the interface locally in the key package there won't be an import cycle but you would need to drop all .GetName() calls. But like you said it's still out of scope and can be its own PR.

exchanges/gateio/gateio_wrapper.go
```diff
@@ -422,9 +423,6 @@ func (e *Exchange) FetchTradablePairs(ctx context.Context, a asset.Item) (curren
 				continue
 			}
 			p := strings.ToUpper(tradables[x].ID)
```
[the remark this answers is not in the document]
gloriousCode: shazbert told me to remove it mr AI. [You can argue with him](https://github.com/thrasher-corp/gocryptotrader/pull/1860#discussion_r2013355772) 😆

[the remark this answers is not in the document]
shazbert: 🤷

shazbert (approved):
LGTM. Thanks @gloriousCode

common/key/key.go
gbjk: It was a super nit - Resolving.

common/key/key.go
gbjk: > If you define the interface locally in the key package there won't be an import cycle but you would need to drop all .GetName() calls. But like you said it's still out of scope and can be its own PR.

I don't understand this, given I was suggesting a (local) interface with just the GetName method:

> For this, I'd use an exchange interface which does GetName() string and just that, since it satisfies identification, and any error handling, but can recast easily.

Note: I'm doing this enough that I'm starting to think of a interfaces/ `Namer`.
The issue we have is that our interfaces intermingle so much that they import cycle.
The way to avoid this is to redeclare micro interfaces either in consumers, or in interfaces packages.
That was the original intention with interfaces anyway.

Resolving - Out of Scope.

exchanges/bitfinex/bitfinex_test.go
gbjk: Okay. Agreed, in this context.

backtester/eventhandlers/exchange/exchange.go
```diff
@@ -119,9 +119,9 @@ func (e *Exchange) ExecuteOrder(o order.Event, dh data.Handler, om *engine.Order
 	}
 
 	if cs.CanUseExchangeLimits || cs.UseRealOrders {
-		// Conforms the amount to the exchange order defined step amount
+		// Validate the amount to the exchange order defined step amount
 		// reducing it when needed
```
gbjk: 🚧 I don't think this comment reflects what is happening
But I also don't think it's adding any information over the well defined method name
Suggest removing.
```suggestion
```

Oh, I see what happened. You were going to use Validate, but then switched it out.

exchanges/binance/binance_test.go
```diff
@@ -2732,33 +2732,36 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
-			assert.Positivef(t, l.MaxAlgoOrders, "MaxAlgoOrders should be positive for %s pair %s", a, p)
-		}
+		t.Run(a.String()+"/"+tests[a].String(), func(t *testing.T) {
+			t.Parallel()
+			err := e.UpdateOrderExecutionLimits(t.Context(), a)
+			require.NoError(t, err, "UpdateOrderExecutionLimits must not error")
+
+			p := tests[a]
+			l, err := e.GetOrderExecutionLimits(a, p)
+			require.NoErrorf(t, err, "GetOrderExecutionLimits must not error for %s pair %s", a, p)
```
gbjk: 🚧 🌴  You've introduced a Run, but kept the `assquire.*f`
This applies to everything below.
```suggestion
			require.NoError(t, err, "GetOrderExecutionLimits must not error")
```

exchanges/okx/okx_test.go
```diff
@@ -3319,23 +3319,21 @@ func TestUpdateTradablePairs(t *testing.T) {
...
-		}
+		t.Run(a.String(), func(t *testing.T) {
+			t.Parallel()
+			require.NoError(t, e.UpdateOrderExecutionLimits(contextGenerate(), a), "UpdateOrderExecutionLimits must not error")
 
-		p, err := e.GetAvailablePairs(a)
-		require.NoErrorf(t, err, "GetAvailablePairs for asset %s must not error", a)
-		require.NotEmptyf(t, p, "GetAvailablePairs for asset %s must not return empty pairs", a)
+			p, err := e.GetAvailablePairs(a)
+			require.NoErrorf(t, err, "GetAvailablePairs for asset %s must not error", a)
```
gbjk: 🚧 🌴  Also using `*f` inside a Run
```suggestion
			require.NoError(t, err, "GetAvailablePairs must not error")
```

gbjk (changes requested):
Good work 👍 

Minor changes only - Will re-review as soon as you throw them back at me.

exchanges/binance/binance_test.go
gloriousCode: whoops sorry! I dont think I understood at the time

exchanges/okx/okx_test.go
gloriousCode: What makes this an issue when iterating over things? You don't know where the issue occurs
We have 135 instances of NoErrorf for assertify and some of them introduced by you.
What is the rule? Is it for the test setup? ie GetAvailablePairs being the problem because it is not the main purpose of the test? Or is it because you want less detail in a loop?

gloriousCode: I think to answer my own question, its when its specifically in its own t.run. So assertify F is not a problem unless it is part of a subtest, where the context can be inferred from the subtest name. I think I just didn't read much into what you were saying before commenting/complaining 😄 
Anyway, it should be resolved now

database/models/postgres/psql_main_test.go
```diff
@@ -95,9 +96,9 @@ func (p *pgTester) setup() error {
 			return err
 		}
 
-		dumpCmd := exec.Command("pg_dump", "--schema-only", p.dbName)
+		dumpCmd := exec.CommandContext(context.TODO(),"pg_dump", "--schema-only", p.dbName)
```
thrasher-: Guessing this was a bulk find + replace for `exec.Command` since this is auto generated model code and ignored by the linter so it can be removed. Same with the others

engine/sync_manager.go
```diff
@@ -404,7 +393,7 @@ func (m *SyncManager) WebsocketUpdate(exchangeName string, p currency.Pair, a as
 		s.IsUsingREST = false
 		if m.config.LogSwitchProtocolEvents {
 			log.Warnf(log.SyncMgr,
-				"%s %s %s: %s Websocket re-enabled, switching from rest to websocket",
+				"%q %q %q: %q Websocket re-enabled, switching from rest to websocket",
```
thrasher-: Just a note this this quoting of the exchange name would be inconsistent with all the other logged messages. I generally say to use %q when someone intentionally wanted to quote something; e.g they already did '%s' where %q accomplishes the same thing

Image

exchange/order/limits/levels_test.go
```diff
@@ -0,0 +1,166 @@
...
+
+	tt = nil
+	val = tt.FloorAmountToStepIncrementDecimal(decimal.NewFromInt(150))
+	assert.Equal(t, "150", val.String())
+}
+
+func TestConformToAmount(t *testing.T) {
+	t.Parallel()
+	tt := &MinMaxLevel{}
+	if tt.FloorAmountToStepIncrement(1.001) != 1.001 {
```
thrasher-:
```suggestion
	require.Equal(t, 1.001, tt.FloorAmountToStepIncrement(1.001))
```

Though I noticed all the others are asserts, so could use that instead

exchanges/bitfinex/bitfinex_test.go
```diff
@@ -148,20 +148,15 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
-				continue
-			}
-			if limits.MinimumBaseAmount == 0 {
-				t.Errorf("UpdateOrderExecutionLimits empty minimum base amount; Pair: %s Expected Limit: %v", pair, limits.MinimumBaseAmount)
+		t.Run(assetItem.String(), func(t *testing.T) {
+			t.Parallel()
+			require.NoError(t, e.UpdateOrderExecutionLimits(t.Context(), assetItem), "UpdateOrderExecutionLimits must not error")
+			for _, pair := range pairs {
+				l, err := e.GetOrderExecutionLimits(assetItem, pair)
+				assert.NoError(t, err)
```
thrasher-:
```suggestion
				require.NoError(t, err)
```

The previous code continued on error but we typically just fail immeditely on something like this now

exchanges/bitstamp/bitstamp_test.go
```diff
@@ -241,40 +241,31 @@ func TestUpdateTradablePairs(t *testing.T) {
...
 	}
 	for assetItem, limitTests := range tests {
-		if err := e.UpdateOrderExecutionLimits(t.Context(), assetItem); err != nil {
-			t.Errorf("Error fetching %s pairs for test: %v", assetItem, err)
-		}
-		for _, limitTest := range limitTests {
-			limits, err := e.GetOrderExecutionLimits(assetItem, limitTest.pair)
-			if err != nil {
-				t.Errorf("Bitstamp GetOrderExecutionLimits() error during TestExecutionLimits; Asset: %s Pair: %s Err: %v", assetItem, limitTest.pair, err)
+		assert.NoError(t, e.UpdateOrderExecutionLimits(t.Context(), assetItem), "UpdateOrderExecutionLimits should not error")
```
thrasher-:
```suggestion
		require.NoError(t, e.UpdateOrderExecutionLimits(t.Context(), assetItem), "UpdateOrderExecutionLimits must not error")
```

exchanges/bitstamp/bitstamp_test.go
```diff
@@ -241,40 +241,31 @@ func TestUpdateTradablePairs(t *testing.T) {
...
+		for _, lt := range limitTests {
+			l, err := e.GetOrderExecutionLimits(assetItem, lt.pair)
+			if assert.NoError(t, err, "GetOrderExecutionLimits should not error") {
```
thrasher-: Think this would be simpler with a require here too

exchanges/okx/okx_test.go
```diff
@@ -3318,23 +3319,21 @@ func TestUpdateTradablePairs(t *testing.T) {
...
 
-		limits, err := e.GetOrderExecutionLimits(a, p[0])
-		if assert.NoErrorf(t, err, "GetOrderExecutionLimits for asset %s and pair %s should not error", a, p[0]) {
-			require.Positivef(t, limits.PriceStepIncrementSize, "PriceStepIncrementSize must be positive for %s", p[0])
-			require.Positivef(t, limits.MinimumBaseAmount, "MinimumBaseAmount must be positive for %s", p[0])
-		}
+			l, err := e.GetOrderExecutionLimits(a, p[0])
+			require.NoError(t, err, "GetOrderExecutionLimits must not error")
+			assert.Positivef(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should be positive for %s", p)
+			assert.Positivef(t, l.MinimumBaseAmount, "PriceStepIncrementSize should be positive for %s", p)
```
thrasher-:
```suggestion
			assert.Positivef(t, l.MinimumBaseAmount, "MinimumBaseAmount should be positive for %s", p)
```

exchanges/exchange_test.go
```diff
@@ -420,29 +421,19 @@ func TestSetCurrencyPairFormat(t *testing.T) {
...
 	spot, err = b.GetPairFormat(asset.Spot, false)
-	if err != nil {
-		t.Fatal(err)
-	}
-	if spot.Delimiter != "~" {
-		t.Error("incorrect pair format delimiter")
-	}
+	require.NoError(t, err, "GetPairFormat must not error")
+	assert.Equal(t, "~", spot.Delimiter, "GetPairFormat should return a format with correct delimiter")
+	assert.Equal(t, "~", spot.Delimiter, "GetPairFormat should return a format with correct delimiter")
```
thrasher-: Duplicated test here

exchanges/binance/binance_test.go
gloriousCode: Rather than going in a manually replacing all those fields since they were all so unique, I redid all `TestUpdateOrderExecutionLimits` to a standard which doesn't include such assertify f's. Hope it's better, can revert if not. I don't believe testing each field is worthwhile so I just check one.

engine/sync_manager.go
gloriousCode: I may have done a big find and replace with malice. I'll revert the areas I've impacted like this

exchanges/binance/binance_test.go
gbjk: 🚧 Soz buddy, no, I don't agree.
I'm not against changing the way we do this, but not if it reduces RealWorld coverage.
That is: Lines that we actually assert are doing the right thing, rather than just executed by a test.

So looking again at #1464 which man-handled this into this shape, I think our coverage isn't at the right unit.
We're covering all the `Fetch*` units implicitly via `Update*`.
So either:
1. Leave TestUpdate doing a full coverage test
1. Move test coverage for for each type into `TestFetch*` - This will mean duplicating the common assertions though, and just have `TestUpdate` testing one thing.

For now, I'd vote that you just revert that change and fix the `assert.*f`

exchanges/okx/okx_test.go
gbjk: I think we're on the same page. I'll just restate it from my view and if I don't say anything you don't agree with then we're good 👍 
* Users should be able to understand the failure by combining the test name with the message
* Assertions should have a description unless the function called was the same name as the unit test, and it's clear what's being tested and why
* ErrorIs doesn't need a description if it's unique and clear enough on it's own

I think maybe we're being a bit more anal about format for these right now to establish good precedents throughout the codebase so that contributors have a good chance of copy/pasting the right thing.

Noting: I'm only a `t.Run` convert because of you 😛

exchanges/okx/okx_test.go
```diff
@@ -3323,16 +3323,12 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
-			require.NotEmptyf(t, p, "GetAvailablePairs for asset %s must not return empty pairs", a)
-
-			l, err := e.GetOrderExecutionLimits(a, p[0])
+			require.NoError(t, e.UpdateOrderExecutionLimits(t.Context(), a), "UpdateOrderExecutionLimits must not error")
+			pairs, err := e.CurrencyPairs.GetPairs(a, true)
+			require.NoError(t, err, "GetPairs must not error")
+			l, err := e.GetOrderExecutionLimits(a, pairs[0])
 			require.NoError(t, err, "GetOrderExecutionLimits must not error")
-			assert.Positivef(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should be positive for %s", p)
-			assert.Positivef(t, l.MinimumBaseAmount, "PriceStepIncrementSize should be positive for %s", p)
```
gbjk: 🚧 Same as binance, I don't think we should reduce coverage like this.

exchanges/binance/binance_test.go
gloriousCode: Unfortunately I don't understand the way you are talking. Are you saying that you want the same amount of assertions on data? There's many changes in that PR and there's a lot of words you are saying despite coverage is the same. Its very hard for me to understand what you want changed to improve things outside of reverting

gbjk: Let me be clearer. Sorry if it sounds ruder as a result.

* The fields you removed are not being tested. They need to be.
( MinPrice, MaxPrice, etc )
Please revert the change that removed testing for them.

* "Coverage" isn't the only thing that matters. We need to know that values are correct.
Example: If someone removes a line that means MaxTotalOrders is not being set, coverage would go _up_. But it's a break

I'll make a clear suggestion at the line for you, to speed this up.

exchanges/binance/binance_test.go
```diff
@@ -2746,60 +2704,31 @@ func TestFormatUSDTMarginedFuturesPair(t *testing.T) {
...
-			assert.Positivef(t, limits.MaxAlgoOrders, "MaxAlgoOrders should be positive for %s pair %s", a, p)
-		}
+		t.Run(a.String(), func(t *testing.T) {
+			t.Parallel()
+			require.NoError(t, e.UpdateOrderExecutionLimits(t.Context(), a), "UpdateOrderExecutionLimits must not error")
+			pairs, err := e.CurrencyPairs.GetPairs(a, false)
+			require.NoError(t, err, "GetPairs must not error")
+			l, err := e.GetOrderExecutionLimits(a, pairs[0])
+			require.NoError(t, err, "GetOrderExecutionLimits must not error")
+			assert.NotZero(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should not be zero")
```
gbjk:
```suggestion
			assert.Positive(t, l.MinPrice, "MinPrice should be positive")
			assert.Positive(t, l.MaxPrice, "MaxPrice should be positive")
			assert.Positive(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should be positive")
			assert.Positive(t, l.MinimumBaseAmount, "MinimumBaseAmount should be positive")
			assert.Positive(t, l.MaximumBaseAmount, "MaximumBaseAmount should be positive")
			assert.Positive(t, l.AmountStepIncrementSize, "AmountStepIncrementSize should be positive")
			assert.Positive(t, l.MarketMaxQty, "MarketMaxQty should be positive")
			assert.Positive(t, l.MaxTotalOrders, "MaxTotalOrders should be positive")
			switch a {
			case asset.Spot, asset.Margin:
				assert.Positive(t, l.MaxIcebergParts, "MaxIcebergParts should be positive")
			case asset.USDTMarginedFutures:
				assert.Positive(t, l.MinNotional, "MinNotional should be positive")
				fallthrough
			case asset.CoinMarginedFutures:
				assert.Positive(t, l.MultiplierUp, "MultiplierUp should be positive")
				assert.Positive(t, l.MultiplierDown, "MultiplierDown should be positive")
				assert.Positive(t, l.MarketMinQty, "MarketMinQty should be positive")
				assert.Positive(t, l.MarketStepIncrementSize, "MarketStepIncrementSize should be positive")
				assert.Positive(t, l.MaxAlgoOrders, "MaxAlgoOrders should be positive")
			}
```

gbjk (changes requested):

exchanges/okx/okx_test.go
```diff
@@ -3323,16 +3323,12 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
+			assert.NotZero(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should not be zero")
```
gbjk: Aside: I'm "going with it" that this should move from positive to "not zero" - Assuming that was deliberate and applies to both fields

```suggestion
			assert.NotZero(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should not be zero")
			assert.NotZero(t, l.MinimumBaseAmount, "MinimumBaseAmount should not be zero")
```

samuael (approved):
ch ACK!

exchanges/bitstamp/bitstamp_test.go
```diff
@@ -241,42 +241,16 @@ func TestUpdateTradablePairs(t *testing.T) {
...
-		}
+	for _, a := range e.GetAssetTypes(false) {
+		t.Run(a.String(), func(t *testing.T) {
+			t.Parallel()
+			require.NoError(t, e.UpdateOrderExecutionLimits(t.Context(), a), "UpdateOrderExecutionLimits must not error")
+			pairs, err := e.CurrencyPairs.GetPairs(a, false)
+			require.NoError(t, err, "GetPairs must not error")
+			l, err := e.GetOrderExecutionLimits(a, pairs[0])
+			require.NoError(t, err, "GetOrderExecutionLimits must not error")
+			assert.NotZero(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should not be zero")
```
gbjk: 🤷 Was this deliberately removing the mock tests?
This is reverting the changes in #1423 which I think had value.
At the very least I'd like to re-add the coverage on the other 2 fields, but also want to ensure this wasn't an accidental hit and run?

I asked before, but also just to check: Is there a reason we're moving from `Positive` to `NotZero` for this field?
```suggestion
			assert.NotEmpty(t, l.Pair, "Pair should not be empty")
			assert.Positive(t, l.PriceStepIncrementSize, "PriceStepIncrementSize should be positive")
			assert.Positive(t, l.AmountStepIncrementSize, "AmountStepIncrementSize should be positive")
			assert.Positive(t, l.MinimumQuoteAmount, "MinimumQuoteAmount should be positive")
```

gloriousCode: I have not updated it since previous discussions where you clarified you wanted the same asserts as prior

gbjk: Sorry. I got excited and had a premature monday morning review. 🫡

PR #1861, opened 2025-03-25
portfolio: Fix CryptoID balance issue and assertify tests

thrasher-:
# PR Description

CryptoID API now requires an API key + rate limiter for balance requests which was causing a test failure. This PR adds support for that, makes the portfolio concurrency safe as well as assertifying all tests + adding additional test coverage.

I think another pass will be required to change some things around like having all providers conforming to an interface, listing which coins they support etc but just wanted to get tests working again.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/config/versions/v5.go
+++ b/config/versions/v5.go
@@ -20,7 +20,7 @@ func init() {
 // Note: Do not be tempted to use a constant for Duration. Whilst defaults are still written to config, we need to manage default upgrades discretely.
 var defaultFuturesTrackingSeekDuration = strconv.FormatInt(int64(time.Hour)*24*365, 10)
 
-var defaultConfig = []byte(`{
+var defaultConfigV5 = []byte(`{
   "enabled": true,
   "verbose": false,
   "activelyTrackFuturesPositions": true,
@@ -37,7 +37,7 @@ func (v *Version5) UpgradeConfig(_ context.Context, e []byte) ([]byte, error) {
 	_, valueType, _, err := jsonparser.Get(e, "orderManager", "enabled")
 	switch {
 	case errors.Is(err, jsonparser.KeyPathNotFoundError), valueType == jsonparser.Null:
-		return jsonparser.Set(e, defaultConfig, "orderManager")
+		return jsonparser.Set(e, defaultConfigV5, "orderManager")
 	case err != nil:
 		return e, err
 	}
--- a/config/versions/v6_test.go
+++ b/config/versions/v6_test.go
@@ -8,6 +8,7 @@ import (
 	"github.com/buger/jsonparser"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	v6 "github.com/thrasher-corp/gocryptotrader/config/versions/v6"
 )
 
 var testData = []byte(`{
@@ -30,7 +31,7 @@ func TestUpgrade(t *testing.T) {
 	t.Parallel()
 	r, err := new(Version6).UpgradeConfig(context.Background(), testData)
 	require.NoError(t, err, "UpgradeConfig must not error")
-	assert.Contains(t, string(r), string(defaultConfigV6))
+	require.True(t, bytes.Contains(r, v6.DefaultConfig))
 
 	r2, err := new(Version6).UpgradeConfig(context.Background(), r)
 	require.NoError(t, err, "UpgradeConfig must not error")
@@ -42,7 +43,7 @@ func TestDowngrade(t *testing.T) {
 
 	r, err := new(Version6).UpgradeConfig(context.Background(), testData)
 	require.NoError(t, err, "UpgradeConfig must not error")
-	require.True(t, bytes.Contains(r, defaultConfigV6))
+	require.True(t, bytes.Contains(r, v6.DefaultConfig))
 
 	r, err = new(Version6).DowngradeConfig(context.Background(), r)
 	require.NoError(t, err, "DowngradeConfig must not error")
--- a/config/versions/v6_test.go
+++ b/config/versions/v6_test.go
@@ -1,0 +1,46 @@
+package versions
+
+import (
+	"bytes"
+	"context"
+	"testing"
+
+	"github.com/buger/jsonparser"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var testData = []byte(`{
+	"portfolioAddresses": {
+	  "addresses": [
+	   {
+		"Address": "1JCe8z4jJVNXSjohjM4i9Hh813dLCNx2Sy",
+		"CoinType": "BTC",
+		"Balance": 0.00108832,
+		"Description": "",
+		"WhiteListed": false,
+		"ColdStorage": false,
+		"SupportedExchanges": ""
+	   }
+	}
+}`)
+
+func TestUpgrade(t *testing.T) {
+	t.Parallel()
+	r, err := new(Version6).UpgradeConfig(context.Background(), testData)
+	require.NoError(t, err, "UpgradeConfig must not error")
+	assert.True(t, bytes.Contains(r, defaultConfigV6))
+}
+
+func TestDowngrade(t *testing.T) {
+	t.Parallel()
+
+	r, err := new(Version6).UpgradeConfig(context.Background(), testData)
+	require.NoError(t, err, "UpgradeConfig must not error")
+	require.True(t, bytes.Contains(r, defaultConfigV6))
+
+	r, err = new(Version6).DowngradeConfig(context.Background(), r)
+	require.NoError(t, err, "DowngradeConfig must not error")
+	_, _, _, err = jsonparser.Get(r, "portfolioAddresses", "providers") //nolint:dogsled // Return values not needed
+	assert.ErrorIs(t, err, jsonparser.KeyPathNotFoundError, "providers should be removed")
+}
--- a/config/versions/v6_test.go
+++ b/config/versions/v6_test.go
@@ -22,6 +22,7 @@ var testData = []byte(`{
 		"ColdStorage": false,
 		"SupportedExchanges": ""
 	   }
+	  ]
 	}
 }`)
 
@@ -29,7 +30,11 @@ func TestUpgrade(t *testing.T) {
 	t.Parallel()
 	r, err := new(Version6).UpgradeConfig(context.Background(), testData)
 	require.NoError(t, err, "UpgradeConfig must not error")
-	assert.True(t, bytes.Contains(r, defaultConfigV6))
+	assert.Contains(t, string(r), string(defaultConfigV6))
+
+	r2, err := new(Version6).UpgradeConfig(context.Background(), r)
+	require.NoError(t, err, "UpgradeConfig must not error")
+	assert.Equal(t, r, r2, "UpgradeConfig should not affect an already upgraded config")
 }
 
 func TestDowngrade(t *testing.T) {
--- a/portfolio/portfolio.go
+++ b/portfolio/portfolio.go
@@ -34,20 +34,17 @@ const (
 )
 
 var (
-	errNotETHAddress           = errors.New("not an Ethereum address")
-	errAddressIsEmptyOrInvalid = errors.New("address is empty or invalid")
 	errProviderNotFound        = errors.New("provider not found")
 	errProviderNotEnabled      = errors.New("provider not enabled")
 	errProviderAPIKeyNotSet    = errors.New("provider API key not set")
 	errPortfolioItemNotFound   = errors.New("portfolio item not found")
-	errUnsupportedCoin         = errors.New("unsupported coin")
 	errNoPortfolioItemsToWatch = errors.New("no portfolio items to watch")
 )
 
 // GetEthereumAddressBalance fetches Ethereum address balance for a given address
 func (b *Base) GetEthereumAddressBalance(ctx context.Context, address string) (float64, error) {
-	if ok, _ := common.IsValidCryptoAddress(address, "eth"); !ok {
-		return 0, errNotETHAddress
+	if err := common.IsValidCryptoAddress(address, "eth"); err != nil {
+		return 0, err
 	}
 
 	apiKey := "freekey"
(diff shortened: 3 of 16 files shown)


portfolio/portfolio.go
```diff
@@ -27,55 +28,74 @@ const (
...
+	errProviderAPIKeyNotSet    = errors.New("provider API key not set")
+	errPortfolioItemNotFound   = errors.New("portfolio item not found")
+	errUnsupportedCoinType     = errors.New("unsupported coin type")
+	errNoPortfolioItemsToWatch = errors.New("no portfolio items to watch")
 )
 
-var errNotEthAddress = errors.New("not an Ethereum address")
+// GetEthereumAddressBalance fetches Ethereum address balance for a given address
+func (b *Base) GetEthereumAddressBalance(ctx context.Context, address string) (float64, error) {
+	if ok, _ := common.IsValidCryptoAddress(address, "eth"); !ok {
```
gloriousCode, line 48: Bringing up unrelated things sorry. What makes this address invalid?
https://etherscan.io/address/0xdadb0d80178819f2319190d340ce9a924f783711

```
[ERROR] | PORTFOLIO | 26/03/2025 10:59:59 | PortfolioWatcher: UpdatePortfolio error: error getting balance for 0xdadB0d80178819F2319190D340ce9A924f783711: not an Ethereum address for currency ETH
```

gloriousCode (approved):
Upgraded config,
Downgraded config,
Synced with all three providers
Tested with enabled set to false
Tested without API keys
Checked coverage

tACK

portfolio/portfolio_test.go
```diff
@@ -1,496 +1,270 @@
...
+	"github.com/stretchr/testify/assert"
 	"github.com/thrasher-corp/gocryptotrader/core"
 	"github.com/thrasher-corp/gocryptotrader/currency"
 )
 
 const (
-	testBTCAddress = "0x1D01TH0R53"
+	testInvalidBTCAddress = "0x1D01TH0R53"
+	testLTCAddress        = "LX2LMYXtuv5tiYEMztSSoEZcafFPYJFRK1"
+	testBTCAddress        = "1A1zP1eP5QGefi2DMPTfTL5SLmv7DivfNa"
```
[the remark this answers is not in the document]
thrasher-: image

portfolio/portfolio.go
```diff
@@ -27,55 +28,74 @@ const (
...
-		http.MethodGet,
-		url,
-		nil,
-		nil,
-		b.Verbose)
+	if err := b.cryptoIDLimiter.Wait(ctx); err != nil {
+		return 0, fmt.Errorf("rate limiter wait error: %w", err)
+	}
+
+	urlPath := fmt.Sprintf("%s/%s/api.dws?q=getbalance&a=%s&key=%s", cryptoIDAPIURL, coinType.Lower(), address, p.APIKey)
```
gbjk, line 96: 🥃 I note that elsewhere recently you've been replacing Sprintf with concat `+`

portfolio/portfolio.go
```diff
@@ -27,55 +28,74 @@ const (
...
+)
+
+var (
+	errNotETHAddress           = errors.New("not an Ethereum address")
+	errAddressIsEmptyOrInvalid = errors.New("address is empty or invalid")
+	errProviderNotFound        = errors.New("provider not found")
+	errProviderNotEnabled      = errors.New("provider not enabled")
+	errProviderAPIKeyNotSet    = errors.New("provider API key not set")
+	errPortfolioItemNotFound   = errors.New("portfolio item not found")
+	errUnsupportedCoinType     = errors.New("unsupported coin type")
```
gbjk, line 42: 🐒 
* I think Coin Type is actually Coin
* And I think Coin should actually be Currency
* So could we just use `currency.ErrCurrencyNotSupported`

portfolio/portfolio.go
```diff
@@ -27,55 +28,74 @@ const (
...
-		return EthplorerResponse{}, errNotEthAddress
+	apiKey := "freekey"
+	if p, ok := b.Providers.GetProvider("ethplorer"); ok && p.APIKey != "" {
+		apiKey = p.APIKey
 	}
 
-	urlPath := fmt.Sprintf(
-		"%s/%s/%s?apiKey=freekey", ethplorerAPIURL, ethplorerAddressInfo, address,
-	)
+	urlPath := fmt.Sprintf("%s/%s/%s?apiKey=%s", ethplorerAPIURL, ethplorerAddressInfo, address, apiKey)
```
gbjk, line 57: 🥃 I note that elsewhere recently you've been replacing Sprintf with concat +

portfolio/portfolio.go
```diff
@@ -472,70 +507,81 @@ func (b *Base) GetPortfolioGroupedCoin() map[currency.Code][]string {
...
-	b.Addresses = port.Addresses
-}
-
 // StartPortfolioWatcher observes the portfolio object
-func (b *Base) StartPortfolioWatcher() {
-	addrCount := len(b.Addresses)
-	log.Debugf(log.PortfolioMgr,
-		"PortfolioWatcher started: Have %d entries in portfolio.\n", addrCount,
-	)
+func (b *Base) StartPortfolioWatcher(ctx context.Context, sleepDelay time.Duration) error {
```
gbjk, line 511: 🐒 It's not really a sleep delay, it's an interval / `checkInterval` / `runInterval`.
I would have expected a sleepDelay to mean that the watcher will wait Duration before starting operations.

```suggestion
func (b *Base) StartPortfolioWatcher(ctx context.Context, interval time.Duration) error {
```

portfolio/portfolio.go
```diff
@@ -472,70 +507,81 @@ func (b *Base) GetPortfolioGroupedCoin() map[currency.Code][]string {
...
+	if len(b.Addresses) == 0 {
+		return errNoPortfolioItemsToWatch
+	}
+
+	if sleepDelay <= 0 {
+		sleepDelay = time.Minute * 10
+	}
```
gbjk, line 518: 🐒 Hidden constant
```suggestion
	if sleepDelay <= 0 {
		sleepDelay = defaultInterval
	}
```

portfolio/portfolio.go
```diff
@@ -472,70 +507,81 @@ func (b *Base) GetPortfolioGroupedCoin() map[currency.Code][]string {
...
+
+	log.Infof(log.PortfolioMgr, "PortfolioWatcher started: Have %d entries in portfolio.\n", len(b.Addresses))
+
+	ticker := time.NewTicker(sleepDelay)
+	defer ticker.Stop()
+
```
gbjk, line 524: 🚧 Do we really want to wait Interval before first fetch?
That is: If the user has set it to 1 hour, justifiably, should they have to wait before getting any data populated?

🥃 Kinda never tried this, but I wonder if we could stuff something down the channel to cause a premature ejaculation?
```suggestion
ticker.C <- time.Now()
```

portfolio/portfolio_test.go
```diff
@@ -501,147 +275,129 @@ func TestGetPortfolioSummary(t *testing.T) {
...
-		t.Error("incorrect balance")
-	}
+	assert.Equal(t, currency.LTC, getTotalsVal(currency.LTC).Coin, "Coin should be LTC")
+	assert.Equal(t, 23.0, getTotalsVal(currency.LTC).Balance, "LTC balance should be correct")
+	assert.Equal(t, 200.0, getTotalsVal(currency.BTC).Balance, "BTC balance should be correct")
+	assert.Equal(t, 69.0+420.0+42, getTotalsVal(currency.ETH).Balance, "ETH balance should be correct")
 }
 
-func TestGetPortfolioGroupedCoin(t *testing.T) {
+func TestGetPortfolioAddressesGroupedByCoin(t *testing.T) {
```
gbjk, line 284: 🐒  Adding this as a in-code placeholder for the review summary comment that I think we should replace `Coin` with `Currency` throughout
```suggestion
func TestGetPortfolioAddressesGroupedByCurrency(t *testing.T) {
```

gbjk (changes requested):
Great work 💪 

Minor changes requested.

Throughout this, I'd like to advocate that we should replace `Coin` with `Currency`.

config/versions/v5.go
```diff
@@ -20,7 +20,7 @@ func init() {
 // Note: Do not be tempted to use a constant for Duration. Whilst defaults are still written to config, we need to manage default upgrades discretely.
 var defaultFuturesTrackingSeekDuration = strconv.FormatInt(int64(time.Hour)*24*365, 10)
 
-var defaultConfig = []byte(`{
+var defaultConfigV5 = []byte(`{
```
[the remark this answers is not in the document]
thrasher-: Was about to tackle then but then realised it's quite widespread with the test vars and 3 versions impacted😅 Can do that in a separate PR

portfolio/portfolio.go
thrasher-: Good point, I've now fired it immediately before starting the ticker

portfolio/portfolio.go
thrasher-: Going to defer this since although Bitcoin is a crypto**currency**, it's also a coin (Bit**coin**) so are often used interchangibly (like coinmarketcap.com rather than cryptocurrencymarketcap) but I do agree that the type at the end can be shedded 👍

portfolio/portfolio.go
```diff
@@ -233,84 +256,88 @@ func (b *Base) AddAddress(address, description string, coinType currency.Code, b
...
-			if err != nil {
-				return err
-			}
+	type balanceRetriever func(ctx context.Context, address string) (float64, error)
 
-			if result.Error.Message != "" {
-				return errors.New(result.Error.Message)
-			}
+	var providerName string
+	var getBalance balanceRetriever
```
shazbert, line 290:
```suggestion
	var getBalance func(ctx context.Context, address string) (float64, error)
```

and remove the type dec `balanceRetriever` above

portfolio/portfolio_test.go
```diff
@@ -1,496 +1,273 @@
...
-	}
+	b := Base{}
+	b.AddExchangeAddress("BallerExchange", currency.LTC, 420)
+	bal, ok := b.GetAddressBalance("BallerExchange", ExchangeAddress, currency.LTC)
+	assert.True(t, ok, "GetAddressBalance should return true for existing address")
+	assert.Equal(t, 420.0, bal, "GetAddressBalance should return the correct balance")
+	b.RemoveExchangeAddress("BallerExchange", currency.LTC)
+	bal, ok = b.GetAddressBalance("BallerExchange", ExchangeAddress, currency.LTC)
+	assert.False(t, ok, "GetAddressBalance should return false for non-existent address")
+	assert.Equal(t, 0.0, bal, "GetAddressBalance should return 0 for non-existent address")
```
shazbert, line 151: assert zero

portfolio/portfolio_test.go
```diff
@@ -501,147 +278,129 @@ func TestGetPortfolioSummary(t *testing.T) {
...
+	go func() {
+		doneCh <- b.StartPortfolioWatcher(ctx, 0) // 0 will set the ticker to 10 minutes
+	}()
+
+	select {
+	case err := <-doneCh:
+		assert.ErrorIs(t, err, context.DeadlineExceeded, "StartPortfolioWatcher should return DeadlineExceeded error")
+	case <-time.After(2 * time.Second):
+		assert.Fail(t, "StartPortfolioWatcher should have deadlined")
 	}
```
shazbert, line 365:
```suggestion
	ctx, cancel := context.WithTimeout(t.Context(), 0)
	cancel()
	assert.ErrorIs(t, b.StartPortfolioWatcher(ctx, 0), context.DeadlineExceeded, "StartPortfolioWatcher should return DeadlineExceeded error")
```

Can cancel it straight away

shazbert (changes requested):
Tested and looks great only some minor things I found. 🚀

shazbert (approved):
Thanks changes ACK!

config/versions/v5.go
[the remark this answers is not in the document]
gbjk: I'm okay with addressing the others in another PR.
But I feel that this PR introduces Version6, and so it's `DefaultConfig` should be in pkg `v6`.

portfolio/portfolio.go
```diff
@@ -508,18 +503,30 @@ func (b *Base) GetPortfolioAddressesGroupedByCoin() map[currency.Code][]string {
...
-	ticker := time.NewTicker(sleepDelay)
+	updatePortfolio := func() {
+		for key, value := range b.GetPortfolioAddressesGroupedByCoin() {
+			if err := b.UpdatePortfolio(ctx, value, key); err != nil {
+				log.Errorf(log.PortfolioMgr, "PortfolioWatcher: UpdatePortfolio error: %s for currency %s", err, key)
+				continue
+			}
+			log.Debugf(log.PortfolioMgr, "PortfolioWatcher: Successfully updated address balance for %s", key)
+		}
+	}
```
gbjk, line 525: 🐒 
This func is not closuring any vars apart from `b`.
So I think it should be a private method on Base instead.

🥃  Just processed what `Base` is here. Bit of an overloaded type name; We should probably drive by that at some point.

portfolio/portfolio_test.go
```diff
@@ -346,20 +350,9 @@ func TestStartPortfolioWatcher(t *testing.T) {
...
-	}()
-
-	select {
-	case err := <-doneCh:
-		assert.ErrorIs(t, err, context.DeadlineExceeded, "StartPortfolioWatcher should return DeadlineExceeded error")
-	case <-time.After(2 * time.Second):
-		assert.Fail(t, "StartPortfolioWatcher should have deadlined")
-	}
+	ctx, cancel := context.WithTimeout(t.Context(), 0)
+	cancel()
```
gbjk, line 354: 🐒 Not sure why we're using a Timeout
```suggestion
	ctx, cancel := context.WithCancel(t.Context())
	cancel()
```

gbjk (approved):
LGTM 👍 

One nit, and I still think Version6's defaults live in `v6.DefaultConfig`

Happy to proceed either way.

gbjk (approved):

config/versions/v6_test.go
```diff
@@ -0,0 +1,52 @@
...
+		"Description": "",
+		"WhiteListed": false,
+		"ColdStorage": false,
+		"SupportedExchanges": ""
+	   }
+	  ]
+	}
+}`)
+
+func TestUpgrade(t *testing.T) {
```
gbjk, line 30: Sorry; We missed this.
```suggestion
func TestVersion6Upgrade(t *testing.T) {
```

Will driveBy as part of #1864

Issue #1764, opened 2024-12-26
Subscription and Websocket multiplexing


gbjk:
# Overview

Now that subscription configuration is nearly universal, we can start to address abstracting the solution, and learning from the issues with multiplexing subscriptions.

Required features:
* Support connections per asset class
* Support runtime subscriptions, including FlushChannels
* Enabling pairs after subscription must work
* Support N+ connections due to subscription cap or rate-limiting

# Design

* Remove Subscriptions and GenerateSubs from belonging to connections

The most pressing reason for that is that in a world where we might establish N+ connections due to connection rate limiting,
We can't possible generate the right subscriptions and (in the darkness) bind them at the connection level.
I believe Subscriptions must be an exchange level entity, which are then allocated to websockets in a one-to-one relationship.

* Router middleware to map connections and subscriptions

* Just-in-time connections after the first connection

* Separate Subscription configuration from runtime subscriptions, so that "All pairs" continues to work with FlushChannels

# References 
* #1603 
* #1580

gbjk:
Requirement added from #1785

PR #1863, opened 2025-03-26
gateio: Fix websocket orderbook incremental updates

shazbert:
# PR Description

* Introduces an orderbook manager for websocket updates to ensure incremental updates are being applied correctly and flushes book when it becomes out of sync.
* Removes maps `map[string]bool` for tracking initial book, this made updates out of sync because the book was never fetched again.
* Adds method function `UpdateOrderbookWithLimit` so that a orderbook limit can be set, if initial orderbook len does not match the level of the subscription inconsistency occurs and there will be gaps in liquidity as it tries to build that book over time. 
* Adds field to orderbook.Update struct `AllowEmpty` so that if no updates have been batched between the update frequency the system can still set updated ID's to determine if it is out of sync. 
* Adds `20ms` handling in `GetIntervalString` for futures book frequency. 

Subscription/expand template

* Adds in interfaces `ValidateSubscription` and `ValidateSubscriptions` which is called in expand template which is optional for an exchange. Just so that if there are any mixed options they can be handled and an appropriate error response can be returned so you don't need to dive into the code to find out. This is until we figure out a list of options for that field and we can abstract it away nicely. 
* The functions in GateIO should also scale to other assets when they are converted over to templating. Only spot is supported right now but all pathways for orderbook handling are done. 
* Reintroduces orderbook `snapshots` and `best bid ask` channels

Fixes # (issue)

```
[ERROR] | WEBSOCKET | 26/03/2025 14:05:14 | exchange GateIO websocket error - connection URL:[wss://fx-ws.gateio.ws/v4/ws/usdt] error: malformed orderbook data
```

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,152 @@
+package gateio
```
gloriousCode, line 1: Just for curiosity's sake, what made our existing orderbook buffer insufficient?

shazbert: It doesn't allow for fetching an initial state and handing of what's required, I guess with time and effort it could be upgraded. 🤷

shazbert: Yeah had this PR open for like 8 months or so trying to abstract and consolidate this very thing across all the exchanges, but it never got reviewed so I [closed it.](https://github.com/thrasher-corp/gocryptotrader/pull/1481)

gloriousCode: I thought I had but turns out it was `Pending` 😄 well even less reason for such a consolidation then

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+func (c *updateCache) applyUpdate(g *Gateio, update *orderbook.Update) error {
+	if update.Asset != asset.Spot {
+		return g.Websocket.Orderbook.Update(update)
+	}
+
+	for _, a := range standardMarginAssetTypes {
+		if enabled, _ := g.CurrencyPairs.IsPairEnabled(update.Pair, a); !enabled {
+			continue
```
gloriousCode, line 143: Would checking the map be better? Or is it chicken and egg? These seem like more expensive checks when you know exactly which pairs and assets are being done

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,152 @@
...
+		return nil
+	}
+
+	if book == nil || book.LastUpdateID+1 < firstID /*orderbook is behind notifications so refresh*/ {
+		cache.updating = true
+		go func() {
+			// REST Orderbook IDs are behind the ws update feed by about 50-100 changes; inline rudimentary delay to cache
+			time.Sleep(time.Second)
+			if err := cache.updateOrderbookAndApply(ctx, g, update.Pair, update.Asset, limit); err != nil {
+				g.Websocket.DataHandler <- err
```
gloriousCode, line 60: So I am forcing an issue, but if this errors, there doesn't seem to be anything else that happens to recover.

On the next update we get to the if statement
`if book.LastUpdateID+1 > update.UpdateID {`
and the update.UpdateID is larger by a degree versus book.LastUpdateID+1, so it seems like the snapshot is behind the update and there are no checks for that if `updateOrderbookAndApply` errors.

Is this okay or am I missing things?

gloriousCode (changes requested):
My head is not fully around this yet, so I beg of forgiveness if my comments don't reflect reality 🙏

exchanges/gateio/ws_ob_update_manager.go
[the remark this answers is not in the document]
shazbert: https://www.youtube.com/shorts/JZDgsQhTp-8?feature=share

exchanges/gateio/ws_ob_update_manager.go
shazbert: Not a bad idea; only issue I see is to the persistence of the map I think it would be funky between assets if we say drop one margin pair, it will still think its enabled.

gloriousCode: That's all good then

exchanges/gateio/ws_ob_update_manager.go
shazbert: Yeah I only added an error because it was in the docs; once synced up I have run this over night and it didn't happen at all. For this to trigger you say `update.UpdateID is larger by a degree versus book.LastUpdateID+1` but its actually the inverse, the updates are behind the book all of a sudden 🤷 . In the off chance it does it starts giving us old updates it will complain and then should catch up? 🤷 I could probably remove this error check altogether if you want.

exchanges/gateio/gateio_websocket.go
```diff
@@ -52,20 +53,23 @@ const (
 
 	subscribeEvent   = "subscribe"
 	unsubscribeEvent = "unsubscribe"
+
+	spotOrderbookUpdateLimit = 100
```
gbjk, line 57: 🐒  Is this a depth?
```suggestion
	spotOrderbookDepthLimit = 100
```

exchanges/gateio/gateio_websocket.go
```diff
@@ -52,20 +53,23 @@ const (
...
 )
 
 var defaultSubscriptions = subscription.List{
 	{Enabled: true, Channel: subscription.TickerChannel, Asset: asset.Spot},
 	{Enabled: true, Channel: subscription.CandlesChannel, Asset: asset.Spot, Interval: kline.FiveMin},
-	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds},
+	// spot.order_book_update is locked to 100ms updates and 100 book size
```
gbjk, line 63: 🐒  `locked` doesn't feel specific enough
```suggestion
	// spot.order_book_update only available with 100ms updates and 100 levels
```

Also: Other exchanges have limitations, e.g. update must be 1s or 100ms, levels must be 100.
I don't know if we need to single this one out for a comment. 🙂

exchanges/gateio/gateio_websocket.go
```diff
@@ -52,20 +53,23 @@ const (
...
+	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds, Levels: spotOrderbookUpdateLimit},
+	{Enabled: false, Channel: spotOrderbookTickerChannel, Asset: asset.Spot, Interval: kline.Interval(10 * time.Millisecond), Levels: 1},
```
gbjk, line 65: 🐒 Feels weird to have a non-const Interval that isn't too strange to put in kline_types.
Can we have `kline.TenMilliseconds` ?

exchanges/gateio/gateio_websocket.go
```diff
@@ -764,6 +733,146 @@ func singleSymbolChannel(name string) bool {
...
+		}
+		for _, p := range s.Pairs {
+			k := key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: s.Asset}
+			loadedChanName, ok := orderbookGuard[k]
+			if !ok {
+				orderbookGuard[k] = n
+				continue
+			}
+			if loadedChanName != n {
+				return fmt.Errorf("%w for `%s %s` between `%s` and `%s`, please enable only one type", subscription.ErrExclusiveSubscription, k.Pair(), k.Asset, loadedChanName, n)
```
gbjk, line 752: Am I right in thinking that this is fixing this error:
`error: malformed orderbook data`
If so, that makes sense why we need this function in general, we can't trust them to check for us, and we need to be checking against all subscriptions.

exchanges/gateio/gateio_websocket.go
```diff
@@ -764,6 +733,146 @@ func singleSymbolChannel(name string) bool {
 	return false
 }
 
+// ValidateSubscriptions implements subscription.SubscriptionsValidator interface
```
gbjk, line 736: 🐒  Since this is doing really just one thing, feels like we can move the comment up for clarity
```suggestion
// ValidateSubscriptions implements subscription.SubscriptionsValidator interface to ensure that conflicting orderbooks aren't subscribed to
```

exchanges/gateio/gateio_websocket.go
```diff
@@ -764,6 +733,146 @@ func singleSymbolChannel(name string) bool {
...
+func (g *Gateio) ValidateSubscriptions(l subscription.List) error {
+	orderbookGuard := map[key.PairAsset]string{}
+	for _, s := range l {
+		n := channelName(s)
+		if !isRestrictedOrderbookChannel(n) {
```
gbjk, line 741: 🐒 `restricted` doesn't feel like the right adjective for this. `shared` or `clashing` maybe
```suggestion
		if !isSharedOrderbookChannel(n) {
```

exchanges/gateio/gateio_websocket.go
```diff
@@ -764,6 +733,146 @@ func singleSymbolChannel(name string) bool {
...
+			continue
+		}
+		for _, p := range s.Pairs {
+			k := key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: s.Asset}
+			loadedChanName, ok := orderbookGuard[k]
```
gbjk, line 746: 🐒 
```suggestion
			existingChanName, ok := orderbookGuard[k]
```

exchanges/gateio/gateio_websocket.go
```diff
@@ -764,6 +733,146 @@ func singleSymbolChannel(name string) bool {
...
+		optionsOrderbookChannel,
+		optionsOrderbookTickerChannel,
+		optionsOrderbookUpdateChannel:
+		return true
+	}
+	return false
+}
+
+// ValidateSubscription implements subscription.SubscriptionValidator interface
+func (g *Gateio) ValidateSubscription(s *subscription.Subscription) error {
```
gbjk, line 778: Whilst I can see why `ValidateSubscriptions` is needed, this function is just doing what other exchanges do inside the templates.
The template should call a func for interval and level fields, and it should panic on invalid values.
So that'd mean that all Interval logic would be in one template func, and all levels in the other.

Am I missing something?

That leaves just rejecting attempts to subscribe to Margin at all.
Are we saying that websockets are enabled for them but subscriptions are not?
Cos otherwise, we should just set the asset to not websocket supported 🙂

exchanges/subscription/template.go
```diff
@@ -69,11 +71,26 @@ func (l List) ExpandTemplates(e IExchange) (List, error) {
...
-			subs = append(subs, expanded...)
+			continue
+		}
+
+		subs = append(subs, expanded...)
+	}
+
+	// Validate the subscriptions after expansion to capture fields that will be used in the template
+	if v, ok := e.(ListValidator); ok {
+		if err := v.ValidateSubscriptions(subs); err != nil {
```
gbjk, line 92: 🚧 
This is only validating the ExpandedTemplates here.
Wouldn't it need to be done at a higher level, when we're subscribing, to also check existing subs that are already subscribed?

gbjk (commented):
Good work 👍 

Initial feedback on a first pass. Some in-depth comments on gateio_websocket, but initially just feedback on the overall direction to understand the limitations we're working against.

Thanks

exchanges/gateio/gateio_websocket.go
shazbert: Yeah good point will remove

exchanges/gateio/gateio_websocket.go
shazbert: That error was to do with incremental updates and the fact that zero values came through which is needed because because it updates the sequence numbers. When multiple book channels are subscribed to though, this throws the internal book completely out.

exchanges/gateio/gateio_websocket.go
shazbert: [fixed for templates](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/d8cca4e2b5db4ef3800c7e285e3ee56772a8a070)

Margin and cross margin are just skipped and return nil so I don't think they should be an issue, they share the same books with spot but I don't know how its going to look once all assets are updated to the new template. Would prefer to leave them for now as it seems out of scope.

exchanges/gateio/gateio_websocket.go
```diff
@@ -634,6 +633,13 @@ func (g *Gateio) GetSubscriptionTemplate(_ *subscription.Subscription) (*templat
 			"channelName":         channelName,
 			"singleSymbolChannel": singleSymbolChannel,
 			"interval":            g.GetIntervalString,
```
gbjk, line 635: 🚨 Don't think we can do this. It's return signature isn't acceptable for a template.
I'll include it in proposed fixes.

exchanges/gateio/gateio_websocket.go
```diff
@@ -774,101 +779,106 @@ func isRestrictedOrderbookChannel(name string) bool {
 	return false
 }
 
-// ValidateSubscription implements subscription.SubscriptionValidator interface
-func (g *Gateio) ValidateSubscription(s *subscription.Subscription) error {
-	channelName := channelName(s)
-	switch s.Asset {
+func ensureOrderbook(channel string, a asset.Item, inString string, levels int) error {
+	inDura, err := time.ParseDuration(inString)
```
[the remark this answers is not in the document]
gbjk: I'll provide a patch/commit to share suggested simplification

gbjk (changes requested):
Definitely in the right direction.

Quick feedback, but I'll provide a patch shortly as well.

exchanges/gateio/gateio_websocket.go
```diff
@@ -764,11 +739,157 @@ func singleSymbolChannel(name string) bool {
...
+	}
+	return nil
+}
+
 const subTplText = `
 {{- with $name := channelName $.S }}
 	{{- range $asset, $pairs := $.AssetPairs }}
 		{{- if singleSymbolChannel $name }}
 			{{- range $i, $p := $pairs -}}
+				{{- $err := ensureOrderbook $name $asset (interval $.S.Interval) $.S.Levels }}
```
gbjk, line 891: Just FYI: If we were to do this, we'd use `with` in text/template.

Also: We're turning s.Interval into a string in order to turn it back into duration again 🤷 

No action required, including it in proposed changes.

exchanges/gateio/gateio_websocket.go
shazbert: It worked when I tested it 🤷 😆

gbjk: I think that's because an error of nil would not do anything.
But if the method errored, I think that error would (kinda silently) end up in the template.

exchanges/gateio/gateio_websocket.go
```diff
@@ -764,11 +739,157 @@ func singleSymbolChannel(name string) bool {
...
+				if levels != 20 {
+					if interval == kline.TwentyMilliseconds {
+						return fmt.Errorf("%w for `%s` only `20` levels supports `%s`", subscription.ErrInvalidInterval, channel, kline.TwentyMilliseconds)
+					}
+					if interval != kline.HundredMilliseconds {
+						return fmt.Errorf("%w for `%s` only `%s` is supported", subscription.ErrInvalidInterval, channel, kline.HundredMilliseconds)
+					}
+				} else if interval != kline.HundredMilliseconds && interval != kline.TwentyMilliseconds {
+					return fmt.Errorf("%w for `%s` only `%s` or `%s` is supported", subscription.ErrInvalidInterval, channel, kline.HundredMilliseconds, kline.TwentyMilliseconds)
+				}
```
gbjk, line 845: I'm confused by this line, because ... the template won't output an interval for `futuresOrderbookUpdateChannel`, will it?

shazbert: No, its ready for when templates do futures as well, so it will scale up without having to worry about it.

exchanges/gateio/gateio_websocket.go
shazbert: It seems like it just escapes in expand templates, any error not handled seems to just terminate template processing and return.

```
[ERROR] | WEBSOCKET | 10/04/2025 09:30:55 | GateIO websocket: subscription failure, orderbook spot : template: master.tmpl:6:46: executing "master.tmpl" at <interval $.S.Interval>: error calling interval: '420ms': interval unsupported by exchange
```

exchanges/gateio/gateio_websocket.go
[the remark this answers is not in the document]
shazbert: I didn't see any patch but I have tried to address your nits [here](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/d805fee2b54b005bd74a308517ea4933715d1115)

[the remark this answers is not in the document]
gbjk: Okay, here's a commit which does the `channelInterval` pattern gbjk@cdaaad6320 and patch:
```
curl https://gist.githubusercontent.com/gbjk/a8af720ff743e14461240bc242786986/raw/ac5cf786cafc235a821dc8114c59a62bc5ae549d/gistfile1.txt | git apply
```

Haven't tested it, but it addresses some naming, reduces code and should work.
The Asset to channel name map will not be necessary at some point, for now it's easiest.

I'm thinking you can duplicate the map for levels, or have it be a map of structs with Levels and Intervals at the bottom level.

HTH 🙏

exchanges/gateio/gateio_websocket.go
gbjk: Oh, cute. [That's actually documented](https://pkg.go.dev/text/template#hdr-Pipelines)

Could probably be using that a lot more!

exchanges/subscription/template.go
gbjk: I don't think we have a good place right now.
We're lacking a cross exchange entrypoint into `Subscribe`, since exchanges have their own implementation of it.
For now, I think the solution is for GateIO's `ValidateSubscriptions` to recast `v` back into `IExchange` to get at the rest of the subscriptions as well as the new `subs`.
That's sub-optimal, but fixes the issue.

I'll bear this problem in mind in #1764

shazbert: Sounds good will give it a crack! 👍

shazbert: Addressed I hope 🙏: https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/fc96cd2a38e98650859ee26dcfb7eb9b6402d718

exchange/websocket/buffer/buffer.go
```diff
@@ -328,6 +328,23 @@ func (w *Orderbook) GetOrderbook(p currency.Pair, a asset.Item) (*orderbook.Base
...
+		return 0, currency.ErrCurrencyPairEmpty
+	}
+	if !a.IsValid() {
+		return 0, asset.ErrInvalidAsset
+	}
+	w.mtx.Lock()
+	defer w.mtx.Unlock()
+	book, ok := w.ob[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	if !ok {
+		return 0, fmt.Errorf("%s %s %s %w", w.exchangeName, p, a, ErrDepthNotFound)
```
gbjk, line 343: 🐒  Can probably ignore, but maybe better.
```suggestion
		return 0, fmt.Errorf("%s %w: %s.%s", w.exchangeName, ErrDepthNotFound, a, p)
```

exchange/websocket/buffer/buffer_test.go
```diff
@@ -742,15 +743,35 @@ func TestGetOrderbook(t *testing.T) {
...
-		b.Pair != ob.Pair {
-		t.Fatal("data on both books should be the same")
-	}
+	require.Equal(t, askLen, len(ob.Asks))
+	require.Equal(t, bidLen, len(ob.Bids))
+	require.Equal(t, b.Asset, ob.Asset)
+	require.Equal(t, b.Exchange, ob.Exchange)
+	require.Equal(t, b.LastUpdateID, ob.LastUpdateID)
+	require.Equal(t, b.PriceDuplication, ob.PriceDuplication)
+	require.Equal(t, b.Pair, ob.Pair)
```
gbjk, line 752: 🐒 These don't feel like `require`, and could ideally do with explanations.
```suggestion
	assert.Equal(t, askLen, len(ob.Asks), "Retrieve should get correct number of Asks")
        ... etc ...
```

exchanges/gateio/gateio_websocket.go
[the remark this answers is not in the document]
gbjk: I'll provide a patch that handles both cases and eradicates the `validateOrderbook` call.

gbjk (commented):
Went through the updates and generally good, but realised that https://github.com/gbjk/gocryptotrader/commit/cdaaad6320 isn't adopted yet.

I'll provide a patch that implements it for both levels and intervals and is fully tested for you

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,172 @@
...
+// SyncOrderbook fetches and synchronises an orderbook snapshot to the limit size so that pending updates can be
+// applied to the orderbook.
+func (c *updateCache) SyncOrderbook(ctx context.Context, g *Gateio, pair currency.Pair, a asset.Item, limit uint64) error {
+	// TODO: When templates are introduced for all assets define channel key and use g.Websocket.GetSubscription(ChannelKey{&subscription.Subscription{Channel: channelName}})
+	// to get the subscription and levels for limit. So that this can scale to config changes. Spot is currently the only
+	// asset with templates but it has only one level.
+	book, err := g.UpdateOrderbookWithLimit(ctx, pair, a, limit)
+
+	c.mtx.Lock() // lock here to prevent ws handle data interference with REST request above
+	defer func() { c.updates = c.updates[:0]; c.updating = false; c.mtx.Unlock() }()
```
gloriousCode, line 109:
```suggestion
	defer func() { 
	   c.updates = c.updates[:0] 
	   c.updating = false
	   c.mtx.Unlock() 
	}()
```
I think is an improvement on readability because its an important piece of this whole function

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return err
+	}
+
+	if a != asset.Spot {
+		if err := g.Websocket.Orderbook.LoadSnapshot(book); err != nil {
+			return err
+		}
+	} else {
+		// Spot, Margin, and Cross Margin books are the same
```
gloriousCode, line 120:
```suggestion
		// Spot, Margin, and Cross Margin books are all classified as spot
```

exchanges/gateio/gateio_wrapper.go
```diff
@@ -180,6 +180,8 @@ func (g *Gateio) SetDefaults() {
 	g.WebsocketResponseMaxLimit = exchange.DefaultWebsocketResponseMaxLimit
 	g.WebsocketResponseCheckTimeout = exchange.DefaultWebsocketResponseCheckTimeout
 	g.WebsocketOrderbookBufferLimit = exchange.DefaultWebsocketOrderbookBufferLimit
+
```
gloriousCode, line 183:
```suggestion
```

gloriousCode (changes requested):
This is a lot of nice code. My issues are mostly about comments. The biggest thing is that I think you have to provide a config update with this PR because otherwise all user's have to fiddle with their configs or they receive errors. I don't think we should be doing that by default

exchanges/gateio/gateio.go
```diff
@@ -322,6 +323,8 @@ func (g *Gateio) GetTicker(ctx context.Context, currencyPair, timezone string) (
 // GetIntervalString returns a string representation of the interval according to the Gateio exchange representation.
 func (g *Gateio) GetIntervalString(interval kline.Interval) (string, error) {
 	switch interval {
+	case kline.TwentyMilliseconds:
+		return "20ms", nil
```
gbjk, line 327: Missing kline.TenMilliseconds.

I'll add it in my patch.

exchanges/gateio/gateio_websocket.go
```diff
@@ -53,20 +54,22 @@ const (
...
 	unsubscribeEvent = "unsubscribe"
+
+	spotOrderbookDepthLimit = 100
 )
 
 var defaultSubscriptions = subscription.List{
 	{Enabled: true, Channel: subscription.TickerChannel, Asset: asset.Spot},
 	{Enabled: true, Channel: subscription.CandlesChannel, Asset: asset.Spot, Interval: kline.FiveMin},
-	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds},
+	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds, Levels: spotOrderbookDepthLimit},
```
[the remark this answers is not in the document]
gbjk: @shazbert I can provide this for you shortly as well 🙂

exchanges/gateio/gateio_websocket.go
```diff
@@ -761,11 +730,144 @@ func singleSymbolChannel(name string) bool {
...
+	}
+	return nil
+}
+
 const subTplText = `
 {{- with $name := channelName $.S }}
 	{{- range $asset, $pairs := $.AssetPairs }}
 		{{- if singleSymbolChannel $name }}
 			{{- range $i, $p := $pairs -}}
+				{{- validateOrderbook $name $asset $.S.Interval $.S.Levels }}
```
gbjk, line 870: 🚧  I propose something like [this](https://github.com/gbjk/gocryptotrader/commit/6d00d19fd5.patch)
```
curl https://github.com/gbjk/gocryptotrader/commit/6d00d19fd5.patch | git apply
```

gbjk, line 870: 🚧  I propose something like [this](https://github.com/gbjk/gocryptotrader/commit/6d00d19fd5.patch)
```
curl https://github.com/gbjk/gocryptotrader/commit/6d00d19fd5.patch | git apply
```

gbjk (changes requested):
Provided proposed pattern for avoiding having a "validation line" in the template.
Also fixes 250ms, and some other issues I spotted during testing, as well as adding coverage on getIntervalString

exchanges/gateio/gateio_websocket.go
shazbert: @gbjk THANKS SIRRRRRR!

exchanges/gateio/gateio_websocket.go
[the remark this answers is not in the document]
gbjk: I didn't spot this earlier, but this line shouldn't be changed.
Right now, we don't support setting params (interval/level) which can't be sent to the server.
If an orderbook defaults to 100 levels, _and cannot be changed or set_, then we should not have it in our config.
I completely understand the rationale for having it there. It's just the way it currently is though.
I've been putting it in the exchange docs under Subscriptions heading.

If it's a huge deal, I could deal with this in the template, but given the overall mess and inconsistency with GateIO's asset to orderbook matrix, I'd prefer if we just didn't.

Another non-sendable param like this is the 250ms default interval for one of them.

exchanges/gateio/gateio_websocket.go
gbjk: ~Something like this~
[Here is the real juice](https://github.com/gbjk/gocryptotrader/commit/14b48f8c8f.patch) 🧃
* Tests are passing
* Fixed options orderbook generation
* Removed silent defaults from config, since we can't mix "params that get sent" with those that don't.
* I haven't yet updated docs; I'll do that in my PR upcoming
```
curl https://github.com/gbjk/gocryptotrader/commit/14b48f8c8f.patch | git apply
```
This is the complete patch, so can `git reset --hard HEAD^` or just apply it on top and it will probably jive okay.

This overlaps so heavily with my upcoming work that I don't want to go too far down the same route here on unifying.
I'll steal a bunch of this work when I rebase on top of this, and on top of the split PR.

shazbert: [2d6ebc7](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/2d6ebc7ca81afb36f666f6da2537d61ed6bbbfd2)

Thanks!

exchanges/gateio/gateio_websocket.go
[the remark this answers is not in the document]
shazbert: Yeah that is fair. 👍

exchange/websocket/buffer/buffer_test.go
shazbert: [b737d85](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/b737d85b39f1899add14a76098c52d8b24f55b5f)

Done

exchanges/gateio/gateio_wrapper.go
shazbert: [14258f8](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/14258f88b55aa9861a170843c13908b1d604a34e)

Nits should be done now; thanks!

exchanges/subscription/template.go
```diff
@@ -72,8 +73,28 @@ func (l List) ExpandTemplates(e IExchange) (List, error) {
...
+			return nil, err
+		}
+		// Expand stored subscriptions to ensure we have the same template applied to validate
+		// against the incoming subscriptions. Pass e using IExchange interface to prevent exposure
+		// to ListValidator and avoid recursion.
+		expStoredSubs, err := storedSubs.ExpandTemplates(struct{ IExchange }{e})
+		if err != nil {
+			return nil, err
+		}
+		if err := v.ValidateSubscriptions(append(subs, expStoredSubs...)); err != nil {
```
gbjk, line 96: 🚧 My concern with this is that it shouldn't be the job of ExpandTemplates to expand other subscriptions not passed in.
Was there a use-case for needing this, as opposed to hypothetically ?
If possible, I'd like to just use the subscriptions as they are.
I can't think of a use-case otherwise for how they got into subs without going through expand anyway.
🐒 I think `slices.Concat` is probably wiser because it doesn't imply touching `subs` underlying array
```suggestion
		if err := v.ValidateSubscriptions(slices.Concat(subs, storedSubs)); err != nil {
```

exchanges/subscription/template.go
```diff
@@ -72,8 +73,28 @@ func (l List) ExpandTemplates(e IExchange) (List, error) {
...
+			return nil, fmt.Errorf("validate subscriptions: %w", err)
```
gbjk, line 97: 🐒 
```suggestion
			return nil, fmt.Errorf("error validating subscriptions: %w", err)
```

gbjk (changes requested):
Minor change request that might no-op if it's justified.

exchanges/subscription/template.go
gbjk: I think for now this solution is good.
Overall I don't think it should be in ExpandTemplates at all, but we lack a unified subscription "engine" right now above exchange level.
I'll be adding that soon.

exchanges/subscription/template.go
shazbert: It was mostly hypothetical, and can be seen as redundant. It's really only for when a sub not expanded gets inserted on error or intentionally. It forces consistency between the two in kind lists and doesn't assume anything. It should just skip anyway if its already expanded? If this is a real pinch point I can revert it; I see `ValidateSubscriptions` as being removed from ExpandTemplates eventually and this will force a better standard comparison. 

As for the `slices.Concat` change, this is predicated that subs cap is >= (subs len + storedSubs len) and you want a forced allocation. For this to leak you going to have to change the subs len. I don't think this is necessary even if we write temp to its array. If you are wondering about its memory foot print its already allocated as such until its collected. So if it didn't get validated it would still have the same cap size anyway.

exchanges/subscription/template.go
shazbert: Yeah this is nicer to digest thanks

exchanges/subscription/template.go
shazbert: It's not a hot path anyway so I pushed that slices.Concat change, it won't hurt it. 🤷

exchanges/gateio/gateio_websocket.go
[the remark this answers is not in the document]
gloriousCode: Thank you for removing the error on running. I still take issue with adding new subscriptions and not doing anything to add them to the user's config though. I am more willing to let it slide since its a GCT-wide subscription issue and an issue already exists for it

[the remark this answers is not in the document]
gbjk: @gloriousCode I don't see that we're adding any new _enabled_ subscriptions 🤷 
I don't think we need to go too far out of our way when we add new _disabled_ subscriptions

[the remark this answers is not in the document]
gloriousCode: I know all that, but I can still whinge about the experience of having to copy paste and fiddle with config structs to test/use things

exchanges/gateio/gateio_websocket_delivery_futures.go
```diff
@@ -92,6 +90,9 @@ func (g *Gateio) GenerateDeliveryFuturesDefaultSubscriptions() (subscription.Lis
 				params["interval"] = "0"
```
gloriousCode, line 90: May I request that all the different asset types on `GenerateSubs` that don't support templating have a `// TODO`? Was the only thing that threw me

gloriousCode (changes requested):
Tested and it all seems good to me. I have some unresolved nits left open that weren't addressed, but that's only about line breaks and comments

exchanges/gateio/gateio_websocket_delivery_futures.go
shazbert: Pushed sir 🫡 [dcea496](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/dcea4961b411fe4157914dd8b8011e857f267d5a)

shazbert: Commentary [a2ed470](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/a2ed4709b086c654bf02fe1755ede89b8541ce74)

exchanges/subscription/template.go
shazbert: @gbjk Needs your input when you get a chance 🙇

gbjk: Sorry for delay.
Yes, I think I'm sticking to "please revert".
Less code is better, when it's not necessary.

shazbert: Reverted: [6654440](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/6654440c1adb3f3e08145a16cd206dc4a7494b2b)

gbjk (approved):
LGTM 🎉

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -99,10 +100,26 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
...
-func (c *updateCache) SyncOrderbook(ctx context.Context, g *Gateio, pair currency.Pair, a asset.Item, limit uint64) error {
-	// TODO: When templates are introduced for all assets define channel key and use g.Websocket.GetSubscription(ChannelKey{&subscription.Subscription{Channel: channelName}})
-	// to get the subscription and levels for limit. So that this can scale to config changes. Spot is currently the only
-	// asset with templates but it has only one level.
+func (c *updateCache) SyncOrderbook(ctx context.Context, g *Gateio, pair currency.Pair, a asset.Item) error {
+	// TODO: When templates are introduced for all assets, scale to dynamic supported limit/levels provided from config.
+	var limit uint64
+	switch a {
+	case asset.Spot:
+		sub := g.Websocket.GetSubscription(spotOrderbookUpdateKey)
```
gbjk, line 108: This is super cute. Never thought of it. Really like the pattern.

exchanges/gateio/gateio_websocket.go
```diff
@@ -58,15 +60,15 @@ const (
 var defaultSubscriptions = subscription.List{
 	{Enabled: true, Channel: subscription.TickerChannel, Asset: asset.Spot},
 	{Enabled: true, Channel: subscription.CandlesChannel, Asset: asset.Spot, Interval: kline.FiveMin},
-	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds},
+	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.TwentyMilliseconds},
```
gbjk, line 63: 🚧 Sorry buddy. This will require an config upgrade.
I have 2 branches in flight that improve the ease of doing this, so I can provide a cherry-pick to make this easier.
You can also solo it however you want.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -99,10 +100,26 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
 
 // SyncOrderbook fetches and synchronises an orderbook snapshot to the limit size so that pending updates can be
 // applied to the orderbook.
-func (c *updateCache) SyncOrderbook(ctx context.Context, g *Gateio, pair currency.Pair, a asset.Item, limit uint64) error {
-	// TODO: When templates are introduced for all assets define channel key and use g.Websocket.GetSubscription(ChannelKey{&subscription.Subscription{Channel: channelName}})
-	// to get the subscription and levels for limit. So that this can scale to config changes. Spot is currently the only
-	// asset with templates but it has only one level.
+func (c *updateCache) SyncOrderbook(ctx context.Context, g *Gateio, pair currency.Pair, a asset.Item) error {
+	// TODO: When templates are introduced for all assets, scale to dynamic supported limit/levels provided from config.
```
gbjk, line 104: 🐒 The word dynamic kinda feels wrong when it's actually "configured".
Dynamic normally means more variable than that.
How's this feel:
```suggestion
	// TODO: When subscription config is added for all assets update limits to use sub.Levels
```

exchanges/gateio/ws_ob_update_manager.go
shazbert: ❤️

exchanges/gateio/gateio_websocket.go
shazbert: Reverted change: [c338bc6](https://github.com/thrasher-corp/gocryptotrader/pull/1863/commits/c338bc6a41fe33636b24fa3f3b9e588e2d36be54)

gbjk (approved):
LGTM 🎉

cranktakular (commented):
The cranktakular review has started, but not yet finished.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,209 @@
...
+	m.mtx.RLock()
+	cache, ok := m.m[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	if !ok {
+		m.mtx.RUnlock()
+		m.mtx.Lock()
+		cache = &updateCache{}
+		m.m[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}] = cache
+		m.mtx.Unlock()
+	} else {
+		m.mtx.RUnlock()
```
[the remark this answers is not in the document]
shazbert: Good spotting, shouldn't be an issue. 👍

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,209 @@
...
+
+// LoadCache loads the cache for the given pair and asset. If the cache does not exist, it creates a new one.
+func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCache {
+	m.mtx.RLock()
+	cache, ok := m.m[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	if !ok {
+		m.mtx.RUnlock()
+		m.mtx.Lock()
+		cache = &updateCache{}
+		m.m[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}] = cache
```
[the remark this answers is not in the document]
shazbert: Not really adds a line, in both cases they shouldn't allocate but I haven't tested it.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,209 @@
...
+		limit = deliveryFuturesUpdateLimit
+	case asset.Options:
+		limit = optionOrderbookUpdateLimit
+	}
+
+	book, err := g.UpdateOrderbookWithLimit(ctx, pair, a, limit)
+
+	c.mtx.Lock() // lock here to prevent ws handle data interference with REST request above
+	defer func() {
+		c.updates = c.updates[:0]
```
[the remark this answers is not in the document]
shazbert: Yeah good point, I'm going to just nil it out because there is a slight benefit but nothing that is going to actually impact anything in a meaningful way.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,209 @@
...
+		lastUpdateID, err := g.Websocket.Orderbook.LastUpdateID(data.update.Pair, a)
+		if err != nil {
+			return err
+		}
+		nextID := lastUpdateID + 1
+		if data.firstUpdateID > nextID {
+			return errOrderbookSnapshotOutdated
+		}
+		if data.update.UpdateID < nextID {
+			continue // skip updates that are behind the current orderbook
```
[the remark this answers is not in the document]
shazbert: Nope, it comes up when you buffer the incoming updates to apply to the REST book, some of them can be out of sync.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -0,0 +1,209 @@
...
+		}
+		if err := c.applyOrderbookUpdate(g, data.update); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+// applyOrderbookUpdate applies an orderbook update to the orderbook
+func (c *updateCache) applyOrderbookUpdate(g *Gateio, update *orderbook.Update) error {
```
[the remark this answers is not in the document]
shazbert: I will drop the method receiver but I will keep it just a lonely function so as to not pollute gateio with unnecessary funcs

thrasher- (changes requested):
I left this running whilst doing a glorious dump which took a decent amount of time, on master I'd get errors randomly popping up whereas this PR appears to have fixed it!

Just minor things and needs a rebase

exchanges/gateio/gateio_websocket.go
```diff
@@ -782,20 +750,166 @@ func singleSymbolChannel(name string) bool {
...
+		optionsOrderbookTickerChannel: {},
+		optionsOrderbookUpdateChannel: {5, 10, 20, 50},
+		optionsOrderbookChannel:       {5, 10, 20, 50},
+	},
+}
+
+func channelLevels(s *subscription.Subscription, a asset.Item) (string, error) {
+	cName := channelName(s)
+	assetChannels, ok := channelLevelsMap[a]
+	if !ok {
```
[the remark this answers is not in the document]
shazbert: Thank @gbjk 🚀

exchanges/gateio/gateio_test.go
```diff
@@ -3366,6 +3414,7 @@ func TestOrderbookChannelIntervals(t *testing.T) {
 	assert.Equal(t, "20ms", i)
 
 	for s, exp := range map[*subscription.Subscription]error{
+		{Asset: asset.Binary, Channel: "unhandle_random_ticker", Interval: kline.OneYear}:                            nil,
```
gbjk, line 3417: 🦍 Slightly uneasy about this. Iterated a few ideas before coming up with this.
 ```suggestion
		{Asset: asset.Binary, Channel: "unknown_channel", Interval: kline.OneYear}:                            nil,
```

File: exchanges/gateio/gateio_test.go [hunk 38]
```diff
@@ -3401,7 +3450,6 @@ func TestOrderbookChannelIntervals(t *testing.T) {
 			if exp != nil {
 				require.ErrorIs(t, err, exp)
 			} else {
-				require.NoError(t, err)
 				switch {
```
gbjk, line 3453: 🦍 This is part.2 from above.
```suggestion
				switch {
				case s.Channel == "unknown_channel":
					assert.Empty(t, i, "orderbookChannelInterval should return empty for unknown channels")
```

gbjk (approved):
I'm happy.

However the implicit suffix of `_ticker` having a magical meaning, and `unhandle_random` niggled at me for a moment.

I've offered an explicit case which is clearer.
Would apply to both levels and intervals.

gbjk (approved):

PR #1866, opened 2025-03-28
Huobi: Fix TestPairFromContractExpiryCode on CW

gbjk:
When testing CW against today, we were using a time.Now().Truncate() and documentation says to not do that.

Driveby refactored the tz out of the loop

## Type of change


shazbert (approved):
Thanks; things you learn.

gloriousCode (approved):
tACK! I did not wish to learn that `Truncate` shouldn't be done that way

thrasher- (approved):

PR #1867, opened 2025-03-29
Config: Eliminate shared config version package vars

### LINKED ISSUE #1864: config versions: Eliminate shared package vars
In this instance, pkg vars are not necessary.
In instances where they might be shared enough, I'd probably add it to the type Version5 struct definition and define it in init, probably.
Tests would have to duplicate content. But then often tests should duplicate content unless it's unwieldy.
Another option is methods on Version5 to yield vars.

Suggest this https://github.com/gbjk/gocryptotrader/commit/b729bee449
[Here is a patch](https://gist.githubusercontent.com/gbjk/af1e2013c82c3f635b013c92bff2d0aa/raw/73769626da0da92c9d96e76e89a093d956ae7779/0001-Config-Fix-V5-spraying-pkg-vars-indiscriminately.patch) :
```
curl https://gist.githubusercontent.com/gbjk/af1e2013c82c3f635b013c92bff2d0aa/raw/73769626da0da92c9d96e76e89a093d956ae7779/0001-Config-Fix-V5-spraying-pkg-vars-indiscriminately.patch | git apply
```

_Originally posted by @gbjk in https://github.com/thrasher-corp/gocryptotrader/pull/1861#discussion_r2013303854_

gbjk:
* Fixes shared vars in both runtime and test packages
* Fixes test func names in Version6

Fixes #1864

## Type of change

### DIFF
--- a/config/versions/v2.go
+++ b/config/versions/v2.go
@@ -13,26 +13,21 @@ func init() {
 	Manager.registerVersion(2, &Version2{})
 }
 
-const (
-	from = "GDAX"
-	to   = "CoinbasePro"
-)
-
 // Exchanges returns just GDAX and CoinbasePro
-func (v *Version2) Exchanges() []string { return []string{from, to} }
+func (v *Version2) Exchanges() []string { return []string{"GDAX", "CoinbasePro"} }
 
 // UpgradeExchange will change the exchange name from GDAX to CoinbasePro
 func (v *Version2) UpgradeExchange(_ context.Context, e []byte) ([]byte, error) {
-	if n, err := jsonparser.GetString(e, "name"); err == nil && n == from {
-		return jsonparser.Set(e, []byte(`"`+to+`"`), "name")
+	if n, err := jsonparser.GetString(e, "name"); err == nil && n == "GDAX" {
+		return jsonparser.Set(e, []byte(`"CoinbasePro"`), "name")
 	}
 	return e, nil
 }
 
 // DowngradeExchange will change the exchange name from CoinbasePro to GDAX
 func (v *Version2) DowngradeExchange(_ context.Context, e []byte) ([]byte, error) {
-	if n, err := jsonparser.GetString(e, "name"); err == nil && n == to {
-		return jsonparser.Set(e, []byte(`"`+from+`"`), "name")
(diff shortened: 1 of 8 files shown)


config/versions/v2.go
```diff
@@ -13,26 +13,21 @@ func init() {
 	Manager.registerVersion(2, &Version2{})
 }
 
-const (
-	from = "GDAX"
-	to   = "CoinbasePro"
-)
```
shazbert, line 19: I don't really get why this is a problem?

shazbert (changes requested):
Thanks, seems to have an import issue: 
```
config\versions\v5.go:9:2: no required module provides package github.com/thrasher-corp/gocryptotrader/config/versions/v5; to add it:
        go get github.com/thrasher-corp/gocryptotrader/config/versions/v5
```

config/versions/v2.go
shazbert: I get why this is a problem now @thrasher- thanks.

gbjk:
> Thanks, seems to have an import issue:

Really sorry about that; It got orphaned.

Fixed [5c61dbf4](http://github.com/gbjk/gocryptotrader/commit/5c61dbf4)

shazbert (changes requested):
One linter issue is left and then looks to go. 

Suggestion: Could we just have a separate package for every version so in future we don't have to worry about pollution, we have a common package between them, and init in versions.go for all supported versions.

```
func init() {
	Manager.registerVersion(0, &v0.Version0{})
	Manager.registerVersion(1, &v1.Version1{})
	Manager.registerVersion(2, &v2.Version2{})
	Manager.registerVersion(3, &v3.Version3{})
	Manager.registerVersion(4, &v4.Version4{})
	Manager.registerVersion(5, &v5.Version5{})
	Manager.registerVersion(6, &v6.Version6{})
}
```

gbjk:
> One linter issue is left and then looks to go.

Fixed gbjk@2ea11b40

shazbert (approved):
LGTM

gloriousCode (approved):
tACK! upgraded, downgraded

thrasher- (approved):
ACK, thanks @gbjk !

PR #1885, opened 2025-04-10
cmd/documentation: Add pagination support for contributors

thrasher-:
# PR Description

After wondering why there were duplicates being added to the contributor list in a non-deterministic fashion, I wondered if it was due to an API pagination limit and turns out it was. This PR adds pagination support to the docs tool after several years 😆Default display list is sorted by contributor count.

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Running the documentation tool

## Checklist


shazbert (approved):
changes ACK!

gloriousCode (approved):
tACK

cmd/documentation/documentation.go
```diff
@@ -168,151 +170,26 @@ func main() {
 		if verbose {
 			fmt.Println("Fetching repository contributor list...")
 		}
-		contributors, err = GetContributorList(config.GithubRepo, verbose)
+		contributors, err = GetContributorList(context.TODO(), config.GithubRepo, verbose)
```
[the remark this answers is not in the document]
gbjk: 🐒 Nitpick:
I agree this shouldn't be `TODO`
I don't like hiding a `context.Background()` like this.
And I realised I've bowed to requests to move contexts next to where they're used when I don't think we should.

I'd move a `ctx := context.Background()` up to near the top of main, for use throughout, _even if it's only used once, in an if condition_.
Putting it in a condition, or being lazy with making a context, especially in main, is a beartrap for later devs.

cmd/documentation/documentation.go
```diff
@@ -168,151 +172,26 @@ func main() {
...
 		if err != nil {
 			log.Fatalf("Documentation Generation Tool - GetContributorList error %s",
 				err)
 		}
 
-		// Github API missing contributors
+		// Github API missing/deleted user contributors
```
gbjk, line 181: 🐒 
Now that pagination is fixed, surely it's *just* deleted, and there's no false positives?
```suggestion
		// Github API deleted user contributors
```

Though I think "missing" covers deleted too, fwiw

cmd/documentation/documentation.go
```diff
@@ -469,21 +348,32 @@ func GetTemplateFiles() (*template.Template, error) {
...
-	if err != nil {
-		return nil, err
-	}
+// GetContributorList fetches a list of contributors from the Github API endpoint
+func GetContributorList(ctx context.Context, repo string, verbose bool) ([]Contributor, error) {
+	var contributors []Contributor
+	vals := url.Values{}
+	vals.Set("per_page", defaultGithubAPIPerPageLimit)
+	page := 1
+	for {
```
gbjk, line 357:
```suggestion
```suggestion
	for page := 1; ; page ++ {
```

```

File: cmd/documentation/documentation.go [hunk 4]
```diff
@@ -469,21 +348,32 @@ func GetTemplateFiles() (*template.Template, error) {
...
+		if err := json.Unmarshal(contents, &resp); err != nil {
+			return nil, err
+		}
+
+		contributors = append(contributors, resp...)
+		if len(resp) < 100 {
+			return contributors, nil
+		}
+
+		page++
```
gbjk, line 375:
```suggestion
```

cmd/documentation/documentation.go
```diff
@@ -469,21 +348,32 @@ func GetTemplateFiles() (*template.Template, error) {
...
-	return resp, json.Unmarshal(contents, &resp)
+		contents, err := common.SendHTTPRequest(ctx, http.MethodGet, common.EncodeURLValues(repo+GithubAPIEndpoint, vals), nil, nil, verbose)
+		if err != nil {
+			return nil, err
+		}
+
+		var resp []Contributor
+		if err := json.Unmarshal(contents, &resp); err != nil {
+			return nil, err
+		}
```
gbjk, line 368: 🤷 
If len(contributors) divides equally into your pages, then will we get an error after the last page?
Is there anything other than number to tell us we were done?

cmd/documentation/documentation.go
```diff
@@ -469,21 +348,32 @@ func GetTemplateFiles() (*template.Template, error) {
...
+
+		contributors = append(contributors, resp...)
+		if len(resp) < 100 {
```
gbjk, line 371: 🚧 
```suggestion
		if len(resp) < defaultGithubAPIPerPageLimit {
```

gbjk (approved):
Great work 💪 
One minor construction change.
Assuming you'll get that right 😉  So approving
Happy to see it again, or merge it with changes.

cmd/documentation/documentation.go
thrasher-: Nope they mean different things because we've done PR's where a person wasn't familiar with opening PR's, so did it for them. They're not technically a deleted user but their contributions were missing.

cmd/documentation/documentation.go
thrasher-: Nope, we'll get an empty result on the next page which means the append will be a no-op and break still. There's a header returned which has the max page limit but not high enough ROI to implement since this works

gbjk (approved):

PR #1683, opened 2024-10-17
credentials: Refactor GetCredentials and CheckCredentials

romanornr:
This pull request introduces several important changes to the `exchanges/credentials.go` file, focusing on improving error handling and credential management. The most significant changes include modifying the `CheckCredentials` function to handle different authentication support scenarios, enhancing error messages in `GetCredentials`, and adding a fallback mechanism for default credentials.

### Error Handling Improvements:
* `func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)`: Enhanced the error message for context credentials store type assertion failures to provide more detailed information.

### Credential Management Enhancements:
* `func (b *Base) CheckCredentials(creds *account.Credentials, isContext bool) error`: Reorganized the logic

* `func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)`: small readability change 

^^^ I mean, if copilot says so 
Test do pass

also 
```
if SubAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string); ok {
		b.API.credMu.RLock()
		creds.SubAccount = SubAccountOverride
		b.API.credMu.RUnlock()
	}

```

Is that lock required? 

EDIT: I was experimenting around to understand the code base. 

Will have to edit the commit, it wasn't supposed to return nil. Not sure, maybe IDE did it automatically since I did saw the comment 

My main question was also about the lock being required. (Don't think it will lead to a noticeable performance gain when removed but can) 

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] Unnecessary improvement attempt 
## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [x ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ x] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #1886, opened 2025-04-11
exchanges/credentials: Improve handling and error reporting

romanornr:
Swapped out empty credential returns with more specific errors to make things clearer and avoid weird behavior. Added safeguards to handle nil credentials in WebSocket data processing. This gives us better error handling and more reliable fallback options when managing credentials.

# PR Description: Revenge arc commit 

Old PR: https://github.com/thrasher-corp/gocryptotrader/pull/1683
It was unclear what my goal was and what I was trying to do. Also, I synced upstream and made a new branch to start with a clean slate. I didn't want to fight with git at 5:30 in the morning with git force push and mess up (sorry) 

### Context Credentials Error Handling Change

```diff
                creds := ctxCredStore.Get()
                if err := b.CheckCredentials(creds, true); err != nil {
-                       return creds, fmt.Errorf("context credentials issue: %w", err)
+                       return nil, fmt.Errorf("context credentials issue: %w", err)
                }
                return creds, nil
        }
```

**Before**: When context credentials failed validation, the function returned the invalid credentials and an error.

**After**: When context credentials fail validation, the function now returns `nil` instead of the invalid credentials.

This change helps create a more reliable error-handling pattern.

Before, the function might have returned invalid credentials along with an error, which could lead to some unexpected issues if the calling code didn't check the error carefully. Now, it reliably returns `nil` for credentials whenever there's an error, which makes the API safer UwU

## Default Credentials Error Handling Improvement

```diff
+       // Fallback to default credentials
        creds := b.API.credentials
-       err := b.CheckCredentials(&creds, false)
-       if err != nil {
-               // NOTE: Return empty credentials on error to limit panic on websocket
-               // handling.
-               return &account.Credentials{}, err
+       if err := b.CheckCredentials(&creds, false); err != nil {
+               // Instead of returning empty credentials, return a specific error
+               return nil, fmt.Errorf("invalid default credentials: %w", err)
        }
```

**Before**: 
- Used a separate variable for the error
- Returned empty credentials (`&account.Credentials{}`) on error
- Had a comment about limiting panic in websocket handling

**After**:
- Uses inline error assignment & returns `nil` with a more specific error message
- Added comment for fallback to default credentials
- Removes the workaround comment about websocket panic

This change makes the error handling more consistent with the context credentials case and improves code clarity.

Instead of returning empty credentials, which could be misinterpreted as valid (I am sure I've seen it), it returns `nil` with a clear error message.

##  Sub-Account Override Handling Improvement

```diff
-       subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string)
-       b.API.credMu.RLock()
-       defer b.API.credMu.RUnlock()
-       if ok {
+
+       // Sub account override if set
+       if subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string); ok {
+               b.API.credMu.RLock()
+               defer b.API.credMu.RUnlock()
                creds.SubAccount = subAccountOverride
        }
+
```

**Before**:
- Declared variables outside the if statement
- Always acquired the read lock, even when no subaccount override was present (from what it seemed like)

**After**:
- Moves the variable declaration into the if statement scope
- Only acquires the lock when a subaccount override is 'actually' present
- No lock overhead for the common case (no subaccount override)
- Writers aren't unnecessarily blocked by readers who don't need the lock
(Don't think this was ever a real bottleneck and this theoretical gain is negligible)

## Error Handling Consistency

The patch sets a standard for error handling: return `nil` for invalid credentials instead of empty ones. This consistency enhances API predictability and usability.

Before this patch, there were two different behaviors:
1. For context credentials: Return invalid credentials with an error
2. For default credentials: Return empty credentials with an error

After the patch, both cases return `nil` with an appropriate error message, creating a unified approach.

### Performance Considerations

The change to the subaccount override handling improves performance by only acquiring the read lock when necessary. This could reduce lock contention and improve overall system performance in high-throughput scenarios with many concurrent API requests.

### Websocket Implementation

The websocket code in `exchanges/btcmarkets/btcmarkets_websocket.go` calls `GetCredentials` in two places:

1. In the `signWsReq` method:
```go
func (b *BTCMarkets) signWsReq(r *WsSubscribe) error {
    creds, err := b.GetCredentials(context.TODO())
    if err != nil {
        return err
    }
    // ... use credentials for signing ...
}
```

2. In the `wsHandleData` method when handling order changes:
```go
creds, err := b.GetCredentials(context.TODO())
if err != nil {
    b.Websocket.DataHandler <- order.ClassificationError{
        Exchange: b.Name,
        OrderID:  orderID,
        Err:      err,
    }
}
clientID := ""
if creds != nil {
    clientID = creds.ClientID
}
```

The websocket code already checks whether the returned credentials are `nil` before using them, so it should correctly handle the new behavior. The patch should align with how the websocket code expects credentials to be managed.

I did notice a comment in the original code about "limiting panic on websocket handling" so I guess that was a workaround for some issue.

### Other Exchange Implementations

Other exchange implementations that call `GetCredentials` will need to handle the possibility of receiving `nil` credentials. Most implementations likely already check for errors, but they should also ensure they handle `nil` credentials appropriately.

## Type of change

- [ x] Bug fix (non-breaking change which fixes an issue)

## How has this been tested

- [ x ] go test ./... -race
- [ ] golangci-lint run  (SICK never heard of it. Might check later, it's 5:30am here)
- [ ] Test X  --> Wait never heard about this either 

## Checklist

- [ x ] I have performed a self-review of my own code  → I asked Claude if I am insane and he said no
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool (wait which tool)
- [ x ] My changes generate no new warnings
- [ x  ] New and existing unit tests pass locally and on Github Actions with my changes

### DIFF
--- a/exchanges/btcmarkets/btcmarkets_websocket.go
+++ b/exchanges/btcmarkets/btcmarkets_websocket.go
@@ -328,6 +328,10 @@ func (b *BTCMarkets) wsHandleData(respRaw []byte) error {
 				Err:      err,
 			}
 		}
+		clientID := ""
+		if creds != nil {
+			clientID = creds.ClientID
+		}
 
 		b.Websocket.DataHandler <- &order.Detail{
 			Price:           price,
@@ -335,7 +339,7 @@ func (b *BTCMarkets) wsHandleData(respRaw []byte) error {
 			RemainingAmount: orderData.OpenVolume,
 			Exchange:        b.Name,
 			OrderID:         orderID,
-			ClientID:        creds.ClientID,
+			ClientID:        clientID,
 			Type:            oType,
 			Side:            oSide,
 			Status:          oStatus,
--- a/exchanges/btcmarkets/btcmarkets_websocket.go
+++ b/exchanges/btcmarkets/btcmarkets_websocket.go
@@ -320,16 +320,14 @@ func (b *BTCMarkets) wsHandleData(respRaw []byte) error {
 			}
 		}
 
-		creds, err := b.GetCredentials(context.TODO())
-		if err != nil {
+		clientID := ""
+		if creds, err := b.GetCredentials(context.TODO()); err != nil {
 			b.Websocket.DataHandler <- order.ClassificationError{
 				Exchange: b.Name,
 				OrderID:  orderID,
 				Err:      err,
 			}
-		}
-		clientID := ""
-		if creds != nil {
+		} else if creds != nil {
 			clientID = creds.ClientID
 		}
 
--- a/exchanges/credentials.go
+++ b/exchanges/credentials.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/common/crypto"
 	"github.com/thrasher-corp/gocryptotrader/config"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/account"
@@ -115,9 +116,7 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 	if value != nil {
 		ctxCredStore, ok := value.(*account.ContextCredentialsStore)
 		if !ok {
-			// NOTE: Return empty credentials on error to limit panic on
-			// websocket handling.
-			return nil, fmt.Errorf("context credentials store type assertion failed for %T: %w", value, errContextCredentialsFailure)
+			return nil, common.GetTypeAssertError("*account.ContextCredentialsStore", value)
 		}
 
 		creds := ctxCredStore.Get()
@@ -127,10 +126,11 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 		return creds, nil
 	}
 
-	// Fallback to default credentials
+	// Fallback to exchange loaded credentials
+	b.API.credMu.RLock()
 	creds := b.API.credentials
+	b.API.credMu.RUnlock()
 	if err := b.CheckCredentials(&creds, false); err != nil {
-		// Instead of returning empty credentials, return a specific error
 		return nil, fmt.Errorf("invalid default credentials: %w", err)
 	}
 
--- a/exchanges/credentials.go
+++ b/exchanges/credentials.go
@@ -23,12 +23,11 @@ var (
 	// undertake an authenticated HTTP request.
 	ErrCredentialsAreEmpty = errors.New("credentials are empty")
 	// Errors related to API requirements and failures
-	errRequiresAPIKey            = errors.New("requires API key but default/empty one set")
-	errRequiresAPISecret         = errors.New("requires API secret but default/empty one set")
-	errRequiresAPIPEMKey         = errors.New("requires API PEM key but default/empty one set")
-	errRequiresAPIClientID       = errors.New("requires API Client ID but default/empty one set")
-	errBase64DecodeFailure       = errors.New("base64 decode has failed")
-	errContextCredentialsFailure = errors.New("context credentials type assertion failure")
+	errRequiresAPIKey      = errors.New("requires API key but default/empty one set")
+	errRequiresAPISecret   = errors.New("requires API secret but default/empty one set")
+	errRequiresAPIPEMKey   = errors.New("requires API PEM key but default/empty one set")
+	errRequiresAPIClientID = errors.New("requires API Client ID but default/empty one set")
+	errBase64DecodeFailure = errors.New("base64 decode has failed")
 )
 
 // SetKey sets new key for the default credentials
--- a/exchanges/credentials.go
+++ b/exchanges/credentials.go
@@ -133,7 +133,6 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 		return nil, fmt.Errorf("error checking credentials: %w", err)
 	}
 
-	// Sub account override if set
 	if subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string); ok {
 		creds.SubAccount = subAccountOverride
 	}
--- a/exchanges/credentials.go
+++ b/exchanges/credentials.go
@@ -120,7 +120,7 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 
 		creds := ctxCredStore.Get()
 		if err := b.CheckCredentials(creds, true); err != nil {
-			return nil, fmt.Errorf("context credentials issue: %w", err)
+			return nil, fmt.Errorf("error checking credentials from context: %w", err)
 		}
 		return creds, nil
 	}
--- a/exchanges/credentials.go
+++ b/exchanges/credentials.go
@@ -117,29 +117,30 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 		if !ok {
 			// NOTE: Return empty credentials on error to limit panic on
 			// websocket handling.
-			return &account.Credentials{}, errContextCredentialsFailure
+			return nil, fmt.Errorf("context credentials store type assertion failed for %T: %w", value, errContextCredentialsFailure)
 		}
 
 		creds := ctxCredStore.Get()
 		if err := b.CheckCredentials(creds, true); err != nil {
-			return creds, fmt.Errorf("context credentials issue: %w", err)
+			return nil, fmt.Errorf("context credentials issue: %w", err)
 		}
 		return creds, nil
 	}
 
+	// Fallback to default credentials
 	creds := b.API.credentials
-	err := b.CheckCredentials(&creds, false)
-	if err != nil {
-		// NOTE: Return empty credentials on error to limit panic on websocket
-		// handling.
-		return &account.Credentials{}, err
+	if err := b.CheckCredentials(&creds, false); err != nil {
+		// Instead of returning empty credentials, return a specific error
+		return nil, fmt.Errorf("invalid default credentials: %w", err)
 	}
-	subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string)
-	b.API.credMu.RLock()
-	defer b.API.credMu.RUnlock()
-	if ok {
+
+	// Sub account override if set
+	if subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string); ok {
+		b.API.credMu.RLock()
+		defer b.API.credMu.RUnlock()
 		creds.SubAccount = subAccountOverride
 	}
+
 	return &creds, nil
 }
 
--- a/exchanges/credentials.go
+++ b/exchanges/credentials.go
@@ -136,8 +136,6 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 
 	// Sub account override if set
 	if subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string); ok {
-		b.API.credMu.RLock()
-		defer b.API.credMu.RUnlock()
 		creds.SubAccount = subAccountOverride
 	}
 
--- a/exchanges/credentials.go
+++ b/exchanges/credentials.go
@@ -130,7 +130,7 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 	creds := b.API.credentials
 	b.API.credMu.RUnlock()
 	if err := b.CheckCredentials(&creds, false); err != nil {
-		return nil, fmt.Errorf("invalid default credentials: %w", err)
+		return nil, fmt.Errorf("error checking credentials: %w", err)
 	}
 
 	// Sub account override if set
--- a/exchanges/credentials_test.go
+++ b/exchanges/credentials_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/config"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/account"
 )
@@ -55,8 +56,8 @@ func TestGetCredentials(t *testing.T) {
 
 	ctx = context.WithValue(context.Background(), account.ContextCredentialsFlag, "pewpew")
 	_, err = b.GetCredentials(ctx)
-	if !errors.Is(err, errContextCredentialsFailure) {
-		t.Fatalf("received: %v but expected: %v", err, errContextCredentialsFailure)
+	if !errors.Is(err, common.ErrTypeAssertFailure) {
+		t.Fatalf("received: %v but expected error to wrap: %v", err, common.ErrTypeAssertFailure)
 	}
 
 	b.API.CredentialsValidator.RequiresBase64DecodeSecret = false
--- a/exchanges/credentials_test.go
+++ b/exchanges/credentials_test.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"testing"
 
+	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/config"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/account"
@@ -56,9 +57,7 @@ func TestGetCredentials(t *testing.T) {
 
 	ctx = context.WithValue(context.Background(), account.ContextCredentialsFlag, "pewpew")
 	_, err = b.GetCredentials(ctx)
-	if !errors.Is(err, common.ErrTypeAssertFailure) {
-		t.Fatalf("received: %v but expected error to wrap: %v", err, common.ErrTypeAssertFailure)
-	}
+	require.ErrorIs(t, err, common.ErrTypeAssertFailure)
 
 	b.API.CredentialsValidator.RequiresBase64DecodeSecret = false
 	fullCred := &account.Credentials{
--- a/gctscript/wrappers/gct/gctwrapper_test.go
+++ b/gctscript/wrappers/gct/gctwrapper_test.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"strings"
 	"testing"
 
 	objects "github.com/d5/tengo/v2"
@@ -194,9 +195,10 @@ func TestAccountInfo(t *testing.T) {
 		t.Fatalf("received: %v but expected: %v", err, nil)
 	}
 	rString, _ := objects.ToString(obj)
-	if rString != `error: "Bitstamp REST or Websocket authentication support is not enabled"` {
-		t.Errorf("received: %v but expected: %v",
-			rString, `error: "Bitstamp REST or Websocket authentication support is not enabled"`)
+	if !strings.Contains(rString, "Bitstamp REST or Websocket authentication support is not enabled") {
+		t.Errorf("received: %v but expected to contain: %v",
+			rString,
+			"Bitstamp REST or Websocket authentication support is not enabled")
 	}
 }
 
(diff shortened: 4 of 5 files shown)


romanornr:
If Travis-ci fails, im going to kms when I wake up

exchanges/credentials.go
```diff
@@ -117,29 +117,30 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 		if !ok {
 			// NOTE: Return empty credentials on error to limit panic on
 			// websocket handling.
```
shazbert, line 119: Remove this lying commentary

exchanges/credentials.go
```diff
@@ -117,29 +117,30 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
-			return &account.Credentials{}, errContextCredentialsFailure
+			return nil, fmt.Errorf("context credentials store type assertion failed for %T: %w", value, errContextCredentialsFailure)
```
shazbert, line 120: For type assertions we have a nice happy little helper `common.GetTypeAssertError()` which wraps whats expected and what was received which you can use here.

exchanges/credentials.go
```diff
@@ -117,29 +117,30 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
 
 		creds := ctxCredStore.Get()
 		if err := b.CheckCredentials(creds, true); err != nil {
-			return creds, fmt.Errorf("context credentials issue: %w", err)
+			return nil, fmt.Errorf("context credentials issue: %w", err)
 		}
 		return creds, nil
 	}
 
+	// Fallback to default credentials
```
shazbert, line 130:
```suggestion
	// Fallback to exchange loaded credentials
```

exchanges/credentials.go
```diff
@@ -117,29 +117,30 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
 	creds := b.API.credentials
-	err := b.CheckCredentials(&creds, false)
-	if err != nil {
-		// NOTE: Return empty credentials on error to limit panic on websocket
-		// handling.
-		return &account.Credentials{}, err
+	if err := b.CheckCredentials(&creds, false); err != nil {
+		// Instead of returning empty credentials, return a specific error
```
shazbert, line 133: commentary isn't needed

exchanges/credentials.go
```diff
@@ -117,29 +117,30 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
 	}
-	subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string)
-	b.API.credMu.RLock()
-	defer b.API.credMu.RUnlock()
-	if ok {
+
+	// Sub account override if set
+	if subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string); ok {
+		b.API.credMu.RLock()
+		defer b.API.credMu.RUnlock()
```
shazbert, line 140: This seems like it can be removed completely as the creds is a new allocation without any pointer to `b.API.credentials`. Then above wrap `b.API.credentials` in read locks instead. 

```go
b.API.credMu.RLock()
creds := b.API.credentials
b.API.credMu.RUnlock()
```

gctscript/wrappers/gct/gctwrapper_test.go
```diff
@@ -194,9 +195,10 @@ func TestAccountInfo(t *testing.T) {
 		t.Fatalf("received: %v but expected: %v", err, nil)
 	}
 	rString, _ := objects.ToString(obj)
-	if rString != `error: "Bitstamp REST or Websocket authentication support is not enabled"` {
-		t.Errorf("received: %v but expected: %v",
-			rString, `error: "Bitstamp REST or Websocket authentication support is not enabled"`)
+	if !strings.Contains(rString, "Bitstamp REST or Websocket authentication support is not enabled") {
```
shazbert, line 198: Can fix up these tests if you want just apply this diff but this is optional:

```diff
diff --git a/gctscript/wrappers/gct/gctwrapper_test.go b/gctscript/wrappers/gct/gctwrapper_test.go
index d41afe942..1b467ff19 100644
--- a/gctscript/wrappers/gct/gctwrapper_test.go
+++ b/gctscript/wrappers/gct/gctwrapper_test.go
@@ -7,10 +7,10 @@ import (
        "os"
        "path/filepath"
        "reflect"
-       "strings"
        "testing"

        objects "github.com/d5/tengo/v2"
+       "github.com/stretchr/testify/require"
        "github.com/thrasher-corp/gocryptotrader/common"
        "github.com/thrasher-corp/gocryptotrader/config"
        "github.com/thrasher-corp/gocryptotrader/engine"
@@ -187,19 +187,12 @@ func TestExchangePairs(t *testing.T) {
 func TestAccountInfo(t *testing.T) {
        t.Parallel()
        _, err := gct.ExchangeAccountInfo()
-       if !errors.Is(err, objects.ErrWrongNumArguments) {
-               t.Fatal(err)
-       }
+       require.ErrorIs(t, err, objects.ErrWrongNumArguments)
        obj, err := gct.ExchangeAccountInfo(ctx, exch, assetType)
-       if err != nil {
-               t.Fatalf("received: %v but expected: %v", err, nil)
-       }
-       rString, _ := objects.ToString(obj)
-       if !strings.Contains(rString, "Bitstamp REST or Websocket authentication support is not enabled") {
-               t.Errorf("received: %v but expected to contain: %v",
-                       rString,
-                       "Bitstamp REST or Websocket authentication support is not enabled")
-       }
+       require.NoError(t, err)
+       rString, ok := objects.ToString(obj)
+       require.True(t, ok)
+       require.Contains(t, rString, "Bitstamp REST or Websocket authentication support is not enabled")
 }

 func TestExchangeOrderQuery(t *testing.T) {
@@ -231,9 +224,7 @@ func TestExchangeOrderCancel(t *testing.T) {
 func TestExchangeOrderSubmit(t *testing.T) {
        t.Parallel()
        _, err := gct.ExchangeOrderSubmit()
-       if !errors.Is(err, objects.ErrWrongNumArguments) {
-               t.Fatal(err)
-       }
+       require.ErrorIs(t, err, objects.ErrWrongNumArguments)

        orderSide := &objects.String{Value: "ASK"}
        orderType := &objects.String{Value: "LIMIT"}
@@ -251,16 +242,11 @@ func TestExchangeOrderSubmit(t *testing.T) {
                orderAmount,
                orderID,
                orderAsset)
-       if err != nil {
-               t.Fatalf("received: %v but expected: %v", err, nil)
-       }
+       require.NoError(t, err)

-       rString, _ := objects.ToString(obj)
-       if !strings.Contains(rString, "Bitstamp REST or Websocket authentication support is not enabled") {
-               t.Errorf("received: %v but expected to contain: %v",
-                       rString,
-                       "Bitstamp REST or Websocket authentication support is not enabled")
-       }
+       rString, ok := objects.ToString(obj)
+       require.True(t, ok)
+       require.Contains(t, rString, "Bitstamp REST or Websocket authentication support is not enabled")
 }

 func TestAllModuleNames(t *testing.T) {
```

exchanges/btcmarkets/btcmarkets_websocket.go
```diff
@@ -328,14 +328,18 @@ func (b *BTCMarkets) wsHandleData(respRaw []byte) error {
 				Err:      err,
 			}
 		}
+		clientID := ""
```
shazbert, line 331: This also needs to be done in CoinbasePro for one item.

shazbert (changes requested):
Thanks for the update, just some minor change requests. 🚀 Please don't kys 🙏

exchanges/credentials.go
romanornr: Totally missed this one, didn't think about that. Great suggestion, just applied

gctscript/wrappers/gct/gctwrapper_test.go
romanornr: I just applied your earlier suggestions, including the coinbase, which I have now done correctly (so I believe)
If that's correct, I will just git apply this patch. Def looks cleaner by using stretchr/testify

Greenlight when the new commit is fine, then I will do this patch. (Prefer to take it step by step)

exchanges/credentials.go
shazbert: You can remove this now:

```go
b.API.credMu.RLock()
defer b.API.credMu.RUnlock()
```

gctscript/wrappers/gct/gctwrapper_test.go
shazbert: Your other changes look good, thanks legend! 🚀

shazbert (changes requested):
Initial changes look good, can commence with the other suggestion and a the removal of the unnecessary lock in credentials.go. 🚀

exchanges/credentials.go
```diff
@@ -115,31 +116,31 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
```
shazbert, line 140: These can be removed now:

```go
b.API.credMu.RLock()
defer b.API.credMu.RUnlock()
```

In `if subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string); ok {`

Then looks good to go.

shazbert (changes requested):
One last thing from me. Thanks!

exchanges/credentials.go
shazbert: Yeah exactly 👍

shazbert (approved):
Well done LGTM.

exchanges/credentials_test.go
```diff
@@ -55,8 +56,8 @@ func TestGetCredentials(t *testing.T) {
 
 	ctx = context.WithValue(context.Background(), account.ContextCredentialsFlag, "pewpew")
 	_, err = b.GetCredentials(ctx)
-	if !errors.Is(err, errContextCredentialsFailure) {
-		t.Fatalf("received: %v but expected: %v", err, errContextCredentialsFailure)
+	if !errors.Is(err, common.ErrTypeAssertFailure) {
+		t.Fatalf("received: %v but expected error to wrap: %v", err, common.ErrTypeAssertFailure)
```
gloriousCode, line 60: If you ever see this before merge:
```suggestion
        require.ErrorIs(t, err, common.ErrTypeAssertFailure)
```

exchanges/credentials.go
```diff
@@ -115,31 +116,29 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
 	if value != nil {
 		ctxCredStore, ok := value.(*account.ContextCredentialsStore)
 		if !ok {
-			// NOTE: Return empty credentials on error to limit panic on
-			// websocket handling.
-			return &account.Credentials{}, errContextCredentialsFailure
```
gloriousCode, line 120: You could remove the unused `errContextCredentialsFailure`

gloriousCode (approved):
tACK!

exchanges/credentials.go
```diff
@@ -115,31 +115,29 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
+	b.API.credMu.RLock()
 	creds := b.API.credentials
-	err := b.CheckCredentials(&creds, false)
-	if err != nil {
-		// NOTE: Return empty credentials on error to limit panic on websocket
-		// handling.
-		return &account.Credentials{}, err
+	b.API.credMu.RUnlock()
+	if err := b.CheckCredentials(&creds, false); err != nil {
+		return nil, fmt.Errorf("invalid default credentials: %w", err)
```
thrasher-, line 133:
```suggestion
		return nil, fmt.Errorf("error checking credentials: %w", err)
```

exchanges/credentials.go
```diff
@@ -115,31 +115,29 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
-			// NOTE: Return empty credentials on error to limit panic on
-			// websocket handling.
-			return &account.Credentials{}, errContextCredentialsFailure
+			return nil, common.GetTypeAssertError("*account.ContextCredentialsStore", value)
 		}
 
 		creds := ctxCredStore.Get()
 		if err := b.CheckCredentials(creds, true); err != nil {
-			return creds, fmt.Errorf("context credentials issue: %w", err)
+			return nil, fmt.Errorf("context credentials issue: %w", err)
```
thrasher-, line 123:
```suggestion
			return nil, fmt.Errorf("error checking context credentials: %w", err)
```

thrasher- (changes requested):
Some nits for consistency, thanks Romano (Aka M to the 🐝) 
image

exchanges/btcmarkets/btcmarkets_websocket.go
```diff
@@ -328,14 +328,18 @@ func (b *BTCMarkets) wsHandleData(respRaw []byte) error {
...
+		if creds != nil {
+			clientID = creds.ClientID
+		}
```
gbjk, line 334: 🚧 Change Request:
I don't think we shouldn't be touching creds at all if err isn't nil.
If err is nil, creds shouldn't be nil, right?

So I'd suggest this: gbjk@ff7d1017c4
You can cherry-pick it and squash into the fixup directly, I expect.
```
diff --git a/exchanges/btcmarkets/btcmarkets_websocket.go b/exchanges/btcmarkets/btcmarkets_websocket.go
index 08a75e8e89..10fe45ad75 100644
--- a/exchanges/btcmarkets/btcmarkets_websocket.go
+++ b/exchanges/btcmarkets/btcmarkets_websocket.go
@@ -320,16 +320,14 @@ func (b *BTCMarkets) wsHandleData(respRaw []byte) error {
 			}
 		}
 
-		creds, err := b.GetCredentials(context.TODO())
-		if err != nil {
+		clientID := ""
+		if creds, err := b.GetCredentials(context.TODO()); err != nil {
 			b.Websocket.DataHandler <- order.ClassificationError{
 				Exchange: b.Name,
 				OrderID:  orderID,
 				Err:      err,
 			}
-		}
-		clientID := ""
-		if creds != nil {
+		} else if creds != nil {
 			clientID = creds.ClientID
 		}
 ```

File: exchanges/credentials.go (hunk 12 above)
Comment (gbjk) [reply exchanges/credentials.go:123]: 🐒 Optional Nitpick
Same as thrasher, but I think that "context credentials" is unnecessarily vague in the error, even though the term is somewhat prolific.
I think when we're throwing blame about it, we can be clearer.
If accepted, this applies to other errors nearby too.
```suggestion
			return nil, fmt.Errorf("error checking credentials from context: %w", err)
```

exchanges/credentials.go
```diff
@@ -115,31 +115,29 @@ func (b *Base) GetCredentials(ctx context.Context) (*account.Credentials, error)
...
+	b.API.credMu.RUnlock()
+	if err := b.CheckCredentials(&creds, false); err != nil {
+		return nil, fmt.Errorf("invalid default credentials: %w", err)
 	}
-	subAccountOverride, ok := ctx.Value(account.ContextSubAccountFlag).(string)
-	b.API.credMu.RLock()
-	defer b.API.credMu.RUnlock()
-	if ok {
+
+	// Sub account override if set
```
gbjk, line 136: 🐒 Nitpick: This comment adds no information we don't already have
Best rule is to use comments to explain either:
* Why are we doing something (as opposed to what are we doing)
* What or How we're doing something if it's too complicated to read immediately as code
In this case, the var name and `ok` give us everything we needed to know.
```suggestion
	// Sub account override if set
```

gctscript/wrappers/gct/gctwrapper_test.go
```diff
@@ -186,18 +187,12 @@ func TestExchangePairs(t *testing.T) {
 func TestAccountInfo(t *testing.T) {
```
gbjk, line 187: 🐒 Nitpick, since you're inside it already
```suggestion
func TestExchangeAccountInfo(t *testing.T) {
```

gctscript/wrappers/gct/gctwrapper_test.go
```diff
@@ -186,18 +187,12 @@ func TestExchangePairs(t *testing.T) {
...
-		t.Fatalf("received: %v but expected: %v", err, nil)
-	}
-	rString, _ := objects.ToString(obj)
-	if rString != `error: "Bitstamp REST or Websocket authentication support is not enabled"` {
-		t.Errorf("received: %v but expected: %v",
-			rString, `error: "Bitstamp REST or Websocket authentication support is not enabled"`)
-	}
+	require.NoError(t, err)
+	rString, ok := objects.ToString(obj)
+	require.True(t, ok)
```
gbjk, line 194: 🐒 Errors which aren't self evident should have comments.
Normal Criteria for not needing a comment:
* Testing the function named in the unit tests
* `NoError`, `NotEmpty` or `ErrorIs` against the return value of that function
* Most other things would benefit from giving context for future maintainers 🙏 
```suggestion
	require.True(t, ok, "ExchangeAccount return value must return correctly from objects.ToString")
```

gctscript/wrappers/gct/gctwrapper_test.go
```diff
@@ -249,16 +242,11 @@ func TestExchangeOrderSubmit(t *testing.T) {
...
+	require.NoError(t, err)
 
-	rString, _ := objects.ToString(obj)
-	if rString != `error: "Bitstamp REST or Websocket authentication support is not enabled"` {
-		t.Errorf("received: [%v] but expected: %v",
-			rString,
-			`error: "Bitstamp REST or Websocket authentication support is not enabled"`)
-	}
+	rString, ok := objects.ToString(obj)
+	require.True(t, ok)
```
gbjk, line 248: 🌴🐒 This is the same nitpick monkey as above. It just moved trees to repeat itself 😄

gbjk (changes requested):
Great work! 👍 Thanks 🙏 

One 🚧 change request and a bunch of 🐒 opinionated pedantry you can accept or ignore at your discretion.

gctscript/wrappers/gct/gctwrapper_test.go
romanornr: "ExchangeAccount return value must return correctly from objects.ToString"
or "ExchangeAccount**Info** return value must return correctly from objects.ToString"

exchanges/btcmarkets/btcmarkets_websocket.go
gbjk: Thanks for applying this change.

gctscript/wrappers/gct/gctwrapper_test.go
gbjk: Sorry for the delay replying.
You're right, the later: `"ExchangeAccountInfo return value must return correctly from objects.ToString"`
Also spotting that the func itself should be renamed to match: `TestExchangeAccountInfo`

exchanges/credentials.go
shazbert: @romanornr this is still pending

shazbert (changes requested):
Thanks for the changes Romano, there is one pending nit request from @thrasher- which needs to be resolved, there are optional nits from @gbjk which can be actioned or not, please notify on what you want to do and we can then merge this. Thanks again for all your work. 🚀

gctscript/wrappers/gct/gctwrapper_test.go
```diff
@@ -249,16 +242,11 @@ func TestExchangeOrderSubmit(t *testing.T) {
...
+	require.NoError(t, err)
 
-	rString, _ := objects.ToString(obj)
-	if rString != `error: "Bitstamp REST or Websocket authentication support is not enabled"` {
-		t.Errorf("received: [%v] but expected: %v",
-			rString,
-			`error: "Bitstamp REST or Websocket authentication support is not enabled"`)
-	}
+	rString, ok := objects.ToString(obj)
+	require.True(t, ok, "OrderSubmit return value must return correctly from objects.ToString")
```
thrasher-, line 248:
```suggestion
	require.True(t, ok, "ExchangeOrderSubmit return value must return correctly from objects.ToString")
```

gbjk (approved):
Thanks 🙏 
All Nitpicks approved too 🐒 ✅

shazbert (approved):
Thanks Romano! ch Ack!

thrasher-:
Thank you @romanornr !
image

PR #1887, opened 2025-04-11
Config: Refactor version packages

gbjk:
* Migrate versions to being encapsulated in their own packages.
Before we end up with too many versions, this changes the pattern to create encapsulation.
We couldn't do this in my earlier patterns due to import cycles, but now that's not an issue.
* All of the tests are moved to blackbox tests so they can make sure the version is registered.
* Renamed test functions to better match the methods they are testing
* Had to lose a shared error, and it didn't seem worth localising

Thanks to @shazbert for the suggestion 🙏 

## Type of change

- [x] Refactor


config/versions/testutils/testutils.go
```diff
@@ -0,0 +1,27 @@
...
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/config/versions"
+)
+
+// RequireVersionRegistered is a test helper which asserts the package calling it has a version registered in versions.Manager.
+// Expects to be called from a test function in a package path matching /versions/vN/
+func RequireVersionRegistered(t *testing.T) {
```
shazbert, line 16: 👻 🪄 Spooky magic 🚀

config/versions/v1/v1.go
```diff
@@ -1,26 +1,21 @@
...
-
-func init() {
-	Manager.registerVersion(1, &Version1{})
-}
+// Version is an ExchangeVersion to upgrade currency pair format for exchanges
+type Version struct{}
 
 // Exchanges returns all exchanges: "*"
-func (v *Version1) Exchanges() []string { return []string{"*"} }
+func (v *Version) Exchanges() []string { return []string{"*"} }
```
shazbert, line 15: You dont need the method receiver across these methods in the versions can just be `(v *Version)` -> `(Version)`

config/versions/v0/v0.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"context"
+)
+
+// Version is a baseline version with no changes, so we can downgrade back to nothing
+// It does not implement any upgrade interfaces
+type Version struct{}
+
+// UpgradeConfig is an empty stub
+func (v *Version) UpgradeConfig(_ context.Context, j []byte) ([]byte, error) {
```
shazbert, line 12: missing coverage on this file

config/versions/v3/v3_test.go
```diff
@@ -1,42 +1,48 @@
-package versions
+package v3_test
```
shazbert, line 1: again with the magic

config/versions/versions.go
```diff
@@ -220,6 +219,14 @@ func (m *manager) registerVersion(ver int, v any) {
 	m.versions[ver] = v
 }
 
+// Version returns a version registered by init or nil if nothing has been registered with that version number
+func (m *manager) Version(version int) any {
```
shazbert, line 223: Suggestion/look away: int -> uint64

shazbert (changes requested):
Nice!

config/versions/v1/v1.go
gbjk: Yep.

I'm going to leave it as a pointer method, though.

Fixed https://github.com/gbjk/gocryptotrader/commit/4fec6cd435

config/versions/v0/v0.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/340a134a8a

config/versions/v3/v3_test.go
gbjk: Blackbox testing FTW 🎉

config/versions/versions.go
gbjk: This would cause overflow issues to mitigate... but fortunately, it actually *should* be `uint16`, which would not.

Fixed https://github.com/gbjk/gocryptotrader/commit/e2e552016d

shazbert (approved):
Ch ACK! 
image ALOT!

gbjk:
🔂  Squashed with no differences afted `szrc`

gloriousCode (changes requested):

config/versions/v3/v3_test.go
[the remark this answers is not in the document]
gbjk: Also: by "these" and "this", you mean blackbox testing?
What is wrong with doing blackbox testing?
IMO: We should be doing it more, not less.

config/versions/testutils/testutils.go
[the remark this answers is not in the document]
gbjk: Okay. I can look at moving from a test helper to a tree walker. I really don't like tree walking in tests because it's not idiomatic Go.

I don't think we could have an auto-register function, without embracing a massive PITA plugin archtecture.
I looked at it and walked away due to complexity.
If a package isn't imported, it's not going to be compiled, and only a plugin architecture could change that.
I also looked heavily at auto-registering by introspecting types in versions package.
It was all too messy, and manually registering simplified everything.

config/versions/versions_test.go
```diff
@@ -114,14 +118,21 @@ func TestLatest(t *testing.T) {
...
 	v, err = m.latest()
 	require.NoError(t, err)
 	assert.Equal(t, uint16(2), v)
 }
+
+func TestVersion(t *testing.T) {
+	assert.Nil(t, Manager.Version(math.MaxUint16))
+	l, err := Manager.latest()
+	require.NoError(t, err, "latest must not error")
+	assert.NotNil(t, Manager.Version(l-1))
```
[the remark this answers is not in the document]
gbjk: > Exposes assumption in test that there was always more than 1 registered version

Sure. If that assumption was wrong, the test would fail.
That assumption feels safe; v0 was baked in for this reason.
I didn't think relying on init was an issue, either.

I'll change it anyway for you though.

config/versions/versions.go
```diff
@@ -211,15 +210,23 @@ func exchangeDeploy(ctx context.Context, patch ExchangeVersion, method func(Exch
...
-	if ver >= len(m.versions) {
-		m.versions = slices.Grow(m.versions, ver+1)[:ver+1]
+	if int(ver) >= len(m.versions) {
+		m.versions = slices.Grow(m.versions, int(ver+1))[:ver+1]
 	}
 	m.versions[ver] = v
 }
 
+// Version returns a version registered by init or nil if nothing has been registered with that version number
+func (m *manager) Version(version uint16) any {
```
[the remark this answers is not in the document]
gbjk: Completely agree. Thanks for spotting.

Fixed https://github.com/gbjk/gocryptotrader/commit/c61aaa1b73

config/versions/versions_test.go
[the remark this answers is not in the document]
gbjk: I agree completely with the change; Thank you 🙏 
Test shouldn't use singletons if they're not testing the singleton's state itself.
And I *really* hate singletons. Only reason I'm okay with `Manager` is that it's lifecycle doesn't pervade runtime.

Fixed https://github.com/gbjk/gocryptotrader/commit/8e0c047474

config/versions/testutils/testutils.go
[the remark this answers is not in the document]
gbjk: > I really don't like tree walking in tests because it's not idiomatic Go.

And yet I already did it with LICENSE checking 🙄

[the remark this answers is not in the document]
gloriousCode: Let's pretend I didn't mention auto-registration 😄 It was a more flippant addition where "if you can already check that each interface implementation is registered, why not register?"

I can think more on it. I just feel like if someone adds a version, they already must register, if they already must register to use it, the purpose of the test is void. It does protect against someone like me who could just remove a registration, but then we're getting into weird territory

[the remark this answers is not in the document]
gbjk: > I just feel like if someone adds a version, they already must register, if they already must register to use it, the purpose of the test is void

This is to help prevent them from not registering it, it passing local tests, but them never realising it's not being run at runtime.
It'd probably get caught during PR manual testing. Probably.
This is just to make sure it's not possible to be missed, like you said.

config/versions/register_test.go
```diff
@@ -0,0 +1,49 @@
...
+
+	err = filepath.WalkDir(versionsDir, func(p string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if !d.IsDir() || p == versionsDir {
+			return nil
+		}
+		verStr := filepath.Base(p)
+		verMatch := regexp.MustCompile(`v(\d+)`).FindStringSubmatch(verStr)
```
shazbert, line 35: suggestion: if somebody starts doing v1.1 folder we might be able to screen that.

shazbert (approved):
Changes ACK, 2 suggestions.

config/versions/register_test.go
gbjk: Sorry, what does "might be able to screen that" mean?
You mean: If someone uses an incorrect version number format, we might be able to error on it?

config/versions/v0/v0_test.go
```diff
@@ -0,0 +1,26 @@
+package v0_test
```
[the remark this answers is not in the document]
gbjk: Unless there's a reason to switch to whitebox testing, when we don't have to, I'd prefer not to.
Opened a direct dialogue.

config/versions/register_test.go
shazbert: Yeah exactly like somebody might introduce a minor version convention, it's pretty low on the probability though.

config/versions/testutils/testutils.go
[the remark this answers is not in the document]
gloriousCode:
```
register_test.go:43: 
        	Error Trace:	/Users/scottg/go/src/github.com/thrasher-corp/gocryptotrader/config/versions/register_test.go:43
        	Error:      	Expected value not to be nil.
        	Test:       	TestVersionsRegistered/v10
        	Messages:   	version.Manager init must register this version
```
This is exactly what I wanted through all my words, this is good. Thank you

config/versions/versions_test.go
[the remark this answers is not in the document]
gloriousCode: Yeah not really a valid point in terms of the exposed assumption. It actually caught me off-guard when writing my suggestion instead.

config/versions/v3/v3_test.go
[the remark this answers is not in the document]
gloriousCode: I was trying to say, if everything is exported, then blackbox testing here has no impact. Though the original test necessitated having the `_test` package. So now it can be either blackboxed or whiteboxed. I have no issue with blackbox testing and look forward to seeing more use of it

config/versions/register_test.go
gloriousCode: This is where I like tests that enforce design. Gk's test ensures noone can release v13.37 and make people sad

gloriousCode (approved):
tACK! I'm happy. Thank you for making those changes

config/versions/register_test.go
shazbert: No I didn't know what I was talking about with this, I don't know how I tested it but when I checked again properly it all errored correctly. 🤷 I obviously am being a silly billy and will resolve this.

config/versions/register_test.go
```diff
@@ -0,0 +1,49 @@
...
+		if len(verMatch) != 2 {
+			return filepath.SkipDir
+		}
+		t.Run(verStr, func(t *testing.T) {
+			version, err := strconv.ParseUint(verMatch[1], 10, 16)
+			require.NoError(t, err, "verMatch must Atoi well")
```
thrasher-, line 41:
```suggestion
			require.NoError(t, err, "verMatch must ParseUint well")
```

config/versions/versions.go
```diff
@@ -211,15 +210,25 @@ func exchangeDeploy(ctx context.Context, patch ExchangeVersion, method func(Exch
...
-func (m *manager) registerVersion(ver int, v any) {
+func (m *manager) registerVersion(ver uint16, v any) {
 	m.m.Lock()
 	defer m.m.Unlock()
-	if ver >= len(m.versions) {
-		m.versions = slices.Grow(m.versions, ver+1)[:ver+1]
+	if int(ver) >= len(m.versions) {
+		m.versions = slices.Grow(m.versions, int(ver+1))[:ver+1]
 	}
 	m.versions[ver] = v
```
thrasher-, line 219: What about a check for version number collisions too?

thrasher- (changes requested):
Two things, then LGTM

config/versions/register_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/733735e253

config/versions/versions.go
gbjk: Used a panic here, because it's significantly more convenient and it'd be a startup issue for a dev anyway.

Fixed https://github.com/gbjk/gocryptotrader/commit/771c689564

thrasher-: Nice, that's what I had in mind too 👍

thrasher- (approved):

PR #1895, opened 2025-04-29
Tests: Use currency.NewBTCUSD and NewBTCUSDT

gbjk:
Simple refactor to use the provided shortcut methods

## Type of change

- [ ] Refactor

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run


gloriousCode (approved):
tACK!

shazbert (approved):
Approved

Suggestion: make a linter so that you don't have to do this chore again and open PR's just for this.

thrasher- (changes requested):
Need to also update the docs, don't need to do tmpl files since there's none in them:

```
find . -type f -name '*.md' -exec sed -i 's/currency\.NewPair(currency\.BTC, currency\.USD)/currency.NewBTCUSD()/g' {} +
```

gbjk:
> Need to also update the docs, don't need to do tmpl files since there's none in them:

Fixed https://github.com/gbjk/gocryptotrader/commit/e6528343d7

I'm fixup'ing this because it's close enough to the original title.

gbjk:
> Suggestion: make a linter so that you don't have to do this chore again and open PR's just for this.

Done.

I've put this in a new file for misc linters. It's the first one that I feel kinda might live with other simple checks that are't part of main Go tests.
I feel like these would be small tests that don't justify their own entire line entry.

Potentially I'm wrong and things like proto-checks and config-json-lint, codespell could go in there too...

config-versions-lint lives on it's own because it's an easy ignorable-negative right up until merge time, so we don't want to muddy others with it's redishness.

.github/workflows/misc.yml
```diff
@@ -0,0 +1,14 @@
...
+on: [push, pull_request]
+jobs:
+  lint:
+    name: miscellaneous checks
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: Check for currency.NewPair(BTC, USD) used instead of currency.NewBTCUSD
+      run: |
+        set +e
```
thrasher-, line 11: Several improvements can be done here:

```suggestion
      run: |
        grep -r -n -E "currency.NewPair\(currency.BTC, currency.USDT?\)" * || exit 0
        echo "::error::Replace currency.NewPair(BTC, USD*) with currency.NewBTCUSD*()"
        exit 1
```
1) Can use '-n' for line numbers with grep to get more precise output since ATM it would just show the file and matches
2) Can rid the +e as it's not needed
3) Can use the GHA`::error::` macro to replace the current colour eyesore (the colour was green but I'd opt for red on an error and green for success) to be consistent with GHA events

thrasher- (changes requested):
Thanks for fixing those other matches! On top of the nits, there's a linter issue which needs to be resolved too

.github/workflows/misc.yml
gbjk: Yep, that's better, thank you 🙏 

Fixed https://github.com/gbjk/gocryptotrader/commit/7ab3337117

gbjk:
> Thanks for fixing those other matches! On top of the nits, there's a linter issue which needs to be resolved too

Linter Fixed https://github.com/gbjk/gocryptotrader/commit/c0d2937dcc

shazbert (approved):
changes ACK. Thanks!

thrasher- (approved):
Thanks for making those changes v2, just needs a rebase!

gbjk:
🔁 Rebased and 🔂 Squashed

PR #1908, opened 2025-05-13
buffer/orderbook: shift orderbook update logic from buffer package to orderbook package

shazbert:
# PR Description

Buffer package:

* Shift orderbook logic to orderbook package
* Change Orderbook struct mutex to RWMutex to slightly relieve concurrent lookups between connections
* `Flushbuffer` method change to not purge entire map on reconnection
* Change name `FlushOrderbook` -> `InvalidateOrderbook` in line with orderbook call
* General clean

NOTE: Buffer package can be RM'd or renamed and repurposed just for looking up websocket orderbooks. It really doesn't serve any useful purpose. Getting out of order updates shouldn't be occurring anyway and we shouldn't be trying to sort them. But at least the main logic is pulled out for this future update. 

Orderbook package:

* LoadSnapshot method signature change to take in orderbook.Base pointer (clean)
* Change field name `UpdateIDProgression` -> `SkipOutOfOrderLastUpdateID` as it's a little bit more intuitive
* Change `Checksum` function signature to just return a generated checksum from stored orderbook for standardisation of error returns 
* removed redundant `updateEntriesByID` field and is now using Action field
* Most of Update processing and checking now occurs under one lock instead of retrieving that field under a read mutex context checking that then getting the next field. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/rpcserver_test.go
+++ b/engine/rpcserver_test.go
@@ -3357,18 +3357,15 @@ func TestGetOrderbookMovement(t *testing.T) {
 	t.Parallel()
 	em := NewExchangeManager()
 	exch, err := em.NewExchangeByName("binance")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
+
 	exch.SetDefaults()
 	b := exch.GetBase()
 	b.Name = fakeExchangeName
 	b.Enabled = true
 
 	cp, err := currency.NewPairFromString("btc-metal")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	b.CurrencyPairs.Pairs = make(map[asset.Item]*currency.PairStore)
 	b.CurrencyPairs.Pairs[asset.Spot] = &currency.PairStore{
@@ -3418,9 +3415,7 @@ func TestGetOrderbookMovement(t *testing.T) {
 	}
 
 	depth, err := orderbook.DeployDepth(req.Exchange, currency.NewPair(currency.BTC, currency.METAL), asset.Spot)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	bid := []orderbook.Tranche{
 		{Price: 10, Amount: 1},
@@ -3434,10 +3429,8 @@ func TestGetOrderbookMovement(t *testing.T) {
 		{Price: 13, Amount: 1},
 		{Price: 14, Amount: 1},
 	}
-	err = depth.LoadSnapshot(bid, ask, 0, time.Now(), time.Now(), true)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = depth.LoadSnapshot(&orderbook.Base{Bids: bid, Asks: ask, LastUpdated: time.Now(), UpdatePushedAt: time.Now(), RestSnapshot: true})
+	require.NoError(t, err)
 
 	_, err = s.GetOrderbookMovement(t.Context(), req)
 	if err.Error() != "quote amount invalid" {
@@ -3470,18 +3463,15 @@ func TestGetOrderbookAmountByNominal(t *testing.T) {
 	t.Parallel()
 	em := NewExchangeManager()
 	exch, err := em.NewExchangeByName("binance")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
+
 	exch.SetDefaults()
 	b := exch.GetBase()
 	b.Name = fakeExchangeName
 	b.Enabled = true
 
 	cp, err := currency.NewPairFromString("btc-meme")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	b.CurrencyPairs.Pairs = make(map[asset.Item]*currency.PairStore)
 	b.CurrencyPairs.Pairs[asset.Spot] = &currency.PairStore{
@@ -3531,9 +3521,7 @@ func TestGetOrderbookAmountByNominal(t *testing.T) {
 	}
 
 	depth, err := orderbook.DeployDepth(req.Exchange, currency.NewPair(currency.BTC, currency.MEME), asset.Spot)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	bid := []orderbook.Tranche{
 		{Price: 10, Amount: 1},
@@ -3547,10 +3535,8 @@ func TestGetOrderbookAmountByNominal(t *testing.T) {
 		{Price: 13, Amount: 1},
 		{Price: 14, Amount: 1},
 	}
-	err = depth.LoadSnapshot(bid, ask, 0, time.Now(), time.Now(), true)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = depth.LoadSnapshot(&orderbook.Base{Bids: bid, Asks: ask, LastUpdated: time.Now(), UpdatePushedAt: time.Now(), RestSnapshot: true})
+	require.NoError(t, err)
 
 	nominal, err := s.GetOrderbookAmountByNominal(t.Context(), req)
 	if !errors.Is(err, nil) {
@@ -3576,18 +3562,15 @@ func TestGetOrderbookAmountByImpact(t *testing.T) {
 	t.Parallel()
 	em := NewExchangeManager()
 	exch, err := em.NewExchangeByName("binance")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
+
 	exch.SetDefaults()
 	b := exch.GetBase()
 	b.Name = fakeExchangeName
 	b.Enabled = true
 
 	cp, err := currency.NewPairFromString("btc-mad")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	b.CurrencyPairs.Pairs = make(map[asset.Item]*currency.PairStore)
 	b.CurrencyPairs.Pairs[asset.Spot] = &currency.PairStore{
@@ -3637,9 +3620,7 @@ func TestGetOrderbookAmountByImpact(t *testing.T) {
 	}
 
 	depth, err := orderbook.DeployDepth(req.Exchange, currency.NewPair(currency.BTC, currency.MAD), asset.Spot)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
 
 	bid := []orderbook.Tranche{
 		{Price: 10, Amount: 1},
@@ -3653,10 +3634,8 @@ func TestGetOrderbookAmountByImpact(t *testing.T) {
 		{Price: 13, Amount: 1},
 		{Price: 14, Amount: 1},
 	}
-	err = depth.LoadSnapshot(bid, ask, 0, time.Now(), time.Now(), true)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = depth.LoadSnapshot(&orderbook.Base{Bids: bid, Asks: ask, LastUpdated: time.Now(), UpdatePushedAt: time.Now(), RestSnapshot: true})
+	require.NoError(t, err)
 
 	req.ImpactPercentage = 9.090909090909092
 	impact, err := s.GetOrderbookAmountByImpact(t.Context(), req)
--- a/exchange/websocket/buffer/buffer.go
+++ b/exchange/websocket/buffer/buffer.go
@@ -10,7 +10,6 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/currency"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/orderbook"
-	"github.com/thrasher-corp/gocryptotrader/log"
 )
 
 const packageError = "websocket orderbook buffer error: %w"
@@ -20,16 +19,6 @@ var (
 	errBufferConfigNil              = errors.New("buffer config is nil")
 	errUnsetDataHandler             = errors.New("datahandler unset")
 	errIssueBufferEnabledButNoLimit = errors.New("buffer enabled but no limit set")
-	errUpdateIsNil                  = errors.New("update is nil")
-	errUpdateNoTargets              = errors.New("update bid/ask targets cannot be nil")
-	errDepthNotFound                = errors.New("orderbook depth not found")
-	errRESTOverwrite                = errors.New("orderbook has been overwritten by REST protocol")
-	errInvalidAction                = errors.New("invalid action")
-	errAmendFailure                 = errors.New("orderbook amend update failure")
-	errDeleteFailure                = errors.New("orderbook delete update failure")
-	errInsertFailure                = errors.New("orderbook insert update failure")
-	errUpdateInsertFailure          = errors.New("orderbook update/insert update failure")
-	errRESTTimerLapse               = errors.New("rest sync timer lapse with active websocket connection")
 	errOrderbookFlushed             = errors.New("orderbook flushed")
 )
 
@@ -55,250 +44,38 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
 
 	w.sortBuffer = c.SortBuffer
 	w.sortBufferByUpdateIDs = c.SortBufferByUpdateIDs
-	w.updateEntriesByID = c.UpdateEntriesByID
 	w.exchangeName = exchangeConfig.Name
 	w.dataHandler = dataHandler
 	w.ob = make(map[key.PairAsset]*orderbookHolder)
 	w.verbose = exchangeConfig.Verbose
-	w.updateIDProgression = c.UpdateIDProgression
-	w.checksum = c.Checksum
-	return nil
-}
-
-// validate validates update against setup values
-func (w *Orderbook) validate(u *orderbook.Update) error {
-	if u == nil {
-		return fmt.Errorf(packageError, errUpdateIsNil)
-	}
-	if len(u.Bids) == 0 && len(u.Asks) == 0 {
-		return fmt.Errorf(packageError, errUpdateNoTargets)
-	}
-	return nil
-}
-
-// Update updates a stored pointer to an orderbook.Depth struct containing a
-// bid and ask Tranches, this switches between the usage of a buffered update
-func (w *Orderbook) Update(u *orderbook.Update) error {
-	if err := w.validate(u); err != nil {
-		return err
-	}
-	w.mtx.Lock()
-	defer w.mtx.Unlock()
-	book, ok := w.ob[key.PairAsset{Base: u.Pair.Base.Item, Quote: u.Pair.Quote.Item, Asset: u.Asset}]
-	if !ok {
-		return fmt.Errorf("%w for Exchange %s CurrencyPair: %s AssetType: %s",
-			errDepthNotFound,
-			w.exchangeName,
-			u.Pair,
-			u.Asset)
-	}
-
-	// out of order update ID can be skipped
-	if w.updateIDProgression && u.UpdateID <= book.updateID {
-		if w.verbose {
-			log.Warnf(log.WebsocketMgr,
-				"Exchange %s CurrencyPair: %s AssetType: %s out of order websocket update received",
-				w.exchangeName,
-				u.Pair,
-				u.Asset)
-		}
-		return nil
-	}
-
-	// Checks for when the rest protocol overwrites a streaming dominated book
-	// will stop updating book via incremental updates. This occurs because our
-	// sync manager (engine/sync.go) timer has elapsed for streaming. Usually
-	// because the book is highly illiquid.
-	isREST, err := book.ob.IsRESTSnapshot()
-	if err != nil {
-		if !errors.Is(err, orderbook.ErrOrderbookInvalid) {
-			return err
-		}
-		// In the event a checksum or processing error invalidates the book, all
-		// updates that could be stored in the websocket buffer, skip applying
-		// until a new snapshot comes through.
-		if w.verbose {
-			log.Warnf(log.WebsocketMgr,
-				"Exchange %s CurrencyPair: %s AssetType: %s underlying book is invalid, cannot apply update.",
-				w.exchangeName,
-				u.Pair,
-				u.Asset)
-		}
-		return nil
-	}
-
-	if isREST {
-		if w.verbose {
-			log.Warnf(log.WebsocketMgr,
-				"%s for Exchange %s CurrencyPair: %s AssetType: %s consider extending synctimeoutwebsocket",
-				errRESTOverwrite,
-				w.exchangeName,
-				u.Pair,
-				u.Asset)
-		}
-		// Instance of illiquidity, this signal notifies that there is websocket
-		// activity. We can invalidate the book and request a new snapshot. All
-		// further updates through the websocket should be caught above in the
-		// IsRestSnapshot() call.
-		return book.ob.Invalidate(errRESTTimerLapse)
-	}
-
-	if w.bufferEnabled {
-		var processed bool
-		processed, err = w.processBufferUpdate(book, u)
-		if err != nil {
-			return err
-		}
-
-		if !processed {
-			return nil
-		}
-	} else {
-		err = w.processObUpdate(book, u)
-		if err != nil {
-			return err
-		}
-	}
-
-	// Publish all state changes, disregarding verbosity or sync requirements.
-	book.ob.Publish()
-	w.dataHandler <- book.ob
-	return nil
-}
-
-// processBufferUpdate stores update into buffer, when buffer at capacity as
-// defined by w.obBufferLimit it well then sort and apply updates.
-func (w *Orderbook) processBufferUpdate(o *orderbookHolder, u *orderbook.Update) (bool, error) {
-	*o.buffer = append(*o.buffer, *u)
-	if len(*o.buffer) < w.obBufferLimit {
-		return false, nil
-	}
-
-	if w.sortBuffer {
-		// sort by last updated to ensure each update is in order
-		if w.sortBufferByUpdateIDs {
-			sort.Slice(*o.buffer, func(i, j int) bool {
-				return (*o.buffer)[i].UpdateID < (*o.buffer)[j].UpdateID
-			})
-		} else {
-			sort.Slice(*o.buffer, func(i, j int) bool {
-				return (*o.buffer)[i].UpdateTime.Before((*o.buffer)[j].UpdateTime)
-			})
-		}
-	}
-	for i := range *o.buffer {
-		err := w.processObUpdate(o, &(*o.buffer)[i])
-		if err != nil {
-			return false, err
-		}
-	}
-	// clear buffer of old updates
-	*o.buffer = nil
-	return true, nil
-}
-
-// processObUpdate processes updates either by its corresponding id or by price level
-func (w *Orderbook) processObUpdate(o *orderbookHolder, u *orderbook.Update) error {
-	// Both update methods require post processing to ensure the orderbook is in a valid state.
-	if w.updateEntriesByID {
-		if err := o.updateByIDAndAction(u); err != nil {
-			return err
-		}
-	} else {
-		if err := o.updateByPrice(u); err != nil {
-			return err
-		}
-	}
-	if w.checksum != nil {
-		compare, err := o.ob.Retrieve()
-		if err != nil {
-			return err
-		}
-		err = w.checksum(compare, u.Checksum)
-		if err != nil {
-			return o.ob.Invalidate(err)
-		}
-		o.updateID = u.UpdateID
-	} else if o.ob.VerifyOrderbook() {
-		compare, err := o.ob.Retrieve()
-		if err != nil {
-			return err
-		}
-		err = compare.Verify()
-		if err != nil {
-			return o.ob.Invalidate(err)
-		}
-	}
-	return nil
-}
-
-// updateByPrice amends amount if match occurs by price, deletes if amount is
-// zero or less and inserts if not found.
-func (o *orderbookHolder) updateByPrice(updts *orderbook.Update) error {
-	return o.ob.UpdateBidAskByPrice(updts)
-}
-
-// updateByIDAndAction will receive an action to execute against the orderbook
-// it will then match by IDs instead of price to perform the action
-func (o *orderbookHolder) updateByIDAndAction(updts *orderbook.Update) error {
-	switch updts.Action {
-	case orderbook.Amend:
-		err := o.ob.UpdateBidAskByID(updts)
-		if err != nil {
-			return fmt.Errorf("%w %w", errAmendFailure, err)
-		}
-	case orderbook.Delete:
-		// edge case for Bitfinex as their streaming endpoint duplicates deletes
-		bypassErr := o.ob.GetName() == "Bitfinex" && o.ob.IsFundingRate()
-		err := o.ob.DeleteBidAskByID(updts, bypassErr)
-		if err != nil {
-			return fmt.Errorf("%w %w", errDeleteFailure, err)
-		}
-	case orderbook.Insert:
-		err := o.ob.InsertBidAskByID(updts)
-		if err != nil {
-			return fmt.Errorf("%w %w", errInsertFailure, err)
-		}
-	case orderbook.UpdateInsert:
-		err := o.ob.UpdateInsertByID(updts)
-		if err != nil {
-			return fmt.Errorf("%w %w", errUpdateInsertFailure, err)
-		}
-	default:
-		return fmt.Errorf("%w [%d]", errInvalidAction, updts.Action)
-	}
 	return nil
 }
 
 // LoadSnapshot loads initial snapshot of orderbook data from websocket
 func (w *Orderbook) LoadSnapshot(book *orderbook.Base) error {
-	// Checks if book can deploy to depth
-	err := book.Verify()
-	if err != nil {
+	if err := book.Verify(); err != nil {
 		return err
 	}
 
-	w.mtx.Lock()
-	defer w.mtx.Unlock()
+	w.m.RLock()
 	holder, ok := w.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}]
+	w.m.RUnlock()
 	if !ok {
+		w.m.Lock()
 		// Associate orderbook pointer with local exchange depth map
-		var depth *orderbook.Depth
-		depth, err = orderbook.DeployDepth(book.Exchange, book.Pair, book.Asset)
+		depth, err := orderbook.DeployDepth(book.Exchange, book.Pair, book.Asset)
 		if err != nil {
 			return err
 		}
 		depth.AssignOptions(book)
 		buffer := make([]orderbook.Update, w.obBufferLimit)
-
 		holder = &orderbookHolder{ob: depth, buffer: &buffer}
 		w.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}] = holder
+		w.m.Unlock()
 	}
 
-	holder.updateID = book.LastUpdateID
-
-	err = holder.ob.LoadSnapshot(book.Bids, book.Asks, book.LastUpdateID, book.LastUpdated, book.UpdatePushedAt, false)
-	if err != nil {
+	book.RestSnapshot = false
+	if err := holder.ob.LoadSnapshot(book); err != nil {
 		return err
 	}
 
@@ -307,38 +84,102 @@ func (w *Orderbook) LoadSnapshot(book *orderbook.Base) error {
 	return nil
 }
 
+// Update updates a stored pointer to an orderbook.Depth struct containing bid and ask Tranches, this switches between
+// the usage of a buffered update
+func (w *Orderbook) Update(u *orderbook.Update) error {
+	w.m.RLock()
+	holder, ok := w.ob[key.PairAsset{Base: u.Pair.Base.Item, Quote: u.Pair.Quote.Item, Asset: u.Asset}]
+	w.m.RUnlock()
+	if !ok {
+		return fmt.Errorf("%w for Exchange %s CurrencyPair: %s AssetType: %s", orderbook.ErrDepthNotFound, w.exchangeName, u.Pair, u.Asset)
+	}
+
+	if w.bufferEnabled {
+		if processed, err := w.processBufferUpdate(holder, u); err != nil || !processed {
+			return err
+		}
+	} else {
+		if err := holder.ob.ProcessUpdate(u); err != nil {
+			return err
+		}
+	}
+
+	// Publish all state changes, disregarding verbosity or sync requirements.
+	holder.ob.Publish()
+	w.dataHandler <- holder.ob
+	return nil
+}
+
+// processBufferUpdate stores update into buffer, when buffer at capacity as
+// defined by w.obBufferLimit it well then sort and apply updates.
+func (w *Orderbook) processBufferUpdate(holder *orderbookHolder, u *orderbook.Update) (bool, error) {
+	*holder.buffer = append(*holder.buffer, *u)
+	if len(*holder.buffer) < w.obBufferLimit {
+		return false, nil
+	}
+
+	if w.sortBuffer {
+		// sort by last updated to ensure each update is in order
+		if w.sortBufferByUpdateIDs {
+			sort.Slice(*holder.buffer, func(i, j int) bool {
+				return (*holder.buffer)[i].UpdateID < (*holder.buffer)[j].UpdateID
+			})
+		} else {
+			sort.Slice(*holder.buffer, func(i, j int) bool {
+				return (*holder.buffer)[i].UpdateTime.Before((*holder.buffer)[j].UpdateTime)
+			})
+		}
+	}
+
+	// clear buffer of old updates on all error pathways as any error will invalidate the orderbook and will require
+	// a new snapshot to be loaded
+	defer func() { *holder.buffer = (*holder.buffer)[:0] }()
+
+	for i := range *holder.buffer {
+		if err := holder.ob.ProcessUpdate(&(*holder.buffer)[i]); err != nil {
+			return false, err
+		}
+	}
+
+	return true, nil
+}
+
 // GetOrderbook returns an orderbook copy as orderbook.Base
 func (w *Orderbook) GetOrderbook(p currency.Pair, a asset.Item) (*orderbook.Base, error) {
-	w.mtx.Lock()
-	defer w.mtx.Unlock()
-	book, ok := w.ob[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
-	if !ok {
-		return nil, fmt.Errorf("%s %s %s %w", w.exchangeName, p, a, errDepthNotFound)
+	if p.IsEmpty() {
+		return nil, currency.ErrCurrencyPairEmpty
 	}
-	return book.ob.Retrieve()
-}
-
-// FlushBuffer flushes w.ob data to be garbage collected and refreshed when a
-// connection is lost and reconnected
-func (w *Orderbook) FlushBuffer() {
-	w.mtx.Lock()
-	w.ob = make(map[key.PairAsset]*orderbookHolder)
-	w.mtx.Unlock()
-}
-
-// FlushOrderbook flushes independent orderbook
-func (w *Orderbook) FlushOrderbook(p currency.Pair, a asset.Item) error {
-	w.mtx.Lock()
-	defer w.mtx.Unlock()
-	book, ok := w.ob[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	if !a.IsValid() {
+		return nil, asset.ErrInvalidAsset
+	}
+	w.m.RLock()
+	holder, ok := w.ob[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	w.m.RUnlock()
 	if !ok {
-		return fmt.Errorf("cannot flush orderbook %s %s %s %w",
-			w.exchangeName,
-			p,
-			a,
-			errDepthNotFound)
+		return nil, fmt.Errorf("%s %w: %s.%s", w.exchangeName, orderbook.ErrDepthNotFound, a, p)
+	}
+	return holder.ob.Retrieve()
+}
+
+// FlushBuffer flushes individual orderbook buffers while keeping the orderbook lookups intact and ready for new updates
+// when a connection is re-established.
+func (w *Orderbook) FlushBuffer() {
+	w.m.Lock()
+	for _, holder := range w.ob {
+		*holder.buffer = (*holder.buffer)[:0]
+	}
+	w.m.Unlock()
+}
+
+// InvalidateOrderbook invalidates the orderbook so no trading can occur on potential corrupted data
+func (w *Orderbook) InvalidateOrderbook(p currency.Pair, a asset.Item) error {
+	w.m.RLock()
+	holder, ok := w.ob[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	w.m.RUnlock()
+	if !ok {
+		return fmt.Errorf("cannot invalidate orderbook %s %s %s %w", w.exchangeName, p, a, orderbook.ErrDepthNotFound)
 	}
 	// error not needed in this return
-	_ = book.ob.Invalidate(errOrderbookFlushed)
+	_ = holder.ob.Invalidate(errOrderbookFlushed)
 	return nil
 }
--- a/exchange/websocket/buffer/buffer_test.go
+++ b/exchange/websocket/buffer/buffer_test.go
@@ -1,9 +1,7 @@
 package buffer
 
 import (
-	"errors"
 	"math/rand"
-	"slices"
 	"strconv"
 	"testing"
 	"time"
@@ -67,64 +65,6 @@ func createSnapshot(pair currency.Pair, bookVerifiy ...bool) (holder *Orderbook,
 	}
 	err = holder.LoadSnapshot(book)
 	return holder, asks, bids, err
-}
-
-func bidAskGenerator() []orderbook.Tranche {
-	response := make([]orderbook.Tranche, 100)
-	for i := range 100 {
-		price := float64(rand.Intn(1000)) //nolint:gosec // no need to import crypo/rand for testing
-		if price == 0 {
-			price = 1
-		}
-		response[i] = orderbook.Tranche{
-			Amount: float64(rand.Intn(10)), //nolint:gosec // no need to import crypo/rand for testing
-			Price:  price,
-			ID:     int64(i),
-		}
-	}
-	return response
-}
-
-func BenchmarkUpdateBidsByPrice(b *testing.B) {
-	cp, err := getExclusivePair()
-	require.NoError(b, err)
-
-	ob, _, _, err := createSnapshot(cp)
-	require.NoError(b, err)
-
-	for b.Loop() {
-		bidAsks := bidAskGenerator()
-		update := &orderbook.Update{
-			Bids:       bidAsks,
-			Asks:       bidAsks,
-			Pair:       cp,
-			UpdateTime: time.Now(),
-			Asset:      asset.Spot,
-		}
-		holder := ob.ob[key.PairAsset{Base: cp.Base.Item, Quote: cp.Quote.Item, Asset: asset.Spot}]
-		require.NoError(b, holder.updateByPrice(update))
-	}
-}
-
-func BenchmarkUpdateAsksByPrice(b *testing.B) {
-	cp, err := getExclusivePair()
-	require.NoError(b, err)
-
-	ob, _, _, err := createSnapshot(cp)
-	require.NoError(b, err)
-
-	for b.Loop() {
-		bidAsks := bidAskGenerator()
-		update := &orderbook.Update{
-			Bids:       bidAsks,
-			Asks:       bidAsks,
-			Pair:       cp,
-			UpdateTime: time.Now(),
-			Asset:      asset.Spot,
-		}
-		holder := ob.ob[key.PairAsset{Base: cp.Base.Item, Quote: cp.Quote.Item, Asset: asset.Spot}]
-		require.NoError(b, holder.updateByPrice(update))
-	}
 }
 
 // BenchmarkBufferPerformance demonstrates buffer more performant than multi
@@ -236,38 +176,6 @@ func BenchmarkNoBufferPerformance(b *testing.B) {
 	}
 }
 
-func TestUpdates(t *testing.T) {
-	t.Parallel()
-	cp, err := getExclusivePair()
-	require.NoError(t, err)
-
-	holder, _, _, err := createSnapshot(cp)
-	require.NoError(t, err)
-
-	book := holder.ob[key.PairAsset{Base: cp.Base.Item, Quote: cp.Quote.Item, Asset: asset.Spot}]
-	err = book.updateByPrice(&orderbook.Update{
-		Bids:       itemArray[5],
-		Asks:       itemArray[5],
-		Pair:       cp,
-		UpdateTime: time.Now(),
-		Asset:      asset.Spot,
-	})
-	assert.NoError(t, err)
-
-	err = book.updateByPrice(&orderbook.Update{
-		Bids:       itemArray[0],
-		Asks:       itemArray[0],
-		Pair:       cp,
-		UpdateTime: time.Now(),
-		Asset:      asset.Spot,
-	})
-	assert.NoError(t, err)
-
-	askLen, err := book.ob.GetAskLength()
-	require.NoError(t, err)
-	assert.Equal(t, 3, askLen)
-}
-
 // TestHittingTheBuffer logic test
 func TestHittingTheBuffer(t *testing.T) {
 	t.Parallel()
@@ -312,7 +220,6 @@ func TestInsertWithIDs(t *testing.T) {
 	require.NoError(t, err)
 
 	holder.bufferEnabled = true
-	holder.updateEntriesByID = true
 	holder.obBufferLimit = 5
 	for i := range itemArray {
 		asks := itemArray[i]
@@ -346,8 +253,6 @@ func TestInsertWithIDs(t *testing.T) {
 	holder, _, _, err = createSnapshot(cp, true)
 	require.NoError(t, err)
 
-	holder.checksum = nil
-	holder.updateIDProgression = false
 	err = holder.Update(&orderbook.Update{
 		UpdateTime: time.Now(),
 		Asset:      asset.Spot,
@@ -436,15 +341,15 @@ func TestOrderbookLastUpdateID(t *testing.T) {
 
 	assert.Equal(t, 1000., itemArray[0][0].Price)
 
-	holder.checksum = func(*orderbook.Base, uint32) error { return errors.New("testerino") }
-
 	// this update invalidates the book
 	err = holder.Update(&orderbook.Update{
-		Asks:       []orderbook.Tranche{{Price: 999999}},
-		Pair:       cp,
-		UpdateID:   -1,
-		Asset:      asset.Spot,
-		UpdateTime: time.Now(),
+		Asks:             []orderbook.Tranche{{Price: 999999}},
+		Pair:             cp,
+		UpdateID:         -1,
+		Asset:            asset.Spot,
+		UpdateTime:       time.Now(),
+		ExpectedChecksum: 1337,
+		GenerateChecksum: func(*orderbook.Base) uint32 { return 1336 },
 	})
 	require.ErrorIs(t, err, orderbook.ErrOrderbookInvalid)
 
@@ -454,27 +359,28 @@ func TestOrderbookLastUpdateID(t *testing.T) {
 	holder, _, _, err = createSnapshot(cp)
 	require.NoError(t, err)
 
-	holder.checksum = func(*orderbook.Base, uint32) error { return nil }
-	holder.updateIDProgression = true
-
 	for i := range itemArray {
 		asks := itemArray[i]
 		err = holder.Update(&orderbook.Update{
-			Asks:       asks,
-			Pair:       cp,
-			UpdateID:   int64(i) + 1,
-			Asset:      asset.Spot,
-			UpdateTime: time.Now(),
+			Asks:                       asks,
+			Pair:                       cp,
+			UpdateID:                   int64(i) + 1,
+			Asset:                      asset.Spot,
+			UpdateTime:                 time.Now(),
+			SkipOutOfOrderLastUpdateID: true,
+			ExpectedChecksum:           1337,
+			GenerateChecksum:           func(snapshot *orderbook.Base) uint32 { return 1337 },
 		})
 		require.NoError(t, err)
 	}
 
 	// out of order
 	err = holder.Update(&orderbook.Update{
-		Asks:     []orderbook.Tranche{{Price: 999999}},
-		Pair:     cp,
-		UpdateID: 1,
-		Asset:    asset.Spot,
+		Asks:                       []orderbook.Tranche{{Price: 999999}},
+		Pair:                       cp,
+		UpdateID:                   1,
+		Asset:                      asset.Spot,
+		SkipOutOfOrderLastUpdateID: true,
 	})
 	require.NoError(t, err)
 
@@ -500,7 +406,7 @@ func TestRunUpdateWithoutSnapshot(t *testing.T) {
 		UpdateTime: time.Now(),
 		Asset:      asset.Spot,
 	})
-	require.ErrorIs(t, err, errDepthNotFound)
+	require.ErrorIs(t, err, orderbook.ErrDepthNotFound)
 }
 
 // TestRunUpdateWithoutAnyUpdates logic test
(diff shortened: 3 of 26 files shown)


samuael (commented):
👏  clean and easy to understand PR.
I couldn't find issues in this round. I'll try another round or approve on the next round.

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,232 @@
...
+	default:
+		return fmt.Errorf("%w [%d]", errInvalidAction, u.Action)
+	}
+	return nil
+}
+
+// updateBidAskByPrice updates the bid and ask spread by supplied updates, this will trim total length of depth level to
+// a specified supplied number
+func (d *Depth) updateBidAskByPrice(update *Update) error {
+	if update.UpdateTime.IsZero() {
```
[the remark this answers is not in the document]
shazbert: Yeah that's a good point what I will do, is defer this and add it to the TODO in `ProcessUpdate` with regard to `SkipOutOfOrderLastUpdateID`, this check will be removed. I am trying to keep it roughly the same even though it's not exactly what should be happening and will impact across exchanges and could lead to a massive re-write and review toil.

exchange/websocket/buffer/buffer_types.go
```diff
@@ -15,14 +15,6 @@ type Config struct {
...
 	// corresponding update IDs.
 	SortBufferByUpdateIDs bool
-	// UpdateEntriesByID will match by IDs instead of price to perform the an
-	// action. e.g. update, delete, insert.
-	UpdateEntriesByID bool
-	// UpdateIDProgression requires that the new update ID be greater than the
-	// prior ID. This will skip processing and not error.
-	UpdateIDProgression bool
-	// Checksum is a package defined checksum calculation for updated books.
-	Checksum func(state *orderbook.Base, checksum uint32) error
```
cranktakular: Why's the checksum being removed? This functionality seems useful for some exchanges.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
-			log.Warnf(log.WebsocketMgr,
-				"Exchange %s CurrencyPair: %s AssetType: %s out of order websocket update received",
-				w.exchangeName,
-				u.Pair,
-				u.Asset)
+	w.m.RLock()
+	holder, ok := w.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}]
+	w.m.RUnlock()
+	if !ok {
+		w.m.Lock()
```
cranktakular: This lock seems early. Fields of w aren't read until a few lines later, and they aren't written until a few lines after that.

If this is where you want to lock it, then you'd need to call unlock before returning the error on line 68; otherwise the mutex stays locked forever when that path's taken.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
+		holder = &orderbookHolder{ob: depth, buffer: &buffer}
+		w.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}] = holder
+		w.m.Unlock()
 	}
 
-	// Checks for when the rest protocol overwrites a streaming dominated book
-	// will stop updating book via incremental updates. This occurs because our
-	// sync manager (engine/sync.go) timer has elapsed for streaming. Usually
-	// because the book is highly illiquid.
-	isREST, err := book.ob.IsRESTSnapshot()
```
cranktakular, line 116: Is it good to remove this check? It seems like a decently large part of the purpose for that variable in the first place.

I see that it filters through Depth and options into triggering some branches elsewhere, but those don't seem too meaningful to me, honestly.

Then again, I could easily be missing something due to my lack of trading experience.

engine/rpcserver_test.go
```diff
@@ -3243,18 +3243,15 @@ func TestGetOrderbookMovement(t *testing.T) {
 	t.Parallel()
 	em := NewExchangeManager()
 	exch, err := em.NewExchangeByName("binance")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "NewExchangeByName must not error")
```

exchange/websocket/buffer/buffer_types.go
```diff
@@ -33,27 +25,16 @@ type Orderbook struct {
...
 }
 
 // orderbookHolder defines a store of pending updates and a pointer to the
 // orderbook depth
 type orderbookHolder struct {
-	ob       *orderbook.Depth
-	buffer   *[]orderbook.Update
-	updateID int64
+	ob     *orderbook.Depth
+	buffer *[]orderbook.Update
```
cranktakular: Slices intrinsically point to the data anyway, so there's not much reason to make this a pointer.

Changing this doesn't seem to break any tests in the package.

engine/rpcserver_test.go
```diff
@@ -3243,18 +3243,15 @@ func TestGetOrderbookMovement(t *testing.T) {
...
 	exch.SetDefaults()
 	b := exch.GetBase()
 	b.Name = fakeExchangeName
 	b.Enabled = true
 
 	cp, err := currency.NewPairFromString("btc-metal")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk: 🐒  Either require needs a message, or just remove the err call.
If accepted, L3268 needs to be `err :=`
```suggestion
	cp := currency.NewPairWithDelimiter("btc", "metal", "-")
```

engine/rpcserver_test.go
```diff
@@ -3302,9 +3299,7 @@ func TestGetOrderbookMovement(t *testing.T) {
 	}
 
 	depth, err := orderbook.DeployDepth(req.Exchange, currency.NewPair(currency.BTC, currency.METAL), asset.Spot)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "orderbook.DeployDepth must not error")
```

engine/rpcserver_test.go
```diff
@@ -3318,10 +3313,8 @@ func TestGetOrderbookMovement(t *testing.T) {
 		{Price: 13, Amount: 1},
 		{Price: 14, Amount: 1},
 	}
-	err = depth.LoadSnapshot(bid, ask, 0, time.Now(), time.Now(), true)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = depth.LoadSnapshot(&orderbook.Base{Bids: bid, Asks: ask, LastUpdated: time.Now(), UpdatePushedAt: time.Now(), RestSnapshot: true})
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "depth.LoadSnapshot must not error")
```

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
-		err = compare.Verify()
-		if err != nil {
-			return o.ob.Invalidate(err)
-		}
-	}
-	return nil
-}
+	// clear buffer of old updates on all error pathways as any error will invalidate the orderbook and will require
+	// a new snapshot to be loaded
+	defer func() { *holder.buffer = (*holder.buffer)[:0] }()
```
cranktakular: Clearing this way doesn't actually delete the data, it's just hidden from view.

Reinstantiating it with an empty slice would work.

engine/rpcserver_test.go
```diff
@@ -3350,18 +3343,15 @@ func TestGetOrderbookAmountByNominal(t *testing.T) {
 	t.Parallel()
 	em := NewExchangeManager()
 	exch, err := em.NewExchangeByName("binance")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "NewExchangeByName must not error")
```

exchange/websocket/buffer/buffer.go
```diff
@@ -336,32 +169,35 @@ func (w *Orderbook) LastUpdateID(p currency.Pair, a asset.Item) (int64, error) {
...
-// connection is lost and reconnected
+// FlushBuffer flushes individual orderbook buffers while keeping the orderbook lookups intact and ready for new updates
+// when a connection is re-established.
 func (w *Orderbook) FlushBuffer() {
-	w.mtx.Lock()
-	w.ob = make(map[key.PairAsset]*orderbookHolder)
-	w.mtx.Unlock()
+	w.m.Lock()
+	for _, holder := range w.ob {
+		*holder.buffer = (*holder.buffer)[:0]
```
cranktakular: Similar comment to above, this doesn't actually delete the data in those buffers. Either setting those buffers to nil, or setting them to new empty slices, would delete the data while still keeping the keys alive.

engine/rpcserver_test.go
```diff
@@ -3350,18 +3343,15 @@ func TestGetOrderbookAmountByNominal(t *testing.T) {
...
 	exch.SetDefaults()
 	b := exch.GetBase()
 	b.Name = fakeExchangeName
 	b.Enabled = true
 
 	cp, err := currency.NewPairFromString("btc-meme")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk: 🐒 
If accepted L3368 needs changing
```suggestion
	cp := currency.NewPairWithDelimiter("btc", "meme", "-")
```

engine/rpcserver_test.go
```diff
@@ -3409,9 +3399,7 @@ func TestGetOrderbookAmountByNominal(t *testing.T) {
 	}
 
 	depth, err := orderbook.DeployDepth(req.Exchange, currency.NewPair(currency.BTC, currency.MEME), asset.Spot)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "orderbook.DeployDepth must not error")
```

engine/rpcserver_test.go
```diff
@@ -3425,10 +3413,8 @@ func TestGetOrderbookAmountByNominal(t *testing.T) {
 		{Price: 13, Amount: 1},
 		{Price: 14, Amount: 1},
 	}
-	err = depth.LoadSnapshot(bid, ask, 0, time.Now(), time.Now(), true)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = depth.LoadSnapshot(&orderbook.Base{Bids: bid, Asks: ask, LastUpdated: time.Now(), UpdatePushedAt: time.Now(), RestSnapshot: true})
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "depth.LoadSnapshot must not error")
```

engine/rpcserver_test.go
```diff
@@ -3450,18 +3436,15 @@ func TestGetOrderbookAmountByImpact(t *testing.T) {
 	t.Parallel()
 	em := NewExchangeManager()
 	exch, err := em.NewExchangeByName("binance")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "NewExchangeByName must not error")
```

engine/rpcserver_test.go
```diff
@@ -3450,18 +3436,15 @@ func TestGetOrderbookAmountByImpact(t *testing.T) {
...
 	exch.SetDefaults()
 	b := exch.GetBase()
 	b.Name = fakeExchangeName
 	b.Enabled = true
 
 	cp, err := currency.NewPairFromString("btc-mad")
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk: 🐒 
If accepted L3461 needs changing
```suggestion
	cp := currency.NewPairWithDelimiter("btc", "mad", "-")
```

engine/rpcserver_test.go
```diff
@@ -3509,9 +3492,7 @@ func TestGetOrderbookAmountByImpact(t *testing.T) {
 	}
 
 	depth, err := orderbook.DeployDepth(req.Exchange, currency.NewPair(currency.BTC, currency.MAD), asset.Spot)
-	if err != nil {
-		t.Fatal(err)
-	}
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "orderbook.DeployDepth must not error")
```

engine/rpcserver_test.go
```diff
@@ -3525,10 +3506,8 @@ func TestGetOrderbookAmountByImpact(t *testing.T) {
 		{Price: 13, Amount: 1},
 		{Price: 14, Amount: 1},
 	}
-	err = depth.LoadSnapshot(bid, ask, 0, time.Now(), time.Now(), true)
-	if err != nil {
-		t.Fatal(err)
-	}
+	err = depth.LoadSnapshot(&orderbook.Base{Bids: bid, Asks: ask, LastUpdated: time.Now(), UpdatePushedAt: time.Now(), RestSnapshot: true})
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "depth.LoadSnapshot must not error")
```

exchange/websocket/buffer/buffer_test.go
```diff
@@ -455,27 +360,28 @@ func TestOrderbookLastUpdateID(t *testing.T) {
...
+			UpdateTime:                 time.Now(),
+			SkipOutOfOrderLastUpdateID: true,
+			ExpectedChecksum:           1337,
+			GenerateChecksum:           func(*orderbook.Base) uint32 { return 1337 },
 		})
 		require.NoError(t, err)
 	}
 
 	// out of order
 	err = holder.Update(&orderbook.Update{
```
cranktakular, line 373: What is this testing? Removing it doesn't lose any coverage. Comment indicates it was testing out of order functionality, but you added a parameter so it would skip that, and this package no longer has handling for that regardless.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
+
+// Update updates a stored pointer to an orderbook.Depth struct containing bid and ask Tranches, this switches between
+// the usage of a buffered update
+func (w *Orderbook) Update(u *orderbook.Update) error {
+	w.m.RLock()
+	holder, ok := w.ob[key.PairAsset{Base: u.Pair.Base.Item, Quote: u.Pair.Quote.Item, Asset: u.Asset}]
+	w.m.RUnlock()
+	if !ok {
+		return fmt.Errorf("%w for Exchange %s CurrencyPair: %s AssetType: %s", orderbook.ErrDepthNotFound, w.exchangeName, u.Pair, u.Asset)
 	}
```
gbjk, line 97: 🤷  There's no `common.NilGuard` here on `holder` or `u`. Perhaps there should be, before we get into processBufferUpdate

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
-	if len(*o.buffer) < w.obBufferLimit {
+func (w *Orderbook) processBufferUpdate(holder *orderbookHolder, u *orderbook.Update) (bool, error) {
+	*holder.buffer = append(*holder.buffer, *u)
+	if len(*holder.buffer) < w.obBufferLimit {
 		return false, nil
 	}
 
 	if w.sortBuffer {
 		// sort by last updated to ensure each update is in order
 		if w.sortBufferByUpdateIDs {
```
gbjk: 🥃 Thinking about methods to unify this. Out of scope, but it's not great we're having to branch this late.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
-			sort.Slice(*o.buffer, func(i, j int) bool {
-				return (*o.buffer)[i].UpdateID < (*o.buffer)[j].UpdateID
+			sort.Slice(*holder.buffer, func(i, j int) bool {
+				return (*holder.buffer)[i].UpdateID < (*holder.buffer)[j].UpdateID
 			})
 		} else {
-			sort.Slice(*o.buffer, func(i, j int) bool {
-				return (*o.buffer)[i].UpdateTime.Before((*o.buffer)[j].UpdateTime)
+			sort.Slice(*holder.buffer, func(i, j int) bool {
+				return (*holder.buffer)[i].UpdateTime.Before((*holder.buffer)[j].UpdateTime)
```
gbjk: 🚧 Change Request: use `slices.SortFunc` as per `sort.Slice` docs
Also leverage cmp.Compare for int64 overflow, and time.Time.Compare
Considered moving the if into the func but that'd be less performant.
```suggestion
			slices.SortFunc(o.buffer, func(a, b orderbook.Update) int {
				return cmp.Compare(a.UpdateID, b.UpdateID)
			})
		} else {
			slices.SortFunc(o.buffer, func(a, b orderbook.Update) int {
				return a.UpdateTime.Compare(b.UpdateTime)
			})
```

exchange/websocket/buffer/buffer.go
gbjk: 🚧 🚨 I think this comment and code does not match what's going to happen.
There's a non-error default return below, right?
Sorry if I'm mistaken.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
 
-	// out of order update ID can be skipped
-	if w.updateIDProgression && u.UpdateID <= book.updateID {
-		if w.verbose {
-			log.Warnf(log.WebsocketMgr,
-				"Exchange %s CurrencyPair: %s AssetType: %s out of order websocket update received",
-				w.exchangeName,
-				u.Pair,
-				u.Asset)
+	w.m.RLock()
```
cranktakular: This panics if w is nil.

exchanges/orderbook/orderbook.go
```diff
@@ -153,28 +154,24 @@ func (b *Base) TotalAsksAmount() (amountCollated, total float64) {
...
 
-// Verify ensures that the orderbook items are correctly sorted prior to being
-// set and will reject any book with incorrect values.
-// Bids should always go from a high price to a low price and
-// Asks should always go from a low price to a higher price
+// Verify ensures that the orderbook items are correctly sorted prior to being set and will reject any book with
+// incorrect values.
+// Bids should always go from a high price to a low price and Asks should always go from a low price to a higher price
 func (b *Base) Verify() error {
 	if !b.VerifyOrderbook {
```
cranktakular: This panics if b is nil.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
-	}
 	return nil
 }
 
-// Update updates a stored pointer to an orderbook.Depth struct containing a
-// bid and ask Tranches, this switches between the usage of a buffered update
-func (w *Orderbook) Update(u *orderbook.Update) error {
-	if err := w.validate(u); err != nil {
+// LoadSnapshot loads initial snapshot of orderbook data from websocket
+func (w *Orderbook) LoadSnapshot(book *orderbook.Base) error {
```
cranktakular: A few lines of coverage are missing in this function.

Call with a Base that has VerifyOrderbook true, and a Tranche with an amount below 0.
Call with a Base that doesn't have anything set.
Call with a Base that has an exchange, pair, and asset set, but no LastUpdated set.

Those three will net you full coverage here.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
 	return nil
 }
 
 // processBufferUpdate stores update into buffer, when buffer at capacity as
 // defined by w.obBufferLimit it well then sort and apply updates.
-func (w *Orderbook) processBufferUpdate(o *orderbookHolder, u *orderbook.Update) (bool, error) {
-	*o.buffer = append(*o.buffer, *u)
-	if len(*o.buffer) < w.obBufferLimit {
+func (w *Orderbook) processBufferUpdate(holder *orderbookHolder, u *orderbook.Update) (bool, error) {
+	*holder.buffer = append(*holder.buffer, *u)
```
cranktakular: This panics if holder or u are nil.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
-	holder, ok := w.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}]
-	if !ok {
-		// Associate orderbook pointer with local exchange depth map
-		var depth *orderbook.Depth
-		depth, err = orderbook.DeployDepth(book.Exchange, book.Pair, book.Asset)
-		if err != nil {
-			return err
+	for i := range *holder.buffer {
+		if err := holder.ob.ProcessUpdate(&(*holder.buffer)[i]); err != nil {
+			return false, err
```
cranktakular, line 140: This currently doesn't have any coverage. Some can be achieved if this function is tested with an empty (but non-nil) Orderbook, orderbookHolder, and Update.

exchanges/orderbook/depth.go
```diff
@@ -92,46 +92,37 @@ func (d *Depth) Retrieve() (*Base, error) {
...
+	d.askTranches.load(incoming.Asks)
 	d.validationError = nil
 	d.Alert()
 	return nil
 }
 
-// invalidate flushes all values back to zero so as to not allow strategy
-// traversal on compromised data. NOTE: This requires locking.
+// invalidate flushes all values back to zero so as to not allow strategy traversal on compromised data.
+// NOTE: This requires locking.
```
cranktakular, line 119: Due to this danger, wouldn't it be better to lock and unlock within this function?

This function's (currently) only called within Invalidate(), where the lock could be shifted without it impacting anything, and in ProcessUpdate(), where its long write lock could be replaced with a much more brief read lock, if the write lock was implemented here instead.

cranktakular (changes requested):
Crank Reviews Part 1

Don't forget to finish the checklist in your original post!

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +44,104 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
 	// Publish all state changes, disregarding verbosity or sync requirements.
-	book.ob.Publish()
-	w.dataHandler <- book.ob
+	holder.ob.Publish()
+	w.dataHandler <- holder.ob
 	return nil
 }
 
 // processBufferUpdate stores update into buffer, when buffer at capacity as
 // defined by w.obBufferLimit it well then sort and apply updates.
```
gbjk: 🚧 
```suggestion
// defined by w.obBufferLimit it well then sorts and apply updates.
```

exchange/websocket/buffer/buffer_types.go
gbjk: 🚧 Yep. ☝️ Exactly this. All of the logic that uses this is much more messy because it's not just using ob.buffer.
```suggestion
	buffer []orderbook.Update
```

exchange/websocket/buffer/buffer.go
```diff
@@ -336,32 +169,35 @@ func (w *Orderbook) LastUpdateID(p currency.Pair, a asset.Item) (int64, error) {
...
+// TODO: Add in reason for invalidation for debugging purposes.
+func (w *Orderbook) InvalidateOrderbook(p currency.Pair, a asset.Item) error {
+	w.m.RLock()
+	holder, ok := w.ob[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	w.m.RUnlock()
 	if !ok {
-		return fmt.Errorf("cannot flush orderbook %s %s %s %w", w.exchangeName, p, a, ErrDepthNotFound)
+		return fmt.Errorf("cannot invalidate orderbook %s %s %s %w", w.exchangeName, p, a, orderbook.ErrDepthNotFound)
 	}
 	// error not needed in this return
```
gbjk: 🐒 This comment doesn't explain why, so it's not actually helping

exchanges/binance/binance_websocket.go
```diff
@@ -742,26 +742,19 @@ func (b *Binance) processJob(p currency.Pair) error {
...
+	if err := b.Websocket.Orderbook.InvalidateOrderbook(p, asset.Spot); err != nil {
+		log.Errorf(log.WebsocketMgr, "%s invalidate orderbook websocket error: %v", b.Name, err)
 	}
-	errClean = b.obm.cleanup(p)
-	if errClean != nil {
-		log.Errorf(log.WebsocketMgr, "%s cleanup websocket error: %v",
-			b.Name,
-			errClean)
+	if err := b.obm.cleanup(p); err != nil {
+		log.Errorf(log.WebsocketMgr, "%s cleanup websocket error: %v", b.Name, err)
```
gbjk: 🐒 Doesn't say that this is about orderbook, and prefer "error Xing Y" format
This same monkey appears in other exchanges' functions too.
```suggestion
		log.Errorf(log.WebsocketMgr, "error during %s websocket orderbook cleanup: %v", b.Name, err)
```

exchanges/binance/binance_websocket.go
```diff
@@ -742,26 +742,19 @@ func (b *Binance) processJob(p currency.Pair) error {
...
-	errClean := b.Websocket.Orderbook.FlushOrderbook(p, asset.Spot)
-	if errClean != nil {
-		log.Errorf(log.WebsocketMgr,
-			"%s flushing websocket error: %v",
-			b.Name,
-			errClean)
+// invalidateAndCleanupOrderbook invalidaates orderbook and cleans local cache
+func (b *Binance) invalidateAndCleanupOrderbook(p currency.Pair) {
+	if err := b.Websocket.Orderbook.InvalidateOrderbook(p, asset.Spot); err != nil {
+		log.Errorf(log.WebsocketMgr, "%s invalidate orderbook websocket error: %v", b.Name, err)
```
gbjk: 🦍  prefer "error Xing Y" format; since it's easy to always structure this way, and the fact it's an error starts the log line. This is a super nit because there's no other issues with the structure of this log.
This same gorilla appears in other exchanges' functions too.
```suggestion
		log.Errorf(log.WebsocketMgr, "%s error during websocket orderbook cleanup: %v", b.Name, err)
```
```suggestion
		log.Errorf(log.WebsocketMgr, "error invalidating %s websocket orderbook: %v", b.Name, err)
```

exchanges/btcmarkets/btcmarkets_websocket.go
```diff
@@ -467,20 +462,9 @@ func (b *BTCMarkets) ReSubscribeSpecificOrderbook(pair currency.Pair) error {
...
-			ob.Pair,
-			ob.Asset,
-			ob.LastUpdateID,
-			checksum,
-			check,
-			errChecksumFailure)
-	}
-	return nil
+// generateChecksum provides assurance on current in memory liquidity
+func generateChecksum(ob *orderbook.Base) uint32 {
```
gbjk: 🚧 I think the name lacks context in the pkg context. Don't think `generate` is the right verb, and I think we don't need a verb.
🐒 I didn't understand this comment sentence.
```suggestion
// orderbookChecksum calculates a checksum for the orderbook liquidity
func orderbookChecksum(ob *orderbook.Base) uint32 {
```

exchanges/orderbook/depth.go
```diff
@@ -92,46 +92,37 @@ func (d *Depth) Retrieve() (*Base, error) {
...
-			d.exchange,
-			d.pair,
-			d.asset,
-			errLastUpdatedNotSet)
-	}
-	d.lastUpdateID = lastUpdateID
-	d.lastUpdated = lastUpdated
-	d.updatePushedAt = updatePushedAt
+	if incoming.LastUpdated.IsZero() {
+		return fmt.Errorf("validation error: %s %s %s - %w", d.exchange, d.pair, d.asset, errLastUpdatedNotSet)
```
gbjk: 🚧 `validation error` feels like it lacks context.
```suggestion
		return fmt.Errorf("error loading orderbook snapshot: %s %s %s - %w", d.exchange, d.pair, d.asset, errLastUpdatedNotSet)
```

exchanges/orderbook/depth.go
```diff
@@ -92,46 +92,37 @@ func (d *Depth) Retrieve() (*Base, error) {
...
-		d.asset,
-		common.AppendError(ErrOrderbookInvalid, withReason))
+	d.validationError = fmt.Errorf("%s %s %s Reason: [%w]", d.exchange, d.pair, d.asset, common.AppendError(ErrOrderbookInvalid, withReason))
 	d.Alert()
 	return d.validationError
 }
 
-// Invalidate flushes all values back to zero so as to not allow strategy
-// traversal on compromised data.
+// Invalidate flushes all values back to zero so as to not allow strategy traversal on compromised data.
```
gbjk: 🦍 
* "flushes all values back to zero" => initialises
* I don't think we need to talk about strategies and compromised data. Callers know why they're choosing to call it 🙂 
```suggestion
// Invalidate initialises the Depth, with a error to explain why it was invalid
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -52,16 +51,16 @@ func (m *wsOBUpdateManager) ProcessUpdate(ctx context.Context, g *Gateio, firstU
...
 		return err
 	}
 
 	if lastUpdateID+1 >= firstUpdateID {
 		return applyOrderbookUpdate(g, update)
 	}
 
-	// Orderbook is behind notifications, flush store to prevent trading on stale data
-	if err := g.Websocket.Orderbook.FlushOrderbook(update.Pair, update.Asset); err != nil && !errors.Is(err, buffer.ErrDepthNotFound) {
+	// Orderbook is behind notifications, invalidate store to prevent trading on stale data
```
gbjk: 🦍 
```suggestion
	// Orderbook is behind notifications, therefore Invalidate store
```

gbjk (changes requested):
Good work 👍 

Partial review, have a bit more to review shortly

exchange/websocket/buffer/buffer_types.go
shazbert: It has been moved to the orderbook package as a field in the Update struct as `GenerateChecksum` [here](https://github.com/thrasher-corp/gocryptotrader/pull/1908/files#diff-aa4be819cfc81dcc5b0c661b54f4aab8ce43a8393453b544ef4547f74be89b79R57)

exchange/websocket/buffer/buffer.go
shazbert: [nice catch](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/0e517d58d0fb1d9ed315851f956da94b1faa4747)

exchange/websocket/buffer/buffer.go
shazbert: The system has been optimised to perform most pre-validation including `restSnapshot` check and post-validation tasks, including `checksum` and `verifyOrderbook`, at the depth level. This approach minimises copy/allocations and eliminates mutex interactions, enhancing performance.

exchange/websocket/buffer/buffer_types.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/14a7c6198d30398d9da6f74cb2d1a15cef827569)

exchange/websocket/buffer/buffer.go
shazbert: Its allocated using make which sets it to `w.obBufferLimit` so wiping it this way allows a reduction in allocations and performance improvement, it resets the len value to 0. When appending it just assigns.

exchange/websocket/buffer/buffer.go
shazbert: [See comment](https://github.com/thrasher-corp/gocryptotrader/pull/1908#discussion_r2141784231)

exchange/websocket/buffer/buffer_test.go
shazbert: It just ensures that the update which is out of order does not set the order ID and that it is correctly skipped. When I eventually [remove this](https://github.com/thrasher-corp/gocryptotrader/pull/1908/files#diff-aa4be819cfc81dcc5b0c661b54f4aab8ce43a8393453b544ef4547f74be89b79R83-R89) I can remove this test altogether. It was previously here so I kept it. Most of this package will be repurposed or removed completely into a centralised orderbook manager. Like what is happening in Gateio so other exchanges can use it as well and we can reduce some code foot print. How we are buffering the updates in this package is not what should be happening at all.

exchange/websocket/buffer/buffer.go
shazbert: Yeah I am not worried about this, `Setup` for the buffer is called when `websocket.Manager Setup()` is called in wrapper and being such a hot path I am opting to not introduce checks like that here.

exchanges/orderbook/orderbook.go
shazbert: If you are passing in nil in the param or you are not checking an error then it's okay for this to cease everything. Happy for disagreement here.

exchange/websocket/buffer/buffer.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/ecf037358e5b148a6f98ecd65e783f574f80ab21)

exchange/websocket/buffer/buffer.go
shazbert: Same as the other nil checks I don't think its necessary but others can disagree

exchange/websocket/buffer/buffer.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/73c067777d7cf1d3c4db1149e39355ddb1a9413a)

exchanges/orderbook/depth.go
shazbert: No it's called multiple times in `incremental_updates.go` in function `Depth` method `ProcessUpdate` under the lock.

exchange/websocket/buffer/buffer.go
shazbert: @gbjk @thrasher- On such a hot path for this system I don't think we need to, I am happy to let it panic. There is some comments above brought up from @cranktakular which I have responded. If it's really required I will put them in.

exchange/websocket/buffer/buffer.go
shazbert: Opting to remove this functionality entirely, as we should not be sorting at all.

exchange/websocket/buffer/buffer.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/7ba5c910f34a6073abac36f4065b328dea6fa7b9)

exchange/websocket/buffer/buffer.go
shazbert: No you are correct is [this better?](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/4f3e0a9f6c9beeba204a445876df7849d13614ee)

exchange/websocket/buffer/buffer.go
gbjk: Agreed from me. Since it was me that raised it, I'll resolve.

exchange/websocket/buffer/buffer.go
gbjk: So confused by this suggestion from me. WTF was I thinking 🤷‍♂️
Resolving.

exchange/websocket/buffer/buffer.go
gbjk: @shazbert Did this get removed, or is it outstanding?

exchange/websocket/buffer/buffer.go
gbjk: I'll comment there :)

exchange/websocket/buffer/buffer.go
shazbert: It's outstanding, I would prefer to defer this for another update, if that's okay. If I remove it now I need to check all exchange implementations and that will also increase review toil. My intention was to pull the core update logic out and not change much of the working state of things for this PR.

exchange/websocket/buffer/buffer.go
```diff
@@ -59,256 +45,103 @@ func (w *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandle
...
-	}
-	return nil
-}
-
-// updateByPrice amends amount if match occurs by price, deletes if amount is
-// zero or less and inserts if not found.
-func (o *orderbookHolder) updateByPrice(updts *orderbook.Update) error {
-	return o.ob.UpdateBidAskByPrice(updts)
-}
+	// Clears the buffer of old updates across all pathways. If an error occurs, the buffer is cleared to prepare for a new snapshot.
```
gbjk: 🦍 This is okay. Prefer
```suggestion
	// Always empty the buffer after processing, even if there's an error
```
🐒 I think I'd prefer to move this earlier, after the L120 check for `len(holder.buffer) < w.obBufferLimit`
That way readers would see it early.

exchange/websocket/buffer/buffer.go
```diff
@@ -336,32 +169,35 @@ func (w *Orderbook) LastUpdateID(p currency.Pair, a asset.Item) (int64, error) {
...
+	w.m.RLock()
+	holder, ok := w.ob[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
+	w.m.RUnlock()
 	if !ok {
-		return fmt.Errorf("cannot flush orderbook %s %s %s %w", w.exchangeName, p, a, ErrDepthNotFound)
+		return fmt.Errorf("cannot invalidate orderbook %s %s %s %w", w.exchangeName, p, a, orderbook.ErrDepthNotFound)
 	}
-	// error not needed in this return
-	_ = book.ob.Invalidate(errOrderbookFlushed)
+	// error not needed in this return as it is only used to trigger the Invalidate method
```
gbjk: 🐒  A rule I try to use when writing is to double check if `it` is clear enough about the subject.
In this case, `it` refers to the error, so I had to read several times before I worked out that it meant `this method`
```suggestion
	// Invalidate returns a formatted version of the error it's passed
	// In this context we don't need that, since this method only returns an error if it cannot invalidate
```

I _know_. 🤷‍♂️ Normally I'm making comments shorter, not longer. 🤷

exchanges/orderbook/tranches.go
```diff
@@ -68,9 +68,7 @@ updates:
 			ts[y].StrAmount = updts[x].StrAmount
 			continue updates
 		}
-		return fmt.Errorf("update error: %w ID: %d not found",
-			errIDCannotBeMatched,
-			updts[x].ID)
+		return fmt.Errorf("update error: %w ID: %d not found", errIDCannotBeMatched, updts[x].ID)
```
gbjk: 🐒 
```suggestion
		return fmt.Errorf("update error: %w; ID: %d not found", errIDCannotBeMatched, updts[x].ID)
```

exchanges/orderbook/orderbook_types.go
```diff
@@ -21,13 +21,13 @@ const (
 // Public errors
 var (
 	ErrOrderbookNotFound = errors.New("cannot find orderbook(s)")
+	ErrPriceNotSet       = errors.New("price cannot be zero")
```
gbjk: 🚧 Escalating this from a monkey because I think want to affirm naming convention on errors and descriptions.

Preference for the err price to match the error text. Especially since it might have been _set_ 🤓 

I've been struggling with this a few times, and recently realised that the distinction between two naming and value approaches, naming the situation that occurred, but describing the situation that's prohibited.
There's only implicit guidance on this in [stlye guide](https://google.github.io/styleguide/go/best-practices.html#error-handling), and in effective go, but I think we can follow it generally.

So I also considered `ErrPriceIsZero`

```suggestion
	ErrPriceZero      = errors.New("price cannot be zero")
```

exchanges/orderbook/orderbook_types.go
```diff
@@ -21,13 +21,13 @@ const (
...
+	ErrExchangeNameUnset = errors.New("orderbook exchange name not set")
```
gbjk: 🚧 Above we used `NotSet` and here we used `Unset`.
Admittedly I've pissed all over that conflict by removing the former.
But also want to clarify that the problem is that it's empty, not that it wasn't set.
Alternative: `ErrNoExchangeName`
```suggestion
	ErrExchangeNameEmpty = errors.New("empty orderbook exchange name")
```

exchanges/orderbook/orderbook.go
```diff
@@ -153,28 +154,24 @@ func (b *Base) TotalAsksAmount() (amountCollated, total float64) {
 	return amountCollated, total
 }
 
-// Verify ensures that the orderbook items are correctly sorted prior to being
-// set and will reject any book with incorrect values.
-// Bids should always go from a high price to a low price and
-// Asks should always go from a low price to a higher price
+// Verify ensures that the orderbook items are correctly sorted prior to being set and will reject any book with
+// incorrect values.
```
gbjk: 🦍 
```suggestion
// Verify ensures that the orderbook items are correctly sorted and all fields are valid
```

exchange/websocket/buffer/buffer_test.go
gbjk: Think that means we should say that in both the `NoError` and the `Equal`:
```
	require.NoError(t, err, "Out of sequence Update must not error")
	assert.Equal(t, int64(len(itemArray)+69420), ob.LastUpdateID, "Out of sequence Update should not change LastUpdateID")
```

exchange/websocket/buffer/buffer_test.go
```diff
@@ -571,8 +482,16 @@ func TestFlushBuffer(t *testing.T) {
 	require.NoError(t, err)
 
 	assert.NotEmpty(t, obl.ob)
+
+	holder, ok := obl.ob[key.PairAsset{Base: cp.Base.Item, Quote: cp.Quote.Item, Asset: asset.Spot}]
+	require.True(t, ok)
```
gbjk: 🚧 
```suggestion
	require.NoError(t, err, "createSnapshot must not error")
	require.NotEmpty(t, obl.ob, "createSnapshot must not return empty")

	k := key.PairAsset{Base: cp.Base.Item, Quote: cp.Quote.Item, Asset: asset.Spot}
	holder, ok := obl.ob[k]
	require.True(t, ok, "createSnapshot must return a orderbook for %v", k)
```

exchange/websocket/buffer/buffer_test.go
```diff
@@ -571,8 +482,16 @@ func TestFlushBuffer(t *testing.T) {
...
+
+	holder.buffer = make([]orderbook.Update, 0, 10)
+	holder.buffer = append(holder.buffer, orderbook.Update{})
+
 	obl.FlushBuffer()
-	assert.Empty(t, obl.ob)
+	assert.Empty(t, holder.buffer, 0)
```
gbjk: 🚧 
```suggestion
	assert.Empty(t, holder.buffer, "FlushBuffer should empty buffer")
```

exchange/websocket/buffer/buffer_test.go
```diff
@@ -571,8 +482,16 @@ func TestFlushBuffer(t *testing.T) {
...
+	assert.Equal(t, 10, cap(holder.buffer))
```
gbjk: 🚧 Definite situation where *why* we're testing is better explicitly stated
```suggestion
	assert.Equal(t, 10, cap(holder.buffer), "FlushBuffer should leave the buffer cap to avoid reallocs")
```

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/thrasher-corp/gocryptotrader/currency"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+)
+
+// Action defines a set of differing states required to implement an incoming orderbook update used in conjunction with
+// UpdateEntriesByID
```
gbjk: 🐒 Don't think we need to say this, especially "in conjunction" :)
🥃 I often regret later not suffixing types like this, and their consts.
Consider `ActionType` and `AmendAction`
```suggestion
// Action defines the behaviour of an orderbook update
```

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+const (
+	// Amend applies amount adjustment by ID
+	Amend Action = iota + 1
+	// Delete removes price level from book by ID
+	Delete
+	// Insert adds price level to book
+	Insert
+	// UpdateInsert on conflict applies amount adjustment or appends new amount to book
+	UpdateInsert
+)
```
gbjk: 🚧 
* Comment can be cleaner, since how we identify records which have IDs seems obvious
* Comments on individual consts seem redundant too
* Amend vs Update consistency; Update more consistent with CRUD; Admitting that Insert is better here
* I often regret not suffixing types like this, and their consts, because they're sitting on method/func names
```suggestion
// ActionType defines the behaviour of an orderbook update
type ActionType uint8

// ActionType constants for use with ProcessUpdate
const (
	UnknownAction ActionType = iota
	InsertAction
	UpdateOrInsertAction
	UpdateAction
	DeleteAction
)
```

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+	ErrDepthNotFound = errors.New("orderbook depth not found")
+	ErrEmptyUpdate   = errors.New("update contains no bids or asks")
+)
+
+var (
+	errInvalidAction          = errors.New("invalid action")
+	errAmendFailure           = errors.New("amend update failure")
+	errDeleteFailure          = errors.New("delete update failure")
+	errInsertFailure          = errors.New("insert update failure")
+	errUpdateInsertFailure    = errors.New("update/insert update failure")
```
gbjk: 🐒 Possibly better naming `Failure` 👉 `Failed`
🦍 Not sure if we need to differentiate anything except insert/update vs Delete. I'm easy on this.
```suggestion
	errUpdateFailed           = errors.New("orderbook update failed")
	errDeleteFailed          = errors.New("orderbook update delete failed")
```

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+	// this should screen zero values as well.
+	if u.SkipOutOfOrderLastUpdateID && d.lastUpdateID >= u.UpdateID {
+		return nil
+	}
+
+	if d.options.restSnapshot {
+		return d.invalidate(errRESTSnapshot)
+	}
+
+	if u.Action != 0 {
```
gbjk: 🚧 Can use a const here? `UnknownAction`

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	if u.ExpectedChecksum != 0 {
+		if u.GenerateChecksum == nil {
+			return d.invalidate(errChecksumGeneratorUnset)
+		}
+		if checksum := u.GenerateChecksum(d.snapshot()); checksum != u.ExpectedChecksum {
+			return d.invalidate(fmt.Errorf("%s %s %s %w: expected '%d', got '%d'", d.exchange, d.pair, d.asset, errChecksumMismatch, u.ExpectedChecksum, checksum))
+		}
+	} else if d.verifyOrderbook {
```
gbjk: 🤷 Just to check: These are mutually exclusive? We don't want to verify if we expect a checksum?

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+		if err := verify(d.snapshot()); err != nil {
+			return d.invalidate(err)
+		}
+	}
+
+	return nil
+}
+
+// // TODO: Confirm no alloc as it shouldn't escape
```
gbjk: 🚧 What does this mean?
🦍 Double `//`

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+		IsFundingRate:          d.options.isFundingRate,
+		PriceDuplication:       d.options.priceDuplication,
+		IDAlignment:            d.options.idAligned,
+		ChecksumStringRequired: d.options.checksumStringRequired,
+	}
+}
+
+// updateByIDAndAction will receive an action to execute against the orderbook it will then match by IDs instead of
+// price to perform the action
+func (d *Depth) updateByIDAndAction(u *Update) error {
```
gbjk: 🐒 I think this is just update 🤷

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+			return fmt.Errorf("%w %w", errUpdateInsertFailure, err)
+		}
+	default:
+		return fmt.Errorf("%w [%d]", errInvalidAction, u.Action)
+	}
+	return nil
+}
+
+// updateBidAskByPrice updates the bid and ask spread by supplied updates, this will trim total length of depth level to
+// a specified supplied number
```
gbjk: 🐒  `specified supplied number` doesn't seem correct, since we're not supplying it here
I'm not even sure if we need to restate that; since maxDepth is maxDepth and it being enforced everywhere feels implicit
```suggestion
// updateBidAskByPrice updates the bid and ask spread and enforces Depth.options.maxDepth
```
or
```
// updateBidAskByPrice updates the bid and ask spread
```

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+		return fmt.Errorf("%s %s %s %w", d.exchange, d.pair, d.asset, ErrLastUpdatedNotSet)
+	}
+	d.bidTranches.updateInsertByPrice(update.Bids, d.options.maxDepth)
+	d.askTranches.updateInsertByPrice(update.Asks, d.options.maxDepth)
+	d.updateAndAlert(update)
+	return nil
+}
+
+// updateBidAskByID amends details by ID
+func (d *Depth) updateBidAskByID(update *Update) error {
```
gbjk, line 174: Ironically, I'm okay with this naming here.
Because it contrasts against `updateBidAskByPrice`

exchanges/orderbook/incremental_updates.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+	if err := d.askTranches.updateByID(update.Asks); err != nil {
+		return err
+	}
+	d.updateAndAlert(update)
+	return nil
+}
+
+// deleteBidAskByID deletes a price level by ID
+func (d *Depth) deleteBidAskByID(update *Update, bypassErr bool) error {
```
gbjk: 🐒 
🌴 If accepted, Also applies to the following 2 funcs
```suggestion
// delete deletes a price level by ID
func (d *Depth) delete(update *Update, bypassErr bool) error {
```

gbjk (changes requested):
Good work and thanks for updates.

More feedback 🙂

exchanges/orderbook/incremental_updates.go
shazbert: oooh! Something I didn't do, wanted to check if this struct was actually allocated or was inlined, I will check it soon and RM

exchanges/orderbook/incremental_updates.go
shazbert: Yeah; if we can verify the liquidity represents the exchange via the checksum then I think it should be good, `but` some exchanges only calculate checksum to ~20 length max either side, so if you have a big book it could be corrupted beyond that you wouldn't know, so there's that edge case. 

It might be safer to introduce it into library usage, push it out of the `else` and all externals have the option to turn it off anyway. 

Or optionally even have it all under a `d.verifyOrderbook` check so you can bypass checksum validation as well so you wouldn't need any of that overhead and cowboy it up. 

@thrasher- @gbjk

exchanges/orderbook/incremental_updates.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/6c4ee9b004a56c9f7cca300f99b4c3aeca681d94)

exchanges/orderbook/incremental_updates.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/4a8223f120d9c1d9a1198952b2c6918d49c437c9)

exchanges/orderbook/incremental_updates.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/f13453acb65af78f67e71aca5baae8b94e747631)

exchange/websocket/buffer/buffer_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/4988fbd67d992bd6edd0ec7fb622e13d77ed88d7

exchange/websocket/buffer/buffer.go
gbjk: Agreed. It was an aside.
Resolving.

exchanges/orderbook/incremental_updates.go
gbjk: Okay, so I think this is okay for now.

But to continue the discussion for another change:
Yes, I think that everything related to __validation__ [1️⃣]  should be in one place, and that checksum should probably be alongside other validation.
So that'd mean, as you suggested:
* Call a validate method always
* Conditions branch separatelly for both checksum and other verifications, so can have both, either, neither

---
1️⃣  Verfiy: Something simple truth is evident, Validate: Check something is accurate

exchanges/orderbook/incremental_updates.go
gbjk: 🚧 Still outstanding, right?

gbjk (changes requested):
Looking good, but outstanding and possibly missed line about allocs

exchanges/orderbook/incremental_updates.go
shazbert: Yeah 😭 soz

exchange/websocket/buffer/buffer.go
cranktakular: Are you saying that this sort of check is still done, just in a different part of the codebase?

exchange/websocket/buffer/buffer.go
cranktakular: Ah, the cap's always the same so the performance argument is stronger. Understandable.

exchange/websocket/buffer/buffer.go
shazbert: Correct, its in the orderbook package now. 🚀 Sorry for not being clearer.

exchanges/orderbook/depth.go
cranktakular: Ah, I see. Some of the other method calls in ProcessUpdate need write locks, such as Depth.update() (through depth.updateBidAskById() and then Tranches.updateByID()).

Should we be more careful, and mark those other methods as needing write locks too? Since currently, that warning's only there for invalidate() and updateAndAlert().

exchanges/orderbook/incremental_updates.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1908/commits/24cdb863ff59e21fdc6c0867b12ab2ea0c0fe32e

exchanges/orderbook/depth.go
shazbert: I can if you want me to, I am all for not worrying about it though, because they are unexported functions. Only reason I added it to invalidate is because of the name being the same as an exported func and it can be used across a few things.

exchange/websocket/buffer/buffer.go
cranktakular: Eh, checking if something is nil is extremely quick. Benchmarking it, its presence/absence is undetectable in a ~200 ns/op benchmark (checking if the rand.Intn()'th element in a slice is nil, assigning _ to 5 if so, assigning _ to 6 otherwise // generating rand.Intn(), assigning _ to 6).

I'd think the only performance risk would be if adding the check loses some cache optimisation, which I think is the sort of detail we'd only want to worry about later, especially since it depends on the size of the rest of the function.

shazbert: It is quick, I am more coming from it's a sub system that is already managed and allocated by the websocket package, if it is nil its okay to panic. In both situations, error or panic we would need to fix it and get a PR out for it ASAP. Allowing it to panic just removes the redundant check. @thrasher- @gbjk If you guys want it introduced happy to.

cranktakular (approved):
No real issues. I've noticed some intermittent test failures in engine and exchange wrapper standards, but I may have just missed them on the master branch through chance. They seem unrelated to your changes.

So, tACK it is!

exchanges/binance/binance_websocket.go
```diff
@@ -742,26 +742,19 @@ func (b *Binance) processJob(p currency.Pair) error {
...
-	errClean := b.Websocket.Orderbook.FlushOrderbook(p, asset.Spot)
-	if errClean != nil {
-		log.Errorf(log.WebsocketMgr,
-			"%s flushing websocket error: %v",
-			b.Name,
-			errClean)
+// invalidateAndCleanupOrderbook invalidaates orderbook and cleans local cache
+func (b *Binance) invalidateAndCleanupOrderbook(p currency.Pair) {
+	if err := b.Websocket.Orderbook.InvalidateOrderbook(p, asset.Spot); err != nil {
+		log.Errorf(log.WebsocketMgr, "error invalidating %s websocket orderbook: %v", b.Name, err)
```
thrasher-:
```suggestion
		log.Errorf(log.WebsocketMgr, "%s error invalidating websocket orderbook: %v", b.Name, err)
```

exchanges/binance/binance_websocket.go
```diff
@@ -742,26 +742,19 @@ func (b *Binance) processJob(p currency.Pair) error {
...
 	}
-	errClean = b.obm.cleanup(p)
-	if errClean != nil {
-		log.Errorf(log.WebsocketMgr, "%s cleanup websocket error: %v",
-			b.Name,
-			errClean)
+	if err := b.obm.cleanup(p); err != nil {
+		log.Errorf(log.WebsocketMgr, "error during %s websocket orderbook cleanup: %v", b.Name, err)
```
thrasher-:
```suggestion
		log.Errorf(log.WebsocketMgr, "%s error during websocket orderbook cleanup: %v", b.Name, err)
```

exchanges/btcmarkets/btcmarkets_test.go
```diff
@@ -772,26 +772,13 @@ func TestGetHistoricTrades(t *testing.T) {
...
-		Asks: []orderbook.Level{
-			{Price: 0.3965, Amount: 44149.815},
-			{Price: 0.3967, Amount: 16000.0},
-		},
-		Bids: []orderbook.Level{
-			{Price: 0.396, Amount: 51.0},
-			{Price: 0.396, Amount: 25.0},
-			{Price: 0.3958, Amount: 18570.0},
-		},
+		Asks: orderbook.Levels{{Price: 0.3965, Amount: 44149.815}, {Price: 0.3967, Amount: 16000.0}},
```
thrasher-: I think the previous style was better haha

exchanges/orderbook/incremental_updates.go
thrasher-: I like the idea of all validation/verification all done under a file like validate.go. A funny thing I just realised in this block is that we use the word verify for checking a snapshot, but then call invalidate to invalidate the book. So that'd probably be best changed to `validate(d.snapshot())`. Also a fan of being able to toggle validation of the OB for cowboy action

shazbert: please re-check good sirs 🖖

thrasher-:
Looks like there's some 💥's

https://github.com/shazbert/gocryptotrader/actions/runs/15719996870/job/44298807945#step:11:742 
https://github.com/shazbert/gocryptotrader/actions/runs/15719996870/job/44298807945#step:11:598

thrasher- (approved):
Changes ACK, tests are a happy bing 🌞

exchanges/orderbook/incremental_updates.go
```diff
@@ -97,17 +97,21 @@ func (d *Depth) ProcessUpdate(u *Update) error {
 		}
 	}
 
+	if !d.verifyOrderbook {
```
gbjk: 🚧  I think since we're here and touching things inappropriately, we might as well go full-catholic
Might need to consider overlap of field name and possible method name, but since it's not a method right now that's okay.
```suggestion
	if !d.validateOrderbook {
```

exchanges/orderbook/validate.go
```diff
@@ -0,0 +1,96 @@
+package orderbook
+
+import (
+	"fmt"
+
+	"github.com/thrasher-corp/gocryptotrader/log"
+)
+
+// checker defines specific functionality to determine ascending/descending
+// validation
```
gbjk: 🐒 Given L30, I really can't process the rules your editor is following for wrapping 😂 
```suggestion
// checker defines specific functionality to determine ascending/descending validation
```

exchanges/orderbook/validate.go
```diff
@@ -0,0 +1,96 @@
...
+type checker func(current, previous Level) error
+
+// asc specifically defines ascending price check
+var asc = func(current, previous Level) error {
```
gbjk: 🦍 
```suggestion
// isAsc specifically defines ascending price check
var isAsc = func(current, previous Level) error {
```

exchanges/orderbook/validate.go
```diff
@@ -0,0 +1,96 @@
...
+	if current.Price < previous.Price {
+		return errPriceOutOfOrder
+	}
+	return nil
+}
+
+// dsc specifically defines descending price check
+var dsc = func(current, previous Level) error {
```
gbjk: 🦍 
```suggestion
// isDsc specifically defines descending price check
var isDsc = func(current, previous Level) error {
```

exchanges/orderbook/validate.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return validate(b)
+}
+
+func validate(b *Book) error {
+	// Checking for both ask and bid lengths being zero has been removed and
+	// a warning has been put in place for some exchanges that return zero
+	// level books. In the event that there is a massive liquidity change where
+	// a book dries up, this will still update so we do not traverse potential
+	// incorrect old data.
```
gbjk: 🚧 Can rephrase this to a permanent perspective, please?
I kind of understand what it's saying, but even looking at the introductory commit eb0571cc9 I'm not sure what was removed exactly.

Now that I'm looking at this more, I think the second comment no longer applies at all.
We're talking about "this will still update", but we're not going to update anything that I can see.

So ... saying that we're just warning and it's not an error is self-evident.
The biggest thing I think I'd want to know is why options are exempt.

Something like:
```suggestion
	// Some exchanges may return empty sides, but it's not an error
	// Options have empty sides too frequently for this warning to be useful
```

🚧 I'm not sure if this comment relates to the above or the next lines of code
	// In the event that there is a massive liquidity change where a book dries up, this will still update so we do not traverse potential incorrect old data.
```

File: exchanges/orderbook/validate.go [hunk 65]
```diff
@@ -0,0 +1,96 @@
...
+		return fmt.Errorf(bidLoadBookFailure, b.Exchange, b.Pair, b.Asset, err)
+	}
+	err = checkAlignment(b.Asks, b.IsFundingRate, b.PriceDuplication, b.IDAlignment, b.ChecksumStringRequired, asc, b.Exchange)
+	if err != nil {
+		return fmt.Errorf(askLoadBookFailure, b.Exchange, b.Pair, b.Asset, err)
+	}
+	return nil
+}
+
+// checkAlignment validates full orderbook
```
gbjk: 🚧 Can we reword this comment to be true 😅 
```suggestion
// checkAlignment validates an orderbook side is sequential and does not contain any tribbles
```

exchanges/orderbook/validate.go
```diff
@@ -0,0 +1,96 @@
...
+func checkAlignment(depth Levels, fundingRate, priceDuplication, isIDAligned, requiresChecksumString bool, c checker, exch string) error {
+	for i := range depth {
+		if depth[i].Price == 0 {
+			switch {
+			case exch == "Bitfinex" && fundingRate: /* funding rate can be 0 it seems on Bitfinex */
```
gbjk, line 63: 🗻 🤢  This is weak sauce. Can't see how to fix it quickly, but obviously we don't want to have hardcoding like this.
Pragmatism wins for today.
Maybe we open an issue for it.

gbjk (changes requested):
Good work 👍 

Two change requests and some nitpicks

exchanges/orderbook/validate.go
shazbert: Yeah an issue can be opened

exchange/websocket/buffer/buffer.go
thrasher-: For dynamically created pointers to objects, we should check for nil so in this context `book *orderbook.Book` but I don't think we need it for `w` here. For things where a dev puts nil into a function argument instead of a pointer to an object, I think it should panic since that's obv a dev issue

thrasher-: Just a side note here, Go 1.25 will be introducing compiler level checks for common nil pointer gotchas: https://tip.golang.org/doc/go1.25#compiler

gbjk (approved):

exchange/websocket/buffer/buffer.go
thrasher-: another p.s: I reckon the `(w *Orderbook)` can be changed to `(o *Orderbook)` whilst we're here 😂 
image

gbjk: As it stands, I'd add a `common.NilGuard` on `o` or `w` here, because we do everywhere, and it's not a bad practice.
Can nilguard them both at the same time.
I'd just do it to end the conversation at this point 😂

PR #1912, opened 2025-05-15
types/time: Update `UnmarshalJSON` method to handle all in-between cases 

shazbert:
# PR Description

Some exchanges give back some weird unix floats this should be sufficient to handle permutations a bit better. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/types/time.go
+++ b/types/time.go
@@ -2,10 +2,14 @@ package types
 
 import (
 	"fmt"
-	"math"
 	"strconv"
 	"strings"
 	"time"
+)
+
+const (
+	onePadding = "0"
+	twoPadding = "00"
 )
 
 // Time represents a time.Time object that can be unmarshalled from a float64 or string.
@@ -44,6 +48,14 @@ func (t *Time) UnmarshalJSON(data []byte) error {
 			return fmt.Errorf("%w for `%v`", strconv.ErrSyntax, string(data))
 		}
 		s = s[:target] + s[target+1:]
+
+		// When decimal point is present, the length of the string must be 13, 16, or 19.
+		switch len(s) {
+		case 12, 15, 18:
+			s += onePadding
+		case 11, 14, 17:
+			s += twoPadding
+		}
 	}
 
 	standard, err := strconv.ParseInt(s, 10, 64)
@@ -53,25 +65,12 @@ func (t *Time) UnmarshalJSON(data []byte) error {
 
 	switch len(s) {
 	case 10:
-		// Seconds
 		*t = Time(time.Unix(standard, 0))
-	case 11, 12:
-		// Milliseconds: 1726104395.5 && 1726104395.56
-		*t = Time(time.UnixMilli(standard * int64(math.Pow10(13-len(s)))))
 	case 13:
-		// Milliseconds
 		*t = Time(time.UnixMilli(standard))
-	case 14:
-		// MicroSeconds: 1726106210903.0
-		*t = Time(time.UnixMicro(standard * 100))
 	case 16:
-		// MicroSeconds
 		*t = Time(time.UnixMicro(standard))
-	case 17:
-		// NanoSeconds: 1606292218213.4578
-		*t = Time(time.Unix(0, standard*100))
 	case 19:
-		// NanoSeconds
 		*t = Time(time.Unix(0, standard))
 	default:
 		return fmt.Errorf("cannot unmarshal %s into Time", string(data))
--- a/types/time_test.go
+++ b/types/time_test.go
@@ -47,6 +47,9 @@ func TestUnmarshalJSON(t *testing.T) {
 	require.NoError(t, json.Unmarshal([]byte(`"1726106210903.0"`), &testTime))
 	assert.Equal(t, time.UnixMicro(1726106210903000), testTime.Time())
 
+	require.NoError(t, json.Unmarshal([]byte(`"1747278712.09328"`), &testTime))
+	assert.Equal(t, time.UnixMicro(1747278712093280), testTime.Time())
+
 	// nanoseconds
 	require.NoError(t, json.Unmarshal([]byte(`"1606292218213.4578"`), &testTime))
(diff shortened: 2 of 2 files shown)


types/time_test.go
```diff
@@ -63,13 +66,14 @@ func TestUnmarshalJSON(t *testing.T) {
...
-// 2716176	       441.9 ns/op	     352 B/op	       6 allocs/op (previous)
+// 5697680	       203.8 ns/op	     168 B/op	       2 allocs/op (current)
+// 5030734	       240.1 ns/op	     168 B/op	       2 allocs/op (previous)
 func BenchmarkUnmarshalJSON(b *testing.B) {
 	var testTime Time
 	for b.Loop() {
-		err := json.Unmarshal([]byte(`"1691122380942.173000"`), &testTime)
-		require.NoError(b, err)
+		if err := json.Unmarshal([]byte(`"1691122380942.173000"`), &testTime); err != nil {
+			require.NoError(b, err)
```
thrasher-, line 75:
```suggestion
			b.Fatal(err) 
```
Since in this case we've opted the orig style for perf, we can just default the whole thing to it to bypass require's overhead as we already know an err has occurred. This is probably a good policy only to use for benchmarks so we get a more accurate benchmark

thrasher- (changes requested):
Gotta love exchange things. One nit

thrasher- (approved):
Looking scrumptious!

samuael (commented):
I like the idea of updating this conversion function. I’ve encountered this issue of timestamp information not matching our pre-specified timestamp length.

types/time.go
```diff
@@ -46,33 +50,28 @@ func (t *Time) UnmarshalJSON(data []byte) error {
 		s = s[:target] + s[target+1:]
```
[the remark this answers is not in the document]
shazbert: I haven't seen this occur, could you please give me a string example and where you are seeing this and I can add that to the test and update.

types/time.go
```diff
@@ -2,12 +2,16 @@ package types
...
 	"fmt"
-	"math"
 	"strconv"
 	"strings"
 	"time"
 )
 
+const (
+	onePadding = "0"
+	twoPadding = "00"
```
gbjk, line 12: 🥃 I don't think there's an efficiency or legibility benefit from these constants; They're used once, and I think it's the same to the compiler either way. I wouldn't be adverse to inlining them

types/time.go
```diff
@@ -46,33 +50,28 @@ func (t *Time) UnmarshalJSON(data []byte) error {
...
-	case 17:
-		// NanoSeconds: 1606292218213.4578
-		*t = Time(time.Unix(0, standard*100))
+		*t = Time(time.UnixMicro(unixTS))
 	case 19:
-		// NanoSeconds
-		*t = Time(time.Unix(0, standard))
+		*t = Time(time.Unix(0, unixTS))
 	default:
 		return fmt.Errorf("cannot unmarshal %s into Time", string(data))
```
gbjk, line 76: 🗻 🦍 Pretty sure we don't need to call string when we're also using `%s`
```suggestion
		return fmt.Errorf("cannot unmarshal %s into Time", data)
```

gbjk (approved):
LGTM 👍  Really nice work.

Just nits, entirely optional, and one of them is a OOS super-nit.

types/time.go
shazbert: Yeah good point, I had this because I had a big list at the start then refactored. 🚀

types/time.go
shazbert: nice

types/time.go
```diff
@@ -18,63 +17,43 @@ type Time time.Time
...
-		if r == '.' {
-			return true
-		}
-		// types.Time may only parse numbers. The below check ensures an error is thrown. time.Time should be used to
-		// parse RFC3339 strings instead.
-		badSyntax = r < '0' || r > '9'
-		return badSyntax
-	})
+	switch s {
+	case "null", "0", ``:
```
gbjk, line 25: 🦍 Think backticks were left over from before and we can just consistently use quotation marks.
```suggestion
	case "null", "0", "":
```

types/time.go
```diff
@@ -18,63 +17,43 @@ type Time time.Time
...
+	case 12, 15, 18:
+		s += "0"
+	case 11, 14, 17:
+		s += "00"
+	}
+
+	unixTS, err := strconv.ParseInt(s, 10, 64)
 	if err != nil {
-		return err
+		return fmt.Errorf("%w, expecting unix timestamp", err)
```
gbjk, line 43: 🐒 
```suggestion
		return fmt.Errorf("error parsing unix timestamp: %w", err)
```

gbjk (approved):
LGTM.
Very minor nits.

types/time.go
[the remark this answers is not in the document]
samuael: It is a possibility that I am talking about.
For instance 1606292218213457.8001 will have 20 digit after concatenation if we just concatenate it as it is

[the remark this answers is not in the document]
shazbert: It's possible but I don't think it's probable, because overflowing 19 decimal places you are ~200 years in the future or your going into picoseconds. I think its safe to keep as is and when we come across that issue then we can open a PR for it, just keeps things a tiny bit cleaner. Is that okay? @samuael

types/time.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1912/commits/4973ffb9ec6d9488f18e9f67ef32577157da8283 👀 😆

gbjk (approved):

thrasher- (approved):
Changes ACK

types/time.go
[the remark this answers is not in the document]
samuael: Cool. I think I overthink about it ...

Issue #1762, opened 2024-12-23
Kraken: Fix WsConnect/WsCancelOrder/TestWsAddOrder race


thrasher-:
# New Issue

Race spotted here: https://github.com/thrasher-corp/gocryptotrader/actions/runs/12462044377/job/34782406720#step:10:594

Looks to be because we're using an authToken global var without any sync.Mutex protection

## Context

```
WARNING: DATA RACE
Write at 0x00000187cda0 by goroutine 197:
  github.com/thrasher-corp/gocryptotrader/exchanges/kraken.(*Kraken).WsConnect()
      /home/runner/work/gocryptotrader/gocryptotrader/exchanges/kraken/kraken_websocket.go:115 +0x519
  github.com/thrasher-corp/gocryptotrader/exchanges/kraken.(*Kraken).WsConnect-fm()
      <autogenerated>:1 +0x33
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).connect()
      /home/runner/work/gocryptotrader/gocryptotrader/exchanges/stream/websocket.go:355 +0x87b
  github.com/thrasher-corp/gocryptotrader/exchanges/stream.(*Websocket).Connect()
      /home/runner/work/gocryptotrader/gocryptotrader/exchanges/stream/websocket.go:326 +0x9d
  github.com/thrasher-corp/gocryptotrader/internal/testing/exchange.MockWsInstance[go.shape.struct { github.com/thrasher-corp/gocryptotrader/exchanges.Base },go.shape.*github.com/thrasher-corp/gocryptotrader/exchanges/kraken.Kraken]()
      /home/runner/work/gocryptotrader/gocryptotrader/internal/testing/exchange/exchange.go:115 +0x5fb
  github.com/thrasher-corp/gocryptotrader/exchanges/kraken.TestWsAddOrder()
      /home/runner/work/gocryptotrader/gocryptotrader/exchanges/kraken/kraken_test.go:1172 +0x124
  testing.tRunner()
      /opt/hostedtoolcache/go/1.23.4/x64/src/testing/testing.go:1690 +0x226
  testing.(*T).Run.gowrap1()
      /opt/hostedtoolcache/go/1.23.4/x64/src/testing/testing.go:1743 +0x44

Previous read at 0x00000187cda0 by goroutine 260:
  github.com/thrasher-corp/gocryptotrader/exchanges/kraken.(*Kraken).wsCancelOrder()
[...]

PR #1926, opened 2025-05-28
Kraken: Protect authToken with RWMutex to prevent race

thrasher-:
From Jules:

This commit introduces a sync.RWMutex to protect the global `authToken` variable in `exchanges/kraken/kraken_websocket.go`.

The race condition occurred due to concurrent read/write access to `authToken` from different goroutines, notably between `WsConnect` (write) and functions like `wsCancelOrder`, `wsAddOrder`, `wsCancelAllOrders`, and `manageSubs` (read).

The fix involves:
- Adding `authTokenMutex.Lock()` before writing to `authToken` in `WsConnect` and `authTokenMutex.Unlock()` after.
- Adding `authTokenMutex.RLock()` before reading `authToken` in `wsAddOrder`, `wsCancelOrder`, `wsCancelAllOrders`, and `manageSubs`, and `authTokenMutex.RUnlock()` after.

This change resolves the data race reported in
https://github.com/thrasher-corp/gocryptotrader/issues/1762. I ran tests in the `exchanges/kraken` package with the `-race` detector, and all tests passed without detecting any race conditions.

# PR Description

Race fix

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/exchanges/kraken/kraken.go
+++ b/exchanges/kraken/kraken.go
@@ -39,7 +39,7 @@ const (
 type Kraken struct {
 	exchange.Base
 	wsAuthToken string
-	wsAuthMu    sync.RWMutex
+	wsAuthMtx   sync.RWMutex
 }
 
 // GetCurrentServerTime returns current server time
--- a/exchanges/kraken/kraken_test.go
+++ b/exchanges/kraken/kraken_test.go
@@ -6,6 +6,7 @@ import (
 	"net/http"
 	"os"
 	"strings"
+	"sync"
 	"testing"
 	"time"
 
@@ -1745,6 +1746,34 @@ func TestEnforceStandardChannelNames(t *testing.T) {
 func TestWebsocketAuthToken(t *testing.T) {
 	t.Parallel()
 	k := new(Kraken)
-	k.wsAuthToken = "meep" // k is a local instance, so we can set this lock free
-	assert.Equal(t, "meep", k.websocketAuthToken())
+	k.setWebsocketAuthToken("meep")
+	const n = 69
+	var wg sync.WaitGroup
+	wg.Add(2 * n)
+
+	start := make(chan struct{})
+	for range n {
+		go func() {
+			defer wg.Done()
+			<-start
+			k.setWebsocketAuthToken("69420")
+		}()
+	}
+	for range n {
+		go func() {
+			defer wg.Done()
+			<-start
+			k.websocketAuthToken()
+		}()
+	}
+	close(start)
+	wg.Wait()
+	assert.Equal(t, "69420", k.websocketAuthToken(), "websocketAuthToken should return correctly after concurrent reads and writes")
+}
+
+func TestSetWebsocketAuthToken(t *testing.T) {
+	t.Parallel()
+	k := new(Kraken)
+	k.setWebsocketAuthToken("69420")
+	assert.Equal(t, "69420", k.websocketAuthToken())
 }
--- a/exchanges/kraken/kraken_websocket.go
+++ b/exchanges/kraken/kraken_websocket.go
@@ -119,9 +119,7 @@ func (k *Kraken) WsConnect() error {
 				k.Websocket.SetCanUseAuthenticatedEndpoints(false)
 				log.Errorf(log.ExchangeSys, "%s - failed to connect to authenticated endpoint: %v\n", k.Name, err)
 			} else {
-				k.wsAuthMu.Lock()
-				k.wsAuthToken = authToken
-				k.wsAuthMu.Unlock()
+				k.setWebsocketAuthToken(authToken)
 				k.Websocket.SetCanUseAuthenticatedEndpoints(true)
 				k.Websocket.Wg.Add(1)
 				go k.wsFunnelConnectionData(k.Websocket.AuthConn, comms)
@@ -1411,7 +1409,13 @@ const subTplText = `
 
 // websocketAuthToken retrieves the current websocket session's auth token
 func (k *Kraken) websocketAuthToken() string {
-	k.wsAuthMu.RLock()
-	defer k.wsAuthMu.RUnlock()
+	k.wsAuthMtx.RLock()
+	defer k.wsAuthMtx.RUnlock()
 	return k.wsAuthToken
+}
+
+func (k *Kraken) setWebsocketAuthToken(token string) {
+	k.wsAuthMtx.Lock()
+	k.wsAuthToken = token
+	k.wsAuthMtx.Unlock()
 }
--- a/exchanges/kraken/kraken_wrapper.go
+++ b/exchanges/kraken/kraken_wrapper.go
@@ -1440,9 +1440,7 @@ func (k *Kraken) AuthenticateWebsocket(ctx context.Context) error {
 		return err
 	}
 
-	k.wsAuthMu.Lock()
-	k.wsAuthToken = resp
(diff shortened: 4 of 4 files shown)


exchanges/kraken/kraken_test.go
```diff
@@ -1741,3 +1741,10 @@ func TestEnforceStandardChannelNames(t *testing.T) {
 		assert.ErrorIsf(t, err, subscription.ErrUseConstChannelName, "Private channel names should not be allowed for %s", n)
 	}
 }
+
+func TestWebsocketAuthToken(t *testing.T) {
+	t.Parallel()
+	k := new(Kraken)
+	k.wsAuthToken = "meep" // k is a local instance, so we can set this lock free
+	assert.Equal(t, "meep", k.websocketAuthToken())
```
samuael: If what want to show here is how synchronized the test is by calling `websocketAuthToken()` method, it can be done with several similar calls running in parallel using `t.Run()`, I believe.

samuael (commented):
The PR is clean and very useful. Thank you for inviting me to review this.
I have 1 simple suggestion but it depends up-to you to decide the change. Except for that, the PR LGTM!

exchanges/kraken/kraken_websocket.go
```diff
@@ -112,7 +111,9 @@ func (k *Kraken) WsConnect() error {
 	go k.wsFunnelConnectionData(k.Websocket.Conn, comms)
 
 	if k.IsWebsocketAuthenticationSupported() {
-		authToken, err = k.GetWebsocketToken(context.TODO())
+		k.wsAuthMu.Lock()
+		k.wsAuthToken, err = k.GetWebsocketToken(context.TODO())
```
cranktakular: Is it a good idea to set the token before checking for an error? If it errors, that would zero out the token even if it was already properly set elsewhere.

cranktakular (commented):
tACK (auth aside), got one suggested change for consideration

cranktakular (approved):
Looks good now!

exchanges/kraken/kraken.go
```diff
@@ -37,6 +38,8 @@ const (
 // Kraken is the overarching type across the kraken package
 type Kraken struct {
 	exchange.Base
+	wsAuthToken string
+	wsAuthMu    sync.RWMutex
```
gbjk: 🐒 We don't normally call these `*Mu`
```suggestion
	wsAuthMutex    sync.RWMutex
```

exchanges/kraken/kraken_test.go
```diff
@@ -1741,3 +1741,10 @@ func TestEnforceStandardChannelNames(t *testing.T) {
 		assert.ErrorIsf(t, err, subscription.ErrUseConstChannelName, "Private channel names should not be allowed for %s", n)
 	}
 }
+
+func TestWebsocketAuthToken(t *testing.T) {
+	t.Parallel()
+	k := new(Kraken)
+	k.wsAuthToken = "meep" // k is a local instance, so we can set this lock free
```
gbjk: 🦍 Very self-explanatory 😄 
```suggestion
	k.wsAuthToken = "meep"
```

exchanges/kraken/kraken_test.go
gbjk: 🚧 I agree with the premise here; It's sane to ensure this exercises the lock.
However t.Run is definitely not the right way to achieve that, because there's zero guarantee of parallelism on the lock.
Something like:
```
k.wsAuthMutex.Lock()
c := make(chan string, 1)
go func(){
    c <- k.websocketAuthToken()
}()
runtime.GoSched()
select {
case <-time.After(time.Milisecond):
    // All good
case <-c:
    t.Error("websocketAuthToken must respect wsAuthMutex.Lock")
}
k.wsAuthMutex.Unlock()
```

exchanges/kraken/kraken_websocket.go
```diff
@@ -112,22 +111,17 @@ func (k *Kraken) WsConnect() error {
...
 				k.Websocket.SetCanUseAuthenticatedEndpoints(false)
-				log.Errorf(log.ExchangeSys,
-					"%v - failed to connect to authenticated endpoint: %v\n",
-					k.Name,
-					err)
+				log.Errorf(log.ExchangeSys, "%s - failed to connect to authenticated endpoint: %v\n", k.Name, err)
 			} else {
+				k.wsAuthMu.Lock()
+				k.wsAuthToken = authToken
+				k.wsAuthMu.Unlock()
```
gbjk: 🚧 We do this twice, so can it please be `k.setWSAuthToken`

gbjk (changes requested):
Good work, overlord.

Just some simple change requests.

exchanges/kraken/kraken_test.go
gbjk: I realised that everything I reply with on this PR needs to be a concrete `suggestion` that can be committed, right?

I'll Make-It-So.

exchanges/kraken/kraken_test.go
```diff
@@ -1745,6 +1746,34 @@ func TestEnforceStandardChannelNames(t *testing.T) {
...
+	for range n {
+		go func() {
+			defer wg.Done()
+			<-start
+			k.websocketAuthToken()
+		}()
+	}
+	close(start)
+	wg.Wait()
+	assert.Equal(t, "69420", k.websocketAuthToken(), "websocketAuthToken should return correctly after concurrent reads and writes")
```
gbjk, line 1771: 🚧 This isn't actually guaranteed to be concurrent.
If go-procs is 1, or it just doesn't want/need to context switch, there's no guarantee of a mutex clash.

gbjk, line 1771: 🚧 ~This isn't actually guaranteed to be concurrent.~
~If go-procs is 1, or it just doesn't want/need to context switch, there's no guarantee of a mutex clash.~

~Additionally, we have no evidence in the test that locking is happening. Remove the locks, and the test still passes, right?~

gbjk (changes requested):
Sorry; Change request on the test.

If it's okay, I'll proposal an alternative?

exchanges/kraken/kraken_test.go
gbjk: Sorry, Retracting this; thrasher correctly pointed out that `-race` provides the coverage to ensure that locks are interleaving correctly.

thrasher-: Since the context of this PR is to fix a race, originally the test func was for just test coverage. But @samuael's suggestion was great in that we should also test to make sure it doesn't throw an issue from the race detector. The race detector is probalistic by nature, so this works:

Without the locks:
```sh
go test -timeout 30s -run ^TestWebsocketAuthToken$ github.com/thrasher-corp/gocryptotrader/exchanges/kraken -v -race -count=1
=== RUN   TestWebsocketAuthToken
=== PAUSE TestWebsocketAuthToken
=== CONT  TestWebsocketAuthToken
==================
WARNING: DATA RACE
Write at 0x00c00026b7e0 by goroutine 49:
  github.com/thrasher-corp/gocryptotrader/exchanges/kraken.(*Kraken).setWebsocketAuthToken()
```
With the locks it passes as expected and the race detector is a happy bing. So in conclusion we test for coverage, that we have fixed the original issue (the race) and that we don't have any random deadlocks with the getters/setters firing off a magical 69 * 2 times

PR #1936, opened 2025-06-04
types/time, exchanges/kraken: Refactor spot/futures time types

thrasher-:
# PR Description

image

- Updated WSFuturesTickerData, WsFuturesTradeData, and other related structs to replace int64 timestamps with the new types.Time.
- Adjusted related test cases to accommodate changes in timestamp handling.
- Modified functions in kraken_wrapper and kraken_websocket to utilise the new Time type for better time management.
- Enhanced JSON unmarshalling in the Time type to handle various timestamp formats, including "0" and "0.0".

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist


types/time.go
```diff
@@ -21,8 +21,7 @@ func (t *Time) UnmarshalJSON(data []byte) error {
 		s = s[1 : len(s)-1]
 	}
 
-	switch s {
-	case "null", "0", "":
+	if s == "null" || strings.Trim(s, "0.") == "" {
```
shazbert:
```suggestion
	if s == "" || s[0] == 'n' || s[0] == '0' {
```

more performant and if it starts with 0 it just thinks its cooked. 😆

shazbert (approved):
One suggestion tACK. 🇰🇷 aken strikes

exchanges/kraken/futures_types.go
```diff
@@ -378,7 +378,7 @@ type FuturesFillsData struct {
 	Fills []struct {
```
shazbert, line 378: Out of scope/separate issue: we should rid anon structs because it can be hard to mock externally if required

shazbert (approved):
Changes ACK.

types/time.go
```diff
@@ -21,8 +21,7 @@ func (t *Time) UnmarshalJSON(data []byte) error {
 		s = s[1 : len(s)-1]
 	}
 
-	switch s {
-	case "null", "0", "":
+	if s == "" || s[0] == 'n' || s[0] == '0' {
```
cranktakular: The last condition here would cause this to return an uninitialised time if the timecode has a leading zero. That doesn't sound unlikely enough for me to be comfortable with this.

types/time_test.go
```diff
@@ -54,6 +60,9 @@ func TestUnmarshalJSON(t *testing.T) {
 	require.NoError(t, json.Unmarshal([]byte(`"1606292218213.4578"`), &testTime))
 	assert.Equal(t, time.Unix(0, 1606292218213457800), testTime.Time())
 
+	require.NoError(t, json.Unmarshal([]byte(`"1560516023.070651"`), &testTime))
```
cranktakular: This new test doesn't seem to add any coverage, or account for any functionality you changed. What is it testing for?

exchanges/kraken/kraken_websocket.go
```diff
@@ -365,8 +365,8 @@ func (k *Kraken) wsProcessOpenOrders(ownOrders any) error {
 					LimitPriceUpper:      val.LimitPrice,
 					ExecutedAmount:       val.ExecutedVolume,
 					Fee:                  val.Fee,
-					Date:                 convert.TimeFromUnixTimestampDecimal(val.OpenTime).Truncate(time.Microsecond),
-					LastUpdated:          convert.TimeFromUnixTimestampDecimal(val.LastUpdated).Truncate(time.Microsecond),
```
cranktakular, line 369: Sure you no longer need to truncate a microsecond from these?

cranktakular (changes requested):

types/time.go
thrasher-: I had strings.Trim(s, "0.") beforehand and would be more precise to catch this, but then Ryan suggested just checking for '0' for perf increases and our tests appears to be happy with it

types/time_test.go
thrasher-: We had no exact demonstrations of Kraken's time which is precision to 6 decimal points. So this confirms that.

exchanges/kraken/kraken_websocket.go
thrasher-: Yep, no difference to tests with the precision they return vs expected output (test addition confirms that)

types/time.go
cranktakular: I do still like the n check.

Plus I'd more worry about this leading to silent issues during runtime. Occasionally dropping returned times to zero. But if it doesn't cause any, this would be great.

thrasher-: I agree, reverted back to the more precise check + 'n' for best of both worlds

cranktakular (approved):
tACK, couldn't check a few changes due to lack of credentials, but other than that it worked well!

shazbert (approved):
Thanks doods. 🐄

exchanges/kraken/futures_types.go
gbjk: 🥃 I think this is back to front. When we externally mock things we will remove anonystructs.
Don't think we need to do it prematurely, because it'll just happen anyway.

exchanges/kraken/kraken_test.go
```diff
@@ -1282,7 +1282,7 @@ func TestWsOpenOrders(t *testing.T) {
 				assert.Equal(t, order.Pending, v.Status, "order status")
 				assert.Equal(t, 0.0, v.Price, "price")
 				assert.Equal(t, 0.0001, v.Amount, "amount")
-				assert.Equal(t, time.UnixMicro(1692851641361371).UTC(), v.Date, "Date")
+				assert.Equal(t, time.UnixMicro(1692851641361371).UTC(), v.Date.UTC(), "Date")
```
gbjk, line 1285: 🗻 These aren't great test descriptions

types/time.go
```diff
@@ -21,8 +21,7 @@ func (t *Time) UnmarshalJSON(data []byte) error {
 		s = s[1 : len(s)-1]
 	}
 
-	switch s {
-	case "null", "0", "":
+	if s == "" || s[0] == 'n' || (s[0] == '0' && strings.Trim(s, "0.") == "") {
```
gbjk: 🚧  So we're saying that if it's `0.0000` then it's nil. That's okay.
But we'd also be parsing `0.0.0.0` as nil, and I think that should be an error condition.

So I'd be tempted to move this after L28, remove a single `.` from the input, and then here we can check if it's just 0s with a `TrimPrefix(s, "0") == ""` ( or any other method).
That way we'd still error on L42 if we get whacky input that's probably an IP address or something, by mistake.

types/time_test.go
```diff
@@ -26,6 +26,12 @@ func TestUnmarshalJSON(t *testing.T) {
 	require.NoError(t, json.Unmarshal([]byte(`"0"`), &testTime))
 	assert.Equal(t, time.Time{}, testTime.Time())
 
+	require.NoError(t, json.Unmarshal([]byte(`"0.0"`), &testTime))
+	assert.Equal(t, time.Time{}, testTime.Time())
+
+	require.NoError(t, json.Unmarshal([]byte(`"0.00000"`), &testTime))
+	assert.Equal(t, time.Time{}, testTime.Time())
+
```
gbjk: I'd like to extend this to include the invalid input `0.0.0.0`

gbjk (changes requested):
LGTM 👍 

One change request 🚧 
I considered making it a nitpick, but types.Thyme 🍃  is core, so it doesn't hurt to be a bit pedantic

gbjk (approved):
Thanks 🙏

PR #1940, opened 2025-06-08
cmd/documentation: Add GitHub token support for contributor list retrieval

thrasher-:
# PR Description

Adds optional GitHub authentication support to the documentation tool

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Running the documentation tool

## Checklist

### DIFF
--- a/cmd/documentation/cmd_templates/documentation.tmpl
+++ b/cmd/documentation/cmd_templates/documentation.tmpl
@@ -11,7 +11,7 @@ Be aware, this tool will:
 - Works off a configuration JSON file located at ``gocryptotrader/cmd/documentation/`` for future use with multiple repositories.
 - Automatically find the directory and file tree for the GoCryptoTrader source code and alert you of undocumented file systems which **need** to be updated.
 - Automatically find the template folder tree.
-- Fetch an updated contributor list and rank on pull request commit amount
+- Fetch an updated contributor list and rank on pull request commit amount. Set the `GITHUB_TOKEN` env or use the `ghtoken` command line flag for optional authentication.
 - Sets up core folder docs for the root directory tree including **LICENSE** and **CONTRIBUTORS**
 
 ### config.json example
--- a/cmd/documentation/documentation.go
+++ b/cmd/documentation/documentation.go
@@ -75,6 +75,7 @@ var (
 	// checking
 	ref          = []string{"gocryptotrader", "cmd", "documentation"}
 	engineFolder = "engine"
+	githubToken  = os.Getenv("GITHUB_TOKEN")
 )
 
 // Contributor defines an account associated with this code base by doing
@@ -123,6 +124,7 @@ type Attributes struct {
 func main() {
 	flag.BoolVar(&verbose, "v", false, "Verbose output")
 	flag.StringVar(&toolDir, "tooldir", "", "Pass in the documentation tool directory if outside tool folder")
+	flag.StringVar(&githubToken, "ghtoken", githubToken, "Github authentication token to use when fetching the contributors list")
 	flag.Parse()
 
 	wd, err := os.Getwd()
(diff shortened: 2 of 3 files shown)


cmd/documentation/documentation.go
```diff
@@ -75,6 +75,7 @@ var (
 	// checking
 	ref          = []string{"gocryptotrader", "cmd", "documentation"}
 	engineFolder = "engine"
+	githubToken  = os.Getenv("GITHUB_TOKEN")
```
shazbert: mentioning don't have to act: `ghtoken` will override a set env var

cmd/documentation/cmd_templates/documentation.tmpl
```diff
@@ -11,7 +11,7 @@ Be aware, this tool will:
 - Works off a configuration JSON file located at ``gocryptotrader/cmd/documentation/`` for future use with multiple repositories.
 - Automatically find the directory and file tree for the GoCryptoTrader source code and alert you of undocumented file systems which **need** to be updated.
 - Automatically find the template folder tree.
-- Fetch an updated contributor list and rank on pull request commit amount
+- Fetch an updated contributor list and rank on pull request commit amount. Set the `GITHUB_TOKEN` env or use the `ghtoken` command line flag for optional authentication.
```
shazbert:
```suggestion
- Fetch an updated contributor list and rank on pull request commit amount. Set the `GITHUB_TOKEN` environment variable or use the `ghtoken` command-line flag for optional authentication.
```

shazbert (approved):
approved with optionals

shazbert (approved):
Ch ACK! Thanks!

cranktakular (changes requested):
Seems to function. Breaks when I enter a nonsense token, and works correctly whether I input the correct token or no token at all. Works both when inputted as a flag, and when inputted as an environmental variable. However, I can't tell whether this fetches things better or not.

However, I would request better error handling. When a garbage token is used, the output is:
```
Using GitHub token for authentication
2025/06/17 17:38:12 Documentation Generation Tool - GetContributorList error json: cannot unmarshal object into Go value of type []main.Contributor
exit status 1
```

This is somewhat indicative, but I think it should be more informative, if possible.

thrasher-:
> Seems to function. Breaks when I enter a nonsense token, and works correctly whether I input the correct token or no token at all. Works both when inputted as a flag, and when inputted as an environmental variable. However, I can't tell whether this fetches things better or not.
> 
> However, I would request better error handling. When a garbage token is used, the output is:
> 
> ```
> Using GitHub token for authentication
> 2025/06/17 17:38:12 Documentation Generation Tool - GetContributorList error json: cannot unmarshal object into Go value of type []main.Contributor
> exit status 1
> ```
> 
> This is somewhat indicative, but I think it should be more informative, if possible.

This is for our CI jobs to get better GH rate limits when we specify a token but then extended the option to users if they care for it, but wasn't the point of this PR

cmd/documentation/documentation.go
```diff
@@ -75,6 +75,7 @@ var (
 	// checking
 	ref          = []string{"gocryptotrader", "cmd", "documentation"}
 	engineFolder = "engine"
+	githubToken  = os.Getenv("GITHUB_TOKEN") // Defaults to this, otherwise will be overridden by the -ghtoken flag if set
```
gbjk: 🐒 `Otherwise` is misleading in this comment
I think we could move the whole thing down to the flag:
```suggestion
	githubToken  string
```
But otherwise:
```
	githubToken  = os.Getenv("GITHUB_TOKEN") // Overridden by the ghtoken flag when set
```

cmd/documentation/documentation.go
```diff
@@ -123,6 +130,7 @@ type Attributes struct {
 func main() {
 	flag.BoolVar(&verbose, "v", false, "Verbose output")
 	flag.StringVar(&toolDir, "tooldir", "", "Pass in the documentation tool directory if outside tool folder")
+	flag.StringVar(&githubToken, "ghtoken", githubToken, "Github authentication token to use when fetching the contributors list")
```
gbjk, line 133: 🐒  Could be
```suggestion
	flag.StringVar(&githubToken, "ghtoken", os.Getenv("GITHUB_TOKEN"), "Github authentication token to use when fetching the contributors list")
```

cmd/documentation/documentation.go
```diff
@@ -354,14 +361,25 @@ func GetContributorList(ctx context.Context, repo string, verbose bool) ([]Contr
...
 
-		contents, err := common.SendHTTPRequest(ctx, http.MethodGet, common.EncodeURLValues(repo+GithubAPIEndpoint, vals), nil, nil, verbose)
+		contents, err := common.SendHTTPRequest(ctx, http.MethodGet, common.EncodeURLValues(repo+GithubAPIEndpoint, vals), headers, nil, verbose)
 		if err != nil {
 			return nil, err
 		}
 
+		var g ghError
+		if err := json.Unmarshal(contents, &g); err == nil && g.Message != "" {
+			return nil, fmt.Errorf("GitHub server error message: %s Status: %s", g.Message, g.Status)
```
gbjk: 🐒 
```suggestion
			return nil, fmt.Errorf("GitHub error message: %q Status: %s", g.Message, g.Status)
```

cmd/documentation/documentation.go
thrasher-: Nits addressed, though will pass on moving to the flags since it's used by both the tests and flag in this one simple trick (TM)

gbjk (approved):

PR #1952, opened 2025-06-22
signaler: improve cross-platform signal compatibility

romanornr:
## Signaler Cross-Platform Signal Compatibility Improvements

This pull improves cross-platform signal handling compatibility. The changes introduce runtime-based platform detection and remove problematic signal handling on Unix systems.

## Problem Statement

The previous signaler implementation had cross-platform compatibility issues:

**Unix Signal Handling Issue**: `os.Kill` was included in the signal list for all platforms, but on Unix-based systems (Linux, macOS), `os.Kill` cannot be caught or ignored by signal handlers: https://man7.org/linux/man-pages/man1/kill.1.html

### Runtime-Based Platform Detection

**Added**: `runtime` package import for platform detection.

**Implementation**: `getPlatformSignals()` function now uses `runtime.GOOS` to determine the target platform and configure signals.

```go
if runtime.GOOS == "windows" {
    signals = append(signals, os.Kill)
}
```

###  Platform-Specific Signal Configuration

**Unix Systems** (Linux, macOS, etc.):
- `os.Interrupt` - Ctrl+C interrupt signal
- `syscall.SIGTERM` - Termination request
- `syscall.SIGABRT` - Abort signal
- **Excluded**: `os.Kill` (cannot be caught on Unix)

**Windows Systems**:
- All Unix signals plus `os.Kill` May be useful for Windows-specific scenarios

### Signal Initialization Refactoring

**Before**:
```go
func init() {
    sigs := []os.Signal{
        os.Interrupt,
        os.Kill,
        syscall.SIGTERM,
        syscall.SIGABRT,
    }
    signal.Notify(s, sigs...)
}
```

**After**:
```go
func init() {
    sigs := getPlatformSignals()
    notifier.Notify(s, sigs...)
}
```

### SignalNotifier Interface Introduction

**Added**: `SignalNotifier` interface for signal notification abstraction:

```go
type SignalNotifier interface {
    Notify(c chan<- os.Signal, sig ...os.Signal)
    Stop(c chan<- os.Signal)
}
```

**Implementation**: `osSignalNotifier` struct provides the production implementation:

```go
type osSignalNotifier struct {}

func (o *osSignalNotifier) Notify(c chan<- os.Signal, sig ...os.Signal) {
    signal.Notify(c, sig...)
}

func (o *osSignalNotifier) Stop(c chan<- os.Signal) {
    signal.Stop(c)
}
```

### Unix-Based Systems (Linux, macOS, FreeBSD, etc.)
- **Catchable Signals**: `SIGINT`, `SIGTERM`, `SIGABRT`
- **Non-Catchable**: `SIGKILL` (os.Kill) - handled directly by kernel
- **Behavior**: Signal handlers can intercept and process catchable signals for graceful shutdown

### Windows Systems
- **Available Signals**: All Unix signals plus `os.Kill`
- **Behavior**: Windows signal handling differs from Unix; `os.Kill` may provide additional termination options
- **Compatibility**: Maintains broader signal coverage for Windows-specific scenarios

## Backward Compatibility

✅ **Fully Backward Compatible**
✅ **Public API**: No changes to public functions or exported types
✅ **Behavior**: `WaitForInterrupt()` function maintains identical behavior
✅ **Integration**: Existing code using the signaler package requires no modifications
✅ **Signal Handling**: Core functionality remains unchanged for end users

```

## Changes


**Import Addition**:
```diff
+ "runtime"
```
*Enables platform detection for cross-platform compatibility*

**Signal Configuration Refactoring**:
```diff
- sigs := []os.Signal{
+ sigs := getPlatformSignals()
+ signal.Notify(s, sigs...)
+}
+
+func getPlatformSignals() []os.Signal {
+   signals := []os.Signal{
        os.Interrupt,
-       os.Kill,
        syscall.SIGTERM,
        syscall.SIGABRT,
    }
- signal.Notify(s, sigs...)
```
*Extracts signal configuration into dedicated function with platform awareness*

**Platform-Specific Logic**:
```diff
+   // Add os.Kill only for windows
+   // os.Kill cannot be caught or ignored on Unix-based systems
+   if runtime.GOOS == "windows" {
+       signals = append(signals, os.Kill)
+   }
+   return signals
```
*Adds Windows-specific signal handling while excluding problematic Unix behavior*

**Interface Introduction**:
```diff
+// SignalNotifier is an interface for signal notification
+type SignalNotifier interface {
+   Notify(c chan<- os.Signal, sig ...os.Signal)
+   Stop(c chan<- os.Signal)
+}
```
*Establishes abstraction layer for improved testability and future extensibility*

I have some more improvements planned, including testing, which I want to do later, but it would be easier for me to start with this small PR first. 

- ✅Bug fix (non-breaking change which fixes an issue)
- ✅ Unix systems exclude `os.Kill` from signal handling
- ✅ Windows systems include `os.Kill` for additional termination options
- ✅ All platforms maintain core signal handling functionality
- ✅ No breaking changes to existing integrations

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [x ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] I have added tests that prove my fix is effective or that my feature works (todo)

### DIFF
--- a/signaler/signaler.go
+++ b/signaler/signaler.go
@@ -3,19 +3,36 @@ package signaler
 import (
 	"os"
 	"os/signal"
+	"runtime"
 	"syscall"
 )
 
 var s = make(chan os.Signal, 1)
 
 func init() {
-	sigs := []os.Signal{
+	sigs := getPlatformSignals()
+	signal.Notify(s, sigs...)
+}
+
+func getPlatformSignals() []os.Signal {
+	signals := []os.Signal{
 		os.Interrupt,
-		os.Kill,
 		syscall.SIGTERM,
 		syscall.SIGABRT,
 	}
-	signal.Notify(s, sigs...)
+
+	// Add os.Kill only for windows
+	// os.Kill cannot be caught or ignored on Unix-based systems
+	if runtime.GOOS == "windows" {
+		signals = append(signals, os.Kill)
+	}
+	return signals
+}
+
+// SignalNotifier is an interface for signal notification
+type SignalNotifier interface {
+	Notify(c chan<- os.Signal, sig ...os.Signal)
+	Stop(c chan<- os.Signal)
 }
 
 // WaitForInterrupt waits until a os.Signal is


signaler/signaler.go
```diff
@@ -3,19 +3,36 @@ package signaler
...
+		signals = append(signals, os.Kill)
+	}
+	return signals
+}
+
+// SignalNotifier is an interface for signal notification
+type SignalNotifier interface {
+	Notify(c chan<- os.Signal, sig ...os.Signal)
+	Stop(c chan<- os.Signal)
 }
```
shazbert: Can RM as this is not hooked up with anything

shazbert: optionally: Don't really need global and other things can consolidate with diff

```diff
diff --git a/signaler/signaler.go b/signaler/signaler.go
index 1061e2685..52417c16a 100644
--- a/signaler/signaler.go
+++ b/signaler/signaler.go
@@ -7,36 +7,19 @@ import (
        "syscall"
 )

-var s = make(chan os.Signal, 1)
-
-func init() {
-       sigs := getPlatformSignals()
-       signal.Notify(s, sigs...)
+// WaitForInterrupt waits until a os.Signal is
+// received and returns the result
+func WaitForInterrupt() os.Signal {
+       s := make(chan os.Signal, 1)
+       signal.Notify(s, getPlatformSignals()...)
+       return <-s
 }

 func getPlatformSignals() []os.Signal {
-       signals := []os.Signal{
-               os.Interrupt,
-               syscall.SIGTERM,
-               syscall.SIGABRT,
-       }
-
-       // Add os.Kill only for windows
-       // os.Kill cannot be caught or ignored on Unix-based systems
+       signals := []os.Signal{os.Interrupt, syscall.SIGTERM, syscall.SIGABRT}
+       // Add os.Kill only for windows os.Kill cannot be caught or ignored on Unix-based systems
        if runtime.GOOS == "windows" {
                signals = append(signals, os.Kill)
        }
        return signals
 }
-
-// SignalNotifier is an interface for signal notification
-type SignalNotifier interface {
-       Notify(c chan<- os.Signal, sig ...os.Signal)
-       Stop(c chan<- os.Signal)
-}
-
-// WaitForInterrupt waits until a os.Signal is
-// received and returns the result
-func WaitForInterrupt() os.Signal {
-       return <-s
-}
```

shazbert (changes requested):
Thanks for PR. Interface isn't hooked up so that is a required change, the other is optional.

romanornr:
I actually needed that interface to write some Tests (which I have now)

I could push that commit and it probably gets automatically added into this. If you guys prefer, I can do that now

shazbert:
> I actually needed that interface to write some Tests (which I have now)
> 
> I could push that commit and it probably gets automatically added into this. If you guys prefer, I can do that now

Yeah tests are good, push em 🫡

signaler/signaler.go
```diff
@@ -1,25 +1,88 @@
...
+// Stop stops the signal registration with the OS
+func (o *osSignalNotifier) Stop(c chan<- os.Signal) {
+	signal.Stop(c)
+}
+
+// NewSignalNotifier creates a new SignalNotifier
+// This function returns a SignalNotifier that uses the operating system's signal handling mechanism
+func NewSignalNotifier() SignalNotifier {
+	return &osSignalNotifier{}
 }
```
shazbert: remove as its not being used

signaler/signaler.go
```diff
@@ -1,25 +1,88 @@
...
+		signals = append(signals, os.Kill)
+	}
+	return signals
+}
+
+// SignalNotifier is an interface for signal notification
+type SignalNotifier interface {
+	Notify(c chan<- os.Signal, sig ...os.Signal)
+	Stop(c chan<- os.Signal)
+}
```
shazbert: we can remove this as we can inject directly into the global `s` to test ` WaitForInterrupt`. Then its just the getPlatform signals which can be tested by our github actions and a runtime.Goos check. This will collapse a lot of bloat as what your testing signals is done in the signal package itself.

shazbert (changes requested):
@romanornr There is a lot of bloat code here for such a minor change. Please consider reducing this to it's bare bones. For example the readme is self explained and doesn't need to be introduced with such verbosity.  Plus some other nits.

romanornr:
I am a changed man from now on 
I won't be the author of confusion 

See new commit

signaler/signaler_test.go
```diff
@@ -0,0 +1,38 @@
...
+		done <- sig
+	}()
+
+	// Give the waiter time to register signal.Notify
+	time.Sleep(50 * time.Millisecond)
+
+	proc, err := os.FindProcess(os.Getpid())
+	if err != nil {
+		t.Fatalf("failed to find process: %v", err)
+	}
```
shazbert: can use test helpers require/assert so this will collapse to:

```suggestion
	require.NoError(t, err, "os.FindProcess must not error"
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,38 @@
...
+
+	if err := proc.Signal(syscall.SIGTERM); err != nil {
+		t.Fatalf("failed to send SIGTERM: %v", err)
```
shazbert: use `require.NoError`

signaler/signaler_test.go
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	select {
+	case sig := <-done:
+		if sig != syscall.SIGTERM {
+			t.Fatalf("expected SIGTERM, got %v", sig)
+		}
+	case <-time.After(2 * time.Second):
+		t.Fatalf("timeout waiting for WaitForInterrupt to return")
+	}
```
shazbert: Can use `require.Eventually` for the entire case with a 2 second timeout or use `require.Equal`

thrasher-:
@codex please review

signaler/signaler_test.go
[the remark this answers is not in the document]
romanornr: 👍

signaler/README.md
```diff
@@ -0,0 +1,60 @@
...
+    * Windows:
+        * Receiving: Go maps console events (Ctrl_C, Ctrl_Break) to os.Interrupt. Our handler will receive os.Interrupt.
+        * SIGTERM is included but typically not delivered on windows.
+        * os.Process.Kill forcibly terminates the process and cannot be caught.
+    * Go:
+        * The only signal values guaranteed to exist in the os package on all systems are os.Interrupt and os.Kill. We still register syscall.SIGTERM for Unix portability. On windows, it is usually not delivered but harmless.
+    
+ * Unix/macOS listen for SIGINT and SIGTERM by default. SIGKILL and SIGSTOP are uncatchable by design and are never registered.
+ * Windows maps console events (Ctrl+C, Ctrl+Break) to os.Interrupt including. SIGTERM is harmess but typically not delivered on windows.
+ * SIGABRT excluded to preserve the default abort+core-dump diagnostics.
```
gbjk: 🚧  Change requests
* We seem to be saying the same thing three times in three paragraphs here
* I don't think we should be explaining the specifics of what we can't do, or what system signal processing is outside our control. It's not normal for apps to describe what they don't process, and there's other sigs that we don't handle which you haven't mentioned (e.g. SIGILL)
* Receives is implicit - all apps receive the signal. We want to say what we do on receiving it.
* I'm kinda thinking we don't need to say anything. The behaviour of all apps on receiving these signals is very well understood to be "cleanup and exit". If we're doing that, then ... we don't need to say that we're being good citizens.
* Honestly ... I think you can remove this entire README.md. I'm sorry to say that because you obviously spent time on it.

signaler/signaler.go
```diff
@@ -6,20 +7,17 @@ import (
...
-func init() {
-	sigs := []os.Signal{
-		os.Interrupt,
-		os.Kill,
-		syscall.SIGTERM,
-		syscall.SIGABRT,
-	}
-	signal.Notify(s, sigs...)
+// WaitForInterrupt blocks until a termination signal is received and returns it.
+// It registers a temporary channel, unregisters it via signal.Stop() and returns the received signal.
```
gbjk: 🚧 This comment is about how the function behaves internally, which isn't information callers need to know.
```suggestion
```

signaler/signaler.go
```diff
@@ -6,20 +7,17 @@ import (
...
 }
 
-// WaitForInterrupt waits until a os.Signal is
-// received and returns the result
-func WaitForInterrupt() os.Signal {
-	return <-s
+// getPlatformSignals returns the appropriate signals to listen for on the current platform.
+func getPlatformSignals() []os.Signal {
+	return []os.Signal{os.Interrupt, syscall.SIGTERM}
 }
```
gbjk: Am I missing something, or ... aren't we always returning the same values?
So ... This function can evaporate and the signal slice can move to L14.

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"os"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWaitForInterrupt_SIGTERM(t *testing.T) {
```
gbjk, line 15: 🚧 Unusual naming... why would we test just one, and why put it in the test name?
```suggestion
func TestWaitForInterrupt(t *testing.T) {
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
...
+	done := make(chan os.Signal, 1)
+
+	go func() {
+		sig := WaitForInterrupt()
+		done <- sig
+	}()
+
+	// Give the waiter time to register signal.Notify
+	time.Sleep(50 * time.Millisecond)
```
gbjk: 🦍 Optional super pedantic nitpick, but the Waiter doesn't need time, it needs a context switch yield.
Better is a way to know the waiter is ready. But that's too complicated to justify changing runtime implementation just to make tests more inspectable for a _very rare_ race.
So my suggestion is a slight change to description, and replacing arbitrary time with specifically yielding.
```suggestion
	// Yield to allow the waiter to register signal.Notify
	runtime.Gosched()
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	proc, err := os.FindProcess(os.Getpid())
+
+	require.NoError(t, err, "os.FindProcess must not error")
+	require.NoError(t, proc.Signal(syscall.SIGTERM), "os.FindProcess must not error")
```
gbjk: 🚧 
```suggestion
	require.NoError(t, proc.Signal(syscall.SIGTERM), "proc.Signal must not error")
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	select {
+	case sig := <-done:
+		require.Equal(t, sig, syscall.SIGTERM, "expected SIGTERM, got %v", sig)
```
gbjk: 🚧 
* Isn't require.Equal going to tell us that?
* And I think the description coming out of Equal cover not needing any description at all
* But also, I think you meant `Equalf`
* And order is wrong
```suggestion
		require.Equal(t, syscall.SIGTERM, sig)
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
...
+	case <-time.After(2 * time.Second):
+		t.Fatalf("timeout waiting for WaitForInterrupt to return")
```
gbjk: 🚧 This is a `require.Eventually` in disguise 🙂 
See if you can manage that, but shout to me if you want help with it.

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
+//go:build !windows
+// +build !windows
```
gbjk: 🚧 We should not be double bagging. We pin go to a recent version, currently 1.25 and [+build has been deprecated since 1.18](https://tip.golang.org/doc/go1.18#go-build-lines)
```suggestion
//go:build !windows
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
+//go:build !windows
```
gbjk: 🚧 This feels wrong.
I don't think we should be avoiding building a test.
I think we should skip the tests inside it based on the platform instead.
If the entire feature was optional, we'd use build tags, but this doesn't feel right.

gbjk (changes requested):
Great work 👍 

Sorry to have some significant changes, because this is a good contribution. Just a little _too_ good 😂

signaler/README.md
```diff
@@ -0,0 +1,45 @@
+# Signaler
```
gbjk: 🚧 Change request:
@romanornr I'm going to stick to the original change request, and simplify it:

Please remove this README.md 🙏 

* We don't need it, it's not telling us anything that we should have in a readme separate for a package, and it's actually contrary to the idiom of generating docs from `docs/` with footer and header sections added

signaler/signaler.go
gbjk: 🚧 Change request:
This change request is outstanding.

signaler/signaler.go
```diff
@@ -6,20 +7,12 @@ import (
...
-// WaitForInterrupt waits until a os.Signal is
-// received and returns the result
+// WaitForInterrupt blocks until a termination signal is received and returns it.
+// It registers a temporary channel, unregisters it via signal.Stop() and returns the received signal.
 func WaitForInterrupt() os.Signal {
-	return <-s
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	s := <-c
+	signal.Stop(c) // unregister to avoid keeping channel referenced/registered
```
gbjk: 🐒 Optional nitpick:
* defer makes it cleaner when/why sometimes
* Comment doesn't help us
```suggestion
	defer signal.Stop(c)
	s := <-c
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,76 @@
...
+	"os"
+	"runtime"
+	"syscall"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestWaitForInterrupt_SIGTERM(t *testing.T) {
```
gbjk: 🐒 Optional nitpick:
These 2 tests are so similar.
Can you unify them with a for loop about what the signal is?

signaler/signaler_test.go
```diff
@@ -0,0 +1,76 @@
...
+	var got os.Signal
+	require.Eventually(t, func() bool {
+		select {
+		case got = <-done:
+			return true
+		default:
+			return false
+		}
+	}, 2*time.Second, 10*time.Millisecond, "timeout waiting for WaitForInterrupt to return")
+	require.Equalf(t, syscall.SIGTERM, got, "expected SIGTERM, got %v", got)
```
gbjk: 🐒 Optional nitpick:
This doesn't need to store `got` and test it after.
That can happen inside the require.
```suggestion
	require.Eventually(t, func() bool {
		select {
		case got = <-done:
			return got == syscall.SIGTERM
		default:
			return false
		}
	}, 2*time.Second, 10*time.Millisecond, "timeout waiting for WaitForInterrupt to receive signal")
```

gbjk: 🥃 Just a note:
* When using require like this, you wouldn't use `Equalf` because we already get told what the value is
* We try to use the form "must XYZ" with positive statements for test descriptions

But I'm not asking for this change - I'm just highlighting that if this code *wasn't* removed as suggested above, then it should follow this pattern.

```suggestion
	require.Equal(t, syscall.SIGTERM, got, "must receive correct signal")
```

signaler/README.md
shazbert: Agreed RM please

signaler/signaler_test.go
romanornr: Am I typing this wrong. I don't get it. I would understand case := <- done but how can got =<- done work?

image

romanornr: I used equalf because when I did it, one of these git checks started to complain

signaler/signaler.go
romanornr: I thought of using defer back when writing this code but ended up not using defer 

Thought "pretty straight forward man, can avoid defer and save 0.00000000001 on overhead" lol (but sure will add defer)

signaler/signaler_test.go
romanornr: Oh wait I see, you're not using %v

romanornr: and even if it is with got := 

I'd still get issues here
are we sure to don't use var got os.Signal? 

image

signaler/signaler_test.go
```diff
@@ -0,0 +1,76 @@
...
+
+	// Yield to allow the waiter to register signal.Notify
+	runtime.Gosched()
+
+	proc, err := os.FindProcess(os.Getpid())
+	require.NoError(t, err, "os.FindProcess must not error")
+	if err := proc.Signal(os.Interrupt); err != nil {
+		// On Windows, delivering os.Interrupt programmatically via os.Process.Signal
+		// may not be supported. If so, skip rather than fail the test as the feature
+		// is still valid (Ctrl+C generates os.Interrupt).
```
gbjk: So, in summary... SIGTERM won't work. And SIGINT might not work.
Excellent.

gbjk: Given this, I can't see why we can't just deliver both signals the same in the test and skip either of them that errors 🤷‍♂️
Though ... that makes this the most relaxed test evah 😂

signaler/signaler.go
gbjk: 🚧 Change request:
I actually think we shouldn't do this.
It's the only thing that's stopping us providing a cleaner API where the caller actually knows that it's ready to proceed, without blocking.

gbjk (changes requested):
🚧 Change request:

Please consider [these proposed API changes](https://github.com/gbjk/gocryptotrader/commit/4971d6f7a)

You can apply this patch with:
```
curl -s https://github.com/gbjk/gocryptotrader/commit/4971d6f7a.patch | git am
```

signaler/signaler.go
```diff
@@ -6,20 +7,9 @@ import (
...
-		syscall.SIGABRT,
-	}
-	signal.Notify(s, sigs...)
-}
-
-// WaitForInterrupt waits until a os.Signal is
-// received and returns the result
-func WaitForInterrupt() os.Signal {
-	return <-s
+// WaitForInterrupt returns a channel to receive termination signalls
```
shazbert:
```suggestion
// WaitForInterrupt returns a channel to receive termination signals
```

signaler/signaler_test.go
```diff
@@ -0,0 +1,37 @@
...
+	for _, sig := range []os.Signal{syscall.SIGTERM, os.Interrupt} {
+		sigC := WaitForInterrupt()
+		proc, err := os.FindProcess(os.Getpid())
+		require.NoError(t, err, "os.FindProcess must not error")
+
+		if err := proc.Signal(sig); err != nil {
+			if runtime.GOOS == "windows" {
+				t.Skipf("proc.Signal(%s) not supported on Windows: %v", sig, err)
+			}
+			require.NoError(t, err, "proc.Signal(%s) must not error", sig)
```
shazbert:
```suggestion
			require.NoErrorf(t, err, "proc.Signal(%s) must not error", sig)
```

gbjk: Sorry, mea culpa 🤚

signaler/signaler.go
```diff
@@ -6,20 +7,9 @@ import (
...
+func WaitForInterrupt() chan os.Signal {
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
+	return c
```
samuael, line 14: Since we are only using the detected signals from the channel and not used the channel (`chan`) from outside this function, why can't we just put the blocking channel listening inside this method and just return the `os.Signal`.

In both cases the method call `WaitForInterrupt()` is serving as a blocking method call.


```suggestion
	return <-c
```

It also limits the scope of the channel variable to only inside the function.

samuael (changes requested):
This is nice and clean. 🫡 .
Just left an optional suggestion

signaler/signaler.go
gbjk: You're ~basically~ asking to revert the change I asked for: https://github.com/gbjk/gocryptotrader/commit/4971d6f7a
This gives a cleaner API, and allows callers to use `select` on it.
It's just "the right thing to do".

TL;DR: Please don't 🙏

romanornr: So just to be clear, only @shazbert changes should be applied?
And i can just press on "commit suggestion" button

shazbert: @romanornr Yes please 👍

romanornr: done

shazbert (approved):
tACK. Thanks.

gloriousCode (approved):
tACK!
Now that this is complete @romanornr , can you fix the market? You had your methods before, think its time again

PR #1959, opened 2025-06-27
Engine: Assertify apiserver_test

gbjk:
Nothing to see here.

image

### DIFF
--- a/engine/apiserver_test.go
+++ b/engine/apiserver_test.go
@@ -46,9 +46,7 @@ func TestStartRESTServer(t *testing.T) {
 
 	m.remoteConfig.DeprecatedRPC.Enabled = true
 	err = m.StartRESTServer()
-	if err != nil {
-		t.Fatal(err)
-	}
+	assert.NoError(t, err)
 }
 
 func TestStartWebsocketServer(t *testing.T) {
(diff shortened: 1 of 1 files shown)


shazbert (approved):
ch ACK.

engine/apiserver_test.go
```diff
@@ -124,93 +122,66 @@ func TestWebsocketStop(t *testing.T) {
...
+	require.NoError(t, err, "NewExchangeByName must not error")
 	bs.SetDefaults()
 	err = man.Add(bs)
 	require.NoError(t, err)
 
 	resp := getAllActiveOrderbooks(man)
-	if resp == nil {
-		t.Error("expected not nil")
-	}
+	require.NotNil(t, resp)
```
samuael: could use `assert`

engine/apiserver_test.go
```diff
@@ -124,93 +122,66 @@ func TestWebsocketStop(t *testing.T) {
...
+	require.NoError(t, err, "NewExchangeByName must not error")
 	bs.SetDefaults()
 	err = man.Add(bs)
 	require.NoError(t, err)
 
 	resp := getAllActiveTickers(man)
-	if resp == nil {
-		t.Error("expected not nil")
-	}
+	require.NotNil(t, resp)
```
cranktakular: Was originally t.Error(), so should be assert not require. There's a few other instances of this that should also be fixed.

cranktakular (changes requested):
Got a few rote nits I want, other than that I can tACK this.

engine/apiserver_test.go
gbjk: Thanks. I've considered the suggestion.
* I think resp being nil would be a show-stopper here
* Anything added in future by other devs is most likely to be touching resp, so this is defensive programming 🙂 

Please resolve if you're swayed by my reasoning, or reply if you don't.

engine/apiserver_test.go
gbjk: Thanks; Same reply as above: I've considered the suggestion.
* I think resp being nil would be a show-stopper here
* Anything added in future by other devs is most likely to be touching resp, so this is defensive programming 🙂

Please resolve if you're swayed by my reasoning, or reply if you don't.

cranktakular: Eh, I think expansions on this could plausibly call under different situations, ignoring resp or getting their own to tinker around with. So I'm not swayed on that front.

Plus, this doesn't apply to one of the other places I was thinking of (line 183, where a require is placed for an error check). From a quick look of the surrounding code, I could imagine require saving concerns there, but I'd want you to explicitly consider it first.

gbjk: Okay. We don't agree, but I'll must make-it-so for expediency.

Fixed gbjk@cb7e5e685

engine/apiserver_test.go
gbjk: Fixed gbjk@cb7e5e685

engine/apiserver_test.go
cranktakular: This doesn't change the require on line 183; do you have any thoughts on the necessity of that one?

gbjk: TL;DR: Don't want to change `require.NoError` and regretting changing `require.NotNil` now

Yep: I didn't agree about not needing to require on Nil, but was okayish with it.
My mind regarding errors is "Is this thing a showstopper, or is it a casual error and we'd want to see the rest of the test run".
Nils when you don't expect them are likely to be a beartrap. If it's nil and you expected it to not be, you don't expect more things in the test to run, because anything after is likely to blow up.
I'm saying that now and I still don't agree with the changes I made 😂 

Putting it another way:
Everything could be a require, right?
The reason we use assert is for convenience when we want to see a wider scope of what went wrong.
Like "A few values were wrong, but otherwise everything was okay".

thrasher- (approved):
Changes ACK, agree with the require left for writeResponse err checking. Other changes look fine since I last looked at this

PR #1968, opened 2025-07-06
exchanges: Limit mock test JSON data size by truncating slices and maps

samuael:
ሰላም 🕊️  ከናንተ ጋር ዪሁን ወንድሞቼ

This PR introduces improvements to mock test data recording to reduce file size and improve efficiency:
- Adds logic to truncate slices and maps during mock recording.
- Updates `HTTPRecord`, `CheckResponsePayload`, `CheckJSON`, and related functions to support the new truncation logic.
- Adjusts unit tests under the mock package to align with the updated behavior.
- Ensures recorded mock files remain concise without sacrificing necessary test context.

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test X

## Checklist


exchanges/binance/binance_live_test.go
```diff
@@ -46,6 +46,6 @@ func TestMain(m *testing.M) {
 	if err := e.UpdateTradablePairs(context.Background(), true); err != nil {
 		log.Fatalf("Binance UpdateTradablePairs error: %s", err)
 	}
-
+	b.HTTPRecording = true
```
shazbert: Is this intentional?

exchanges/mock/common_test.go
```diff
@@ -105,14 +62,10 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
 	}
 
 	payload, err := json.Marshal(test1)
-	if err != nil {
-		t.Error("marshal error", err)
-	}
+	assert.NoErrorf(t, err, "marshal error: %v", err)
```
shazbert: require?

exchanges/mock/common_test.go
```diff
@@ -125,17 +78,9 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
 	}
 
 	payload, err = json.Marshal(test2)
-	if err != nil {
-		t.Error("marshal error", err)
-	}
+	assert.NoErrorf(t, err, "marshal error: %v", err)
```
shazbert: require?

exchanges/mock/common_test.go
```diff
@@ -125,17 +78,9 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
-	if err != nil {
-		t.Error("DeriveURLValsFromJSON error", err)
-	}
-
-	if vals["val"][0] != "1" {
-		t.Error("DeriveURLValsFromJSON unexpected value",
-			vals["val"][0])
-	}
+	assert.NoErrorf(t, err, "DeriveURLValsFromJSON error: %v", err)
+	assert.Equalf(t, "1", vals["val"][0], "DeriveURLValsFromJSON unexpected value: ^%v", vals["val"][0])
```
shazbert: vals["val"] need len check or it will panic on failure

exchanges/mock/recording.go
```diff
@@ -245,54 +239,38 @@ func HTTPRecord(res *http.Response, service string, respContents []byte) error {
...
-func CheckResponsePayload(data []byte) ([]byte, error) {
-	items, err := GetExcludedItems()
-	if err != nil {
-		return nil, err
-	}
-
+func CheckResponsePayload(data []byte, items Exclusion) ([]byte, error) {
 	var intermediary any
-	err = json.Unmarshal(data, &intermediary)
+	err := json.Unmarshal(data, &intermediary)
```
shazbert: if short

shazbert (changes requested):
Nice not much from me. Thanks Boss.

gloriousCode (changes requested):
This is lovely! Minimising the giant json test files down would great. Would you be able to re-record an exchange with it and ensure all tests pass with the change? I would like it for all exchanges, but let's stick to one at first

edit: you are on the case already 
# 🫡  🕊️

shazbert (changes requested):
Massive reduction nice work! Just need to fix up Binance tests and should be good to go.

exchanges/binance/binance_test.go
```diff
@@ -1090,12 +1090,7 @@ func TestGetExchangeInfo(t *testing.T) {
 	t.Parallel()
 	info, err := e.GetExchangeInfo(t.Context())
 	require.NoError(t, err, "GetExchangeInfo must not error")
-	if mockTests {
```
shazbert, line 1093: reinstate mock as this seems like it will eventually break

exchanges/binance/binance_test.go
```diff
@@ -1105,6 +1100,10 @@ func TestFetchTradablePairs(t *testing.T) {
 		t.Error("Binance FetchTradablePairs(asset asets.AssetType) error", err)
 	}
 
+	enabledPair, err := e.GetEnabledPairs(asset.Spot)
+	require.NoError(t, err)
+	println(enabledPair[0].String())
```
shazbert: RM

exchanges/binance/cfutures_types.go
```diff
@@ -77,7 +77,7 @@ type FuturesCandleStick struct {
 
 // UnmarshalJSON unmarshals FuturesCandleStick data from JSON
 func (f *FuturesCandleStick) UnmarshalJSON(data []byte) error {
-	return json.Unmarshal(data, &[11]any{&f.OpenTime, &f.Open, &f.High, &f.Low, &f.Close, &f.Volume, &f.CloseTime, &f.BaseAssetVolume, &f.NumberOfTrades, &f.TakerBuyVolume, &f.TakerBuyBaseAssetVolume})
+	return json.Unmarshal(data, &[12]any{&f.OpenTime, &f.Open, &f.High, &f.Low, &f.Close, &f.Volume, &f.CloseTime, &f.BaseAssetVolume, &f.NumberOfTrades, &f.TakerBuyVolume, &f.TakerBuyBaseAssetVolume})
```
shazbert: revert

exchanges/binance/binance_test.go
```diff
@@ -1163,7 +1162,10 @@ func TestGetAggregatedTrades(t *testing.T) {
 }
 
 func TestGetSpotKline(t *testing.T) {
-	t.Parallel()
+	if !mockTests {
+		t.Parallel()
+		e.MockDataSliceLimit = 24
+	}
```
shazbert: a bunch of these should be reverted and verbose removed

exchanges/binance/binance.go
```diff
@@ -709,12 +709,13 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ePath exchange.URL, path
...
-		Verbose:       e.Verbose,
-		HTTPDebugging: e.HTTPDebugging,
-		HTTPRecording: e.HTTPRecording,
+		Method:             http.MethodGet,
+		Path:               endpointPath + path,
+		Result:             result,
+		Verbose:            e.Verbose,
+		HTTPDebugging:      e.HTTPDebugging,
+		HTTPRecording:      e.HTTPRecording,
+		MockDataSliceLimit: e.MockDataSliceLimit,
```
shazbert: Either this change needs to be replicated across the entire code for `MockDataSliceLimit`, or you drop these fields and use a context by value specifically for the mock system and then it will be constrained per specific request/nested requests for that call.

exchanges/binance/binance_live_test.go
```diff
@@ -40,7 +40,8 @@ func TestMain(m *testing.M) {
 			}
 		}
 	}
-
+	ctx := context.Background()
```
shazbert: re-use the context below or just nest the context.Background call

exchanges/binance/binance_mock_test.go
```diff
@@ -28,10 +28,10 @@ func TestMain(m *testing.M) {
 	if err := testexch.MockHTTPInstance(e); err != nil {
 		log.Fatalf("Binance MockHTTPInstance error: %s", err)
 	}
-
+	ctx := context.Background()
```
shazbert: re-use the context below or just nest the context.Background call

shazbert (changes requested):
Thanks for the changes, just some nits for the recent commits.

exchanges/binance/binance.go
samuael: I believe having this field is relevant wherever there is the HTTPRecording field in the codebase.

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -249,6 +249,19 @@ Please refer to our [coding guidelines](/docs/CODING_GUIDELINES.md).
...
+	item := &request.Item{  
+		Method:        		http.MethodGet,
+		Path:          		path,
+		Result:        		result,
+		Verbose:       		e.Verbose,
+		HTTPDebugging: 		e.HTTPDebugging,
+		HTTPRecording: 		e.HTTPRecording,
+		MockDataSliceLimit: e.MockDataSliceLimit,
+	}
+=======
```
shazbert: Conflict artifact

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -259,6 +272,7 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, path string, result any)
         HTTPDebugging: e.HTTPDebugging,
         HTTPRecording: e.HTTPRecording,
     }
+>>>>>>> dcf596c72bb4c6e5f6531d9e583a0307cbefa34a
```
shazbert: 👀

exchanges/binance/binance.go
shazbert: Its still relevant in a context, so instead of having to worry about having those fields correctly set, we can just force a context check, then you can set up a global context with value with a struct with those fields with `func mock.Context(ctx, mock.Fields) ctx` for example or even just for one test function. What you have done is sufficient though. 🚀

shazbert (changes requested):
One this left over then looks good to go, thanks boss

exchanges/binance/binance.go
samuael: I am not sure as to whether I understood what you mean ... but If you considered these values as Global they are not. Or could you explain what you would like in a code(better way of communication) especially where and how the Mock Values/Struct to context setting function should be implemented?

shazbert: Now you have done what you have done, the context isn't needed it but its basically just this

```go
type Mock struct {
	HTTPDebugging      bool
	HTTPRecording      bool
	MockDataSliceLimit int
}

const contextMockFlag = "mock"

func WithMock(ctx context.Context, mock *Mock) context.Context {
	return context.WithValue(ctx, contextMockFlag, mock)
}

func GetMock(ctx context.Context) (*Mock, bool) {
	mock, ok := ctx.Value(contextMockFlag).(*Mock)
	return mock, ok
}
```

Then have a global context or in test invoke `ctx = request.WithMock(ctx, &Mock{values})` like how we have `request.WithVerbose` and `account.DeployCredentialsToContext` Then all the fields can be removed for request.Item.

shazbert (approved):
ch ACK. 🚀

exchanges/binance/binance.go
samuael: This is a very holy idea. Don't you think we may need to repeatedly call the **mock.WithMock()** in every unit test

shazbert: Yeah most likely need a global ctx, but defer this, can be done later when needed. especially for mocking websocket requests/responses.

gbjk: I agree with @shazbert 's direction, but I'd also fear like him that "we're not ready yet", because we don't yet propagate ctx everywhere.
Nearly every single time I've written a func or method in my own code and thought "This one definitely _definitely_ does not need context" it's turned out to be wrong 😂 
That's partly because I pull the logging context, trace and sampling, out of context, but it's also true for anything like exactly this.

In summary: Let's get there so we don't have to worry about whether we have context everywhere we need to do stuff like this.

Resolving.

exchanges/binance/binance_test.go
```diff
@@ -2322,7 +2322,11 @@ func TestGetHistoricCandles(t *testing.T) {
 	}
 
 	startTime = time.Date(2020, 1, 1, 0, 0, 0, 0, time.UTC)
-	_, err := e.GetHistoricCandles(t.Context(), currency.NewBTCUSDT(), asset.Spot, kline.Interval(time.Hour*7), startTime, end)
+
+	enabledPairs, err := e.GetEnabledPairs(asset.Spot)
+	require.NoError(t, err)
```
gbjk: 🚧 🐒 🌴  Please add comments where the failure is not related to the unit test.
They really help when the test fails to immediately see the cause without needing to check the SHA and line number. But also it just helps be clear about exact expectations, which are _sometimes_ unclear.
This will apply to all `assert` and `require` that are not exercising the function named in the test, and whenever the test is not self explanatory.
This nitpick has been upgraded to a change request because I've said it for months and I think it's always actioned, so I'll drop the pretence it's optional 😂 
```suggestion
	require.NoError(t, err, "GetEnabledPairs must not error")
```

exchanges/exchange_types.go
```diff
@@ -256,6 +257,10 @@ type Base struct {
 	ValidateOrderbook bool
 	order.ExecutionLimits
 
+	// Used in HTTP recording to limit the number of slices per record,
+	// helping reduce the size of the recorded mock data.
+	MockDataSliceLimit int
```
gbjk: 🚧  The main change here was to use `HTTPRecording` so people will immediately associate this with the field above.
But in doing so I wanted to reduce it to a suffix comment so we can move it up and put it next to HTTPRecording so everything is more self explanatory.
Can you move this to line 244?
```suggestion
	HTTPMockDataSliceLimit int	// Use with HTTPRecording to reduce the size of recorded mock data
```

exchanges/mock/common_test.go
```diff
@@ -17,69 +20,24 @@ func TestMatchURLVals(t *testing.T) {
...
-			expected,
-			received)
-	}
-
-	received = MatchURLVals(nonceVal1, nonceVal2)
-	if received != expected {
-		t.Errorf("MatchURLVals error expected %v received %v",
-			expected,
-			received)
+	args := []struct {
```
gbjk:
```suggestion
	tests := []struct {
```

exchanges/mock/common_test.go
```diff
@@ -17,69 +20,24 @@ func TestMatchURLVals(t *testing.T) {
...
+		Src1     url.Values
+		Src2     url.Values
+		Expected bool
```
gbjk: 🚧
* It's usual to name comparrisson vars `a` and `b`, and that's probably better here than `Src1`
* Don't need to export
```suggestion
		a     url.Values
		b     url.Values
		exp bool
```

exchanges/mock/common_test.go
```diff
@@ -17,69 +20,24 @@ func TestMatchURLVals(t *testing.T) {
...
+		{testVal2, testVal, false},
+		{testVal, testVal3, false},
+		{nonceVal1, testVal2, false},
+		{emptyVal, emptyVal, true},
+		{testVal, testVal, true},
+		{nonceVal1, nonceVal2, true},
+	}
+	for a := range args {
+		received := MatchURLVals(args[a].Src1, args[a].Src2)
+		assert.Equalf(t, args[a].Expected, received, "MatchURLVals error expected %v received %v", args[a].Expected, received)
```
gbjk: 🚧 
* Please don't use `Equalf` and then embed the same values we're testing. It's exactly what `assertify` is for 🙂 
* In this case, the message doesn't help us at all. We'll just see "expected true received false"
* Consistent naming of `tc` or `tt` please, as before 👍 
```suggestion
	for _, tc := range tests {
		got := MatchURLVals(tc.a, tc.b)
		assert.Equalf(t, tc.exp, got, "MatchURLVals should return correctly for (%q, %q)", tc.a, tc.b)
```

exchanges/mock/common_test.go
```diff
@@ -105,14 +63,10 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
 	}
 
 	payload, err := json.Marshal(test1)
-	if err != nil {
-		t.Error("marshal error", err)
-	}
+	require.NoErrorf(t, err, "marshal error: %v", err)
```
gbjk: 🚧 🌴 This change applies to several places below. Please look for any further `assert.*f` or `require.*f` that incorrectly include the test case again.
```suggestion
	require.NoErrorf(t, err, "json.Marshal must not error")
```

exchanges/mock/common_test.go
```diff
@@ -125,17 +79,13 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
-	}
-
-	if vals["val"][0] != "1" {
-		t.Error("DeriveURLValsFromJSON unexpected value",
-			vals["val"][0])
+	require.NoErrorf(t, err, "marshal error: %v", err)
+
+	values, err := DeriveURLValsFromJSONMap(payload)
+	require.NoErrorf(t, err, "DeriveURLValsFromJSON error: %v", err)
+	require.Len(t, values, 7)
```
gbjk: 🚧 If this fails, `require.Len` is going to give us a mess to read.
Better to use `Equal` + `len` when the slice contains a lot of elements or possibly large data.
```suggestion
	require.Equal(t, 7, len(values), `DeriveURLValsFromJSONMap must return the correct number of values")
```

exchanges/mock/common_test.go
```diff
@@ -125,17 +79,13 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
+	for key, val := range values {
+		require.Len(t, val, 1)
+		assert.Equalf(t, val[0], test2[key], "DeriveURLValsFromJSON unexpected value: ^%v", val[0])
```
gbjk: 🤷‍♂️ Note: The previous comment about Equalf should catch this, but also to note that we had a rogue `^` here.

exchanges/mock/recording.go
```diff
@@ -30,7 +30,7 @@ type HTTPResponse struct {
 
 // HTTPRecord will record the request and response to a default JSON file for
 // mocking purposes
```
gbjk, line 35:
```suggestion
// mocking purposes
// mockDataSliceLimit defaults to 5
```

exchanges/mock/recording.go
```diff
@@ -47,7 +47,10 @@ func HTTPRecord(res *http.Response, service string, respContents []byte) error {
 		return errors.New("service not supplied cannot access correct mock file")
 	}
 	service = strings.ToLower(service)
-
+	if mockDataSliceLimit == 0 {
+		// Set the mock slice data size limit to a default of 5
```
gbjk: 🚧 
* This comment belongs in the function comments, which I've suggested above 👍

exchanges/mock/recording.go
```diff
@@ -47,7 +47,10 @@ func HTTPRecord(res *http.Response, service string, respContents []byte) error {
...
+		mockDataSliceLimit = 5
```
gbjk: 🐒 If we make this a const then we don't need a comment at this point
```suggestion
		mockDataSliceLimit = defaultDataSliceLimit
```

exchanges/mock/recording.go
```diff
@@ -311,22 +291,37 @@ const (
...
+func CheckJSON(data any, excluded *Exclusion, limit int) (any, error) {
+	if value, ok := data.([]any); ok {
 		var sData []any
-		for i := range d {
-			v := d[i]
-			switch v.(type) {
-			case map[string]any, []any:
-				checkedData, err := CheckJSON(v, excluded)
+		for i := range value {
+			switch subvalues := value[i].(type) {
```
gbjk: 🚧 This is not `subvalues`; it's just one 🙂 
```suggestion
			switch subvalue := value[i].(type) {
```

exchanges/mock/recording.go
```diff
@@ -311,22 +291,37 @@ const (
...
+			case []any:
+				if len(subvalues) > 0 {
+					if _, ok := subvalues[0].(map[string]any); ok && len(subvalues) > limit {
```
gbjk: 🚨 🚧 Don't we need to check that `limit` is not zero here?
Though I'd be fine with bailing on the func when limit is 0.
Or if the func is private and called from somewhere we know we've ensured that limit is not zero, that's okay.
~Though that makes me wonder if that defaulting would be better here?~
It's recursive, so that wouldn't be great

exchanges/mock/recording.go
```diff
@@ -311,22 +291,37 @@ const (
 )
 
 // CheckJSON recursively parses json data to retract keywords, quite intensive.
-func CheckJSON(data any, excluded *Exclusion) (any, error) {
-	if d, ok := data.([]any); ok {
+func CheckJSON(data any, excluded *Exclusion, limit int) (any, error) {
```
gbjk, line 296: 🤷 Does this need to be exported?

exchanges/mock/recording_test.go
```diff
@@ -3,39 +3,36 @@ package mock
...
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/encoding/json"
 )
 
 func TestGetFilteredHeader(t *testing.T) {
+	items, err := GetExcludedItems()
+	require.NoError(t, err)
+	assert.NotNil(t, items)
```
gbjk:
```suggestion
	assert.NotNil(t, items, "GetExcludedItems should not return nil")
```

exchanges/mock/recording_test.go
```diff
@@ -3,39 +3,36 @@ package mock
...
-	"strings"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/encoding/json"
 )
 
 func TestGetFilteredHeader(t *testing.T) {
+	items, err := GetExcludedItems()
```
gbjk: 🤷 Does this func need to be exported?

exchanges/mock/recording_test.go
```diff
@@ -3,39 +3,36 @@ package mock
...
-	fMap, err := GetFilteredHeader(&resp)
-	if err != nil {
-		t.Error(err)
-	}
-
-	if fMap.Get("Key") != "" {
-		t.Error("risky vals where not replaced correctly")
-	}
+	fMap := GetFilteredHeader(&resp, items)
+	assert.Empty(t, fMap.Get("Key"), "risky vals where not replaced correctly")
```
gbjk: 🚧 test descriptions should follow the positive format, not negative format.
i.e. State what should happen, instead of what should not.
```suggestion
	assert.Empty(t, fMap.Get("Key"), "risky values should be removed")
```

exchanges/mock/recording_test.go
```diff
@@ -3,39 +3,36 @@ package mock
...
-	cleanVals, err := GetFilteredURLVals(shadyVals)
-	if err != nil {
-		t.Error("GetFilteredURLVals error", err)
-	}
-
-	if strings.Contains(cleanVals, superSecretData) {
-		t.Error("Super secret data found")
-	}
+	cleanVals := GetFilteredURLVals(shadyVals, items)
+	assert.NotContains(t, cleanVals, superSecretData, "Super secret data found")
```
gbjk: 🚧 As above. Exercise for reader.
>  test descriptions should follow the positive format, not negative format.
i.e. State what should happen, instead of what should not.
```suggestion
	assert.NotContains(t, cleanVals, superSecretData, "exclusion real_name should be removed")
```

exchanges/mock/recording.go
samuael: Maybe! That was not my decision and I didn't think of unexporting it because I always think there may be a situation when we need to use these methods from outside

exchanges/mock/recording_test.go
samuael: Same answer for this. Maybe in the future they may be needed from outside the package

exchanges/mock/common_test.go
```diff
@@ -105,14 +63,10 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
-	if err != nil {
-		t.Error("marshal error", err)
-	}
+	require.NoError(t, err, "json.Marshal must not error")
 
 	_, err = DeriveURLValsFromJSONMap(payload)
-	if err != nil {
-		t.Error("DeriveURLValsFromJSON error", err)
-	}
+	assert.NoError(t, err, "DeriveURLValsFromJSON error should not error")
```
gbjk: 🥃 Just an aside that the test comment is optional here, because it's in a unit test of the same name

exchanges/mock/common_test.go
```diff
@@ -105,14 +63,10 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
 	}
```
gbjk, line 60: 🗻 🥃 Just realised how unusual the above test struct is.
Bonus points if you drive by abstract the type and make it saner.

Also really don't have much idea what's going on with this test - Is it ... ensuring that it doesn't blow up when given ... invalid data ... or ?

exchanges/mock/common_test.go
gbjk: 🗻 🚧
* I'm really not sure this test is exercising anything meaningful at all
* Just spent a bit longer ~groking~ (urgh. Can't say *that* anymore!) understanding this test, and realised that `test2` is actually the expectation.
```suggestion
		assert.Equalf(t, test2[key], val[0], "DeriveURLValsFromJSON should return the correct value for %s", key)
```

But actually I think it's worth reviewing / rewriting this whole test.
So this is kinda both ... out of scope and a change re-request at the same time.
At the least, let's fix the Equalf.

exchanges/mock/recording.go
```diff
@@ -311,22 +294,37 @@ const (
...
-		for i := range d {
-			v := d[i]
-			switch v.(type) {
-			case map[string]any, []any:
-				checkedData, err := CheckJSON(v, excluded)
+		for i := range value {
+			switch subvalue := value[i].(type) {
+			case []any:
+				if len(subvalue) > 0 {
+					if _, ok := subvalue[0].(map[string]any); ok && len(subvalue) > limit {
```
gbjk: 🚧 I can't see that [this thread](https://github.com/thrasher-corp/gocryptotrader/pull/1968#discussion_r2275295228) has been resolved?

exchanges/mock/recording.go
gbjk: Has this been addressed?

exchanges/mock/recording_test.go
gbjk: I don't think that's a good reason to export things. 😄

But don't action it if you don't want.

exchanges/mock/recording.go
gbjk: I don't think that's a good reason to export things. 😄

But don't action it if you don't want.

exchanges/mock/recording_test.go
```diff
@@ -3,39 +3,36 @@ package mock
...
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/encoding/json"
 )
 
 func TestGetFilteredHeader(t *testing.T) {
+	items, err := GetExcludedItems()
+	require.NoError(t, err)
+	assert.NotNil(t, items, "GetExcludedItems should not error")
```
gbjk: 🚧 This is not what I added as a suggestion. I think it belongs on the line above, and this one should be about returning not-nil.

gbjk (changes requested):
Thanks for those changes.

You've asked for re-review, but not all of the threads were actioned.
I've copied them over for clarity. Please check again 😄

exchanges/mock/recording.go
samuael: Just for safety I have added the limit >0 check. Thanks!

exchanges/mock/recording.go
```diff
@@ -30,7 +33,8 @@ type HTTPResponse struct {
 
 // HTTPRecord will record the request and response to a default JSON file for
 // mocking purposes
-func HTTPRecord(res *http.Response, service string, respContents []byte) error {
+// mockDataSliceLimit defaults to 5
+func HTTPRecord(res *http.Response, service string, respContents []byte, mockDataSliceLimit int) error {
```
gloriousCode, line 37: [optional] If you could write a test and have it cover this function in a predictable way,  I'd really appreciate it.
If in the process of writing a test, you change this large function into smaller, tested pieces then that's also cool.

exchanges/mock/recording.go
```diff
@@ -311,22 +294,37 @@ const (
...
-			v := d[i]
-			switch v.(type) {
-			case map[string]any, []any:
-				checkedData, err := CheckJSON(v, excluded)
+		for i := range value {
+			switch subvalue := value[i].(type) {
+			case []any:
+				if limit > 0 && len(subvalue) > 0 {
+					if _, ok := subvalue[0].(map[string]any); ok && len(subvalue) > limit {
+						subvalue = subvalue[:limit]
```
gloriousCode: [required] This area of code does not receive any test coverage from this entire package. Please add

gloriousCode (changes requested):
I love this PR. I used to remove json data manually like an idiot. Nice work

shazbert (approved):
ch ACK.

exchanges/mock/common_test.go
```diff
@@ -42,20 +42,17 @@ func TestMatchURLVals(t *testing.T) {
 }
 
 func TestDeriveURLValsFromJSON(t *testing.T) {
+	type data struct {
```
gbjk: 🐒 Sorry; `data` is bad name for a type.
( _IMO it's rarely a good name for a variable, either_ )
```suggestion
	type test struct {
```

exchanges/mock/recording_test.go
```diff
@@ -58,6 +60,13 @@ func TestCheckResponsePayload(t *testing.T) {
 	assert.Equal(t, expected, string(data))
 }
 
+func TestGetExcludedItems(t *testing.T) {
+	exclusionList, err := GetExcludedItems()
+	require.NoErrorf(t, err, "GetExcludedItems error: %v", err)
```
gbjk: 🚧 
* Please do not use `assertify.*f` with the values that will be printed.
This is a regression - I'm confused why it keeps coming back.
[It was raised on this comment earlier](https://github.com/thrasher-corp/gocryptotrader/pull/1968#discussion_r2275280382).
* If we were going to have a comment, it should say "must not error"
```suggestion
	require.NoError(t, err)
```

exchanges/mock/recording_test.go
```diff
@@ -58,6 +60,13 @@ func TestCheckResponsePayload(t *testing.T) {
...
+	assert.NotEmpty(t, exclusionList.Headers)
+	assert.NotEmpty(t, exclusionList.Variables)
```
gbjk: 🚧 
```suggestion
	assert.NotEmpty(t, exclusionList.Headers, "Headers must not be empty")
	assert.NotEmpty(t, exclusionList.Variables, "Variables must not be empty")
```

exchanges/mock/recording_test.go
```diff
@@ -67,95 +76,70 @@ type TestStructLevel0 struct {
...
-			StringVal: "somestringstuff",
-			FloatVal:  3.14,
-		},
-		{
-			StringVal: "somestringstuff",
-			FloatVal:  3.14,
-			IntVal:    1337,
+		MixedSlice: []any{
+			[]map[string]any{{"id": 0}, {"id": 2}, {"id": 3}, {"id": 4}, {"id": 5}, {"id": 6}, {}},
+			[]any{float64(1586994000000), "6615.23000000", 'a', 1234, false, 17866372632, 0},
```
gbjk: 🚧 🚨  Test failure on 32-bit

exchanges/mock/recording_test.go
```diff
@@ -67,95 +76,70 @@ type TestStructLevel0 struct {
...
+		IntVal:    1337,
+	},
+	{},
+	{},
+	{},
+}
 
+func TestCheckJSON(t *testing.T) {
 	exclusionList, err := GetExcludedItems()
 	assert.NoErrorf(t, err, "GetExcludedItems error: %v", err)
```
gbjk: 🚧 DriveBy please
```suggestion
	require.NoError(t, err, "GetExcludedItems must not error")
```

exchanges/mock/recording_test.go
```diff
@@ -180,26 +164,42 @@ func TestCheckJSON(t *testing.T) {
 	assert.Len(t, newStruct, 4)
 	assert.Empty(t, newStruct[0].StructVal.BadVal, "Value not wiped correctly")
 	assert.Empty(t, newStruct[0].StructVal.BadVal2, "Value not wiped correctly")
-	assert.Empty(t, newStruct[0].StructVal.OtherData[0].BadVal, "Value not wiped correctly")
-	assert.Empty(t, newStruct[0].StructVal.OtherData[0].BadVal2, "Value not wiped correctly")
-	assert.Empty(t, newStruct[0].StructVal.OtherData[0].OtherData.BadVal, "Value not wiped correctly")
-	assert.Empty(t, newStruct[0].StructVal.OtherData[0].OtherData.BadVal2, "Value not wiped correctly")
-	assert.Len(t, newStruct[0].StructVal.OtherData, 4)
+	assert.Empty(t, newStruct[0].StructVal.OtherData.BadVal, "Value not wiped correctly")
+	assert.Empty(t, newStruct[0].StructVal.OtherData.BadVal2, "Value not wiped correctly")
```
gbjk:
```suggestion
	assert.Empty(t, newStruct[0].StructVal.OtherData.BadVal, "BadVal should be removed")
	assert.Empty(t, newStruct[0].StructVal.OtherData.BadVal2, "BadVal2 should be removed")
```

exchanges/mock/recording_test.go
```diff
@@ -180,26 +164,42 @@ func TestCheckJSON(t *testing.T) {
...
-	exclusionList, err := GetExcludedItems()
-	require.NoErrorf(t, err, "GetExcludedItems error: %v", err)
-	assert.NotEmpty(t, exclusionList.Headers)
-	assert.NotEmpty(t, exclusionList.Variables)
+	_, err = os.Stat(filePath)
+	require.NoError(t, err, "file not created properly")
+
+	defer func() {
+		_ = os.Remove(filePath)
+		_ = os.Remove(outputDirPath)
```
gbjk: 🚧 Surprised there's no linter warnings on these.
Is it possible to wrap them in NoError?
```suggestion
		require.NoErrorf(t, os.Remove(filePath), "Remove test exclusion file %q must not error", filePath)
		require.NoErrorf(t, os.Remove(outputDirPath), "Remove test exclusion dir %q must not error", outputDirPath)
```

exchanges/mock/recording_test.go
```diff
@@ -180,26 +164,42 @@ func TestCheckJSON(t *testing.T) {
...
+	}()
+
+	content, err := json.Marshal(testVal)
+	require.NoError(t, err, "json.Marshal nust not error")
+	assert.NotNil(t, content, "json.Marshal nust not return nil")
```
gbjk: 🚧 
Is the second one definitely a assert?
Feels like it's a require.
```suggestion
	require.NoError(t, err, "json.Marshal must not error")
	assert.NotNil(t, content, "json.Marshal should not return nil")
```

exchanges/mock/common_test.go
gbjk: Okay. This wasn't addressed but nevermind.

exchanges/mock/recording_test.go
gbjk: 🚧 This was not addressed ... it was in [this comment](https://github.com/thrasher-corp/gocryptotrader/pull/1968#discussion_r2280421704)

exchanges/mock/common_test.go
samuael: `'test'` is a keyword my holy brother 😁

exchanges/mock/server_test.go
```diff
@@ -38,89 +37,54 @@ func TestNewVCRServer(t *testing.T) {
...
-	if err != nil {
-		t.Fatal("marshal error", err)
-	}
+	require.NoError(t, err, "json.Marshal must not error")
 
 	err = os.WriteFile(testFile, payload, os.ModePerm)
-	if err != nil {
-		t.Fatal("marshal error", err)
-	}
+	require.NoError(t, err, "json.Marshal must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "WriteFile must not error")
```

exchanges/mock/server_test.go
```diff
@@ -38,89 +37,54 @@ func TestNewVCRServer(t *testing.T) {
...
 
 	_, err = common.SendHTTPRequest(t.Context(),
 		http.MethodGet,
 		"http://localhost:300/somethingElse?"+queryString,
 		nil,
 		bytes.NewBufferString(""), true)
-	if err == nil {
-		t.Error("Sending http request expected an error")
-	}
+	assert.Error(t, err, "Sending http request expected an error")
```
thrasher-:
```suggestion
	assert.Error(t, err, "SendHTTPRequest expected an error")
```

Our style is method, func or variable goes frst followed by the rest, will need to check this for the others on this PR

exchanges/mock/server_test.go
```diff
@@ -38,89 +37,54 @@ func TestNewVCRServer(t *testing.T) {
...
-	if res.Amount != 1 {
-		t.Error("response error expected 1 but received:",
-			res.Amount)
-	}
-
-	if res.Currency != "bitcoin" {
-		t.Error("response error expected \"bitcoin\" but received:",
-			res.Currency)
-	}
+	assert.NoErrorf(t, err, "unmarshal error: %v", err)
```
thrasher-:
```suggestion
	assert.NoError(t, err, "Unmarshal should not error")
```

exchanges/mock/server_test.go
```diff
@@ -38,89 +37,54 @@ func TestNewVCRServer(t *testing.T) {
...
+	assert.Equalf(t, 8000.0, res.Price, "response error expected 8000 but received: %f", res.Price)
+	assert.Equalf(t, 1.0, res.Amount, "response error expected 1 but received: %f", res.Amount)
+	assert.Equalf(t, "bitcoin", res.Currency, "response error expected \"bitcoin\" but received: %s", res.Currency)
 
 	// clean up test.json file
 	err = os.Remove(testFile)
-	if err != nil {
-		t.Fatal("Remove error", err)
-	}
+	require.NoErrorf(t, err, "Remove error: %v", err)
```
thrasher-: Same here

exchanges/mock/recording_test.go
```diff
@@ -81,130 +84,122 @@ type TestStructLevel1 struct {
...
-	if err != nil {
-		t.Error("GetExcludedItems error", err)
-	}
-
-	vals, err := CheckJSON(testVal, &exclusionList)
-	if err != nil {
-		t.Error("Check JSON error", err)
-	}
+	vals, err := CheckJSON(input, &exclusionList, 4)
+	assert.NoErrorf(t, err, "Check JSON error: %v", err)
```
thrasher-: Even though these are carry over messages from the prior code, can be adjusted to:

`assert.NoError(t, err, "CheckJSON should not error")`

For these and others

thrasher- (changes requested):
Thank you for tackling this issue, OCD was triggered hard with the massive JSON files.

Tests and coverage all look good despite the shedding.

Just some minor nits; great work!

exchanges/mock/common_test.go
gbjk: Okay.... I don't know what `class` means in this context, but I'll resolve.

exchanges/mock/recording_test.go
```diff
@@ -3,39 +3,38 @@ package mock
...
 
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/encoding/json"
 )
 
 func TestGetFilteredHeader(t *testing.T) {
+	items, err := getExcludedItems()
+	require.NoError(t, err, "getExcludedItems must not error")
+	assert.NotNil(t, items)
```
gbjk: 🚧 Sorry, still not addressed.
```suggestion
	assert.NotNil(t, items, "getExcludedItems should not return nil")
```

exchanges/mock/recording_test.go
```diff
@@ -3,39 +3,38 @@ package mock
...
-		t.Error("risky vals where not replaced correctly")
-	}
+	fMap := GetFilteredHeader(&resp, items)
+	assert.Empty(t, fMap.Get("Key"), "risky values should be removed")
 }
 
 func TestGetFilteredURLVals(t *testing.T) {
+	items, err := getExcludedItems()
+	require.NoError(t, err, "getExcludedItems must not error")
+	assert.NotNil(t, items)
```
gbjk:
```suggestion
	assert.NotNil(t, items, "getExcludedItems should not return nil")
```

exchanges/mock/recording_test.go
```diff
@@ -81,130 +84,122 @@ type TestStructLevel1 struct {
...
-	if len(exclusionList.Variables) == 0 {
-		t.Error("Variable exclusion list not popoulated")
+	response := &http.Response{
+		Request: &http.Request{
+			Method: http.MethodGet,
+			URL:    &url.URL{},
+		},
 	}
+	err = HTTPRecord(response, "mock", content, 4)
+	require.NoError(t, err)
```
gbjk:
```suggestion
	require.NoError(t, err, "HTTPRecord must not error")
```

exchanges/mock/server_test.go
```diff
@@ -24,9 +25,7 @@ const (
 
 func TestNewVCRServer(t *testing.T) {
 	_, _, err := NewVCRServer("")
-	if err == nil {
-		t.Error("NewVCRServer error cannot be nil")
-	}
+	assert.Error(t, err, "NewVCRServer error cannot be nil")
```
gbjk: 🚧 What error are we expecting here?
Please use `assert.ErrorIs` if possible, or `assert.ErrorContains` if it's really not possible.
I normally slam in io.EOF and see what I get back and go hunting from there.
Also should have description saying why this error is there, if it's not obvious
```suggestion
	assert.ErrorIs(t, err, errNameMustMentionPenguins)
```

exchanges/mock/server_test.go
gbjk: 🚧 🌴 Same as above please.
Please look for any other examples I've missed where we are just saying "it must error".
Our test *must* always say what error they're expecting, and why, if it's not self-explanatory from the error name or text.

exchanges/mock/server_test.go
```diff
@@ -38,89 +37,54 @@ func TestNewVCRServer(t *testing.T) {
...
 		bytes.NewBufferString(""), true)
-	if err != nil {
-		t.Error("Sending http request error", err)
-	}
-
-	if !strings.Contains(string(r), "404 page not found") {
-		t.Error("Was not expecting any value returned:", r)
-	}
+	assert.NoErrorf(t, err, "Sending http request error: %v", err)
+	assert.Containsf(t, string(r), "404 page not found", "Was not expecting any value returned: %s", string(r))
```
gbjk: 🚧 This error message does not make sense with assertify.
```suggestion
	assert.Containsf(t, string(r), "404 page not found", "SendHTTPRequest return should only contain 404")
```

gbjk (changes requested):
Thanks for the changes.

A few more. One of them is a real zombie; it's survived 3 reviews now.

exchanges/mock/server_test.go
samuael: I hope the hope satisfies your request. The error is raised from the http package saying `'connection refused'` and I have added that accordingly.

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -249,16 +249,17 @@ Please refer to our [coding guidelines](/docs/CODING_GUIDELINES.md).
...
-    // generate functionality below. 
-    item := &request.Item{  
-        Method:        http.MethodGet,
-        Path:          path,
-        Result:        result,
-        Verbose:       e.Verbose,
-        HTTPDebugging: e.HTTPDebugging,
-        HTTPRecording: e.HTTPRecording,
-    }
+	// This is used to generate the *http.Request, used in conjunction with the
```
thrasher-: Please use space indentation here since markdown linters complain about this. ATM there's a mixture between spaces and tabs

communications/slack/slack_test.go
```diff
@@ -281,7 +281,7 @@ func TestHandleReconnectResponse(t *testing.T) {
 	}
 
 	data, err := json.Marshal(testURL)
-	require.NoError(t, err, "json.Marshal must not error")
+	require.NoError(t, err, "WriteFile must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/binance/binance_live_test.go
```diff
@@ -40,12 +40,12 @@ func TestMain(m *testing.M) {
...
+	e.setupOrderbookManager(ctx)
```
thrasher-: Can remove this

exchanges/binance/binance_mock_test.go
```diff
@@ -28,10 +28,10 @@ func TestMain(m *testing.M) {
 	if err := testexch.MockHTTPInstance(e); err != nil {
 		log.Fatalf("Binance MockHTTPInstance error: %s", err)
 	}
-
-	if err := e.UpdateTradablePairs(context.Background(), true); err != nil {
+	ctx := context.Background()
+	e.setupOrderbookManager(ctx)
```
thrasher-: Same for here

exchanges/binance/binance_test.go
```diff
@@ -1091,7 +1091,7 @@ func TestGetExchangeInfo(t *testing.T) {
 	info, err := e.GetExchangeInfo(t.Context())
 	require.NoError(t, err, "GetExchangeInfo must not error")
 	if mockTests {
-		exp := time.Date(2024, 5, 10, 6, 8, 1, int(707*time.Millisecond), time.UTC)
+		exp := time.UnixMilli(1754603741167).UTC()
```
thrasher-: Prefer the original style as it's more human friendly so it's easier to tell what time it is without conversion

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -402,13 +403,14 @@ func (e *Exchange) SendAuthHTTPRequest(ctx context.Context, ePath exchange.URL,
...
         fullPath := common.EncodeURLValues(endpointPath+path, params) + "&signature=" + hex.EncodeToString(hmacSigned)
         return &request.Item{
-            Method:        method,
-            Path:          fullPath,
-            Headers:       headers,
-            Result:        &interim,
-            Verbose:       e.Verbose,
-            HTTPDebugging: e.HTTPDebugging,
-            HTTPRecording: e.HTTPRecording,
+            Method:        		    method,
```
thrasher-: Same for here

exchanges/mock/common_test.go
```diff
@@ -17,102 +20,51 @@ func TestMatchURLVals(t *testing.T) {
...
 			Number:     3.14,
 			SomeString: "hello, peoples",
 		},
 	}
 
 	payload, err := json.Marshal(test1)
-	if err != nil {
-		t.Error("marshal error", err)
-	}
+	require.NoError(t, err, "WriteFile must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/mock/common_test.go
```diff
@@ -125,17 +77,13 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
-
-	vals, err := DeriveURLValsFromJSONMap(payload)
-	if err != nil {
-		t.Error("DeriveURLValsFromJSON error", err)
-	}
-
-	if vals["val"][0] != "1" {
-		t.Error("DeriveURLValsFromJSON unexpected value",
-			vals["val"][0])
+	require.NoError(t, err, "WriteFile must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/mock/recording_test.go
```diff
@@ -46,22 +45,26 @@ func TestCheckResponsePayload(t *testing.T) {
 	}
 
 	payload, err := json.Marshal(testbody)
-	if err != nil {
-		t.Fatal("json marshal error", err)
-	}
+	require.NoError(t, err, "json marshal must not error")
```
thrasher-, line 48:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/mock/recording_test.go
```diff
@@ -46,22 +45,26 @@ func TestCheckResponsePayload(t *testing.T) {
...
+	assert.Equal(t, expected, string(data))
+}
 
-	if string(data) != expected {
-		t.Error("unexpected returned data")
-	}
+func TestGetExcludedItems(t *testing.T) {
+	exclusionList, err := getExcludedItems()
+	require.NoError(t, err, "getExcludedItems must not error")
+	require.NotEmpty(t, exclusionList.Headers, "Headers must not be empty")
```
thrasher-:
```suggestion
	assert.NotEmpty(t, exclusionList.Headers, "Headers should not be empty")
```

exchanges/mock/recording_test.go
```diff
@@ -46,22 +45,26 @@ func TestCheckResponsePayload(t *testing.T) {
...
+	require.NotEmpty(t, exclusionList.Variables, "Variables must not be empty")
```
thrasher-:
```suggestion
	assert.NotEmpty(t, exclusionList.Variables, "Variables should not be empty")
```

exchanges/mock/recording_test.go
```diff
@@ -81,130 +84,123 @@ type TestStructLevel1 struct {
...
-		t.Error("Check JSON error", err)
-	}
+	vals, err := CheckJSON(input, &exclusionList, 4)
+	assert.NoError(t, err, "CheckJSON should not error")
 
 	payload, err := json.Marshal(vals)
-	if err != nil {
-		t.Fatal("json marshal error", err)
-	}
+	require.NoError(t, err, "Marshal json must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/mock/recording_test.go
```diff
@@ -81,130 +84,123 @@ type TestStructLevel1 struct {
...
-	}
-
-	if newStruct.StructVal.OtherData.BadVal != 0 {
-		t.Error("Value not wiped correctly")
-	}
-
-	if newStruct.StructVal.OtherData.BadVal2 != 0 {
-		t.Error("Value not wiped correctly")
-	}
+	require.NoError(t, err, "Umarshal json must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Umarshal must not error")
```

exchanges/mock/recording_test.go
```diff
@@ -81,130 +84,123 @@ type TestStructLevel1 struct {
...
-	exclusionList, err := GetExcludedItems()
-	if err != nil {
-		t.Error("GetExcludedItems error", err)
-	}
-
-	if len(exclusionList.Headers) == 0 {
-		t.Error("Header exclusion list not popoulated")
-	}
+	content, err := json.Marshal(testVal)
+	require.NoError(t, err, "WriteFile must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/mock/server_test.go
```diff
@@ -38,89 +37,54 @@ func TestNewVCRServer(t *testing.T) {
 		Amount:   1,
 		Currency: "bitcoin",
 	})
-	if err != nil {
-		t.Fatal("marshal error", err)
-	}
+	require.NoError(t, err, "WriteFile must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/mock/server_test.go
```diff
@@ -38,89 +37,54 @@ func TestNewVCRServer(t *testing.T) {
...
 
 	testValue := HTTPResponse{Data: rp, QueryString: queryString, BodyParams: queryString}
 	test1.Routes["/test"][http.MethodGet] = []HTTPResponse{testValue}
 
 	payload, err := json.Marshal(test1)
-	if err != nil {
-		t.Fatal("marshal error", err)
-	}
+	require.NoError(t, err, "WriteFile must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/order/order_test.go
```diff
@@ -1741,7 +1741,7 @@ func TestMarshalOrder(t *testing.T) {
 		Price:      1000,
 	}
 	j, err := json.Marshal(orderSubmit)
-	require.NoError(t, err, "json.Marshal must not error")
+	require.NoError(t, err, "WriteFile must not error")
```
thrasher-:
```suggestion
	require.NoError(t, err, "Marshal must not error")
```

exchanges/exchange_types.go
```diff
@@ -230,19 +230,21 @@ type API struct {
...
+	API                        API
+	BaseCurrencies             currency.Currencies
+	CurrencyPairs              currency.PairsManager
+	Features                   Features
+	HTTPTimeout                time.Duration
+	HTTPRecording              bool
+	HTTPMockDataSliceLimit     int // Use with HTTPRecording to reduce the size of recorded mock data
+	HTTPDebugging              bool
+	BypassConfigFormatUpgrades bool
+
```
thrasher-: Can rm this line

thrasher- (changes requested):
Almost there, just some more refinement changes

exchanges/mock/server_test.go
```diff
@@ -59,32 +59,32 @@ func TestNewVCRServer(t *testing.T) {
 		"http://localhost:300/somethingElse?"+queryString,
 		nil,
 		bytes.NewBufferString(""), true)
-	assert.Error(t, err, "Sending http request expected an error")
+	assert.Error(t, err, "SendHTTPRequest should return a connection refused error")
```
gbjk: 🚧 Sorry, no, this doesn't address my request to use `ErrorIs` or `ErrorContains`
```suggestion
	assert.ErrorContains(t, err, "connection refused", "SendHTTPRequest should return the correct error")
```

exchanges/mock/server_test.go
gbjk: I'm afraid not. I'll re-raise it

exchanges/mock/server_test.go
```diff
@@ -59,32 +59,32 @@ func TestNewVCRServer(t *testing.T) {
...
 	// Expected good outcome
 	r, err := common.SendHTTPRequest(t.Context(),
 		http.MethodGet,
 		deets,
 		nil,
 		bytes.NewBufferString(""), true)
-	assert.NoErrorf(t, err, "Sending http request error: %v", err)
-	assert.Containsf(t, string(r), "404 page not found", "Was not expecting any value returned: %s", string(r))
+	assert.NoError(t, err, "SendHTTPRequest should not error")
+	assert.Containsf(t, string(r), "404 page not found", "SendHTTPRequest return should only contain 404")
```
gbjk: 🚧  Sorry, this is my fault - my suggestion didn't remove the `f`s.
```suggestion
	assert.Contains(t, string(r), "404 page not found", "SendHTTPRequest return should only contain 404")
```

gbjk (changes requested):
Thanks for those changes.
Just one zombie 🧟 and one issue with `Containsf` without arguments.

exchanges/mock/recording.go
```diff
@@ -343,7 +328,7 @@ func CheckJSON(data any, excluded *Exclusion, limit int) (any, error) {
 
 	if len(context) == 0 {
 		// Nil for some reason, should error out before in json.Unmarshal
-		return nil, nil
+		return context, nil
```
shazbert: context var is a package name conflict, rn to payload or similar

gloriousCode: Yes please rename this plus contextValue

exchanges/bitstamp/bitstamp.go
```diff
@@ -488,12 +488,13 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path st
 		return err
```
gloriousCode, line 488: OOS: Update bitstamp_live_test.go to use `e.Name` not `b.Name`

exchanges/mock/recording.go
gloriousCode: This has not been addressed

gloriousCode (changes requested):
Thank you for your changes. Still one main pending thing

exchanges/mock/recording.go
gloriousCode: Hello, I wish to inform you I was WRONG and it has been addressed. Thank you

exchanges/mock/server_test.go
```diff
@@ -59,7 +76,17 @@ func TestNewVCRServer(t *testing.T) {
 		"http://localhost:300/somethingElse?"+queryString,
 		nil,
 		bytes.NewBufferString(""), true)
-	assert.ErrorContains(t, err, "connection refused", "SendHTTPRequest should return the correct error")
+	assert.Error(t, err, "SendHTTPRequest should return the correct error")
```
thrasher-:
```
	var netErr *net.OpError
	assert.ErrorAs(t, err, &netErr, "SendHTTPRequest should return a net.OpError for an invalid host")
```
	
I reckon something like this is sufficient due to different messages between different OS's

thrasher- (changes requested):

thrasher- (approved):
Changes ACK!

exchanges/mock/server_test.go
```diff
@@ -38,89 +38,57 @@ func TestNewVCRServer(t *testing.T) {
...
 
 	_, err = common.SendHTTPRequest(t.Context(),
 		http.MethodGet,
 		"http://localhost:300/somethingElse?"+queryString,
 		nil,
 		bytes.NewBufferString(""), true)
-	if err == nil {
-		t.Error("Sending http request expected an error")
-	}
+	assert.Error(t, err, "SendHTTPRequest should return the correct error")
```
gbjk: 🚧 I think this line was meant to be removed. The following line does the same thing, but better.
Also, the test message is inaccurate for the test.
```suggestion
```

samuael: I don't know which version you are referring to.
The last commit had this line removed. Could you check it again?

gbjk: It does not look like that to me:
https://github.com/thrasher-corp/gocryptotrader/pull/1968/commits/def3f9aa6

And also here:
Screenshot 2025-08-25 at 19 57 53

gbjk (approved):

Issue #1947, opened 2025-06-12
Bithumb: Add zero filter for orderbook amount


thrasher-:
Based off https://api.bithumb.com/public/orderbook/btc, exchange things:

```
      {
        "price": "148360000",
        "quantity": "0.0002"
      },
      {
        "price": "148371000",
        "quantity": "0.0000"
      },
      {
        "price": "148375000",
        "quantity": "0.0842"
      },
```

Bithumb will need a filter just like what we do with Bitstamp

PR #1969, opened 2025-07-07
Bithumb: Filter zero quantity values in GetOrderBook

Polizo96:
bugfix: added a filter to eliminate zero quantity of bids and asks at bithumb::GetOrderBook

feature: added assertions in TestGetOrderBook

# PR Description

Added a filter to elimite zero values of quantity in both bids and asks in the response of GetOrderBook
Additionally, a check in test file to verify that the changes are working correctly. (also made the assertion.NotEmpty conditionally because wasnt sure if this should be empty or not)

Fixes # (issue)
#1947 

## How has this been tested

Tested with debugger and with the additions of the bithumb_test

### DIFF
--- a/exchanges/bithumb/bithumb.go
+++ b/exchanges/bithumb/bithumb.go
@@ -132,6 +132,30 @@ func (b *Bithumb) GetOrderBook(ctx context.Context, symbol string) (*Orderbook, 
 		return nil, errors.New(response.Message)
 	}
 
+	filteredBids := make([]struct {
+		Quantity float64 `json:"quantity,string"`
+		Price    float64 `json:"price,string"`
+	}, 0, len(response.Data.Bids))
+
+	for _, bid := range response.Data.Bids {
+		if bid.Quantity > 0 {
+			filteredBids = append(filteredBids, bid)
+		}
+	}
+	response.Data.Bids = filteredBids
+
+	filteredAsks := make([]struct {
+		Quantity float64 `json:"quantity,string"`
+		Price    float64 `json:"price,string"`
+	}, 0, len(response.Data.Asks))
+
+	for _, ask := range response.Data.Asks {
+		if ask.Quantity > 0 {
+			filteredAsks = append(filteredAsks, ask)
+		}
+	}
+	response.Data.Asks = filteredAsks
+
 	return &response, nil
 }
 
--- a/exchanges/bithumb/bithumb.go
+++ b/exchanges/bithumb/bithumb.go
@@ -132,31 +132,19 @@ func (b *Bithumb) GetOrderBook(ctx context.Context, symbol string) (*Orderbook, 
 		return nil, errors.New(response.Message)
 	}
 
-	filteredBids := make([]struct {
-		Quantity float64 `json:"quantity,string"`
-		Price    float64 `json:"price,string"`
-	}, 0, len(response.Data.Bids))
-
-	for _, bid := range response.Data.Bids {
-		if bid.Quantity > 0 {
-			filteredBids = append(filteredBids, bid)
-		}
-	}
-	response.Data.Bids = filteredBids
-
-	filteredAsks := make([]struct {
-		Quantity float64 `json:"quantity,string"`
-		Price    float64 `json:"price,string"`
-	}, 0, len(response.Data.Asks))
-
-	for _, ask := range response.Data.Asks {
-		if ask.Quantity > 0 {
-			filteredAsks = append(filteredAsks, ask)
-		}
-	}
-	response.Data.Asks = filteredAsks
+	response.Data.Bids = response.Data.Bids.FilterZeros()
+	response.Data.Asks = response.Data.Asks.FilterZeros()
 
 	return &response, nil
+}
+
+func (o OrderbookLevels) FilterZeros() (c OrderbookLevels) {
+	for _, l := range o {
+		if l.Quantity > 0 {
+			c = append(c, l)
+		}
+	}
+	return c
 }
 
 // GetAssetStatus returns the withdrawal and deposit status for the symbol
--- a/exchanges/bithumb/bithumb.go
+++ b/exchanges/bithumb/bithumb.go
@@ -132,19 +132,21 @@ func (b *Bithumb) GetOrderBook(ctx context.Context, symbol string) (*Orderbook, 
 		return nil, errors.New(response.Message)
 	}
 
-	response.Data.Bids = response.Data.Bids.FilterZeros()
-	response.Data.Asks = response.Data.Asks.FilterZeros()
+	response.Data.Bids.FilterZeroQuantities()
+	response.Data.Asks.FilterZeroQuantities()
 
 	return &response, nil
 }
 
-func (o OrderbookLevels) FilterZeros() (c OrderbookLevels) {
-	for _, l := range o {
-		if l.Quantity > 0 {
-			c = append(c, l)
+// FilterZeroQuantities filters out orderbook levels with zero quantity
+func (o *OrderbookLevels) FilterZeroQuantities() {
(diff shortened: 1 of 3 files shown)


gbjk (changes requested):
Thanks for the contribution.

I think we would be better using named types, and a method.

Something like [this](http://github.com/gbjk/gocryptotrader/commit/526babe86).

exchanges/bithumb/bithumb.go
```diff
@@ -132,6 +132,30 @@ func (b *Bithumb) GetOrderBook(ctx context.Context, symbol string) (*Orderbook,
 		return nil, errors.New(response.Message)
 	}
 
+	filteredBids := make([]struct {
+		Quantity float64 `json:"quantity,string"`
+		Price    float64 `json:"price,string"`
+	}, 0, len(response.Data.Bids))
+
```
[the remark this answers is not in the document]
cranktakular: This is enforced by the compiler upon append. So don't worry about this.

It would be nice to de-anonymise and share the struct, but that seems a touch out of scope.

cranktakular (approved):
tACK

exchanges/bithumb/bithumb.go
```diff
@@ -132,9 +132,21 @@ func (b *Bithumb) GetOrderBook(ctx context.Context, symbol string) (*Orderbook,
 		return nil, errors.New(response.Message)
 	}
 
+	response.Data.Bids = response.Data.Bids.FilterZeros()
+	response.Data.Asks = response.Data.Asks.FilterZeros()
+
 	return &response, nil
 }
 
+func (o OrderbookLevels) FilterZeros() (c OrderbookLevels) {
```
thrasher-: Exported methods/types needs a comment for this and the below

exchanges/bithumb/bithumb.go
```diff
@@ -132,9 +132,21 @@ func (b *Bithumb) GetOrderBook(ctx context.Context, symbol string) (*Orderbook,
...
+	for _, l := range o {
```
thrasher-: Can be improved by preallocating as this issue is intermittment and doesn't always happen:

```
func (o OrderbookLevels) FilterZeros() (c OrderbookLevels) {
	c = make(OrderbookLevels, 0, len(o))
	for i := range o {
		if o[i].Quantity == 0 {
                        continue
                }
		c = append(c, o[i])
	}
	return c
}
```

When adding new methods, we prefer to have direst test coverage for it rather than in a different test

gbjk: Note: I think if we're going to pre-alloc it, which I agree with, I'd remove the named return var and make it:
```
	c: = make(OrderbookLevels, 0, len(o))
```

But also: If we're going to pre-alloc it but then filter it, we should shrink:
```
	return slices.Clip(c)
```

exchanges/bithumb/bithumb_types.go
```diff
@@ -330,3 +324,10 @@ type StatusAll struct {
 	} `json:"data"`
 	Message string `json:"message"`
 }
+
+type OrderbookLevel struct {
```
gbjk, line 329: My mistake; Sorry.
None of `OrderbookLevel` or `OrderbookLevels`, or the functions on them, need to be public.
```suggestion
type orderbookLevel struct {
```

exchanges/bithumb/bithumb_types.go
```diff
@@ -330,3 +324,10 @@ type StatusAll struct {
...
+	Quantity float64 `json:"quantity,string"`
+	Price    float64 `json:"price,string"`
+}
+
+type OrderbookLevels []OrderbookLevel
```
gbjk, line 335:
```suggestion
type orderbookLevels []OrderbookLevel
```

gbjk (changes requested):
Sorry - my mistake, the new types don't need to be public.
I was thinking they did for json, but nope, it's all good with them private.

exchanges/bithumb/bithumb_types.go
gloriousCode: I think it would be nicer to export the structs. Its more strange to have exported func `GetOrderBook` return an exported type `Orderbook` with exported property `Bids`, be an unexported type with exported fields `Quantity` and `Levels`

exchanges/bithumb/bithumb.go
shazbert: Suggestion: We don't need the extra allocation we can just filter in place on the original and drop the return. We probably don't need the added overhead of slices.Clip as this will be garbage collected as soon as we unify to `orderbook.Level/s` and store it in the orderbook package.

exchanges/bithumb/bithumb_types.go
shazbert: Happy to keep all these exported as suggested by @gloriousCode

shazbert (changes requested):
Thanks for the PR, my 2 cents, all optional.

exchanges/bithumb/bithumb_types.go
gbjk: Sorry for the delay replying.

I'm happy with that too; Sound reasoning.

exchanges/bithumb/bithumb.go
```diff
@@ -132,9 +132,23 @@ func (b *Bithumb) GetOrderBook(ctx context.Context, symbol string) (*Orderbook,
...
 	}
 
+	response.Data.Bids.FilterZeroQuantities()
+	response.Data.Asks.FilterZeroQuantities()
+
 	return &response, nil
 }
 
+// FilterZeroQuantities filters out orderbook levels with zero quantity
+func (o *OrderbookLevels) FilterZeroQuantities() {
```
gloriousCode, line 142: [Almost OOS] I'm happy with the fix and all, but if zero quantity is a problem for REST, is it also an issue for websocket?
```go
func (b *Bithumb) wsHandleData(respRaw []byte) error {
...
case "orderbookdepth":
		var orderbooks WsOrderbooks
		err = json.Unmarshal(resp.Content, &orderbooks)
		if err != nil {
			return err
		}
		for _, x := range orderbooks.List {
			if x.Quantity == 0 {
				panic("woah man")
			}
		}
```

```
[DEBUG] | WEBSOCKET | 08/08/2025 16:05:53 | Bithumb wss://pubwss.bithumb.com/pub/ws: Message received: {"type":"orderbookdepth","content":{"datetime":"1754633153415044","list":[{"total":"0","orderType":"ask","quantity":"0","price":"164165000","symbol":"BTC_KRW"},{"total":"1","orderType":"ask","quantity":"0.0257","price":"160512000","symbol":"BTC_KRW"}]}}
panic: woah man

goroutine 104 [running]:
github.com/thrasher-corp/gocryptotrader/exchanges/bithumb.(*Bithumb).wsHandleData(0x140002b8908, {0x14000953000, 0xfd, 0x200})
	/Users/butts/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/bithumb/bithumb_websocket.go:163 +0x61c
github.com/thrasher-corp/gocryptotrader/exchanges/bithumb.(*Bithumb).wsReadData(0x140002b8908)
	/Users/butts/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/bithumb/bithumb_websocket.go:73 +0x90
created by github.com/thrasher-corp/gocryptotrader/exchanges/bithumb.(*Bithumb).WsConnect in goroutine 115
	/Users/butts/go/src/github.com/thrasher-corp/gocryptotrader/exchanges/bithumb/bithumb_websocket.go:54 +0x180
```

thrasher-: Since level's `updateInsertByPrice` will delete prices with an amount <= 0 it's a happy bing and you won't see zeroes when streaming the book via ws

PR #1974, opened 2025-07-15
workflows: Add check for common LLM targeted unicode characters

thrasher-:
# PR Description

Adds bad bing (zero width encoded messages directed at LLMs) protection. Despite testing this on Copilot and Jules, they didn't execute on it but it's good to detect these things anyway.

Thanks to @cranktakular for sharing this article: https://www.promptfoo.dev/blog/invisible-unicode-threats/

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Encoding a secret message in a .md file and getting LLMs to read from it

## Checklist

### DIFF
--- a/.github/workflows/misc.yml
+++ b/.github/workflows/misc.yml
@@ -51,3 +51,9 @@ jobs:
         echo "::error::Replace !errors.Is(err, target) with testify equivalents"
         exit 1
 
+    - name: Check for zero width unicode characters
+      run: |
+        grep -r -I --color=always -P "[\x{200B}\x{200C}\x{200D}\x{FEFF}]" . || exit 0
+        echo "::error::Remove zero width unicode characters"
+        exit 1
+
--- a/.github/workflows/misc.yml
+++ b/.github/workflows/misc.yml
@@ -53,7 +53,7 @@ jobs:
 
     - name: Check for zero width unicode characters
       run: |
-        grep -r -I --color=always -P "[\x{200B}\x{200C}\x{200D}\x{FEFF}]" . || exit 0
+        grep -r -n -I --color=always --exclude-dir=.git -P "[\x{200B}\x{200C}\x{200D}\x{FEFF}]" . || exit 0
         echo "::error::Remove zero width unicode characters"
         exit 1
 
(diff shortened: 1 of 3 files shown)


.github/workflows/misc.yml
```diff
@@ -51,3 +51,9 @@ jobs:
         echo "::error::Replace !errors.Is(err, target) with testify equivalents"
         exit 1
 
+    - name: Check for zero width unicode characters
+      run: |
+        grep -r -n -I --color=always --exclude-dir=.git -P "[\x{200B}\x{200C}\x{200D}\x{FEFF}]" . || exit 0
```
cranktakular: Hex 2060 should also be added https://unicode-explorer.com/articles/space-characters

More speculative, but I wonder if a whitelist might be ideal, in case other risky characters are included. Although I'm not sure how feasible such an approach would be.

shazbert (approved):
ch ACK.

.github/workflows/misc.yml
```diff
@@ -51,3 +51,9 @@ jobs:
         echo "::error::Replace !errors.Is(err, target) with testify equivalents"
         exit 1
 
+    - name: Check for zero width unicode characters
+      run: |
+        grep -r -n -I --color=always --exclude-dir=.git -P "[\x{200B}\x{200C}\x{200D}\x{FEFF}\x{2060}]" . || exit 0
```
samuael: Searched for other invisible (zero width characters) and found some characters we can add.
```suggestion
        grep -r -n -I --color=always --exclude-dir=.git -P "[\x{00A0}\x{2000}-\x{200F}\x{2028}-\x{202F}\x{205F}-\x{206F}\x{FEFF}\x{3000}]" .
```

samuael (commented):
Just minor suggestion

.github/workflows/misc.yml
thrasher-: Added more robust checks for these which now includes format, separator and mark checks which covers all the invisible things that can be thrown at LLMs

shazbert (approved):
LGTM and comes up with changes as well from copy pasted cooked html. 👍 🔥 ACK

PR #1923, opened 2025-05-26
accounts: Move to instance methods, fix races and isolate tests

gbjk:
This PR moves the account service from being a global instance with package functions to being an instance pattern with an optional global instance. Most methods for accounts move to the exchange interface, since all account updates would require a working exchange.

Fixes #1902 ; The original motivation for this PR was to fix races on Bybit.

Also addresses issues with when running tests with -count=N+ where we need to be more careful about the implications of `init` functions.
Reduces boilerplate and complexity in all exchanges when interacting with exchanges.

* `exchanges/account` ➡️ `exchange/accounts`
* exchange interface
  * `UpdateAccountInfo`  ➡️ `UpdateAccountBalances`  
  * `GetCachedAccountInfo` ➡️ `GetCachedSubAccounts`
  * ⭐ `SubscribeAccountBalances`
  * ⭐ `GetCachedCurrencyBalances`
* Add `Accounts` to exchange base
* Replace global service proxy methods with instance methods
  * `account.Process` ➡️  `e.Accounts.Save`
  * `account.ProcessChange` ➡️ `e.Accounts.Update`
* Accounts api changes:
  * `Service` ➡️  `store`
  * Protect `account.global` with atomic.Pointer
  * `NewStore` and `GetStore`
  * `NewAccounts` and `MustNewAccounts`
  * `account.SubscribeToExchangeAccount` ➡️ `e.SubscribeAccountBalanc`
  * `a.Subscribe` also available on the Account instance
  * `account.GetExchangeAccounts` ➡️ `account.GetStore().GetExchangeAccounts`
  * ⭐ `a.CurrencyBalances`
  * ⭐ `a.SubAccounts`
  * `account.GetBalance` ➡️ `account.GetStore().GetBalance`
  * `service.Save` ➡️ `Accounts.Save`
  * `service.Update` ➡️ `Accounts.Update`
  * `SubAccounts` and `s.Merge`
  * `CurrencyBalances` with `b.Add` and `b.Set`
  * All updates to accounts can be monitored through Subscribe, replacing the complexity of protected balances
* Deprecated apiserver:
  * Remove `AllEnabledAccountInfo`
  * Update docs to harden warnings now time has passed
* rpcserver:
  * `GetAccountInfo` ➡️ `GetAccountBalances` 
  * `UpdateAccountInfo` ➡️ `UpdateAccountBalances`
  * `GetAccountInfoStream ` ➡️ `GetAccountBalancesStream`
* Dispatcher changes:
[...]

PR #1980, opened 2025-07-24
bybit: fix UpdateAccountInfo free and hold calculations

shazbert:
# PR Description

For free field available to withdraw is a separate to tradable amounts. New calculation derives diff from borrow amount as wallet balance includes negative amounts that are borrowed. Then reduces the result by the locked amount which can be a limit order.

Hold is just the locked amount. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/bybit/bybit_test.go
+++ b/exchanges/bybit/bybit_test.go
@@ -25,7 +25,6 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/exchanges/kline"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/margin"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
-	"github.com/thrasher-corp/gocryptotrader/exchanges/request"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/sharedtestvalues"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/subscription"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
@@ -2874,7 +2873,7 @@ func TestUpdateAccountInfo(t *testing.T) {
 		sharedtestvalues.SkipTestIfCredentialsUnset(t, e)
 	}
 
-	r, err := e.UpdateAccountInfo(request.WithVerbose(t.Context()), asset.Spot)
+	r, err := e.UpdateAccountInfo(t.Context(), asset.Spot)
 	require.NoError(t, err, "UpdateAccountInfo must not error")
 	require.NotEmpty(t, r, "UpdateAccountInfo must return account info")
 
--- a/exchanges/bybit/bybit_test.go
+++ b/exchanges/bybit/bybit_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/exchanges/kline"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/margin"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/request"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/sharedtestvalues"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/subscription"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
@@ -2873,7 +2874,7 @@ func TestUpdateAccountInfo(t *testing.T) {
 		sharedtestvalues.SkipTestIfCredentialsUnset(t, e)
 	}
 
-	r, err := e.UpdateAccountInfo(t.Context(), asset.Spot)
+	r, err := e.UpdateAccountInfo(request.WithVerbose(t.Context()), asset.Spot)
 	require.NoError(t, err, "UpdateAccountInfo must not error")
 	require.NotEmpty(t, r, "UpdateAccountInfo must return account info")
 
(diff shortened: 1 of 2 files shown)


exchanges/bybit/bybit_test.go
```diff
@@ -2886,15 +2886,15 @@ func TestUpdateAccountInfo(t *testing.T) {
 			case 0:
 				assert.Equal(t, currency.USDC, r.Accounts[0].Currencies[x].Currency, "Currency should be USDC")
 				assert.Equal(t, -30723.63021638, r.Accounts[0].Currencies[x].Total, "Total amount should match")
-				assert.Equal(t, -30723.63021638, r.Accounts[0].Currencies[x].Hold, "Hold amount should match")
+				assert.Equal(t, 0.0, r.Accounts[0].Currencies[x].Hold, "Hold amount should match")
 				assert.Equal(t, 30723.630216383714, r.Accounts[0].Currencies[x].Borrowed, "Borrowed amount should match")
-				assert.Equal(t, 0.0, r.Accounts[0].Currencies[x].Free, "Free amount should match")
+				assert.Equal(t, 3.714376362040639e-09, r.Accounts[0].Currencies[x].Free, "Free amount should match")
```
gbjk: These are mock tests, right?
Can we just simplify the amount we're expecting here or do we really need to exercise this level of precision?

exchanges/bybit/bybit_wrapper.go
```diff
@@ -572,9 +570,9 @@ func (e *Exchange) UpdateAccountInfo(ctx context.Context, assetType asset.Item)
 			balance := account.Balance{
 				Currency: balances.List[i].Coin[c].Coin,
 				Total:    balances.List[i].Coin[c].WalletBalance.Float64(),
-				Free:     balances.List[i].Coin[c].AvailableToWithdraw.Float64(),
+				Free:     (balances.List[i].Coin[c].BorrowAmount.Float64() + balances.List[i].Coin[c].WalletBalance.Float64()) - balances.List[i].Coin[c].Locked.Float64(),
```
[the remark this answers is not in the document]
gbjk: I'm fine with not addressing this, because of the overlap with #1923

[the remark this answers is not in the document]
gloriousCode: I disagree on the basis that 1923 has not yet been reviewed by everyone and will probably take awhile. Plus merging more readable code now is better than waiting. The rebase impact to you will be the same as it is now - very painful 😭 
So I think improving readability here is best

exchanges/bybit/bybit_test.go
gloriousCode: Even better would be changing the result to one that populates all the fields that have been updated for calculations. Better for testing

gloriousCode (changes requested):
Looks good, I agree with the AI 🤖

exchanges/bybit/bybit_test.go
shazbert: Well it depends, it seems like this is a real world example. I don't have borrow amounts to inject an update but this does. The problem with adjusting to be exact would obfuscate the truncation that takes place on the total `"walletBalance":"-30723.63021638"` to 8 decimal places but the borrowed amount `"borrowAmount":"30723.630216383711792744"` is truncated to 18 decimal places. I could strip the result on the mock by truncating the borrowAmount or if I detect a negative value I flip free to zero, but that might be disingenuous. How I am calculating this could be completely incorrect as well. But it seems more correct than what it was. Thoughts?

shazbert: @gloriousCode @gbjk bump

gbjk: I'm fine if you don't address this.

But:
* `seems like this is a real world example` - If I'm not really sure about the source, I never trust mock data
* Truncation is going to happen or it isn't but it doesn't feel like it's part of our code or testing - Nothing we're doing is going to be exercised by having bigger numbers. I think you're just exercising Go, and I think they might already have some unit tests.

exchanges/bybit/bybit_wrapper.go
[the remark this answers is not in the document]
gbjk: For me, readable would be:
```suggestion
				Free:     b.BorrowAmount.Float64() + b.WalletBalance.Float64() - b.Locked.Float64(),
```

I just realised: What's with the parenthesis around the addition ... 🤷

exchanges/bybit/bybit_test.go
shazbert: addressed: https://github.com/thrasher-corp/gocryptotrader/pull/1980/commits/e42403d0437fd4a4d1662d1ba3df951f53f8e736

exchanges/bybit/bybit_wrapper.go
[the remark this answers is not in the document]
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1980/commits/74bc1b49ec1fcb75959c4abc351a2638d84c9cbf

exchanges/bybit/bybit_types.go
```diff
@@ -895,7 +895,7 @@ type WalletBalance struct {
 		TotalWalletBalance     types.Number `json:"totalWalletBalance"`
 		AccountLTV             types.Number `json:"accountLTV"` // Account LTV: account total borrowed size / (account total equity + account total borrowed size).
 		TotalMaintenanceMargin types.Number `json:"totalMaintenanceMargin"`
-		Coin                   []struct {
+		Coin                   []*struct {
```
cranktakular, line 898: Is changing this to a slice of pointers to structs really a good idea? The slice itself is just a pointer to the data, its length, and its capacity. I'm not sure whether it's better for this memory in particular to be contiguous or not, but I generally don't see this done.

cranktakular (approved):
Got one change I want you to consider, and besides that, I can tACK this.

thrasher-:
@codex please review

exchanges/bybit/bybit_types.go
gloriousCode: The main reason it is a pointer is for the following code:
```go
for _, b := range balances.List[i].Coin {
```
(_though I would have gone with c to represent coin instead of b for balance - totally outrageous_)

If its not a pointer, go duplicates the contents of the struct, which is quite large. Making it a pointer has its own performance impacts, but they are less than copying large structs.

I'd suggest normally that it should be made into its own WalletBallenceCoin struct instead of an inline one, but I don't see this being used other than exactly where it has been

shazbert:
The values are still cooked, I need to rejig them. PLEASE WAIT. 👻

shazbert:
This will need full scrutiny again

exchanges/bybit/bybit_types.go
gbjk: I agree with switching to a pointer.

I'd be tempted to move to a defined type as well, but I'll leave that to you

exchanges/bybit/bybit_wrapper.go
```diff
@@ -680,18 +678,26 @@ func (e *Exchange) UpdateAccountInfo(ctx context.Context, assetType asset.Item)
...
+			balanceDiff := truncBorrow + c.WalletBalance.Float64()
+
+			freeBalance := balanceDiff - c.Locked.Float64()
+			if assetType == asset.Spot && c.AvailableBalanceForSpot.Float64() != 0 {
+				freeBalance = c.AvailableBalanceForSpot.Float64()
 			}
-			currencyBalance = append(currencyBalance, balance)
+
+			currencyBalance = append(currencyBalance, account.Balance{
+				Currency: c.Coin,
```
gbjk: 🦍 `Coin.Coin` is pretty hateful.
I don't share your enthusiasm for allowing their (often crazy) naming to pervade our interfaces.
I'd like to rename it to `Currency` 🙂

thrasher- (approved):
tACK with a live account in which the balances reflect the API more concretely

Issue #1961, opened 2025-06-30
gateio: orderbook update manager test failure


shazbert:
Investigate why this test is being silly. 

Test seen:
https://github.com/thrasher-corp/gocryptotrader/actions/runs/15963488621/job/45019756470?pr=1948#step:11:342

Pic: 
<img width="695" alt="Image" src="https://github.com/user-attachments/assets/3e102345-6bfe-4a2f-b636-873eed94e4f1" />

PR #1989, opened 2025-07-31
GateIO: Update websocket orderbook manager

shazbert:
# PR Description

* Issue: Fetching orderbook data without storing it was not supported for WebSocket management.
    - Fix: Added a new fetchOrderbook subfunction in wrapper.go to retrieve orderbook data without storing it, enabling efficient use in WebSocket management routines.
* Bug: Rate limiter could cause excessive memory usage if requests were heavily delayed.
    - Fix: Introduced a deadline mechanism in addition to the existing delay, ensuring the rate limiter does not lead to unbounded memory consumption by timing out stalled routines.
* Bug: Gate.io REST orderbooks were often out of sync, especially for trading pairs with no activity, and REST responses did not reflect past update IDs.
    - Fix: Implemented a synchronization mechanism that waits for an appropriate update ID via an unbuffered channel before applying the full orderbook cache, ensuring consistency between REST and WebSocket data for inactive pairs.
* Bug: Orderbook sync errors could leave stale data in the store, potentially causing memory leaks or sync failures.
    - Fix: Updated error handling to clear the orderbook store on all error pathways, ensuring garbage collection and allowing the syncer routine to restart cleanly.

Fixes # #1961 

## Type of change

## How has this been tested

- [x] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/gateio/ws_ob_update_manager_test.go
+++ b/exchanges/gateio/ws_ob_update_manager_test.go
@@ -384,7 +384,11 @@ func TestOBManagerProcessOrderbookUpdateHTTPMocked(t *testing.T) {
 	})
 	require.NoError(t, err, "ProcessOrderbookUpdate must not error")
 
-	time.Sleep(time.Millisecond * 100) // allow some time for rest mock request routine to complete
+	// Wait for the background sync goroutine to complete and orderbook to be synced
+	require.Eventually(t, func() bool {
+		_, err := e.Websocket.Orderbook.LastUpdateID(currency.NewBTCUSDT(), asset.Spot)
+		return err == nil
+	}, time.Second*5, time.Millisecond*50, "orderbook must eventually be synced")
 
 	err = m.ProcessOrderbookUpdate(t.Context(), e, 27596272448, &orderbook.Update{
 		UpdateID:   27596272449,


exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,46 +44,45 @@ type pendingUpdate struct {
...
 
-	lastUpdateID, err := g.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
+	lastUpdateID, err := e.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
 	if err != nil && !errors.Is(err, orderbook.ErrDepthNotFound) {
 		return err
 	}
 
-	if lastUpdateID+1 >= firstUpdateID {
-		return applyOrderbookUpdate(g, update)
+	if lastUpdateID+1 == firstUpdateID {
```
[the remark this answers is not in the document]
shazbert: When applying once the REST book is loaded this is better and ensures there are no dropped messages.

exchanges/subscription/keys.go
```diff
@@ -122,3 +122,29 @@ func (k IgnoringAssetKey) Match(eachKey MatchableKey) bool {
...
+
+var _ MatchableKey = ChannelKey{} // Enforce ChannelKey must implement MatchableKey
+
+// MustChannelKey is a helper function to create a ChannelKey from a subscription channel
+func MustChannelKey(channel string) ChannelKey {
+	if channel == "" {
+		panic("channel must not be empty")
+	}
+	return ChannelKey{Subscription: &Subscription{Channel: channel}}
+}
```
gloriousCode, line 140: [verbose whinge]
I am not totally against the execution of the func. My stance on `panic` has softened over time. 
The way you are using this function in this PR makes me think this isn't necessary to add at all. All you are doing is making a shorthand 
`var spotOrderbookUpdateKey = subscription.ChannelKey{Subscription: &subscription.Subscription{Channel: subscription.OrderbookChannel}}`

You aren't really needing to check that the const `subscription.OrderbookChannel` becomes `""` in a PR update, because that's an insane check

Hardly seems worth introducing into the codebase

gloriousCode (changes requested):
Had a look over and seems like it's all good. I should test it though. Next time!

exchanges/subscription/keys.go
shazbert: > Hardly seems worth introducing into the codebase

Yeah I am still opting to keep it. It's mostly for future when we need to get the bare bones channel sub info for dynamic info across channels/exchanges which will happen when Gateio gets updated with another template pass. It's now in its area of concern and it's not like its a massive introduction of unused code. Maybe I am not really understanding why this is so contentious?

gloriousCode: Its not.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,46 +43,45 @@ type pendingUpdate struct {
...
+		return applyOrderbookUpdate(e, update)
 	}
 
 	// Orderbook is behind notifications, therefore Invalidate store
```
cranktakular: I don't think this terminology is correct now. Execution can now reach here if our Orderbook is "ahead" (due to receiving information out of order, and being behind on a previous pass; or by the exchange sending the same data twice). This may not change the actual handling, but might change the comment; "Orderbook and notifications are desynced" maybe?

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,46 +43,45 @@ type pendingUpdate struct {
...
 	go func() {
-		select {
-		case <-ctx.Done():
-			return
-		case <-time.After(m.snapshotSyncDelay):
-			if err := cache.SyncOrderbook(ctx, g, update.Pair, update.Asset); err != nil {
-				g.Websocket.DataHandler <- fmt.Errorf("failed to sync orderbook for %v %v: %w", update.Pair, update.Asset, err)
-			}
+		if err := cache.SyncOrderbook(ctx, e, update.Pair, update.Asset, m.delay, m.deadline); err != nil {
+			log.Errorf(log.ExchangeSys, "%s websocket orderbook manager: failed to sync orderbook for %v %v: %v", e.Name, update.Pair, update.Asset, err)
```
cranktakular, line 127: Why was the %w changed to a %v here?

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -97,85 +104,164 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
...
+	case asset.Options:
+		return optionOrderbookUpdateLimit, nil
+	default:
+		return 0, fmt.Errorf("%w: %q", asset.ErrNotSupported, a)
+	}
+}
+
+// waitForUpdate waits for the next update with the specified ID to be available in the cache that exceeds the next
+// update ID, this ensures that an update can be applied to the orderbook. This is needed for illiquid pairs and the
+// REST book ID's are out of sync.
```
cranktakular: Grammar nits: "waits for next update with the specified ID to be available in the cache that exceeds the next update ID" mashes two explanations together. Given the code, I think it should be "waits for an update to be available in the cache with an ID that exceeds the specified next update ID".

"This is needed for illiquid pairs and the REST book ID's are out of sync" > "This is needed when the REST book IDs are out of sync and for illiquid pairs"

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -97,85 +104,164 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
...
+
+// applyPendingUpdates applies all pending updates to the orderbook
+func (c *updateCache) applyPendingUpdates(e *Exchange) error {
+	var updateApplied bool
 	for _, data := range c.updates {
-		lastUpdateID, err := g.Websocket.Orderbook.LastUpdateID(data.update.Pair, a)
+		bookLastUpdateID, err := e.Websocket.Orderbook.LastUpdateID(data.update.Pair, data.update.Asset)
 		if err != nil {
-			return err
+			return fmt.Errorf("applying pending updates: %w", err)
```
cranktakular: I don't think it's particularly customary, within our codebase, to wrap errors in this way - and I'd want some consistency for this.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -97,85 +104,164 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
...
 		}
-		nextID := lastUpdateID + 1
-		if data.firstUpdateID > nextID {
-			return errOrderbookSnapshotOutdated
+
+		nextUpdateID := bookLastUpdateID + 1 // `baseId+1`
```
cranktakular: This comment doesn't seem to provide more information than the code itself.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -97,85 +104,164 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
...
+
+		// Dump all notifications which satisfy `u` < `baseId+1`
+		if data.update.UpdateID < nextUpdateID {
```
cranktakular, line 233: If each pendingUpdate only contains a pointer to a single update, why is there a distinction between data.firstUpdateID and data.update.UpdateID such that different ones would be used in different situations?

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -97,85 +104,164 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
...
+	}
 
-	c.mtx.Lock() // lock here to prevent ws handle data interference with REST request above
-	defer func() {
-		c.updates = nil
-		c.updating = false
-		c.mtx.Unlock()
-	}()
+	// prevents rate limiter from blocking across multiple enabled pairs and connections, which consumes resources
+	// when appending to the update cache.
```
cranktakular: After talk on Slack, I'd suggest rewording to "Setting deadline to error out instead of waiting if the rate limiter goes wild". If you even want a comment for that, in the first place.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -97,85 +104,164 @@ func (m *wsOBUpdateManager) LoadCache(p currency.Pair, a asset.Item) *updateCach
...
+		c.clearWithLock()
+		return err
+	}
+
+	if err := c.waitForUpdate(ctxWDeadline, book.LastUpdateID+1); err != nil {
+		c.clearWithLock()
 		return err
 	}
 
+	c.mtx.Lock() // lock here to prevent ws handle data interference with REST request above
```
cranktakular: Comments that are complete sentences should start with capital letters https://google.github.io/styleguide/go/decisions.html#comment-sentences

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -102,49 +105,114 @@ func TestSyncOrderbook(t *testing.T) {
...
+	ctx, cancel := context.WithDeadline(t.Context(), time.Now())
+	defer cancel()
+	err = cache.waitForUpdate(ctx, 1338)
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+
+	cache.ch = make(chan int64, 1) // Reset channel to avoid deadlock
+	var wg sync.WaitGroup
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
```
cranktakular: If you rebase to master, you can replace these three lines with wg.Go( func () {

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -153,36 +221,65 @@ func TestApplyPendingUpdates(t *testing.T) {
...
 
 func TestApplyOrderbookUpdate(t *testing.T) {
 	t.Parallel()
 
 	e := new(Exchange)
 	require.NoError(t, testexch.Setup(e), "Setup must not error")
-	require.NoError(t, e.UpdateTradablePairs(t.Context(), false))
 
-	pair := currency.NewBTCUSDT()
+	pair := currency.NewPair(currency.BABY, currency.BABYDOGE)
```
cranktakular: Does it really make sense to use a non-USDT pair for the USDTMarginedFutures asset like this? I think you might want to stick with the good old BTC-USDT.

exchanges/gateio/gateio_wrapper.go
```diff
@@ -655,6 +655,17 @@ func (e *Exchange) UpdateOrderbook(ctx context.Context, p currency.Pair, a asset
 
 // UpdateOrderbookWithLimit updates and returns the orderbook for a currency pair with a set orderbook size limit
 func (e *Exchange) UpdateOrderbookWithLimit(ctx context.Context, p currency.Pair, a asset.Item, limit uint64) (*orderbook.Book, error) {
+	book, err := e.fetchOrderbook(ctx, p, a, limit)
+	if err != nil {
+		return nil, err
+	}
+	if err := book.Process(); err != nil {
+		return nil, err
```
cranktakular, line 627: Neither of these error cases are hit during testing, missing some coverage; could you expand tests to catch them?

exchanges/subscription/keys.go
```diff
@@ -122,3 +122,29 @@ func (k IgnoringAssetKey) Match(eachKey MatchableKey) bool {
...
+// ChannelKey is a key type for finding a single subscription by its channel, this will match first found.
+// For use with exchange websocket method GetSubscription.
+type ChannelKey struct {
+	*Subscription
+}
+
+var _ MatchableKey = ChannelKey{} // Enforce ChannelKey must implement MatchableKey
+
+// MustChannelKey is a helper function to create a ChannelKey from a subscription channel
+func MustChannelKey(channel string) ChannelKey {
```
cranktakular, line 135: MustChannelKey reads as a strange function name for what it's doing. Could it simply be CreateChannelKey?

exchanges/subscription/keys.go
```diff
@@ -122,3 +122,29 @@ func (k IgnoringAssetKey) Match(eachKey MatchableKey) bool {
 		eachSub.Levels == k.Levels &&
 		eachSub.Interval == k.Interval
 }
+
+// ChannelKey is a key type for finding a single subscription by its channel, this will match first found.
```
cranktakular, line 126: In what sense does it match first found? As far as I can tell from these subscriptions, it doesn't check a slice/map matching on the first it encounters. Rather, it just matches based on the channel.

exchanges/subscription/keys_test.go
```diff
@@ -150,3 +150,27 @@ func TestGetSubscription(t *testing.T) {
...
+
+func TestChannelKeyMatch(t *testing.T) {
+	t.Parallel()
+	key := ChannelKey{&Subscription{Channel: TickerChannel}}
+	try := &DummyKey{&Subscription{Channel: OrderbookChannel}, t}
+
+	require.Panics(t, func() { key.Match(nil) }, "Match on a nil must panic")
+	require.False(t, key.Match(try), "Gate 1: Match must reject a bad Channel")
+	try.Channel = TickerChannel
+	require.True(t, key.Match(try), "Gate 1: Match must accept a good Channel")
```
cranktakular: Wording nit: These aren't "good"/"bad" channels, they're both valid. I think it's better to say they're "different"/"identical".

exchanges/gateio/ws_ob_update_manager.go
shazbert: This is now a log; can't wrap.

exchanges/gateio/ws_ob_update_manager.go
shazbert: `data.update.UpdateID` is the `lastUpdateID` I didn't change name because it impacts across exchanges and wanted to keep this PR contained as much as possible.

exchanges/gateio/ws_ob_update_manager_test.go
shazbert: It's for segregation, pair is irrelevant, unlike the asset type. This interacts with orderbook package storage and it will conflict.

exchanges/gateio/gateio_wrapper.go
shazbert: Added a line; too much fiddle for `book.Process`

exchanges/subscription/keys.go
shazbert: Must is a prefix for something that panics on error which is idiomatic in Go. https://google.github.io/styleguide/go/decisions.html#must-functions

exchanges/subscription/keys.go
shazbert: When we call `e.Websocket.GetSubscription` this will iterate over the stored subscriptions and will match the first available subscription with channel key.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,235 @@ type pendingUpdate struct {
...
-	lastUpdateID, err := g.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
-	if err != nil && !errors.Is(err, orderbook.ErrDepthNotFound) {
-		return err
-	}
-
-	if lastUpdateID+1 >= firstUpdateID {
-		return applyOrderbookUpdate(g, update)
+	var inUpdateErr error
+	// Error disregarded as all error pathways require a new snapshot
+	if lastUpdateID, _ := e.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset); lastUpdateID+1 == firstUpdateID {
```
cranktakular: Consider accounting for errors anyway, as future changes to LastUpdateID may introduce ways to error out here.

thrasher-:
@codex please review

exchanges/gateio/gateio_wrapper.go
gloriousCode: Error cases like these are rarely the point of these tests. Making `book.Process()` fail should be the goal of `TestProcess`. The rules that make `book.Process()` fail may change, that is not a fault of gateio's implementation

Though I do believe UpdateOrderbookWithLimit should have had a dedicated function test. Or at least now `fetchOrderbook` should. The logic for 
```
if a != asset.Spot && !available {
return nil, fmt.Errorf("%v instrument %v does not have orderbook data", a, p)
``` 
should be tested

exchanges/gateio/ws_ob_update_manager.go
gloriousCode: I agree. LastUpdateID is part of a different package that could yet be utilised across many exchanges and may change

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,235 @@ type pendingUpdate struct {
...
+			if recentPendingUpdateID >= nextUpdateID {
+				return nil // Update is now available
+			}
+		}
+	}
 }
 
-// ApplyPendingUpdates applies all pending updates to the orderbook
-func (c *updateCache) applyPendingUpdates(g *Exchange, a asset.Item) error {
+// applyPendingUpdates applies all pending updates to the orderbook
```
gloriousCode, line 220:
```suggestion
// applyPendingUpdates applies all pending updates to the orderbook
// requires lock
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,235 @@ type pendingUpdate struct {
...
+	if !a.IsValid() {
+		return nil, fmt.Errorf("%w: %q", asset.ErrInvalidAsset, a)
+	}
 	m.mtx.RLock()
 	cache, ok := m.lookup[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
 	m.mtx.RUnlock()
 	if !ok {
 		m.mtx.Lock()
-		cache = &updateCache{}
+		cache = &updateCache{ch: make(chan int64)}
```
gloriousCode: could put cache outside the lock ~because it is a huge deal~

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,235 @@ type pendingUpdate struct {
...
+		return err
+	}
+
+	c.mtx.Lock() // Lock here to prevent ws handle data interference with REST request above.
+	defer func() {
+		c.clearNoLock()
+		c.mtx.Unlock()
+	}()
+
 	if a != asset.Spot {
```
gloriousCode: What about if I just have margin enabled? I might be wrong, but it looks like this doesn't work out when margin/cross margin is enabled and spot isn't

exchanges/gateio/ws_ob_update_manager.go
shazbert: It's for optimisation. It's a hot path and I didn't want to errors.Is call just for `orderbook.ErrOrderbookInvalid`. `currency.ErrCurrencyPairEmpty`, `asset.ErrInvalidAsset`, are checked in the cache. On all errors a lastUpdateID is zero.

> LastUpdateID is part of a different package that could yet be utilised across many exchanges and may change

It could either add or drop errors, in any case the lastUpdateID would be valid or invalid. I would like to pass on this.

exchanges/gateio/ws_ob_update_manager.go
shazbert: You are correct somewhat the asset is hard coded to spot in the calling function. The margin/cross margin is turned off for websocket and has a TODO: in wrapper. Would need to upgrade templating for it. They default to REST. This was just a carry over from previous versions.

exchanges/gateio/gateio_wrapper.go
shazbert: [test added](https://github.com/thrasher-corp/gocryptotrader/pull/1989/commits/a98e57ef3500d3676cf0d09600ab2af6f99617a1)

exchanges/gateio/ws_ob_update_manager.go
shazbert: [done](https://github.com/thrasher-corp/gocryptotrader/pull/1989/commits/a98e57ef3500d3676cf0d09600ab2af6f99617a1)

exchanges/gateio/ws_ob_update_manager.go
gloriousCode: That's fair, thanks for clarifying.  I was looking the GitHub diffs. I'll verify against master and then resolve 🌈

exchanges/gateio/ws_ob_update_manager.go
gloriousCode: I hear you. I think the way this section is done with the initial expected failure, ignored error & the other error.Is etc could use a relook. I'll have a look at this

exchanges/gateio/gateio_wrapper.go
```diff
@@ -641,20 +641,32 @@ func (e *Exchange) UpdateOrderbook(ctx context.Context, p currency.Pair, a asset
...
-		}
-		if a != asset.Spot && !available {
-			return nil, fmt.Errorf("%v instrument %v does not have orderbook data", a, p)
+		if a != asset.Spot {
+			available, err := e.checkInstrumentAvailabilityInSpot(p)
+			if err != nil {
+				return nil, err
+			}
+			if !available {
+				return nil, fmt.Errorf("%w: %q %q", errNoOrderbookDataAvailable, p, a)
```
gbjk: 🚧 This error message doesn't feel right to me.
It reads like the orderbook data isn't available. But it's actually the entire symbol isn't available in spot, right?
So I think the error needs to at least compound two errors maybe:
`error fetching orderbook data: no spot instrument available for margin PINGU-WOMBAT`
```suggestion
				return nil, fmt.Errorf("%w: %w for %q %q", errFetchingOrderbook, errNoSpotInstrument, a, p)
```

exchanges/gateio/gateio_wrapper.go
```diff
@@ -641,20 +641,32 @@ func (e *Exchange) UpdateOrderbook(ctx context.Context, p currency.Pair, a asset
...
 	switch a {
 	case asset.Spot, asset.Margin, asset.CrossMargin:
-		var available bool
-		available, err = e.checkInstrumentAvailabilityInSpot(p)
-		if err != nil {
-			return nil, err
-		}
-		if a != asset.Spot && !available {
-			return nil, fmt.Errorf("%v instrument %v does not have orderbook data", a, p)
+		if a != asset.Spot {
```
gbjk: 🥃 🦍  I'd move this to:
```suggestion
		case asset.Spot:
		    doStuff...
		    fallthrough
		case asset.Margin, asset.CrossMargin:
```

exchanges/gateio/gateio_wrapper_test.go
```diff
@@ -0,0 +1,73 @@
...
+	testexch "github.com/thrasher-corp/gocryptotrader/internal/testing/exchange"
+)
+
+func TestFetchOrderbook(t *testing.T) {
+	t.Parallel()
+
+	testexch.UpdatePairsOnce(t, e)
+
+	availSpot, err := e.GetAvailablePairs(asset.Spot)
+	require.NoError(t, err)
```
gbjk: 🚧 Please provide test messages to tests that aren't the unit test
🌴 This issue lives in many places below
```suggestion
	require.NoError(t, err, "GetAvailablePairs must not error")
```

exchanges/gateio/gateio_wrapper_test.go
```diff
@@ -0,0 +1,73 @@
...
+				require.ErrorIs(t, err, tc.err)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, e.Name, got.Exchange)
+			assert.True(t, tc.pair.Equal(got.Pair))
+			assert.Equal(t, tc.a, got.Asset)
+			// options books are consistently empty but should not exceed limit 1
+			assert.LessOrEqual(t, len(got.Asks), 1)
+			assert.LessOrEqual(t, len(got.Bids), 1)
```
gbjk: 🚧 nearly always comments around tests are actually hidden test descriptions
It makes me wonder if we should have Equal Len 1 for Spot, and then LessOrEqual for options
```suggestion
			// options books are consistently empty but should not exceed limit 1
			assert.LessOrEqual(t, len(got.Asks), 1, "Asks count should not exceed limit, but may be empty especially for options")
			assert.LessOrEqual(t, len(got.Bids), 1, "Bids count should not exceed limit, but may be empty especially for options")
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+	}
+
 	cache.mtx.Lock()
 	defer cache.mtx.Unlock()
 
 	if cache.updating {
 		cache.updates = append(cache.updates, pendingUpdate{update: update, firstUpdateID: firstUpdateID})
+		select {
+		case cache.ch <- update.UpdateID: // Notify SyncOrderbook of most recent update ID for inspection
+		default:
```
gbjk, line 79: 🤷‍♂️ If we don't send it, do we need to log that it's backlogged?

exchanges/gateio/ws_ob_update_manager.go
gbjk: 🚧 I'm also not sold that hotpathishness should excuse not checking errors.
Noting that ... that cost is only incurred if there *is* an error.

So long as `LastUpdateID` is in another package, @gloriousCode's comment is valid that we have to check the error and do something with it.
I wouldn't worry about the ns you're wasting if there is an error.

The point is that in the future the error might be something entirely unexpected that causes us to blow the cache constantly, right?

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
 	}
 
-	lastUpdateID, err := g.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
-	if err != nil && !errors.Is(err, orderbook.ErrDepthNotFound) {
-		return err
-	}
-
-	if lastUpdateID+1 >= firstUpdateID {
-		return applyOrderbookUpdate(g, update)
+	var inUpdateErr error
```
gbjk: 🦍 
```suggestion
	var updateErr error
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+		return nil, currency.ErrCurrencyPairEmpty
+	}
+	if !a.IsValid() {
+		return nil, fmt.Errorf("%w: %q", asset.ErrInvalidAsset, a)
+	}
 	m.mtx.RLock()
 	cache, ok := m.lookup[key.PairAsset{Base: p.Base.Item, Quote: p.Quote.Item, Asset: a}]
 	m.mtx.RUnlock()
 	if !ok {
+		cache = &updateCache{ch: make(chan int64)}
```
gbjk: 🥃 I'd often want to reach for some arbitrary buffer on a channel like this.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
-		limit = optionOrderbookUpdateLimit
+func (c *updateCache) SyncOrderbook(ctx context.Context, e *Exchange, pair currency.Pair, a asset.Item, delay, deadline time.Duration) error {
+	limit, err := c.extractOrderbookLimit(e, a)
+	if err != nil {
+		c.clearWithLock()
+		return err
 	}
 
-	book, err := g.UpdateOrderbookWithLimit(ctx, pair, a, limit)
+	// Rest requests can be behind websocket updates by a large margin, we need to wait here so as to allow the cache to
```
gbjk:
```suggestion
	// REST requests can be behind websocket updates by a large margin, so we wait here to allow the cache to
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
 
-	c.mtx.Lock() // lock here to prevent ws handle data interference with REST request above
-	defer func() {
-		c.updates = nil
-		c.updating = false
-		c.mtx.Unlock()
-	}()
+	// Setting deadline to error out instead of waiting for rate limiter delay
+	// which excessively builds a backlog of pending updates.
+	ctxWDeadline, cancel := context.WithDeadline(ctx, time.Now().Add(deadline))
```
gbjk: 🤷‍♂️ Normally this would just be `ctx` to avoid using the wrong one in future updates below.
Right?
```suggestion
    var cancel context.CancelFunc
	ctx, cancel = context.WithDeadline(ctx, time.Now().Add(deadline))
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+		c.clearWithLock()
 		return err
 	}
 
+	if err := c.waitForUpdate(ctxWDeadline, book.LastUpdateID+1); err != nil {
+		c.clearWithLock()
+		return err
+	}
+
+	c.mtx.Lock() // Lock here to prevent ws handle data interference with REST request above.
```
gbjk: 🥃 This feels racey.
Normally when I do this, I'd want to re-check that state hasn't changed between L144 and L155,
that the `LastUpdateID` is still the same thing.

exchanges/gateio/ws_ob_update_manager.go
gbjk: 🚧 I think that's worth having a comment about in the code, please

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+
+// TODO: When subscription config is added for all assets update limits to use sub.Levels
+func (c *updateCache) extractOrderbookLimit(e *Exchange, a asset.Item) (uint64, error) {
+	switch a {
+	case asset.Spot:
+		sub := e.Websocket.GetSubscription(spotOrderbookUpdateKey)
+		if sub == nil {
+			return 0, fmt.Errorf("%w for %q", subscription.ErrNotFound, spotOrderbookUpdateKey)
+		}
+		// There is no way to set levels when we subscribe for this specific subscription case.
```
gbjk: 🤷 
```suggestion
		// There is no way to set levels when we subscribe for this specific channel
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+		// Extract limit from interval e.g. 20ms == 20 limit book and 100ms == 100 limit book.
```
gbjk: 🐒 From memory there's only 2.
Feels like we should validate that it's what we expect, because _exchange things_

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+	case asset.Options:
+		return optionOrderbookUpdateLimit, nil
+	default:
+		return 0, fmt.Errorf("%w: %q", asset.ErrNotSupported, a)
+	}
+}
+
+// waitForUpdate waits for an update to be available in the cache with an ID
+// that exceeds the specified next update ID. This is needed when the REST book
+// IDs are out of sync and for illiquid pairs.
```
gbjk: 🚧 
* Mismatch in comment and the implementation - we're waiting for equal or greater than 
* Don't need to say why functions might be used 😄 
```suggestion
// waitForUpdate waits for an update with an ID >= nextUpdateID
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+func (c *updateCache) waitForUpdate(ctx context.Context, nextUpdateID int64) error {
+	var updateListLastUpdateID int64
+	c.mtx.Lock()
+	updateListLastUpdateID = c.updates[len(c.updates)-1].update.UpdateID
+	c.mtx.Unlock()
+
+	if updateListLastUpdateID >= nextUpdateID {
+		return nil // No need to wait, the update is already in the cache
```
gbjk: 🦍 
```suggestion
		return nil
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+	}
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case recentPendingUpdateID := <-c.ch:
+			if recentPendingUpdateID >= nextUpdateID {
+				return nil // Update is now available
```
gbjk: 🦍 
```suggestion
				return nil
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+			}
+		}
+	}
 }
 
-// ApplyPendingUpdates applies all pending updates to the orderbook
-func (c *updateCache) applyPendingUpdates(g *Exchange, a asset.Item) error {
+// applyPendingUpdates applies all pending updates to the orderbook
+// requires lock
```
gbjk:
```suggestion
// Does not lock cache
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+func (c *updateCache) applyPendingUpdates(e *Exchange) error {
+	var updateApplied bool
```
gbjk: 🦍 ```suggestion
	var updated bool
```

File: exchanges/gateio/ws_ob_update_manager.go [hunk 38]
```diff
@@ -35,147 +43,236 @@ type pendingUpdate struct {
...
+	c.updating = false
+}
+
 // applyOrderbookUpdate applies an orderbook update to the orderbook
 func applyOrderbookUpdate(g *Exchange, update *orderbook.Update) error {
 	if update.Asset != asset.Spot {
 		return g.Websocket.Orderbook.Update(update)
 	}
 
+	var updateApplied bool
```
gbjk:
```suggestion
	var updated bool
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -184,23 +281,12 @@ func applyOrderbookUpdate(g *Exchange, update *orderbook.Update) error {
...
-
-type channelKey struct {
-	*subscription.Subscription
-}
-
-func (k channelKey) Match(eachKey subscription.MatchableKey) bool {
-	return k.Subscription.Channel == eachKey.GetSubscription().Channel
-}
+	if !updateApplied {
+		return fmt.Errorf("apply orderbook update: %q %q %w", update.Pair, update.Asset, currency.ErrPairNotEnabled)
```
gbjk: 🐒 I think errors should start by saying there's an error
Also: Should this be an err constant?
```suggestion
		return fmt.Errorf("error applying orderbook update: %q %q %w", update.Pair, update.Asset, currency.ErrPairNotEnabled)
```

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -196,9 +299,29 @@ func TestApplyOrderbookUpdate(t *testing.T) {
...
+		Asks:         []orderbook.Level{{Price: 1, Amount: 1}},
+		LastUpdated:  time.Now(),
+		LastPushed:   time.Now(),
+		LastUpdateID: 1336,
+	})
+	require.NoError(t, err)
 
-	update.Pair = currency.NewPair(currency.BABY, currency.BABYDOGE)
+	err = e.Base.SetPairs([]currency.Pair{pair}, asset.Spot, true)
+	require.NoError(t, err)
```
gbjk, line 337:
```suggestion
	require.NoError(t, err, "SetPairs must not error")
```

exchanges/gateio/gateio_wrapper_test.go
shazbert: For spot, when they bring on new pairs, there's usually no liquidity; then this will break intermittently.

shazbert: I hard coded a liquid pair so I will introduce this

exchanges/gateio/ws_ob_update_manager.go
shazbert: No I don't believe so, the update is still in the cache anyway, just means there is no receiver while its still fetching snapshot.

exchanges/gateio/ws_ob_update_manager.go
shazbert: If there was a buffer it would prefill old N update(s) up until a receiver is there. There might be a need for such things, but I don't see it right now?

exchanges/gateio/ws_ob_update_manager.go
shazbert: Is this regarding a stored book change; if by some reason it gets flushed with REST book update as long as the updates are in line with this fetched book which isn't stored this will flush it again. So this potential issue I don't think would be long standing.

exchanges/gateio/ws_ob_update_manager.go
shazbert: Please check new commit as I think I have addressed this now. Hot path stays the same, error is returned back to caller if it is not expected. It will just chatter and cycle on error.

gloriousCode: Thank you for updating it to not ignore the error. I am not much of a fan of how that is looking. I think the design of the `updateCache` struct could be improved to limit the amount of weird pathways going on. I will still take a look and present a diff

exchanges/gateio/gateio_wrapper.go
```diff
@@ -658,16 +658,16 @@ func (e *Exchange) fetchOrderbook(ctx context.Context, p currency.Pair, a asset.
...
-				return nil, fmt.Errorf("%w: %q %q", errNoOrderbookDataAvailable, p, a)
-			}
+	case asset.Margin, asset.CrossMargin:
+		available, err := e.checkInstrumentAvailabilityInSpot(p)
+		if err != nil {
+			return nil, err
 		}
+		if !available {
+			return nil, fmt.Errorf("%w: %w for %q %q", errFetchingOrderbook, errNoSpotInstrument, a, p)
+		}
```
gbjk: 🦍 👑 
```suggestion
		if available, err := e.checkInstrumentAvailabilityInSpot(p); err != nil {
			return nil, err
		} else if !available {
			return nil, fmt.Errorf("%w: %w for %q %q", errFetchingOrderbook, errNoSpotInstrument, a, p)
		}
```

exchanges/gateio/ws_ob_update_manager.go
gbjk: Erm. Okay 🤷  I guess I'll leave it.
My rationale doesn't seem to apply.

The pattern I always expect is that after switching from a RLock to a WLock we then check the state once more, incase another goro got a WLock inbetween and processed a subsequent update to the one we're about to load.

If you understand what I'm saying and you don't think it's relevant, that's fine.

Leaving you to consider and resolve.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +45,240 @@ type pendingUpdate struct {
...
 
-	if a != asset.Spot {
-		if err := g.Websocket.Orderbook.LoadSnapshot(book); err != nil {
+	c.mtx.Lock() // Lock here to prevent ws handle data interference with REST request above.
+	defer func() {
+		c.clearNoLock()
+		c.mtx.Unlock()
+	}()
+
+	if a != asset.Spot { // Regarding Spot, Margin and Cross Margin, the asset is hard coded to `spot in the calling function
```
gbjk: 🚧 
* Sorry; This comment doesn't feel like it helps at all. I know what it's about and I still struggle to process it
* Stray ` in the middle of this
* I think if we're saying this is Spot only, and that mitigates some risk... .then we should remove the non-spot handling for now, and push it to a mothball/gateio_margin_ob branch
I tried to rewrite the comment in any way that says "Actually, this code is never called" and then can't let it live.

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,147 +45,240 @@ type pendingUpdate struct {
...
+		if err := e.Websocket.Orderbook.LoadSnapshot(book); err != nil {
 			return err
 		}
 	} else {
 		// Spot, Margin, and Cross Margin books are all classified as spot
```
gbjk: 🚧 This comment seems misleading, given the if above.
```suggestion
```

exchanges/gateio/ws_ob_update_manager.go
shazbert: Yeah I see what you are saying. The updates are processed linearly by one reader routine so this shouldn't be an issue. Unless we have the same subscription in multiple connections then yes this will be cooked but then we have opened up a potential can of worms.

exchanges/gateio/ws_ob_update_manager.go
shazbert: You make a good point I will RM it! Thanks!

shazbert: [fixed](https://github.com/thrasher-corp/gocryptotrader/pull/1989/commits/f7cf456ea7d79afea1418d9c832cd89bb36dda5b)

gbjk (approved):
Noice 👍

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -153,52 +236,54 @@ func TestApplyPendingUpdates(t *testing.T) {
...
 
-	update.Asset = asset.Spot
-	err = applyOrderbookUpdate(e, update)
-	require.ErrorIs(t, err, orderbook.ErrDepthNotFound)
+func TestClearWithLock(t *testing.T) {
+	t.Parallel()
+	cache := &updateCache{updates: []pendingUpdate{{update: &orderbook.Update{}}}, updating: true}
+	cache.clearWithLock()
+	require.Empty(t, cache.updates)
+	require.False(t, cache.updating)
```
samuael: `require` to `assert`

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -153,52 +236,54 @@ func TestApplyPendingUpdates(t *testing.T) {
...
 
-	update.Pair = currency.NewPair(currency.BABY, currency.BABYDOGE)
-	err = applyOrderbookUpdate(e, update)
-	require.NoError(t, err)
+func TestClearWithNoLock(t *testing.T) {
+	t.Parallel()
+	cache := &updateCache{updates: []pendingUpdate{{update: &orderbook.Update{}}}, updating: true}
+	cache.clearNoLock()
+	require.Empty(t, cache.updates)
+	require.False(t, cache.updating)
```
samuael: same here

exchanges/subscription/keys_test.go
```diff
@@ -150,3 +150,27 @@ func TestGetSubscription(t *testing.T) {
...
+
+func TestChannelKeyMatch(t *testing.T) {
+	t.Parallel()
+	key := ChannelKey{&Subscription{Channel: TickerChannel}}
+	try := &DummyKey{&Subscription{Channel: OrderbookChannel}, t}
+
+	require.Panics(t, func() { key.Match(nil) }, "Match on a nil must panic")
+	require.False(t, key.Match(try), "Match must reject a different channel")
+	try.Channel = TickerChannel
+	require.True(t, key.Match(try), "Match must accept an identical channel")
```
samuael: `require` to `assert`

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -102,49 +113,121 @@ func TestSyncOrderbook(t *testing.T) {
...
+	require.ErrorIs(t, err, context.DeadlineExceeded)
+
+	cache.ch = make(chan int64, 1) // Reset channel to avoid deadlock
+	var wg sync.WaitGroup
+	wg.Go(func() {
+		err = cache.waitForUpdate(t.Context(), 1338)
+	})
+	cache.ch <- 1338
+	wg.Wait()
+	require.NoError(t, err)
```
samuael: For sake of consistency, we should stop using `require`  at the end of the unit test.

exchanges/gateio/gateio_wrapper.go
```diff
@@ -641,21 +641,31 @@ func (e *Exchange) UpdateOrderbook(ctx context.Context, p currency.Pair, a asset
...
 		return nil, err
 	}
 	var o *Orderbook
 	switch a {
-	case asset.Spot, asset.Margin, asset.CrossMargin:
-		var available bool
-		available, err = e.checkInstrumentAvailabilityInSpot(p)
-		if err != nil {
+	case asset.Margin, asset.CrossMargin:
+		if available, err := e.checkInstrumentAvailabilityInSpot(p); err != nil {
```
samuael, line 640: Why do we check the spot tradable pairs when working on margin and cross margin? could we not have a tradable pairs for these too even if the tradable pairs are similar? There may be cases when we need independent tradable pairs for margin and cross margin

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,172 +44,217 @@ type pendingUpdate struct {
...
-	defer func() {
-		c.updates = nil
-		c.updating = false
-		c.mtx.Unlock()
-	}()
+// waitForUpdate waits for an update with an ID >= nextUpdateID
+func (c *updateCache) waitForUpdate(ctx context.Context, nextUpdateID int64) error {
+	var updateListLastUpdateID int64
+	c.mtx.Lock()
+	updateListLastUpdateID = c.updates[len(c.updates)-1].update.UpdateID
```
samuael:
```suggestion
	updateListLastUpdateID := c.updates[len(c.updates)-1].update.UpdateID
```
and remove the declaration above

samuael (commented):
Minor change requests

exchanges/gateio/gateio_wrapper.go
shazbert: This is just to do with orderbooks; there are assets which can be borrowed but don't have any books.

exchanges/gateio/ws_ob_update_manager.go
shazbert: 🔥

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -277,13 +277,13 @@ func TestClearWithLock(t *testing.T) {
 	cache := &updateCache{updates: []pendingUpdate{{update: &orderbook.Update{}}}, updating: true}
 	cache.clearWithLock()
 	require.Empty(t, cache.updates)
-	require.False(t, cache.updating)
+	assert.False(t, cache.updating)
```
gbjk: 🐒 Optional nitpick:
```
	assert.False(t, cache.updating, "updating should be correct after clearWithLock")
```

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -277,13 +277,13 @@ func TestClearWithLock(t *testing.T) {
...
 }
 
 func TestClearWithNoLock(t *testing.T) {
 	t.Parallel()
 	cache := &updateCache{updates: []pendingUpdate{{update: &orderbook.Update{}}}, updating: true}
 	cache.clearNoLock()
 	require.Empty(t, cache.updates)
-	require.False(t, cache.updating)
+	assert.False(t, cache.updating)
```
gbjk: 🐒 Optional nitpick:
```
	assert.False(t, cache.updating, "updating should be correct after clearWithNoLock")
```

exchanges/subscription/keys_test.go
```diff
@@ -166,7 +166,7 @@ func TestChannelKeyMatch(t *testing.T) {
 	require.Panics(t, func() { key.Match(nil) }, "Match on a nil must panic")
 	require.False(t, key.Match(try), "Match must reject a different channel")
 	try.Channel = TickerChannel
-	require.True(t, key.Match(try), "Match must accept an identical channel")
+	assert.True(t, key.Match(try), "Match must accept an identical channel")
```
gbjk: 🚧 Change request:
This should be failing lint.
```
	assert.True(t, key.Match(try), "Match should accept an identical channel")
```

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -153,52 +236,54 @@ func TestApplyPendingUpdates(t *testing.T) {
...
+	cache := &updateCache{updates: []pendingUpdate{{update: &orderbook.Update{}}}, updating: true}
+	cache.clearWithLock()
+	require.Empty(t, cache.updates)
+	assert.False(t, cache.updating)
+}
 
-	update.Pair = currency.NewPair(currency.BABY, currency.BABYDOGE)
-	err = applyOrderbookUpdate(e, update)
-	require.NoError(t, err)
+func TestClearWithNoLock(t *testing.T) {
```
gbjk: 🚧 Change request:
```suggestion
func TestClearNoLock(t *testing.T) {
```

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -102,49 +113,121 @@ func TestSyncOrderbook(t *testing.T) {
...
+	err = cache.SyncOrderbook(t.Context(), e, pair, asset.Spot, 0, time.Second)
+	require.ErrorContains(t, err, context.DeadlineExceeded.Error())
+
+	err = e.Base.SetPairs([]currency.Pair{pair}, asset.Spot, true)
+	require.NoError(t, err)
+	cache.updates = []pendingUpdate{{update: &orderbook.Update{Pair: pair, Asset: asset.Spot, UpdateID: math.MaxInt64}}}
+	err = cache.SyncOrderbook(t.Context(), e, pair, asset.Spot, 0, time.Second)
+	require.ErrorIs(t, err, orderbook.ErrOrderbookInvalid)
+
+	err = e.Base.SetPairs([]currency.Pair{pair}, asset.USDTMarginedFutures, true)
```
gloriousCode, line 187: I would like to see a successful test scenario of `SyncOrderbook`. The coverage of `SyncOrderbook` makes it look like its all good, but you can only test a failure because either:
- If you use a valid currencypair for `USDTMarginedFutures`, it calls `fetchOrderbook` returning the latest live `LastUpdateID`, you can't know the `LastUpdateID` in advance and so you can't setup the orderbook appropriately to run a test
- You have to use `BABY-BABYDOGE` as the pair, loading a custom ob to get around that, but that is not a valid pair for `USDTMarginedFutures` and so that errors

Are you sure you cannot update the test, or the design of `SyncOrderbook` to have a successful testable scenario?

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -35,172 +44,216 @@ type pendingUpdate struct {
...
+		case cache.ch <- update.UpdateID: // Notify SyncOrderbook of most recent update ID for inspection
+		default:
+		}
 		return nil
 	}
 
-	lastUpdateID, err := g.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
-	if err != nil && !errors.Is(err, orderbook.ErrDepthNotFound) {
-		return err
+	lastUpdateID, updateErr := e.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
```
gloriousCode: I find that this function relies on wackiness to get by. I understand that it is battletested etc etc, but I think it could still be improved!

- You call it the first time with the expectation that it will fail on certain checks and nil out the errors
- You prioritise `err == nil` even though it does nothing to speed it up, making the main pathway of the function indented and the failure state appear as the main pathway
- You ignore specific errors because you expect it to fail there and continue on because you want to setup the cache
- This is wasteful calling initially, naturally when things are up and running it becomes less important
- Could be a me problem, but because of the above it took me awhile to understand how this all worked. 

If you knew the _state_ of a cache, you'd be able to know what functions to call. 
- An initialised state would only setup the cache, this would then transition to a queuing state and append the update
- An update while queuing would continue to append
- When the Sync deadline is triggered, an applying state would apply the updates and no new updates would be queued. This would transition to a synchronised state
   - There is a chance that you may wish to still append updates while applying updates, but `cache` is locked, so the updates would all be waiting at the `cache` mutex gate anyway. 
- A synchronised state would only call ApplyUpdate
- If an error occurs on any of these states, it would transition to Desynchronised and would restart the process
- It also removes the `updating` bool since `state` is much more expressive

See this https://gist.github.com/gloriousCode/6368280ba40b7c2684d3c98a123ff397

I think it simplifies things and makes it more readable and maintainable

shazbert: All valid points, this will be abstracted away to reduce code foot print for other orderbook managers, so it was intentional to just pivot around the orderbook ID mismatch because a tighter coupling with the orderbook will eventually happen. But I don't know when this will be done so what you suggest is a better way for now. 

Could you please update the gist so I can git apply it properly your other branch seems to be outdated to what this looks like. It seems like you have done a lot of work to spruce this up but I cannot see the ramifications of this change until its in prod. Then I will update tests. 

Alternatively you can open a PR to address this refactor after this is merged because this has some critical bug fixes and it fixes a test issue. I am also a bit time constrained on other things.

The reason why I blab on about battletested etc etc is because it works and its handling quite a bit of volume between disconnection. There were a few things which made the older design insufficient but that was merged as is without too much design like issues and this is just built upon that.

gloriousCode: That diff was generated on your latest commit from 22 hours ago. Here is a take 2: https://gist.github.com/gloriousCode/5e187d55c9a7b058240b7a811dc7cdbe hopefully this somehow does it.
I don't agree with a bunch there, you don't have to use my diff, but I will need my issues addressed before I can give a tack

gloriousCode: PLEASE ACCEPT MY APOLOGIES RIGHT THIS SECOND!
My patch had a panic on reconnect. It was caused by double-calling `invalidateCache` which cleared updates and panicked on a len check. I wrongfully assumed it was due to lock delays, but the double call caused all sorts of state issues and multiple go-routine spawns.

DESPITE THAT, its not so bad. Here is a freshly baked diff which addresses things and makes other minor changes:
https://gist.github.com/gloriousCode/5236c71f72d7ff85dae60bd2090dfe91

exchanges/gateio/ws_ob_update_manager_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1989/commits/58060e24bd69ed9e93d7706238f1f85d5f95d6a7

exchanges/gateio/ws_ob_update_manager.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/1989/commits/5e4ca4a7646c4edcc9318b331f8402bf0a3ae4a8

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -33,12 +34,20 @@ type wsOBUpdateManager struct {
 }
 
 type updateCache struct {
-	updates  []pendingUpdate
-	updating bool
-	ch       chan int64
-	mtx      sync.Mutex
+	updates []pendingUpdate
+	ch      chan int64
+	mtx     sync.Mutex
```
gbjk: 🐒 Optional nitpick:
Don't think `mtx` is our normal acronym. I normally see just `m`.
```suggestion
	m     sync.Mutex
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -33,12 +34,20 @@ type wsOBUpdateManager struct {
...
+	state   cacheState
 }
 
+type cacheState uint32
+
+const (
+	cacheStateInitialised cacheState = iota
```
gbjk: 🚧 Change request:
Please don't have a meaningful empty value.
```suggestion
	cacheStateUninitialised cacheState = iota
	cacheStateInitialised	
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -57,42 +66,65 @@ func (m *wsOBUpdateManager) ProcessOrderbookUpdate(ctx context.Context, e *Excha
...
+	return nil
+}
 
-	lastUpdateID, updateErr := e.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
-	if updateErr == nil && lastUpdateID+1 == firstUpdateID {
-		if updateErr = e.Websocket.Orderbook.Update(update); updateErr == nil {
-			return nil
+// handleSynchronisedState checks that the incoming update can be applied and applies it, otherwise invalidates the cache
+// assumes lock already active on cache
+func (m *wsOBUpdateManager) handleSynchronisedState(ctx context.Context, e *Exchange, cache *updateCache, firstUpdateID int64, update *orderbook.Update) error {
```
gbjk: 🚧 Change request:
This method name is written purely from the condition that it's called from, and that's reflected by how the comment doesn't reflect the name.
It feels like it's just `applyUpdateImmediately` which is really `applyUpdate.
```suggestion
func (m *wsOBUpdateManager) applyUpdate(ctx context.Context, e *Exchange, cache *updateCache, firstUpdateID int64, update *orderbook.Update) error {
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -57,42 +66,65 @@ func (m *wsOBUpdateManager) ProcessOrderbookUpdate(ctx context.Context, e *Excha
...
 	}
+	return nil
+}
 
-	lastUpdateID, updateErr := e.Websocket.Orderbook.LastUpdateID(update.Pair, update.Asset)
-	if updateErr == nil && lastUpdateID+1 == firstUpdateID {
-		if updateErr = e.Websocket.Orderbook.Update(update); updateErr == nil {
-			return nil
+// handleSynchronisedState checks that the incoming update can be applied and applies it, otherwise invalidates the cache
+// assumes lock already active on cache
```
gbjk: 🐒 Optional nitpick:
* Simplified wording
* When talking about locks, we're not assuming there is one... we're just not locking it. If that makes sense.
```suggestion
// applyUpdate verifies and applies an orderbook update
// Invalidates the cache on error
// Does not benefit from concurrent lock protection
```

exchanges/gateio/ws_ob_update_manager.go
```diff
@@ -57,42 +66,65 @@ func (m *wsOBUpdateManager) ProcessOrderbookUpdate(ctx context.Context, e *Excha
...
+		return m.handleInvalidCache(ctx, e, firstUpdateID, update, cache)
 	}
+	return nil
+}
 
-	// Orderbook notifications are desynced, therefore invalidate store.
-	if err := e.Websocket.Orderbook.InvalidateOrderbook(update.Pair, update.Asset); err != nil && !errors.Is(err, orderbook.ErrDepthNotFound) {
+// handleInvalidCache invalidates the existing orderbook, clears the update queue and reinitialises the orderbook cache
+// assumes lock already active on cache
+func (m *wsOBUpdateManager) handleInvalidCache(ctx context.Context, e *Exchange, firstUpdateID int64, update *orderbook.Update, cache *updateCache) error {
```
gbjk: 🚧 Change request:
```suggestion
func (m *wsOBUpdateManager) invalidateCache(ctx context.Context, e *Exchange, firstUpdateID int64, update *orderbook.Update, cache *updateCache) error {
```

exchanges/gateio/ws_ob_update_manager_test.go
```diff
@@ -37,6 +43,7 @@ func TestProcessOrderbookUpdate(t *testing.T) {
 	})
 	require.NoError(t, err)
 
+	// demonstrate an update is queued
```
gbjk: 🐒 Optional nitpick:
🌴 This occurs a few times
* Comment belongs somewhere in a test message below
* Even if it didn't `demonstrate` 👉 `check` or `test` or `ensure`

gbjk (changes requested):

samuael (approved):
It's a ch ACK!

PR #1995, opened 2025-08-09
Deribit: Replace bespoke message IDs with uuid v7

gbjk:
This PR is a prototype for changes that can be made to most of the exchanges.

UUID V7 is a time ordered unique identifier. I'd generally been clinging to V4 until recently)

By using standard IDs directly in our exchange implementations we reduce inter-linking, testing plane, and overall code.
The aim is that by the end of this websocket itself won't need an message ID generator func or field, and most exchanges will use the V7 uuid. Where not possible, they'll have whatever they need, but called directly.

I'm opting to use `Must()`, which will panic on error, because random number generator failing should be nearly impossible, and we'd nearly definitely panic soon anyway on crypto. I've never seen a `Must(V4)` panic in many years.

Out-Of-Scope:
* First 3 attempts at this I started to refactor and deduplicate the various types in deribit (request types, response types, error types). That turned out to be many yaks hiding behind one-another, so I've avoided touching it too much.
* This is just one exchange first, to keep the review fast and easy. Once merged, I'll tackle the rest in one or maybe two PRs, depending on any outliers that are easier to deal with first.

Motivation:
* When refactoring inside multi-connection websockets I started to hack wildly at the message generator, got frustrated, and now we're here. Welcome 👋 

## Type of change

### DIFF
--- a/exchanges/deribit/deribit_websocket.go
+++ b/exchanges/deribit/deribit_websocket.go
@@ -106,13 +106,6 @@ var defaultSubscriptions = subscription.List{
 	{Enabled: true, Asset: asset.All, Channel: subscription.MyTradesChannel, Interval: kline.HundredMilliseconds, Authenticated: true},
 }
 
-var pingMessage = WsSubscriptionInput{
-	ID:             2,
-	JSONRPCVersion: rpcVersion,
-	Method:         "public/test",
-	Params:         map[string][]string{},
-}
-
 // WsConnect starts a new connection with the websocket API
 func (e *Exchange) WsConnect() error {
 	ctx := context.TODO()
@@ -227,15 +220,11 @@ func (e *Exchange) wsHandleData(ctx context.Context, msgRaw []byte) error {
 		return fmt.Errorf("%s - err %s could not parse websocket data: %s", e.Name, err, msgRaw)
 	}
 	if response.Method == "heartbeat" {
-		return e.Websocket.Conn.SendJSONMessage(ctx, request.Unset, pingMessage)
+		go e.wsSendHeartbeat(ctx)
+		return nil
 	}
-	if response.ID > 2 {
-		if !e.Websocket.Match.IncomingWithData(response.ID, msgRaw) {
-			return fmt.Errorf("can't send ws incoming data to Matched channel with RequestID: %d", response.ID)
-		}
-		return nil
-	} else if response.ID > 0 {
-		return nil
+	if response.ID != "" {
+		return e.Websocket.Match.RequireMatchWithData(response.ID, msgRaw)
 	}
 	channels := strings.Split(response.Params.Channel, ".")
 	switch channels[0] {
@@ -332,6 +321,17 @@ func (e *Exchange) wsHandleData(ctx context.Context, msgRaw []byte) error {
 		}
 	}
 	return nil
+}
+
+func (e *Exchange) wsSendHeartbeat(ctx context.Context) {
+	msg := WsSubscriptionInput{
+		ID:             uuid.Must(uuid.NewV7()).String(),
+		JSONRPCVersion: rpcVersion,
+		Method:         "public/test",
+	}
+	if err := e.Websocket.Conn.SendJSONMessage(ctx, request.Unset, msg); err != nil {
+		log.Errorf(log.ExchangeSys, "%v %s: %s\n", e.Name, errSendingHeartbeat, err)
+	}
 }
 
 func (e *Exchange) processUserOrders(respRaw []byte, channels []string) error {
--- a/exchanges/deribit/deribit_websocket.go
+++ b/exchanges/deribit/deribit_websocket.go
@@ -207,17 +207,17 @@ func (e *Exchange) wsReadData(ctx context.Context) {
 	}
 }
 
-func (e *Exchange) wsHandleData(ctx context.Context, respRaw []byte) error {
-	var response WsResponse
-	err := json.Unmarshal(respRaw, &response)
+func (e *Exchange) wsHandleData(ctx context.Context, msgRaw []byte) error {
+	var response wsResponse
+	err := json.Unmarshal(msgRaw, &response)
 	if err != nil {
-		return fmt.Errorf("%s - err %s could not parse websocket data: %s", e.Name, err, respRaw)
+		return fmt.Errorf("%s - err %s could not parse websocket data: %s", e.Name, err, msgRaw)
 	}
 	if response.Method == "heartbeat" {
 		return e.Websocket.Conn.SendJSONMessage(ctx, request.Unset, pingMessage)
 	}
 	if response.ID > 2 {
-		if !e.Websocket.Match.IncomingWithData(response.ID, respRaw) {
+		if !e.Websocket.Match.IncomingWithData(response.ID, msgRaw) {
 			return fmt.Errorf("can't send ws incoming data to Matched channel with RequestID: %d", response.ID)
 		}
 		return nil
@@ -229,78 +229,78 @@ func (e *Exchange) wsHandleData(ctx context.Context, respRaw []byte) error {
 	case "announcements":
 		announcement := &Announcement{}
 		response.Params.Data = announcement
-		err = json.Unmarshal(respRaw, &response)
+		err = json.Unmarshal(msgRaw, &response)
 		if err != nil {
 			return err
 		}
 		e.Websocket.DataHandler <- announcement
 	case "book":
-		return e.processOrderbook(respRaw, channels)
+		return e.processOrderbook(msgRaw, channels)
 	case "chart":
-		return e.processCandleChart(respRaw, channels)
+		return e.processCandleChart(msgRaw, channels)
 	case "deribit_price_index":
 		indexPrice := &wsIndexPrice{}
-		return e.processData(respRaw, indexPrice)
+		return e.processData(msgRaw, indexPrice)
 	case "deribit_price_ranking":
 		priceRankings := &wsRankingPrices{}
-		return e.processData(respRaw, priceRankings)
+		return e.processData(msgRaw, priceRankings)
 	case "deribit_price_statistics":
 		priceStatistics := &wsPriceStatistics{}
-		return e.processData(respRaw, priceStatistics)
+		return e.processData(msgRaw, priceStatistics)
 	case "deribit_volatility_index":
 		volatilityIndex := &wsVolatilityIndex{}
-		return e.processData(respRaw, volatilityIndex)
+		return e.processData(msgRaw, volatilityIndex)
 	case "estimated_expiration_price":
 		estimatedExpirationPrice := &wsEstimatedExpirationPrice{}
-		return e.processData(respRaw, estimatedExpirationPrice)
+		return e.processData(msgRaw, estimatedExpirationPrice)
 	case "incremental_ticker":
-		return e.processIncrementalTicker(respRaw, channels)
+		return e.processIncrementalTicker(msgRaw, channels)
 	case "instrument":
 		instrumentState := &wsInstrumentState{}
-		return e.processData(respRaw, instrumentState)
+		return e.processData(msgRaw, instrumentState)
 	case "markprice":
 		markPriceOptions := []wsMarkPriceOptions{}
-		return e.processData(respRaw, markPriceOptions)
+		return e.processData(msgRaw, markPriceOptions)
 	case "perpetual":
 		perpetualInterest := &wsPerpetualInterest{}
-		return e.processData(respRaw, perpetualInterest)
+		return e.processData(msgRaw, perpetualInterest)
 	case platformStateChannel:
 		platformState := &wsPlatformState{}
-		return e.processData(respRaw, platformState)
+		return e.processData(msgRaw, platformState)
 	case "quote": // Quote ticker information.
-		return e.processQuoteTicker(respRaw, channels)
+		return e.processQuoteTicker(msgRaw, channels)
 	case "rfq":
 		rfq := &wsRequestForQuote{}
-		return e.processData(respRaw, rfq)
+		return e.processData(msgRaw, rfq)
 	case "ticker":
-		return e.processInstrumentTicker(respRaw, channels)
+		return e.processInstrumentTicker(msgRaw, channels)
 	case "trades":
-		return e.processTrades(respRaw, channels)
+		return e.processTrades(msgRaw, channels)
 	case "user":
 		switch channels[1] {
 		case "access_log":
 			accessLog := &wsAccessLog{}
-			return e.processData(respRaw, accessLog)
+			return e.processData(msgRaw, accessLog)
 		case "changes":
-			return e.processUserOrderChanges(respRaw, channels)
+			return e.processUserOrderChanges(msgRaw, channels)
 		case "lock":
 			userLock := &WsUserLock{}
-			return e.processData(respRaw, userLock)
+			return e.processData(msgRaw, userLock)
 		case "mmp_trigger":
 			data := &WsMMPTrigger{
 				Currency: channels[2],
 			}
-			return e.processData(respRaw, data)
+			return e.processData(msgRaw, data)
 		case "orders":
-			return e.processUserOrders(respRaw, channels)
+			return e.processUserOrders(msgRaw, channels)
 		case "portfolio":
 			portfolio := &wsUserPortfolio{}
-			return e.processData(respRaw, portfolio)
+			return e.processData(msgRaw, portfolio)
 		case "trades":
-			return e.processTrades(respRaw, channels)
+			return e.processTrades(msgRaw, channels)
 		default:
 			e.Websocket.DataHandler <- websocket.UnhandledMessageWarning{
-				Message: e.Name + websocket.UnhandledMessage + string(respRaw),
+				Message: e.Name + websocket.UnhandledMessage + string(msgRaw),
 			}
 			return nil
 		}
@@ -313,7 +313,7 @@ func (e *Exchange) wsHandleData(ctx context.Context, respRaw []byte) error {
 			}
 		default:
 			e.Websocket.DataHandler <- websocket.UnhandledMessageWarning{
-				Message: e.Name + websocket.UnhandledMessage + string(respRaw),
+				Message: e.Name + websocket.UnhandledMessage + string(msgRaw),
 			}
 			return nil
 		}
@@ -325,7 +325,7 @@ func (e *Exchange) processUserOrders(respRaw []byte, channels []string) error {
 	if len(channels) != 4 && len(channels) != 5 {
 		return fmt.Errorf("%w, expected format 'user.orders.{instrument_name}.raw, user.orders.{instrument_name}.{interval}, user.orders.{kind}.{currency}.raw, or user.orders.{kind}.{currency}.{interval}', but found %s", errMalformedData, strings.Join(channels, "."))
 	}
-	var response WsResponse
+	var response wsResponse
 	orderData := []WsOrder{}
 	response.Params.Data = orderData
 	err := json.Unmarshal(respRaw, &response)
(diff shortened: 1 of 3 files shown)


exchanges/deribit/deribit_websocket.go
```diff
@@ -105,44 +106,49 @@ var defaultSubscriptions = subscription.List{
...
 	e.Websocket.Wg.Add(1)
 	go e.wsReadData(ctx)
+	go e.wsStartHeartbeat(ctx)
 	if e.Websocket.CanUseAuthenticatedEndpoints() {
-		err = e.wsLogin(ctx)
-		if err != nil {
+		if err := e.wsLogin(ctx); err != nil {
 			log.Errorf(log.ExchangeSys, "%v - authentication failed: %v\n", e.Name, err)
 			e.Websocket.SetCanUseAuthenticatedEndpoints(false)
+			return err
```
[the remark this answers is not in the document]
gbjk: You're right, it's not usual to return an error on auth connection in other exchanges.

Fixed https://github.com/gbjk/gocryptotrader/commit/e37b94aec

exchanges/deribit/deribit_websocket.go
```diff
@@ -207,100 +213,96 @@ func (e *Exchange) wsReadData(ctx context.Context) {
...
-	if response.ID > 2 {
-		if !e.Websocket.Match.IncomingWithData(response.ID, respRaw) {
-			return fmt.Errorf("can't send ws incoming data to Matched channel with RequestID: %d", response.ID)
-		}
-		return nil
-	} else if response.ID > 0 {
+		go e.wsSendHeartbeat(ctx)
 		return nil
 	}
+	if response.ID != "" {
```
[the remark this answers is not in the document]
gbjk: RequireMatchWIthData returns an error and it's all good.

exchanges/deribit/deribit_websocket.go
```diff
@@ -208,22 +213,18 @@ func (e *Exchange) wsReadData(ctx context.Context) {
...
 	if response.Method == "heartbeat" {
-		return e.Websocket.Conn.SendJSONMessage(ctx, request.Unset, pingMessage)
-	}
-	if response.ID > 2 {
-		if !e.Websocket.Match.IncomingWithData(response.ID, respRaw) {
-			return fmt.Errorf("can't send ws incoming data to Matched channel with RequestID: %d", response.ID)
-		}
-		return nil
-	} else if response.ID > 0 {
+		go e.wsSendHeartbeat(ctx)
```
shazbert, line 222: response isn't being handled

```console
{"params":{"type":"heartbeat"},"method":"heartbeat","jsonrpc":"2.0"}
[DEBUG] | WEBSOCKET | 11/08/2025 07:37:42 | Deribit wss://www.deribit.com/ws/api/v2: Sending message: {"jsonrpc":"2.0","id":"019895ea-fdca-760e-8536-4cc6c167facf","method":"public/test"}
{"jsonrpc":"2.0","id":"019895ea-fdca-760e-8536-4cc6c167facf","result":{"version":"1.2.26"},"usIn":1754861862630620,"usOut":1754861862630694,"usDiff":74,"testnet":false}
[ERROR] | WEBSOCKET | 11/08/2025 07:37:42 | exchange Deribit websocket error - '019895ea-fdca-760e-8536-4cc6c167facf' websocket response to request signature not matched with data {"jsonrpc":"2.0","id":"019895ea-fdca-760e-8536-4cc6c167facf","result":{"version":"1.2.26"},"usIn":1754861862630620,"usOut":1754861862630694,"usDiff":74,"testnet":false}
```

exchanges/deribit/deribit_websocket.go
```diff
@@ -105,44 +106,48 @@ var defaultSubscriptions = subscription.List{
...
 			e.Websocket.SetCanUseAuthenticatedEndpoints(false)
 		}
 	}
-	return e.Websocket.Conn.SendJSONMessage(ctx, request.Unset, setHeartBeatMessage)
+	return nil
+}
+
+func (e *Exchange) wsStartHeartbeat(ctx context.Context) {
+	msg := wsInput{
+		ID:             uuid.Must(uuid.NewV7()).String(),
```
shazbert: Suggestion: The common.Counter is preferable across these response matches because the connection doesn't track IDs they can conflict between sessions. and it has the lowest overhead. But as this is POC to see integration 🤷 doesn't matter; once I start trading on it I will convert them.

shazbert (changes requested):
Thanks, just one issue.

exchanges/deribit/deribit_websocket.go
gbjk: I don't understand how you're getting this.
Your GCT is sending a request because you got a message telling you to, right?
I never get that message 😢  And I tried really hard to work out why.
Might need to pair with you to work out what's different.
[See this thread](https://gocryptotrader.slack.com/archives/C05QQD37JBG/p1754822546884969?thread_ts=1754822517.771929&cid=C05QQD37JBG)

gbjk: Mystery solved: heartbeats only sent if no subscriptions are turned on.
We're lucky you were running like that 🙏

exchanges/deribit/deribit_websocket.go
gbjk: UUID v7 is about 30ns slower than the existing `GenerateMessageID`.
`counter.Counter` is 20x faster
```
BenchmarkCounterUUID-8                   5696677               188.7 ns/op            64 B/op          2 allocs/op
BenchmarkCounterCommon-8               171930782               6.977 ns/op           0 B/op            0 allocs/op
BenchmarkCounterGenerateMessageID-8      7548716               158.4 ns/op            48 B/op          3 allocs/op
```
I think that's acceptable, IMO.

gbjk: But actually, I realised I didn't sniff *which* uuid pkg we're using.
We're using gofrs, which isn't at all bad. It's faster than the ubiquitous google implementation.
But flexstack have a fork of gofrs that is just a tad quicker, ~and that might be worth dropping in~:
```
BenchmarkCounterUUIDGoogle-8             3012028               371.7 ns/op            64 B/op          2 allocs/op
BenchmarkCounterUUIDFlexstack-8          7014712               170.2 ns/op            48 B/op          1 allocs/op
BenchmarkCounterUUIDSamborkent-8         6750514               179.0 ns/op            48 B/op          1 allocs/op
BenchmarkCounterUUIDGofrs-8              6586630               190.4 ns/op            64 B/op          2 allocs/op
BenchmarkCounterCommon-8                168591043                7.024 ns/op           0 B/op          0 allocs/op
BenchmarkCounterGenerateMessageID-8      7555176               158.1 ns/op            48 B/op          3 allocs/op
```

exchanges/deribit/deribit_websocket.go
gbjk: I think the best thing to do here is to catch the heartbeat but still use a unique id.

Fixed https://github.com/gbjk/gocryptotrader/commit/95e92c546

exchanges/deribit/deribit_websocket.go
shazbert: In another PR. Happy to swap the default `defaultGenerateMessageID` for GenerateMessageID to UUID (either googs or whatever UUID package you think is right) and convert the return to string for `ClientOrderID's`. The current default implementation can't guarantee unique numbers anyway, so I am okay with the added overhead. This is helpful for ClientOrderID's as well when it is required and the caller hasn't set it themselves.

I will still convert this over to the `common.Counter` when I start trading that exchange for just message responses if the exchange supports it. Using the `RequestIDGenerator` field when setting up the connection.

shazbert (approved):
Thanks tACK.

exchanges/deribit/deribit_websocket.go
gbjk: That's helpful to know; Any plans I have for this need to allow for swapping out ID generators in all circumstances.
On that basis, I think I need to redo this PR slightly. It's not flexible enough for you.

gbjk:
@shazbert As a result of [this comment](https://github.com/thrasher-corp/gocryptotrader/pull/1995#discussion_r2271468309), I've restructured this to allow you to drop in:
```
func (e *Exchange) MessageID() string {
	return e.messageCounter.IncrementAndGet()
}
```
and I tested that it'll work okay 👍

exchanges/deribit/deribit_websocket.go
```diff
@@ -105,44 +105,48 @@ var defaultSubscriptions = subscription.List{
...
 	}
 	var dialer gws.Dialer
-	err := e.Websocket.Conn.Dial(ctx, &dialer, http.Header{})
-	if err != nil {
+	if err := e.Websocket.Conn.Dial(ctx, &dialer, http.Header{}); err != nil {
 		return err
 	}
 	e.Websocket.Wg.Add(1)
 	go e.wsReadData(ctx)
+	go e.wsStartHeartbeat(ctx)
```
shazbert, line 120: suggestion: can be inlined and erro returned as the connection could terminate randomly anyway if traffic is low.

exchanges/deribit/deribit_websocket.go
```diff
@@ -321,11 +324,22 @@ func (e *Exchange) wsHandleData(ctx context.Context, respRaw []byte) error {
 	return nil
 }
 
+func (e *Exchange) wsSendHeartbeat(ctx context.Context) {
+	msg := WsSubscriptionInput{
+		ID:             "hb-" + e.MessageID(),
```
shazbert, line 330: suggestion: Can in this routine send message return response and check it. but this works anyway 🤷

exchanges/deribit/deribit_ws_endpoints.go
```diff
@@ -2304,7 +2305,7 @@ func (e *Exchange) SendWSRequest(ctx context.Context, epl request.EndpointLimit,
 	}
 	input := &WsRequest{
 		JSONRPCVersion: rpcVersion,
-		ID:             e.Websocket.Conn.GenerateMessageID(true),
+		ID:             uuid.Must(uuid.NewV7()).String(),
```
shazbert:
```suggestion
		ID:             e.MessageID(),
```

exchanges/interfaces.go
```diff
@@ -92,6 +90,7 @@ type IBotExchange interface {
 	PrintEnabledPairs()
 	IsVerbose() bool
 	GetCurrencyTradeURL(ctx context.Context, a asset.Item, cp currency.Pair) (string, error)
+	MessageID() string
```
shazbert: suggestion: Probably doesn't need to be exposed but up to you.

exchanges/deribit/deribit_websocket.go
gbjk: _By "inlined" I think you meant "called immediately without goro", not "inline the function"._

Okay. So you're saying:
* If we fail to start the heartbeat, that's a fatal issue and connection needs to return the error
* That's worth waiting / halting-the-world to know about

Are you sure?

Personally I'm not convinced. I don't like delaying funcs for ancillary bootstrapping.

exchanges/deribit/deribit_websocket.go
gbjk: I chose this route because I don't see the value in wasting our resources doing matching on heartbeats.
My first implementation used `SendMessageReturnResponse` as suggested.

Heartbeats are currently rare though, and only when we have no traffic.
I didn't know that before. It's not documented.
I'm tempted to change back as you suggest, but since heartbeat being no-traffic only isn't documented, I don't want to implement for that if we don't have to.

exchanges/deribit/deribit_ws_endpoints.go
gbjk: Cheeky minx hiding there.

Fixed https://github.com/gbjk/gocryptotrader/commit/0651bc120

exchanges/interfaces.go
gbjk: Agreed. I have no use-case for exposing it right now.
If we ever want to `MessageID` on `IBotExchange` we can add it then.

Fixed https://github.com/gbjk/gocryptotrader/commit/9a5db8244

gbjk:
> changes look good just minor things and a linter issue needs resolving. thanks!

Linter should be fixed. Sloppy. Thanks!

exchanges/deribit/deribit_websocket.go
shazbert: I don't mind can stay as-is; it's not a show stopper. from an external POV it would cease and then you reconnect 🤷

exchanges/deribit/deribit_websocket.go
shazbert: Whatever you want to do man. 🚀

exchanges/deribit/deribit_websocket.go
gbjk: Agreed. Leaving as-is.

exchanges/deribit/deribit_websocket.go
gbjk: Also happy as-is then 👍

shazbert (approved):
Thanks! 🫶

exchanges/exchange.go
```diff
@@ -1962,3 +1963,9 @@ func (*Base) WebsocketSubmitOrder(context.Context, *order.Submit) (*order.Submit
 func (*Base) WebsocketSubmitOrders(context.Context, []*order.Submit) (responses []*order.SubmitResponse, err error) {
 	return nil, common.ErrFunctionNotSupported
 }
+
+// MessageID returns a universally unique id using UUID V7
+// Additional parameters may be added to provide context for the message id
```
gloriousCode:
```suggestion
```
Don't think it adds much to say you can lengthen the string

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -358,6 +358,12 @@ Alternatively you can use `request.WithVerbose(context.Background())` as the `co
 
 Ensure each endpoint is implemented and has an associated test to improve test coverage and increase confidence
 
+#### Message IDs
+
+If the exchange uses unique string message IDs, then your API functions can call e.MessageID() to get a UUID V7 message ID.
+However if the exchange does not support that format then you should override MessageID with an appropriate implementation.
+For Example: Consider `common.Counter` if universal uniqueness is not a requirement and you need an integer id.
```
gloriousCode: Use e.MessageID() to get a UUIDv7 if the exchange supports unique string IDs. Otherwise, override MessageID with a suitable alternative—e.g., common.Counter for simple integer IDs if uniqueness isn't critical.

exchanges/exchange.go
gbjk: Obviously the comment was not clear.
I'm trying to communicate to future maintainers that there's no roadblock to adding params to the method signature.
And I'm definite I want to be saying this, to steer future dev against over complicating things trying to go around this.

How about:
```
// In the future additional params may be added to method signature to provide context for the message id for overriding exchange implementations
```

Fixed https://github.com/gbjk/gocryptotrader/commit/dc4d76f6c

docs/ADD_NEW_EXCHANGE.md
gbjk: Yikes. Yeah.
Struggling to write this correctly in the context of #2008 's change to add MessageSequence.

Minor formatting change to your suggestion:
```
Use e.MessageID() to get a UUIDv7 if the exchange supports unique string IDs. Otherwise override MessageID with a suitable alternative.
For example: Consider common.Counter for simple integer IDs if uniqueness isn't critical.
```

I'll also rewrite in #2008 and can meet you there in a few weeks 😂 

Fixed https://github.com/gbjk/gocryptotrader/commit/5a5307b6f

gbjk: PR 2008 Counterpart, but can/should/will be reviewed later. So .... not sure why I'm saying this.

gbjk@529f8aec8

exchanges/deribit/deribit_websocket.go
```diff
@@ -105,44 +105,48 @@ var defaultSubscriptions = subscription.List{
...
+	if err == nil {
+		var resp wsResponse
+		err = json.Unmarshal(respRaw, &resp)
+		if resp.Error.Code != 0 || resp.Error.Message != "" {
+			err = fmt.Errorf("code: %d message: %s", resp.Error.Code, resp.Error.Message)
+		}
+	}
+	if err != nil {
+		log.Errorf(log.ExchangeSys, "%v %s: %s\n", e.Name, errStartingHeartbeat, err)
+	}
```
gloriousCode: I don't think the saving of lines is worth the finickiness of the way this is done
```suggestion
	respRaw, err := e.Websocket.Conn.SendMessageReturnResponse(ctx, request.Unset, msg.ID, msg)
	if err != nil {
		log.Errorf(log.ExchangeSys, "%v %s: %s\n", e.Name, errStartingHeartbeat, err)
		return
	}
	var resp wsResponse
	if err := json.Unmarshal(respRaw, &resp); err != nil {
		log.Errorf(log.ExchangeSys, "%v %s: %s\n", e.Name, errStartingHeartbeat, err)
	}
	if resp.Error.Code != 0 || resp.Error.Message != "" {
		log.Errorf(log.ExchangeSys, "%s code: %d message: %s", errStartingHeartbeat, resp.Error.Code, resp.Error.Message)
	}
```
Its only two extra lines to have a more standard "handle err immediately, return early, minimise indentation of critical path"

gbjk: Fixed 3ea223a51

thrasher-:
@codex please review

exchanges/deribit/deribit_websocket.go
```diff
@@ -105,44 +105,49 @@ var defaultSubscriptions = subscription.List{
...
+	if err != nil {
+		log.Errorf(log.ExchangeSys, "%v %s: %s\n", e.Name, errStartingHeartbeat, err)
+		return
+	}
+	var resp wsResponse
+	if err := json.Unmarshal(respRaw, &resp); err != nil {
+		log.Errorf(log.ExchangeSys, "%v %s: %s\n", e.Name, errStartingHeartbeat, err)
+	}
+	if resp.Error.Code != 0 || resp.Error.Message != "" {
+		log.Errorf(log.ExchangeSys, "%s code: %d message: %s", errStartingHeartbeat, resp.Error.Code, resp.Error.Message)
```
thrasher-:
```suggestion
		log.Errorf(log.ExchangeSys, "%s %s code: %d message: %s", e.Name, errStartingHeartbeat, resp.Error.Code, resp.Error.Message)
```

Otherwise this would be too ambiguous

gbjk: Fixed gbjk@b99f7b39e

Rebase squashed immediately

exchanges/exchange.go
```diff
@@ -1959,3 +1960,9 @@ func (*Base) WebsocketSubmitOrder(context.Context, *order.Submit) (*order.Submit
 func (*Base) WebsocketSubmitOrders(context.Context, []*order.Submit) (responses []*order.SubmitResponse, err error) {
 	return nil, common.ErrFunctionNotSupported
 }
+
+// MessageID returns a universally unique id using UUID V7
+// In the future additional params may be added to method signature to provide context for the message id for overriding exchange implementations
+func (b *Base) MessageID() string {
+	return uuid.Must(uuid.NewV7()).String()
```
gloriousCode, line 1967: Looking at [the docs](https://pkg.go.dev/github.com/google/uuid#MustParse), this does not appear to be the intended use-case for the `Must` function. This is introducing the risk of random panics into GCT. I am willing to take that risk and I don't believe this should ever create such a scenario. So really this is just a reserved "but I said it", so I can be smug one day maybe..

gloriousCode: oh silly me. I was reading "must parse". look at that egg on my face

thrasher-: Despite the func mixup, your point is still valid that it theoretically can happen; things like insufficient entropy on a given system; neutrons smashing hardware causing bits to shift randomly, hardware/electrical faults with non-ECC memory also causing problems but you'd have cascading issues everywhere else throughout the application (anything using crypto, random number generation etc). That's also assuming no lib upstream issues though we should be able to catch those with tests whenever they perform updates. Can test in prod and if it ever becomes too much of a hassle albeit extremely rare we can adjust 🚀

gbjk: FWIW: I did acknowledge risk and mitigation in the PR description 🙂

PR #2012, opened 2025-08-22
linters: Add modernise tool check and fix issues

thrasher-:
# PR Description

Adds a Go modernise tool check to GHA, consolidates some existing checks into the misc checks workflow and fixes the reported issues based on the new Go 1.25 and future features. Adds a new make recipe `make moderise` to make testing easy. I've also opted to use AU english as opposed to US despite the tool being spelt with a z. Happy to change it though if people want it aligned like that

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Running the moderniser tool

## Checklist


gloriousCode (approved):
I'm happy with the changes and have checked out the documentation for wg.Go. The other areas that could do wg.Go require more effort

tACK

.github/workflows/misc.yml
```diff
@@ -62,4 +67,27 @@ jobs:
...
+            echo "jq differences found in $file! Please run 'make lint_configs'"
+            exit 1
+          else
+            rm $processed_file
+            echo "No differences found in $file 🌞"
+          fi
+        done
+    
+    - name: Check config versions are continuous
+      run: go test ./config/versions/ -tags config_versions -run Continuity
```
gbjk: The idea behind this being in a separate workflow is that it's Okay for it to fail right up until a PR is merged.
So what I wanted with it being separate was that we can quickly see to ignore that entire workflow on PRs during review.

If we move it into misc.yml then everyone has to open up the CI for that workflow to check if it was that or not.

I should have left it as a comment inside the yaml file.

So, I think my justification stands and config version continuity should remain a separate and easy to ignore workflow, but with a new comment explaining why not to fold it into misc.

gbjk (changes requested):
Nice! 🎉 

Only one change request from me

.github/workflows/misc.yml
thrasher-: Yep I understand and remember your motivation; similar to how I've left the spellchecker and proto checks alone. The main caveat is how infrequently we do config upgrades, if misc fails for whatever reason they'd have to check it once every blue moon

gbjk: Okay. I've considered that.
I stand by it, though, because for a given PR that does introduce a config upgrade, they'd be checking that same thing every time they garden their PR.
🙏

gbjk (approved):

PR #2029, opened 2025-09-08
app/feature: Quickspy

gloriousCode:
# PR Description

This is a feature I made in my superTrader5000 and I use it extensively. I have removed some unnecessary things for GCT and made it much more robust for public use instead of being held up by twigs and `panic("help")`. I think this feature would be of use for traders and noobies who just want to get some data without needing to understand many of GCT's systems

The `quickspy` package provides a means to quickly request and receive data for an individual exchange, asset and currency pair. For the times you just really want to get some data without fussing about with configs - setting currency pairs or API keys - or `SetDefaults()` functions.

## Features
- Quickly creates an exchange with only the selected asset and currency pair enabled
- Supports a range of focus data types allowing for a more tailored approach to data retrieval
- Supports both REST and Websocket data retrieval methods
- Supports both public and authenticated data retrieval methods
- Three types of quickspy implementations:
  - `Quickspy` - supports multiple focus data types, has the finest level of control over data retrieval method and frequency
  - `QuickerSpy` - supports a single focus data type, prioritises websocket and allows for control over the quickspy instance
  - `QuickestSpy` - supports a single focus data type, prioritises websocket and returns a chan of data for the caller to consume

For more details, check out the readmes.

#### Quicknotes
- a few things in /cmd/quickspy were AI generated, I'll fix up anything janky 🤖 
- These features rely on GCT standards and wrappers, as such, if you come across issues like "gateio delivery futures FARTCOIN/BTC is not streaming alpha", then a separate PR should be raised to address that
- My last PR had a few out of scope PR requests. I will be resisting them as this kind of feature will be very easy to request extras and its already big enough

## Type of change

## How has this been tested
- [x] go test ./... -race
- [x] golangci-lint run
- [x] Coverage of all new functionality


exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,634 @@
...
+	}
+	b.Config.Features.Subscriptions = newSubs
+	b.Features.Subscriptions = newSubs
+	if err := b.Websocket.EnableAndConnect(); err != nil {
+		if !errors.Is(err, websocket.ErrWebsocketAlreadyEnabled) {
+			return fmt.Errorf("%s: %w", q.key.ExchangeAssetPair, err)
+		}
+		// Because EnableAndConnect errors if its already enabled, but also wants to connect
+		// we have to do this silly handling, making everyone suffer
+		// the complaint was generated by AI, there must be a lot of complaining in scraped comments
```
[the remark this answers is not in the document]
gloriousCode: You generated those unprofessional comments first! 😆

shazbert:
@codex review plz?

shazbert:
😢

thrasher-:
> @codex review plz?

I'm on it

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,634 @@
...
+		return err
+	}
+	focus.setSuccessful()
+	return nil
+}
+
+// Shutdown stops all routines and websocket connections
+func (q *QuickSpy) Shutdown() {
+	q.credContext.Done()
+}
```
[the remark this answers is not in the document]
gloriousCode: Thank you Mr Ai! I did do this wrong and you taught me things. I've opted to use a shutdown channel. Quickspy will still respect a user's context cancel(), but if one didn't make a context with a cancel they can still close it via channel

thrasher-:
> I think this feature would be of use for traders and noobies who just want to get some data without needing to understand many of GCT's systems

From this perspective: "I think this feature would be of use for traders and noobies who just want to get some data without needing to understand many of GCT's systems" it would be very useful to newcomers since some of the most common questions we get are things of this nature and people say that GCT has a steep learning curve. Also helps flesh out helpers in the lib more so we can improve streaming service one liners (like this: https://github.com/ccxt/go-binance?tab=readme-ov-file#depth). The point about more code to maintain is a fair point but I think it'll help us improve the lib over time to be more helpful

cmd/documentation/exchange_templates/exchange_quickspy_readme.tmpl
```diff
@@ -0,0 +1,119 @@
...
+The `quickspy` package provides a means to quickly request and receive data for an individual exchange, asset and currency pair. For the times you just really want to get some data without fussing about with configs - setting currency pairs or API keys - or `SetDefaults()` functions.
+
+## Features
+
+- Quickly creates an exchange with only the selected asset and currency pair enabled
+- Supports a range of focus data types allowing for a more tailored approach to data retrieval
+- Supports both REST and Websocket data retrieval methods
+- Supports both public and authenticated data retrieval methods
+- Three types of quickspy implementations:
+  - `Quickspy` - supports multiple focus data types, has the finest level of control over data retrieval method and frequency
```
shazbert:
```suggestion
  - `QuickSpy` - supports multiple focus data types, has the finest level of control over data retrieval method and frequency
```

shazbert: or all lower case and then update headings

cmd/documentation/exchange_templates/exchange_quickspy_readme.tmpl
```diff
@@ -0,0 +1,119 @@
...
+
+### Quickerspy for account info focus type with credentials provided by context
+```go
+func main() {
+	ctx := context.Background()
+	credentials := &account.Credentials{
+		Key:    "abc",
+		Secret: "123",
+	}
+	ctx = account.DeployCredentialsToContext(ctx, credentials)
```
shazbert:
```suggestion
	ctx := account.DeployCredentialsToContext(context.Background(), &account.Credentials{
		Key:    "abc",
		Secret: "123",
	})
```

cmd/documentation/exchange_templates/exchange_quickspy_readme.tmpl
```diff
@@ -0,0 +1,119 @@
...
+		log.Fatalf("could not create quickspy instance: %v", err)
+	}
+	parseData(ctx, qs)
+}
+
+func parseData(ctx context.Context, c <-chan any) {
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
shazbert: suggestion: Should return `ctx.Err()`

cmd/documentation/exchange_templates/exchange_quickspy_readme.tmpl
```diff
@@ -0,0 +1,119 @@
...
+| URL | Yes | No | No | No |
+
+## Usage
+
+There are multiple ways to utilise a quickspy. See `/cmd/quickspy` for a basic way of establishing a single purpose quickspy that subscribes to data and prints it to console.
+
+### Quickspy with two focus types
+```go
+func main() {
+	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
```
shazbert: suggestion: defer cancel

cmd/quickspy/main.go
```diff
@@ -0,0 +1,461 @@
...
+	// Credentials (only applied when supplied or required by focus)
+	var creds *account.Credentials
+	if quickspy.RequiresAuth(fType) {
+		creds = &account.Credentials{
+			Key:             strings.TrimSpace(*apiKey),
+			Secret:          strings.TrimSpace(*apiSecret),
+			SubAccount:      strings.TrimSpace(*subAccount),
+			ClientID:        strings.TrimSpace(*clientID),
+			OneTimePassword: strings.TrimSpace(*otp),
+			PEMKey:          strings.TrimSpace(*pemKey),
```
shazbert: suggestion: Might not be the best trimming PEM key here. Might need to validate it after to see if it can be used like:

```go
if creds.PEMKey != "" {
    if block, _ := pem.Decode([]byte(creds.PEMKey)); block == nil {
        return fmt.Errorf("invalid PEM key format")
    }
}
```

cmd/quickspy/main.go
```diff
@@ -0,0 +1,461 @@
...
+
+// emit writes NDJSON events to stdout.
+func emit(ev eventEnvelope) {
+	if err := enc.Encode(ev); err != nil {
+		_, _ = fmt.Fprintf(os.Stderr, "failed to emit event: %v %v\n", ev, err)
+	}
+}
+
+func parseFocusType(s string) quickspy.FocusType {
+	// returns (focusType, useWebsocketDefault)
```
shazbert: RM?

cmd/quickspy/main.go
```diff
@@ -0,0 +1,461 @@
...
+	case "activeorders", "orders":
+		return quickspy.ActiveOrdersFocusType
+	case "orderexecution", "executionlimits", "limits":
+		return quickspy.OrderLimitsFocusType
+	case "url", "tradeurl", "trade_url":
+		return quickspy.URLFocusType
+	case "contract":
+		return quickspy.ContractFocusType
+	default:
+		return quickspy.UnsetFocusType
```
shazbert: return error as unhandled `%q` of s

cmd/quickspy/main.go
```diff
@@ -0,0 +1,461 @@
...
+		lvl := b.Bids[i]
+		cumulativeBids += lvl.Amount
+		outPrintf("%s% -14.8f % -14.8f % -14.8f%s\n", ansiGreen, lvl.Price, lvl.Amount, cumulativeBids, ansiReset)
+	}
+}
+
+func renderTicker(t *ticker.Price) {
+	if t == nil {
+		return
+	}
```
shazbert, line 282: param checks can be removed and below as its in a case

cmd/quickspy/main.go
```diff
@@ -0,0 +1,461 @@
...
+}
+
+func renderPrettyPayload(payload any, bookLevels int) {
+	switch v := payload.(type) {
+	case *orderbook.Book:
+		renderOrderbook(v, bookLevels)
+	case []ticker.Price:
+		if len(v) > 0 {
+			renderTicker(&v[0])
+		}
```
shazbert: why the first one?

cmd/quickspy/main.go
```diff
@@ -0,0 +1,461 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case d := <-c:
+			switch {
+			case cfg.JSONOnly:
+				emit(eventEnvelope{Timestamp: time.Now().UTC(), Focus: cfg.FocusType.String(), Data: d})
+			default:
+				clearScreen()
+				heading := fmt.Sprintf("%s | %s | %s", cfg.Exchange, cfg.Asset, cfg.Pair)
```
shazbert: can be declared up top above for loop

cmd/quickspy/main.go
```diff
@@ -0,0 +1,461 @@
...
+		return
+	}
+	outPrintf("%sFunding:%s latest=%.6f predicted=%.6f next=%s checked=%s\n",
+		ansiBold, ansiReset,
+		fr.LatestRate.Rate.InexactFloat64(), fr.PredictedUpcomingRate.Rate.InexactFloat64(),
+		fr.TimeOfNextRate.UTC().Format(time.RFC3339), fr.TimeChecked.UTC().Format(time.RFC3339))
+}
+
+// intMin for dealing with ints
+func intMin(a, b int) int {
```
shazbert, line 471: Use inbuilt func `min` and rm test

cmd/quickspy/main_test.go
```diff
@@ -0,0 +1,227 @@
...
+	ctx, cancel := context.WithCancel(context.Background())
+	ch := make(chan any, 1)
+	cfg := &appConfig{FocusType: quickspy.TickerFocusType, JSONOnly: true}
+	ch <- "test-data"
+	done := make(chan error, 1)
+	go func() { done <- streamData(ctx, ch, cfg) }()
+	// Allow goroutine to process the first item
+	time.Sleep(50 * time.Millisecond)
+	cancel()
+	select {
```
shazbert, line 76: suggestion: Can use Eventually function from assert/require

exchange/quickspy/data.go
```diff
@@ -0,0 +1,498 @@
...
+	q.m.Lock()
+	q.data.AccountBalance = payload
+	q.m.Unlock()
+	focus.stream(payload)
+	focus.setSuccessful()
+	return nil
+}
+
+func (q *QuickSpy) handleWSAccountChanges(data []account.Change) error {
+	focus := q.focuses.GetByFocusType(AccountHoldingsFocusType)
```
shazbert, line 51: missing nil guard check either rm them all or add to all.

exchange/quickspy/data.go
```diff
@@ -0,0 +1,498 @@
...
+	}
+	resp, err := q.exch.UpdateTicker(q.credContext, q.key.ExchangeAssetPair.Pair(), q.key.ExchangeAssetPair.Asset)
+	if err != nil {
+		return fmt.Errorf("%s %q %w", q.key.ExchangeAssetPair, focus.focusType.String(), err)
+	}
+	q.m.Lock()
+	q.data.Ticker = resp
+	q.m.Unlock()
+	focus.stream(resp)
+
```
shazbert: rm line just to be the same as others.

exchange/quickspy/focus.go
```diff
@@ -0,0 +1,248 @@
...
+	ErrUnsetFocusType                  = errors.New("focus type is unset")
+	ErrUnsupportedFocusType            = errors.New("unsupported focus type")
+	ErrInvalidRESTPollTime             = errors.New("invalid REST poll time")
+	ErrInvalidAssetForFocusType        = errors.New("invalid asset for focus type")
+	ErrCredentialsRequiredForFocusType = errors.New("credentials required for this focus type")
+	ErrNoCredentials                   = errors.New("no credentials provided")
+)
+
+// FocusType is an identifier for data types that quickspy can gather
+type FocusType int
```
shazbert: uint(n) variant okay?

exchange/quickspy/focus.go
```diff
@@ -0,0 +1,248 @@
...
+
+// FocusData contains information on what data quickspy should gather
+// how it should be gathered as well as a channel for delivering that data
+type FocusData struct {
+	focusType             FocusType
+	useWebsocket          bool
+	restPollTime          time.Duration
+	m                     *sync.RWMutex
```
shazbert: Does this need to be a pointer? Stops copy if not.

exchange/quickspy/focus.go
```diff
@@ -0,0 +1,248 @@
...
+
+// Validate checks if the FocusData instance is good to go
+// It is assumed that this will be called before use rather than willy-nilly
+func (f *FocusData) Validate(k *CredentialsKey) error {
+	if err := common.NilGuard(f); err != nil {
+		return err
+	}
+	if err := common.NilGuard(k); err != nil {
+		return err
+	}
```
shazbert:
```suggestion
	if err := common.NilGuard(f, k); err != nil {
		return err
	}
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/trade"
+	"github.com/thrasher-corp/gocryptotrader/log"
+)
+
+// NewQuickSpy returns a running quickspy if everything passed in is valid
+func NewQuickSpy(ctx context.Context, k *CredentialsKey, focuses []*FocusData) (*QuickSpy, error) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
shazbert: not need can be removed, all ctx params must be populated.

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+func NewQuickerSpy(ctx context.Context, k *key.ExchangeAssetPair, focus FocusType) (*QuickSpy, error) {
+	if err := common.NilGuard(k); err != nil {
+		return nil, err
+	}
+	if err := focus.Valid(); err != nil {
+		return nil, err
+	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
shazbert: can be rm'd

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+	focusData := NewFocusData(focus, false, useWS, time.Second)
+	ck := &CredentialsKey{
+		ExchangeAssetPair: *k,
+		Credentials:       account.GetCredentialsFromContext(ctx),
+	}
+	q, err := NewQuickSpy(ctx, ck, []*FocusData{focusData})
+	if err != nil {
+		return nil, err
+	}
+	return q, nil
```
shazbert:
```suggestion
	return NewQuickSpy(ctx, ck, []*FocusData{focusData})
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+func NewQuickestSpy(ctx context.Context, k *key.ExchangeAssetPair, focus FocusType) (chan any, error) {
+	if err := common.NilGuard(k); err != nil {
+		return nil, err
+	}
+	if err := focus.Valid(); err != nil {
+		return nil, err
+	}
+	if ctx == nil {
+		ctx = context.Background()
+	}
```
shazbert: rm

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+}
+
+// GetAndWaitForFocusByKey is a convenience function to wait for a quickspy to be setup and have data
+// before utilising the desired focus type
+func (q *QuickSpy) GetAndWaitForFocusByKey(ctx context.Context, focusType FocusType, timeout time.Duration) (*FocusData, error) {
+	focus, err := q.GetFocusByKey(focusType)
+	if err != nil {
+		return nil, err
+	}
+	timer := time.NewTimer(timeout)
```
shazbert: can use `time.After()`

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+	}
+
+	if !b.Features.Supports.Websocket {
+		return fmt.Errorf("exchange %s has no websocket. Websocket requirement was enabled", q.key.ExchangeAssetPair.Exchange)
+	}
+	if err := common.NilGuard(b.Websocket); err != nil {
+		return fmt.Errorf("%s %w", q.key.ExchangeAssetPair, err)
+	}
+	// allows routing of all websocket data to our custom one
+	b.Websocket.ToRoutine = q.dataHandlerChannel
```
shazbert: Swap instead of creating a data handler channel?

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+		err = b.Websocket.Connect()
+		if err != nil {
+			return fmt.Errorf("%s: %w", q.key.ExchangeAssetPair, err)
+		}
+	}
+	err := q.validateSubscriptions(newSubs)
+	if err != nil {
+		return err
+	}
+	return nil
```
shazbert:
```suggestion
	return q.validateSubscriptions(newSubs)
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+		if !errors.Is(err, websocket.ErrWebsocketAlreadyEnabled) {
+			return fmt.Errorf("%s: %w", q.key.ExchangeAssetPair, err)
+		}
+		// EnableAndConnect returns an error if the websocket is already enabled,
+		// but a connection still needs to be established. In this case, we manually
+		// call Connect to ensure the websocket is connected.
+		err = b.Websocket.Connect()
+		if err != nil {
+			return fmt.Errorf("%s: %w", q.key.ExchangeAssetPair, err)
+		}
```
shazbert:
```suggestion
		if err := b.Websocket.Connect(); err != nil {
			return fmt.Errorf("%s: %w", q.key.ExchangeAssetPair, err)
		}
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+	return nil
+}
+
+func (q *QuickSpy) run() {
+	if q.AnyRequiresWebsocket() {
+		q.wg.Go(func() {
+			err := q.handleWS()
+			if err != nil {
+				log.Errorf(log.QuickSpy, "%s websocket handler error: %v", q.key.ExchangeAssetPair, err)
+			}
```
shazbert:
```suggestion
			if err := q.handleWS(); err != nil {
				log.Errorf(log.QuickSpy, "%s websocket handler error: %v", q.key.ExchangeAssetPair, err)
			}
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+		})
+	}
+	for _, focus := range q.focuses.List() {
+		if focus.useWebsocket {
+			continue
+		}
+		q.wg.Add(1) // wg.Go doesn't work here as we have to pass in the focus variable
```
shazbert: You sure it should just close over the variable?

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+		go func(f *FocusData) {
+			defer q.wg.Done()
+			err := q.runRESTRoutine(f)
+			if err != nil {
+				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+					return
+				}
+				f.stream(err)
+			}
```
shazbert:
```suggestion
			if err := q.runRESTRoutine(f); err != nil {
				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
					return
				}
				f.stream(err)
			}
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+				log.Errorf(log.QuickSpy, "%s %s", q.key.ExchangeAssetPair, err)
+			}
+		}
+	}
+}
+
+func (q *QuickSpy) handleWSData(d any) error {
+	switch data := d.(type) {
+	case account.Change:
+		return q.handleWSAccountChange(&data)
```
shazbert, line 460: Why is this a pointer to an `any` and below?

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+		select {
+		case <-q.shutdown:
+			return nil
+		case <-q.credContext.Done():
+			return q.credContext.Err()
+		case <-timer.C:
+			err := q.processRESTFocus(f)
+			if err != nil {
+				return err
+			}
```
shazbert:
```suggestion
			if err := q.processRESTFocus(f); err != nil {
				return err
			}
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+	if f.useWebsocket {
+		return nil
+	}
+	timer := time.NewTimer(0)
+	for {
+		select {
+		case <-q.shutdown:
+			return nil
+		case <-q.credContext.Done():
+			return q.credContext.Err()
```
shazbert: OOS: 🍺 It might be better for future metrics aggregation to have this param only and populate down between callers.

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+				return nil
+			}
+			timer.Reset(f.restPollTime)
+		}
+	}
+}
+
+func (q *QuickSpy) processRESTFocus(f *FocusData) error {
+	err := q.handleFocusType(f.focusType, f)
+	if err != nil {
```
shazbert, line 505: suggestion: can swap and not indent all this code but up to you. Or use if short on what's there.

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+
+// DumpJSON conveniently gives you JSON output of all gathered data
+func (q *QuickSpy) DumpJSON() ([]byte, error) {
+	q.m.RLock()
+	defer q.m.RUnlock()
+	b, err := json.MarshalIndent(q.data, "", "  ")
+	if err != nil {
+		return nil, err
+	}
+	return b, nil
```
shazbert:
```suggestion
	return json.MarshalIndent(q.data, "", "  ")
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	case <-focus.hasBeenSuccessfulChan:
+		return nil
+	}
+}
+
+// WaitForInitialDataWithTimer waits for initial data for a focus type or cancels when ctx is done.
+func (q *QuickSpy) WaitForInitialDataWithTimer(ctx context.Context, focusType FocusType, tt time.Duration) error {
```
shazbert:
```suggestion
func (q *QuickSpy) WaitForInitialDataWithTimeout(ctx context.Context, focusType FocusType, timeout time.Duration) error {
```

exchange/quickspy/quickspy.go
```diff
@@ -0,0 +1,640 @@
...
+	if tt == 0 {
+		return fmt.Errorf("%w: timer cannot be 0", errTimerNotSet)
+	}
+	focus := q.focuses.GetByFocusType(focusType)
+	if focus == nil {
+		return fmt.Errorf("%w %q", errKeyNotFound, focusType)
+	}
+	t := time.NewTimer(tt)
```
shazbert: use time.After

exchange/quickspy/quickspy_types.go
```diff
@@ -0,0 +1,86 @@
...
+	exch exchange.IBotExchange
+	// Key contains exchange, pair, and asset information
+	key *CredentialsKey
+	// focuses is a map of focus types to focus options
+	// Don't access directly, use functions to handle locking
+	focuses *FocusStore
+	// dataHandlerChannel is used for receiving data from websockets
+	dataHandlerChannel chan any
+	// m is used for concurrent read/write operations
+	m *sync.RWMutex
```
shazbert: Does this need to be a pointer?

exchange/quickspy/store.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+// Upsert adds or updates FocusData in the store.
+// If data is nil, the method does nothing. Use Remove to delete entries.
+func (s *FocusStore) Upsert(key FocusType, data *FocusData) {
+	s.m.Lock()
+	defer s.m.Unlock()
+	if data == nil {
+		return
+	}
```
shazbert, line 28: can be above lock or removed

exchange/quickspy/store.go
```diff
@@ -0,0 +1,78 @@
...
+
+// GetByFocusType returns FocusData if exists
+func (s *FocusStore) GetByFocusType(key FocusType) *FocusData {
+	s.m.RLock()
+	defer s.m.RUnlock()
+	data, ok := s.s[key]
+	if !ok {
+		return nil
+	}
+	return data
```
shazbert:
```suggestion
	return s.s[key]
```

exchange/quickspy/store.go
```diff
@@ -0,0 +1,78 @@
...
+// Note: order of the slice is not guaranteed
+// Note: is unsafe
+func (s *FocusStore) List() []*FocusData {
+	s.m.RLock()
+	defer s.m.RUnlock()
+	list := make([]*FocusData, 0, len(s.s))
+	for _, v := range s.s {
+		list = append(list, v)
+	}
+	return list
```
shazbert: suggestion: you can one liner this but I don't know the performance sillyness:

```go
slices.Collect(maps.Values(s.s))
```

exchanges/binance/binance_websocket.go
```diff
@@ -99,7 +99,12 @@ func (e *Exchange) WsConnect() error {
 	e.Websocket.Wg.Add(1)
 	go e.wsReadData()
 
-	e.setupOrderbookManager(ctx)
+	for _, sub := range e.Features.Subscriptions {
+		if sub.Channel == subscription.OrderbookChannel { // don't presume orderbook is part of the supplied subscriptions
+			e.setupOrderbookManager(ctx)
+			break
+		}
+	}
```
shazbert: Does this cause issues though on connect?

exchange/quickspy/quickspy.go
gloriousCode: For this one, it has been found to be an account.Change which is the current way we pass in account things (at least per websocketroutine_manager), so when passing it around, I make it a pointer. `d` is any, `data` is `account.Change`.

exchanges/binance/binance_websocket.go
gloriousCode: I had a blown up binance when I didn't subscribe to the otherbook. I'll reverify for you, but I believe I added it after panics

cmd/quickspy/main.go
gloriousCode: AI things

cmd/quickspy/main.go
gloriousCode: AI things

exchange/quickspy/data.go
gloriousCode: I didn't for slices since the for loop would do nothing, I will add the step for you

exchange/quickspy/quickspy.go
gloriousCode: Want me to remove it from `func (r *Requester) SendPayload`?

shazbert: Yes please!

exchange/quickspy/quickspy.go
gloriousCode: ?

exchanges/binance/binance_websocket.go
gloriousCode: Confirming I don't get a panic there 🤷 must have been a spooky ghost

exchange/quickdata/quickdata.go
```diff
@@ -0,0 +1,614 @@
...
+			defer q.wg.Done()
+			if err := q.runRESTRoutine(f); err != nil {
+				if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+					return
+				}
+				f.stream(err)
+			}
+		}(focus)
+	}
+}
```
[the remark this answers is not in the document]
gloriousCode: I did just look and the only errors are about keys not being found; unimportant errors and the error could not be streamed to a channel as it does not exist. It is logging the event of weird websocket subscription data input

cmd/quickdata/app/focus.go
```diff
@@ -0,0 +1,244 @@
...
+// the flag is only set once, preventing multiple goroutines from setting it
+// simultaneously. If the flag has already been set, it returns immediately
+// without doing anything further.
+func (f *FocusData) setSuccessful() {
+	f.m.RLock()
+	if f.hasBeenSuccessful {
+		f.m.RUnlock()
+		return
+	}
+	f.m.RUnlock()
```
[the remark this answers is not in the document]
gloriousCode: But that is why there is an additional check

exchange/quickdata/quickspy_types.go
```diff
@@ -0,0 +1,86 @@
...
+	ExchangeAssetPair key.ExchangeAssetPair `json:"exchangeAssetPair"`
+}
+
+// QuickData is a struct that holds data on a single exchange pair asset
+// Its purpose is to continuously generate metadata on the market
+// it is write-side oriented vs the comparer
+type QuickData struct {
+	// credContext is the context for credentials
+	// also used for cancelling goroutines
+	credContext context.Context //nolint:containedctx // allows for repeated credential usage and cancellation. Happy to hear of other ways
```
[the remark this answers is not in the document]
gloriousCode: This is good and I will give it a go for you Mr AI

thrasher-:
@codex please review

cmd/documentation/exchange_templates/exchange_quickdata_readme.tmpl
```diff
@@ -0,0 +1,117 @@
...
+### QuickestData to stream ticker data as fast as possible
+```go
+func main() {
+	ctx, _ := signal.NotifyContext(context.Background(), syscall.SIGINT, syscall.SIGTERM)
+	k := key.NewExchangeAssetPair("Binance", asset.Spot, currency.NewBTCUSDT())
+	qs, err := quickdata.NewQuickestData(ctx, &k, quickdata.TickerFocusType)
+	if err != nil {
+		log.Fatalf("could not create quickData instance: %v", err)
+	}
+	parseData(ctx, qs)
```
shazbert: need to handle error here

cmd/quickdata/main_test.go
```diff
@@ -225,3 +240,21 @@ func TestStreamData_DefaultRendering(t *testing.T) {
...
+	return nil
+}
+
+func transferStuff(a any) <-chan any {
+	ch := make(chan any, 1)
+	ch <- a
+	return ch
+}
+
+func TestButts(t *testing.T) {
```
shazbert: rm and above? 😆

exchange/quickdata/quickdata_test.go
```diff
@@ -282,35 +280,35 @@ func TestLatestData(t *testing.T) {
...
 		t.Parallel()
-		q := mustQuickSpy(t, TickerFocusType)
-		_, err := q.LatestData(999)
+		q := mustQuickData(t, TickerFocusType)
+		_, err := q.LatestData(111)
 		require.ErrorIs(t, err, errKeyNotFound)
 	})
 }
 
 func TestWaitForInitialDataWithTimer(t *testing.T) {
```
shazbert: Update test name

cmd/quickdata/main_test.go
gloriousCode: WHOOPS 😄 ty

cmd/documentation/cmd_templates/quickdata_readme.tmpl
```diff
@@ -0,0 +1,38 @@
+{{define "cmd quickdata" -}}
+{{template "header" .}}
+
+## Current Features for {{.Name}}
+QuickData is an applet to quickly fetch data from an exchange without needing to fuss about with configs or setting currency pairs or calling functions like `SetDefaults()`.
```
gbjk: 🚧
* applet is a loaded term, right?
* capitalisation is a bit misleading, isn't it? 
```suggestion
quickdata fetches data targeted data from an exchange without needing to edit config by using sane defaults.
```

🦍 Optional pedantic nitpick:
I see this moved from QuickSpy to `QuickData`.
It still doesn't feel right to me.
I _kinda_ get why you wanna re-iterate `quick`, but it feels redundant across the package and command name if we get the command name tweaked.

Could you consider command names based on verbs like `peek`, `focus`, `tail`, `spew`, `cat`.
Cute names that are more evocative might include `telescope`, `zoom`.
And it sniffs data, so if you want to get super cute I'd call it corgi or something.
I like naming things 🤷‍♂️  Gives them identity and a life cycle.

cmd/documentation/cmd_templates/quickdata_readme.tmpl
```diff
@@ -0,0 +1,38 @@
...
+
+## Usage
+```go
+go run . --exchange="Binance" --asset="Spot" --pair="BTCUSDT" -datatype=Ticker
```
gbjk: 🚧 Change request:

```suggestion
go run . --exchange="Binance" --asset="Spot" --pair="BTCUSDT" -data=Ticker
```

cmd/documentation/cmd_templates/quickdata_readme.tmpl
```diff
@@ -0,0 +1,38 @@
...
+	- "orderbook", "order_book", "ob", "book":
+	- "kline", "candles", "candle", "ohlc":
+	- "trades", "trade":
+	- "openinterest", "oi":
+	- "fundingrate", "funding":
+	- "accountholdings", "account", "holdings", "balances":
+	- "activeorders", "orders":
+	- "orderexecution", "executionlimits", "limits":
+	- "url", "tradeurl", "trade_url":
+	- "contract":
```
gbjk: 🚧 Change request:
I think you were right to just point the user at the command help.
~And you have haemorrhoids.~ And there's trailing colons here.
```suggestion
```

exchange/quickdata/quickspy_types.go
```diff
@@ -0,0 +1,82 @@
+package quickdata
```
gbjk: 🚧
* Please rename this file, at least to `quickdata_types.go`
* But ideally to just `types.go`
* But really ideally, don't have a types file. IMO, types go next to the methods that use them, and that's more idiomatic. The argument against that for request/response types in exchanges is valid, but outside of that it's cleaner for maintenance and everything if they're next to the code that uses them.

exchange/quickdata/quickspy_types.go
```diff
@@ -0,0 +1,82 @@
...
+	// m is used for concurrent read/write operations
+	m sync.RWMutex
+	// wg is used for synchronizing goroutines
+	wg sync.WaitGroup
+	// Data contains all the market data
+	data *Data
+	// shutdown is a channel used to signal shutdown of goroutines
+	// while a stored context can be cancelled to do the same, we don't pass around the cancel func
+	// the cancel func is the only way to close the context
+	shutdown chan any
```
gbjk: 🚧 It's great these are private ... therefore nearly all of these comments can be removed.

gbjk: 🚧 Change request:

```suggestion
	shutdownC chan any
```

cmd/documentation/exchange_templates/exchange_quickdata_readme.tmpl
```diff
@@ -0,0 +1,119 @@
+{{define "exchange quickdata" -}}
+{{template "header" .}}
+## Overview
+
+The `quickdata` package provides a means to quickly request and receive data for an individual exchange, asset and currency pair. For the times you just really want to get some data without fussing about with configs - setting currency pairs or API keys - or `SetDefaults()` functions.
```
gbjk:
```suggestion
The `quickdata` package provides a means to quickly request and receive data for an individual exchange, asset and currency pair. For the times you just really want to get some data without config and minimal setup.
```

cmd/documentation/exchange_templates/exchange_quickdata_readme.tmpl
```diff
@@ -0,0 +1,119 @@
...
+
+## Features
+
+- Quickly creates an exchange with only the selected asset and currency pair enabled
```
gbjk:
```suggestion
- Creates an exchange with only the selected asset and currency pair enabled
```

cmd/documentation/exchange_templates/exchange_quickdata_readme.tmpl
```diff
@@ -0,0 +1,119 @@
...
+- Supports a range of focus data types allowing for a more tailored approach to data retrieval
```
gbjk:
```suggestion
- Supports a range of focused data types
```

exchange/quickdata/quickspy_types.go
```diff
@@ -0,0 +1,82 @@
...
+// CredentialsKey is a struct that holds credentials and exchange/pair/asset info
+type CredentialsKey struct {
+	// Credentials is optional, if nil public data only
+	Credentials       *account.Credentials  `json:"credentials,omitempty"`
+	ExchangeAssetPair key.ExchangeAssetPair `json:"exchangeAssetPair"`
+}
+
+// QuickData is a struct that holds data on a single exchange pair asset
+// Its purpose is to continuously generate metadata on the market
+// it is write-side oriented vs the comparer
```
gbjk: 🐒 Optional nitpick:
I don't understand this comment. Maybe I will later, delving more into the ecosystem.
But I'm not sure I should have to, and I'm not sure it's helping
```suggestion
// it is write-side oriented vs the comparer
```

exchange/quickdata/quickspy_types.go
```diff
@@ -0,0 +1,82 @@
...
+
+// CredentialsKey is a struct that holds credentials and exchange/pair/asset info
+type CredentialsKey struct {
+	// Credentials is optional, if nil public data only
+	Credentials       *account.Credentials  `json:"credentials,omitempty"`
+	ExchangeAssetPair key.ExchangeAssetPair `json:"exchangeAssetPair"`
+}
+
+// QuickData is a struct that holds data on a single exchange pair asset
+// Its purpose is to continuously generate metadata on the market
```
gbjk: 🐒 Optional nitpick:
I don't think this is metadata, is it?
I think it's just `reporting data from the market`
```suggestion
// Its purpose is to continuously generate metadata on the market
```

exchange/quickdata/quickspy_types.go
```diff
@@ -0,0 +1,82 @@
...
+// it is write-side oriented vs the comparer
+type QuickData struct {
```
gbjk: 🐒 Optional nitpick:
This stutter might be removed. I'm trying to work out the name of this thing.
I'm thinking it's an instance or manager and you'd `quickdata.NewInstance` `NewManager`, `NewRunner`, etc.

exchange/quickdata/quickdata.go
```diff
@@ -0,0 +1,618 @@
...
+	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/orderbook"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/subscription"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/trade"
+	"github.com/thrasher-corp/gocryptotrader/log"
+)
+
+// NewQuickData returns a running quickData if everything passed in is valid
+func NewQuickData(ctx context.Context, k *CredentialsKey, focuses []*FocusData) (*QuickData, error) {
```
gbjk: 🚧 Change request:
`New` should definitely not automatically `Run`.
Can we separate out the run into a public method please?
```suggestion
// NewQuickData returns a running quickData if everything passed in is valid
func NewQuickData(ctx context.Context, k *CredentialsKey, focuses []*FocusData) (*QuickData, error) {
```

gbjk (changes requested):
Great work.
Initial pass, many more to come  🙂

cmd/documentation/cmd_templates/quickdata_readme.tmpl
gloriousCode: I'll redo, but I will not include language like "sane" 😄

exchange/quickdata/quickspy_types.go
gloriousCode: Sometimes yes, sometimes no. I get the idiomatic part of go types, sometimes I find it incredibly messy, but other times I don't care. I'll have another look, but if I hate it because there are too many types required it'll go into `types.go`

exchange/quickdata/quickspy_types.go
gloriousCode: This may have been done by an AI, sorry

exchange/quickdata/quickdata.go
gloriousCode: The only place I'm going to auto `Run` is `QuickestData` as it just returns a chan and not a `QuickData`. I'll update the others

exchange/quickdata/quickspy_types.go
gloriousCode: I'll assume you want `shutdown` renamed `shutdownC` and dismiss the code deletion

exchange/quickdata/quickspy_types.go
gloriousCode: I don't have issue with stutter that matches package name. It looks like you wanted to instead focus on the function names`NewQuickData`, `NewQuickerData` and `NewQuickestData`. I'll consider the stutter with a more appropriate name since you want that changed too

exchange/quickdata/quickspy_types.go
gloriousCode: Even better, I've removed the channel completely and rely on context and banning contexts which can't cancel in some fashion

cmd/quickdata/README.md
```diff
@@ -36,18 +36,22 @@ go run . --exchange="binance" --asset="spot" --pair="btc-usdt" --data="orders" -
...
+| Funding Rate | "fundingrate", "funding" |
+| Account Info | "accountholdings", "account", "holdings", "balances" |
+| Active Orders | "activeorders", "orders" |
+| Order Execution Limits | "orderexecution", "executionlimits", "limits" |
+| URL | "url", "tradeurl", "trade_url" |
+| Contract Info | "contract" |
+
 
 ## Further Reading
 For more  details about QuickData, see the [quickData package documentation](/exchange/quickData/README.md).
```
shazbert: These links seem to be broken 
```suggestion
For more  details about QuickData, see the [quickData package documentation](/exchange/quickdata/README.md).
```

🌴 There seems to be a casing issue across multiple places e.g. `/cmd/quickData`

exchange/quickdata/quickdata.go
```diff
@@ -24,8 +28,45 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/log"
 )
 
+var (
+	errKeyNotFound                    = errors.New("key not found")
+	errNoKey                          = errors.New("no key provided")
```
shazbert: suggestion: Could be a key package error 🤷

cmd/quickdata/app/quickdata_test.go
```diff
@@ -547,3 +523,32 @@ func TestHandleWSData(t *testing.T) {
 	assert.NoError(t, qs.handleWSData(trade.Data{}))
 	assert.NoError(t, qs.handleWSData([]trade.Data{}))
 }
+
+func TestRun(t *testing.T) {
+	t.Parallel()
+	qs := mustQuickData(t, TickerFocusType)
+	require.ErrorIs(t, qs.Run(context.Background()), ErrContextMustBeAbleToFinish)
```
shazbert, line 531: Does t.Context() have timeout already? if so maybe below needs to have a context.Background as well?

gloriousCode: It does have a timeout. It doesn't really matter in the below scenario since it just needs to be able to cancel and I'm manually cancelling it.

exchange/quickdata/README.md
```diff
@@ -0,0 +1,158 @@
...
+
+## Context
+
+All quickData implementations require a context to run. The context must have a deadline, whether from SIGINT, cancel function or deadline. This allows for graceful shutdown of the instance.
+The context can also be used to provide API credentials for authenticated data retrieval methods.
+See Usage section below.
+
+## Usage
+
+There are multiple ways to utilise a quickData. See `/cmd/quickData` for a basic way of establishing a single purpose quickData that subscribes to data and prints it to console.
```
shazbert:
```suggestion
There are multiple ways to utilise a quickData. See `/cmd/quickdata` for a basic way of establishing a single purpose quickData that subscribes to data and prints it to console.
```

shazbert (approved):
Ch ACK. Just one optional minor when other nits come in. 🚀

shazbert (approved):
ch ACK on new stuff

cmd/quickdata/main.go
```diff
@@ -0,0 +1,468 @@
...
+	start := trs[0].Timestamp
+	end := trs[n-1].Timestamp
+	if end.Before(start) {
+		start, end = end, start
+	}
+	var baseVol, quoteVol float64
+	var buys, sells int
+	for i := range trs {
+		baseVol += trs[i].Amount
+		quoteVol += trs[i].Amount * trs[i].Price
```
samuael: Shouldn't rename this `.Amount` variable to something like `.Quantity`? because most of the time, it is used to express the quote amount and here it is the base amount.

cmd/quickdata/main.go
```diff
@@ -0,0 +1,468 @@
...
+		if trs[i].Side&order.Buy == order.Buy || trs[i].Side&order.Bid == order.Bid || trs[i].Side&order.Long == order.Long {
```
samuael: Can use `trs[i].Side.IsLong()` and `trs[i].Side.IsShort()` methods for such checks.

cmd/quickdata/main_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+func TestParseFlags_Success(t *testing.T) {
+	// backup and restore os.Args and flag.CommandLine
+	origArgs := os.Args
+	defer func() { os.Args = origArgs }()
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	os.Args = []string{"quickData", "--exchange", "BiNaNcE", "--asset", "spot", "--pair", "ETH-BTC", "--data", "ticker", "--poll", "10s", "--book-levels", "20", "--json"}
+	cfg := parseFlags()
+	require.Equal(t, "binance", cfg.Exchange)
+	require.Equal(t, asset.Spot, cfg.Asset)
```
samuael: all these below can be `assert`

cmd/quickdata/main_test.go
```diff
@@ -0,0 +1,242 @@
...
+}
+
+func TestParseFlags_AuthRequired(t *testing.T) {
+	origArgs := os.Args
+	defer func() { os.Args = origArgs }()
+	flag.CommandLine = flag.NewFlagSet(os.Args[0], flag.ExitOnError)
+	os.Args = []string{"quickData", "--exchange", "okx", "--asset", "spot", "--pair", "BTC-USDT", "--data", "account", "--apiKey", "k123", "--apiSecret", "s456"}
+	cfg := parseFlags()
+	require.Equal(t, quickdata.AccountHoldingsFocusType, cfg.FocusType)
+	require.NotNil(t, cfg.Credentials)
```
samuael: same here

cmd/quickdata/main.go
```diff
@@ -0,0 +1,468 @@
...
+		return
+	}
+}
+
+// parseFlags parses CLI flags, validates them, applies sensible defaults, and returns appConfig.
+func parseFlags() *appConfig {
+	exch := flag.String("exchange", "", "Exchange name, e.g. binance, okx")
+	assetStr := flag.String("asset", "spot", "Asset type, e.g. spot, futures, perpetualswap")
+	pairStr := flag.String("pair", "BTC-USDT", "Currency pair, e.g. BTC-USDT or ETHUSD")
+	focusStr := flag.String("data", "ticker", "Data type: ticker, orderbook, kline, trades, openinterest, fundingrate, accountholdings, activeorders, orderexecution, url, contract")
```
samuael: shouldn't we have better way of printing this? These flags like `orderbook, kline, ...` has alias that are being used.

cmd/quickdata/main.go
```diff
@@ -0,0 +1,468 @@
...
+		start, end = end, start
+	}
+	var baseVol, quoteVol float64
+	var buys, sells int
+	for i := range trs {
+		baseVol += trs[i].Amount
+		quoteVol += trs[i].Amount * trs[i].Price
+		if trs[i].Side&order.Buy == order.Buy || trs[i].Side&order.Bid == order.Bid || trs[i].Side&order.Long == order.Long {
+			buys++
+		} else if trs[i].Side&order.Sell == order.Sell || trs[i].Side&order.Ask == order.Ask || trs[i].Side&order.Short == order.Short {
```
samuael: `trs[i].Side.IsShort()`

cmd/quickdata/main.go
```diff
@@ -0,0 +1,468 @@
...
+		if kl[i].LowPrice < low {
+			low = kl[i].LowPrice
+		}
+		totalVol += kl[i].Volume
+	}
+	avgVol := totalVol / float64(n)
+	span := end.Sub(start)
+	interval := kl[n-1].Interval
+	if interval == "" && n > 1 {
+		interval = kl[0].Interval
```
samuael: Why could the first element will have the `kline interval` information but not the last `KlineData`? What am I missing here?

gloriousCode: Weird AI things, good spot

cmd/quickdata/main.go
gloriousCode: So much better!

cmd/quickdata/main.go
gloriousCode: That is part of the trade package and I'll say that's out of scope for this part here

cmd/quickdata/main.go
gloriousCode: I've pushed it to include all the aliases, let me know what you think

cmd/quickdata/main_test.go
```diff
@@ -105,8 +105,8 @@ func TestRenderFunctions_NoPanicAndOutput(t *testing.T) {
 	_ = captureOutput(func() { renderKlines(kl) })
 	_ = captureOutput(func() { renderKlines(nil) })
 
-	// account holdings
-	holdings := &account.Holdings{Exchange: "binance", Accounts: []account.SubAccount{{ID: "spot", AssetType: asset.Spot, Currencies: []account.Balance{{Currency: currency.BTC, Total: 1, Free: 0.5, Hold: 0.5}}}}}
+	// accounts. holdings
```
shazbert: revert

shazbert (changes requested):
nit and one misc issue 🚀

cmd/quickdata/main_test.go
gloriousCode: I ham-fisted merging accounts changes soz

shazbert: Is it thinly sliced leg ham?

exchange/quickdata/quickspy_types.go
gbjk: I think we should have an issue about stutter that matches package name 🙂 
It's 
The function names I listed would match the type name.
e.g. type quickdata.Manager
I'll resolve and let you do what you like with it.

cmd/quickdata/app/data.go
```diff
@@ -0,0 +1,509 @@
...
+	if err := common.NilGuard(data); err != nil {
+		return err
+	}
+	focus := q.focuses.GetByFocusType(TickerFocusType)
+	if focus == nil {
+		return fmt.Errorf("%w %q", errKeyNotFound, TickerFocusType)
+	}
+	var payload *ticker.Price
+	switch {
+	case len(data) == 0:
```
samuael, line 132: No need to add this case.

cmd/quickdata/app/data.go
```diff
@@ -0,0 +1,509 @@
...
+	oi, err := q.exch.GetOpenInterest(ctx, key.PairAsset{
+		Base:  q.Key.Pair().Base.Item,
+		Quote: q.Key.Pair().Quote.Item,
+		Asset: q.Key.Asset,
+	})
+	if err != nil {
+		return fmt.Errorf("%s %q %w", q.Key, focus.focusType.String(), err)
+	}
+	if len(oi) != 1 {
+		return nil
```
samuael, line 354: Shouldn't this have an error return or some explanation?

cmd/quickdata/app/quickdata.go
```diff
@@ -0,0 +1,668 @@
...
+		sm.Upsert(focuses[i].focusType, focuses[i])
+	}
+
+	q := &QuickData{
+		Key:                k,
+		dataHandlerChannel: make(chan any, 10),
+		focuses:            sm,
+		data:               &Data{Key: k},
+	}
+	err := q.setupExchange()
```
samuael, line 93:
```suggestion
	if err := q.setupExchange(); err != nil {
```

cmd/quickdata/app/quickdata.go
```diff
@@ -0,0 +1,668 @@
...
+	b.CurrencyPairs.Pairs[q.Key.Asset] = &currency.PairStore{
+		AssetEnabled:  true,
+		RequestFormat: rFmt,
+		ConfigFormat:  cFmt,
+	}
+
+	if err := b.CurrencyPairs.StorePairs(q.Key.Asset, currency.Pairs{cFmtPair}, false); err != nil {
+		return err
+	}
+	if err := b.CurrencyPairs.StorePairs(q.Key.Asset, currency.Pairs{cFmtPair}, true); err != nil {
```
samuael, line 306:
```suggestion
	return b.CurrencyPairs.StorePairs(q.Key.Asset, currency.Pairs{cFmtPair}, true)
```

cmd/quickdata/app/quickdata.go
```diff
@@ -0,0 +1,668 @@
...
+			if f.isOnceOff {
+				return nil
+			}
+			timer.Reset(f.restPollTime)
+		}
+	}
+}
+
+func (q *QuickData) processRESTFocus(ctx context.Context, f *FocusData) error {
+	err := q.handleFocusType(ctx, f.focusType, f)
```
samuael, line 504:
```suggestion
	if err := q.handleFocusType(ctx, f.focusType, f); err != nil {
```

samuael (changes requested):
Just a quick reviews and left few comments.

CONTRIBUTORS
```diff
@@ -20,51 +20,55 @@ junnplus | https://github.com/junnplus
 geseq | https://github.com/geseq
 marcofranssen | https://github.com/marcofranssen
 140am | https://github.com/140am
-TaltaM | https://github.com/TaltaM
+Copilot | https://github.com/apps/copilot-swe-agent
```
gbjk: 🥃 Aside: Wondering if we should be excluding copilot

cmd/documentation/exchange_templates/exchange_quickdata_readme.tmpl
```diff
@@ -0,0 +1,135 @@
+{{define "exchange quickdata" -}}
+{{template "header" .}}
+## Overview
+
+This package provides a means to quickly request and receive data for an individual exchange, asset and currency pair. For the times you just really want to get some data without config and minimal setup.
```
gbjk: 🦍 Optional pedantic nitpick:
```suggestion
This package provides a means to quickly see data for an individual exchange, asset and currency pair. For the times you just really want to get some data without config and minimal setup.
```

cmd/quickdata/main_test.go
```diff
@@ -0,0 +1,242 @@
...
+	"github.com/thrasher-corp/gocryptotrader/exchanges/fundingrate"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/futures"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/orderbook"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/trade"
+)
+
+// captureOutput redirects os.Stdout for the duration of f and returns what was written.
+func captureOutput(f func()) string {
```
gbjk: 🐒 Optional nitpick:
I don't think we should have to do this.
quickdata is hardcoded to use stdout, and that's why we're fangling with it.
I'd expect instead both users and tests to be able to control the filehandle used.
That also pre-supposes that the command runs an App instance.

cmd/quickdata/main.go
```diff
@@ -0,0 +1,481 @@
...
+	UseWebsocket bool
+	PollInterval time.Duration
+	BookLevels   int
+	Credentials  *accounts.Credentials
+	JSONOnly     bool
+}
+
+func main() {
+	outPrintln("Hello! 🌞")
+	defer outPrintln("\nGoodbye! 🌚")
```
gbjk: 🐒 Optional nitpick:
Would want logging/output to be configurable based on an instance and/or context, so that users and tests can decide where things like this are going.

So basically this main.go moves to just running App.Run() or something, and all the functions in main here becomes methods.
That allows for easier testing, configurability and extension.

I probably wouldn't even have that App in here. cmd/quickdata/main.go would be absolutely tiny, and everything would be somewhere else.
I say somewhere else, cos I feel like the right place would be:
` \- app/quickdata/app.go` type `quickdata.App` and `quickdata.NewApp()` with `q.Run()` etc.

It doesn't feel right to have `exchange/quickdata` when it's not really an exchange thing. exchanges don't use it. It uses exchanges. So `app` or `applet` or whatever.

I'm keeping this as a 🐒 because it's not a deal breaker.

cmd/quickdata/main.go
```diff
@@ -0,0 +1,481 @@
...
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/fundingrate"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/futures"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/order"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/orderbook"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/trade"
+)
+
+var enc = json.NewEncoder(os.Stdout)
```
gbjk, line 31: 🐒 Optional nitpick:
output should be configurable on an instance.

cmd/quickdata/main.go
```diff
@@ -0,0 +1,481 @@
...
+	subAccount := flag.String("subAccount", "", "Sub-account (optional)")
+	clientID := flag.String("clientID", "", "Client ID (optional)")
+	otp := flag.String("otp", "", "One-time password (optional)")
+	pemKey := flag.String("pemKey", "", "PEM key (optional)")
+
+	flag.Parse()
+
+	if strings.TrimSpace(*exch) == "" || strings.TrimSpace(*assetStr) == "" || strings.TrimSpace(*pairStr) == "" || strings.TrimSpace(*focusStr) == "" {
+		_, _ = fmt.Fprintln(os.Stderr, "missing required flags: --exchange, --asset, --pair, --focusType")
+		_, _ = fmt.Fprintln(os.Stderr, "please read the readme for more information")
```
gbjk, line 115: * 🤷‍♂️ Clarification please: Did something make you declare the return types for `fmt.FPrint` ?
Just asking cos I don't think we use them normally elsewhere.
* Can this not just be one comma separated thing? Or concat with `\n`
* If you accept the challenge to move to an instance, the would use context or instance for logger.
```suggestion
		fmt.Fprintln(os.Stderr,
			"missing required flags: --exchange, --asset, --pair, --focusType",
			"please read the readme for more information",
			)
```

cmd/quickdata/main.go
```diff
@@ -0,0 +1,481 @@
...
+		flag.Usage()
+		os.Exit(2)
+	}
+
+	// Parse asset
```
gbjk, line 120: 🦍 Optional pedantic nitpick:
These comments aren't adding any useful information. There's a few of them.

cmd/quickdata/main.go
```diff
@@ -0,0 +1,481 @@
...
+	if err != nil {
+		fatalErr(err)
+	}
+
+	pollDur, err := time.ParseDuration(*pollStr)
+	if err != nil {
+		fatalErr(fmt.Errorf("invalid poll duration: %w", err))
+	}
+
+	// Credentials (only applied when supplied or required by focus)
```
gbjk: * 🚧 Change request: Doesn't seem true that applied when supplied?
This suggestion is superseded by the one the gorilla is holding.
```suggestion
	// Credentials (only applied when required by focus)
```
* 🦍 Both the fact that these Credentials are Credentials, but also that they're only required by focus, seem self-evident
```suggestion
```

cmd/quickdata/main.go
```diff
@@ -0,0 +1,481 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case d := <-c:
+			switch {
+			case cfg.JSONOnly:
+				emit(eventEnvelope{Timestamp: time.Now().UTC(), Focus: cfg.FocusType.String(), Data: d})
+			default:
+				clearScreen()
```
gbjk, line 186: 🐒 Optional nitpick:
In the context of what I've been saying about configurability, I think you'd check the term capabilities and decide whether to do things like this.
There's many contexts ( pipes `|` and redirects `>` ) where I definitely wouldn't want you making opinionated choices about manhandling my terminal for me 🙂

cmd/quickdata/main.go
```diff
@@ -0,0 +1,481 @@
...
+				outPrintf("%s%s%s\n", ansiBold, heading, ansiReset)
```
gbjk: 🚧 Change request:
I think we're not using Sprintf, dropped in favour of concating
```suggestion
				outPrintln(ansiBold + heading + ansiReset)
```

exchanges/coinbase/README.md
```diff
@@ -5,7 +5,7 @@
 
 [![Build Status](https://github.com/thrasher-corp/gocryptotrader/actions/workflows/tests.yml/badge.svg?branch=master)](https://github.com/thrasher-corp/gocryptotrader/actions/workflows/tests.yml)
 [![Software License](https://img.shields.io/badge/License-MIT-orange.svg?style=flat-square)](https://github.com/thrasher-corp/gocryptotrader/blob/master/LICENSE)
-[![GoDoc](https://godoc.org/github.com/thrasher-corp/gocryptotrader?status.svg)](https://godoc.org/github.com/thrasher-corp/gocryptotrader/exchanges/Coinbase)
+[![GoDoc](https://godoc.org/github.com/thrasher-corp/gocryptotrader?status.svg)](https://godoc.org/github.com/thrasher-corp/gocryptotrader/exchanges/coinbase)
```
gbjk, line 8: 🚧 🤷‍♂️ Maybe a Change request:
I think this change needs to be made in `cmd/documentation/sub_templates/status.tmpl` and it's caller. Though I went hunting for how sub_templates is loaded, etc and didn't find anything

exchanges/kucoin/kucoin_wrapper.go
```diff
@@ -1921,6 +1921,8 @@ func (e *Exchange) GetFuturesContractDetails(ctx context.Context, item asset.Ite
 }
 
 // GetLatestFundingRates returns the latest funding rates data
+// if no pair is supplied it returns all funding rates, but no predicted rate
+// if an individual pair is supplied it returns the predicted rate as well
```
gbjk, line 1925:
```suggestion
// If no pair is supplied it returns all funding rates.
// Only returns a predicted rate with a single pair.
```

exchanges/coinbase/README.md
gloriousCode: This output likely would have only changed as a result of generating docs, I have no stake in if coinbase is capitalised

gbjk: Oh, then this is completely the other way round!
Resolving 👍

cmd/quickdata/main.go
gloriousCode: I agree with the part about moving things from /exchange/. Rather than a new clutter of a folder with the context of people disliking it taking up space in GCT, would `cmd` still suffice? 
`/cmd/quickdata/` where `/exchange/quickdata` moves to `/cmd/quickdata/engine`. I move various rendering stuff to `/cmd/quickdata/engine/renderer.go` to keep `/cmd/quickdata/main.go` minimal

I'm more wary of log configurability, because its increasing complexity to have a config or something and `main.go` is mostly an example. I may be overcomplicating your request. What does log configurability look like to you?

cmd/documentation/exchange_templates/exchange_quickdata_readme.tmpl
gloriousCode: Fair to hate on request and receive, but "see" is too casual. I could accept it for the main.go readme, since you are seeing the output, but you are receiving proper structs that you can do things with. So I'll change to `quickly encromulate data` or something else

cmd/quickdata/main.go
gloriousCode: I like being explicit about ignoring return types, but I can remove those and be more standard

cmd/quickdata/main.go
gloriousCode: I don't like the idea of `renderer.go` anymore. Same feelings apply to only wanting `main.go` to be a tiny file. I feel like it is closer to a helpful demo application of retrieving data quickly, it doesn't really need log customisation or a more standard way of rendering that data (ala renderer.go) because its much better to actually get and utilise the data than using cmd to show it.

cmd/documentation/exchange_templates/exchange_quickdata_readme.tmpl
gbjk: 🦍 is easily placated - resolving is fine.
But `grab, `review`, `stream`.

Thought actually .. just `request` seems fine.

cmd/quickdata/main.go
gbjk: I think we're agreed that exchange is the wrong place.
I don't like the idea of `/cmd/` being more than light shims, but it's up to you.

My review overall comment wasn't really threadable, so I'll repeat it here, cos it still feels right to me:

> I think I'd prefer for quickdata itself to live in a side ecosystem to GCT, in it's own repo, gct-quickdata. It could still depend on GCT, and everything would be just as fast, but it would reduce GCT core size for people who don't use it

Is that viable?

gloriousCode: I'll tag in @thrasher here because @shazbert also requested it be in a  separate repo and we're discussing the difficulty of where to place the work

thrasher-: Can put it into a separate repo and have references to it in the README

cmd/gctcli/futures.go
```diff
@@ -582,17 +572,7 @@ func getManagedPosition(c *cli.Context) error {
...
-		includePredictedRate = c.Bool("includepredictedrate")
-	} else if c.Args().Get(6) != "" {
-		includePredictedRate, err = strconv.ParseBool(c.Args().Get(6))
-		if err != nil {
-			return err
-		}
-	}
-
-	err = futures.CheckFundingRatePrerequisites(getFundingData, includePredictedRate, includeFundingEntries)
+	err = futures.CheckFundingRatePrerequisites(getFundingData, includeFundingEntries)
```
samuael, line 575: `if err := ...`

cmd/quickdata/app/store_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestNewStore(t *testing.T) {
+	t.Parallel()
+	fs := NewFocusStore()
+	require.NotNil(t, fs)
+	require.NotNil(t, fs.s)
```
samuael, line 14: if you don't have a comment with `"  ..  must .. "` comment, just use `assert`

cmd/quickdata/app/quickdata_test.go
```diff
@@ -0,0 +1,567 @@
...
+}
+
+func TestCheckRateLimits(t *testing.T) {
+	t.Parallel()
+	q := mustQuickData(t, TickerFocusType)
+	e, err := engine.NewSupportedExchangeByName(q.Key.Exchange)
+	require.NoError(t, err)
+	b := e.GetBase()
+	err = q.setupExchangeDefaults(e, b)
+	require.NoError(t, err)
```
samuael, line 179: for these assertions, I believe putting them inside the `assert`/`require` call

cmd/quickdata/main_test.go
```diff
@@ -0,0 +1,144 @@
...
+	r, w, _ := os.Pipe()
+	os.Stdout = w
+	enc = json.NewEncoder(os.Stdout)
+	ctx, cancel := context.WithCancel(context.Background())
+	ch := make(chan any, 1)
+	cfg := &appConfig{Exchange: "testexch", Asset: asset.Spot, Pair: pair, FocusType: app.TradesFocusType, JSONOnly: false, BookLevels: 10}
+	ch <- []trade.Data{{Price: 101, Amount: 0.5, Timestamp: time.Now(), Side: order.Buy}}
+	done := make(chan error, 1)
+	go func() { done <- streamData(ctx, ch, cfg) }()
+	time.Sleep(80 * time.Millisecond)
```
samuael, line 127: Why this specific delay, any explanation?

PR #2039, opened 2025-09-08
testing/log/experimental: make logging quieter

gloriousCode:
# PR Description

There are some logs during testing that are unhelpful and don't actually add anything to the process. It wastes time and makes my eyes sad.
 Take this test:
```go
func TestLogs(t *testing.T) {
	t.Error("noo")
}
```

output `master`:
```
2025/09/08 15:17:16 Running upgrade to config version 1
2025/09/08 15:17:16 Running upgrade to config version 2
2025/09/08 15:17:16 Running upgrade to config version 3
2025/09/08 15:17:16 Running upgrade to config version 4
2025/09/08 15:17:16 Running upgrade to config version 5
2025/09/08 15:17:16 Running upgrade to config version 6
2025/09/08 15:17:16 Running upgrade to config version 7
2025/09/08 15:17:16 Running upgrade to config version 8
2025/09/08 15:17:16 Version management finished
Exchange Bitflyer Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange Bitstamp Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange COINUT Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange CoinbasePro Websocket orderbook buffer limit value not set, defaulting to 5.
exchange Deribit authenticated API support disabled due to default/empty APIKey/Secret/ClientID values
Exchange Deribit Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange EXMO Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange GateIO: [options] Removing enabled pair(s) [BTC_USDT-20230217-28000-P] from enabled pairs list, as it isn't located in the available pairs list.
exchange Kucoin authenticated API support disabled due to default/empty APIKey/Secret/ClientID values
Exchange Kucoin Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange LBank Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange Poloniex Websocket orderbook buffer limit value not set, defaulting to 5.
Exchange Yobit Websocket orderbook buffer limit value not set, defaulting to 5.
Currency file update duration invalid, defaulting to 168h0m0s
Currency foreign exchange update duration invalid, defaulting to 1m0s
Okx cannot set pairs manager, pairs already set
=== RUN   TestLogs
    okx_test.go:188: noo
--- FAIL: TestLogs (0.00s)
```

My branch:
```
=== RUN   TestLogs
    okx_test.go:188: noo
--- FAIL: TestLogs (0.00s)

FAIL

```

Now extrapolate that across all our test failures! aaaaah!
Perhaps you find value in the output like the gist, but it contains no data on the actual test failure.

edit: I'm happy to add build tags or something here, with default being opt out. So someone can opt into the yuck

### DIFF
--- a/internal/testing/exchange/exchange.go
+++ b/internal/testing/exchange/exchange.go
@@ -30,9 +30,7 @@ import (
 func Setup(e exchange.IBotExchange) error {
 	// quiet test setup logs as they are unhelpful
 	log.SetOutput(io.Discard)
-	defer func() {
-		log.SetOutput(os.Stdout)
-	}()
+	defer log.SetOutput(os.Stdout)
 
 	cfg := &config.Config{}
 	root, err := testutils.RootPathFromCWD()


exchanges/exchange.go
```diff
@@ -671,28 +671,30 @@ func (b *Base) UpdatePairs(incoming currency.Pairs, a asset.Item, enabled, force
...
-					"%s Updating %s pairs [%v] - Added: %s.\n",
-					b.Name,
-					updateType,
-					strings.ToUpper(a.String()),
-					diff.New)
-			}
-			if len(diff.Remove) > 0 {
+		if b.Verbose {
+			if force {
 				log.Debugf(log.ExchangeSys,
```
shazbert, line 676: suggestion: you can one liner these logs if you want to

internal/testing/exchange/exchange.go
```diff
@@ -27,8 +28,13 @@ import (
 
 // Setup takes an empty exchange instance and loads config for it from testdata/configtest and connects a NewTestWebsocket
 func Setup(e exchange.IBotExchange) error {
-	cfg := &config.Config{}
+	// quiet test setup logs as they are unhelpful
+	log.SetOutput(io.Discard)
+	defer func() {
+		log.SetOutput(os.Stdout)
+	}()
```
shazbert:
```suggestion
	defer log.SetOutput(os.Stdout)
```

exchanges/exchange.go
```diff
@@ -671,28 +671,30 @@ func (b *Base) UpdatePairs(incoming currency.Pairs, a asset.Item, enabled, force
...
-				log.Debugf(log.ExchangeSys,
-					"%s Updating %s pairs [%v] - Added: %s.\n",
-					b.Name,
-					updateType,
-					strings.ToUpper(a.String()),
-					diff.New)
-			}
-			if len(diff.Remove) > 0 {
+		if b.Verbose {
+			if force {
```
shazbert, line 675: 🥤 Force seems weird and seems like it can be rm'd but definitely not for this PR.

shazbert (changes requested):
In the course of my exhaustive and meticulous examination of the submitted pull request, I have painstakingly identified a singular matter of utmost concern that demands immediate attention, coupled with an additional, carefully considered recommendation for enhancement, which I humbly propose for your thorough consideration.

[less verbose] one thing and a suggestion.

exchanges/exchange.go
shazbert: #2043

exchanges/exchange.go
shazbert: #2043 done here

exchanges/exchange.go
thrasher-: Was used when we needed to force update pairs when we did config upgrades per exchange but that code has all been removed now, so should  be safe to remove

shazbert (approved):
Thanks 🫶.

thrasher-:
@codex please review

thrasher- (changes requested):
Aside from the config version upgrade, the logger is actually adding value with its output. Several issues it has highlighted:

Deribit and Kucoin have auth set to true when it should be false
We're calling SetDefaults twice https://github.com/thrasher-corp/gocryptotrader/blob/master/internal/testing/exchange/exchange.go#L41-L47 which sets an already set pair manager again
We can set the ws buffer limit and other values it's complaining about (most exchanges have this already)

For the config version I'd recommend upgrading our test config since we have full test coverage as it stands, or another approach is to have an isolated config which just tests that separate to our main test config (like we did with engine).

Then all these issues will be resolved without muting the logger

gloriousCode:
Can't believe you don't want my absolutely brute force approach! 😄 Was thinking of having it behind build tags, but I'll look to modify the config

Issue #2026, opened 2025-09-04
Implement OpenTelemetry (OTEL) for Metrics Aggregation in GoCryptoTrader


shazbert:
## Description
Add OpenTelemetry (OTEL) support across the GoCryptoTrader library to enable metrics collection for Prometheus/Grafana dashboards, enhancing observability of core functionalities.

## Scope
- Integrate OTEL for metrics collection across the library (can be split into multiple PRs).
- Target components:
  - **gRPC**: Instrument gRPC server/client calls (e.g., latency, request count, errors).
  - **WebSocket**: Track manager status, connection states, throughput, and request metrics.
  - **REST API**: Monitor all HTTP requests (e.g., latency, status codes, request volume).
- Ensure metrics are exported to Prometheus for visualization in Grafana.

## Design Considerations
- Use `context.Context` for OTEL instrumentation to ensure seamless propagation of tracing context without modifying existing function signatures.
- Implement full context propagation from the engine to all API boundaries (gRPC, WebSocket, REST).
- Adhere to OTEL best practices for naming conventions and metric types (e.g., counters, histograms for latency, gauges for connection states).
- Minimize performance overhead by using efficient OTEL collectors and batching.
- Ensure backward compatibility with existing logging and monitoring setups.

## Additional Suggestions
- Add support for OTEL tracing alongside metrics for end-to-end request visibility.
- Provide configuration options to enable/disable OTEL or specific metric types.
- Include documentation for setting up Prometheus/Grafana with GoCryptoTrader.
- Add example Grafana dashboards for common metrics (e.g., request latency, error rates).
- Write unit tests to validate OTEL metric output and context propagation.

## Acceptance Criteria
- Metrics are correctly emitted and visible in Prometheus/Grafana.
- Context propagation works across all targeted components without breaking existing functionality.
- Performance impact is minimal (<5% overhead).
- Code adheres to GoCryptoTrader’s style guide and passes CI checks.

## References
- [OpenTelemetry Go SDK](https://opentelemetry.io/docs/languages/go/)
- [Prometheus Go Client](https://prometheus.io/docs/instrumenting/clientlibs/#go)

Issue #1951, opened 2025-06-19
orderbook: peak processing spike


shazbert:
Investigate why orderbook peak processing time spikes to ~20x average while loading either incremental updates or snapshots, consistently noticing on external ops while using library. This external instance does not interact with engine. This is only subscribed to orderbooks only. 

see: https://github.com/thrasher-corp/gocryptotrader/pull/1734 for metric usage wrt websocket/manager

```
2025-06-19 21:36:00 DBG Connection: wss://fx-ws.gateio.ws/v4/ws/usdt Operations/Second: 2022.35, Avg Processing/Operation: 8.489µs, Errors: 0 Peak: 180.351µs Cause: {"channel":"futures.order_book_update","event":"update","result":{"t":1750368932681,"U":17646757515,... subLogger=WEBSOCKET
2025-06-19 21:36:00 DBG Connection: wss://api.gateio.ws/ws/v4/ Operations/Second: 1231.43, Avg Processing/Operation: 10.974µs, Errors: 0 Peak: 181.911µs Cause: {"time":1750368933,"time_ms":1750368933522,"channel":"spot.order_book_update","event":"update","resu... subLogger=WEBSOCKET
```

gloriousCode:
[This ](https://github.com/thrasher-corp/gocryptotrader/pull/1734#discussion_r1868464912)could still be helpful to track it easier. Would give you the time, the json and could easily have breakpoints for if current > 10x average. That focusses on if JSON is the annoyance, say its a massive update json.

Next thing I'd suggest is if you don't use it externally, remove mux.publish from order book processing. That's if unnecessary atomic checks are slowing things down at scale on one field `atomic.LoadInt32(&m.d.subscriberCount)`

I'd also check the potential chokepoint of the data handler and go back to this for `LoadSnapshot` etc
```go
        select {
		case w.dataHandler <- holder.ob:
		default:
			log.Errorf(log.WebsocketMgr, "help")
			w.dataHandler <- holder.ob
	}
```

What does the `orderbook.Depth` `lastUpdated`, `updatePushedAt` and `insertedAt` say? If you put in similar reporting metrics at the order book you can at least see if your order book processing is lagging. 

Finally, on the order book metrics, you could also have a "receivedAt" to help distinguish between external time data and your own

edit: Also, if its something only rare, then I can't imagine its worth the chase

thrasher-:
[...]

PR #2066, opened 2025-09-22
websocket: Initial refactor for websocket handling to support context propagation

shazbert:
* Adds package `message` for the websocket data handler so there is no circular dependencies
    - This design stops blocking and returns an error if the buffer is full when using method `Send`
    - `Send` takes in context that can be consumed for future metrics/trace. e.g. external consumer using signal (orderbook change) can get metrics all the way back from the websocket reader routine. 
    
* This removes `observeData` method on websocket manager. This was just moving data to another channel just to indicate blockage which had a bit of overhead, the above design eliminates the need for that mechanism.  

* Tries to keep things contained in websocket only a few context.TODO's and these will be opened up in a future PR:
    - exchange wrappers
    - orderbook
    - ticker
    - trades etc.

This complements issue #2026 for a context based OTEL solution
Also complements issue #2068 

This also will help trace along with profiling #1951 as it seems to be just a garbage collection spike, optionally can track GC cycles on full trace for that signal. 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -75,6 +75,7 @@ var (
 	ErrGettingField              = errors.New("error getting field")
 	ErrSettingField              = errors.New("error setting field")
 	ErrParsingWSField            = errors.New("error parsing websocket field")
+	ErrMalformedData             = errors.New("malformed data")
 )
 
 var (
--- a/currency/pair.go
+++ b/currency/pair.go
@@ -7,8 +7,12 @@ import (
 	"unicode"
 )
 
+// Public errors
 var (
-	errCannotCreatePair       = errors.New("cannot create currency pair")
+	ErrCreatingPair = errors.New("error creating currency pair")
+)
+
+var (
 	errDelimiterNotFound      = errors.New("delimiter not found")
 	errDelimiterCannotBeEmpty = errors.New("delimiter cannot be empty")
 )
@@ -70,7 +74,7 @@ func NewPairWithDelimiter(base, quote, delimiter string) Pair {
 // with or without delimiter
 func NewPairFromString(currencyPair string) (Pair, error) {
 	if len(currencyPair) < 3 {
-		return EMPTYPAIR, fmt.Errorf("%w from %s string too short to be a currency pair", errCannotCreatePair, currencyPair)
+		return EMPTYPAIR, fmt.Errorf("%w from %s string too short to be a currency pair", ErrCreatingPair, currencyPair)
 	}
 
 	for x := range currencyPair {
--- a/currency/pair_methods.go
+++ b/currency/pair_methods.go
@@ -61,7 +61,7 @@ func (p *Pair) UnmarshalJSON(d []byte) error {
 	// incorrectly converted to DUS-KUSDT, ELKRW (Bithumb) which will convert
 	// converted to ELK-RW and HTUSDT (Lbank) which will be incorrectly
 	// converted to HTU-SDT.
-	return fmt.Errorf("%w from %s cannot ensure pair is in correct format, please use exchange method MatchSymbolWithAvailablePairs", errCannotCreatePair, pair)
+	return fmt.Errorf("%w from %s cannot ensure pair is in correct format, please use exchange method MatchSymbolWithAvailablePairs", ErrCreatingPair, pair)
 }
 
 // MarshalJSON conforms type to the marshaler interface
--- a/currency/pair_test.go
+++ b/currency/pair_test.go
@@ -54,7 +54,7 @@ func TestPairUnmarshalJSON(t *testing.T) {
 	assert.Equal(t, "usd", p.Quote.String(), "Quote should be correct")
 	assert.Equal(t, "_", p.Delimiter, "Delimiter should be correct")
 
-	assert.ErrorIs(t, p.UnmarshalJSON([]byte(`"btcusd"`)), errCannotCreatePair, "UnmarshalJSON with no delimiter should error")
+	assert.ErrorIs(t, p.UnmarshalJSON([]byte(`"btcusd"`)), ErrCreatingPair, "UnmarshalJSON with no delimiter should error")
 
 	assert.NoError(t, p.UnmarshalJSON([]byte(`""`)), "UnmarshalJSON should not error on empty value")
 	assert.Equal(t, EMPTYPAIR, p, "UnmarshalJSON empty value should give EMPTYPAIR")
--- a/currency/pairs_test.go
+++ b/currency/pairs_test.go
@@ -54,7 +54,7 @@ func TestPairsFromString(t *testing.T) {
 	_, err := NewPairsFromString("", "")
 	assert.ErrorIs(t, err, errNoDelimiter)
 	_, err = NewPairsFromString("", ",")
-	assert.ErrorIs(t, err, errCannotCreatePair)
+	assert.ErrorIs(t, err, ErrCreatingPair)
 
 	pairs, err := NewPairsFromString("ALGO-AUD,BAT-AUD,BCH-AUD,BSV-AUD,BTC-AUD,COMP-AUD,ENJ-AUD,ETC-AUD,ETH-AUD,ETH-BTC,GNT-AUD,LINK-AUD,LTC-AUD,LTC-BTC,MCAU-AUD,OMG-AUD,POWR-AUD,UNI-AUD,USDT-AUD,XLM-AUD,XRP-AUD,XRP-BTC", ",")
 	require.NoError(t, err)
--- a/exchanges/deribit/deribit_types.go
+++ b/exchanges/deribit/deribit_types.go
@@ -42,7 +42,6 @@ var (
 	errInvalidID                           = errors.New("invalid id")
 	errInvalidMarginModel                  = errors.New("missing margin model")
 	errInvalidEmailAddress                 = errors.New("invalid email address")
-	errMalformedData                       = errors.New("malformed data")
 	errWebsocketConnectionNotAuthenticated = errors.New("websocket connection is not authenticated")
 	errResolutionNotSet                    = errors.New("resolution not set")
 	errInvalidDestinationID                = errors.New("invalid destination id")
--- a/exchanges/deribit/deribit_websocket.go
+++ b/exchanges/deribit/deribit_websocket.go
@@ -331,7 +331,7 @@ func (e *Exchange) wsSendHeartbeat(ctx context.Context) {
 
 func (e *Exchange) processUserOrders(ctx context.Context, respRaw []byte, channels []string) error {
 	if len(channels) != 4 && len(channels) != 5 {
-		return fmt.Errorf("%w, expected format 'user.orders.{instrument_name}.raw, user.orders.{instrument_name}.{interval}, user.orders.{kind}.{currency}.raw, or user.orders.{kind}.{currency}.{interval}', but found %s", errMalformedData, strings.Join(channels, "."))
+		return fmt.Errorf("%w, expected format 'user.orders.{instrument_name}.raw, user.orders.{instrument_name}.{interval}, user.orders.{kind}.{currency}.raw, or user.orders.{kind}.{currency}.{interval}', but found %s", common.ErrMalformedData, strings.Join(channels, "."))
 	}
 	var response wsResponse
 	orderData := []WsOrder{}
@@ -379,7 +379,7 @@ func (e *Exchange) processUserOrders(ctx context.Context, respRaw []byte, channe
 
 func (e *Exchange) processUserOrderChanges(ctx context.Context, respRaw []byte, channels []string) error {
 	if len(channels) < 4 || len(channels) > 5 {
-		return fmt.Errorf("%w, expected format 'trades.{instrument_name}.{interval} or trades.{kind}.{currency}.{interval}', but found %s", errMalformedData, strings.Join(channels, "."))
+		return fmt.Errorf("%w, expected format 'trades.{instrument_name}.{interval} or trades.{kind}.{currency}.{interval}', but found %s", common.ErrMalformedData, strings.Join(channels, "."))
 	}
 	var response wsResponse
 	changeData := &wsChanges{}
@@ -489,7 +489,7 @@ func (e *Exchange) processTrades(ctx context.Context, respRaw []byte, channels [
 	}
 
 	if len(channels) < 3 || len(channels) > 5 {
-		return fmt.Errorf("%w, expected format 'trades.{instrument_name}.{interval} or trades.{kind}.{currency}.{interval}', but found %s", errMalformedData, strings.Join(channels, "."))
+		return fmt.Errorf("%w, expected format 'trades.{instrument_name}.{interval} or trades.{kind}.{currency}.{interval}', but found %s", common.ErrMalformedData, strings.Join(channels, "."))
 	}
 	var response wsResponse
 	var tradeList []wsTrade
@@ -535,7 +535,7 @@ func (e *Exchange) processTrades(ctx context.Context, respRaw []byte, channels [
 
 func (e *Exchange) processIncrementalTicker(ctx context.Context, respRaw []byte, channels []string) error {
 	if len(channels) != 2 {
-		return fmt.Errorf("%w, expected format 'incremental_ticker.{instrument_name}', but found %s", errMalformedData, strings.Join(channels, "."))
+		return fmt.Errorf("%w, expected format 'incremental_ticker.{instrument_name}', but found %s", common.ErrMalformedData, strings.Join(channels, "."))
 	}
 	a, cp, err := getAssetPairByInstrument(channels[1])
 	if err != nil {
@@ -566,7 +566,7 @@ func (e *Exchange) processIncrementalTicker(ctx context.Context, respRaw []byte,
 
 func (e *Exchange) processInstrumentTicker(ctx context.Context, respRaw []byte, channels []string) error {
 	if len(channels) != 3 {
-		return fmt.Errorf("%w, expected format 'ticker.{instrument_name}.{interval}', but found %s", errMalformedData, strings.Join(channels, "."))
+		return fmt.Errorf("%w, expected format 'ticker.{instrument_name}.{interval}', but found %s", common.ErrMalformedData, strings.Join(channels, "."))
 	}
 	return e.processTicker(ctx, respRaw, channels)
 }
@@ -619,7 +619,7 @@ func (e *Exchange) processData(ctx context.Context, respRaw []byte, result any) 
 
 func (e *Exchange) processCandleChart(ctx context.Context, respRaw []byte, channels []string) error {
 	if len(channels) != 4 {
-		return fmt.Errorf("%w, expected format 'chart.trades.{instrument_name}.{resolution}', but found %s", errMalformedData, strings.Join(channels, "."))
+		return fmt.Errorf("%w, expected format 'chart.trades.{instrument_name}.{resolution}', but found %s", common.ErrInvalidResponse, strings.Join(channels, "."))
 	}
 	a, cp, err := getAssetPairByInstrument(channels[2])
 	if err != nil {
@@ -661,15 +661,15 @@ func (e *Exchange) processOrderbook(respRaw []byte, channels []string) error {
 		asks := make(orderbook.Levels, 0, len(orderbookData.Asks))
 		for x := range orderbookData.Asks {
 			if len(orderbookData.Asks[x]) != 3 {
-				return errMalformedData
+				return common.ErrMalformedData
 			}
 			price, okay := orderbookData.Asks[x][1].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid orderbook price", errMalformedData)
+				return fmt.Errorf("%w, invalid orderbook price", common.ErrMalformedData)
 			}
 			amount, okay := orderbookData.Asks[x][2].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid amount", errMalformedData)
+				return fmt.Errorf("%w, invalid amount", common.ErrMalformedData)
 			}
 			asks = append(asks, orderbook.Level{
 				Price:  price,
@@ -679,17 +679,17 @@ func (e *Exchange) processOrderbook(respRaw []byte, channels []string) error {
 		bids := make(orderbook.Levels, 0, len(orderbookData.Bids))
 		for x := range orderbookData.Bids {
 			if len(orderbookData.Bids[x]) != 3 {
-				return errMalformedData
+				return common.ErrMalformedData
 			}
 			price, okay := orderbookData.Bids[x][1].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid orderbook price", errMalformedData)
+				return fmt.Errorf("%w, invalid orderbook price", common.ErrMalformedData)
 			} else if price == 0.0 {
 				continue
 			}
 			amount, okay := orderbookData.Bids[x][2].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid amount", errMalformedData)
+				return fmt.Errorf("%w, invalid amount", common.ErrMalformedData)
 			}
 			bids = append(bids, orderbook.Level{
 				Price:  price,
@@ -730,17 +730,17 @@ func (e *Exchange) processOrderbook(respRaw []byte, channels []string) error {
 		asks := make(orderbook.Levels, 0, len(orderbookData.Asks))
 		for x := range orderbookData.Asks {
 			if len(orderbookData.Asks[x]) != 2 {
-				return errMalformedData
+				return common.ErrMalformedData
 			}
 			price, okay := orderbookData.Asks[x][0].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid orderbook price", errMalformedData)
+				return fmt.Errorf("%w, invalid orderbook price", common.ErrMalformedData)
 			} else if price == 0 {
 				continue
 			}
 			amount, okay := orderbookData.Asks[x][1].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid amount", errMalformedData)
+				return fmt.Errorf("%w, invalid amount", common.ErrMalformedData)
 			}
 			asks = append(asks, orderbook.Level{
 				Price:  price,
@@ -750,17 +750,17 @@ func (e *Exchange) processOrderbook(respRaw []byte, channels []string) error {
 		bids := make([]orderbook.Level, 0, len(orderbookData.Bids))
 		for x := range orderbookData.Bids {
 			if len(orderbookData.Bids[x]) != 2 {
-				return errMalformedData
+				return common.ErrMalformedData
 			}
 			price, okay := orderbookData.Bids[x][0].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid orderbook price", errMalformedData)
+				return fmt.Errorf("%w, invalid orderbook price", common.ErrMalformedData)
 			} else if price == 0 {
 				continue
 			}
 			amount, okay := orderbookData.Bids[x][1].(float64)
 			if !okay {
-				return fmt.Errorf("%w, invalid amount", errMalformedData)
+				return fmt.Errorf("%w, invalid amount", common.ErrMalformedData)
 			}
 			bids = append(bids, orderbook.Level{
 				Price:  price,
--- a/exchanges/gateio/gateio.go
+++ b/exchanges/gateio/gateio.go
@@ -195,6 +195,7 @@ type Exchange struct {
 
 	messageIDSeq  common.Counter
 	wsOBUpdateMgr *wsOBUpdateManager
+	wsOBResubMgr  *wsOBResubManager
 }
 
 // ***************************************** SubAccounts ********************************
--- a/exchanges/gateio/gateio_types.go
+++ b/exchanges/gateio/gateio_types.go
@@ -2092,6 +2092,17 @@ type WsOrderbookUpdate struct {
 	Asks          orderbook.LevelsArrayPriceAmount `json:"a"`
 }
 
+// WsOrderbookUpdateWithSnapshot represents websocket orderbook update push data
+type WsOrderbookUpdateWithSnapshot struct {
+	UpdateTime    types.Time                       `json:"t"`
+	Full          bool                             `json:"full"`
+	Channel       string                           `json:"s"`
+	FirstUpdateID int64                            `json:"U"`
+	LastUpdateID  int64                            `json:"u"`
+	Bids          orderbook.LevelsArrayPriceAmount `json:"b"`
+	Asks          orderbook.LevelsArrayPriceAmount `json:"a"`
+}
+
 // WsOrderbookSnapshot represents a websocket orderbook snapshot push data
 type WsOrderbookSnapshot struct {
 	UpdateTime   types.Time                       `json:"t"`
--- a/exchanges/gateio/gateio_websocket.go
+++ b/exchanges/gateio/gateio_websocket.go
@@ -46,6 +46,7 @@ const (
 	spotOrderbookTickerChannel = "spot.book_ticker"       // Best bid or ask price
 	spotOrderbookUpdateChannel = "spot.order_book_update" // Changed order book levels
 	spotOrderbookChannel       = "spot.order_book"        // Limited-Level Full Order Book Snapshot
+	spotOrderbookV2            = "spot.obu"
 	spotOrdersChannel          = "spot.orders"
 	spotUserTradesChannel      = "spot.usertrades"
 	spotBalancesChannel        = "spot.balances"
@@ -64,6 +65,7 @@ var defaultSubscriptions = subscription.List{
 	{Enabled: true, Channel: subscription.OrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds},
 	{Enabled: false, Channel: spotOrderbookTickerChannel, Asset: asset.Spot, Interval: kline.TenMilliseconds, Levels: 1},
 	{Enabled: false, Channel: spotOrderbookChannel, Asset: asset.Spot, Interval: kline.HundredMilliseconds, Levels: 100},
+	{Enabled: false, Channel: spotOrderbookV2, Asset: asset.Spot, Levels: 50},
 	{Enabled: true, Channel: spotBalancesChannel, Asset: asset.Spot, Authenticated: true},
 	{Enabled: true, Channel: crossMarginBalanceChannel, Asset: asset.CrossMargin, Authenticated: true},
 	{Enabled: true, Channel: marginBalancesChannel, Asset: asset.Margin, Authenticated: true},
@@ -191,6 +193,8 @@ func (e *Exchange) WsHandleSpotData(ctx context.Context, conn websocket.Connecti
 		return e.processOrderbookUpdate(ctx, push.Result, push.Time)
 	case spotOrderbookChannel:
 		return e.processOrderbookSnapshot(push.Result, push.Time)
+	case spotOrderbookV2:
+		return e.processOrderbookUpdateWithSnapshot(ctx, conn, push.Result, push.Time, asset.Spot)
 	case spotOrdersChannel:
 		return e.processSpotOrders(ctx, respRaw)
 	case spotUserTradesChannel:
@@ -322,7 +326,7 @@ func (e *Exchange) processCandlestick(ctx context.Context, incoming []byte) erro
 	}
 	icp := strings.Split(data.NameOfSubscription, currency.UnderscoreDelimiter)
 	if len(icp) < 3 {
-		return errors.New("malformed candlestick websocket push data")
+		return fmt.Errorf("%w: candlestick websocket", common.ErrMalformedData)
 	}
 	currencyPair, err := currency.NewPairFromString(strings.Join(icp[1:], currency.UnderscoreDelimiter))
 	if err != nil {
@@ -404,6 +408,59 @@ func (e *Exchange) processOrderbookSnapshot(incoming []byte, lastPushed time.Tim
 		}
 	}
 	return nil
+}
+
+func (e *Exchange) processOrderbookUpdateWithSnapshot(ctx context.Context, conn websocket.Connection, incoming []byte, lastPushed time.Time, a asset.Item) error {
+	var data WsOrderbookUpdateWithSnapshot
+	if err := json.Unmarshal(incoming, &data); err != nil {
+		return err
+	}
+
+	channelParts := strings.Split(data.Channel, ".")
+	if len(channelParts) < 3 {
+		return fmt.Errorf("%w: %q", common.ErrMalformedData, data.Channel)
+	}
+
+	pair, err := currency.NewPairFromString(channelParts[1])
+	if err != nil {
+		return err
+	}
+
+	if data.Full {
+		if err := e.Websocket.Orderbook.LoadSnapshot(&orderbook.Book{
+			Exchange:     e.Name,
+			Pair:         pair,
+			Asset:        a,
+			LastUpdated:  data.UpdateTime.Time(),
+			LastPushed:   lastPushed,
+			LastUpdateID: data.LastUpdateID,
+			Bids:         data.Bids.Levels(),
+			Asks:         data.Asks.Levels(),
+		}); err != nil {
+			return err
+		}
+		e.wsOBResubMgr.CompletedResubscribe(pair, a)
+		return nil
+	}
+
+	if e.wsOBResubMgr.IsResubscribing(pair, a) {
+		return nil // Drop incremental updates; waiting for a fresh snapshot
+	}
+
+	lastUpdateID, err := e.Websocket.Orderbook.LastUpdateID(pair, a)
+	if err != nil || lastUpdateID+1 != data.FirstUpdateID {
+		return common.AppendError(err, e.wsOBResubMgr.Resubscribe(ctx, e, conn, data.Channel, pair, a))
+	}
+	return e.Websocket.Orderbook.Update(&orderbook.Update{
+		Pair:       pair,
+		Asset:      a,
+		UpdateTime: data.UpdateTime.Time(),
+		LastPushed: lastPushed,
+		UpdateID:   data.LastUpdateID,
+		Bids:       data.Bids.Levels(),
+		Asks:       data.Asks.Levels(),
+		AllowEmpty: true,
+	})
 }
 
 func (e *Exchange) processSpotOrders(ctx context.Context, data []byte) error {
@@ -599,11 +656,12 @@ func (e *Exchange) GetSubscriptionTemplate(_ *subscription.Subscription) (*templ
 	return template.New("master.tmpl").
 		Funcs(sprig.FuncMap()).
 		Funcs(template.FuncMap{
-			"channelName":         channelName,
-			"singleSymbolChannel": singleSymbolChannel,
-			"orderbookInterval":   orderbookChannelInterval,
-			"candlesInterval":     candlesChannelInterval,
-			"levels":              channelLevels,
+			"channelName":             channelName,
+			"singleSymbolChannel":     singleSymbolChannel,
+			"orderbookInterval":       orderbookChannelInterval,
+			"candlesInterval":         candlesChannelInterval,
+			"levels":                  channelLevels,
+			"compactOrderbookPayload": isCompactOrderbookPayload,
 		}).Parse(subTplText)
 }
 
@@ -691,7 +749,7 @@ func channelName(s *subscription.Subscription) string {
 // singleSymbolChannel returns if the channel should be fanned out into single symbol requests
 func singleSymbolChannel(name string) bool {
 	switch name {
-	case spotCandlesticksChannel, spotOrderbookUpdateChannel, spotOrderbookChannel:
+	case spotCandlesticksChannel, spotOrderbookUpdateChannel, spotOrderbookChannel, spotOrderbookV2:
 		return true
 	}
 	return false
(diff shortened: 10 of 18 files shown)


thrasher-:
@codex please review

exchanges/btse/btse_websocket.go
```diff
@@ -189,27 +190,16 @@ func (e *Exchange) wsHandleData(_ context.Context, respRaw []byte) error {
...
-				e.Websocket.DataHandler <- order.ClassificationError{
-					Exchange: e.Name,
-					OrderID:  notification.Data[i].OrderID,
-					Err:      err,
-				}
+				return err
 			}
-			oSide, err = order.StringToOrderSide(notification.Data[i].OrderMode)
+
+			oSide, err = order.StringToOrderSide(strings.ReplaceAll(notification.Data[i].OrderMode, "MODE_", ""))
```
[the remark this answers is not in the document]
shazbert: This was just to fix a test, the other implementations don't seem to have the same handling which leads me to believe the test websocket payload might be wrong but this is all out of scope.

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -547,14 +516,13 @@ func (e *Exchange) handleWSEvent(respRaw []byte) error {
...
-				log.Errorf(log.ExchangeSys, "%s %s 'code': %s from message: %s", e.Name, common.ErrParsingWSField, err, respRaw)
-			}
-			return fmt.Errorf("WS auth subscription error; Status: %s Error Code: %d", status, errCode)
+			return e.Websocket.DataHandler.Send(ctx, glob)
 		}
+		errCode, err := jsonparser.GetInt(respRaw, "code")
+		if err != nil {
+			log.Errorf(log.ExchangeSys, "%s %s 'code': %s from message: %s", e.Name, common.ErrParsingWSField, err, respRaw)
+		}
+		return fmt.Errorf("WS auth subscription error; Status: %s Error Code: %d", status, errCode)
```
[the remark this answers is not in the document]
shazbert: status return might not ever be "ERROR" and this doesn't take in an unhandled status event; as this is out of scope I will defer this.

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -244,27 +214,27 @@ func (e *Exchange) wsHandleData(_ context.Context, respRaw []byte) error {
...
 		case wsNotification:
-			return e.handleWSNotification(d, respRaw)
+			return e.handleWSNotification(ctx, d, respRaw)
 		case wsOrderSnapshot:
 			if snapBundle, ok := d[2].([]any); ok && len(snapBundle) > 0 {
 				if _, ok := snapBundle[0].([]any); ok {
 					for i := range snapBundle {
 						if positionData, ok := snapBundle[i].([]any); ok {
-							e.wsHandleOrder(positionData)
+							return e.wsHandleOrder(ctx, positionData)
```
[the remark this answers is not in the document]
shazbert: This I will do.

exchanges/btcmarkets/btcmarkets_test.go
```diff
@@ -612,6 +614,10 @@ func TestWsHeartbeats(t *testing.T) {
 }
 
 func TestWsOrders(t *testing.T) {
+	ctx := account.DeployCredentialsToContext(t.Context(), &account.Credentials{
+		Key:    "testkey",
+		Secret: base64.StdEncoding.EncodeToString([]byte("testsecret")),
+	})
```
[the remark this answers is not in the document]
shazbert: Unless others want me to do this I will leave this open.

exchange/websocket/buffer/buffer.go
```diff
@@ -107,7 +111,7 @@
 
 	// Publish all state changes, disregarding verbosity or sync requirements.
 	holder.ob.Publish()
-	o.dataHandler <- holder.ob
+	o.dataHandler.Send(ctx, holder.ob)
```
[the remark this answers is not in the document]
shazbert: 🔥

thrasher-:
@codex please review

exchange/message/message.go
```diff
@@ -0,0 +1,46 @@
...
+type Payload struct {
+	Ctx  context.Context //nolint:containedctx // context needed for tracing/metrics
+	Data any
+}
+
+// NewRelay creates a new Relay instance with a specified buffer size
+func NewRelay(buffer uint) *Relay {
+	if buffer == 0 {
+		panic("buffer size must be greater than 0")
+	}
```
gloriousCode: Why must it be a buffered channel?

exchange/websocket/buffer/buffer.go
```diff
@@ -24,7 +26,7 @@ var (
 )
 
 // Setup sets private variables
-func (o *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandler chan<- any) error {
+func (o *Orderbook) Setup(exchangeConfig *config.Exchange, c *Config, dataHandler *message.Relay) error {
 	if exchangeConfig == nil { // exchange config fields are checked in websocket package prior to calling this, so further checks are not needed
```
gloriousCode:
```suggestion
	if err := common.NilGuard(exchangeConfig, c, dataHandler); err != nil {
```

exchange/websocket/buffer/buffer.go
```diff
@@ -107,8 +110,7 @@ func (o *Orderbook) Update(u *orderbook.Update) error {
 
 	// Publish all state changes, disregarding verbosity or sync requirements.
 	holder.ob.Publish()
-	o.dataHandler <- holder.ob
-	return nil
+	return o.dataHandler.Send(ctx, holder.ob)
```
gloriousCode:
```suggestion
	return o.dataHandler.Send(context.TODO(), holder.ob)
```
not fussed

exchange/websocket/manager.go
```diff
@@ -407,13 +406,17 @@ func (m *Manager) getConnectionFromSetup(c *ConnectionSetup) *connection {
...
 	defer m.m.Unlock()
-	return m.connect()
+	return m.connect(ctx)
 }
 
-func (m *Manager) connect() error {
+func (m *Manager) connect(ctx context.Context) error {
+	if err := common.NilGuard(m.DataHandler); err != nil {
+		return err
+	}
```
gloriousCode: Seems like an odd place for your style of code. Seems like a failure on the setup level and not the connect level.
If the buffer gets reset on shutdown/connect, it still seems to be an odd place for this. I think it should be elsewhere, not here

exchange/websocket/buffer/buffer.go
```diff
@@ -2,13 +2,15 @@ package buffer
```
gloriousCode, line 2: I understand that this is an initial refactor, but it also seems odd that nothing in the manager or buffer responds to context.Done
edit: This seems to be much more about telemetry than using context with `ctx.Done()`

exchange/websocket/manager.go
```diff
@@ -1010,24 +988,22 @@ func (m *Manager) observeConnection(t *time.Timer) (exit bool) {
...
 // the connection is shutdown and will be reconnected by the connectionMonitor routine.
-func (m *Manager) monitorTraffic() func() bool {
-	timer := time.NewTimer(m.trafficTimeout)
-	return func() bool { return m.observeTraffic(timer) }
+func (m *Manager) monitorTraffic(context.Context) func() bool {
+	return func() bool { return m.observeTraffic(m.trafficTimeout) }
 }
 
-func (m *Manager) observeTraffic(t *time.Timer) bool {
+func (m *Manager) observeTraffic(timeout time.Duration) bool {
```
gloriousCode: pass ctx and ctx.Done handling in `observeTraffic`?

exchanges/binance/binance_websocket.go
```diff
@@ -586,8 +568,6 @@ func (e *Exchange) manageSubs(ctx context.Context, op string, subs subscription.
 
 	if err != nil {
 		err = fmt.Errorf("%w; Channels: %s", err, strings.Join(subs.QualifiedChannels(), ", "))
-		e.Websocket.DataHandler <- err
```
gloriousCode: ?

exchange/message/message.go
```diff
@@ -0,0 +1,46 @@
...
+
+// Relay defines a channel relay for messages
+type Relay struct {
+	C    <-chan Payload
+	comm chan Payload
+}
+
+// Payload represents a relayed message with a context
+type Payload struct {
+	Ctx  context.Context //nolint:containedctx // context needed for tracing/metrics
```
gloriousCode: I think the interesting thing about what you're doing is that its not to do with say handling `ctx.Done()` across large planes to handle spinning up and shutting down say an entire exchange - It is instead about tracing metrics. In such a scenario, I have consulted our AI overlord

> You generally should not store a full context in a long‑lived or widely shared struct; that is why the linter warns. For propagating tracing/metrics over a channel, extract only the needed identifiers (trace/span/baggage) into the payload, then reconstruct a fresh context on the receiving side. This avoids leaking cancellations/timeouts and satisfies the linter intent.

>Below replaces the embedded context with trace metadata (example using OpenTelemetry). The sender extracts IDs, the receiver reconstructs a span context.

```go
// Go
package message

import (
	"context"
	"errors"
	"fmt"

	"go.opentelemetry.io/otel/trace"
)

var errChannelBufferFull = errors.New("channel buffer is full")

type Relay struct {
	C    <-chan Payload
	comm chan Payload
}

// Payload carries data plus minimal trace metadata instead of a full context.
type Payload struct {
	Data    any
	TraceID string
	SpanID  string
	// Add baggage or metrics labels as needed, e.g. map[string]string
}

func NewRelay(buffer uint) *Relay {
	if buffer == 0 {
		panic("buffer size must be greater than 0")
	}
	comm := make(chan Payload, buffer)
	return &Relay{comm: comm, C: comm}
}

// Send extracts trace IDs from ctx and enqueues a lightweight payload.
func (r *Relay) Send(ctx context.Context, data any) error {
	sc := trace.SpanFromContext(ctx).SpanContext()
	traceID := ""
	spanID := ""
	if sc.IsValid() {
		traceID = sc.TraceID().String()
		spanID = sc.SpanID().String()
	}
	select {
	case r.comm <- Payload{
		Data:    data,
		TraceID: traceID,
		SpanID:  spanID,
	}:
		return nil
	default:
		return fmt.Errorf("%w: failed to relay <%T>", errChannelBufferFull, data)
	}
}

// ReconstructContext creates a new context carrying the span context.
// This produces a new root (no parent linkage) but preserves IDs for metrics.
func (p Payload) ReconstructContext() context.Context {
	if p.TraceID == "" || p.SpanID == "" {
		return context.Background()
	}
	tid, err1 := trace.TraceIDFromHex(p.TraceID)
	sid, err2 := trace.SpanIDFromHex(p.SpanID)
	if err1 != nil || err2 != nil {
		return context.Background()
	}
	sc := trace.NewSpanContext(trace.SpanContextConfig{
		TraceID: tid,
		SpanID:  sid,
		// Adjust flags if needed (e.g. Remote: true, TraceFlags: trace.FlagsSampled)
	})
	return trace.ContextWithSpanContext(context.Background(), sc)
}

func (r *Relay) Close() {
	close(r.comm)
}
```

> If fuller propagation (parent/links, baggage) is required, serialize those into the payload (e.g. a map) using your propagator’s Inject/Extract functions. This keeps contexts ephemeral and avoids embedding them directly. 

I do find its answers more appropriate versus storing potentially expired contexts and not handling `Done()` anywhere.

I'm not using OpenTelemetry yet, but I [read this](https://opentelemetry.io/docs/specs/otel/trace/api/#context-interaction) and it seems to coincide with the AI overlords view. Specifically this:
>    Get the currently active span from the implicit context. This is equivalent to getting the implicit context, then extracting the Span from the context.
    Set the currently active span into a new context, and make that the implicit context. This is equivalent to combining the current implicit context’s values with the Span to create a new context, which is then made the current implicit context.

I will leave this here for you to ponder

final edit: Additional thing to ponder https://go.dev/blog/context-and-structs#storing-context-in-structs-leads-to-confusion

gloriousCode (commented):
I appreciate context usage, but the nolint leaves me uncertain. I will leave the AI output for you to consider

exchange/message/message.go
gbjk: I actually agree with this.
I often embed contexts and share them, but never through channels.
I hadn't thought about that last time round.

exchange/websocket/manager.go
shazbert: Yeah this was a bit of a weird one, I think it was a few tests where calling connect without calling `NewManager()` and it was silently deadlocking. This protects against the chance of that happening again.

exchange/websocket/buffer/buffer.go
shazbert: Yeah this is purely telemetry

exchange/message/message.go
shazbert: Happy to do this but in another PR just to reduce review toil and focus on what is currently being done. We don't consume it in the library yet anyway, externally I will be, and I think I will opt for the map suggestion so we can scale it up from the context. @gloriousCode is that okay? Can we add this to the telemetry issue? #2026

gloriousCode: This seems like a fundamental part of the PR. I'll get back to you when I can look into things again

engine/rpcserver.go
```diff
@@ -5884,3 +5884,14 @@ func (s *RPCServer) GetCurrencyTradeURL(ctx context.Context, r *gctrpc.GetCurren
...
 }
+
+// rpcContextToLongLivedSession converts a short-lived incoming context to a long-lived outgoing context, this is due
+// to the incoming context being cancelled when the RPC call completes.
+func (s *RPCServer) rpcContextToLongLivedSession(ctx context.Context) context.Context {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		md = metadata.New(nil) // Fallback to empty metadata
+	}
+	// TODO: Capture RPC server context for cancellations and use it here
```
gloriousCode: This is a very specific edgecase I'm not sure needs to be a todo, especially in the case of storing a context in a struct. Feels like trying to rush against the request. 
eg
- Send SubmitOrder via client
- Server is alive, so accepts request
- Server sends SubmitOrder to exchange
- You desperately attempt to cancel the server to stop that FOMO 120x Fart/USD order
- You hope that the context cancelFunc hits before the exchange submits the order

The TODO can still stay, hopefully its for a different reason

exchange/message/message.go
gloriousCode: **TL;DR** Alrighty, I think this can be deferred to another PR now. 

- The exported Context isn’t currently used within GCT, so my earlier points aren’t blocking right now. That said, GK, AI, OTEL, Go devs, golangci-lint and myself recommend against the current approach

- Regarding the comment about “reducing review toil and focusing on what’s being done” — this PR is what’s introducing the pattern, so it’s not really review toil, it’s part of ensuring the change is introduced better from the start.

exchanges/binance/binance_websocket.go
```diff
@@ -133,39 +133,39 @@ func (e *Exchange) setupOrderbookManager(ctx context.Context) {
...
 		resp := e.Websocket.Conn.ReadMessage()
 		if resp.Raw == nil {
 			return
 		}
-		err := e.wsHandleData(resp.Raw)
-		if err != nil {
-			e.Websocket.DataHandler <- err
+		if err := e.wsHandleData(ctx, resp.Raw); err != nil {
+			if errSend := e.Websocket.DataHandler.Send(ctx, err); errSend != nil {
+				log.Errorf(log.WebsocketMgr, "%s %s: %s %s", e.Name, e.Websocket.Conn.GetURL(), errSend, err)
```
gloriousCode, line 162: I do appreciate the requirement to have a buffer chan with err versus a holdup

exchanges/binanceus/binanceus_websocket.go
```diff
@@ -100,43 +100,41 @@ func (e *Exchange) WsConnect() error {
...
 		resp := e.Websocket.Conn.ReadMessage()
 		if resp.Raw == nil {
 			return
 		}
-		err := e.wsHandleData(resp.Raw)
-		if err != nil {
-			e.Websocket.DataHandler <- err
+		if err := e.wsHandleData(ctx, resp.Raw); err != nil {
+			if errSend := e.Websocket.DataHandler.Send(ctx, err); errSend != nil {
+				log.Errorf(log.WebsocketMgr, "%s: %s %s", e.Name, errSend, err)
```
gloriousCode: Consistency with including URL plz. This is the only instance without

docs/ADD_NEW_EXCHANGE.md
```diff
@@ -849,7 +849,7 @@ func (e *Exchange) wsReadData() {
 
             err := e.wsHandleData(resp.Raw)
             if err != nil {
-                e.Websocket.DataHandler <- err
+                e.Websocket.DataHandler.Send(ctx, err)
```
gloriousCode: There are two more instances of DataHandler usage that needs to be updated

engine/rpcserver.go
```diff
@@ -5884,3 +5884,14 @@ func (s *RPCServer) GetCurrencyTradeURL(ctx context.Context, r *gctrpc.GetCurren
 		Url: url,
 	}, nil
 }
+
+// rpcContextToLongLivedSession converts a short-lived incoming context to a long-lived outgoing context, this is due
+// to the incoming context being cancelled when the RPC call completes.
+func (s *RPCServer) rpcContextToLongLivedSession(ctx context.Context) context.Context {
```
gloriousCode: Add test

```go

func TestRPCContextToLongLivedSession(t *testing.T) {
	t.Parallel()
	s := &RPCServer{}
	t.Run("preserve metadata", func(t *testing.T) {
	    t.Parallel()
		incomingMD := metadata.Pairs(
			"authorization", "Bearer token",
			"x-id", "123",
		)
		inCtx, cancel := context.WithCancel(metadata.NewIncomingContext(context.Background(), incomingMD))
		defer cancel()

		newCtx := s.rpcContextToLongLivedSession(inCtx)
		require.NotNil(t, newCtx)

		outMD, ok := metadata.FromOutgoingContext(newCtx)
		require.True(t, ok)
		require.Len(t, outMD, len(incomingMD))
		assert.ElementsMatch(t, incomingMD.Get("authorization"), outMD.Get("authorization"))
		assert.ElementsMatch(t, incomingMD.Get("x-id"), outMD.Get("x-id"))

		cancel()
		assert.Eventually(t, func() bool {
			return inCtx.Err() != nil
		}, time.Second, 10*time.Millisecond)
		assert.Eventually(t, func() bool {
			return newCtx.Err() == nil
		}, 500*time.Millisecond, 20*time.Millisecond)
	})

	t.Run("no metadata", func(t *testing.T) {
	    t.Parallel()
		inCtx, cancel := context.WithCancel(context.Background())
		defer cancel()

		newCtx := s.rpcContextToLongLivedSession(inCtx)
		require.NotNil(t, newCtx)

		outMD, ok := metadata.FromOutgoingContext(newCtx)
		require.True(t, ok)
		assert.Len(t, outMD, 0)

		cancel()
		assert.Eventually(t, func() bool { return inCtx.Err() != nil }, time.Second, 10*time.Millisecond)
		assert.Eventually(t, func() bool { return newCtx.Err() == nil }, 500*time.Millisecond, 20*time.Millisecond)
	})
}
```

engine/rpcserver.go
shazbert: It was more to do if we have a context generated in engine that we could possibly capture when we do things like SetProxyAddress and it restarts and flushes the websocket system then atleast if we ctrl+c it this will die as well buuut in saying that, this is not how we should be using context and there should be Shutdown funcs being called. So I will remove it for now. If we ever get a problem in future we can always just add that in 🤷.

As for the submit order, it's probably going to be outbound before you even think `oh f*&k` 😆

docs/ADD_NEW_EXCHANGE.md
shazbert: I found one other than this one, is there another one?

engine/rpcserver.go
shazbert: Thank you kindly

exchanges/binance/binance_websocket.go
shazbert: 🕊️

docs/ADD_NEW_EXCHANGE.md
gloriousCode: `e.Websocket.DataHandler <- resultData.FillsData`

`e.Websocket.DataHandler <- &ticker.Price{`

Is my ctrl f. Soz should have been specific

shazbert: Thank you kindly: https://github.com/thrasher-corp/gocryptotrader/pull/2066/commits/fc7df14d3cb77bc912ecf52b65f5b639852e23db

exchange/message/message.go
gbjk: I'm on the other side. I've done exactly this and regretted it, and pulled it out, exactly because of confused cancellation paths.
Also because if we're using context to store values, then the receiving context's values are equally important.

Instead of overlord's suggestion of having a bespoke Payload specific implementation of context serialising for transit, I'd say that it's worth having `common.FreezeCtx `and `common.ThawCtx` that provides a struct that's a stateless value store safe for sending down channels, essentially between contexts. We can also have `common.MergeCtx` to take the values from a `ThawCtx` and add them to an existing `Context`.
The point here is that we can ensure all values in a context can be caught and sent.
Normally I use structs for context keys, and whilst it's a bit of extra scaffolding to also have them in a slice for checking, it's not too painful.

Soz. Thought about this a bit before replying.

exchange/stream/relay.go
```diff
@@ -0,0 +1,47 @@
...
+	"context"
+	"errors"
+	"fmt"
+)
+
+var errChannelBufferFull = errors.New("channel buffer is full")
+
+// Relay defines a channel relay for messages
+type Relay struct {
+	C    <-chan Payload
```
gbjk, line 15: 🥃 🗻  Just a thought: 
This feels like we're introducing a duplicate of `websocketDataReceiver` 's fan out to datahandlers pattern.
I found myself wanting to say "Okay. I've subscribed to X. Now I just want to get that data" a few times. And I have to Register a Datahandler and discard what I'm not interested in.
Worse, I'm subscribing to all exchanges when I do that, so I'm filtering a lot.

It'd be good if we could merge that into here, and allow targetted relaying.

This is out of scope, though

gbjk (changes requested):
I'd like to avoid passing the context down the channel, as per [this thread](https://github.com/thrasher-corp/gocryptotrader/pull/2066#discussion_r2501403057)

exchange/message/message.go
shazbert: Alright fellows, hopefully this addresses everything.

common/common.go
```diff
@@ -695,3 +696,64 @@ func SetIfZero[T comparable](p *T, def T) bool {
...
+	context.Context //nolint:containedctx // Using context.WithValue will nest contexts and cause lookup latency
+	frozen          FrozenContext
+}
+
+func (m *mergeCtx) Value(key any) any {
+	if val, ok := m.frozen.values[key]; ok {
+		return val
+	}
+	return m.Context.Value(key)
+}
```
gbjk: 🚧 Change request:
I think this code was maybe a reaction to the `fatcontext` linter.
The idiomatic way to set keys on contexts is with `context.WithValue`.
It's fine that they're nested, that's the way go contexts were designed.
The linter's author's use-case was to not call `context.WithContext` N+ times in a for loop and keep overwriting the same key.
They highlighted this in [the blog post](https://gabnotes.org/posts/context-induced-performance-bottleneck-in-go).
They should have been using a shadow var, and that's the solution they expect to be used for this linter warning.
None of that is relevant to us.
We're looping over a list of values and going to return a context with all of them.
Calling `WithValue` is the right thing to do.
```suggestion
// MergeCtx adds the frozen values to an existing context
func MergeCtx(ctx context.Context, fc FrozenContext) (context.Context, error) {
	for k, v := range fc.values {
		if ctx.Value(k) != nil {
			return nil, fmt.Errorf("%w: %q", errDuplicateContextKey, k)
		}
		ctx = context.WithValue(ctx, k, v) //nolint:fatcontext // false-positive; linter catches non-shadow contexts, but we return this ctx
	}
	return ctx, nil
}
```

gbjk (changes requested):
Great work, thanks.

Mini-review.

I'm also thinking that the justification for `exchange/message` being a package might be solvable by using the consumer defined interface pattern. I'll look into that and report back

common/common.go
shazbert: I benched it and it indeed nests contexts within each other even with different key values, so the implementation I have done can stay:

```go
// 100000000	        12.17 ns/op	       0 B/op	       0 allocs/op (with mergeCtx)
//
//	38971	     26070 ns/op	       0 B/op	       0 allocs/op (using context.WithValue)
func BenchmarkXxx(b *testing.B) {
	m := FrozenContext{
		values: make(map[any]any),
	}
	for i := range 10000 {
		m.values[i] = i
	}

	ctx, err := ThawCtx(m)
	require.NoError(b, err)

	b.ResetTimer()

	for b.Loop() {
		if val := ctx.Value(5000); val == nil {
			b.Fatal("value not found")
		}
	}
}
```
Can you please check yourself to see if I have done something silly @gbjk

gbjk: @shazbert Whoa. Yeah, you've benched something nobody should ever do, because you've stored thousands of keys.
I wasn't saying nesting contexts is a bad thing.
I was saying nesting thousands of contexts like the author did is a bad thing.

If you're in a situation where you're storing a large number of keys then storing a single struct would make sense.
But we're not. If we store up to 5 keys, then accessing them shouldn't take longer.
See [this bench](https://gist.githubusercontent.com/gbjk/3269ff29ff1bdbe078b8ecf937903784/raw/f50bfbb05922d0749051ac164b4f0cb7c1bc89a8/bench.go)
Accessing these keys is about the same time (faster for 4, slower for 10).

**However, plot twist**
Freezing the context takes N* longer, though.
And our use-case freezes the context a lot more than normal, since we're doing it for every message we want to relay, right?
So If we're storing even 2 keys, all the time, then we should use your method.

If you agree, just resolve.

common/common.go
```diff
@@ -695,3 +696,64 @@ func SetIfZero[T comparable](p *T, def T) bool {
...
+	defer contextKeysMu.Unlock()
+	if !slices.Contains(contextKeys, key) {
+		contextKeys = append(contextKeys, key)
+	}
+}
+
+// FrozenContext holds captured context values
+type FrozenContext struct {
+	values map[any]any
+}
```
gbjk: 🚧 Change request:
Very minor but might as well.
This doesn't need to be a struct.
See this patch: gbjk@91321316f
You can apply this patch with `git apply` or `git am`:
```
curl -s https://github.com/gbjk/gocryptotrader/commit/91321316f.patch | git am
```

common/common.go
shazbert: Leave it up to @thrasher- @gloriousCode they can resolve or whatnot 

Id opt to stay as is so that if somebody has an ungodly amount (I don't know why) of things for context at least its collapsed at this point 🤷.

Also I can do what metadata does and have a single key lookup for the map on a context and use context.WithValue() so we don't need to do the nolint on a struct as an option, if we want to keep the storage of the map?

common/common.go
shazbert: Half did it, and waiting for further direction from other people: https://github.com/thrasher-corp/gocryptotrader/pull/2066/commits/e42ae00e90ad9f6f063338a6c70eb6d2e94e1e7a

shazbert: ❤️

common/common.go
gbjk: I think it's fine as is. No need to use a real context.WithValue.
I'd lean towards saying: Use this context thing from the start, though.

Make it an app context that implements context.Context, backed by context.Context for lifecycle management but with a different value store permanently attached to `WithValue` and `Value` for speed.

Because request cycle values _are_ a thing, a context is the right place to store them, and native context.* is bad at it.

The nice thing about that is that nested context for any *other* reason, like cancellation, would still un-nest to find your context's value store. And you can still delegate to the underlying context Value like you do.

common/common.go
```diff
@@ -695,3 +696,62 @@ func SetIfZero[T comparable](p *T, def T) bool {
...
+			return nil, fmt.Errorf("%w: %q", errDuplicateContextKey, k)
+		}
+	}
+	return &mergeCtx{Context: ctx, frozen: fc}, nil
+}
+
+// mergeCtx is a context that merges values from a frozen context and a parent context.
+type mergeCtx struct {
+	context.Context //nolint:containedctx // Using context.WithValue will nest contexts and cause lookup latency
+	frozen          FrozenContext
```
gbjk: 🚧 Change request:
* This comment has nothing to do with the linter issue
* As we showed, it's the setting that's more painful, not the lookup, so we should write that. Otherwise people will read it, say "no it doesn't and revert".
```suggestion
	context.Context //nolint:containedctx // mergeCtx implements context.Context
	frozen          FrozenContext
```

common/common.go
```diff
@@ -695,3 +696,62 @@ func SetIfZero[T comparable](p *T, def T) bool {
...
+func MergeCtx(ctx context.Context, fc FrozenContext) (context.Context, error) {
+	for k := range fc {
+		if ctx.Value(k) != nil {
+			return nil, fmt.Errorf("%w: %q", errDuplicateContextKey, k)
+		}
+	}
+	return &mergeCtx{Context: ctx, frozen: fc}, nil
+}
+
+// mergeCtx is a context that merges values from a frozen context and a parent context.
```
gbjk:
```suggestion
// mergeCtx is a context that merges values from a frozen context and a parent context.
// frozen values are stored in FrozenContext instead of nested context.WithValue because of the performance of calling WithValue N+ times on messages being frozen
```

shazbert:
> Two small change requests that are unimportant.
> 
> https://github.com/gbjk/gocryptotrader/commit/3506efa04.patch
> 
> I was looking at the option of what we discussed, and saw advantages to moving to contextStore type. The issue is that it becomes mutable. I haven't finished this, but it's 90% of the way there. I'll just throw this ball to you and you can use it, drop it, whatever.
> 
> The main realisation I had was that we're paying the cost of WithValue already N+ times with the current design. Yes, we avoid it at Thaw. But we paid it originally putting values into the context on the other side before Freeze.

Yeah I like it. All the values that are put in by the library are captured across this boundary by freezecontext so you don't need to set the required values you want in common, it will auto scale. Anything else are dropped and it keeps its speed. 

Will see if @thrasher- or @gloriousCode wants this for this PR. Then I will drop it in. Or we can open another PR for it after.

common/common.go
gloriousCode: Thank you both for looking into solutions so thoroughly! I was only after a `message` package specific helper to grab OTELs metrics. I find the current committed solution addresses the problem I raised. I hope its not too complex on the `Thawed` side, Shazbert.

I think there's room for some improvements to be made, but design nicety improvements should be their own PR. I think the only place to try and improve upon now is efficiency. This looks like it could have an impact since it occurs for every websocket message. I'll look into that today and leave review comments

common/common.go
```diff
@@ -695,3 +696,63 @@ func SetIfZero[T comparable](p *T, def T) bool {
...
+	return MergeCtx(context.Background(), fc)
+}
+
+// MergeCtx adds the frozen values to an existing context
+func MergeCtx(ctx context.Context, fc FrozenContext) (context.Context, error) {
+	for k := range fc {
+		if ctx.Value(k) != nil {
+			return nil, fmt.Errorf("%w: %q", errDuplicateContextKey, k)
+		}
+	}
```
gloriousCode: I think this `for` loop should go. At scale, traversing a map and a context just incase there's existing values is likely to be a bottleneck. 

You are specifically asking to apply values from your frozen context, so I think it is okay just to apply them.

You may wish make it a new func entirely where you have `ApplyCtx` which doesn't error and doesn't loop instead and use `MergeCtx` for debugging
Here is my work:

# BenchmarkMergeContext Results

| Benchmark | ns/op | B/op | allocs/op |
| --- | --- | --- | --- |
| BenchmarkMergeContext/original\_sequential | 121.9 | 24 | 1 |
| BenchmarkMergeContext/original\_parallel | 20.83 | 24 | 1 |
| BenchmarkMergeContext/noerr\_sequential | 22.53 | 24 | 1 |
| BenchmarkMergeContext/noerr\_parallel | 0.1033 | 0 | 0 |

common/common.go
```diff
@@ -695,3 +696,63 @@ func SetIfZero[T comparable](p *T, def T) bool {
...
+}
+
+// FrozenContext holds captured context values
+type FrozenContext map[any]any
+
+// FreezeCtx captures values from the context for registered keys
+func FreezeCtx(ctx context.Context) FrozenContext {
+	contextKeysMu.RLock()
+	defer contextKeysMu.RUnlock()
```
gloriousCode, line 720: I have trialed the following designs:
1) Shorten RLock time via `slices.Clone` (clone_keys)
2) Use a `sync.Pool` for `FrozenContext` allocations (pooling)
3) Pass in `keys` instead of having a set of global mutexes via signature `func FreezeCtxKeys3[K comparable](ctx context.Context, keys []K) FrozenContext` (no-globals)
4) 1A combination of pooling and local keys (no-globals+pooling)

The winner is 2, but it does require that you `Release` each context, which I can understand not wanting to do. Otherwise, your design works fine 🌈 

[You can look at my work here](https://gist.github.com/gloriousCode/e9c3de183fc95791638f69d798712bdb)

I have the following results:
# BenchmarkFreezeCtx Results

| Benchmark | ns/op | B/op | allocs/op |
| --- | --- | --- | --- |
| BenchmarkFreezeCtx/original\_sequential | 477.6 | 336 | 2 |
| BenchmarkFreezeCtx/original\_parallel | 256.5 | 336 | 2 |
| BenchmarkFreezeCtx/clone\_keys\_sequential | 549.1 | 464 | 3 |
| BenchmarkFreezeCtx/clone\_keys\_parallel | 300.5 | 464 | 3 |
| BenchmarkFreezeCtx/pooling\_sequential | 441.6 | 0 | 0 |
| BenchmarkFreezeCtx/pooling\_parallel | 157.3 | 0 | 0 |
| BenchmarkFreezeCtx/no-globals\_sequential | 846.3 | 592 | 18 |
| BenchmarkFreezeCtx/no-globals\_parallel | 334.6 | 592 | 18 |
| BenchmarkFreezeCtx/no-globals+pooling\_sequential | 846.6 | 256 | 16 |
| BenchmarkFreezeCtx/no-globals+pooling\_parallel | 287.1 | 256 | 16 |

common/common.go
```diff
@@ -695,3 +696,63 @@ func SetIfZero[T comparable](p *T, def T) bool {
...
+		contextKeys = append(contextKeys, key)
+	}
+}
+
+// FrozenContext holds captured context values
+type FrozenContext map[any]any
+
+// FreezeCtx captures values from the context for registered keys
+func FreezeCtx(ctx context.Context) FrozenContext {
+	contextKeysMu.RLock()
```
gloriousCode:
```go
	contextKeysMu.RLock()
	keys := slices.Clone(contextKeys)
	contextKeysMu.RUnlock()
```

common/common.go
shazbert: Will remove the check

common/common.go
shazbert: Pooling will make it harder for the consumer so I might defer this. Did you see what GK did?: https://github.com/gbjk/gocryptotrader/commit/3506efa04.patch

common/common.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2066/commits/239d160a65190e8b9f888e29881c5c8aa3599ed6

common/common.go
gloriousCode: I tried so hard and I got so far but in the end the patch couldn't apply and my brain couldn't parse it as text.

I also didn't really expect `sync.Pool` to be practical

thrasher-:
> > Two small change requests that are unimportant.
> > https://github.com/gbjk/gocryptotrader/commit/3506efa04.patch
> > I was looking at the option of what we discussed, and saw advantages to moving to contextStore type. The issue is that it becomes mutable. I haven't finished this, but it's 90% of the way there. I'll just throw this ball to you and you can use it, drop it, whatever.
> > The main realisation I had was that we're paying the cost of WithValue already N+ times with the current design. Yes, we avoid it at Thaw. But we paid it originally putting values into the context on the other side before Freeze.
> 
> Yeah I like it. All the values that are put in by the library are captured across this boundary by freezecontext so you don't need to set the required values you want in common, it will auto scale. Anything else are dropped and it keeps its speed.
> 
> Will see if @thrasher- or @gloriousCode wants this for this PR. Then I will drop it in. Or we can open another PR for it after.

Happy as is, it works. Can always adjust it once the patch is more fleshed out and in a different PR

engine/rpcserver.go
```diff
@@ -5884,3 +5884,13 @@ func (s *RPCServer) GetCurrencyTradeURL(ctx context.Context, r *gctrpc.GetCurren
...
```
gbjk: 🐒 Optional nitpick:
Okay; I kinda get it, except `LongLivedSession` seems ... well weird.
We're basically just stripping out the lifecycle from the context, right?
So ... isn't this just a very fancy `context.WithoutCancel` ?

exchange/message/message.go
```diff
@@ -0,0 +1,48 @@
+package message
```
gbjk: 🚧 Change request:
`exchange/message` feels like a very overloaded term and it feels like you're domain-squatting  it a bit here.
I get that this *could* represent any message queue.
But right now we don't have an idea of any other type of messages here.
And this is fundamentally a relay, not a message thing.

So I'd like to see this in `exchange/websocket/relay/relay.go` instead.
But I'm also thinking it's even better in `exchange/websocket/buffer/relay.go` 🎉

exchange/websocket/manager.go
gbjk: In that case, I think we want some sort of `ErrNotInitialised` maybe ?
But also confused how it was getting past the `IsEnabled` in that case.

exchange/websocket/manager.go
```diff
@@ -972,11 +948,13 @@ func (m *Manager) observeConnection(t *time.Timer) (exit bool) {
 		}
 		// Speedier reconnection, instead of waiting for the next cycle.
 		if m.IsEnabled() && (!m.IsConnected() && !m.IsConnecting()) {
-			if connectErr := m.Connect(); connectErr != nil {
+			if connectErr := m.Connect(ctx); connectErr != nil {
 				log.Errorln(log.WebsocketMgr, connectErr)
 			}
 		}
-		m.DataHandler <- err // hand over the error to the data handler (shutdown and reconnection is priority)
+		if err := m.DataHandler.Send(ctx, err); err != nil { // hand over the error to the data handler (shutdown and reconnection is priority)
```
gbjk: 🐒 Optional nitpick:
This comment isn't helping me at all. Except I have to spend 20 seconds trying to grok what it thought was important to tell me, which I'm concluding is ... nothing.
```suggestion
		if err := m.DataHandler.Send(ctx, err); err != nil {
```

exchange/message/message.go
shazbert: My reasoning for placing this in `exchange/message` was that the relay is intentionally generic — it doesn’t depend on WebSockets or exchange code, and its API (message.Relay) is transport-agnostic by design.

If we move it into websocket/relay, it implicitly couples the type to a subsystem it doesn’t rely on and might make it harder to reuse later for other messaging paths (e.g., FIX, -- these are all a strech --> RPC, internal event pumps, REST batch queues).

That said, I’m happy to relocate it if the consensus is that the Relay should remain WebSocket-specific for now.

@thrasher- @gloriousCode 

Also Buffer naming I think will be subject to change because no websocket orderbooks should buffer events or sort them.

Also what about `exchange/stream/relay.go`?

engine/rpcserver.go
shazbert: Yeah good point, any other naming suggestions and I will change it, and should this function live somewhere like common?

exchange/websocket/manager.go
shazbert: I removed it: https://github.com/thrasher-corp/gocryptotrader/pull/2066/commits/9f5fbda6006ae68123b358ec8c269add130207f1

exchange/message/message.go
thrasher-: I think `exchange/stream/relay.go` is the best. Removes the websocket coupling since it's protocol agnostic and I agree that message is too generic

engine/rpcserver.go
gbjk: I mean... if it's `context.WithoutCancel` you don't need this function... right?
Just drop in replace it.

exchange/message/message.go
gbjk: * `exchange/stream/relay.go` is good 👍

🥃  _Your reasoning about buffer was backwards. Just because websocket orderbooks are in a package called buffer doesn't mean that nothing else can be, or that when they move out we couldn't leave a relay in there. But that's just an aside._

shazbert: I was more meaning, its going to change, then we will have to change the package name for all relay instances and it doesn't really reflect `buffer` that's all.

engine/rpcserver.go
shazbert: Well I feel like an idiot. 😆

shazbert: Nice catch: https://github.com/thrasher-corp/gocryptotrader/pull/2066/commits/ca24dc1445f170783cb8608e92f43c0cef392634

exchange/message/message.go
shazbert: done: https://github.com/thrasher-corp/gocryptotrader/pull/2066/commits/db647bfee93993ea8194893a68c65939f33d0533

exchange/stream/relay.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/thrasher-corp/gocryptotrader/common"
+)
+
+var errChannelBufferFull = errors.New("channel buffer is full")
+
```
gbjk, line 12: 🐒 Optional nitpick:
Can Relay be an interface please, and then type Relay doesn't leave this package :)
I noted we have concrete types (`*Relay`) in websocket manager, etc.
I think we might as well get this right first time.
NewRelay returns a `Relay` interface and nothing outside of here knows about it.
Were we to get circular dependencies (which I acknowledge we won't with this tiny package now) - consumers would just define the interface again there, or move the interface to common/interfaces/relay.go

shazbert: Interfaces are usually defined by the consuming package, so adding an interface felt like premature abstraction and for me personally makes it more complicated. 

That said, if we want to future-proof this now and avoid leaks of internal fields, I’m happy to wrap `Relay` in a small interface but I am going to have to change relay.C field functionality @thrasher- @gloriousCode @gbjk

thrasher-: Agree with @shazbert and this is applicable based on the go dev guidelines: https://go.dev/wiki/CodeReviewComments#interfaces

I think it's safer to let the consumer define the interface later if/when they specifically need it, rather than locking it in here and prematurely risking interface pollution

gbjk: Agreed. Resolving.

exchange/stream/relay.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// Send sends a message to the channel receiver
+// This is non-blocking and returns an error if the channel buffer is full
+func (r *Relay) Send(ctx context.Context, data any) error {
+	select {
+	case r.comm <- Payload{Ctx: common.FreezeCtx(ctx), Data: data}:
+		return nil
+	default:
+		return fmt.Errorf("%w: failed to relay <%T>", errChannelBufferFull, data)
```
gloriousCode: Not sure here, but am I silly for thinking Relay could have a connection URL on creation. Allowing for this error to be more meaningful. But your context itself might contain all of this info in the first place.

I only say it when looking at how its used:
```
	if err := e.Websocket.DataHandler.Send(ctx, trades[i]); err != nil {
				return err
			}
```
that won't really add much. But it doesn't do much right now either

exchange/stream/relay.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+}
+
+// Close closes the relay channel
+func (r *Relay) Close() {
```
gloriousCode, line 46: Should this be called when a websocket is shutdown? Should `NewRelay` also be called right after it? (like `ShutdownC`)
There isn't really a way right now that Calls `NewRelay` without re-setting up the exchange. But there's also nothing that calls close. But feels a bit more correct that if it does close, then restarting the ws fixes it. Get what I'm saying? I feel like I can't describe it. Its not super important, but it feels better setup

gloriousCode (changes requested):
I think this is looking good. I've been trying to break things in various ways. I'll continue tomorrow. For now I thought I'd give you one thing which I think is solid to do

exchange/stream/relay.go
shazbert: Yeah I see your point, unfortunately Relay doesn’t have strict ownership, so closing it during WS shutdown could cause send on closed channel panics. In Go, channel close is a signaling mechanism, not a cleanup mechanism.
Close() is currently used only in tests to terminate receivers deterministically, which is fine, but I don’t think it should be called implicitly during websocket shutdown unless we introduce stronger lifecycle guarantees.

gloriousCode: Yeah I get that too. I was thinking about if something did call close, for the ability to recover. But that resulting panic could highlight 1) a design issue where something shouldn't be calling close that needs to be resolved 2) something to fix when it happens

shazbert: So a panic here would likely be timing-related rather than a clear misuse, especially in highly subscribed connections where we close out the connection and its still reading from the ws buffer and trying to transport concurrently. I will slap it in see what happens though. 🤷

gloriousCode: I agreed with you and didn't want it 😄

gloriousCode (approved):
tACK! Thanks for working through the issues

exchange/stream/relay.go
shazbert: Fine I won't slap it in then! Slappity stopped!

PR #2073, opened 2025-10-04
request/rate limit: add context value check and fix bug 

shazbert:
# PR Description

* Change `RateLimit` function to be method of `RateLimiterWithWeight`
* Have single time value on all delay calls so that reservations can be cancelled correctly
* Add function `WithNoDelayPermitted` so that any delay will error for e.g. outbound orders that are time critical

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/request/limit.go
+++ b/exchanges/request/limit.go
@@ -7,6 +7,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	"github.com/thrasher-corp/gocryptotrader/common"
 	"golang.org/x/time/rate"
 )
 
@@ -15,9 +16,9 @@ var (
 	ErrRateLimiterAlreadyDisabled = errors.New("rate limiter already disabled")
 	ErrRateLimiterAlreadyEnabled  = errors.New("rate limiter already enabled")
 
-	errLimiterSystemIsNil       = errors.New("limiter system is nil")
-	errInvalidWeightCount       = errors.New("invalid weight count must equal or greater than 1")
-	errSpecificRateLimiterIsNil = errors.New("specific rate limiter is nil")
+	errLimiterSystemIsNil = errors.New("limiter system is nil")
+	errInvalidWeightCount = errors.New("invalid weight count must equal or greater than 1")
+	errNoDelayPermitted   = errors.New("no delay permitted")
 )
 
 // RateLimitNotRequired is a no-op rate limiter
@@ -46,18 +47,17 @@ type RateLimitDefinitions map[any]*RateLimiterWithWeight
 // refers to the number or weighting of the request. This is used to define
 // the rate limit for a specific endpoint.
 type RateLimiterWithWeight struct {
-	*rate.Limiter
-	Weight
+	limiter *rate.Limiter
+	weight  Weight
 }
 
 // Reservations is a slice of rate reservations
 type Reservations []*rate.Reservation
 
-// CancelAll cancels all potential reservations to free up rate limiter for
-// context cancellations and deadline exceeded cases.
-func (r Reservations) CancelAll() {
+// CancelAll cancels all potential reservations to free up rate limiter for context cancellations and deadline exceeded cases.
+func (r Reservations) CancelAll(at time.Time) {
 	for x := range r {
-		r[x].Cancel()
+		r[x].CancelAt(at)
 	}
 }
 
@@ -92,7 +92,7 @@ func NewWeightedRateLimitByDuration(interval time.Duration) *RateLimiterWithWeig
 // GetRateLimiterWithWeight couples a rate limiter with a weight count into an
 // accepted defined rate limiter with weight struct
 func GetRateLimiterWithWeight(l *rate.Limiter, weight Weight) *RateLimiterWithWeight {
-	return &RateLimiterWithWeight{l, weight}
+	return &RateLimiterWithWeight{limiter: l, weight: weight}
 }
 
 // NewBasicRateLimit returns an object that implements the limiter interface
@@ -116,7 +116,7 @@ func (r *Requester) InitiateRateLimit(ctx context.Context, e EndpointLimit) erro
 
 	rateLimiter := r.limiter[e]
 
-	err := RateLimit(ctx, rateLimiter)
+	err := rateLimiter.RateLimit(ctx)
 	if err != nil {
 		return fmt.Errorf("cannot rate limit request %w for endpoint %d", err, e)
 	}
@@ -133,44 +133,51 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
 	return r.limiter
 }
 
-// RateLimit is a function that will rate limit a request based on the rate
-// limiter provided. It will return an error if the context is cancelled or
-// deadline exceeded.
-func RateLimit(ctx context.Context, rateLimiter *RateLimiterWithWeight) error {
-	if rateLimiter == nil {
-		return errSpecificRateLimiterIsNil
+// RateLimit is a function that will rate limit a request based on the rate limiter provided. It will return an error if
+// the context is cancelled, deadline exceeded or if no delay is permitted via the context and a delay is required.
+func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
+	if err := common.NilGuard(r); err != nil {
+		return err
 	}
 
-	if rateLimiter.Weight <= 0 {
+	if r.weight == 0 {
 		return errInvalidWeightCount
 	}
 
+	tn := time.Now()
+	reservations := make(Reservations, r.weight)
 	var finalDelay time.Duration
-	reservations := make(Reservations, rateLimiter.Weight)
-	for i := Weight(0); i < rateLimiter.Weight; i++ {
-		// Consume 1 weight at a time as this avoids needing burst capacity in the limiter,
-		// which would otherwise allow the rate limit to be exceeded over short periods
-		reservations[i] = rateLimiter.Reserve()
-		finalDelay = reservations[i].Delay()
+	for i := range r.weight {
+		// Consume 1 weight at a time as this avoids needing burst capacity in the limiter, which would otherwise allow
+		// the rate limit to be exceeded over short periods
+		reservations[i] = r.limiter.ReserveN(tn, 1)
+		finalDelay = reservations[i].DelayFrom(tn)
 	}
 
-	if dl, ok := ctx.Deadline(); ok && dl.Before(time.Now().Add(finalDelay)) {
-		reservations.CancelAll()
-		return fmt.Errorf("rate limit delay of %s will exceed deadline: %w",
-			finalDelay,
-			context.DeadlineExceeded)
+	if finalDelay == 0 {
+		return nil
 	}
 
-	tick := time.NewTimer(finalDelay)
+	if hasNoDelayPermitted(ctx) {
+		reservations.CancelAll(tn)
+		return errNoDelayPermitted
+	}
+
+	// Check deadline before waiting to fail fast
+	if dl, ok := ctx.Deadline(); ok {
+		if dl.Before(tn.Add(finalDelay)) {
+			reservations.CancelAll(tn)
+			return fmt.Errorf("rate limit delay of %s will exceed deadline: %w", finalDelay, context.DeadlineExceeded)
+		}
+	}
+
 	select {
-	case <-tick.C:
+	case <-time.After(finalDelay):
 		return nil
 	case <-ctx.Done():
-		tick.Stop()
-		reservations.CancelAll()
+		reservations.CancelAll(tn)
 		return ctx.Err()
 	}
-	// TODO: Shutdown case
 }
 
 // DisableRateLimiter disables the rate limiting system for the exchange
@@ -193,4 +200,16 @@ func (r *Requester) EnableRateLimiter() error {
 		return fmt.Errorf("%s %w", r.name, ErrRateLimiterAlreadyEnabled)
 	}
 	return nil
+}
+
+type noDelayPermittedKey struct{}
+
+// WithNoDelayPermitted adds a value to the context that indicates that no delay is permitted for rate limiting
+func WithNoDelayPermitted(ctx context.Context) context.Context {
+	return context.WithValue(ctx, noDelayPermittedKey{}, struct{}{})
+}
+
+func hasNoDelayPermitted(ctx context.Context) bool {
+	_, ok := ctx.Value(noDelayPermittedKey{}).(struct{})
+	return ok
 }
--- a/exchanges/request/limit_test.go
+++ b/exchanges/request/limit_test.go
@@ -1,0 +1,164 @@
+package request
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/common"
+	"golang.org/x/time/rate"
+)
+
+func TestRateLimit(t *testing.T) {
+	t.Parallel()
+
+	err := (*RateLimiterWithWeight)(nil).RateLimit(t.Context())
+	assert.ErrorIs(t, err, common.ErrNilPointer)
+
+	r := &RateLimiterWithWeight{limiter: rate.NewLimiter(rate.Limit(10), 1)}
+	err = r.RateLimit(t.Context())
+	assert.ErrorIs(t, err, errInvalidWeightCount, "should return errInvalidWeightCount for zero weight")
+
+	r = NewRateLimitWithWeight(time.Second, 10, 1)
+	start := time.Now()
+	err = r.RateLimit(t.Context())
+	elapsed := time.Since(start)
+	require.NoError(t, err, "rate limit must not error")
+	assert.Less(t, elapsed, time.Millisecond*50, "should complete quickly for first request")
+
+	r = NewRateLimitWithWeight(time.Second, 10, 5)
+	start = time.Now()
+	err = r.RateLimit(t.Context())
+	elapsed = time.Since(start)
+	require.NoError(t, err, "rate limit must not error")
+	assert.Less(t, elapsed, time.Millisecond*600, "should complete within reasonable time for weight 5")
+
+	r = NewRateLimitWithWeight(100*time.Millisecond, 1, 1)
+	start = time.Now()
+	err = r.RateLimit(WithNoDelayPermitted(t.Context()))
+	require.NoError(t, err, "first rate limit call must not error and must be immediate")
+	firstElapsed := time.Since(start)
+	assert.Less(t, firstElapsed, 50*time.Millisecond, "first call should be immediate")
+
+	start = time.Now()
+	err = r.RateLimit(t.Context())
+	require.NoError(t, err, "second rate limit call must not error")
+	secondElapsed := time.Since(start)
+	assert.GreaterOrEqual(t, secondElapsed, 90*time.Millisecond, "second call should be delayed by approximately 100ms")
+	assert.Less(t, secondElapsed, 150*time.Millisecond, "delay should not be excessive")
+
+	err = r.RateLimit(WithNoDelayPermitted(t.Context()))
+	assert.ErrorIs(t, err, errNoDelayPermitted, "should return correct error")
+
+	ctx, cancel := context.WithCancel(t.Context())
+	cancel()
+	err = r.RateLimit(ctx)
+	assert.ErrorIs(t, err, context.Canceled, "should return correct error")
+
+	ctx, cancel = context.WithDeadline(t.Context(), time.Now())
+	defer cancel()
+
+	err = r.RateLimit(ctx)
+	assert.ErrorIs(t, err, context.DeadlineExceeded, "should return correct error")
+}
+
+func TestReservationsCancelAll(t *testing.T) {
+	t.Parallel()
+
+	var reservations Reservations
+	require.NotPanics(t, func() { reservations.CancelAll(time.Now()) }, "cancelling empty reservations must not panic")
+
+	r := NewRateLimitWithWeight(time.Millisecond*50, 1, 1)
+
+	err := r.RateLimit(t.Context())
+	require.NoError(t, err, "must not error and must be immediate when reservations are set")
+
+	err = r.RateLimit(WithNoDelayPermitted(t.Context()))
+	require.ErrorIs(t, err, errNoDelayPermitted, "must error when no delay permitted this will reset reservations")
+
+	tn := time.Now()
+	err = r.RateLimit(t.Context())
+	elapsed := time.Since(tn)
+	require.NoError(t, err, "RateLimit must not error")
+	assert.LessOrEqual(t, elapsed, 60*time.Millisecond, "should be delayed by approximately 50ms when reservations are reset on error")
+}
+
+func TestNewRateLimit(t *testing.T) {
+	t.Parallel()
+
+	r := NewRateLimit(time.Second, 10)
+	require.NotNil(t, r, "limiter must not be nil")
+	assert.Equal(t, rate.Limit(10), r.Limit(), "limit should be 10 per second")
+
+	r = NewRateLimit(time.Second, 0)
+	require.NotNil(t, r, "limiter must not be nil")
+	assert.Equal(t, rate.Inf, r.Limit(), "limit should be infinite on zero actions")
+
+	r = NewRateLimit(time.Second, -1)
+	require.NotNil(t, r, "limiter must not be nil")
+	assert.Equal(t, rate.Inf, r.Limit(), "limit should be infinite on negative actions")
+
+	r = NewRateLimit(0, 10)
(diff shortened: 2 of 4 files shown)


exchanges/request/limit.go
```diff
@@ -133,44 +133,51 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
...
-		return errSpecificRateLimiterIsNil
+// RateLimit is a function that will rate limit a request based on the rate limiter provided. It will return an error if
+// the context is cancelled, deadline exceeded or if no delay is permitted via the context and a delay is required.
+func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
+	if err := common.NilGuard(r); err != nil {
+		return err
 	}
 
-	if rateLimiter.Weight <= 0 {
+	if r.weight == 0 {
```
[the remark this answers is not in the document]
shazbert: it's a uint my dude

thrasher-:
@codex please review

exchanges/request/limit.go
```diff
@@ -4,20 +4,23 @@ import (
...
 
+	"github.com/thrasher-corp/gocryptotrader/common"
 	"golang.org/x/time/rate"
 )
 
 // Defines rate limiting errors
 var (
 	ErrRateLimiterAlreadyDisabled = errors.New("rate limiter already disabled")
 	ErrRateLimiterAlreadyEnabled  = errors.New("rate limiter already enabled")
+	ErrNoDelayPermitted           = errors.New("no delay permitted")
```
gbjk: 🦍 Optional pedantic nitpick:
* `permitted` vs `allowed` - I think `allowed` is more idiosyncratic
* `NoXY` vs `XNotY`
* Considered if this should be `ErrInvalidDelay`
```suggestion
	ErrDelayNotAllowed           = errors.New("delay not allowed")
```

exchanges/request/limit.go
```diff
@@ -4,20 +4,23 @@ import (
...
 
-	errLimiterSystemIsNil       = errors.New("limiter system is nil")
-	errInvalidWeightCount       = errors.New("invalid weight count must equal or greater than 1")
-	errSpecificRateLimiterIsNil = errors.New("specific rate limiter is nil")
+	errLimiterSystemIsNil = errors.New("limiter system is nil")
```
gbjk: 🐒 Optional nitpick:
Should this exist, or should we use common.ErrNilGuard and/or common.ErrNilPointer

exchanges/request/limit.go
```diff
@@ -4,20 +4,23 @@ import (
...
+	errInvalidWeightCount = errors.New("invalid weight count must equal or greater than 1")
```
gbjk: 🐒 Optional nitpick:
```suggestion
	errInvalidWeight = errors.New("weight must be equal-or-greater than 1")
```

exchanges/request/limit.go
```diff
@@ -46,19 +49,9 @@ type RateLimitDefinitions map[any]*RateLimiterWithWeight
 // refers to the number or weighting of the request. This is used to define
```
gbjk: 🐒 Optional nitpick:
L45 says `weight count` and I'd prefer it to just be `weight`.
It's not counting anything 🙂

exchanges/request/limit.go
```diff
@@ -133,44 +126,59 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
...
 }
 
-// RateLimit is a function that will rate limit a request based on the rate
-// limiter provided. It will return an error if the context is cancelled or
-// deadline exceeded.
-func RateLimit(ctx context.Context, rateLimiter *RateLimiterWithWeight) error {
-	if rateLimiter == nil {
-		return errSpecificRateLimiterIsNil
+// RateLimit is a function that will rate limit a request based on the rate limiter provided. It will return an error if
+// the context is cancelled, deadline exceeded or if no delay is permitted via the context and a delay is required.
```
gbjk: 🐒 Optional nitpick:
* This is a method, so "on the rate limiter provided" is implied
* Erroring on context cancelled or deadline exceeded is implicit always when using contexts, so we can remove those comments. And conversely: Any function *not* respecting the context given to it is a bug 🐛 
* I've attempted to get the wording right for the verb for this function ... `throttles` seemed kinda-okay, but I've also considered `delays`
```suggestion
// RateLimit throttles a request based on weight, delaying the request 
// Errors if no delay is permitted via the context and a delay is required
```

exchanges/request/limit.go
```diff
@@ -133,44 +126,59 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
...
-		// Consume 1 weight at a time as this avoids needing burst capacity in the limiter,
-		// which would otherwise allow the rate limit to be exceeded over short periods
-		reservations[i] = rateLimiter.Reserve()
-		finalDelay = reservations[i].Delay()
+	r.m.Lock()
+	tn := time.Now()
+	reserved := make([]*rate.Reservation, 0, r.weight)
+	for range r.weight {
+		// This avoids needing burst capacity in the limiter, which would otherwise allow the rate limit to be exceeded over short periods
+		reserved = append(reserved, r.limiter.ReserveN(tn, 1))
```
gbjk, line 133: 🤷‍♂️ Clarification please: Are we holding this weapon correctly?
It seems ... that `RateLimit` is ... maybe not actually `RateLimit` and that `InitiateRateLimit` is maybe `RateLimit`.

I'm raising this without enough understanding of it, and I may retract the question later, but I figured I'd just ask early.

exchanges/request/limit.go
```diff
@@ -133,44 +126,59 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
...
-		tick.Stop()
-		reservations.CancelAll()
 		return ctx.Err()
+	case <-time.After(finalDelay):
+		return nil
+	}
+}
+
+// cancelAll cancels all reservations at a specific time, this must be called in a lock with reservations so that
+// correct reimbursement takes place due to concurrency.
```
gbjk: 🤷‍♂️ Clarification please: 
* Kinda wishing this could lock itself, and comparing the way this needs a WLock to the previous RLock that's implicit in the caller
🐒 Optional nitpick: Wording feels painful, and we don't need to say `due to concurrency`
```suggestion
// cancelAll cancels all reservations at a specific time
// Does not provide locking protection, so callers can maintain a single lock throughout
```

exchanges/request/limit.go
shazbert: Short answer no I don't think we are. There is a lot problems here naming wise but wanted to keep it light, I was going to shove this entire file in to `exchange/rate (or something similar)` but that will take time. Once you get a better understanding of it, the different functions names will be good for a cleaner call. But might be better for a refactor PR.

exchanges/request/limit.go
shazbert: Yeah it would be cleaner to lock by itself but if that were introduced then there might be temptation for it to be used in the select case in RateLimit and on `ctx.Done()` which may lead to some reimbursements that shouldn't happen.

exchanges/request/limit.go
gbjk: Okay; so if I review this lightly and accept that it's not making things *worse*, then we can revisit another day.

gbjk (approved):
Thanks 👍 

I'm not yet confident in the overall design and usage of rate.Limiter, but I know that if I start sniffing that it'd be way bigger, and out-of-scope of a bugfix PR.

exchanges/request/limit.go
```diff
@@ -110,13 +101,13 @@ func (r *Requester) InitiateRateLimit(ctx context.Context, e EndpointLimit) erro
...
-	if r.limiter == nil {
-		return fmt.Errorf("cannot rate limit request %w", errLimiterSystemIsNil)
+	if err := common.NilGuard(r.limiter); err != nil {
+		return err
 	}
 
 	rateLimiter := r.limiter[e]
 
-	err := RateLimit(ctx, rateLimiter)
+	err := rateLimiter.RateLimit(ctx)
```
samuael:
```suggestion
	if err := rateLimiter.RateLimit(ctx); err != nil {
```

exchanges/request/limit.go
```diff
@@ -133,44 +124,59 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
...
+		return nil
 	}
 
-	if dl, ok := ctx.Deadline(); ok && dl.Before(time.Now().Add(finalDelay)) {
-		reservations.CancelAll()
-		return fmt.Errorf("rate limit delay of %s will exceed deadline: %w",
-			finalDelay,
-			context.DeadlineExceeded)
+	if hasDelayNotAllowed(ctx) {
+		cancelAll(reserved, tn)
```
samuael, line 143: I couldn't see where this tn is updated ... it still represents time.Now() and it is being set as a Reservation cancellation timestamp

shazbert: The tn var is never updated, it's a static value. There is a slight drift issue (gc cycles) with delays when they have a weight and a slight race that occurs when accessing this concurrently, so keeping the time value the same and having it protected by a mutex makes it more consistent due to the token bucket design.  Passing that value into the cancelAll function allows all the reservations that are done at that exact timestamp to be freed correctly.

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,245 @@
...
+	assert.ErrorIs(t, err, common.ErrNilPointer)
+
+	r := &RateLimiterWithWeight{limiter: rate.NewLimiter(rate.Limit(1), 1)}
+	err = r.RateLimit(t.Context())
+	assert.ErrorIs(t, err, errInvalidWeight, "should return errInvalidWeightCount for zero weight")
+
+	r = NewRateLimitWithWeight(time.Second, 10, 1)
+	start := time.Now()
+	err = r.RateLimit(t.Context())
+	elapsed := time.Since(start)
```
thrasher-: I was going to nit these since they'd be non-used variables if RateLimit was to ever error but then it could throw off your millisecond timer 😂

thrasher- (approved):

exchanges/request/limit.go
```diff
@@ -133,44 +119,59 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
...
+func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
+	if err := common.NilGuard(r); err != nil {
+		return err
+	}
+
+	if r.weight == 0 {
+		return errInvalidWeight
+	}
+
+	r.m.Lock()
```
cranktakular: Pretty sure this lock should be moved up before the "if r.weight == 0" line. Otherwise, another goroutine could set it to 0 immediately after it's checked, letting us proceed with that invalid weight.

If we're not worried about weight changing, we could put reserved's instantiation before this lock.

exchanges/request/limit.go
```diff
@@ -133,44 +119,59 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
 	return r.limiter
 }
 
-// RateLimit is a function that will rate limit a request based on the rate
-// limiter provided. It will return an error if the context is cancelled or
-// deadline exceeded.
-func RateLimit(ctx context.Context, rateLimiter *RateLimiterWithWeight) error {
-	if rateLimiter == nil {
-		return errSpecificRateLimiterIsNil
+// RateLimit throttles a request based on weight, delaying the request
```
cranktakular: Documentation comments like these [should always be full sentences with periods at the end](https://google.github.io/styleguide/go/decisions#comment-sentences).

This lives in a few other functions touched, here.

exchanges/request/limit.go
```diff
@@ -194,3 +195,15 @@ func (r *Requester) EnableRateLimiter() error {
 	}
 	return nil
 }
+
+type delayNotAllowedKey struct{}
+
+// WithDelayNotAllowed adds a value to the context that indicates that no delay is allowed for rate limiting
+func WithDelayNotAllowed(ctx context.Context) context.Context {
+	return context.WithValue(ctx, delayNotAllowedKey{}, struct{}{})
```
cranktakular, line 198: John Golang disapproves

`Use context Values only for request-scoped data that transits processes and APIs, not for passing optional parameters to functions.`

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,245 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/common"
+	"golang.org/x/time/rate"
+)
+
+func TestRateLimit(t *testing.T) {
```
cranktakular, line 16: You should use synctest for this and any other test functions that deal with time. Uses fake time so it runs way faster, and reflects ordinary operation consistently.

That fake time also lets you make these time checks more precise.

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,245 @@
...
+	start = time.Now()
+	err = r.RateLimit(WithDelayNotAllowed(t.Context()))
+	require.NoError(t, err, "first rate limit call must not error and must be immediate")
+	firstElapsed := time.Since(start)
+	assert.Less(t, firstElapsed, 50*time.Millisecond, "first call should be immediate")
+
+	start = time.Now()
+	err = r.RateLimit(t.Context())
+	require.NoError(t, err, "second rate limit call must not error")
+	secondElapsed := time.Since(start)
```
cranktakular: Why do you move to using firstElapsed and secondElapsed? Can't you just keep reusing elapsed like you did earlier?

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,245 @@
...
+	var routineErr error
+	wg := sync.WaitGroup{}
+	wg.Go(func() { routineErr = r.RateLimit(t.Context()) })
+	ctx, cancel := context.WithCancel(t.Context())
+	cancel()
+	time.Sleep(10 * time.Millisecond)
+	err = r.RateLimit(ctx)
+	assert.ErrorIs(t, err, context.Canceled, "should return correct error")
+	wg.Wait()
+	assert.NoError(t, routineErr, "routine should complete successfully will providing friction for above context cancellation")
```
cranktakular: Is the word here, and in the next similar error message, meant to be "while" instead of "will"?

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,245 @@
...
+
+	wg.Go(func() { routineErr = r.RateLimit(t.Context()) })
+	ctx, cancel = context.WithDeadline(t.Context(), time.Now())
+	defer cancel()
+	time.Sleep(10 * time.Millisecond)
+
```
cranktakular: Odd newline; there isn't one in a similar position above.

exchanges/request/limit.go
```diff
@@ -110,17 +101,12 @@ func (r *Requester) InitiateRateLimit(ctx context.Context, e EndpointLimit) erro
 	if atomic.LoadInt32(&r.disableRateLimiter) == 1 {
 		return nil
 	}
-	if r.limiter == nil {
-		return fmt.Errorf("cannot rate limit request %w", errLimiterSystemIsNil)
+	if err := common.NilGuard(r.limiter); err != nil {
+		return err
```
cranktakular, line 100: Can you get coverage of this line?

cranktakular (changes requested):
A few things I would like changed

exchanges/request/limit.go
shazbert: nice catch

exchanges/request/limit.go
shazbert: I am not going to update the full function signature from wrapper to accommodate this request. You can add an issue and open a PR for it when you have time. It needs to accommodate accounts.DeployCredentialsToContext, request.WithVerbose and this. Can be an optional param but this is out of scope.

exchanges/request/limit_test.go
shazbert: YES! For things like this you can provide suggestions and then I can commit it straight way and pull it into my branch

exchanges/request/limit_test.go
shazbert: If you want to send me a patch I will update it, but I am not going to use sync test for this. The tests currently work.

exchanges/request/limit.go
```diff
@@ -13,7 +13,7 @@ import (
 	"golang.org/x/time/rate"
 )
 
-// Defines rate limiting errors
+// Defines rate limiting errors.
```
gbjk: 🐒 🗻  Optional nitpick:
Since you touched it...
This wouldn't read correctly in godoc as-is.
Mountain monkey because you didn't introduce it.
```suggestion
// Rate limiting errors.
```

exchanges/request/limit_test.go
```diff
@@ -62,17 +63,16 @@ func TestRateLimit(t *testing.T) {
 	err = r.RateLimit(ctx)
 	assert.ErrorIs(t, err, context.Canceled, "should return correct error")
 	wg.Wait()
-	assert.NoError(t, routineErr, "routine should complete successfully will providing friction for above context cancellation")
+	assert.NoError(t, routineErr, "routine should complete successfully while providing friction for above context cancellation")
```
gbjk: 🦍 Optional pedantic nitpick:
I have no idea what "while providing friction" means.
Do we need lube?
For clarity, as a maintainer, I'd have no idea what anything after "complete successfully" means, or how it should impact the test expectations.
And now ... I'm thinking it should be:
```suggestion
	assert.NoError(t, routineErr, "waitgroup should be Done when context is canceled")
```

exchanges/request/limit_test.go
```diff
@@ -243,3 +243,24 @@ func TestCancelAll(t *testing.T) {
...
+	assert.ErrorIs(t, err, ErrRequestSystemIsNil, "should return correct error")
+
+	r = &Requester{}
+	atomic.StoreInt32(&r.disableRateLimiter, 1)
+	err = r.InitiateRateLimit(t.Context(), Unset)
+	assert.NoError(t, err, "should not error when rate limiter is disabled")
+
+	atomic.StoreInt32(&r.disableRateLimiter, 0)
+	err = r.InitiateRateLimit(t.Context(), Unset)
+	assert.ErrorIs(t, err, common.ErrNilPointer, "should return correct error when limiter is nil")
```
gbjk: 🦍 Optional pedantic nitpick:
I've switched to always using `ErrorContains` on these, just to make sure the NilPointer is really the one I was expecting.

gbjk (approved):

exchanges/request/limit_test.go
cranktakular: [my-changes.patch](https://github.com/user-attachments/files/23733918/my-changes.patch)

shazbert: That patch was broken @cranktakular but applied it manually

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func TestRateLimit_Linear_WithFailure(t *testing.T) {
+	t.Parallel()
+
+	synctest.Test(t, func(t *testing.T) {
+		t.Helper()
+		r := NewRateLimitWithWeight(time.Second, 10, 1)
+		tn := time.Now()
+		var wg sync.WaitGroup
```
thrasher-: This is defined but running synchronously since we're not using Go like the others

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,281 @@
...
+}
+
+func TestRateLimit_Linear(t *testing.T) {
+	t.Parallel()
+
+	synctest.Test(t, func(t *testing.T) {
+		t.Helper()
+		r := NewRateLimitWithWeight(time.Second, 10, 1)
+		tn := time.Now()
+		var wg sync.WaitGroup
```
thrasher-: Same here

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,281 @@
...
+		}
+		assert.Less(t, time.Since(tn), time.Second, "should complete within reasonable time")
+	})
+}
+
+func TestRateLimit_Linear_WithFailure(t *testing.T) {
+	t.Parallel()
+
+	synctest.Test(t, func(t *testing.T) {
+		t.Helper()
```
thrasher-: This can be removed since it's usually reserved for utility functions and for the others

thrasher-: Sorry sir, fixed upstream https://github.com/kulti/thelper/issues/48 but need to bump our golangci-lint to rid the need for nolints

exchanges/request/limit_test.go
```diff
@@ -16,137 +17,132 @@ import (
...
+		var routineErr error
+		wg := sync.WaitGroup{}
+		wg.Go(func() { routineErr = r.RateLimit(t.Context()) })
+		ctx, cancel := context.WithCancel(t.Context())
+		cancel()
+		time.Sleep(10 * time.Millisecond)
+		err = r.RateLimit(ctx)
+		assert.ErrorIs(t, err, context.Canceled, "should return correct error")
+		wg.Wait()
+		assert.NoError(t, routineErr, "waitgroup should be Done when context is canceled")
```
gbjk: 🤷‍♂️ I am new to synctest. I guess we all are.
But familiarising myself with it, I thought this is exactly when we'd have expected to use `synctest.Wait()`. I got that impression from the [original](https://go.dev/blog/synctest#testing-time) and [new](https://go.dev/blog/testing-time#waiting-for-quiescence) blogposts.

I don't have anything actionable here, cos I'll need to play with it more.
It's just a sense that we should be actually using `synctest.Wait()` inside the bubble.

Feel free to resolve. We can merge and then consider later.

gbjk (approved):
Good work 👍 

My review is low value, because I'm just getting up to speed with synctest, which is most of what's changed here since my last review.

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,262 @@
...
```
gbjk: 🚧 Change request:
At the least, the comment here is incorrect.
But the more I look at Waitgroup usage here, the more I'm sure we're not using `synctest.Wait` when we should.

gbjk: This is what I came up with: gbjk@e8eae9a2c
```
curl -s https://github.com/gbjk/gocryptotrader/commit/e8eae9a2c.patch | git apply
```

Note: I might be missing something, but I couldn't see any value of the background RateLimits.
They're not testing anything we didn't test before. The contexts aren't inter-mingled.
I couldn't grok what they're adding _now_ other than complexity.
(Before maybe they added something, though)

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,256 @@
...
+		err = r.RateLimit(t.Context())
+		assert.ErrorIs(t, err, errInvalidWeight, "should return errInvalidWeightCount for zero weight")
+
+		r = NewRateLimitWithWeight(time.Second, 10, 1)
+		start := time.Now()
+		err = r.RateLimit(t.Context())
+		synctest.Wait()
+		elapsed := time.Since(start)
+		require.NoError(t, err, "rate limit must not error")
+		assert.Less(t, elapsed, time.Millisecond*50, "should complete quickly for first request")
```
gbjk: 🦍 Optional pedantic nitpick:
This gorilla is a 🍆 , but he thinks we can save some lines in a test file 🙄 
```suggestion
		assert.Less(t, time.Since(start), time.Millisecond*50, "should complete quickly for first request")
		require.NoError(t, err, "rate limit must not error")
```

gbjk (approved):

thrasher- (approved):
Changes ACK, thanks!

exchanges/request/limit_test.go
```diff
@@ -0,0 +1,252 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/common"
+	"golang.org/x/time/rate"
+)
+
+func TestRateLimit(t *testing.T) {
+	t.Parallel()
+
+	synctest.Test(t, func(t *testing.T) { //nolint:thelper // false positive
```
gloriousCode: I get the opposite locally, but I see lint is ticked online 🌈 So nothing is wrong
```
exchanges/request/limit_test.go:19:40: directive `//nolint:thelper // false positive` is unused for linter "thelper" (nolintlint)
	synctest.Test(t, func(t *testing.T) { //nolint:thelper // false positive
```

thrasher-: this patch will fix that: 

```console
Invoke-WebRequest -Uri "https://gist.githubusercontent.com/thrasher-/de6a9b2eedf228aabaeabb5da98e688c/raw/61f22b77b1eb7523271c75587fcdb7918fb344ba/woof.patch" | ForEach-Object { $_.Content | git apply }
```

shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2073/commits/24133d40e1c67ab37ed22e02eb663a7ff67491b8

PR #2051, opened 2025-09-11
Okx: Fix authenticated websocket login

dazi005:
# PR Description

Please include a summary of the change, feature or issue which this pull request addresses. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes #1970 
if i use SendAuthenticatedWebsocketRequest,i will got err
image

e.Websocket.AuthConn.SendJSONMessage(ctx, request.Unset, op)   can login

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2076, opened 2025-10-07
NilGuard: return on first nil

gloriousCode:
# PR Description
I think an improvement to `common.NilGuard` is to return on the first nil, rather than to append all nils. This allows for (lack of a better term) compounding nil checks, eg calling `common.NilGuard(e, e.Websocket, e.Websocket.Match)`. Calling that snippet now would lead to a panic if `e` was nil. [See this](https://github.com/thrasher-corp/gocryptotrader/pull/2051#discussion_r2409345028)for a real world example

I think it still achieves its goal of guarding against nils. Since the error returned is always a slice of `common.ErrNilPointer`, there isn't much loss in the slice return as you cannot really do too much extra with each err returned to inspect which param is nil. 

## Type of change

## How has this been tested
- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/common/common.go
+++ b/common/common.go
@@ -84,7 +84,7 @@ var (
 )
 
 // NilGuard returns an ErrNilPointer with the type of the first nil argument
-func NilGuard(ptrs ...any) (errs error) {
+func NilGuard(ptrs ...any) error {
 	for _, p := range ptrs {
 		/* 	Internally interfaces contain a type and a value address
 		Obviously can't compare to nil, since the types won't match, so we look into the interface
@@ -93,10 +93,10 @@ func NilGuard(ptrs ...any) (errs error) {
 		We optimize here by converting to [2]uintptr and just checking the address, instead of casting to a local eface type
 		*/
 		if (*[2]uintptr)(unsafe.Pointer(&p))[1] == 0 {
-			errs = AppendError(errs, fmt.Errorf("%w: %T", ErrNilPointer, p))
+			return fmt.Errorf("%w: %T", ErrNilPointer, p)
 		}
 	}
-	return errs
+	return nil
 }
 
 // MatchesEmailPattern ensures that the string is an email address by regexp check
--- a/common/common_test.go
+++ b/common/common_test.go
@@ -670,11 +670,6 @@ func TestNilGuard(t *testing.T) {
 
 	err = NilGuard(&s, nil, (*int)(nil))
 	assert.ErrorIs(t, err, ErrNilPointer)
-	assert.ErrorContains(t, err, "*int")
-	var mErr *multiError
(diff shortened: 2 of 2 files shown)


thrasher-:
@codex please review

common/common.go
```diff
@@ -83,8 +83,9 @@ var (
 	errHTTPClientInvalid       = errors.New("custom http client cannot be nil")
 )
 
-// NilGuard returns an ErrNilPointer with the type of the first nil argument
-func NilGuard(ptrs ...any) (errs error) {
+// NilGuard returns an ErrNilPointer with the type of the first nil value found in the arguments.
+// It returns immediately upon encountering the first nil, which may not be the first argument.
```
gbjk, line 87: 🚧 Change request:
* Existing comment was both incorrect for the original implementation, and valid for the new implementation
* This explantation kinda feels overcooked - It just returns for the first nil argument. End of story.
```suggestion 
// NilGuard returns an ErrNilPointer with the type of the first nil argument
```

common/common_test.go
```diff
@@ -670,11 +670,6 @@ func TestNilGuard(t *testing.T) {
 
 	err = NilGuard(&s, nil, (*int)(nil))
 	assert.ErrorIs(t, err, ErrNilPointer)
-	assert.ErrorContains(t, err, "*int")
-	var mErr *multiError
-	require.ErrorAs(t, err, &mErr, "err must be a multiError")
-	assert.Len(t, mErr.Unwrap(), 2, "Should get 2 errors back")
-
```
gbjk, line 677: 🚧 Change request:
* I agree that you needed to change this,  I think you need to keep the assert on type, and switch around the args so you get that. In other words: We want to cover that we get past a non-nil input and find a subsequent error
* I think you need coverage of not blowing up
```suggestion
    type nilWrap struct {
        i *int
    }
    var nilWrap *nilWrap
	err = NilGuard(&s, &nilWrap, nilWrap.i)
	assert.ErrorIs(t, err, ErrNilPointer)
	assert.ErrorContains(t, err, "*nilWrap")
```

~Note, Haven't tested this at all. It's just communicating my feedback in code.~

gbjk (changes requested):
👍 Strong approval.
**If this works!**

I wanted to do this, but I also felt it was important to gather all the errors together.

Having released this beast into the wild, I now see that it's often a PITA to wield, and if we can return on the first error, we should.

I actually thought what you're suggesting would not work.
I thought that go would panic on the *call* to NilGuard, where it tries to access the nested value.

It does on my system....

Am I missing something?

gloriousCode:
My goodness, thanks for the review gk. I realise I forgot how golang works and this won't be an effective solution. I'll look into seeing if we can clean things up still though. But that can be another PR rather than clogging up our already clogged PRs

PR #2090, opened 2025-10-19
common.go: Update ErrorCollector to use mutex and simplify error collection in concurrent operations

shazbert:
# PR Description

* leverages `sync.WaitGroup Go` method to auto scale so we don't need to specify routine count 
* uses `common.AppendError` on field errs protected by mutex so we don't need to specify chan len
* privatises fields so only have method calls, reducing caller complexity
* have func closure return an error, so that error management is done internally 
* drop `CollectErrors` func so `ErrorCollector` struct does not need to be heap allocated

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/request/request_test.go
+++ b/exchanges/request/request_test.go
@@ -307,9 +307,8 @@ func TestDoRequest(t *testing.T) {
 	require.False(t, respErr.Error, "Error must be false")
 
 	// Check client side rate limit
-	const numReqs = 5
 	var ec common.ErrorCollector
-	for range numReqs {
+	for range 5 {
 		ec.Go(func() error {
 			var resp struct {
 				Response bool `json:"response"`
@@ -339,9 +338,8 @@ func TestDoRequest_Retries(t *testing.T) {
 	r, err := New("test", new(http.Client), WithBackoff(func(int) time.Duration { return 0 }))
 	require.NoError(t, err, "New requester must not error")
 
-	const numReqs = 4
 	var ec common.ErrorCollector
-	for range numReqs {
+	for range 4 {
 		ec.Go(func() error {
 			var resp struct {
 				Response bool `json:"response"`


common/common.go
```diff
@@ -553,31 +553,31 @@ func ExcludeError(err, excl error) error {
...
-	e.Wg.Wait()
-	close(e.C)
-	for err := range e.C {
-		if err != nil {
-			errs = AppendError(errs, err)
-		}
+// Go runs a function in a goroutine and collects any error it returns
+func (e *ErrorCollector) Go(f func() error) {
+	if f == nil {
+		panic("ErrorCollector.Go received a nil function")
```
gloriousCode: I would argue that if things are the caller's responsibility, which is something you often advocate, then this should just return rather than panic. A user who is calling this can check if the function that they are passing in is nil before doing so and can panic on their own terms ie a param based function pass in, instead of inline

common/common.go
```diff
@@ -553,31 +553,31 @@ func ExcludeError(err, excl error) error {
...
-// CollectErrors returns an ErrorCollector with WaitGroup and Channel buffer set to n
-func CollectErrors(n int) *ErrorCollector {
-	e := &ErrorCollector{
-		C: make(chan error, n),
-	}
-	e.Wg.Add(n)
-	return e
+// Collect waits for the internal wait group to be done and returns an error collection
+func (e *ErrorCollector) Collect() (errs error) {
+	e.wg.Wait()
```
gloriousCode, line 565: Something that gives pause is that if you notice that there is something wrong with a function passed in, eg takes longer than expected, you are now stuck

I would have `AwaitAndCollect` which is this function and a function `Collect` which just locks the mutex and returns errs, but that may not be suitable for you.

If you want silly increases in complexity, you could also pass in a context here and check for `ctx.Done()`/`e.wg.Wait()`

edit: im not super fussed here either, we have to trust our functions! Having an out could be important if stuck

common/common.go
```diff
@@ -553,31 +553,31 @@ func ExcludeError(err, excl error) error {
...
+	e.m.Lock()
+	defer e.m.Unlock()
+	return e.errs
```
gloriousCode, line 568: After collection, you don't clear errors, so it's not reusable. Not huge, but anyone who does retry something, will not have any idea that the error they get back is the one from last time

common/common.go
```diff
@@ -553,31 +553,31 @@ func ExcludeError(err, excl error) error {
...
+func (e *ErrorCollector) Go(f func() error) {
+	if f == nil {
+		panic("ErrorCollector.Go received a nil function")
 	}
-	return
+	e.wg.Go(func() {
+		err := f()
+		e.m.Lock()
+		e.errs = AppendError(e.errs, err)
+		e.m.Unlock()
```
gloriousCode: Not big, but surely a check `if err != nil` costs less than a blind lock/unlock + function call
ie
```
if err := f(); err != nil {
    e.m.Lock()
    e.errs = AppendError(e.errs, err)
    e.m.Unlock()
}

File: common/common.go (hunk 1 above)
Comment (shazbert) [reply common/common.go]: It's usually a closure right? Even in sync.Waitgroup `Go` func call it panics if func is nil. I only just added this check outside of the lower go routine for easier debugging.

Comment (gloriousCode) [reply common/common.go]: It does not panic on nil. If it did I'd argue to not introduce an extra one, but much less strongly. I don't know what you mean about closure though.

This isn't something I'm very fussed on, but I don't see the need to introduce it

Comment (shazbert) [reply common/common.go]:
```
[signal 0xc0000005 code=0x0 addr=0x0 pc=0x7ff711d843e2]

goroutine 10 [running]:
sync.(*WaitGroup).Go.func1()
	C:/Program Files/Go/src/sync/waitgroup.go:239 +0x42
created by sync.(*WaitGroup).Go in goroutine 9
	C:/Program Files/Go/src/sync/waitgroup.go:237 +0x73
FAIL	github.com/thrasher-corp/gocryptotrader/exchanges	0.233s
```

```go
func TestSomething(t *testing.T) {
	wg := sync.WaitGroup{}
	wg.Go(nil)
	wg.Wait()
}
```

It most certainly does. Its like us passing in a RequestParam struct pointer it panics if nil. I am not seeing an issue here.

gloriousCode: Hey no fair, you said wg.Go panics on nil, not wait

shazbert: it panics on nil `created by sync.(*WaitGroup).Go in goroutine 9` wait is usually a pattern called after.

common/common.go
shazbert: > Something that gives pause is that if you notice that there is something wrong with a function passed in, eg takes longer than expected, you are now stuck

The same can be said when using sync.WaitGroup Go method and then calling Wait. We seem to be okay. 

> I would have AwaitAndCollect which is this function and a function Collect which just locks the mutex and returns errs, but that may not be suitable for you.

I don't see the need of having another func `Collect` because you're bypassing the point of waiting until the routines have finished that are called above. Which seems inconsistent with what this helper type is trying to do. 

> If you want silly increases in complexity, you could also pass in a context here and check for ctx.Done()/e.wg.Wait()

If there is some friction in what a func calls then they should have a context anyway like HTTP requests which is safely closed over. I don't see the added bonus of that change.

common/common.go
shazbert: Its not really used so that it can escape, but I guess I can nil it out 🤷 won't hurt things.

common/common.go
gloriousCode: Yeah I already had a change of heart and edited the post before your comment

common/common.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2090/commits/1dfd3d221ef5f938e5204133af99456af0fbc1b1

common/common.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2090/commits/1dfd3d221ef5f938e5204133af99456af0fbc1b1

thrasher-:
@codex please review

common/common.go
```diff
@@ -553,31 +553,34 @@ func ExcludeError(err, excl error) error {
...
-	close(e.C)
-	for err := range e.C {
-		if err != nil {
-			errs = AppendError(errs, err)
+	e.wg.Wait()
+	e.m.Lock()
+	defer e.m.Unlock()
+	out := e.errs
+	e.errs = nil
+	return out
```
gbjk: 🦍 Optional pedantic nitpick:

```suggestion
	defer func(){ e.errs = nil }()
	defer e.m.Unlock()
	return e.errs
```
Alternatively:
```suggestion
	defer func(){
		e.errs = nil
		e.m.Unlock()
	}()
	return e.errs
```

common/common.go
```diff
@@ -553,31 +553,34 @@ func ExcludeError(err, excl error) error {
...
-func CollectErrors(n int) *ErrorCollector {
-	e := &ErrorCollector{
-		C: make(chan error, n),
-	}
-	e.Wg.Add(n)
-	return e
-}
-
-// Collect runs waits for e.Wg to be Done, closes the error channel, and return a error collection
+// Collect waits for the internal wait group to be done and returns an error collection
```
gbjk, line 562: 🐒 Optional nitpick:
I've assumed the point here of niling `e.errs` is to make `e` reusable.
But we don't actually tell the user that in the docs.
So I'd say either we don't do it, or we say we're doing it.
```suggestion
// Collect waits for the internal wait group to be done and returns an error collection
// State is reset after each Collect, so successive calls are okay
```

common/common.go
```diff
@@ -553,31 +553,34 @@ func ExcludeError(err, excl error) error {
...
+	defer e.m.Unlock()
+	out := e.errs
+	e.errs = nil
+	return out
+}
+
+// Go runs a function in a goroutine and collects any error it returns
+func (e *ErrorCollector) Go(f func() error) {
+	if f == nil {
+		panic("ErrorCollector.Go received a nil function")
```
gbjk: 🐒 Optional nitpick:
* This is _kinda_ the same topic as the previous thread here, but my suggestion seems solid to me: Why panic when we have a way to give the user back the error?
* I'm also kinda okay with panicing though.
* And also okay if we don't use NilGuard but still put the result in e.errs
```suggestion
	if err := common.NilGuard(f); err != nil {
			e.m.Lock()
			e.errs = AppendError(e.errs, err)
			e.m.Unlock()
```

gbjk (approved):
Lovely work.

Only optional change requests, so approving.

common/common.go
shazbert: I think its more to do with my preference these days; it reduces complexity and I have a love of panics for when things are structural. e.g. if this errors because you have supplied nil, likely means you need to bring the program down anyway and address this happy little accident. Will leave this open.

gbjk: * 🤷‍♂️ What happens if we don't have this check? Won't it panic anyway? Is it really worth rethrowing the panic differently?

* I'm debating the same. If we allow it to panic, the caller's app has the choices about how to handle fatal structural failures in their code:
  * Let the app die completely
  * Restart the app from environment, probably with backoff handling for fast restarting and spawning
  * Catch the panic inside the app and handle gracefully
 
So by leaving panics coming from external *mis-use* as just that, I think we provide the clearest feedback.
That feels like okay to me.

shazbert: > What happens if we don't have this check? Won't it panic anyway? Is it really worth rethrowing the panic differently? 

Yeah you are right, I only wanted it outside of the go routine below to be a bit clearer for debugging and testing it, but it can be removed if you want, it's what happens when method `Go` is called on `sync.Waitgroup` type?

gloriousCode: I think the only upside of our own panic than the normal panic is that it shows more quickly that it was a result of calling this specific function, despite the stack trace doing the same thing if this panic wasn't here. I don't gracefully recover from panics, so I can't say whether it's easier to recover from our own panic than the normal panic ala `errors.Is`. Would be happy to understand that part more and was what I was trying to ask about in my initial review

edit: needed sync.Mutex

gbjk: Okay. Having seen how it smells when we let the panic happen in the Go func, screw that cartman.
So I vote we use our consistent Nil error handling so they can handle it properly:
```suggestion
	if err := common.NilGuard(f); err != nil {
		panic(err)
```

Smells better:
```
panic: nil pointer: common.CollectorFunc [recovered, repanicked]
goroutine 9 [running]:
        /opt/homebrew/Cellar/go/1.25.1/libexec/src/runtime/panic.go:783 +0x120
github.com/thrasher-corp/gocryptotrader/common.(*ErrorCollector).Go(0xc000103710, 0x0)
        /Users/gbjk/go/src/github.com/gbjk/gocryptotrader/common/common.go:577 +0x190
github.com/thrasher-corp/gocryptotrader/common.TestErrorCollector(0xc000248000)
```

And this gives them a chance to catch the panic properly:
```
	func() {
		defer func() {
			if err := recover(); err != nil {
				t.Log(err)
			}
		}()
		e.Go(nil)
```
Yielding: `● nil pointer: common.CollectorFunc`

I don't think they can do anything about that, really, but it's nice to be nice.

common/common_test.go
```diff
@@ -590,23 +590,22 @@ func TestGenerateRandomString(t *testing.T) {
...
+	var e ErrorCollector
+	require.Panics(t, func() { e.Go(nil) }, "Go with nil function must panic")
 	for i := range 4 {
-		go func() {
+		e.Go(func() error {
 			if i%2 == 0 {
-				e.C <- errors.New("Collected error")
-			} else {
-				e.C <- nil
+				return errors.New("Collected error")
```
gloriousCode:
```suggestion
				return errors.New("collected error")
```
Huge deal, untackable

shazbert: 👍

exchanges/request/request_test.go
```diff
@@ -343,13 +340,9 @@ func TestDoRequest_Retries(t *testing.T) {
 	require.NoError(t, err, "New requester must not error")
 
 	const numReqs = 4
```
thrasher-: Since neqRegs is now only used in the for loop, can move it to there

exchanges/request/request_test.go
```diff
@@ -308,12 +308,9 @@ func TestDoRequest(t *testing.T) {
 
 	// Check client side rate limit
 	const numReqs = 5
```
thrasher-: Same with this

exchanges/request/request_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2090/commits/4f08d17468a228f0911a4bcada4e7efd77187175

exchanges/request/request_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2090/commits/4f08d17468a228f0911a4bcada4e7efd77187175

thrasher- (approved):
Changes ACK, thanks!

PR #2091, opened 2025-10-20
HitBTC: Fix panic in FetchTradablePairs

gbjk:
Fixes panic in engine test for TestGetDefaultConfig

```
❯ go test ./engine/... -run TestGetDefaultConfig
--- FAIL: TestGetDefaultConfigurations (2.01s)
    --- FAIL: TestGetDefaultConfigurations/hitbtc (0.39s)
panic: runtime error: slice bounds out of range [:-1] [recovered, repanicked]
```

The code was using strings.Index to find the quote currency position in the symbol ID, then slicing the ID. When Index returns -1 (not found), it caused a panic with 'slice bounds out of range [:-1]'.

Fixed by using the BaseCurrency and QuoteCurrency fields directly from the Symbol struct, which is the correct approach and what the API provides.

## Type of change

### DIFF
--- a/exchanges/hitbtc/hitbtc_wrapper.go
+++ b/exchanges/hitbtc/hitbtc_wrapper.go
@@ -177,21 +177,22 @@ func (e *Exchange) Setup(exch *config.Exchange) error {
 }
 
 // FetchTradablePairs returns a list of the exchanges tradable pairs
-func (e *Exchange) FetchTradablePairs(ctx context.Context, _ asset.Item) (currency.Pairs, error) {
+func (e *Exchange) FetchTradablePairs(ctx context.Context, a asset.Item) (currency.Pairs, error) {
+	if a != asset.Spot {
+		return nil, fmt.Errorf("%w: %q", asset.ErrNotSupported, a)
+	}
+
 	symbols, err := e.GetSymbolsDetailed(ctx)
 	if err != nil {
 		return nil, err
 	}
 
 	pairs := make([]currency.Pair, len(symbols))
-	for x := range symbols {
-		index := strings.Index(symbols[x].ID, symbols[x].QuoteCurrency)
-		var pair currency.Pair
-		pair, err = currency.NewPairFromStrings(symbols[x].ID[:index], symbols[x].ID[index:])
-		if err != nil {
+	for i, s := range symbols {
+		// s.QuoteCurrency is actually settlement currency, so trim the base currency to get the real quote currency
+		if pairs[i], err = currency.NewPairFromStrings(s.BaseCurrency, strings.TrimPrefix(s.ID, s.BaseCurrency)); err != nil {
 			return nil, err
 		}
-		pairs[x] = pair
 	}
 	return pairs, nil
 }
--- a/exchanges/hitbtc/hitbtc.go
+++ b/exchanges/hitbtc/hitbtc.go
@@ -103,8 +103,8 @@ func (e *Exchange) GetSymbols(ctx context.Context, symbol string) ([]string, err
 
 // GetSymbolsDetailed is the same as above but returns an array of symbols with
 // all their details.
-func (e *Exchange) GetSymbolsDetailed(ctx context.Context) ([]Symbol, error) {
-	var resp []Symbol
+func (e *Exchange) GetSymbolsDetailed(ctx context.Context) ([]*Symbol, error) {
+	var resp []*Symbol
 	return resp, e.SendHTTPRequest(ctx, exchange.RestSpot, apiV2Symbol, &resp)
 }
 
--- a/exchanges/hitbtc/hitbtc_test.go
+++ b/exchanges/hitbtc/hitbtc_test.go
@@ -992,9 +992,14 @@ func TestGetOrderInfo(t *testing.T) {
 
 func TestFetchTradablePairs(t *testing.T) {
 	t.Parallel()
-	if _, err := e.FetchTradablePairs(t.Context(), asset.Spot); err != nil {
-		t.Fatal(err)
-	}
+
+	_, err := e.FetchTradablePairs(t.Context(), asset.Futures)
+	assert.ErrorIs(t, err, asset.ErrNotSupported)
+
+	r, err := e.FetchTradablePairs(t.Context(), asset.Spot)
+	require.NoError(t, err)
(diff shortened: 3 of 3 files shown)


thrasher-:
@codex please review

exchanges/hitbtc/hitbtc_wrapper.go
```diff
@@ -185,9 +184,8 @@ func (e *Exchange) FetchTradablePairs(ctx context.Context, _ asset.Item) (curren
 
 	pairs := make([]currency.Pair, len(symbols))
 	for x := range symbols {
-		index := strings.Index(symbols[x].ID, symbols[x].QuoteCurrency)
 		var pair currency.Pair
-		pair, err = currency.NewPairFromStrings(symbols[x].ID[:index], symbols[x].ID[index:])
+		pair, err = currency.NewPairFromStrings(symbols[x].BaseCurrency, symbols[x].QuoteCurrency)
```
shazbert: So what was causing this was:

```
{ID:COAIUSDT BaseCurrency:COAI QuoteCurrency:BTC QuantityIncrement:0.001 TickSize:0.001 TakeLiquidityRate:0.0025 ProvideLiquidityRate:0.0012 FeeCurrency:BTC}
```

Which now converts COAIUSDT -> COAI-BTC which is an invalid symbol. QuoteCurrency looks to be a settlement currency? So there are a lot of pairs now converting quote from USDT -> USD. Which is also invalid.

So I suggest instead of using quote currency we roll back and use Base currency in our index and then handle the potential -1 issue. 

```go
index := strings.Index(symbols[x].ID, symbols[x].BaseCurrency)
```

suggestion: Also a driveby here would be to handle the param asset type to ensure that asset.Spot was passed in.

shazbert (changes requested):
Thanks for PR, one issue I found with the update.

exchanges/hitbtc/hitbtc_wrapper.go
gbjk: Thanks for the suggestions.
I don't think there's any value on indexing the symbol by the base currency... it should be the first character, right?
But we can just use trim prefix.
Added a test to make sure we get sanity back out.
Wanted to make sure that the USDT pairs are coming out correctly, but couldn't see a way to do that and ... this'll do.

Fixed https://github.com/gbjk/gocryptotrader/commit/ac5f4336a and https://github.com/gbjk/gocryptotrader/commit/9b50bea1f

samuael (commented):
del

exchanges/hitbtc/hitbtc_wrapper.go
```diff
@@ -177,21 +177,23 @@ func (e *Exchange) Setup(exch *config.Exchange) error {
...
 	}
 
 	pairs := make([]currency.Pair, len(symbols))
-	for x := range symbols {
-		index := strings.Index(symbols[x].ID, symbols[x].QuoteCurrency)
-		var pair currency.Pair
-		pair, err = currency.NewPairFromStrings(symbols[x].ID[:index], symbols[x].ID[index:])
+	for i, s := range symbols {
+		// s.QuoteCurrency is actually settlement currency, so trim the base to get the real Quote
+		pair, err := currency.NewPairFromStrings(s.BaseCurrency, strings.TrimPrefix(s.ID, s.BaseCurrency))
```
samuael:
```suggestion
		pairs[i], err = currency.NewPairFromStrings(s.BaseCurrency, strings.TrimPrefix(s.ID, s.BaseCurrency))
```

exchanges/hitbtc/hitbtc_wrapper.go
```diff
@@ -177,21 +177,23 @@ func (e *Exchange) Setup(exch *config.Exchange) error {
 }
 
 // FetchTradablePairs returns a list of the exchanges tradable pairs
-func (e *Exchange) FetchTradablePairs(ctx context.Context, _ asset.Item) (currency.Pairs, error) {
+func (e *Exchange) FetchTradablePairs(ctx context.Context, a asset.Item) (currency.Pairs, error) {
+	if a != asset.Spot {
+		return nil, fmt.Errorf("%w: %q", asset.ErrNotSupported, a)
+	}
```
thrasher-, line 183: Please add a newline here just for clarity as everything else is grouped by purpose

exchanges/hitbtc/hitbtc_wrapper.go
```diff
@@ -177,21 +177,23 @@ func (e *Exchange) Setup(exch *config.Exchange) error {
...
 		return nil, err
 	}
 
 	pairs := make([]currency.Pair, len(symbols))
-	for x := range symbols {
-		index := strings.Index(symbols[x].ID, symbols[x].QuoteCurrency)
-		var pair currency.Pair
-		pair, err = currency.NewPairFromStrings(symbols[x].ID[:index], symbols[x].ID[index:])
+	for i, s := range symbols {
+		// s.QuoteCurrency is actually settlement currency, so trim the base to get the real Quote
```
thrasher-:
```suggestion
		// s.QuoteCurrency is actually settlement currency, so trim the base currency to get the real quote currency
```

exchanges/hitbtc/hitbtc_wrapper.go
gbjk: Agreed. Also my style.

Fixed https://github.com/gbjk/gocryptotrader/commit/e3388a00e

🏀 Slam dunk; Resolving.

exchanges/hitbtc/hitbtc_wrapper.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/645387110

🏀 Slam dunk; Resolving.

gbjk:
🔂  Squashed, since last changes were no-ops

exchanges/hitbtc/hitbtc_wrapper.go
gbjk: Yeah, that's cute, go-on-then.
I've assumed you intended to suggest removing line 5.
Also folding the err condition down.

Fixed https://github.com/gbjk/gocryptotrader/commit/1a4702667

Rebasing again, so this commit will be orphaned but you can still view it

PR #2092, opened 2025-10-21
exchanges: Use singular futures settlement currency

gloriousCode:
# PR Description
Minor change to simplify using settlement currencies. Only BTSE had multiple settlement currencies and it is an [auto-convert-behind-the-scenes-using-value-scaling](https://support.btse.com/en/support/solutions/articles/43000732282-multi-asset-settlement) feature derived from USDT. I've added a field just for BTSE to use.

## How has this been tested
- [x] go test ./... -race
- [x] golangci-lint run

### DIFF
--- a/exchanges/bitmex/bitmex.go
+++ b/exchanges/bitmex/bitmex.go
@@ -93,7 +93,7 @@ const (
 	bitmexEndpointUserRequestWithdraw   = "/user/requestWithdrawal"
 
 	constSatoshiBTC = 1e-08
-	countLimit      = int32(1000)
+	countLimit      = uint32(1000)
 
 	perpetualContractID         = "FFWCSX"
 	spotID                      = "IFXXXP"
--- a/exchanges/bitmex/bitmex_wrapper.go
+++ b/exchanges/bitmex/bitmex_wrapper.go
@@ -605,7 +605,7 @@ allTrades:
 				ts = tradeData[i].Timestamp
 			}
 		}
-		if int32(len(tradeData)) != countLimit {
+		if len(tradeData) != int(countLimit) {
 			break allTrades
 		}
 	}
--- a/exchanges/bitmex/bitmex_parameters.go
+++ b/exchanges/bitmex/bitmex_parameters.go
@@ -186,7 +186,7 @@ type GenericRequestParams struct {
 	Columns string `json:"columns,omitempty"`
 
 	// Count - Number of results to fetch.
-	Count int32 `json:"count,omitempty"`
+	Count uint32 `json:"count,omitempty"`
 
 	// EndTime - Ending date filter for results.
 	EndTime string `json:"endTime,omitempty"`


thrasher-:
@codex please review

exchanges/bitmex/bitmex_wrapper.go
```diff
@@ -1010,28 +1009,22 @@ func (e *Exchange) GetFuturesContractDetails(ctx context.Context, item asset.Ite
...
-		return nil, err
-	}
-
-	resp := make([]futures.Contract, 0, len(marketInfo))
+	var resp []futures.Contract
 	switch item {
 	case asset.PerpetualContract:
+		marketInfo, err := e.GetInstruments(ctx, &GenericRequestParams{
+			Count:  1000,
+			Filter: `{"typ": "` + perpetualContractID + `"}`,
```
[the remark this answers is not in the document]
gloriousCode: Lies:
```
{
  "error": {
    "message": "Maximum result count is 1000. Please use the start & count params to paginate.",
    "name": "ValidationError"
  }
}
```
source: https://www.bitmex.com/api/explorer/#!/Instrument/Instrument_get

exchanges/btse/btse_wrapper.go
```diff
@@ -1129,16 +1129,17 @@ func (e *Exchange) GetFuturesContractDetails(ctx context.Context, item asset.Ite
...
-			Asset:                item,
-			SettlementCurrencies: settlementCurrencies,
-			StartDate:            startTime,
-			EndDate:              endTime,
-			SettlementType:       contractSettlementType,
-			IsActive:             marketSummary[i].Active,
-			Type:                 ct,
+			Exchange:                       e.Name,
+			Name:                           cp,
+			Underlying:                     currency.NewPair(currency.NewCode(marketSummary[i].Base), currency.NewCode(marketSummary[i].Quote)),
```
shazbert: suggestion/oos: Base and Quote could be currency.Code types.

exchanges/bitmex/bitmex_wrapper.go
```diff
@@ -1049,37 +1042,38 @@ func (e *Exchange) GetFuturesContractDetails(ctx context.Context, item asset.Ite
...
-			cp, err = currency.NewPairFromStrings(marketInfo[x].RootSymbol, marketInfo[x].Symbol[len(marketInfo[x].RootSymbol):])
-			if err != nil {
-				return nil, err
-			}
-			underlying, err = currency.NewPairFromStrings(marketInfo[x].RootSymbol, marketInfo[x].SettlCurrency)
+			cp, err := currency.NewPairFromStrings(marketInfo[x].RootSymbol, marketInfo[x].Symbol[len(marketInfo[x].RootSymbol):])
 			if err != nil {
 				return nil, err
 			}
+			underlying := currency.NewPair(cp.Base, marketInfo[x].SettlementCurrency)
```
shazbert: I think this could be moved to the field below instead of making a var.

exchanges/futures/contract.go
```diff
@@ -30,13 +30,14 @@ type Contract struct {
...
 	// Optional values if the exchange offers them
-	SettlementCurrencies currency.Currencies
-	MarginCurrency       currency.Code
-	Multiplier           float64
-	MaxLeverage          float64
-	LatestRate           fundingrate.Rate
-	FundingRateFloor     decimal.Decimal
-	FundingRateCeiling   decimal.Decimal
+	SettlementCurrency             currency.Code
+	AdditionalSettlementCurrencies currency.Currencies
```
shazbert, line 34: 🍻 Really should change `Currencies` -> `Codes` but not here.

exchanges/btse/btse_wrapper.go
gloriousCode: Main reason against is this line:
```
cp, err := currency.NewPairFromString(marketSummary[i].Base, marketSummary[i].Symbol[len(marketSummary[i].Base):])
```
Feels more convoluted with
```
cp := currency.NewPair(marketSummary[i].Base, currency.NewCode(marketSummary[i].Symbol[len(marketSummary[i].Base.String()):]))
```

There may be a better way once I sleep more

exchanges/btse/btse_wrapper.go
```diff
@@ -1077,77 +1077,54 @@ func (e *Exchange) GetMarketSummary(ctx context.Context, symbol string, spot boo
...
 	marketSummary, err := e.GetMarketSummary(ctx, "", false)
 	if err != nil {
 		return nil, err
 	}
-	resp := make([]futures.Contract, 0, len(marketSummary))
+	resp := make([]futures.Contract, len(marketSummary))
 	for i := range marketSummary {
-		var cp currency.Pair
-		cp, err = currency.NewPairFromStrings(marketSummary[i].Base, marketSummary[i].Symbol[len(marketSummary[i].Base):])
+		cp, err := currency.NewPairFromStrings(marketSummary[i].Base, marketSummary[i].Symbol[len(marketSummary[i].Base):])
```
gbjk: 🐒 Optional nitpick:
I think that this is too painful to do inline and without explaining that .Quote currency is the settlement currency, and do something more like:
```suggestion
		// .Quote is the settlement currency; Get the real quote currency from Symbol
		quoteCurr := marketSummary[i].Symbol[len(marketSummary[i].Base):]
		cp, err := currency.NewPairFromStrings(marketSummary[i].Base, quote)
```

exchanges/btse/btse_wrapper.go
gbjk: I've asked for that line to be changed anyway.
I'd be supportive of using currency.Code for both.

exchanges/futures/contract.go
gbjk: I don't understand why - It's a settlement currency, not a settlement code ?

gbjk (approved):
Good work; Only minor change requests. Substantially approved.

exchanges/futures/contract.go
shazbert: @gbjk I meant the type `currency.Currencies` -> `currency.Codes`

gloriousCode:
I had to do a few extra bits and bobs to make my changes and requested changes work. So please re-looky.

currency/currencies.go
```diff
@@ -55,6 +55,15 @@ func (c Currencies) Join() string {
 
 // UnmarshalJSON conforms type to the umarshaler interface
 func (c *Currencies) UnmarshalJSON(d []byte) error {
+	if d[0] == '[' {
+		var currs []Code
+		err := json.Unmarshal(d, &currs)
+		if err != nil {
```
gbjk: 🐒 Optional nitpick:

```suggestion
		if err := json.Unmarshal(d, &currs); err != nil {
```

exchanges/btse/btse_wrapper.go
```diff
@@ -1018,27 +1018,27 @@ func (m *MarketPair) StripExponent() (string, error) {
...
+			baseCurr = currency.TRUMP
+			quoteStr = strings.TrimPrefix(m.Symbol, baseCurr.String())
 		} else {
-			s := strings.Split(m.Symbol, m.Base) // e.g. RUNEPFC for RUNE-USD futures pair
+			s := strings.Split(m.Symbol, m.Base.String()) // e.g. RUNEPFC for RUNE-USD futures pair
 			if len(s) <= 1 {
 				return currency.EMPTYPAIR, errInvalidPairSymbol
 			}
-			quoteCurr = s[1]
+			quoteStr = s[1]
```
gbjk: 🤷‍♂️ Clarification please:
Example doesn't seem to match what we're doing here.
`quoteStr` will be PFC surely?

exchanges/btse/btse_wrapper.go
```diff
@@ -1086,10 +1086,9 @@ func (e *Exchange) GetFuturesContractDetails(ctx context.Context, item asset.Ite
 	}
 	resp := make([]futures.Contract, len(marketSummary))
 	for i := range marketSummary {
-		cp, err := currency.NewPairFromStrings(marketSummary[i].Base, marketSummary[i].Symbol[len(marketSummary[i].Base):])
-		if err != nil {
-			return nil, err
-		}
+		// quote field is the settlement currency, create the quote from the symbol
```
gbjk: 🐒 Optional nitpick:
* On a recent PR thrasher corrected `quote` to `quote currency`.
So that's why my suggestion used the phrasing it did
* I don't agree with `quote field` because the field is public - Hence the suggestion of `Quote`
```suggestion
		// Quote field is the settlement currency, create the quote currency from the symbol
```

gbjk (changes requested):
Thanks.

Just need to resolve the `RUNEPFC` - either comment feels misleading or code is wrong.

exchanges/btse/btse_wrapper.go
gloriousCode: I've reused the other comment I added about the Quote field. You're right that the original comment has ambiguity

gbjk (approved):

exchanges/btse/btse_test.go
```diff
@@ -642,10 +641,7 @@ func TestWsUnexpectedData(t *testing.T) {
 
 func TestGetFuturesContractDetails(t *testing.T) {
 	t.Parallel()
-	_, err := e.GetFuturesContractDetails(t.Context(), asset.Spot)
-	assert.ErrorIs(t, err, futures.ErrNotFuturesAsset, "GetFuturesContractDetails should error correctly on Spot")
-
-	_, err = e.GetFuturesContractDetails(t.Context(), asset.USDTMarginedFutures)
+	_, err := e.GetFuturesContractDetails(t.Context(), asset.USDTMarginedFutures)
```
thrasher-:
```suggestion
	_, err := e.GetFuturesContractDetails(t.Context(), asset.Margin)
```
So the comment below matches 👀

exchanges/bitmex/bitmex_wrapper.go
```diff
@@ -558,10 +558,9 @@ func (e *Exchange) GetHistoricTrades(ctx context.Context, p currency.Pair, asset
 	if err != nil {
 		return nil, err
 	}
-	limit := 1000
 	req := &GenericRequestParams{
 		Symbol:  p.String(),
-		Count:   int32(limit),
+		Count:   int32(1000),
```
thrasher-: Can make this a const countLimit = 1000 since it's used as a magic number in 4 different areas

currency/currencies.go
```diff
@@ -55,6 +55,14 @@ func (c Currencies) Join() string {
 
 // UnmarshalJSON conforms type to the umarshaler interface
 func (c *Currencies) UnmarshalJSON(d []byte) error {
+	if d[0] == '[' {
+		var currs []Code
+		if err := json.Unmarshal(d, &currs); err != nil {
+			return err
+		}
+		*c = currs
+		return nil
```
shazbert:
```suggestion
		return json.Unmarshal(d, (*[]Code)(c))
```

You can break recursion without the allocation here

shazbert:
```go
// UnmarshalJSON conforms type to the umarshaler interface
func (c *Currencies) UnmarshalJSON(d []byte) error {
	if d[0] != '[' {
		d = []byte(`[` + string(d) + `]`)
	}
	return json.Unmarshal(d, (*[]Code)(c))
}
```

Can collapse the entire thing if ya want

gloriousCode: You're out of control! I'm surprised at the result
```
// GloriousButts
// BenchmarkButts-10    	  513199	      2344 ns/op	    1448 B/op	      29 allocs/op
// ShazButts
// BenchmarkButts-10    	  568749	      2094 ns/op	    1184 B/op	      24 allocs/op
func BenchmarkButts(b *testing.B) {
	var unmarshalHere Currencies
	j := []byte(`["btc","usd","ltc","bro","things"]`)
	for range b.N {
		_ = json.Unmarshal(j, &unmarshalHere)
	}
}
```

gbjk: I'd marry that code.

exchanges/bitmex/bitmex.go
```diff
@@ -96,15 +93,7 @@ const (
...
-
-	// ContractPerpetual perpetual contract type
-	ContractPerpetual = iota
-	// ContractFutures futures contract type
-	ContractFutures
-	// ContractDownsideProfit downside profit contract type
-	ContractDownsideProfit
-	// ContractUpsideProfit upside profit contract type
-	ContractUpsideProfit
+	countLimit      = int32(1000)
```
shazbert: suggestion: maybe a uint instead to stop the gosec linter issue

PR #2093, opened 2025-10-26
GateIO: Subscription templating for all assets

gbjk:
This PR is aimed primarily at fixing the asset templating for GateIO.
Along the way it's also simplified generateSubs.
Some things are small steps towards simplifying the multi-connection handling, and removing a lot of the boilerplating we see duplicated, which will come in the next PR.

* GateIO: Subscription templating for all assets
* Bybit: Fix messageFilter types
* Websocket: Filter subs by connection
* Binance: Move sub asset type assertion to config version
* Subscriptions:
 * Abstracted default GenerateSubs
 * Move subscriptions to ws Manager
 * Abstracted default GenerateSubs
* Exchanges:
  * Context for ErrFunctionNotSupported error
  * `e.SubscribeToWebsocketChannels` 👉 `ws.SubscribeToChannels()`
  * `e.UnsubscribeToWebsocketChannels` 👉 `ws.UnsubscribeChannels()`
* Docs: Updates to coding guidelines aimed at better AI tests

## Type of change


thrasher-:
@codex please review

exchange/websocket/manager.go
```diff
@@ -472,74 +479,65 @@ func (m *Manager) connect() error {
...
-					log.Warnf(log.WebsocketMgr, "%s websocket: no subscriptions generated", m.exchangeName)
-				}
-				continue
+		if len(connSubs) == 0 {
+			// If no subscriptions are generated, we skip the connection
+			if m.verbose {
+				log.Warnf(log.WebsocketMgr, "%s websocket: no subscriptions generated [conn:%d] [URL:%s]", m.exchangeName, i+1, cW.setup.URL)
 			}
+			continue
 		}
```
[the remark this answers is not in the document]
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/34494cef4

exchange/websocket/manager.go
```diff
@@ -472,102 +479,91 @@ func (m *Manager) connect() error {
...
 
-		m.connections[conn] = m.connectionManager[i]
-		m.connectionManager[i].connection = conn
+		m.connections[conn] = cW
+		cW.connection = conn
 
 		m.Wg.Add(1)
-		go m.Reader(context.TODO(), conn, m.connectionManager[i].setup.Handler)
+		go m.Reader(context.TODO(), conn, cW.setup.Handler)
```
shazbert: can use `m.Wg.Go()`

exchange/websocket/manager_test.go
```diff
@@ -239,64 +243,54 @@ func TestConnectionMessageErrors(t *testing.T) {
...
 	err = ws.Connect()
-	require.ErrorIs(t, err, errDastardlyReason)
-
-	ws.connectionManager[0].setup.Handler = func(context.Context, Connection, []byte) error {
-		return errDastardlyReason
-	}
-	err = ws.Connect()
-	require.ErrorIs(t, err, errDastardlyReason)
+	require.ErrorContains(t, err, "setup.Subscriber error")
+	require.NoError(t, ws.shutdown()) // Subscription errors do not disconnect
```
shazbert: comment should be in `NoError` call

exchange/websocket/manager_test.go
```diff
@@ -1036,11 +1026,11 @@ func TestSetupNewConnection(t *testing.T) {
 	require.ErrorIs(t, err, errWebsocketDataHandlerUnset)
 
 	connSetup.Handler = func(context.Context, Connection, []byte) error { return nil }
-	connSetup.MessageFilter = []string{"slices are super naughty and not comparable"}
+	connSetup.MessageFilter = naughtyFilter{"slices are super naughty and not comparable"}
 	err = multi.SetupNewConnection(connSetup)
 	require.ErrorIs(t, err, errMessageFilterNotComparable)
```
shazbert: I think this error may not be required and can be removed, because we don't use this as an immutable map lookup key anymore.

exchanges/exchange.go
```diff
@@ -175,14 +175,17 @@ func (b *Base) SetFeatureDefaults() {
 func (b *Base) SetSubscriptionsFromConfig() {
 	b.settingsMutex.Lock()
 	defer b.settingsMutex.Unlock()
+	if b.Websocket == nil {
+		return
```
shazbert, line 177: This looks like we can return an error. As we shouldn't call this without the websocket being setup.

shazbert (changes requested):
Just the most basic of basic reviews while I get my head around changes

exchange/websocket/manager.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/019b60b5f

exchange/websocket/manager_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/08357260a

exchange/websocket/manager_test.go
gbjk: I don't know; We still use `GetConnection` and expect to get just one connection back from it in Bybit and GateIO.
We still expect to be able to compare the message filters when we're adding them.
And right now all the message filters are easily compared.

So I think it can stay. Am I missing something?

exchanges/exchange.go
gbjk: TL;DR:
Fixed https://github.com/gbjk/gocryptotrader/commit/9fa693ae9

But you might ask me to revert based on blast radius and discussion below.

🥃  Yeah, alright I guess....
Feels like overkill because it's only called from `SetFeatureDefaults` (which would have to have error return added), which is only called from `SetupDefaults`.
And also "we shouldn't call this without websocket being setup" is the same as it returning if called with a nil Websocket. Moving it to an if condition in the caller is ... just the same thing.
So this didn't feel wrong.

exchange/websocket/manager.go
```diff
@@ -472,102 +479,90 @@ func (m *Manager) connect() error {
...
 		}
 
-		m.connections[conn] = m.connectionManager[i]
-		m.connectionManager[i].connection = conn
+		m.connections[conn] = cW
+		cW.connection = conn
 
-		m.Wg.Add(1)
-		go m.Reader(context.TODO(), conn, m.connectionManager[i].setup.Handler)
+		m.Wg.Go(func() { m.Reader(context.TODO(), conn, cW.setup.Handler) })
```
shazbert: Need to rm the `defer m.Wg.Done()` in m.Reader or it will panic at the disco.

exchange/websocket/manager_test.go
shazbert: It implements message filter already or it's nil, so regardless of implementation outside of this I don't think it really matters doing the check at all. Keeping the check in will stop the ability to define a struct with a cache map or slice field for whatever reason, if there are some weird future connection permutations. 

But if we ever come across the need for that functionality this can be deferred to later, I will leave it up to you if you want to remove it.

shazbert (changes requested):
Just on recent changes

exchange/websocket/connection.go
```diff
@@ -473,3 +474,29 @@ func (c *connection) RequireMatchWithData(signature any, incoming []byte) error
 func (c *connection) IncomingWithData(signature any, data []byte) bool {
 	return c.Match.IncomingWithData(signature, data)
 }
+
+// MessageFilter defines the interface for types to use in ConnectionSetup to match messages and subscriptions
+// TODO: This isn't really a message filter, and this naming will be fixed in an upcoming PR
+type MessageFilter interface {
```
gloriousCode: Why not now?

exchange/websocket/manager.go
```diff
@@ -155,6 +158,11 @@ type ManagerSetup struct {
 	RateLimitDefinitions request.RateLimitDefinitions
 }
 
+// IExchange is the interface exchanges must implement to use the manager
+type IExchange interface {
+	subscription.IExchange
+}
```
gloriousCode, line 161: Plannig expansion? I'm not seeing a reason not to just be using `subscription.IExchange` instead. Or just internalise the function signatures here if concern is about dependency

exchange/websocket/manager.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/a88a4427a

Though now I'm eyeing `monitorFrame` with suspicion.
How would you feel about [this](https://github.com/gbjk/gocryptotrader/commit/5e1e8e70e60073c1b130ee8e29da1b92b75563c9) ?
If you're happy with the simplification of your pattern, I'll fix the tests as well.

exchange/websocket/manager_test.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/afc95acbe

Note: I'm seeing a few tests hang after this, but I don't know if I just broke it.
I'll sniff it shortly

exchange/websocket/manager.go
gbjk: I've fixed up the tests for it and I'm pretty happy, so I've pushed it to the branch too.

Fixed https://github.com/gbjk/gocryptotrader/commit/be82ea61c

exchanges/exchange.go
gbjk: @shazbert Unresolved this thread because it broke engine tests for exchanges that don't support websockets.

I think that the best phrasing is "we shouldn't do anything if the websocket isn't setup".
I don't agree that means "we shouldn't call this", because I like to avoid action at a distance;
I didn't want to wrap calls to this in an `if b.Websocket != nil` and then return an error if we call with nil even though we only have one internal caller.

So I've done kinda a manual revert on it.
I've moved this to being a private function, and unwound the error stacks we created for this.
It'll return if it's called without a websocket, and that's fine IMO.
It's private, and it's called from only one place.

At the same time I spotted that the original (and assertified) tests about feature defaults weren't testing the right thing [here](https://github.com/gbjk/gocryptotrader/commit/9fa693ae9#diff-e35cc6b92d77e3da9f99162a5679baf42e387a496d37416413333729a80c1bb6L247-L251) because what we're really trying to see (I think) is that the config is updated from features in those situations.

Fixed https://github.com/gbjk/gocryptotrader/commit/9932ebb38

config/versions/v11/v11.go
```diff
@@ -0,0 +1,53 @@
...
+			if assetType == "" {
+				if sub, err = jsonparser.Set(sub, []byte(`"spot"`), "asset"); err != nil {
+					errs = common.AppendError(errs, err)
+					return
+				}
+			}
+		}
+		newSubs = append(newSubs, sub)
+	}
+	_, err := jsonparser.ArrayEach(e, subsFn, "features", "subscriptions")
```
samuael, line 39: This can be written in a short form, right?

```suggestion
	if _, err := jsonparser.ArrayEach(e, subsFn, "features", "subscriptions"); err != nil {
```

exchange/websocket/subscriptions.go
```diff
@@ -23,6 +23,16 @@ var (
 	errSubscriptionsExceedsLimit = errors.New("subscriptions exceeds limit")
 )
 
+// GenerateSubscriptions fans out subscriptions using ExpandTemplates
+func (m *Manager) GenerateSubscriptions() (subscription.List, error) {
+	if err := common.NilGuard(m); err != nil {
+		return nil, err
+	} else if err := common.NilGuard(m.Exchange); err != nil {
```
samuael, line 30:
```suggestion
	} else if ommon.NilGuard(m.Exchange) != nil {
```

exchange/websocket/subscriptions_test.go
```diff
@@ -47,8 +48,8 @@ func TestSubscribeUnsubscribe(t *testing.T) {
 	assert.NoError(t, ws.SubscribeToChannels(nil, nil), "Subscribe to an nil List should not error")
 	assert.NoError(t, ws.UnsubscribeChannels(nil, subs), "Unsubscribing should not error")
 
-	ws.Subscriber = func(subscription.List) error { return errDastardlyReason }
-	assert.ErrorIs(t, ws.SubscribeToChannels(nil, subs), errDastardlyReason, "Should error correctly when error returned from Subscriber")
+	ws.Subscriber = func(subscription.List) error { return errors.New("Subscriber error") }
+	assert.ErrorContains(t, ws.SubscribeToChannels(nil, subs), "Subscriber error", "Should error correctly when error returned from Subscriber")
```
samuael, line 91:
```suggestion
	assert.ErrorContains(t, ws.SubscribeToChannels(nil, subs), "Subscriber error Should error correctly when error returned from Subscriber")
```

samuael (commented):
a quick review on subscriptions.go

config/versions/v11/v11.go
gbjk: * I'm confused by your suggestion. You realise you're dropping `&& !errors.Is(err, jsonparser.KeyPathNotFoundError) { ` ?
* And that's the reason it's not short form. It felt like it'd just become too long
* While it *could* be written like that, it wouldn't save anything, because I still need an `err` var for line 44 to work.
If applied, your suggestion would look like this:
```
@@ -36,10 +36,10 @@ func (v *Version) UpgradeExchange(_ context.Context, e []byte) ([]byte, error) {
 		}
 		newSubs = append(newSubs, sub)
 	}
-	_, err := jsonparser.ArrayEach(e, subsFn, "features", "subscriptions")
-	if err != nil && !errors.Is(err, jsonparser.KeyPathNotFoundError) {
+	if _, err := jsonparser.ArrayEach(e, subsFn, "features", "subscriptions"); err != nil && !errors.Is(err, jsonparser.KeyPathNotFoundError) {
 		return e, fmt.Errorf("error upgrading subscription assets: %w", err)
 	}
+	var err error
 	for i, s := range newSubs {
 		e, err = jsonparser.Set(e, s, "features", "subscriptions", "["+strconv.Itoa(i)+"]")
```

That's why I haven't done it. It doesn't save anything and makes the code less readable.

I'm going to assume this answer is acceptable and resolve this thread unilaterally.

exchange/websocket/subscriptions.go
gbjk: 👎 🤷‍♂️ But ... we're *using* `err` in the next line.
Also prefer to not have to make this change:
```
@@ -7,6 +7,7 @@ import (
 	"slices"
 
+	ommon "github.com/thrasher-corp/gocryptotrader/common"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/subscription"
 )
```

Unilaterally resolving this 🙂

exchange/websocket/subscriptions_test.go
gbjk: 👎 Sorry; This suggestion doesn't make sense. `Subsciber error` is the string we are looking for, and `Should error correctly when error returned from Subscriber` is the message we'll show when the test fails.

Unilaterally resolving this 🙂

exchange/websocket/manager.go
```diff
@@ -1083,3 +1031,29 @@ func (m *Manager) GetConnection(messageFilter any) (Connection, error) {
 
 	return nil, fmt.Errorf("%s: %w associated with message filter: '%v'", m.exchangeName, ErrRequestRouteNotFound, messageFilter)
 }
+
+func (w *connectionWrapper) generateSubscriptions(managerSubs subscription.List) (connSubs subscription.List, err error) {
```
gloriousCode: I think this is an improvement to the func
```go
func (w *connectionWrapper) generateSubscriptions(managerSubs subscription.List) (subscription.List, error) {
	if w.setup.GenerateSubscriptions != nil {
		return w.setup.GenerateSubscriptions()
	}
	connSubs := make(subscription.List, 0, len(managerSubs))
	f := w.setup.MessageFilter
	if f == nil {
		return nil, errNoMessageFilter
	}
	for _, s := range managerSubs {
		if f.MatchesSub(s) {
			connSubs = append(connSubs, s)
		}
	}
	return connSubs, nil
}
```

- Exits sooner
- Reduces indentation of secondary path
- Doesn't rely on named return vars (not inherently bad)
- Allocates the response slice

config/versions/v11/v11.go
samuael, line 39:
```suggestion
	if _, err := jsonparser.ArrayEach(e, subsFn, "features", "subscriptions"); err != nil && !errors.Is(err, jsonparser.KeyPathNotFoundError) {
```

exchange/websocket/manager.go
```diff
@@ -905,140 +881,112 @@ func drain(ch <-chan error) {
...
-			if *dropped != 0 {
-				log.Infof(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer recovered; %d messages were dropped", m.exchangeName, dropped)
-				*dropped = 0
-			}
-		default:
-			if *dropped == 0 {
-				// If this becomes prone to flapping we could drain the buffer, but that's extreme and we'd like to avoid it if possible
-				log.Warnf(log.WebsocketMgr, "%s exchange websocket ToRoutine channel buffer full; dropping messages", m.exchangeName)
+		case <-m.ShutdownC:
+			done = true
```
samuael: You can just return

internal/testing/subscriptions/subscriptions.go
```diff
@@ -13,7 +13,7 @@ import (
 // EqualLists is a utility function to compare subscription lists and show a pretty failure message
 // It overcomes the verbose depth of assert.ElementsMatch spewConfig
 // Duplicate of exchange/subscription/subscription:equalList
-func EqualLists(tb testing.TB, a, b subscription.List) bool {
+func EqualLists(tb testing.TB, a, b subscription.List, msg ...string) bool {
```
samuael, line 16: You could just pass a single string and check if it is not an empty string (`""`) instead of a length check.

samuael (commented):
This reduces significant unnecessary code and makes it clear. 🫡 
I couldn't find many issues for now, but I've left 3 comments for now.

config/versions/v11/v11.go
```diff
@@ -0,0 +1,53 @@
...
+	"strconv"
+
+	"github.com/buger/jsonparser"
+	"github.com/thrasher-corp/gocryptotrader/common"
+)
+
+// Version is an ExchangeVersion to ensure all binance subscriptions have asset set
+type Version struct{}
+
+// Exchanges returns just Bitmex
```
gloriousCode:
```suggestion
// Exchanges returns just Binance.
```

config/versions/v11/v11.go
```diff
@@ -0,0 +1,53 @@
...
+func (v *Version) Exchanges() []string { return []string{"Binance"} }
+
+// UpgradeExchange adds asset to all fluffins
```
gloriousCode: Improve comment please

config/versions/v11/v11.go
```diff
@@ -0,0 +1,53 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/buger/jsonparser"
+	"github.com/thrasher-corp/gocryptotrader/common"
+)
+
+// Version is an ExchangeVersion to ensure all binance subscriptions have asset set
```
gloriousCode:
```suggestion
// Version is an ExchangeVersion to ensure that any Binance subscriptions which do not have an asset set, are set to spot.
```

config/versions/v11/v11_test.go
```diff
@@ -0,0 +1,42 @@
...
+	assert.NoError(t, err, "UpgradeExchange should not error with no subscriptions")
+
+	in := []byte(`{"name":"Binance","features":{"subscriptions":[{"channel":"1","asset":4}]}}`)
+	_, err = new(v11.Version).UpgradeExchange(t.Context(), in)
+	assert.ErrorContains(t, err, "Value is not a string: 4", "UpgradeExchange should error correctly")
+
+	in = []byte(`{"name":"Binance","features":{"subscriptions":[{"channel":"1"},{"channel":"2","asset":"all"}]}}`)
+	out, err := new(v11.Version).UpgradeExchange(t.Context(), in)
+	require.NoError(t, err)
+
```
gloriousCode, line 31: Feels like `exp` should be tested here to ensure things worked. Then retested like you do below to ensure that things stay the same

exchange/websocket/connection.go
```diff
@@ -473,3 +474,29 @@ func (c *connection) RequireMatchWithData(signature any, incoming []byte) error
...
+	MatchesSub(*subscription.Subscription) bool
+}
+
+// AssetFilter is a MessageFilter for filtering connections by Asset
+type AssetFilter asset.Item
+
+var _ MessageFilter = AssetFilter(asset.Spot)
```
gloriousCode: I think these are still worth leaving the comment saying its there for ensuring interface implementation. Its a very easy line to snip with no repercussions

exchange/websocket/manager.go
```diff
@@ -206,6 +211,9 @@ func (m *Manager) Setup(s *ManagerSetup) error {
 	if s.Features == nil {
 		return fmt.Errorf("%w: ManagerSetup.Features", common.ErrNilPointer)
 	}
+	if s.Exchange == nil {
+		return fmt.Errorf("%w: ManagerSetup.Exchange", common.ErrNilPointer)
+	}
```
gloriousCode, line 219:
```suggestion
	if err := common.NilGuard(s.ExchangeConfig.Features , s.Features, s.Exchange); err != nil {
		return err
	}
```
Low priority

exchange/websocket/manager.go
```diff
@@ -334,11 +340,8 @@ func (m *Manager) SetupNewConnection(c *ConnectionSetup) error {
 		if c.Connector == nil {
 			return fmt.Errorf("%w: %w", errConnSetup, errWebsocketConnectorUnset)
 		}
-		if c.GenerateSubscriptions == nil && !c.SubscriptionsNotRequired {
-			return fmt.Errorf("%w: %w", errConnSetup, errWebsocketSubscriptionsGeneratorUnset)
-		}
 		if c.Subscriber == nil && !c.SubscriptionsNotRequired {
-			return fmt.Errorf("%w: %w", errConnSetup, errWebsocketSubscriberUnset)
+			return fmt.Errorf("%w (%s): %w", errConnSetup, c.URL, errWebsocketSubscriberUnset)
```
gloriousCode:
```suggestion
			return fmt.Errorf("%w %q: %w", errConnSetup, c.URL, errWebsocketSubscriberUnset)
```

exchange/websocket/manager.go
```diff
@@ -472,102 +462,90 @@ func (m *Manager) connect() error {
...
 
 	// TODO: Implement concurrency below.
-	for i := range m.connectionManager {
-		var subs subscription.List
-		if !m.connectionManager[i].setup.SubscriptionsNotRequired {
-			if m.connectionManager[i].setup.GenerateSubscriptions == nil {
-				multiConnectFatalError = fmt.Errorf("cannot connect to [conn:%d] [URL:%s]: %w ", i+1, m.connectionManager[i].setup.URL, errWebsocketSubscriptionsGeneratorUnset)
-				break
-			}
+	for i, cW := range m.connectionManager {
```
gloriousCode: connectionWanager?
cm?
c?
I don't believe localised acronym variables have mixed casing like you're doing

exchange/websocket/manager.go
gloriousCode: I agree

exchange/websocket/manager.go
```diff
@@ -905,140 +881,112 @@ func drain(ch <-chan error) {
...
-	case err := <-m.ReadMessageErrors:
-		if errors.Is(err, errConnectionFault) {
-			log.Warnf(log.WebsocketMgr, "%v websocket has been disconnected. Reason: %v", m.exchangeName, err)
-			if m.IsConnected() {
-				if shutdownErr := m.Shutdown(); shutdownErr != nil {
-					log.Errorf(log.WebsocketMgr, "%v websocket: connectionMonitor shutdown err: %s", m.exchangeName, shutdownErr)
+// monitorConnection observes the connection and attempts to reconnect if the connection is lost
+func (m *Manager) monitorConnection() {
+	t := time.NewTimer(m.connectionMonitorDelay)
+	for done := false; !done; {
```
gloriousCode:
```go
t := time.NewTicker(m.connectionMonitorDelay)
	defer t.Stop()
	for {
```
And just return instead of `done`

exchange/websocket/manager.go
```diff
@@ -905,140 +881,112 @@ func drain(ch <-chan error) {
...
-		}
-		if m.IsConnected() {
-			go func() { // Without this the m.Shutdown() call below will deadlock
-				if err := m.Shutdown(); err != nil {
-					log.Errorf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown err: %s", m.exchangeName, err)
-				}
-			}()
+// monitorTraffic monitors to see if there has been traffic within the trafficTimeout time window.
+// If there is no traffic the connection is shutdown and will be reconnected by the connectionMonitor routine.
+func (m *Manager) monitorTraffic() {
```
gloriousCode, line 969: here too

exchange/websocket/manager.go
```diff
@@ -1083,3 +1031,29 @@ func (m *Manager) GetConnection(messageFilter any) (Connection, error) {
...
+	}
+	return connSubs, err
+}
+
+func (m *Manager) generateSubs() (subscription.List, error) {
+	if m.GenerateSubs != nil {
+		return m.GenerateSubs()
+	}
+	return m.GenerateSubscriptions()
+}
```
gloriousCode: Is there really no way to differentiate these three similarly named functions? No comment to suggest using this over the other ones either?

exchange/websocket/manager_test.go
```diff
@@ -1159,110 +1142,143 @@ func TestDrain(t *testing.T) {
...
+
+	go ws.monitorConsumers()
+
+	ws.DataHandler <- "test-1"
+	ws.DataHandler <- "test-2"        // Expect to be dropped
+	time.Sleep(10 * time.Millisecond) // Allow dropping to actually happen
+	assert.Equal(t, "test-1", <-ws.ToRoutine, "Should be able to drain expected value from ToRoutine")
+	ws.DataHandler <- "test-3"
+	time.Sleep(10 * time.Millisecond) // Allow dropping to actually happen
+	assert.Equal(t, "test-3", <-ws.ToRoutine, "Should restore delivery after dropping 1 message")
```
gloriousCode:
```suggestion
	ws.DataHandler <- "test-1"
	ws.DataHandler <- "test-2" // Expect to be dropped

	assert.Eventually(t, func() bool {
		select {
		case got := <-ws.ToRoutine:
			return got == "test-1"
		default:
			return false
		}
	}, time.Second, 10*time.Millisecond, "Should be able to drain expected value from ToRoutine")

	ws.DataHandler <- "test-3"
	assert.Eventually(t, func() bool {
		select {
		case got := <-ws.ToRoutine:
			return got == "test-3"
		default:
			return false
		}
	}, time.Second, 10*time.Millisecond, "Should restore delivery after dropping 1 message")
```
Is that a better translation? I understand its more code, but I try and avoid time.Sleep in testing

exchange/websocket/manager_test.go
```diff
@@ -1159,110 +1142,143 @@ func TestDrain(t *testing.T) {
...
-	require.True(t, ok)
-	require.ErrorIs(t, err, errConnectionFault)
-	// Handle outta closure shell
-	innerShell := ws.monitorConnection()
+
+	go ws.monitorConnection()
+
+	time.Sleep(50 * time.Millisecond)
+	assert.True(t, ws.connectionMonitorRunning.Load(), "enabled websocket should keep monitor running")
+	assert.Equal(t, connectedState, ws.state.Load(), "enabled websocket should remain in connected state")
```
gloriousCode:
```suggestion
	assert.Eventually(t, func() bool {
		return ws.connectionMonitorRunning.Load() && ws.state.Load() == connectedState
	}, 100*time.Millisecond, 5*time.Millisecond, "enabled websocket should keep monitor running and remain connected")
	
```

exchange/websocket/manager_test.go
```diff
@@ -1159,110 +1142,143 @@ func TestDrain(t *testing.T) {
...
+
 	ws.TrafficAlert <- struct{}{}
-	require.False(t, innerShell())
+	ws.setState(connectedState)
+
+	time.Sleep(40 * time.Millisecond)
+	assert.Equal(t, connectedState, ws.state.Load(), "monitorTraffic should not shutdown when receiving traffic")
+
+	time.Sleep(40 * time.Millisecond)
+	assert.Equal(t, disconnectedState, ws.state.Load(), "monitorTraffic should shutdown when no traffic received")
```
gloriousCode:
```suggestion
	assert.Eventually(t, func() bool {
		return ws.state.Load() == disconnectedState
	}, time.Second, 5*time.Millisecond, "monitorTraffic should shutdown when no traffic received")

	ws.m.Lock() // Prevents race with shutdown
	ws.setState(connectingState)
	go ws.monitorTraffic()
	ws.m.Unlock()

	assert.Eventually(t, func() bool {
		return ws.state.Load() == connectingState
	}, time.Second, 5*time.Millisecond, "monitorTraffic should not shutdown when connecting")

	ws.TrafficAlert <- struct{}{}
	ws.setState(connectedState)

	assert.Eventually(t, func() bool {
		return ws.state.Load() == connectedState
	}, time.Second, 5*time.Millisecond, "monitorTraffic should not shutdown when receiving traffic")

	assert.Eventually(t, func() bool {
		return ws.state.Load() == disconnectedState
	}, time.Second, 5*time.Millisecond, "monitorTraffic should shutdown when no traffic received")
```

exchange/websocket/subscriptions.go
```diff
@@ -275,52 +285,48 @@ func (m *Manager) FlushChannels() error {
...
 		return m.updateChannelSubscriptions(nil, m.subscriptions, newSubs)
 	}
 
-	for x := range m.connectionManager {
-		if m.connectionManager[x].setup.SubscriptionsNotRequired {
-			continue
-		}
-
-		newSubs, err := m.connectionManager[x].setup.GenerateSubscriptions()
+	for _, cW := range m.connectionManager {
```
gloriousCode: Same thing regarding `cW` being non-idiomatic and seemingly unrelated

gloriousCode (changes requested):
Looks good so far, mostly focused on /websocket package

config/versions/v11/v11.go
gbjk: Thanks, but this was a deliberate style choice.
The error condition is too long to read clearly like that.

exchange/websocket/manager.go
gbjk: Thanks.
Several passes of optimisation removed the need for a complicated `done` instead of just return.
This also applied below, so please review whole commit.

Fixed https://github.com/gbjk/gocryptotrader/commit/d403b38b

internal/testing/subscriptions/subscriptions.go
gbjk: Thanks, yes, I'm aware. I chose this style deliberately.

exchange/websocket/manager.go
gbjk: Yes, the branch this comes from has a bigger signature.
But it's also about saying that websocket manager consumers *aren't* just subscription consumers.
Except, right now, they are.
I guess it makes no difference, but if this isn't a big deal i'd like to leave it as is for clarity.

exchange/websocket/manager.go
gbjk: Nice, and thank you for spotting.
Slight Samendment that if we're going to over allocate the slice then we'll want to Clip it.

Fixed https://github.com/gbjk/gocryptotrader/commit/9e370a2f

config/versions/v11/v11.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/687a3341

config/versions/v11/v11.go
gbjk: Thanks!
Under what circumstances was I so rushed that `fluffins` what a necessary shortcut from just typing subscriptions in the first place?! 🤦 

Fixed https://github.com/gbjk/gocryptotrader/commit/0a2ac92a

config/versions/v11/v11.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/0a2ac92a

config/versions/v11/v11_test.go
gbjk: Agreed

Fixed https://github.com/gbjk/gocryptotrader/commit/e5354a1f

exchange/websocket/connection.go
gbjk: Thanks; These shouldn't be here - I add them as compile time assertions when working, but they should be moved to tests before merging.

Fixed https://github.com/gbjk/gocryptotrader/commit/a79044b99

exchange/websocket/manager.go
gbjk: The messages weren't acceptably clear to me, hence the Errorf.
Sometimes it's fine and "clear enough", but in this case it seemed to obscure.

exchange/websocket/manager.go
gbjk: Sorry for the blast radius for this, but if it applies to this one it applies to all of them.
This pattern is either cute or objectionable. Let me know which. 😂 

Fixed https://github.com/gbjk/gocryptotrader/commit/d8035c2d

exchange/websocket/manager.go
gbjk: connectionWrapper, but you're right that it's wrong/bad.

Fixed https://github.com/gbjk/gocryptotrader/commit/a6ef6225

Not a fixup because compound bases. Can probably merge it into another commit later though.

exchange/websocket/manager.go
gbjk: [We're using Go > 1.23 - defer t.Stop is redundant - GC will act correctly](https://go.dev/wiki/Go123Timer)

Return instead of done actioned already:
Fixed https://github.com/gbjk/gocryptotrader/commit/d403b38

exchange/websocket/manager.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/d403b38

exchange/websocket/manager.go
gbjk: Agreed. I kept feeling that.

* `manager.generateSubs` is now private and set to the default during setup
* `manager.generateSubs()` func is therefore removed
* `connection.generateSubscriptions` moved to subscriptions.go and documented
* Rogue comment in coinbase removed
* internal testing no longer sets the (now private) GenerateSubs. The exchanges without templating are: Coinut, Poloniex and BinanceUS. None of them hate tests that exercise subs that expect an empty/clean websocket for baseline. This was my thing anyway, so if it doesn't make sense, don't worry
* Removed a really weird thing in ADD_NEW_EXCHANGE that had an (now redundant) `generateSubscriptions` with reference to binance spot sub backwards compat 😂  

Fixed https://github.com/gbjk/gocryptotrader/commit/3d62415d
and
Fixed https://github.com/gbjk/gocryptotrader/commit/8e9ca33e

exchange/websocket/manager_test.go
gbjk: I nearly always agree with you on this, but in this one case I deliberately removed the `assert.Eventually` because I thought it was cumbersome.

Looking again, though, waiting 1 second is correct and less fragile.

Fixed https://github.com/gbjk/gocryptotrader/commit/a37d3994

exchange/websocket/manager_test.go
gbjk: Thanks

Fixed https://github.com/gbjk/gocryptotrader/commit/b34adca9

exchange/websocket/manager_test.go
gbjk: Thanks

Fixed https://github.com/gbjk/gocryptotrader/commit/b34adca9

exchange/websocket/subscriptions.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/67d2edc8

Also not a fixup, will see if it's base becomes cleaner later.

exchange/websocket/connection_test.go
```diff
@@ -59,3 +60,13 @@ func TestIncomingWithData(t *testing.T) {
 	require.Len(t, ch, 1, "must have one item in channel")
 	assert.Equal(t, []byte("test"), <-ch)
 }
+
+func TestAssetFilter(t *testing.T) {
+	t.Parallel()
+	assert.Implements(t, (*MessageFilter)(nil), AssetFilter(asset.Spot))
```
gloriousCode, line 66: 👍 Nice

exchange/websocket/manager_test.go
gloriousCode: A few of your commit links 404, it looks like not everything has been pushed

exchanges/binance/binance_wrapper.go
```diff
@@ -219,14 +213,14 @@ func (e *Exchange) Setup(exch *config.Exchange) error {
...
 	err = e.Websocket.Setup(&websocket.ManagerSetup{
-		ExchangeConfig:        exch,
-		DefaultURL:            binanceDefaultWebsocketURL,
-		RunningURL:            ePoint,
-		Connector:             e.WsConnect,
-		Subscriber:            e.Subscribe,
-		Unsubscriber:          e.Unsubscribe,
-		GenerateSubscriptions: e.generateSubscriptions,
-		Features:              &e.Features.Supports.WebsocketCapabilities,
+		Exchange:       e,
```
samuael, line 216: This is a bit weird to pass `e` as an input while the receiver of the websocket is `e`. itself. There has to be a better approach to accesing e.

exchanges/exchange_test.go
```diff
@@ -2484,24 +2457,33 @@ func TestGetCachedOpenInterest(t *testing.T) {
...
 func TestSetSubscriptionsFromConfig(t *testing.T) {
 	t.Parallel()
-	b := Base{Config: &config.Exchange{Features: &config.FeaturesConfig{}}}
+	b := Base{
+		Verbose: true,
+		Config: &config.Exchange{
+			Features: &config.FeaturesConfig{},
+		},
+	}
+	assert.NotPanics(t, func() { b.setSubscriptionsFromConfig() }, "should not panic with a nil Websocket")
```
samuael:
```suggestion
	assert.NotPanics(t, b.setSubscriptionsFromConfig, "should not panic with a nil Websocket")
```

exchange/websocket/manager.go
```diff
@@ -328,37 +340,24 @@ func (m *Manager) SetupNewConnection(c *ConnectionSetup) error {
...
-			return errMessageFilterNotComparable
-		}
-
-		for x := range m.connectionManager {
-			// Below allows for multiple connections to the same URL with different outbound request signatures. This
-			// allows for easier determination of inbound and outbound messages. e.g. Gateio cross_margin, margin on
-			// a spot connection.
-			if m.connectionManager[x].setup.URL == c.URL && c.MessageFilter == m.connectionManager[x].setup.MessageFilter {
-				return fmt.Errorf("%w: %w", errConnSetup, errDuplicateConnectionSetup)
+		if err := func() error {
```
samuael:
```suggestion
    var err error
	switch {
	case c.URL == "":
		err = errDefaultURLIsEmpty
	case c.Connector == nil:
		err = errWebsocketConnectorUnset
	case c.Subscriber == nil && !c.SubscriptionsNotRequired:
		err = errWebsocketSubscriberUnset
	case c.Unsubscriber == nil && m.features.Unsubscribe && !c.SubscriptionsNotRequired:
		err = errWebsocketUnsubscriberUnset
	case c.Handler == nil:
		err = errWebsocketDataHandlerUnset
	}
		if err != nil {
			return fmt.Errorf("%w %q: %w", errConnSetup, c.URL, err)
		}
```

samuael: I don't see the advantage of wrapping this section with a disposable function

samuael (commented):
One of these comments is a note for when/If _Poloniex_ is merged earlier that this PR.

exchanges/interfaces.go
```diff
@@ -77,8 +77,6 @@ type IBotExchange interface {
 	EnableRateLimiter() error
 	GetServerTime(ctx context.Context, ai asset.Item) (time.Time, error)
 	GetWebsocket() (*websocket.Manager, error)
-	SubscribeToWebsocketChannels(channels subscription.List) error
```
gloriousCode, line 80: Please add this to your PR description. You have removed the ability for websockets to make additional subscription requests outside of the connection setup.

I understand it is no longer fit for purpose, but I would like the comment there

exchange/websocket/manager.go
gloriousCode: Main thing is that your comment in code communicates none of that and is not there for clarity
I don't mind it being there

exchanges/binance/binance_wrapper.go
gloriousCode: I also don't oppose it staying. Unless you have something in mind, I'm happy to continue/accept future PRs to address it.

exchanges/bybit/bybit_wrapper.go
```diff
@@ -303,7 +303,7 @@ func (e *Exchange) Setup(exch *config.Exchange) error {
 		Handler: func(_ context.Context, conn websocket.Connection, resp []byte) error {
 			return e.wsHandleData(conn, asset.USDTMarginedFutures, resp)
 		},
-		MessageFilter: asset.USDTMarginedFutures, // Unused but it allows us to differentiate between the two linear futures types.
+		MessageFilter: websocket.AssetFilter(asset.USDTMarginedFutures), // Unused but it allows us to differentiate between the two linear futures types.
```
gloriousCode:
```
[ERROR] | WEBSOCKET | 12/01/2026 14:49:26 | Bybit websocket: subscription failure, message filter is not set
```

I cannot tell you which connection isn't setup properly, but it looks like you've missed some setting of new things. Good time to recheck all

exchange/websocket/subscriptions.go
```diff
@@ -23,6 +23,36 @@ var (
...
+// Calls connection specific GenerateSubscriptions if set. Currently only bybit uses this.
+// Otherwise the websocket manager subscriptions are filtered by MessageFilter
+func (w *connectionWrapper) generateSubscriptions(managerSubs subscription.List) (subscription.List, error) {
+	if w.setup.GenerateSubscriptions != nil {
+		return w.setup.GenerateSubscriptions()
+	}
+	connSubs := make(subscription.List, 0, len(managerSubs))
+	f := w.setup.MessageFilter
+	if f == nil {
+		return nil, errNoMessageFilter
```
gloriousCode, line 46: Are you able to impart anything more  helpful? See my error output, I do not know which thing has an issue and I will instead leave to you to address

exchange/websocket/subscriptions.go
```diff
@@ -23,6 +23,36 @@ var (
...
+	} else if err := common.NilGuard(m.Exchange); err != nil {
+		return nil, fmt.Errorf("m.Exchange: %w", err)
+	}
+	return m.Subscriptions.ExpandTemplates(m.Exchange)
+}
+
+// generateSubscriptions returns the subscriptions for a specific connection.
+// Calls connection specific GenerateSubscriptions if set. Currently only bybit uses this.
+// Otherwise the websocket manager subscriptions are filtered by MessageFilter
+func (w *connectionWrapper) generateSubscriptions(managerSubs subscription.List) (subscription.List, error) {
```
gloriousCode, line 39: I request test coverage per function, even unexported to allow for much better oversight on achieving each code scenario from clearly defined test setups. That I'm getting errors for bybit makes me request that you should make a test for this function. I think whatever the fix is for Bybit may optionally get its own test to avoid anything in future, but its probably one of those things that when its set, its set and doesn't need anything further

gloriousCode (changes requested):

exchange/websocket/connection.go
```diff
@@ -92,7 +93,7 @@ type ConnectionSetup struct {
 	Authenticate func(ctx context.Context, conn Connection) error
 	// MessageFilter defines the criteria used to match messages to a specific connection.
 	// The filter enables precise routing and handling of messages for distinct connection contexts.
-	MessageFilter any
+	MessageFilter MessageFilter
```
shazbert, line 96: suggestion: shift MessageFilter to the subscription package as subscription.Filter? along with all the sub types.

exchange/websocket/manager.go
```diff
@@ -905,140 +889,111 @@ func drain(ch <-chan error) {
...
+				log.Warnf(log.WebsocketMgr, "%v websocket: has not received a traffic alert in %v. Reconnecting", m.exchangeName, m.trafficTimeout)
+			}
+			if m.IsConnected() {
+				go func() { // Without this the m.Shutdown() call below will deadlock
+					if err := m.Shutdown(); err != nil {
+						log.Errorf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown err: %s", m.exchangeName, err)
+					}
+				}()
+			}
+			return
```
shazbert, line 993: GetConnection below seems like it could do to change the `any` type to `MessageFilter` also might need a func Equals or Is method to get around types that are not comparable maybe? Tests need to be updated when that change occurs.

exchanges/binance/binance_wrapper.go
gbjk: I don't understand "while the receiver of the websocket is e. itself".
e is not "the receiver of the websocket".
Websocket is a field on `Base`, an embedded field in e's `Exchange` type.
Any shared methods are actually on `Base`, not `Exchange`. They don't get to access `e`.
Field values can't magically access the container that holds a pointer to them.

Better to view "Exchange" field as "any thing that can implement the type of websocket.IExchange

That said, I share your distaste at having to do this.
But I had the same problem originally when writing `ExpandTemplates`. I can't see any way around it, and I've tried relatively hard.
Bottom line is that we eventually need to call real methods on `e` itself, not Base, and to do that we need `e` itself.

exchanges/exchange_test.go
gbjk: Thanks.

Fixed https://github.com/gbjk/gocryptotrader/commit/8616eb4d

exchange/websocket/manager.go
gbjk: Oh, yeah, that's better.
I went snow blind and resorted to a familiar pattern. But we're not doing anything complicated inside, so it's crazy. Thanks.

Fixed https://github.com/gbjk/gocryptotrader/commit/b122d060

exchanges/interfaces.go
gbjk: Done.

Note: I'm not intending, and don't think, that we limit subscription requests to connection setup.
Aside: But I do need to move towards tighter guiderails for how subs enter the system and what state they're in.

So as-of-now, you can still add them through the Manager itself. And in the future we still want that to be the case, we just need to make sure they're validated and expanded.

exchanges/bybit/bybit_wrapper.go
gbjk: Frustrating. I'm sure I had this at one point. Thanks for spotting.
Fixed both the lack of context, and also the issue.

Fixed https://github.com/gbjk/gocryptotrader/commit/21e727e5

exchange/websocket/subscriptions.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/21e727e5

exchange/websocket/subscriptions.go
gbjk: Fixed https://github.com/gbjk/gocryptotrader/commit/da6e0dc4

exchange/websocket/manager.go
```diff
@@ -328,37 +340,23 @@ func (m *Manager) SetupNewConnection(c *ConnectionSetup) error {
...
+		case c.Unsubscriber == nil && m.features.Unsubscribe && !c.SubscriptionsNotRequired:
+			err = errWebsocketUnsubscriberUnset
+		case c.Handler == nil:
+			err = errWebsocketDataHandlerUnset
 		}
-		if c.Handler == nil {
-			return fmt.Errorf("%w: %w", errConnSetup, errWebsocketDataHandlerUnset)
-		}
-
-		if c.MessageFilter != nil && !reflect.TypeOf(c.MessageFilter).Comparable() {
```
shazbert, line 350: suggestion: I still think it should be comparable. So we can at least make sure we can get a websocket connection from `GetConnection` or split it out as WebSocket key/identifier and filter fields. I need to think more on this.

exchange/websocket/manager.go
```diff
@@ -905,140 +888,111 @@ func drain(ch <-chan error) {
...
-			go func() { // Without this the m.Shutdown() call below will deadlock
-				if err := m.Shutdown(); err != nil {
-					log.Errorf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown err: %s", m.exchangeName, err)
-				}
-			}()
+// monitorTraffic monitors to see if there has been traffic within the trafficTimeout time window.
+// If there is no traffic the connection is shutdown and will be reconnected by the connectionMonitor routine.
+func (m *Manager) monitorTraffic() {
+	t := time.NewTimer(m.trafficTimeout)
+	for done := false; !done; {
```
shazbert, line 971:
```suggestion
	for {
```

exchange/websocket/manager.go
```diff
@@ -905,140 +888,111 @@ func drain(ch <-chan error) {
...
+		select {
+		case <-m.ShutdownC:
+			if m.verbose {
+				log.Debugf(log.WebsocketMgr, "%v websocket: trafficMonitor shutdown message received", m.exchangeName)
+			}
+			return
+		case <-t.C:
```
shazbert, line 978: might be able to use `time.After`

shazbert (changes requested):
Had a quick skim to get my head around changes will go through this more tomorrow, 🚀

exchange/websocket/manager.go
gloriousCode: Yeah there is an `==` check on message filters which this code ensures it won't have issues retrieving the connection. 
I missed this being removed, so I'll have to relook at how things are being done

gbjk:
Status update: Waiting for #2105 to merge, then I'll address feedback, fixup sub handling for poloniex, and rebase on master.

PR #2109, opened 2025-11-18
okx/websocket: scale connections to subscriptions for multi-connection websocket

shazbert:
# PR Description

This is a quick refactor to support multiple concurrent connections due to exchange subscription count limitations. Introduced connection pooling with capacity tracking and fixed several subscription-related bugs.

If there are limitations on subscriptions per connection for an exchange it will need to be migrated to multi-connection to work. This is not compatible with the old system. 

#### websocket 
* renamed connectionWrapper to websocket 
* Connection now has internalised a subscription store for tracking capacity and I guess now can be a nice lookup for the exchange wrappers. 
* Replaced single-connection logic with a []websocket keeps global subscription for now.
* Added connectAndSubscribe() – shared logic used by both initial connect() and FlushChannels()
* Introduced scaleConnectionsToSubscriptions() to dynamically create, reuse, and close connections based on current subscription count

#### OKX 
* Migrated to multi-connection mode for all connections
* Fixed subscription duplication bug in handleSubscription by extracting payload creation into chunkRequests()
* Removed unnecessary nil checks on raw WebSocket responses
* Standardized API endpoint resolution using GetURL() for consistency
* Integrated business connection
→ Note: Default asset subscriptions are incomplete and conflict with spread template expansion – update TODO for follow-up template pass
* Fix margin/spot double subscribe between connections.

Compliments issue #1764 

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/okx/okx_test.go
+++ b/exchanges/okx/okx_test.go
@@ -6173,7 +6173,7 @@ func TestBusinessWSCandleSubscriptions(t *testing.T) {
 		}
 		return len(got) == 3
 	}
-	assert.Eventually(t, check, 5*time.Second, time.Nanosecond)
+	assert.Eventually(t, check, 5*time.Second, time.Millisecond)
 	require.Equal(t, 3, len(got), "must receive candles for all three subscriptions")
 	require.NoError(t, got.ContainsAll(p, true), "must receive candles for all subscribed pairs")
 }


thrasher-:
@codex please review

exchange/websocket/subscriptions.go
```diff
@@ -224,20 +224,26 @@ func (m *Manager) GetSubscriptions() subscription.List {
 // The subscription state is not considered when counting existing subscriptions
 func (m *Manager) checkSubscriptions(conn Connection, subs subscription.List) error {
 	var subscriptionStore *subscription.Store
+	var usedCapacity int
 	if wrapper, ok := m.connections[conn]; ok && conn != nil {
 		subscriptionStore = wrapper.subscriptions
+		if subscriptionStore == nil {
+			return fmt.Errorf("%w: Websocket.subscriptions", common.ErrNilPointer)
+		}
+		connStore := wrapper.connectionSubs[conn]
```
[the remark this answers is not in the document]
gloriousCode: I suggest the normal `, ok` check instead

exchange/websocket/manager.go
```diff
@@ -517,65 +520,38 @@ func (m *Manager) connect() error {
...
+		for _, batchedSubs := range common.Batch(subs, m.MaxSubscriptionsPerConnection) {
+			if err := m.connectAndSubscribe(context.TODO(), m.connectionManager[i], batchedSubs); err != nil {
+				if errors.Is(err, errFatalError) {
+					multiConnectFatalError = fmt.Errorf("cannot connect to [conn:%d] [URL:%s]: %w ", i+1, m.connectionManager[i].setup.URL, err)
+					break
+				}
+				subscriptionError = common.AppendError(subscriptionError, fmt.Errorf("subscription error on [conn:%d] [URL:%s]: %w ", i+1, m.connectionManager[i].setup.URL, err))
+			}
+			if m.verbose {
+				log.Debugf(log.WebsocketMgr, "%s websocket: [URL:%s] connected. [Subscribed: %d]", m.exchangeName, m.connectionManager[i].setup.URL, len(subs))
```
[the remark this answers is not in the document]
gloriousCode: I suggest both batchSubs and total subs

exchange/websocket/manager.go
```diff
@@ -60,6 +60,8 @@ var (
 	errExchangeConfigEmpty                  = errors.New("exchange config is empty")
 	errCannotObtainOutboundConnection       = errors.New("cannot obtain outbound connection")
 	errMessageFilterNotComparable           = errors.New("message filter is not comparable")
+	errFatalError                           = errors.New("fatal error")
```
gbjk: 🐒 Optional nitpick:
Surely this is a common error?

exchange/websocket/manager.go
```diff
@@ -517,65 +520,38 @@ func (m *Manager) connect() error {
...
 
-		// TODO: Add window for max subscriptions per connection, to spawn new connections if needed.
-
-		conn := m.getConnectionFromSetup(m.connectionManager[i].setup)
-
-		if err := m.connectionManager[i].setup.Connector(context.TODO(), conn); err != nil {
-			multiConnectFatalError = fmt.Errorf("%v Error connecting %w", m.exchangeName, err)
-			break
+		if m.connectionManager[i].setup.SubscriptionsNotRequired && len(subs) == 0 {
+			subs = subscription.List{{}} // Dummy subscription to utilise the connect and subscribe function below
```
gloriousCode:
```suggestion
		if m.connectionManager[i].setup.SubscriptionsNotRequired && len(subs) == 0 {
			subs = subscription.List{{}} // Dummy subscription to utilise common.Batch to call connectAndSubscribe in a loop
```

exchange/websocket/manager.go
```diff
@@ -602,6 +578,55 @@ func (m *Manager) connect() error {
 	return subscriptionError
 }
 
+func (m *Manager) connectAndSubscribe(ctx context.Context, wrapper *connectionWrapper, subs subscription.List) error {
+	if m.MaxSubscriptionsPerConnection > 0 && len(subs) > m.MaxSubscriptionsPerConnection {
+		return fmt.Errorf("%w %w: max subs allowed %d, requested %d", errFatalError, errSubscriptionsExceedsLimit, m.MaxSubscriptionsPerConnection, len(subs))
```
gloriousCode: `MaxSubscriptionsPerConnection` may be better as part of`wrapper` setup rather than the `Manager` setup. Allowing for `MaxSubscriptionsPerConnection` to be per type of connection....
But that seems more of a `// TODO` if it actually comes up
... So dw

exchange/websocket/manager.go
```diff
@@ -1073,11 +1098,14 @@ func (m *Manager) GetConnection(messageFilter any) (Connection, error) {
...
 
 	for _, wrapper := range m.connectionManager {
-		if wrapper.setup.MessageFilter == messageFilter {
-			if wrapper.connection == nil {
-				return nil, fmt.Errorf("%s: %s %w associated with message filter: '%v'", m.exchangeName, wrapper.setup.URL, ErrNotConnected, messageFilter)
-			}
-			return wrapper.connection, nil
+		if wrapper.setup.MessageFilter != messageFilter {
+			continue
+		}
```
gloriousCode: Thank you

exchange/websocket/manager_test.go
```diff
@@ -308,7 +310,12 @@ func TestConnectionMessageErrors(t *testing.T) {
 	err = ws.Connect()
 	require.NoError(t, err)
 
-	err = ws.connectionManager[0].connection.SendRawMessage(t.Context(), request.Unset, gws.TextMessage, []byte("test"))
+	var conn Connection
+	for conn = range ws.connectionManager[0].connectionSubs {
+		break
+	}
```
gloriousCode: ?

exchange/websocket/subscriptions.go
```diff
@@ -350,3 +335,103 @@ func (m *Manager) updateChannelSubscriptions(c Connection, store *subscription.S
 	}
 	return nil
 }
+
+// scaleConnectionsToSubscriptions scales connections to subscriptions based off current subscription list
+func (m *Manager) scaleConnectionsToSubscriptions(ctx context.Context, wrapper *connectionWrapper, incoming subscription.List) error {
```
gloriousCode: I'm going to request 100% coverage of this function _within_ TestScaleConnectionsToSubscriptions. The cases that aren't covered are important for scaling.
There is some coverage via other functions, I don't think that cuts it for this important function - plus there's still gaps despite that

It's okay if it ends up not being 100%, but I really want you to try and leave a comment on why you can't get the remainder so I can see if I can help

exchanges/okx/okx_test.go
```diff
@@ -6114,26 +6125,36 @@ func TestBusinessWSCandleSubscriptions(t *testing.T) {
...
-		err := e.BusinessSubscribe(t.Context(), subscription.List{{Channel: ch, Pairs: p[i : i+1]}})
+		err := e.BusinessSubscribe(t.Context(), conn, subscription.List{{Channel: ch, Pairs: p[i : i+1]}})
 		require.NoErrorf(t, err, "BusinessSubscribe %s-%s must not error", ch, p[i])
 	}
 
+	timeout := time.NewTimer(5 * time.Second)
+	defer timeout.Stop()
 	var got currency.Pairs
-	assert.Eventually(t, func() bool {
+check:
```
gloriousCode: Assert eventually is better. But maybe I haven't looked at the reason why it needed to be changed

exchange/websocket/manager.go
```diff
@@ -159,9 +161,9 @@ type ManagerSetup struct {
 // attempting a new connection. It also contains the subscriptions that are
 // associated with the specific connection.
 type connectionWrapper struct {
-	setup         *ConnectionSetup
-	subscriptions *subscription.Store
-	connection    Connection
+	setup          *ConnectionSetup
+	subscriptions  *subscription.Store
+	connectionSubs map[Connection]*subscription.Store
```
gbjk: 🤷‍♂️ Clarification please: 
I'll put the bulk of my overall PR question here so we can discuss in a thread.
1. Isn't this circular, with the same conn in `m.connections` and then again inside it?
`m.connections[CONN].connectionSubs[CONN] = store`
2. I feel like I'm being really stupid ... but ... we never seem to add more than one connection to this map?
When we call `connectAndSubscribe` in `scaleConnectionsToSubscriptions` we're doing it with the same wrapper, and we're going to nuke the existing entry.
I don't see where we ever actually create multiple subs.

gbjk (commented):
Nice work and I understand the direction.

But I'm really confused by the implementation. It's probably just me.
I've left one main thread for us to discuss, but I'll also reach out on slack.

exchange/websocket/manager_test.go
shazbert: Grabs first available connection, want me to use GetConnection method func? Or write a comment?

exchange/websocket/manager.go
shazbert: Yeah can be both in future, as it's usually a per exchange configuration.

exchange/websocket/manager.go
shazbert: You are welcome

exchanges/okx/okx_test.go
```diff
@@ -4012,6 +4012,10 @@ var pushDataMap = map[string]string{
 	"Balance Save Error":                    `{"arg": {"channel": "balance_and_position","uid": "77982378738415880"},"data": [{"pTime": "1597026383085","eventType": "snapshot","balData": [{"ccy": "BTC","cashBal": "1","uTime": "1597026383085"}],"posData": [{"posId": "1111111111","tradeId": "2","instId": "BTC-USD-191018","instType": "FUTURES","mgnMode": "cross","posSide": "long","pos": "10","ccy": "BTC","posCcy": "","avgPx": "3320","uTIme": "1597026383085"}]}]}`,
 }
 
+type ConnectionFixture struct {
+	websocket.Connection
+}
+
```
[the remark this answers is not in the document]
shazbert: Will rm it completely as it currently isn't needed.

exchanges/okx/okx_test.go
shazbert: There was a lot of data coming through on update, will drop tick size down.

exchange/websocket/manager_test.go
gloriousCode: Well that's pretty fun!
Bit of an extra tangent sorry: Looking at all the usages of connectionSubs, is a map the most appropriate? In nearly every instance of usage, you iterate over it via `for ..., ... range ...connectionSubs {`. I'm also not seeing where you write to the map.
So would a slice be more appropriate?

shazbert: Yeah I can see what a slice does, and seeing as there shouldn't be too many connections anyway the map lookup isn't going to give much a performance boost. I only see around 30 being the "minimum" max subscription size.  So around 100 connections could spawn off a 3000 subs. 

As for the writing to the map its done in `connectAndSubscribe` I only used the map for ease of implementation and  it's not really a hot path. 

```go
	m.connections[conn] = wrapper
	connectionStore := subscription.NewStore()
	wrapper.connectionSubs[conn] = connectionStore
```

exchange/websocket/manager.go
shazbert: It was/is but it works, It's now a slice; they were being added. I have been using gctcli and it seems work to scale up and down. to tests the spawn connection cleanup you can drop the MaxSubscriptionsPerConnection to ~2 and then you will see it have some fun.

exchange/websocket/subscriptions.go
gloriousCode: https://gist.github.com/gloriousCode/1961142207365adca24582a9013fa0a2

shazbert: I have added patch thanks! https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/3b105b42b9e20ef9f41272d76a2f91b71bc43a4c

shazbert:
@dazi005 Can you please test this branch when you get a chance?

dazi005:
> @dazi005 Can you please test this branch when you get a chance?

Sure, no problem. I'll arrange a test later.

exchanges/okx/okx_websocket.go
```diff
@@ -16,7 +16,6 @@ import (
```
gloriousCode, line 16:
```
[DEBUG] | WEBSOCKET | 01/12/2025 10:29:30 | Okx wss://ws.okx.com:8443/ws/v5/public: Message received: {"event":"error","msg":"Wrong URL or channel:candles,instId:BTC-USD-260626-80000-C doesn't exist. Please use the correct URL, channel and parameters referring to API document.","code":"60018","connId":"4bf01114"}
```

edit: I think there has been a change since I last ran OKX where candles are now limited to `spot`. So I don't think this applies
edit 2: Looks like there is no more candle subscriptions for okx 😅

exchanges/okx/okx_websocket.go
```diff
@@ -16,7 +16,6 @@ import (
 
 	"github.com/buger/jsonparser"
```
gloriousCode, line 17:
```
[ERROR] | WEBSOCKET | 01/12/2025 10:35:39 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx BONK-USDT spot Reason: [orderbook data integrity compromised, Okx BONK-USDT spot checksum mismatch: expected '2480260311', got '1006267793']
```
I think this one may be due to rate limiting because I am also getting these:
```
[ERROR] | REQUESTER | 01/12/2025 10:35:31 | Okx request has failed. Retrying request in 100ms, attempt 1
```

exchanges/okx/okx_websocket.go
```diff
@@ -16,7 +16,6 @@ import (
...
 	gws "github.com/gorilla/websocket"
```
gloriousCode, line 18:
```
[ERROR] | WEBSOCKET | 01/12/2025 10:29:32 | subscription error on [conn:3] [URL:wss://ws.okx.com:8443/ws/v5/business]: subscription failure: subscription already in state: sprd-tickers spread BTC/USD-SWAP_BTC-USD-260327, duplicate subscription: sprd-tickers spread BTC/USD-SWAP_BTC-USD-260327 , subscription error on [conn:3] [URL:wss://ws.okx.com:8443/ws/v5/business]: subscription failure: subscription already in state: block-tickers perpetualswap LTC/USD-SWAP, duplicate subscription: block-tickers perpetualswap LTC/USD-SWAP 
```

I think the most interesting observation about this above error is that I get the exact same sub error on multiple goes of testing

exchanges/okx/okx_websocket.go
```diff
@@ -16,7 +16,6 @@ import (
...
-	"github.com/thrasher-corp/gocryptotrader/common"
```
gloriousCode, line 19:
```
[WARN]  | WEBSOCKET | 01/12/2025 10:44:35 | Okx websocket has been disconnected. Reason: websocket: close 1006 (abnormal closure): unexpected EOF: connection fault ("wss://ws.okx.com:8443/ws/v5/public")

```

I get this one rather frequently

gloriousCode: Hard to say, but this one may prompt a new field say `MaxWebsocketConnections`:
> Connection count limit
>The limit will be set at 30 WebSocket connections per specific WebSocket channel per sub-account. Each WebSocket connection is identified by the unique connId.

exchanges/okx/okx_websocket.go
shazbert: Yeah I think it was intentionally routed by a separate business connection and subscription. I kept the TestBusinessWSCandleSubscriptions func just to demonstrate that. It will have to be correctly managed with a template update. But I will unresolve this for other peoples input if its coming up on general subscriptions.

exchanges/okx/okx_websocket.go
shazbert: I did a quick count and check and even if all available pairs are subscribed its around 5205 subs / 240 subs per connection its around 22 connections. 

The culprit which I think could be more probable of causing disconnection is to do with our `websocket routine manager/engine` processing which causes a lot of latency. Trying to log everything to stdout etc. 

I will add in the additional field, no reason why we can't screen that out.

exchanges/okx/okx_websocket.go
shazbert: added in request rate limiting now

exchanges/okx/okx_websocket.go
shazbert: Fixed bug in business subscriptions

exchanges/okx/okx_websocket.go
shazbert: You were right, I tested it with 30 subs per connections and it was stable, 240 subs per conn was in the old API version 🤷, it will disconnect periodically if we are writing to stdout when running the library. But for what's in scope it seems like its stable. Thanks for bringing this up. This will need a thorough test from you as well.

exchanges/okx/okx_websocket.go
gloriousCode: I am still getting some form of rate limiting, but I think its REST and it doesn't appear to be breaking orderbooks:
```
[DEBUG] | REQUESTER | 02/12/2025 17:35:39 | Okx attempt 1 request path: https://www.okx.com/api/v5/sprd/books?size=50&sprdId=BTC-USDT-251226_BTC-USDT-260626
[DEBUG] | REQUESTER | 02/12/2025 17:35:39 | Okx request header [Content-Type]: [application/json]
[DEBUG] | REQUESTER | 02/12/2025 17:35:39 | Okx request type: GET
[DEBUG] | REQUESTER | 02/12/2025 17:35:39 | Okx request body: 
[ERROR] | REQUESTER | 02/12/2025 17:35:39 | Okx request has failed. Retrying request in 100ms, attempt 1
```
Its also reduced a bit. It may really just be out of date OKx limits.

Thanks for your updates though. This doesn't require immediate remedy, I'll try and do more debugging than staring at logs and yelling out "GOTCHA SHAZBERT"

exchange/websocket/manager.go
```diff
@@ -517,66 +527,39 @@ func (m *Manager) connect() error {
...
-		m.connectionManager[i].connection = conn
-
-		m.Wg.Add(1)
-		go m.Reader(context.TODO(), conn, m.connectionManager[i].setup.Handler)
-
-		if m.connectionManager[i].setup.Authenticate != nil && m.CanUseAuthenticatedEndpoints() {
-			if err := m.connectionManager[i].setup.Authenticate(context.TODO(), conn); err != nil {
-				multiConnectFatalError = fmt.Errorf("%s websocket: [conn:%d] [URL:%s] failed to authenticate %w", m.exchangeName, i+1, conn.URL, err)
-				break
+		for _, batchedSubs := range common.Batch(subs, m.MaxSubscriptionsPerConnection) {
```
samuael, line 530: Are you suggesting we don't run the `connectAndSubscribe()` method if `SubscriptionsNotRequired`? Because if the length of subscription `subs` is zero, the _loop will not be executed even once_. Meaning, we don't have a connection created for that given setup.

I believe a connection must be established even if it doesn't allow subscription. Some exchanges require **trade**  calls on a separate websocket connection. Just for trade calls with no susbcription.

exchange/websocket/manager.go
```diff
@@ -654,15 +687,16 @@ func (m *Manager) shutdown() error {
 	var nonFatalCloseConnectionErrors error
 
 	// Shutdown managed connections
-	for x := range m.connectionManager {
-		if m.connectionManager[x].connection != nil {
-			if err := m.connectionManager[x].connection.Shutdown(); err != nil {
+	for _, wrapper := range m.connectionManager {
+		for _, connSubs := range wrapper.connectionsSubs {
+			if err := connSubs.Connection.Shutdown(); err != nil {
```
samuael: Should we be not checking if this connection is still active?

exchange/websocket/subscriptions.go
```diff
@@ -224,20 +224,34 @@ func (m *Manager) GetSubscriptions() subscription.List {
 // The subscription state is not considered when counting existing subscriptions
 func (m *Manager) checkSubscriptions(conn Connection, subs subscription.List) error {
 	var subscriptionStore *subscription.Store
+	var usedCapacity int
 	if wrapper, ok := m.connections[conn]; ok && conn != nil {
-		subscriptionStore = wrapper.subscriptions
+		if subscriptionStore = wrapper.subscriptions; subscriptionStore == nil {
```
samuael: We should not be assigning variables if there is an error check before a code section where they will be used.

```suggestion
		if wrapper.subscriptions; subscriptionStore == nil {
```

exchange/websocket/subscriptions.go
```diff
@@ -224,20 +224,34 @@ func (m *Manager) GetSubscriptions() subscription.List {
...
+		if connSubStore == nil {
+			return fmt.Errorf("%w: connection subscription store not found", common.ErrNilPointer)
+		}
+		usedCapacity = connSubStore.Len()
 	} else {
 		subscriptionStore = m.subscriptions
-	}
-	if subscriptionStore == nil {
-		return fmt.Errorf("%w: Websocket.subscriptions", common.ErrNilPointer)
+		if subscriptionStore == nil {
```
samuael: `subscriptionStore` is checked if it is nil. So, we can move this to the else statement and remove the 	`var subscriptionStore *subscription.Store` declaration. And, use only the `usedCapacity` variable since it is the only variable used below.

samuael (changes requested):
This PR is very cool. Some exchanges set a subscription limits and this PR has a clear approach to bypass it. 🫡 .

exchange/websocket/manager.go
shazbert: Above this call it inserts a dummy sub to kick this off, now that you have brought this up, maybe this is a bad design choice. Will have a think more on this. 

```go
if m.connectionManager[i].setup.SubscriptionsNotRequired && len(subs) == 0 {
			subs = subscription.List{{}} // Dummy subscription to utilise common.Batch to call connectAndSubscribe in a loop
		}
```

exchange/websocket/manager.go
shazbert: In shutdown the underlying conn is closed, unless you don't believe that resources are released for any reason. I don't think we need the added check?

exchange/websocket/subscriptions.go
shazbert: Why though, we return immediately on error and this reduces code footprint?

exchange/websocket/subscriptions.go
shazbert: `subscriptionStore` is used below to see if there is duplicated subscriptions? Can you send me diff please maybe I am misunderstanding.

exchange/websocket/subscriptions.go
gloriousCode: I mostly just find it ugly. It's also not really less footprint, just on the same line. Plus you also don't do it in your else statement.

Only because you asked why. I'm not going to request its change vs everything else in this PR

shazbert: Have pushed up to get rid of ugliness. @samuael In future can you please suggest code that can be committed without errors, because the suggestion didn't make sense to me.

exchange/websocket/manager.go
shazbert: Hopefully this makes it more intuitive: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/7324c0748980a219307c7d47ce9fe565b470f0e6

exchange/websocket/subscriptions.go
samuael: Ohh my bad. I will, sir.
And, 👍  for applying the change

exchange/websocket/manager.go
samuael: I hope this isn't a silly question. Do we remove the instance from the `connectionsSubs` list when disconnected?

exchange/websocket/manager.go
samuael: Yes. This one is better. 👍

exchange/websocket/manager.go
shazbert: It should be, if you have time can you please check.

exchange/websocket/manager.go
```diff
@@ -159,9 +160,14 @@ type ManagerSetup struct {
...
-	subscriptions *subscription.Store
-	connection    Connection
+	setup           *ConnectionSetup
+	subscriptions   *subscription.Store
+	connectionsSubs []connectionSubscriptions
+}
+
+type connectionSubscriptions struct {
+	Connection    Connection
+	Subscriptions *subscription.Store
```
gloriousCode: Both can be unexported without issue. Resolve if it is somehow references externally

exchange/websocket/manager.go
gloriousCode: The original diff and fix are really just a symptom of having `connectAndSubscribe` be one function. The only other issue is the amount of things called `connect` in the package that its all so confusing

pseudocode:
```go
...
fatalErr = m.connectoroo(context.TODO(), m.connectionManager[i], nil)
if if m.connectionManager[i].setup.SubscriptionsNotRequired && len(subs) == 0 {
  continue
}
for submaroo := range. batcheroo(subs, subLim) {
    fatalErr = m.submaroo(context.TODO(), m.connectionManager[i], submaroo)
}
```

I'll look at things some more (could have changed since I wrote this comment) before advising anything

exchange/websocket/connection.go
```diff
@@ -60,9 +60,11 @@ type Connection interface {
...
 // ConnectionSetup defines variables for an individual stream connection
 type ConnectionSetup struct {
-	ResponseCheckTimeout     time.Duration
-	ResponseMaxLimit         time.Duration
-	RateLimit                *request.RateLimiterWithWeight
+	ResponseCheckTimeout time.Duration
+	ResponseMaxLimit     time.Duration
+	RateLimit            *request.RateLimiterWithWeight
+	// ConnectionRateLimit returns a new rate limiter for each connection instance
+	ConnectionRateLimit      func() *request.RateLimiterWithWeight
```
gbjk: 🚧 Change request:
```suggestion
	ConnectionRateLimiter      func() *request.RateLimiterWithWeight
```

exchange/websocket/manager.go
```diff
@@ -602,6 +592,56 @@ func (m *Manager) connect() error {
...
+	if err := wrapper.setup.Connector(ctx, conn); err != nil {
+		return fmt.Errorf("%w: %w", common.ErrFatal, err)
+	}
+
+	if !conn.IsConnected() {
+		return fmt.Errorf("%w: %w", common.ErrFatal, ErrNotConnected)
+	}
+
+	m.connections[conn] = wrapper
+	connectionStore := subscription.NewStore()
```
gbjk: 🚧 Change request:
Really bad name for a store of things that aren't connections.
```suggestion
	connSubsStore := subscription.NewStore()
```

exchange/websocket/manager.go
gbjk: 🚧 Change request:
If Connections can have subscriptions, why not simplify both "multi" and normal handling and put the subs inside the Connection.

My problem with this current pattern is that there's loads of code in Manager related to connections and subs that seems like it should live in the `*connection` itself.

exchange/websocket/manager.go
```diff
@@ -159,9 +160,14 @@ type ManagerSetup struct {
 // attempting a new connection. It also contains the subscriptions that are
 // associated with the specific connection.
 type connectionWrapper struct {
```
gbjk: 🚧 Change request:
This name has become overloaded and confusing.
Indulge me summarising and tell me if I've got it right, please.
We have:
* m.connections which is a map of all connections to their setup (connectionWrapper)
* The connectionWrapper has subscriptions, which is a fan out of all the subscriptions configured for the entire websocket
* The connection Wrapper then has slice containing a circular reference back to that same connection again, along with the subscriptions specific to the connection

I'm not sure we can make it simpler in one step, but introducing a simple extra name would help.
1. We have a connection to a websocket, and that has connected subs in it (See change request to move them inside *connection
2. We then want to store the config for that connection. That feels like it's actually a `Weboscket`.

So we have websockets, which define the endpoints we'll connect to, and the subs we want to have overall.
And then connections to those websockets.

I think you'd still have circular references here, but that seems okay.
Sometimes you need to find a websocket from it's connections, and sometimes you need to find connections for a websocket.

The abstraction I think we're missing is to have Websocket be a thing, which probably just renames `ConnectionSetup`.

I've got a lot of work in this area outstanding to PR 😢 
I don't know if it's better to Frankenstein some of it here or not.
You can review it [here](https://github.com/gbjk/gocryptotrader/blob/feature/websocket_multi_v4/exchange/websocket/manager.go)

Kinda die inside when I see how much is there that isn't PR'd yet.
I was trying to get one step forward with the PR for GateIO's sub api pathways. That's stalled without review attention, and I don't want to do more work on ws_multi_v4 until it's merged.

exchange/websocket/manager.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/8943b1662caf374a5027ffc66fa2eee7ea8287c2

exchange/websocket/manager.go
shazbert: The summary seems right. Have moved subs into connection. Will now change `connectionWrapper` to `websocket`. 

> The abstraction I think we're missing is to have Websocket be a thing, which probably just renames ConnectionSetup.

I am not changing `ConnectionSetup` name just yet am I?

> I don't know if it's better to Frankenstein some of it here or not.

Can just iterate on top of this change for now, because this works for what I need it to do and I am time constrained on other things.

> Kinda die inside when I see how much is there that isn't PR'd yet.

🕊️ 

> I was trying to get one step forward with the PR for GateIO's sub api pathways. That's stalled without review attention, and I don't want to do more work on ws_multi_v4 until it's merged.

Noted

shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/a60f90d7fdea37dc9ebde73269e8fde05d1406a4

exchange/websocket/connection.go
```diff
@@ -56,15 +56,17 @@ type Connection interface {
 	IncomingWithData(signature any, data []byte) bool
 	// MatchReturnResponses sets up a channel to listen for an expected number of responses.
 	MatchReturnResponses(ctx context.Context, signature any, expected int) (<-chan MatchedResponse, error)
+	// SubStore returns the subscription store for the connection
+	SubStore() *subscription.Store
```
gbjk: 🦍 Optional pedantic nitpick:
```suggestion
	// Subscriptions returns the subscription store for the connection
	Subscriptions() *subscription.Store
```

exchange/websocket/connection.go
```diff
@@ -111,6 +113,7 @@ type Response struct {
 
 // connection contains all the data needed to send a message to a websocket connection
 type connection struct {
+	*subscription.Store
```
gbjk: 🚧 Change request:
Not sure why we're embedding the pointer like this?
It seems wrong... Subscription store feels like it shouldn't intermingle with connection that openly.
```suggestion
	subscriptions *subscription.Store
```

shazbert: Yeah good catch, originally I wanted the methods exposed on the interface so we don't need to call the method to get the store but that was annoying.

exchange/websocket/manager.go
```diff
@@ -1072,13 +1103,14 @@ func (m *Manager) GetConnection(messageFilter any) (Connection, error) {
...
-			}
-			return wrapper.connection, nil
+	for _, ws := range m.connectionManager {
+		if ws.setup.MessageFilter != messageFilter {
+			continue
+		}
+		if len(ws.connections) == 0 {
+			return nil, fmt.Errorf("%s: %s %w associated with message filter: '%v'", m.exchangeName, ws.setup.URL, ErrNotConnected, messageFilter)
 		}
+		return ws.connections[0], nil
```
samuael, line 1087: Why would we have multiple connection per `messageFilter`? If so, should't we have additional filter to exactly know which connection it is?

samuael (commented):
Great stuff 👍 . Just a single comment or question I have on this round

exchange/websocket/manager.go
shazbert: Main reason is that I have never been using it as a message filter, but a key to a `websocket` which gives back its first connection for outbound trading purposes.  @gbjk Has some updates to this, I think that can retrieve a connection from its subscription.  

> If so, should't we have additional filter to exactly know which connection it is? 

A specific connection isn't really that important, knowing what subscriptions there are and what is subscribed is, but I think you really need to figure out intention when getting that connection and what you need it to do, because the subscription and connections handling (with scaling now) are all done in websocket manager.

exchange/websocket/manager.go
```diff
@@ -602,6 +583,55 @@ func (m *Manager) connect() error {
 	return subscriptionError
 }
 
+func (m *Manager) connectAndSubscribe(ctx context.Context, ws *websocket, subs subscription.List) error {
+	if m.MaxSubscriptionsPerConnection > 0 && len(subs) > m.MaxSubscriptionsPerConnection {
```
samuael, line 585: Since you're here, have you not considered having a variable to decide the maximum connection that must be established if specified? Because I know that several exchanges have a limit to the maximum WebSocket connections you can create per IP or account.

samuael: Screenshot 2025-12-11 at 9 54 39 in the morning

gbjk: I'm not convinced we can do very much if we do hardcode that value, though.
We hardcode the max subs per connection because it helps us see when to scale up connections.
Knowing the documented limit for connections per ip locally would only be of benefit in giving an error message if they don't give one themselves.
And we'd need tests to check that the limit isn't changing (something I've done elsewhere when we've decided to clientside constants from their documenation).

TL;RA: Is there a benefit for us to do that?

gloriousCode: Id imagine it would stop all connections being cut server-side due to being over the limit. Without that, GCT would just loop endlessly connecting and being disconnected.
Happy for that to be deferred for when it impacts

edit: what am I on about? You already added max connections 😆

shazbert: @samuael Yes I did consider and for now out of scope. I don't really want to introduce a ceiling until something breaks and it's required.

exchange/websocket/connection_test.go
```diff
@@ -59,3 +60,11 @@ func TestIncomingWithData(t *testing.T) {
 	require.Len(t, ch, 1, "must have one item in channel")
 	assert.Equal(t, []byte("test"), <-ch)
 }
+
+func TestConnectionSubscriptions(t *testing.T) {
+	t.Parallel()
+	ws := &connection{}
+	require.Nil(t, ws.Subscriptions())
+	ws.subscriptions = subscription.NewStore()
+	require.NotNil(t, ws.Subscriptions())
```
gloriousCode, line 70:
```suggestion
	require.NotNil(t, ws.Subscriptions())
	assert.Equal(t, ws.subscriptions, ws.Subscriptions())
```

exchange/websocket/manager.go
```diff
@@ -155,13 +154,13 @@ type ManagerSetup struct {
 	RateLimitDefinitions request.RateLimitDefinitions
 }
 
-// connectionWrapper contains the connection setup details to be used when
-// attempting a new connection. It also contains the subscriptions that are
-// associated with the specific connection.
-type connectionWrapper struct {
+// websocket contains the connection setup details to be used when attempting a new connection. It also contains the
+// global subscription store that are associated with the connections. Each connection will have its own subscription
```
gloriousCode:
```suggestion
// websocket contains the connection setup details to be used when attempting a new connection. Its subscription store
// knows of all subscriptions for each connection. Each connection will have its own subscription
```
Huge change request, okay if you defer to another PR

I just thought the use of `global` wasn't truly accurate as its not a global. But Im being lame and pedantic 💁‍♂️

exchange/websocket/manager.go
gloriousCode: Good news, things are better than when I wrote that comment 🌈 separating them takes more lines and is more ugly.
Only I'd be suggesting is renaming it to `addNewConnectionAndSubscribe`

exchanges/okx/okx_websocket.go
```diff
@@ -236,55 +235,19 @@ var subscriptionNames = map[string]string{
 }
 
 // WsConnect initiates a websocket connection
-func (e *Exchange) WsConnect() error {
```
gloriousCode, line 239: At this kind of scale its harder to pinpoint exactly why this is failing as much as it is:
image
Its pages and pages of this ^ This seems to specifically affect SPOT orderbooks.
I'm happy for this to be raised as a separate issue that is starting to show as a result of this scaling.

exchanges/okx/okx_websocket.go
```diff
@@ -328,123 +291,77 @@ func (e *Exchange) authenticateConnection(ctx context.Context, conn websocket.Co
...
+			if err != nil {
+				return nil, err
+			}
+			if len(chunk) > maxConnByteLen {
+				// Remove last added channel and argument as they exceed max byte length
+				channels = channels[:len(channels)-1]
+				arguments = arguments[:len(arguments)-1]
+				subs = subs[i:]
+				break
+			}
```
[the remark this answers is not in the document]
gloriousCode: While I appreciate this one, we'll discover that OKX isn't working pretty quickly as a result of this

exchanges/okx/ratelimit.go
```diff
@@ -16,7 +16,7 @@ const (
 )
```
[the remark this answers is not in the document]
gloriousCode: Why did you make this change @shazbert ? It does not seem to be about `RateLimitNotRequired` since `bybit` has that without adding `+1`

exchange/websocket/manager.go
samuael: Thank you all for contributing your thoughts about this. I also don't think it is urgent, but I brought this up because I thought it would be easier to do it now. However, I also agree on the idea of bringing it to attentions when it is necessary. @shazbert 👍 .

exchange/websocket/connection_test.go
```diff
@@ -67,4 +67,5 @@ func TestConnectionSubscriptions(t *testing.T) {
 	require.Nil(t, ws.Subscriptions())
 	ws.subscriptions = subscription.NewStore()
 	require.NotNil(t, ws.Subscriptions())
+	assert.Equal(t, ws.subscriptions, ws.Subscriptions())
```
gbjk: 🚧 Change request:
```suggestion
	testsubs.EqualLists(t, ws.subscriptions, ws.Subscriptions())
```

gbjk (changes requested):
One small thing as a result of changes since last approval.

exchanges/okx/okx_websocket.go
shazbert: Deferred to: #2136

exchanges/okx/ratelimit.go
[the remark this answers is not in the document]
shazbert: Our ping handlers where using `request.Unset` which defaulted to `placeOrderEPL` potentially limiting that outbound ability. So just moved everything up by one so it either will error on REST or default to the websocket connection rate limiter. That was my intention anyway.

exchange/websocket/connection_test.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/c48eaf0e08a00e8972e4a21a2218368377e11219

exchange/websocket/manager.go
```diff
@@ -60,6 +60,7 @@ var (
 	errExchangeConfigEmpty                  = errors.New("exchange config is empty")
```
gloriousCode, line 61: Got a couple of house cleaning bits now that this work has more complexity. 
Now that connections scale, these public functions access `connections` without a lock and can be called asynchronously:
`UnsubscribeChannels`
`SubscribeToChannels`
`RemoveSubscriptions`
`AddSuccessfulSubscriptions`
`AddSubscriptions`
`FlushSubscriptions`

Some of these it may not apply exactly, but `FlushSubscriptions` which I point out below has no lock and could be called asynchronously and ruin your day

exchange/websocket/manager.go
```diff
@@ -1072,13 +1103,14 @@ func (m *Manager) GetConnection(messageFilter any) (Connection, error) {
...
 	}
 
-	for _, wrapper := range m.connectionManager {
-		if wrapper.setup.MessageFilter == messageFilter {
-			if wrapper.connection == nil {
-				return nil, fmt.Errorf("%s: %s %w associated with message filter: '%v'", m.exchangeName, wrapper.setup.URL, ErrNotConnected, messageFilter)
-			}
-			return wrapper.connection, nil
+	for _, ws := range m.connectionManager {
+		if ws.setup.MessageFilter != messageFilter {
```
gloriousCode, line 1081: OOS: Petition to change MessageFilter to `EndPointID`. Its definitely not a connection identifier as we now have a slice of them per endpoint. As you say in your own words, it doesn't filter a websocket message

exchange/websocket/manager.go
```diff
@@ -1073,12 +1114,13 @@ func (m *Manager) GetConnection(messageFilter any) (Connection, error) {
...
-			if wrapper.connection == nil {
-				return nil, fmt.Errorf("%s: %s %w associated with message filter: '%v'", m.exchangeName, wrapper.setup.URL, ErrNotConnected, messageFilter)
-			}
-			return wrapper.connection, nil
+		if wrapper.setup.MessageFilter != messageFilter {
+			continue
+		}
+		if len(wrapper.connectionsSubs) == 0 {
+			return nil, fmt.Errorf("%s: %s %w associated with message filter: '%v'", m.exchangeName, wrapper.setup.URL, ErrNotConnected, messageFilter)
 		}
```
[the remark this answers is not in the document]
gloriousCode: I think it would be good to leave a comment about returning [0]. It will be a confusing thing in future after we forget everything in the new year. Doesn't have to be this AI comment, but just describe how returning _a_ connection is important rather than a connection in which to inspect for specific subscriptions

exchange/websocket/subscriptions.go
```diff
@@ -280,58 +295,37 @@ func (m *Manager) FlushChannels() error {
 		if err != nil {
 			return err
 		}
-		return m.updateChannelSubscriptions(nil, m.subscriptions, newSubs)
+		return m.updateChannelSubscriptions(m.subscriptions, newSubs)
 	}
 
-	for x := range m.connectionManager {
-		if m.connectionManager[x].setup.SubscriptionsNotRequired {
+	for _, ws := range m.connectionManager {
```
gloriousCode, line 310: Lock, if `GenerateSubscrptions` causes deadlock, you definitely need to lock for `scaleConnectionsToSubscriptions`

exchange/websocket/manager.go
```diff
@@ -60,6 +60,7 @@ var (
 	errExchangeConfigEmpty                  = errors.New("exchange config is empty")
 	errCannotObtainOutboundConnection       = errors.New("cannot obtain outbound connection")
```
gloriousCode, line 62: Bug: `monitorFrame` has a pointer to a waitgroup as a paramter, but uses `m.wg.Done()` instead
```go
if wg != nil {
		defer m.Wg.Done()
	}
```
To
```go
if wg != nil {
    defer wg.Done()
}
```

gloriousCode, line 62: Bug:  `observeData` has a `dropped` `*int` but uses `dropped` in the log, use `*dropped` to show the value instead of an interpretation of the pointer address
```go
func TestButts(t *testing.T) {
	var d int = 42
	dd := &d
	t.Logf("%d", dd)
}
```

exchange/websocket/manager.go
gloriousCode, line 1081: future problem: If `messageFilter` ever gets more complex `!=` may not be sufficient in future and may become confusing as to why its not working.

exchange/websocket/subscriptions.go
```diff
@@ -28,9 +28,9 @@ func (m *Manager) UnsubscribeChannels(conn Connection, channels subscription.Lis
 	if len(channels) == 0 {
```
gloriousCode, line 29: cleanup: `errSubscriptionsExceedsLimit` comment is inaccurate and can be removed

exchange/websocket/subscriptions.go
```diff
@@ -28,9 +28,9 @@ func (m *Manager) UnsubscribeChannels(conn Connection, channels subscription.Lis
...
 		return nil // No channels to unsubscribe from is not an error
 	}
-	if wrapper, ok := m.connections[conn]; ok && conn != nil {
-		return m.unsubscribe(wrapper.subscriptions, channels, func(channels subscription.List) error {
-			return wrapper.setup.Unsubscriber(context.TODO(), conn, channels)
+	if ws, ok := m.connections[conn]; ok && conn != nil {
```
gloriousCode: Low stakes OOS: Wouldn't you want to know if a `nil` `conn` has been passed in? Or if a `conn` is passed in, but isn't part of the `manager`?
```go
	if m.useMultiConnectionManagement {
		if err := common.NilGuard(conn); err != nil {
			return err
		}
		ws, ok := m.connections[conn]
		if !ok {
			return fmt.Errorf("%w: %q", errConnectionNotFound, conn.GetURL())
		}
		return m.unsubscribe(ws.subscriptions, channels, func(channels subscription.List) error {
			return ws.setup.Unsubscriber(context.TODO(), conn, channels)
		})
	}
```

gloriousCode (changes requested):
Thanks for looking into the okx orderbook checksum issue. These are mostly cleanups for potential problems in future

exchange/websocket/manager.go
gloriousCode: heh, I forgot I required this code along time ago:
```go
if c.MessageFilter != nil && !reflect.TypeOf(c.MessageFilter).Comparable() {
			return errMessageFilterNotComparable
		}
```

So actually nevermind

exchange/websocket/manager.go
shazbert: 🔥

exchange/websocket/manager.go
gloriousCode: I'm happy to defer this to its own PR. The async panic potential is preexisting. I think the solution needs to be considered on its own and not amongst the rest of this

shazbert: Will make an issue for a future PR

exchanges/okx/okx_wrapper.go
```diff
@@ -65,7 +66,7 @@ func (e *Exchange) SetDefaults() {
 		log.Errorln(log.ExchangeSys, err)
 	}
 
-	// TODO: Disabled until spread/business websocket is implemented
+	// TODO: Remove when spread/business websocket templates across connections are completed
 	if err := e.DisableAssetWebsocketSupport(asset.Spread); err != nil {
```
gloriousCode: Is this needed anymore?
```
[DEBUG] | WEBSOCKET | 14/01/2026 14:39:05 | Okx wss://ws.okx.com:8443/ws/v5/business: Message received: {"arg":{"channel":"sprd-books5","sprdId":"LTC-USDT_LTC-USDT-SWAP"},"data":[{"asks":[["-0.04","644.8","1"]],"bids":[["-0.08","100.2","1"]],"ts":"1768361945106","seqId":1768271692951126}]}
[DEBUG] | SYNC | 14/01/2026 14:39:05 | Okx Orderbook sync complete LTC-USDT_LTC-USDT-SWAP [12/12].
[INFO]  | SYNC | 14/01/2026 14:39:05 | Okx websocket LTC-USDT_LTC-USDT-SWAP SPREAD ORDERBOOK: Bids len: 1 Amount: 100.200000 LTC. Total value: -8.016 Asks len: 1 Amount: 644.800000 LTC. Total value: -25.791999999999998
```

Resolve if OOS

exchange/websocket/manager.go
```diff
@@ -388,6 +384,10 @@ func (m *Manager) getConnectionFromSetup(c *ConnectionSetup) *connection {
 		// the match from the global match and have a match per connection.
 		match = NewMatch()
 	}
+	rateLimit := c.RateLimit
```
gloriousCode, line 386: Any chance this function could be renamed to `createConnectionFromSetup`?  Its not getting a connection

exchange/websocket/manager.go
```diff
@@ -602,6 +583,55 @@ func (m *Manager) connect() error {
 	return subscriptionError
 }
 
+func (m *Manager) connectAndSubscribe(ctx context.Context, ws *websocket, subs subscription.List) error {
```
gloriousCode: could this be renamed to `createConnectAndSubscribe`? Or just `createConnectSubscribe`

I got way too hung up on the func `scaleConnections` func when this is the big important one

gloriousCode: Could this function have a dedicated unit test?

shazbert: Done sir: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/4b469a5541c2db65bbaec80fa1c15c7180092141

exchange/websocket/subscriptions.go
```diff
@@ -350,3 +353,137 @@ func (m *Manager) updateChannelSubscriptions(c Connection, store *subscription.S
...
+		if err := conn.Shutdown(); err != nil {
+			log.Warnf(log.WebsocketMgr, "%v websocket: failed to shutdown connection: %v", m.exchangeName, err)
+		}
+	}
+	ws.connections = clean
+	return nil
+}
+
+// unsubscribeFromConnection unsubscribes from a connection and removes subscriptions from the store
+func (m *Manager) unsubscribeFromConnection(conn Connection, subs subscription.List) (subscription.List, error) {
```
gloriousCode: This and you unsubuscribe is confusing

exchanges/okx/okx_types.go
```diff
@@ -3196,6 +3197,7 @@ type SubscriptionInfo struct {
 type WSSubscriptionInformationList struct {
 	Operation string             `json:"op"`
 	Arguments []SubscriptionInfo `json:"args"`
+	subs      subscription.List  `json:"-"`
```
gloriousCode: unexported properties don't get json'd in the first place. I don't mind that you're being explicit, but its not needed as a FYI
edit: maybe sonic is different

exchange/websocket/subscriptions.go
```diff
@@ -350,3 +353,137 @@ func (m *Manager) updateChannelSubscriptions(c Connection, store *subscription.S
...
+		delete(m.connections, conn)
+		if err := conn.Shutdown(); err != nil {
+			log.Warnf(log.WebsocketMgr, "%v websocket: failed to shutdown connection: %v", m.exchangeName, err)
+		}
+	}
+	ws.connections = clean
+	return nil
+}
+
+// unsubscribeFromConnection unsubscribes from a connection and removes subscriptions from the store
```
gloriousCode:
```suggestion
// unsubscribeFromConnection unsubscribes for a connection and removes subscriptions from the connection's store
```

exchange/websocket/subscriptions.go
```diff
@@ -350,3 +353,137 @@ func (m *Manager) updateChannelSubscriptions(c Connection, store *subscription.S
...
+	missing := store.Missing(subs)
+	for _, r := range remove {
+		if err := store.Remove(r); err != nil {
+			return nil, err
+		}
+	}
+	return missing, nil
+}
+
+// subscribeToConnection subscribes to a connection and adds subscriptions to the store
```
gloriousCode:
```suggestion
// subscribeToConnection subscribes for a connection and adds subscriptions to the connection's store
```

exchange/websocket/subscriptions.go
```diff
@@ -350,3 +353,137 @@ func (m *Manager) updateChannelSubscriptions(c Connection, store *subscription.S
 	}
 	return nil
 }
+
+// scaleConnectionsToSubscriptions scales connections to subscriptions based off current subscription list
```
gloriousCode:
```suggestion
// scaleConnectionsToSubscriptions scales connections to subscriptions based off current subscription list and subscription limit
```

gloriousCode (approved):
I've just left some tiny clarification comments, but otherwise I think this works. tACK!

exchanges/okx/okx_types.go
shazbert: Nice, will change; if it wiggins (doubtful) I will fix it later.

shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/019cca7ee7c0f1ac6c5a1e3e7ea7fa06d7acf275

exchange/websocket/manager_test.go
```diff
@@ -1046,7 +1046,7 @@ func TestConnectionShutdown(t *testing.T) {
 	t.Parallel()
 	wc := connection{shutdown: make(chan struct{})}
 	err := wc.Shutdown()
-	assert.ErrorIs(t, err, common.ErrNilPointer, "Shutdown should error correctly")
+	assert.NoError(t, err, "Shutdown should not error when connection is nil")
```
gloriousCode:
```suggestion
	assert.NoError(t, err, "Shutdown should not error when connection.Connection is nil")
```
A small difference only because `connection` is not nil

thrasher-:
Master with `-exchanges=okx -enableallpairs` as expected:
```
[ERROR] | WEBSOCKET | 22/01/2026 14:06:57 | subscriptions exceeds limit: current subscriptions: 0, incoming subscriptions: 7884, max subscriptions per connection: 240 - please reduce enabled pairs
```

This PR scales across connections but I did notice this which I'm not getting on master (yet):

```
[ERROR] | WEBSOCKET | 22/01/2026 14:05:35 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx BARD-USDT spot Reason: [orderbook data integrity compromised, Okx BARD-USDT spot checksum mismatch: expected '3926326186', got '1517177850']
[WARN]  | SYNC | 22/01/2026 14:05:35 | Okx LTC-USD SPOT: No ticker update after 1m0s, switching from websocket to rest
[ERROR] | WEBSOCKET | 22/01/2026 14:05:35 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx BARD-USDT spot Reason: [orderbook data integrity compromised, Okx BARD-USDT spot checksum mismatch: expected '3926326186', got '1517177850']
[WARN]  | SYNC | 22/01/2026 14:05:35 | Okx MON-USDT-SWAP PERPETUALSWAP: No ticker update after 1m0s, switching from websocket to rest
[WARN]  | SYNC | 22/01/2026 14:05:35 | Okx MON-USDT-SWAP PERPETUALSWAP: No orderbook update after 1m0s, switching from websocket to rest
[ERROR] | WEBSOCKET | 22/01/2026 14:05:35 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx BARD-USDT spot Reason: [orderbook data integrity compromised, Okx BARD-USDT spot checksum mismatch: expected '3926326186', got '1517177850']
[WARN]  | SYNC | 22/01/2026 14:05:35 | Okx LUNA-USD SPOT: No ticker update after 1m0s, switching from websocket to rest
[ERROR] | WEBSOCKET | 22/01/2026 14:05:35 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx BARD-USDT spot Reason: [orderbook data integrity compromised, Okx BARD-USDT spot checksum mismatch: expected '3926326186', got '1517177850']
[WARN]  | SYNC | 22/01/2026 14:05:35 | Okx MAGIC-USD SPOT: No ticker update after 1m0s, switching from websocket to rest
[WARN]  | SYNC | 22/01/2026 14:05:35 | Okx MOODENG-USDT-SWAP PERPETUALSWAP: No ticker update after 1m0s, switching from websocket to rest
[ERROR] | WEBSOCKET | 22/01/2026 14:05:35 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx ASTER-USDT spot Reason: [orderbook data integrity compromised, Okx ASTER-USDT spot checksum mismatch: expected '777506638', got '728563276']
[WARN]  | SYNC | 22/01/2026 14:05:35 | Okx MOODENG-USDT-SWAP PERPETUALSWAP: No orderbook update after 1m0s, switching from websocket to rest
[ERROR] | WEBSOCKET | 22/01/2026 14:05:35 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx BARD-USDT spot Reason: [orderbook data integrity compromised, Okx BARD-USDT spot checksum mismatch: expected '3926326186', got '1517177850']
[WARN]  | SYNC | 22/01/2026 14:05:36 | Okx HMSTR-USDT MARGIN: No ticker update after 1m0s, switching from websocket to rest
[ERROR] | WEBSOCKET | 22/01/2026 14:05:36 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx BARD-USDT spot Reason: [orderbook data integrity compromised, Okx BARD-USDT spot checksum mismatch: expected '3926326186', got '1517177850']
[ERROR] | WEBSOCKET | 22/01/2026 14:05:36 | exchange Okx websocket error - connection URL:[wss://ws.okx.com:8443/ws/v5/public] error: Okx ASTER-USDT spot Reason: [orderbook data integrity compromised, Okx ASTER-USDT spot checksum mismatch: expected '777506638', got '728563276']
```

shazbert:
@dazi005 Can you please check this branch out when you have a moment just to make sure everything is working for you?

dazi005:
> @dazi005 Can you please check this branch out when you have a moment just to make sure everything is working for you?

Okay, thank you so much for everything you've done. I will verify this branch and will contact you promptly if there are any issues.
image
image

shazbert:
> > @dazi005 Can you please check this branch out when you have a moment just to make sure everything is working for you?
> 
> Okay, thank you so much for everything you've done. I will verify this branch and will contact you promptly if there are any 


@dazi005 Fixed that potential panic issue with auth payloads: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/bd865228e6f496012db4046916e9a05844d9f5cc

If you want to recheck that would be awesome. I also need to fix up a blocked channel for business subscription tests, wondering if I should just drop it instead.

dazi005:
> > > @dazi005 Can you please check this branch out when you have a moment just to make sure everything is working for you?
> > 
> > 
> > Okay, thank you so much for everything you've done. I will verify this branch and will contact you promptly if there are any
> 
> @dazi005 Fixed that potential panic issue with auth payloads: [bd86522](https://github.com/thrasher-corp/gocryptotrader/commit/bd865228e6f496012db4046916e9a05844d9f5cc)
> 
> If you want to recheck that would be awesome. I also need to fix up a blocked channel for business subscription tests, wondering if I should just drop it instead.

Just now, I have tested it and everything is normal. The business you mentioned requires a business account, which I haven't used. If there are more important things, you can consider giving up first

exchanges/okx/okx_websocket.go
```diff
@@ -304,14 +304,18 @@ func (e *Exchange) Unsubscribe(ctx context.Context, conn websocket.Connection, c
...
 	}
-
 	for _, req := range requests {
-		if err := conn.SendJSONMessage(ctx, websocketRequestEPL, req); err != nil {
-			return err
+		// Zero outbound arguments can happen for when margin or spot and a subscription already exists, this will
+		// still create a connection but will technically be a no-op for tracking subscriptions.
+		// TODO: Handle no-op subscription tracking more gracefully as we don't need a connection for it.
+		if len(req.Arguments) != 0 {
+			if err := conn.SendJSONMessage(request.WithVerbose(ctx), websocketRequestEPL, req); err != nil {
```
gloriousCode: Should with verbose stay?

exchanges/okx/okx_test.go
```diff
@@ -6102,11 +6102,42 @@ func TestGenerateSubscriptions(t *testing.T) {
 	testsubs.EqualLists(t, exp, append(public, private...))
 }
 
+// TODO: Implement channel subscriptions for business ws and remove this test
 func TestBusinessWSCandleSubscriptions(t *testing.T) {
 	t.Parallel()
 	e := new(Exchange)
 	require.NoError(t, testexch.Setup(e), "Setup must not error")
-	require.NoError(t, e.Websocket.Connect())
+
```
gloriousCode, line 3910: Hmm 
--- PASS: TestBusinessWSCandleSubscriptions (23.87s)
I would like to request another looky for this test as I think 23 seconds seems excessive

exchanges/okx/okx_websocket.go
shazbert: nar

exchanges/okx/okx_test.go
shazbert: thats to do with the subscription rate limiting for 7 seconds per request 😆 for some reason

shazbert: Yeah so for the connection rate limiter in wrapper when setting up the new connection, it has 480 requests per hour which is 8 requests per minute so all requests afterwards are restricted by ~7 seconds. 

```
Okx wss://ws.okx.com:8443/ws/v5/business: Sending message: {"op":"subscribe","args":[{"channel":"sprd-public-trades","sprdId":"ETH-USDT_ETH-USD-250627"},{"channel":"sprd-books5","sprdId":"ETH-USDT_ETH-USD-250627"},{"channel":"sprd-tickers","sprdId":"ETH-USDT_ETH-USD-250627"},{"channel":"public-struc-block-trades"}]}
Okx wss://ws.okx.com:8443/ws/v5/business: Sending message: {"op":"subscribe","args":[{"channel":"candle1D","instId":"BTC-USDT"}]}
Okx wss://ws.okx.com:8443/ws/v5/business: Sending message: {"op":"subscribe","args":[{"channel":"mark-price-candle1M","instId":"ETH-USDT"}]}
Okx wss://ws.okx.com:8443/ws/v5/business: Sending message: {"op":"subscribe","args":[{"channel":"index-candle1H","instId":"OKB-USDT"}]}
```

So minimum after the first request outbound you are looking at 21 seconds at a minimum. 

I have merged them into one request so its around ~9s now to complete: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/e21b452ae4e9dfa0feef162e0da602d683dafb01

exchanges/okx/okx_test.go
```diff
@@ -6081,59 +6083,102 @@ func TestGenerateSubscriptions(t *testing.T) {
...
+		case []CandlestickMarkPrice:
+			if len(v) > 0 {
+				got = got.Add(v[0].Pair)
 			}
 		default:
 		}
 		return len(got) == 3
-	}, 4*time.Second, 100*time.Millisecond, "Should eventually get candles from the datahandler")
+	}
+	assert.Eventually(t, check, 5*time.Second, time.Nanosecond)
```
gloriousCode: I don't think the check needs to be every nanosecond really. Millisecond should be fine. Optional because its not going to affect much but maybe my cpu will get slightly warmer

shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2109/commits/a3fbcee8f4979be70e5eda21068eb5e764501a41

Changed but if we do increase the sub count this might get smashed and extend the test time. It blocked on the reading the websocket data channel.

gloriousCode: I really don't think that will and we can adapt when you do that

thrasher-:
Two separate things I've found when testing: jobBuffer of 5000 gets cooked and fails to relay any message (OKX again with all pairs enabled), but it's clearly scaling the connections and pairs. Need to overhaul engine websocketDataReceiver / syncer / OB evaluations to be async instead of processing things serially
image

PR #2115, opened 2025-11-25
mock/recording: Update query checks/comparison and payload saving

samuael:
This PR fixes an issue where request body arrays could not be correctly recorded or fetched from mock files. The underlying problem was caused by attempting to unmarshal a JSON array into a url.Values (which expects a map[string]interface{}), resulting in the error:
`place order failed: authenticated request failed mock recording failure json: cannot unmarshal array into Go value of type map[string]interface {}`.

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test X

## Checklist


thrasher-:
@codex please review

exchanges/mock/common_test.go
```diff
@@ -87,3 +88,52 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
+				SomeString: "hello, there",
+			},
+		},
+		{},
+	}
+	payload, err := json.Marshal(test1)
+	require.NoError(t, err, "Marshal must not error")
+
+	values, err := DeriveURLValsFromJSONSlice(payload)
+	require.NoError(t, err, "DeriveURLValsFromJSONMap must not error")
```
shazbert:
```suggestion
	values, err := DeriveURLValsFromJSONSlice(payload)
	require.NoError(t, err, "DeriveURLValsFromJSONSlice must not error")
```

exchanges/mock/recording_test.go
```diff
@@ -203,4 +205,30 @@ func TestHTTPRecord(t *testing.T) {
 	}
 	err = HTTPRecord(response, "mock", content, 4)
 	require.NoError(t, err, "HTTPRecord must not error")
+
+	fullURL, err := url.Parse("https://api.abc.com/test/payload")
+	require.NoError(t, err)
+	assert.NotNil(t, fullURL)
+
+	println("Content: ", string(content))
```
shazbert: we could RM this

shazbert (changes requested):
Nice work, not much from me!

exchanges/mock/recording.go
```diff
@@ -185,21 +183,48 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
-							// data
-							mockResponses = slices.Delete(mockResponses, i, i+1)
+							if len(reqVals) != len(mockVals) {
+								continue
+							}
 							found = true
+							for j := range reqVals {
+								found = found && MatchURLVals(reqVals[j], mockVals[j])
+							}
+							if found {
```
gbjk: 🚧 Change request:
This seems to be saying "If all elements match".
You're saying it's true, unless one of the elements doesn't match, and then it's false.

That seems like it's just slices.EqualFunc

```suggestion
							if found = slices.EqualFunc(reqVals, mockVals, MatchURLVals); found {
```

exchanges/mock/recording.go
```diff
@@ -185,21 +183,48 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
+							}
+
+							mockVals, jErr := DeriveURLValsFromJSONMap([]byte(mockResponses[i].BodyParams))
+							if jErr != nil {
+								return jErr
+							}
+
+							if MatchURLVals(reqVals, mockVals) {
+								// if found will delete instance and overwrite with new
+								// data
```
gbjk: Same as above, don't know what this means.

exchanges/mock/recording.go
```diff
@@ -185,21 +183,48 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
+							if jErr != nil {
+								return jErr
+							}
+
+							mockVals, jErr := DeriveURLValsFromJSONMap([]byte(mockResponses[i].BodyParams))
+							if jErr != nil {
+								return jErr
+							}
+
+							if MatchURLVals(reqVals, mockVals) {
```
gbjk:
```suggestion
							if found = MatchURLVals(reqVals, mockVals); found {
```

exchanges/mock/recording.go
```diff
@@ -185,21 +183,48 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
+								// if found will delete instance and overwrite with new
+								// data
+								mockResponses = slices.Delete(mockResponses, i, i+1)
+								found = true
```
gbjk:
```suggestion
```

gbjk (changes requested):
Looks good. Just one simplification

shazbert (changes requested):
@samuael linter issues then things look good. thanks!

exchanges/mock/recording.go
```diff
@@ -177,29 +174,50 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
+							mockVals, jErr := DeriveURLValsFromJSONMap([]byte(mockResponses[i].BodyParams))
+							if jErr != nil {
+								return jErr
+							}
+
+							if found = MatchURLVals(reqVals, mockVals); found {
+								// if found will delete instance and overwrite with new
+								// data
+								mockResponses = slices.Delete(mockResponses, i, i+1)
+							}
```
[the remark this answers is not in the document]
samuael: This is a valid comment. However, as far as my knowledge there is not a single exchange return a primitive type data as a JSON payload for REST requests. So, I believe, until we encounter such exchange and response, we should hold on to this way of slice and map JSON payload handling approach only.

exchanges/mock/common.go
```diff
@@ -63,8 +81,7 @@ func DeriveURLValsFromJSONMap(payload []byte) (url.Values, error) {
 		case map[string]any, []any, nil:
 			vals.Add(k, fmt.Sprintf("%v", val))
 		default:
-			log.Println(reflect.TypeOf(val))
-			return vals, errors.New("unhandled conversion type, please add as needed")
+			return vals, fmt.Errorf("unhandled conversion type: %v, please add as needed", reflect.TypeOf(val))
```
gloriousCode:
```suggestion
			return vals, fmt.Errorf("unhandled conversion type: %T, please add as needed", val)
```

exchanges/mock/common_test.go
```diff
@@ -41,30 +41,41 @@ func TestMatchURLVals(t *testing.T) {
 	}
 }
 
+type class struct {
```
gloriousCode: I understand its for a test and its pre-existing, but lets rename this to something more constructive

exchanges/mock/common_test.go
```diff
@@ -87,3 +98,72 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
 		assert.Equalf(t, test2[key], val[0], "DeriveURLValsFromJSON should return the correct value for %s", key)
 	}
 }
+
+func TestDeriveURLValsFromJSONSlice(t *testing.T) {
+	t.Parallel()
+	_, err := DeriveURLValsFromJSONSlice([]byte(``))
```
gloriousCode:
```go
t.Parallel()
	_, err := DeriveURLValsFromJSONSlice([]byte(``))
	require.NoError(t, err, "DeriveURLValsFromJSONSlice must not error")

	tcs := []struct {
		Things     []string `json:"things"`
		Data       class    `json:"data"`
		Counter    int      `json:"counter"`
		IsEvenNum  bool     `json:"numbers"`
		Number     float64  `json:"number"`
		SomeString string   `json:"somestring"`
	}{
		{
			Things: []string{"hello", "world"},
			Data: class{
				Counter:    1,
				Numbers:    []int{1, 3, 3, 7, 9},
				Number:     3.14,
				SomeString: "hello, peoples",
			},
			Counter:    1,
			IsEvenNum:  false,
			Number:     3.14,
			SomeString: "Test case 1",
		},
		{
			Things: []string{"hello", "thrasher"},
			Data: class{
				Counter:    2,
				Numbers:    []int{1, 9, 9, 9},
				Number:     3.14529,
				SomeString: "hello, team",
			},
			IsEvenNum:  true,
			Number:     3,
			SomeString: "Test case 2",
		},
		{
			Things: []string{"hello", "Ethiopia"},
			Data: class{
				Counter:    3,
				Numbers:    []int{2, 0, 1, 8},
				Number:     2018,
				SomeString: "hello, there",
			},
			IsEvenNum:  true,
			SomeString: "Test case 3",
			Number:     3,
		},
		{
			SomeString: "Test case 4",
		},
		{},
	}
	payload, err := json.Marshal(tcs)
	require.NoError(t, err, "Marshal must not error")

	values, err := DeriveURLValsFromJSONSlice(payload)
	require.NoError(t, err, "DeriveURLValsFromJSONSlice must not error")
	assert.Len(t, values, len(tcs))

	for i := range tcs {
		t.Run(tcs[i].SomeString, func(t *testing.T) {
			t.Parallel()
			elemPayload, err := json.Marshal(tcs[i])
			require.NoError(t, err, "Marshal must not error")

			val, err := DeriveURLValsFromJSONMap(elemPayload)
			require.NoError(t, err, "DeriveURLValsFromJSONMap must not error")
			assert.True(t, MatchURLVals(values[i], val), "MatchURLVals should be true")
		})
	}
```

exchanges/mock/recording.go
```diff
@@ -177,29 +174,49 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
 
-						if MatchURLVals(reqVals, mockVals) {
-							// if found will delete instance and overwrite with new
-							// data
-							mockResponses = slices.Delete(mockResponses, i, i+1)
-							found = true
+							if len(reqVals) != len(mockVals) {
+								continue
+							}
```
gloriousCode: Somewhat redundant with the `slices.EqualFunc` below calling `MatchURLVals` where the first line of `MatchURLVals` is 
```go
if len(v1) != len(v2) {
		return false
	}
```

Can remove

gloriousCode (changes requested):
Nice work Sam

shazbert (changes requested):
not much from me, just merge master and address scotts nits.

exchanges/mock/recording.go
```diff
@@ -177,29 +174,46 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
 						}
 
 					case applicationJSON, textPlain:
-						reqVals, jErr := DeriveURLValsFromJSONMap([]byte(body))
-						if jErr != nil {
-							return jErr
-						}
+						trimmedStoredBody := strings.TrimSpace(mockResponses[i].BodyParams)
+						trimmedRequestBody := strings.TrimSpace(body)
+						if strings.HasPrefix(trimmedStoredBody, "[") && strings.HasPrefix(trimmedRequestBody, "[") {
```
gloriousCode: A slightly modified version:
https://gist.github.com/gloriousCode/89893d62886fc18034873b1bd8bdde8a
It has an error if something is unexpected. It also changes your test a bit

gloriousCode (changes requested):
I think this is good to go. I made a diff, have a look at see if you want it. It just makes a couple of things more explicit. But if you don't think its necessary, I'm happy to approve this

exchanges/mock/recording.go
samuael: Diff applied 👍

exchanges/mock/recording.go
```diff
@@ -177,29 +179,53 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
 
 					case applicationJSON, textPlain:
-						reqVals, jErr := DeriveURLValsFromJSONMap([]byte(body))
-						if jErr != nil {
-							return jErr
-						}
+						trimmedStoredBody := strings.TrimSpace(mockResponses[i].BodyParams)
+						trimmedRequestBody := strings.TrimSpace(body)
+						storedShape := getJSONBodyShape(trimmedStoredBody)
+						requestShape := getJSONBodyShape(trimmedRequestBody)
```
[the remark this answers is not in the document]
gloriousCode: Hi Sam, check out this patch. See if it works for you: https://gist.github.com/gloriousCode/17ececb097c0dcfa473043b1dbe9643d

Bug fix: mock playback now supports JSON array request bodies.

Old behavior: array bodies were recorded, but playback parsed JSON as object-only, so matching could fail.
New behavior: playback detects JSON shape and matches arrays with array logic, objects with existing object logic.
Covered both application/json and text/plain (X-Gemini-Payload).
Added regression tests to show old path fails and new path succeeds.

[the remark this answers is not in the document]
samuael: Ohh. Thank you so much for the patch sir. I was confused for I did not read/understand the review comment properly.

Patch applied.

gloriousCode:
Looks like I broke lint in the patch for you. I'm happy to approve once build issues are sorted 🌈

gloriousCode (approved):
tACK!

exchanges/mock/common_test.go
```diff
@@ -87,3 +98,78 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
+		{
+			Things: []string{"hello", "Ethiopia"},
+			Data: multiTypeHolder{
+				Counter:    3,
+				Numbers:    []int{2, 0, 1, 8},
+				Number:     2018,
+				SomeString: "hello, there",
+			},
+			IsEvenNum:  true,
+			SomeString: "case 3",
```
cranktakular: This third case seems kind of unneceessary, it's not really introducing any new testable features that aren't covered by the previous two. Rearranging the order these fields are populated doesn't even matter past the marshaller (which we're not testing).

Either delete this one, or the second case.

exchanges/mock/common_test.go
```diff
@@ -87,3 +98,78 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
+
+	for i := range tcs {
+		t.Run("Test :"+tcs[i].SomeString, func(t *testing.T) {
+			t.Parallel()
+			elemPayload, err := json.Marshal(tcs[i])
+			require.NoError(t, err, "Marshal must not error")
+
+			val, err := DeriveURLValsFromJSONMap(elemPayload)
+			require.NoError(t, err, "DeriveURLValsFromJSONMap must not error")
+			assert.True(t, MatchURLVals(values[i], val), "MatchURLVals should be true")
```
cranktakular, line 176: I'm not a fan of tests where the validation basically recreates the function being tested. Since that means that it can't actually detect most incorrect outputs from the function.

In some cases that can be necessary, but I think in this case it might be practical to manually create the url.Values you expect.

But if it's not, you can just ignore this.

exchanges/mock/common.go
```diff
@@ -40,15 +40,38 @@ func MatchURLVals(v1, v2 url.Values) bool {
 	return true
 }
 
+// DeriveURLValsFromJSONSlice converts a JSON array into a slice of url.Values by processing each array element as a JSON object
+func DeriveURLValsFromJSONSlice(payload []byte) ([]url.Values, error) {
```
cranktakular: Really nitpicky, but it's probably better to call this DeriveURLValsFromJSONArray. Since JSON technically doesn't have slices.

exchanges/mock/common.go
```diff
@@ -40,15 +40,38 @@ func MatchURLVals(v1, v2 url.Values) bool {
...
+		return []url.Values{}, nil
+	}
+	var intermediary []json.RawMessage
+	if err := json.Unmarshal(payload, &intermediary); err != nil {
+		return nil, err
+	}
+
+	vals := make([]url.Values, len(intermediary))
+	for i := range intermediary {
+		result, err := DeriveURLValsFromJSONMap(intermediary[i])
```
cranktakular: This function returns an error at this step if the payload is an array of arrays.

If that isn't acceptable, a workaround should be found.

If that is acceptable, that may merit a mention in the function comment, and should be included in the test for more coverage.

exchanges/mock/common.go
```diff
@@ -61,10 +84,13 @@ func DeriveURLValsFromJSONMap(payload []byte) (url.Values, error) {
 		case float64:
 			vals.Add(k, strconv.FormatFloat(val, 'f', -1, 64))
 		case map[string]any, []any, nil:
-			vals.Add(k, fmt.Sprintf("%v", val))
+			b, err := json.Marshal(val)
```
cranktakular, line 96: Probably out of scope for this PR, but I think this indicates a fundamental flaw in this approach. If we're going to have to marshal (and presumably eventually unmarshal) some url.Values, why not just use json the whole way through (or even just for requests where that's more appropriate)?

exchanges/mock/common_test.go
```diff
@@ -87,3 +98,78 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
+	_, err := DeriveURLValsFromJSONSlice([]byte(``))
+	require.NoError(t, err, "DeriveURLValsFromJSONSlice must not error")
+
+	tcs := []struct {
+		Things     []string        `json:"things"`
+		Data       multiTypeHolder `json:"data"`
+		Counter    int             `json:"counter"`
+		IsEvenNum  bool            `json:"numbers"`
+		Number     float64         `json:"number"`
+		SomeString string          `json:"somestring"`
```
cranktakular: This is the same struct as is used in TestDeriveURLValsFromJSON; you should probably pull it out instead of defining it twice.

exchanges/mock/common.go
```diff
@@ -40,15 +40,38 @@ func MatchURLVals(v1, v2 url.Values) bool {
...
 }
 
+// DeriveURLValsFromJSONSlice converts a JSON array into a slice of url.Values by processing each array element as a JSON object
+func DeriveURLValsFromJSONSlice(payload []byte) ([]url.Values, error) {
+	if len(payload) == 0 {
+		return []url.Values{}, nil
+	}
+	var intermediary []json.RawMessage
+	if err := json.Unmarshal(payload, &intermediary); err != nil {
+		return nil, err
```
cranktakular, line 54: Please get coverage for this line.

exchanges/mock/recording_test.go
```diff
@@ -37,27 +39,64 @@ func TestGetFilteredURLVals(t *testing.T) {
...
+		DataPointer *checkclass `json:"datapointer,omitempty"`
+		Login       int         `json:"login,omitempty"`
+		IsEvenNum   bool        `json:"pass,omitempty"`
+		Balance     float64     `json:"bsb,omitempty"`
+		RealName    string      `json:"real_name,omitempty"`
+	}
+	inputs := []struct {
+		in  any
+		exp []byte
+		err error
```
cranktakular: This field seems unused. Deleting it causes no errors when compiling or running these tests.

exchanges/mock/recording.go
```diff
@@ -177,29 +179,53 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
+						switch {
+						case storedShape == jsonBodyArray && requestShape == jsonBodyArray:
+							reqVals, jErr := DeriveURLValsFromJSONSlice([]byte(body))
+							if jErr != nil {
+								return jErr
```
cranktakular, line 199: Is it feasible to get test coverage for this line?

exchanges/mock/recording.go
```diff
@@ -177,29 +179,53 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
+							}
 
-						mockVals, jErr := DeriveURLValsFromJSONMap([]byte(mockResponses[i].BodyParams))
-						if jErr != nil {
-							return jErr
-						}
+							mockVals, jErr := DeriveURLValsFromJSONSlice([]byte(mockResponses[i].BodyParams))
+							if jErr != nil {
+								return jErr
```
cranktakular, line 204: Is it feasible to get test coverage for this line?

exchanges/mock/recording.go
```diff
@@ -177,29 +179,53 @@ func HTTPRecord(res *http.Response, service string, respContents []byte, mockDat
...
+							}
+
+							if found = MatchURLVals(reqVals, mockVals); found {
+								// if the incoming query matches an existing, we will delete the existing mock record and overwrite with new data
+								mockResponses = slices.Delete(mockResponses, i, i+1)
+							}
+						case storedShape != requestShape:
+							return fmt.Errorf("%w request=%s stored=%s", errMismatchedJSONBodyShape, requestShape, storedShape)
+						default:
+							return fmt.Errorf("%w request=%s stored=%s", errUnsupportedJSONBodyShape, requestShape, storedShape)
```
cranktakular, line 230: Is it feasible to get test coverage for this line?

exchanges/mock/server.go
```diff
@@ -165,14 +167,28 @@ func RegisterHandler(pattern string, mock map[string][]HTTPResponse, mux *http.S
...
-
-				payload, err := MatchAndGetResponse(httpResponses, reqVals, false)
-				if err != nil {
-					log.Fatal("Mock Test Failure - MatchAndGetResponse error ", err)
+				var payload json.RawMessage
+				switch getJSONBodyShape(strings.TrimSpace(string(readBody))) {
+				case jsonBodyArray:
+					reqVals, err := DeriveURLValsFromJSONSlice(readBody)
+					if err != nil {
+						log.Fatalf("DeriveURLValsFromJSONSlice Mock Test Failure - %v", err)
```
cranktakular, line 175: Is it feasible to get test coverage for this line?

exchanges/mock/server.go
```diff
@@ -165,14 +167,28 @@ func RegisterHandler(pattern string, mock map[string][]HTTPResponse, mux *http.S
...
+					}
+
+					payload, err = MatchAndGetResponseJSONSlice(httpResponses, reqVals, false)
+					if err != nil {
+						log.Fatal("Mock Test Failure - MatchAndGetResponseJSONSlice error ", err)
```
cranktakular, line 180: Is it feasible to get test coverage for this line?

exchanges/mock/server.go
```diff
@@ -189,14 +205,28 @@ func RegisterHandler(pattern string, mock map[string][]HTTPResponse, mux *http.S
...
-
-				payload, err := MatchAndGetResponse(httpResponses, reqVals, false)
-				if err != nil {
-					log.Fatal("Mock Test Failure - MatchAndGetResponse error ", err)
+				var payload json.RawMessage
+				switch getJSONBodyShape(strings.TrimSpace(string(jsonThings))) {
+				case jsonBodyArray:
+					reqVals, err := DeriveURLValsFromJSONSlice(jsonThings)
+					if err != nil {
+						log.Fatalf("Mock Test Failure - %v", err)
```
cranktakular, line 213: Is it feasible to get test coverage for this line?

exchanges/mock/server.go
```diff
@@ -189,14 +205,28 @@ func RegisterHandler(pattern string, mock map[string][]HTTPResponse, mux *http.S
...
+					}
+
+					payload, err = MatchAndGetResponseJSONSlice(httpResponses, reqVals, false)
+					if err != nil {
+						log.Fatal("Mock Test Failure - MatchAndGetResponseJSONSlice error ", err)
```
cranktakular, line 218: Is it feasible to get test coverage for this line?

exchanges/mock/server.go
```diff
@@ -286,5 +316,33 @@ func MatchAndGetResponse(mockData []HTTPResponse, requestVals url.Values, isQuer
...
+func MatchAndGetResponseJSONSlice(mockData []HTTPResponse, requestVals []url.Values, isQueryData bool) (json.RawMessage, error) {
+	for i := range mockData {
+		var data string
+		if isQueryData {
+			data = mockData[i].QueryString
+		} else {
+			data = mockData[i].BodyParams
+		}
+
+		if !json.Valid([]byte(data)) || getJSONBodyShape(strings.TrimSpace(data)) != jsonBodyArray {
```
cranktakular, line 329: Is the valid json check here really that useful? Most mock data that isn't valid json would happen to not start with a "[" (that character gets escaped in query strings), and that which does could just keep executing and hit an error.

Not too important, since at worst, this just wastes a little time in a part that isn't particularly time sensitive.

exchanges/mock/server.go
```diff
@@ -286,5 +316,33 @@ func MatchAndGetResponse(mockData []HTTPResponse, requestVals url.Values, isQuer
...
+			continue
+		}
+
+		mockVals, err := DeriveURLValsFromJSONSlice([]byte(data))
+		if err != nil {
+			return nil, err
```
cranktakular, line 335: Is it feasible to get test coverage for this line?

exchanges/mock/server.go
```diff
@@ -286,5 +316,33 @@ func MatchAndGetResponse(mockData []HTTPResponse, requestVals url.Values, isQuer
...
+	return nil, errNoDataMatched
+}
+
+// MatchAndGetResponseJSONSlice matches incoming request values against array
+// based body params from the mock data and returns the payload.
+func MatchAndGetResponseJSONSlice(mockData []HTTPResponse, requestVals []url.Values, isQueryData bool) (json.RawMessage, error) {
+	for i := range mockData {
+		var data string
+		if isQueryData {
+			data = mockData[i].QueryString
```
cranktakular: If the data's querydata, that applies to the entire function call. It will hit the continue statement on every loop (unless there's a json slice in this QueryString field, which sounds improper), and ultimately match no data, returning an error.

This both seems like odd functionality (shouldn't querydata still try to match with the requestVals?), and like functionality that could be better achieved differently (the if isQueryData check could come outside the for loop, and immediately return the error).

cranktakular (changes requested):
I've found some things I'd like changed

exchanges/mock/common_test.go
samuael: Skipped for now!

exchanges/mock/common.go
samuael: Makes sense 🥇

exchanges/mock/common.go
samuael: I think the newly added `DeriveURLValsFromJSONArrayAsMap` will solve this issue.

exchanges/mock/common.go
samuael: Great point. I think if a slight difference on the two values(Input and loaded from mock) could not cause a mismatch in the comparison, using a JSON comparison could be a good idea.

exchanges/mock/server.go
samuael: Execusimua for not applying this. Yes. What you said is correct and it doesn't make sense to have a query string payload with a slice of JSON payload.

Change applied. Thank you!

exchanges/mock/server.go
samuael: No. I don't see how I can reach there since it get caught in the `json.Valid` check before reaching there. Causing the loop to continue on that check.

exchanges/mock/server.go
samuael: The query payload check for slice bodies payload is removed and the isQueryData returns an error. Therefore, having the validity check with save time from further calling the `DeriveURLValsFromJSONArray` with the invalid slice data.

Screenshot 2026-04-07 at 6 48 21 in the evening

exchanges/mock/common.go
cranktakular: You gave this a thumbs up, but it doesn't seem to get coverage on my end.

samuael: test coverage added.

exchanges/mock/server.go
samuael: no.

exchanges/mock/server.go
samuael: These `log.Fatal` calls run inside HTTP handler goroutines in test-only mocks. Because `log.Fatal `invokes `os.Exit(1)`, it kills the test process, making in-process coverage impossible without refactoring (e.g., to `http.Error`) or using subprocess tests. The gap is intentional if reached, the test fails. Covering these lines would require changing the error-handling contract.

exchanges/mock/server.go
samuael: These `log.Fatal` calls run inside HTTP handler goroutines in test-only mocks. Because `log.Fatal` invokes `os.Exit(1)`, it kills the test process, making in-process coverage impossible without refactoring (e.g., to `http.Error`) or using subprocess tests. The gap is intentional if reached, the test fails. Covering these lines would require changing the error-handling contract.

exchanges/mock/server.go
samuael: These `log.Fatal` calls run inside HTTP handler goroutines in test-only mocks. Because `log.Fatal` invokes `os.Exit(1)`, it kills the test process, making in-process coverage impossible without refactoring (e.g., to `http.Error`) or using subprocess tests. The gap is intentional if reached, the test fails. Covering these lines would require changing the error-handling contract.

exchanges/mock/server.go
samuael: These `log.Fatal` calls run inside HTTP handler goroutines in test-only mocks. Because `log.Fatal` invokes `os.Exit(1)`, it kills the test process, making in-process coverage impossible without refactoring (e.g., to `http.Error`) or using subprocess tests. The gap is intentional if reached, the test fails. Covering these lines would require changing the error-handling contract.

exchanges/mock/common.go
```diff
@@ -40,19 +40,25 @@ func MatchURLVals(v1, v2 url.Values) bool {
 	return true
 }
 
-// DeriveURLValsFromJSONSlice converts a JSON array into a slice of url.Values by processing each array element as a JSON object
-func DeriveURLValsFromJSONSlice(payload []byte) ([]url.Values, error) {
+// DeriveURLValsFromJSONArray converts a JSON array into a slice of url.Values by processing each array element as a JSON object
```
gloriousCode, line 46: So I'm trying to test your work against Poloniex.
I am doing the following:
- Enable live testing
- Enable http recording
- Run all tests
- The following tests fail for the same reason:
  - `TestUpdateOrderExecutionLimits`
  - `TestGetTickers`
  - `TestGetTickers`
  - `TestGetTicker`
  - `TestGetBorrowRate`

```
        	            	mock recording failure invalid character 'S' after top-level value, request &{GET https://api.poloniex.com/markets/borrowRatesInfo HTTP/1.1 1 1 map[] <nil> <nil> 0 [] false api.poloniex.com map[] map[] <nil> map[]   <nil> <nil> <nil>  0x147a819110e0 <nil> [] map[]}: resp: &{200 OK 200 HTTP/2.0 2 0 map[Content-Type:[application/json] Date:[Tue, 05 May 2026 06:12:55 GMT] Vary:[accept-encoding] Via:[1.1 4d4befe10b69aac25e127b8999e718d2.cloudfront.net (CloudFront)] Web-Filter:[skip-authentication] X-Amz-Cf-Id:[TMCg-IC2NE1-mxgykG7nw2e-XYIAJ8m3C8dQtwT0AiQTKt-l1Kw46Q==] X-Amz-Cf-Pop:[SYD3-P3] X-Cache:[Miss from cloudfront]] 0x147a82c4cb60 -1 [] false true map[] 0x147a820b0280 0x147a81982300}
        	Test:       	TestGetBorrowRate
```

Is this within the scope of your changes?
These tests pass when testing against live with recording disabled

gloriousCode: Hello Sam. I have a fix!
```diff
diff --git a/exchanges/mock/recording.go b/exchanges/mock/recording.go
index 7d2897467..de4bb5226 100644
--- a/exchanges/mock/recording.go
+++ b/exchanges/mock/recording.go
@@ -1,6 +1,7 @@
 package mock
 
 import (
+       "bytes"
        "encoding/base64"
        "errors"
        "fmt"
@@ -25,6 +26,7 @@ const defaultDataSliceLimit = 5
 var (
        errMismatchedJSONBodyShape  = errors.New("mismatched JSON body shape")
        errUnsupportedJSONBodyShape = errors.New("unsupported JSON body shape")
+       errUnexpectedTrailingJSON   = errors.New("unexpected trailing JSON content")
 )
 
 // HTTPResponse defines expected response from the end point including request
@@ -313,7 +315,7 @@ func GetFilteredURLVals(vals url.Values, items Exclusion) string {
 // that should not be there.
 func CheckResponsePayload(data []byte, items Exclusion, mockDataSliceLimit int) ([]byte, error) {
        var intermediary any
-       if err := json.Unmarshal(data, &intermediary); err != nil {
+       if err := decodeLeadingJSONValue(data, &intermediary); err != nil {
                return nil, err
        }
 
@@ -325,6 +327,20 @@ func CheckResponsePayload(data []byte, items Exclusion, mockDataSliceLimit int)
        return json.MarshalIndent(payload, "", " ")
 }
 
+func decodeLeadingJSONValue(data []byte, target any) error {
+       decoder := json.NewDecoder(bytes.NewReader(data))
+       if err := decoder.Decode(target); err != nil {
+               return err
+       }
+
+       remainder := bytes.TrimSpace(data[decoder.InputOffset():])
+       if len(remainder) == 0 || bytes.Equal(remainder, []byte(`""`)) {
+               return nil
+       }
+
+       return fmt.Errorf("%w: %q", errUnexpectedTrailingJSON, remainder)
+}
+
 // Reflection consts
 const (
        Float64 = "float64"
diff --git a/exchanges/mock/recording_test.go b/exchanges/mock/recording_test.go
index 011888a4d..c216f311f 100644
--- a/exchanges/mock/recording_test.go
+++ b/exchanges/mock/recording_test.go
@@ -54,6 +54,7 @@ type checkclass struct {
 }
 
 func TestCheckResponsePayload(t *testing.T) {
+       t.Parallel()
        type someJSON struct {
                Secret      string      `json:"secret,omitempty"`
                Data        checkclass  `json:"data"`
@@ -105,6 +106,27 @@ func TestCheckResponsePayload(t *testing.T) {
        }
 }
 
+func TestCheckResponsePayloadTrailingData(t *testing.T) {
+       t.Parallel()
+
+       items, err := getExcludedItems()
+       require.NoError(t, err, "getExcludedItems must not error")
+
+       data, err := CheckResponsePayload([]byte(`[{"symbol":"BTC_USDT"}]""`), items, 5)
+       require.NoError(t, err, "CheckResponsePayload must not error with trailing payload data")
+       assert.JSONEq(t, `[{"symbol":"BTC_USDT"}]`, string(data), "CheckResponsePayload should preserve the leading valid JSON value")
+}
+
+func TestCheckResponsePayloadUnexpectedTrailingData(t *testing.T) {
+       t.Parallel()
+
+       items, err := getExcludedItems()
+       require.NoError(t, err, "getExcludedItems must not error")
+
+       _, err = CheckResponsePayload([]byte(`[{"symbol":"BTC_USDT"}]{"extra":true}`), items, 5)
+       require.ErrorIs(t, err, errUnexpectedTrailingJSON, "CheckResponsePayload must reject unexpected trailing JSON data")
+}
+
 func TestGetExcludedItems(t *testing.T) {
        t.Parallel()
        exclusionList, err := getExcludedItems()

```
Here is a breakdown
> Poloniex mock recording was failing because the recorder treated any response with extra trailing characters after valid JSON as fully invalid, even when the main payload was usable. Its been fixed by parsing the main JSON while only allowing the known trailing "" edge case (and rejecting other trailing data), then added tests and verified Poloniex plus lint/misc checks all pass.

shazbert: 🚀

gloriousCode: Sam. This was only a temporary thing.....

Im going to resolve it 😆

shazbert: 💨

gloriousCode (approved):
With the thing resolved, I'm happy with your changes. tACK

exchanges/mock/common.go
samuael: Thank you for the patch 🙇  I have applied the patch with an additional unit test to `decodeLeadingJSONValue`.

thrasher- (changes requested):
Looks like there's test failures which need to be fixed

exchanges/mock/common.go
```diff
@@ -61,10 +90,50 @@ func DeriveURLValsFromJSONMap(payload []byte) (url.Values, error) {
...
+
+// DeriveURLValsFromJSONArrayAsMap converts a JSON array into url.Values,
+// using indices as keys and stringified values
+func DeriveURLValsFromJSONArrayAsMap(payload []byte) (url.Values, error) {
+	vals := url.Values{}
+	if len(payload) == 0 {
+		return vals, nil
+	}
+	if getJSONBodyShape(strings.TrimSpace(string(payload))) != jsonBodyArray {
+		return vals, errJSONMapPayloadMustBeObject
```
cranktakular: This isn't quite correct. The payload must be an array, not an object.

Also, this logic is a little odd. As-is, it would progress even if the bodyshape was unknown.

exchanges/mock/common.go
```diff
@@ -61,10 +90,50 @@ func DeriveURLValsFromJSONMap(payload []byte) (url.Values, error) {
...
+		default:
+			return vals, fmt.Errorf("unhandled conversion type: %T, please add as needed", val)
+		}
+	}
+
+	return vals, nil
+}
+
+// DeriveURLValsFromJSONArrayAsMap converts a JSON array into url.Values,
+// using indices as keys and stringified values
```
cranktakular, line 110: Is it actually useful for the keys to be the indices? I find it hard to imagine where we'd practically want to use that.

exchanges/mock/common_test.go
```diff
@@ -86,4 +100,117 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
+		Number:     3,
+		SomeString: "case 3",
+	},
+	{
+		SomeString: "case 4",
+	},
+	{},
+}
+
+func TestDeriveURLValsFromJSONSlice(t *testing.T) {
```
cranktakular: The name of the test function has the outdated version of the function name.

exchanges/mock/common.go
```diff
@@ -61,10 +90,50 @@ func DeriveURLValsFromJSONMap(payload []byte) (url.Values, error) {
...
+		case string:
+			vals.Add(strconv.Itoa(k), val)
+		case bool:
+			vals.Add(strconv.Itoa(k), strconv.FormatBool(val))
+		case float64:
+			vals.Add(strconv.Itoa(k), strconv.FormatFloat(val, 'f', -1, 64))
+		case map[string]any, []any, nil:
+			b, err := json.Marshal(val)
+			if err != nil {
+				return vals, err
```
cranktakular, line 135: I think we usually avoid returning incomplete values when returning an error. Please consider just returning nil for this, and the following error case.

exchanges/mock/common_test.go
```diff
@@ -86,4 +100,117 @@ func TestDeriveURLValsFromJSON(t *testing.T) {
...
+
+	values, err := DeriveURLValsFromJSONArray(payload)
+	require.NoError(t, err, "DeriveURLValsFromJSONArray must not error")
+	assert.Len(t, values, 1)
+
+	// Testing JSON array as map
+	payload, err = json.Marshal(tcs)
+	require.NoError(t, err, "Marshal must not error")
+
+	value, err := DeriveURLValsFromJSONArrayAsMap(payload)
```
cranktakular, line 195: Why do you bother creating new values to capture these responses each time? Seems wasteful.

exchanges/mock/server.go
```diff
@@ -286,5 +318,30 @@ func MatchAndGetResponse(mockData []HTTPResponse, requestVals url.Values, isQuer
...
 	}
-	return nil, errors.New("no data could be matched")
+	return nil, errNoDataMatched
+}
+
+// MatchAndGetResponseJSONSlice matches incoming request values against array
+// based body params from the mock data and returns the payload.
+func MatchAndGetResponseJSONSlice(mockData []HTTPResponse, requestVals []url.Values, isQueryData bool) (json.RawMessage, error) {
+	if isQueryData {
+		return nil, errNoDataMatched
```
cranktakular: Why even have isQueryData as a parameter if it'd make the function immediately error out? Shouldn't that bool being set either cause it to pull the data from the QueryString, or just not exist and expect callers to not call it in a situation where isQueryData would be set to true?

exchanges/mock/server_test.go
```diff
@@ -90,3 +96,332 @@ func TestNewVCRServer(t *testing.T) {
...
+	w := httptest.NewRecorder()
+	MessageWriteJSON(w, http.StatusOK, nil)
+	assert.Equal(t, http.StatusOK, w.Code, "status code should be correct")
+	assert.JSONEq(t, applicationJSON, w.Header().Get(contentType), "Content-Type should be set")
+	assert.Empty(t, w.Body.String(), "nil data should produce empty body")
+
+	w = httptest.NewRecorder()
+	MessageWriteJSON(w, http.StatusCreated, map[string]string{"key": "value"})
+	assert.Equal(t, http.StatusCreated, w.Code, "status code should be correct")
+	assert.JSONEq(t, applicationJSON, w.Header().Get(contentType), "Content-Type should be set")
```
cranktakular: JSONEq is invalid here, and in the similar situation above. The applicationJSON string is not, itself, valid JSON.

exchanges/mock/common.go
cranktakular: This line still doesn't seem to get coverage on my end.

exchanges/mock/recording.go
cranktakular: You gave this a thumbs up, but it doesn't seem to get coverage on my end.

exchanges/mock/recording.go
cranktakular: You gave this a thumbs up, but it doesn't seem to get coverage on my end.

cranktakular (changes requested):
Still got a few changes I'd like to see.

exchanges/mock/common.go
samuael: The logic below only works for array, and this check is to make sure only an array JSON payload is accepted.

exchanges/mock/common.go
samuael: We use it to set it as a value to url.Values and use it to easily compare the values with _MatchURLVals()_

exchanges/mock/server_test.go
samuael: I used JSONEq because the linter was treating applicationJSON as a JSON and was causing a linter issue when we use the Equal() assertion method.

exchanges/mock/recording.go
samuael: Now it is reachable. 🫡

exchanges/mock/common.go
cranktakular: Oh, doh, I misread.

Still, the error is wrong. The payload must be an array.

cranktakular (approved):
Only got one minor change left that I want (changing one error). (EDIT: This has now been fixed)

Other than that, I can tACK this.

exchanges/mock/recording.go
```diff
@@ -334,6 +336,20 @@ const (
...
+		return err
+	}
+
+	remainder := bytes.TrimSpace(data[decoder.InputOffset():])
+	if len(remainder) == 0 || bytes.Equal(remainder, []byte(`""`)) {
+		return nil
+	}
+
+	return fmt.Errorf("%w: %q", errUnexpectedTrailingJSON, remainder)
+}
```
gloriousCode, line 362: Sorry Sam. We support both default golang json and sonic. There is an error when reviewing it.

While I would argue that this shouldn't matter, it is causing failures for the CLI build. I have made a compatable solution:
```go

func decodeLeadingJSONValue(data []byte, target any) error {
	reader := bytes.NewReader(data)
	decoder := json.NewDecoder(reader)
	if err := decoder.Decode(target); err != nil {
		return err
	}

	bufferedRemainder, err := io.ReadAll(decoder.Buffered())
	if err != nil {
		return fmt.Errorf("could not read buffered trailing JSON data: %w", err)
	}

	readerRemainder, err := io.ReadAll(reader)
	if err != nil {
		return fmt.Errorf("could not read trailing JSON data: %w", err)
	}

	remainder := bytes.TrimSpace(append(bufferedRemainder, readerRemainder...))
	if len(remainder) == 0 || bytes.Equal(remainder, []byte(`""`)) {
		return nil
	}

	return fmt.Errorf("%w: %q", errUnexpectedTrailingJSON, remainder)
}
```

Tests:
```go

func TestDecodeLeadingJSONValue(t *testing.T) {
	t.Parallel()

	t.Run("invalid json", func(t *testing.T) {
		t.Parallel()
		var v any
		var syntaxErr *json.SyntaxError
		require.ErrorAs(t, decodeLeadingJSONValue([]byte(`not-json`), &v), &syntaxErr, "decodeLeadingJSONValue must return a JSON syntax error")
	})

	tcs := []struct {
		name         string
		input        []byte
		wantErr      error
		wantAnyError bool
	}{
		{name: "valid json no trailing", input: []byte(`[{"symbol":"BTC_USDT"}]`)},
		{name: "trailing empty string", input: []byte(`[{"symbol":"BTC_USDT"}]""`)},
		{name: "trailing whitespace then empty string", input: []byte(`[{"symbol":"BTC_USDT"}]   ""`)},
		{name: "trailing non-empty string", input: []byte(`[{"symbol":"BTC_USDT"}]"extra"`), wantErr: errUnexpectedTrailingJSON},
		{name: "trailing empty string then extra object", input: []byte(`[{"symbol":"BTC_USDT"}]""{"extra":true}`), wantErr: errUnexpectedTrailingJSON},
		{name: "trailing empty string then malformed trailing json", input: []byte(`[{"symbol":"BTC_USDT"}]""{`), wantAnyError: true},
		{name: "unexpected trailing object", input: []byte(`[{"symbol":"BTC_USDT"}]{"extra":true}`), wantErr: errUnexpectedTrailingJSON},
		{name: "unexpected trailing value", input: []byte(`{"a":1}null`), wantErr: errUnexpectedTrailingJSON},
		{name: "malformed trailing json", input: []byte(`{"a":1}{`), wantAnyError: true},
	}
	for _, tc := range tcs {
			t.Parallel()
			var v any
			err := decodeLeadingJSONValue(tc.input, &v)
			switch {
			case tc.wantErr != nil:
				require.ErrorIs(t, err, tc.wantErr, "decodeLeadingJSONValue must return expected error")
			case tc.wantAnyError:
				require.Error(t, err, "decodeLeadingJSONValue must error")
			default:
				require.NoError(t, err, "decodeLeadingJSONValue must not error")
			}
		})
	}
}
```

gloriousCode (changes requested):
LAST THING I THINK SAM

gloriousCode (approved):
tACK!

thrasher- (changes requested):
Looks like this test needs to be fixed and then this looks good to go:

https://github.com/thrasher-corp/gocryptotrader/actions/runs/26292792129/job/77397232421?pr=2115#step:11:236

thrasher- (approved):
Thanks Sam!

Issue #2151, opened 2026-01-23
websocket/manager: internal/testing/exchange support for multi-connection


gloriousCode:
Came across this while testing Poloniex

As an eager tester, I cannot utilise `/internal/testing/exchange/SetupWs(tb testing.TB, e exchange.IBotExchange)` for a `useMultiConnectionManagement` connection following reasons:
- `b.Features.Subscriptions = subscription.List{}` removes subscriptions for the connection
- in `/websocket/manager.go:connect(ctx context.Context)` for `m.useMultiConnectionManagement` there is a requirement for subscriptions with the following
```go
if len(subs) == 0 {
	// If no subscriptions are generated, we skip the connection
	if m.verbose {
		log.Warnf(log.WebsocketMgr, "%s websocket: no subscriptions generated", m.exchangeName)
	}
	continue
}
```
- There is no way to override and enable `SubscriptionsNotRequired` in such a scenario so the connection is never made

It would be ideal to have a function within `/internal/testing/exchange` which sets up a connection for a `useMultiConnectionManagement` connection such as `gateio_websocket_request_test.go`'s `newExchangeWithWebsocket`

For now I will advise a code change within Poloniex to suit that PR to allow testing

PR #2156, opened 2026-01-27
exchanges: redesign websocket stream consumer

samuael:
This PR addresses the recurring `error: channel buffer is full: failed to relay` issue observed while handling websocket stream data.

The fix ensures that websocket messages are consistently consumed and properly relayed, preventing channel congestion and dropped messages.

- Added a listener to DataHandler in the SetupWs function of testexch to ensure incoming messages are actively consumed.
- Introduced a unit test to validate correct DataHandler behavior and guard against regressions.
- Fixed minor issues on `WebsocketRoutineManager`'s `websocketDataReceiver()` method incoming data handler.
- Made a minor improvement to the Send method of HataHandler to support the updated flow.

Fixes # (issue)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run
- [x] Test X

## Checklist


thrasher-:
@codex please review

exchange/stream/relay.go
```diff
@@ -34,10 +35,12 @@ func NewRelay(buffer uint) *Relay {
 // Send sends a message to the channel receiver
 // This is non-blocking and returns an error if the channel buffer is full
 func (r *Relay) Send(ctx context.Context, data any) error {
+	relayTimer := time.NewTimer(time.Millisecond * 10)
```
shazbert: Please revert this as is not an issue, this drops data by design. What it indicates is that there is a reader issue either not present or slow at processing.

exchange/stream/relay.go
```diff
@@ -34,10 +35,11 @@ func NewRelay(buffer uint) *Relay {
...
 	select {
 	case r.comm <- Payload{Ctx: common.FreezeContext(ctx), Data: data}:
 		return nil
-	default:
+	case <-relayTimer.C:
 		return fmt.Errorf("%w: failed to relay <%T>", errChannelBufferFull, data)
 	}
```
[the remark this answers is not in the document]
shazbert: Just a side note could have used `time.After`

internal/testing/exchange/exchange.go
```diff
@@ -206,6 +207,18 @@ func SetupWs(tb testing.TB, e exchange.IBotExchange) {
 	err = w.Connect(context.TODO())
 	require.NoError(tb, err, "Connect must not error")
 
+	w.Wg.Go(func() {
```
shazbert: An automatic reader routine might not be the best if we want to consume in tests. Maybe it should be a helper func that takes in a websocket instance? Erroring on a nil data is cool 👍

shazbert (changes requested):
Thanks Boss! Just some minor suggestions.

engine/websocketroutine_manager.go
```diff
@@ -171,12 +171,16 @@ func (m *WebsocketRoutineManager) websocketDataReceiver(ws *websocket.Manager) e
...
 				if payload.Data == nil {
 					log.Errorf(log.WebsocketMgr, "exchange %s nil data sent to websocket", ws.GetName())
+					continue
 				}
 				m.mu.RLock()
 				for x := range m.dataHandlers {
-					if err := m.dataHandlers[x](ws.GetName(), payload.Data); err != nil {
-						log.Errorln(log.WebsocketMgr, err)
-					}
+					m.wg.Go(func() {
```
gloriousCode: I'm not sure I agree with spawning a go routine for every single websocket message received for each data handler registered. The time it could take to handle the spawned go routine could be longer than it could take to handle the data. 

We would also need to ensure that data is processed in order

I understand that there can be annoyances for how this data is processed, but I would rather the effort be spent working out where in `websocketDataHandler` data is being processed slowly and resolve that first before doing something like this. 

So I think this change is out of scope of the actual problem you are encountering to do with testing in Poloniex

gloriousCode (changes requested):
I appreciate you looking into what's going on with the new relay. We definitely need to update tests to handle it.

gloriousCode:
Hi Sam, would something like this be an improvement for you? https://gist.github.com/gloriousCode/18394d5b5275862a53684bd413c9c364
This does the following
- Stops multi-connection websockets auto-subscribing causing datahandler relay errors
- Clears out all the other connections so you limit tests only for the connection you want to use

engine/websocketroutine_manager.go
samuael: This is a very good insight, sir.
I am thinking of creating the worker goroutines and let the function do the listening. Creating the listener functions(workers) based on the number of websocket connections the exchange creates.

**_Assuming that the stream data consuming should be faster or equal in speed to reading incoming the data from the websocket connection._**

samuael:
> Hi Sam, would something like this be an improvement for you? https://gist.github.com/gloriousCode/18394d5b5275862a53684bd413c9c364 This does the following
> 
>     * Stops multi-connection websockets auto-subscribing causing datahandler relay errors
> 
>     * Clears out all the other connections so you limit tests only for the connection you want to use

Thank you, sir. This is helpful.
However, for the long term, I believe, handling the issue from the DataHandler side should be the better.

exchange/stream/relay.go
samuael: Why did we not use the `ctx.Done()` to decide to drop the data?

internal/testing/exchange/exchange.go
samuael: I accept this but put the error check to inside the Send method. Since we are having issue of the stream not being available when we need to pass the data, we should put the `nil` guarg check here.

I used `NilGuard()` here for consistency.

samuael: If we were expecting an error (`Send(ctx context.Context, data any) error`) when sending payload to the `DataHandler`, we must be expecting this nil check error too.

samuael:
> Hi Sam, would something like this be an improvement for you? https://gist.github.com/gloriousCode/18394d5b5275862a53684bd413c9c364 This does the following
> 
>     * Stops multi-connection websockets auto-subscribing causing datahandler relay errors
> 
>     * Clears out all the other connections so you limit tests only for the connection you want to use

If we have to merge Poloniex first, I will follow your implementation.

engine/websocketroutine_manager.go
```diff
@@ -163,26 +163,29 @@ func (m *WebsocketRoutineManager) websocketDataReceiver(ws *websocket.Manager) e
...
-			case payload := <-ws.DataHandler.C:
-				if payload.Data == nil {
-					log.Errorf(log.WebsocketMgr, "exchange %s nil data sent to websocket", ws.GetName())
-				}
-				m.mu.RLock()
-				for x := range m.dataHandlers {
-					if err := m.dataHandlers[x](ws.GetName(), payload.Data); err != nil {
-						log.Errorln(log.WebsocketMgr, err)
-					}
+	for range ws.NumConnections() {
```
shazbert: Revert this please; Connections will be dynamic and dependent on pair count and subscription max count. So this would need to scale and tie into a websocket manager driven event. 

There is a possible performance driven solution where you load balance to your CPU threads but this just adds in more complexity without performance here. I would keep a single routine per exchange instance and let the go schedular do its job and consume one at a time in that routine. You can get a lot done just by doing that, the added overhead of routines will pressure the go schedular. 

The biggest boost you will get is upstream and optimising engine.

exchange/stream/relay.go
```diff
@@ -34,10 +34,13 @@ func NewRelay(buffer uint) *Relay {
 // Send sends a message to the channel receiver
 // This is non-blocking and returns an error if the channel buffer is full
 func (r *Relay) Send(ctx context.Context, data any) error {
+	if err := common.NilGuard(data); err != nil {
+		return err
+	}
```
shazbert: This can be removed, a client could use this package for anything they want including nils. This adds in latency and the objects might not be nillable anyway.

exchange/stream/relay.go
```diff
@@ -34,10 +34,13 @@ func NewRelay(buffer uint) *Relay {
...
 	select {
 	case r.comm <- Payload{Ctx: common.FreezeContext(ctx), Data: data}:
 		return nil
-	default:
+	case <-ctx.Done():
```
shazbert: Please revert: This will block the send and the reader operation. Reader operations must always be free for metrics aggregation which will track error counts.

internal/testing/exchange/exchange.go
```diff
@@ -206,9 +207,27 @@ func SetupWs(tb testing.TB, e exchange.IBotExchange) {
...
 	require.NoError(tb, err, "Connect must not error")
 
+	for range w.NumConnections() {
+		w.Wg.Add(1)
+		go streamDataConsumer(w)
+	}
 	setupWsOnce[e] = true
 }
 
+func streamDataConsumer(w *websocket.Manager) {
```
shazbert, line 236: This seems fine; will need to test it first.

shazbert (changes requested):
Had a quick check and its almost there. Got some things to consider first.

engine/websocketroutine_manager.go
samuael: According to the number of messages, including tickers that are coming to this channel, having only a single worker to consume the payloads is not efficient. I think, we should have a better way of consuming the payloads.


The workers can be for **all exchanges** or **n-workers per exchange**

exchange/stream/relay.go
samuael: I will replace it with `if data == nil {}` check. What I did was move the check from inside the consumer to the sender.

At least, it will help in checking the error before loading the buffer.

engine/websocketroutine_manager.go
```diff
@@ -38,6 +38,7 @@ func setupWebsocketRoutineManager(exchangeManager iExchangeManager, orderManager
 		orderManager:    orderManager,
 		syncer:          syncer,
 		currencyConfig:  cfg,
+		logSequence:     make(chan logCaller, 100),
```
shazbert: I’m struggling to see a clear efficiency or latency win here.

This appears to move blocking from the logging backend to a buffered channel, which under sustained load still blocks the websocket goroutines once the buffer fills. It also introduces additional allocations and a shutdown risk if handleLogs exits while producers are still sending.

The main culprit as well is not necessarily these updates but orderbook update frequency would could be in real time, this puts massive pressure on the system.

exchange/stream/relay.go
```diff
@@ -34,6 +34,9 @@ func NewRelay(buffer uint) *Relay {
 // Send sends a message to the channel receiver
 // This is non-blocking and returns an error if the channel buffer is full
 func (r *Relay) Send(ctx context.Context, data any) error {
+	if data == nil {
+		return common.ErrNilPointer
+	}
```
shazbert: This is a generic relay and as such this check is not required, this can be used to signal anything even a nil being pushed down as a signal.

shazbert (changes requested):
Just some things I noticed on recent changes. That need to be clarified.

exchange/stream/relay.go
samuael: I did this because we are not processing nil values on the handler side, and I didn’t want them to generate unnecessary traffic, so I added a nil check. If you believe a nil value should be treated as a signal, I’ve removed the check. However, I haven’t seen any place where nil values are actually processed, nor do I see how we would use nil as a signal when there are several clearer and more explicit ways to do so.

engine/websocketroutine_manager.go
samuael: What I was expecting was a request to add more consumers for the log call after you see the log processor. 

Anyways, I have removed this and replaced it with asynchronous call of summary processors of orderbook and tickers. Such as `printOrderSummary`, `PrintOrderbookSummary`, and `PrintTickerSummary` These calls do not expect a value return like err value and I believe calling them asynchronously will help execute the validations until the lock calls inside are released.

thrasher-:
Closing as the multiconn auto-subscribe flood this addressed was fixed by #2205 (issue #2151) and the unconditional drainer in `SetupWs` would race the many tests that read `DataHandler.C` directly on the shared exchange instance

PR #2157, opened 2026-01-28
engine/kucoin: change validation logic for credentials

shazbert:
# PR Description

With regards to Kucoin when futures was the only thing enabled it would fail when trying to verify/validate the credentials key and turn it off. 

engine:

* Add logic that sets spot first then a futures item then falls back to first available asset.
* **I stopped the setting of false to the credentials key in config as it was getting annoying** but if you want me to revert I can.

Kucoin: 

* Remove unused function and update the test to use the correct function

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -772,7 +772,7 @@ func (bot *Engine) LoadExchange(name string) error {
 		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
 		var preferredAsset asset.Item
 		switch {
-		case enabledAssets.Contains(asset.Spot): // prioritise spot due to wide usage across GCT
+		case enabledAssets.Contains(asset.Spot): // prioritise validating credentials with spot due to wide usage across GCT
 			preferredAsset = asset.Spot
 		default:
 			for _, a := range enabledAssets { // second priority to futures if spot isn't available


thrasher-:
@codex please review

engine/engine.go
```diff
@@ -768,26 +759,39 @@ func (bot *Engine) LoadExchange(name string) error {
...
+	if err := bot.ExchangeManager.Add(exch); err != nil {
 		return err
 	}
 
 	b := exch.GetBase()
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
-		err = exch.ValidateAPICredentials(context.TODO(), asset.Spot)
-		if err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", b.Name, err)
+		enabled := b.CurrencyPairs.GetAssetTypes(true)
```
gloriousCode: This shadows `var enabled = &gctrpc.GenericResponse{Status: "enabled"}` not huge, but my IDE text is green because of you. This should make you feel truly terrible

engine/engine.go
```diff
@@ -768,26 +759,39 @@ func (bot *Engine) LoadExchange(name string) error {
...
+					bias = a
+					break
+				}
+			}
+			if bias == 0 {
+				bias = enabled[0] // last resort pick first available
+			}
+		}
+
+		if err := exch.ValidateAPICredentials(context.TODO(), bias); err != nil {
```
gloriousCode: I don't understand why you aren't doing something like this
```go
ea := b.CurrencyPairs.GetAssetTypes(true)
if err := exch.ValidateAPICredentials(context.TODO(), ea[0]); err != nil { // nolint: enabled assets have already been cleared in the config package
```
The problem with the old way was that it chose spot, even if spot wasn't enabled. Now you choose something that is enabled. The only exchange that will potentially have issue is Kraken due to the dual API key issue - but that's not even really mitigated with your above code either

shazbert: Yeah my way was somewhat silly. I just didn't want to get back options asset randomly and then we don't have account support for it then it cooked itself.  We could range across and validate all assets as well and update the credentials manager later just to ensure that the asset key association is correct.

shazbert:
> tACK. ehh it's okay about how it's done really. I just didn't like how much space the code took up and that `enabled` was shadowed and the variable name `bias`

Yeah will see what other people say and change my bad names in one go

engine/engine.go
gloriousCode: Yeah that's all fair. I think I was feeling whingey before eating

engine/engine.go
```diff
@@ -768,26 +759,39 @@ func (bot *Engine) LoadExchange(name string) error {
...
 	}
 
 	b := exch.GetBase()
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
-		err = exch.ValidateAPICredentials(context.TODO(), asset.Spot)
-		if err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", b.Name, err)
+		enabled := b.CurrencyPairs.GetAssetTypes(true)
+
+		var bias asset.Item
```
thrasher-:
```suggestion
		var selectedAsset asset.Item
```

Or perhaps preferredAsset since we prioritise spot

thrasher- (changes requested):

engine/engine.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2157/commits/df446a6c34c8ef0ba94a4fc215cc6dd3b30feca3

I did preferred

engine/engine.go
```diff
@@ -767,26 +758,38 @@ func (bot *Engine) LoadExchange(name string) error {
...
 
 	b := exch.GetBase()
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
-		err = exch.ValidateAPICredentials(context.TODO(), asset.Spot)
-		if err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", b.Name, err)
+		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
+		var preferredAsset asset.Item
+		switch {
+		case enabledAssets.Contains(asset.Spot): // prioritise spot due to wide usage across GCT
```
gloriousCode:
```suggestion
		case enabledAssets.Contains(asset.Spot): // prioritise validating credentials with spot due to wide usage across GCT
```
I think that's the only critique I have here. It helps highlight the purpose of what's going on within this `if` statement a bit earlier.

gloriousCode (approved):

engine/engine.go
shazbert: Agreed merged

PR #2159, opened 2026-02-01
exchanges: Update `UpdateTickers` method so that available pairs are used instead of enabled pairs.

shazbert:
# PR Description

Externally run strategies will have enabled pairs set but dynamically changing pairs create potential holes when pricing balances when using get tickers functionality. 

* Load everything the function returns, provided its maintained in the available pairs list. 
* Removes GetTickers functionality for assets that does consecutive REST calls. This may break certain engine batching calls but this can be mapped out per asset functionality if required. 

note: It's imperative we remove wrapper impediments for REST calls e.g. if an asset is not enabled but supported. A client should be able to call these functions outside of enabled pairs/assets trading scope for hedging and pricing. This also includes returned websocket stream data. 

gloriousEdit: 
>  From my perspective wrapper/exchange layer's job is to translate exchange API responses into internal types. Policy decisions about which pairs to surface belong in the application layer (engine, gRPC handlers, strategies). Embedding an "enabled" filter inside a REST-response-processing loop conflates data acquisition with business logic

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2171, opened 2026-02-12
Currecy/Asset: Environment variable to Ignore enabled currency/asset check/error

gloriousCode:
# PR Description
PRs like #2159 with messages like
> note: It's imperative we remove wrapper impediments for REST calls e.g. if an asset is not enabled but supported. A client should be able to call these functions outside of enabled pairs/assets trading scope for hedging and pricing. This also includes returned websocket stream data.

Make me feel FEELINGS! I think it is totally the wrong direction to take. So rather than change every single wrapper function and every single websocket implementation, I thought this opt-in check would be a better way to address the issue.

Just add environment var `IGNORE_ENABLED_CHECK=true`. It allows processing of non-enabled currencies and assets through wrapper and websocket functions

Fixes # (issue)
- Shazbert

## How has this been tested
- TestIgnoreEnabledCheckBypassPaths

# Feature on:
Config:
image

```
go run . submitorder --exchange="gateio" --asset="margin" --pair="btc-usdt" --side="sell" --type="MARKET" --amount="1"
2026/02/13 09:29:36 rpc error: code = Unknown desc = GateIO unsuccessful HTTP status code: 403 raw response: {"message":"Request API key does not have spot write permission","label":"FORBIDDEN"}
, authenticated request failed
```

```
go run . getticker --exchange="gateio" --asset="margin" --pair="eth-usdt"
{
 "pair": {
  "delimiter": "-",
  "base": "eth",
  "quote": "usdt"
 },
 "last_updated": 1770935492,
 "last": 1928.25,
 "high": 2001.62,
 "low": 1896.71,
 "bid": 1928.25,
 "ask": 1928.26,
 "volume": 105785.5351
}%   
```

# Feature off:

```
go run . submitorder --exchange="gateio" --asset="margin" --pair="btc-usdt" --side="sell" --type="MARKET" --amount="1"
2026/02/13 09:41:23 rpc error: code = Unknown desc = margin asset type not enabled
```

```
go run . getticker --exchange="gateio" --asset="margin" --pair="eth-usdt"
2026/02/13 09:35:56 rpc error: code = Unknown desc = margin asset type not enabled
exit status 1
```

### DIFF
--- a/currency/manager.go
+++ b/currency/manager.go
@@ -31,6 +31,7 @@ var (
 	errPairConfigFormatNil = errors.New("pair config format is nil")
 )
 
+// ignoreEnabledCheckEnv setting this environment variable to true will ignore the enabled check for assets and pairs and not return errors
 const ignoreEnabledCheckEnv = "IGNORE_ENABLED_CHECK"
 
 var ignoreEnabledCurrencyAssetCheck bool
@@ -39,17 +40,13 @@ func init() {
 	ignoreEnabledCurrencyAssetCheck, _ = strconv.ParseBool(os.Getenv(ignoreEnabledCheckEnv))
 }
 
-func ignoreEnabledCheck() bool {
-	return ignoreEnabledCurrencyAssetCheck
-}
-
 // GetAssetTypes returns a list of stored asset types
 func (p *PairsManager) GetAssetTypes(enabled bool) asset.Items {
 	p.mutex.RLock()
 	defer p.mutex.RUnlock()
 	assetTypes := make(asset.Items, 0, len(p.Pairs))
 	for k, ps := range p.Pairs {
-		if enabled && !ps.AssetEnabled && !ignoreEnabledCheck() {
+		if enabled && !ps.AssetEnabled && !ignoreEnabledCurrencyAssetCheck {
 			continue
 		}
 		assetTypes = append(assetTypes, k)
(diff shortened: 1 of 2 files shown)


thrasher-:
@codex please review

shazbert:
So there is two competing philosophies here. 

PR #2159 (mine) — Remove the enabled-pairs filter from UpdateTickers wrapper functions entirely. Match against available pairs instead. The wrapper layer becomes agnostic — it fetches and stores everything the exchange supports, and filtering is left to higher layers (engine, strategy, CLI).

This — Keep the existing enabled-pairs checks in place but add an IGNORE_ENABLED_CHECK env var that globally bypasses them when set to true. The wrapper layer keeps its filter but callers can opt out.

From my perspective wrapper/exchange layer's job is to translate exchange API responses into internal types. Policy decisions about which pairs to surface belong in the application layer (engine, gRPC handlers, strategies). Embedding an "enabled" filter inside a REST-response-processing loop conflates data acquisition with business logic. #2159 pushes that boundary outward.

Main isssues with this PR:

* A package-level init() + global var checked inside hot paths (IsPairEnabled, GetAssetTypes, etc.) is a hidden side-channel. It makes functions behave differently depending on process-level state that isn't visible in their signatures or documentation.
* It circumvents IsPairEnabled to return true for pairs that are only available — this subtly violates the function's contract and could confuse callers who rely on IsPairEnabled meaning "the user explicitly enabled this pair".

@thrasher- @samuael

gloriousCode:
I agree this shouldn't be a PR. It was mostly to help you achieve your goals of not having filtering without requiring massive amounts of code changes and not needing to call "enable all pairs" every time a new pair is added

PR #2200, opened 2026-03-06
engine, backtester: propagate runtime lifecycle and harden graceful shutdown

thrasher-:
## Summary

This PR hardens runtime/shutdown lifecycle handling across engine + backtester, and finalises the noctx/lifecycle workstream.

### Key changes

- Propagates a runtime-scoped cancellable context through startup/runtime paths.
- Replaces noctx listener usage with context-aware listener setup.
- Improves gRPC shutdown behavior:
  - graceful stop with timeout fallback
  - suppresses expected server-stopped shutdown noise
  - avoids concurrent shutdown call hangs via non-blocking shutdown signaling
- Fixes websocket routine manager start/stop race conditions around cancel/shutdown channels.
- Prevents subsystem re-enable during active shutdown lifecycle.
- Fixes order-manager shutdown cancellation flow so `CancelOrdersOnShutdown` executes correctly in real stop paths.
- Improves data-history shutdown semantics so context cancellation is treated as shutdown/abort (not persisted as misleading failed job outcomes).
- Adds/updates targeted regression tests for the above paths.

## Why

This closes the noctx lifecycle gap and addresses shutdown race/edge-case behavior observed during review:
- startup/shutdown context consistency
- concurrent shutdown safety
- websocket manager state/race correctness
- cleaner shutdown behavior under cancellation

## Validation

Ran locally:

- `go test ./engine -count=1`
- `go test ./backtester/engine -count=1`
- `go test -race ./engine -run TestWebsocketRoutineManagerConcurrentStartStop -count=1`
- `go test -race ./engine -run TestShutdownConcurrentRequestsDoNotBlock -count=1`
- `golangci-lint run ./engine/... ./backtester/engine/...`
- `codespell` on touched files
- Simulated `.github/workflows/misc.yml` checks (pass)

## Notes

- Patch is intentionally squashed for review clarity.
- Includes both production fixes and regression coverage for race/shutdown behavior.

### DIFF
--- a/contrib/spellcheck/exclude_lines.txt
+++ b/contrib/spellcheck/exclude_lines.txt
@@ -52,3 +52,7 @@
 	pressXToJSON = `[0,"oc",[34930659963,null,1574955083558,"tETHUSD",1574955083558,1574955354487,0.201104,0.201104,"EXCHANGE LIMIT",null,null,null,0,"CANCELED",null,null,120,0,0,0,null,null,null,0,0,null,null,null,"BFX",null,null,null]]`
 	"Quotes":                                `{"arg": {"channel":"quotes"},"data":[{"validUntil":"1608997227854","uTime":"1608267227834","cTime":"1608267227834","legs":[{"px":"0.0023","sz":"25.0","instId":"BTC-USD-220114-25000-C","side":"sell","tgtCcy":""},{"px":"0.0045","sz":"25","instId":"BTC-USD-220114-35000-C","side":"buy","tgtCcy":""}],"quoteId":"25092","rfqId":"18753","traderCode":"SATS","quoteSide":"sell","state":"canceled","clQuoteId":""}]}`,
 	// unimplemented method is ever invoked, so we test this at initialization
+	assert.ErrorIs(t, ctx.Err(), context.Canceled, "runtime context should be cancelled")
+	require.ErrorIs(t, ctxAfterShutdown.Err(), context.Canceled, "shutdown-requested runtime context must stay cancelled")
+	require.ErrorIs(t, ctx.Err(), context.Canceled, "shutdown-requested runtime context must be cancelled")
+	assert.ErrorIs(t, ctx.Err(), context.Canceled, "runtime context should be cancelled on shutdown request")


main.go
```diff
@@ -143,21 +144,37 @@ func main() {
 
 	engine.Bot.Settings.PrintLoadedSettings()
 
+	engine.Bot.EnsureRuntimeContext()
+	var shutdownRequested atomic.Bool
+	go waitForInterrupt(settings.Shutdown, engine.Bot, &shutdownRequested)
+
 	if err = engine.Bot.Start(); err != nil {
+		if shutdownRequested.Load() {
+			engine.Bot.Stop()
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Great catch — addressed in commit https://github.com/thrasher-/gocryptotrader/commit/3e8a0c3bf23b916daada0dcecda074bfad05d33e. The startup-abort path in `main.go` now returns early when shutdown was requested during `Start()`, and no longer calls `Stop()` on a potentially partially initialised engine.

backtester/main.go
```diff
@@ -218,19 +219,27 @@ func main() {
...
-			fmt.Printf("Could not start RPC server. Error: %v\n", err)
+		s, setupErr := backtest.SetupRPCServer(c, runManager)
+		if setupErr != nil {
+			fmt.Printf("Could not setup RPC server. Error: %v\n", setupErr)
+			os.Exit(1)
+		}
+		startErr := backtest.StartRPCServer(ctx, s)
+		if startErr != nil && ctx.Err() == nil {
+			fmt.Printf("Could not start RPC server. Error: %v\n", startErr)
 			os.Exit(1)
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Agreed, and updated in commit https://github.com/thrasher-/gocryptotrader/commit/3e8a0c3bf23b916daada0dcecda074bfad05d33e. `backtester/main.go` no longer calls `os.Exit(1)` from inside the RPC goroutine; RPC setup/start errors are sent over an error channel and handled from the main goroutine so cancellation/cleanup occurs before process exit.

engine/datahistory_manager.go
```diff
@@ -716,16 +722,19 @@ func (m *DataHistoryManager) processCandleData(job *DataHistoryJob, exch exchang
...
 		job.Asset,
 		job.Interval,
 		startRange,
 		endRange)
 	if err != nil {
+		if ctxErr := ctx.Err(); ctxErr != nil && errors.Is(err, ctxErr) {
+			return nil, err
+		}
 		r.Result += "could not get candles: " + err.Error() + ". "
 		r.Status = dataHistoryStatusFailed
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Addressed in commit https://github.com/thrasher-/gocryptotrader/commit/3e8a0c3bf23b916daada0dcecda074bfad05d33e. For context-cancellation cases, the path now returns the cancellation error (`return nil, err`) so shutdown/abort is explicit. For non-cancellation API errors, existing data-history behavior is preserved by recording failure details in the job result and returning `r, nil` intentionally.

engine/datahistory_manager.go
```diff
@@ -769,15 +778,18 @@ func (m *DataHistoryManager) processTradeData(job *DataHistoryJob, exch exchange
...
 		startRange,
 		endRange)
 	if err != nil {
+		if ctxErr := ctx.Err(); ctxErr != nil && errors.Is(err, ctxErr) {
+			return nil, err
+		}
 		r.Result += "could not get trades: " + err.Error() + ". "
 		r.Status = dataHistoryStatusFailed
-		return r, nil //nolint:nilerr // error is returned in the job result
+		return r, nil
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Same treatment here in commit https://github.com/thrasher-/gocryptotrader/commit/3e8a0c3bf23b916daada0dcecda074bfad05d33e. Cancellation now returns a real context error; non-cancellation fetch errors continue to be encoded in the job result (`r`) by design to preserve existing data-history semantics.

backtester/engine/grpcserver.go
```diff
@@ -86,22 +90,40 @@ func StartRPCServer(server *GRPCServer) error {
...
+			log.Errorln(log.GRPCSys, serveErr)
 			return
 		}
 	}()
 
 	log.Debugln(log.GRPCSys, "GRPC server started!")
 
 	if server.config.GRPC.GRPCProxyEnabled {
-		return server.StartRPCRESTProxy()
+		return server.startRPCRESTProxy(ctx)
```
shazbert: If the proxy server fails for any reason, either initialise the proxy before starting Serve, or stop and close the gRPC server before returning the proxy error.

backtester/engine/grpcserver.go
```diff
@@ -123,15 +145,25 @@ func (s *GRPCServer) StartRPCRESTProxy() error {
...
+		if shutdownErr := server.Shutdown(shutdownCtx); shutdownErr != nil && !errors.Is(shutdownErr, http.ErrServerClosed) {
+			log.Errorf(log.GRPCSys, "GRPC proxy server shutdown failed: %s\n", shutdownErr)
 		}
+	}()
 
-		if err = server.ListenAndServe(); err != nil {
-			log.Errorf(log.GRPCSys, "GRPC proxy failed to server: %s\n", err)
+	go func() {
+		if serveErr := server.ListenAndServe(); serveErr != nil && !errors.Is(serveErr, http.ErrServerClosed) {
+			log.Errorf(log.GRPCSys, "GRPC proxy failed to server: %s\n", serveErr)
```
shazbert: Across these logs, the `\n` can be removed as it gets screened in the log package.

backtester/main.go
```diff
@@ -218,51 +219,71 @@ func main() {
...
-	interrupt := <-signaler.WaitForInterrupt()
-	log.Infof(log.Global, "Captured %v, shutdown requested\n", interrupt)
-	if btCfg.StopAllTasksOnClose {
-		log.Infoln(log.Global, "Stopping all running tasks on close")
-		var stopped []*backtest.TaskSummary
-		stopped, err = runManager.StopAllTasks()
-		if err != nil {
-			log.Errorln(common.Backtester, err)
+		startErr := backtest.StartRPCServer(ctx, s)
+		if startErr != nil && ctx.Err() == nil {
```
shazbert, line 233: maybe if the context is erroring we should send that through the channel? as an else if case?

engine/engine.go
```diff
@@ -32,26 +32,30 @@ import (
...
+	gctScriptManager         *gctscript.GctScriptManager
+	WebsocketRoutineManager  *WebsocketRoutineManager
+	WithdrawManager          *WithdrawManager
+	dataHistoryManager       *DataHistoryManager
+	currencyStateManager     *CurrencyStateManager
+	Settings                 Settings
+	uptime                   time.Time
+	GRPCShutdownSignal       chan struct{}
+	runtimeCtx               context.Context //nolint:containedctx // runtime-scoped cancellation context for startup + subsystem propagation
+	runtimeCancel            context.CancelFunc
```
shazbert, line 55: suggestion: Any chance we can use Start with updated parameter with context and just use that without all the runtime ctx methods? Then if we need to stop it the caller just calls cancel and then the Stop method?

engine/exchange_manager.go
```diff
@@ -146,9 +147,16 @@ func (m *ExchangeManager) Shutdown(shutdownTimeout time.Duration) error {
...
-		name string
-		key  string
-		err  error
+		name     string
+		key      string
+		err      error
+		duration time.Duration
+	}
+
+	type pendingShutdown struct {
```
shazbert, line 156: Potentially, this type can be merged to the above type. And the loops merged into one having both startedAt and duration times fields, so that duration.

engine/helpers.go
```diff
@@ -125,6 +126,15 @@ func (bot *Engine) SetSubsystem(subSystemName string, enable bool) error {
 		return errNilConfig
 	}
 
+	if enable && bot.isRuntimeShutdownRequested() {
```
shazbert, line 129: I think its better to have a ctx param on these helper funcs.

engine/helpers.go
```diff
@@ -622,6 +632,7 @@ func (bot *Engine) GetExchangeCryptocurrencyDepositAddress(ctx context.Context,
 
 // GetAllExchangeCryptocurrencyDepositAddresses obtains an exchanges deposit cryptocurrency list
 func (bot *Engine) GetAllExchangeCryptocurrencyDepositAddresses() map[string]ExchangeDepositAddresses {
+	runtimeCtx := bot.getRuntimeContext()
```
shazbert, line 635: Like above the ctx should be populated through the paramater

exchanges/bitfinex/bitfinex_websocket.go
```diff
@@ -143,20 +143,20 @@ func (e *Exchange) WsConnect() error {
...
-			e.Websocket.SetCanUseAuthenticatedEndpoints(false)
+		} else {
+			e.Websocket.Wg.Add(1)
+			go e.wsReadData(ctx, e.Websocket.AuthConn)
+			err = e.WsSendAuth(ctx)
+			if err != nil {
+				log.Errorf(log.ExchangeSys,
+					"%v - authentication failed: %v\n",
+					e.Name,
+					err)
```
shazbert: WE can one line these logs.

main.go
```diff
@@ -143,21 +144,38 @@ func main() {
...
 	engine.Bot.Settings.PrintLoadedSettings()
 
+	engine.Bot.EnsureRuntimeContext()
+	var shutdownRequested atomic.Bool
+	go waitForInterrupt(settings.Shutdown, engine.Bot, &shutdownRequested)
+
 	if err = engine.Bot.Start(); err != nil {
+		if shutdownRequested.Load() {
+			// Startup may have been interrupted before full subsystem initialisation.
+			// Avoid calling Stop() on a partially-initialized engine.
```
shazbert:
```suggestion
			// Avoid calling Stop() on a partially-initialised engine.
```

shazbert (changes requested):
thanks!

backtester/engine/grpcserver.go
thrasher-: Jarvis: Addressed. Startup now initialises proxy setup before launching gRPC `Serve`, startup returns immediately on proxy bind/setup failures, and failure cleanup closes listeners so they are not left open. Added regression coverage in `TestStartRPCServerProxyFailureStopsListener`. Commit: https://github.com/thrasher-/gocryptotrader/commit/bafeb07b9c14def269c0751b44d1fb641e50b5a5

backtester/main.go
thrasher-: Jarvis: Good call-out. I am intentionally leaving this unchanged in this patch: pushing raw context cancellation through the startup error channel can turn normal shutdown into a startup failure signal. Current flow handles shutdown via context path; I suggest addressing this in a focused lifecycle follow-up so cancellation semantics stay explicit.

engine/engine.go
thrasher-: Jarvis: Agreed this is a reasonable direction. I am deferring this in the current patch because it broadens lifecycle API surface (`Start(ctx, ...)` + cancel/Stop flow) beyond the scoped startup-safety fix here. Happy to do it as a dedicated follow-up refactor.

engine/exchange_manager.go
thrasher-: Jarvis: This is a good cleanup idea, but I am deferring it for now to keep this patch tightly scoped to correctness/stability changes. I can include type/loop consolidation in the same follow-up refactor set as the runtime-context API simplification.

engine/helpers.go
thrasher-: Jarvis: Agreed on explicit `ctx` parameter threading as a clearer long-term pattern. I have deferred this in this patch to avoid widening scope, and will fold it into the follow-up lifecycle/context refactor.

engine/helpers.go
thrasher-: Jarvis: Same as above — explicit `ctx` threading is cleaner, but I have deferred this in the current scoped fix set and will include it in the follow-up lifecycle/context refactor.

backtester/engine/grpcserver.go
thrasher-: Jarvis: Addressed. Removed trailing `\n` usage in the touched gRPC proxy log paths to keep logger output formatting consistent. Commit: https://github.com/thrasher-/gocryptotrader/commit/bafeb07b9c14def269c0751b44d1fb641e50b5a5

exchanges/bitfinex/bitfinex_websocket.go
thrasher-: Jarvis: Addressed. Simplified this logging call to a single-line `log.Errorf(...)` form. Commit: https://github.com/thrasher-/gocryptotrader/commit/bafeb07b9c14def269c0751b44d1fb641e50b5a5

main.go
thrasher-: Jarvis: Addressed. Applied the inline wording simplification for that comment path. Commit: https://github.com/thrasher-/gocryptotrader/commit/bafeb07b9c14def269c0751b44d1fb641e50b5a5

engine/engine.go
```diff
@@ -554,17 +573,118 @@ func (bot *Engine) Start() error {
 		}
 	}
 
+	startSuccessful = true
 	return nil
 }
 
+func (bot *Engine) getRuntimeContext() context.Context {
```
gloriousCode, line 580: While none of the below functions are very complex. I don't see why you cannot add dedicated unit tests for each of them. Direct tests are important (before you say there is indirect coverage)

thrasher-: jarvis: Addressed in 8ea33da9. Added dedicated unit tests for the runtime helper block in `engine/engine_test.go`, covering `getRuntimeContext`, `getRuntimeCancel`, `cancelRuntimeContext`, `clearRuntimeContext`, `setRuntimeShutdownRequested`, `isRuntimeShutdownRequested`, `EnsureRuntimeContext`, and `RequestShutdown`, including the shutdown reuse and edge-case branches in `EnsureRuntimeContext()`. Focused tests, focused `-race`, and `golangci-lint run ./engine/...` pass. Full `go test ./engine -count=1` is still blocked by the unrelated Bitfinex ticker failure already present on `master`.

gloriousCode (approved):
tACK! Thanks Dirvus

PR #2205, opened 2026-03-10
websocket, internal/testing: support multiconn WS helpers

### LINKED ISSUE #2151: websocket/manager: internal/testing/exchange support for multi-connection

thrasher-:
## PR Description
Fix internal testing websocket helpers so they work with multi-connection managers.

This addresses issue #2151 by updating the shared websocket test helpers instead of relying on per-exchange workarounds.

### What changed
- added `SetSubscriptionsNotRequired()` support for multi-connection managers
- added `SetAllConnectionURLs()` so test helpers can redirect all managed websocket connections to a single mock server
- updated `internal/testing/exchange.SetupWs` to use the multi-connection-safe helper path
- updated `internal/testing/exchange.MockWsInstance` to use the multi-connection-safe helper path
- added direct unit coverage for the new manager helper methods
- added helper-level regression coverage for:
  - `MockWsInstance` with a real multi-connection exchange
  - `SetupWs` with a synthetic multi-connection setup
- added a focused Poloniex proof-of-concept test showing the shared helper works on the exchange that originally triggered the issue

### Why
`SetupWs` and `MockWsInstance` previously assumed single-connection websocket behaviour. That broke down once exchanges used `UseMultiConnectionManagement`, because empty generated subscriptions prevented connections from being established and test helpers had no clean way to mark subscriptions as not required.

This change keeps the fix in shared test infrastructure so exchanges do not need ad-hoc helper logic.

## Type of change
- [x] Bug fix
- [x] Tests

## How has this been tested
- `go test ./exchange/websocket -count=1`
- `go test ./internal/testing/exchange -count=1 -v`
- `go test ./exchanges/bybit -count=1`
- `go test ./exchanges/poloniex -count=1`
- `golangci-lint run ./exchange/websocket ./internal/testing/exchange ./exchanges/poloniex`
- `codespell`
- `gofumpt`

Closes #2151


exchange/websocket/manager.go
```diff
@@ -772,6 +772,67 @@ func (m *Manager) IsEnabled() bool {
...
+		m.GenerateSubs = func() (subscription.List, error) { return subscription.List{}, nil }
+		return
+	}
+
+	for _, ws := range m.connectionManager {
+		if ws.setup == nil {
+			ws.setup = &ConnectionSetup{}
+		}
+		ws.setup.SubscriptionsNotRequired = true
+	}
```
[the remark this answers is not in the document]
thrasher-: Jarvis: I’m keeping this guard as-is. In normal flow multi-connection entries are created via SetupNewConnection with a non-nil setup, so this is a cheap defensive safeguard rather than a primary code path, and I’d rather keep the helper resilient here than widen it into a broader error-return change.

exchange/websocket/manager.go
```diff
@@ -772,6 +772,67 @@ func (m *Manager) IsEnabled() bool {
...
+
+	for _, ws := range m.connectionManager {
+		if ws.setup == nil {
+			ws.setup = &ConnectionSetup{}
+		}
+		ws.setup.URL = u
+		for _, conn := range ws.connections {
+			conn.SetURL(u)
+		}
+	}
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Same reasoning here — I’m keeping the guard. The nil setup case should not occur in normal multi-connection setup flow, and the defensive check keeps the two helper methods consistent without changing behaviour for valid managers.

internal/testing/exchange/exchange.go
```diff
@@ -115,15 +115,16 @@ func MockWsInstance[T any, PT interface {
...
-	}
+
+	wsURL := "ws" + strings.TrimPrefix(s.URL, "http")
+	err = b.Websocket.SetAllConnectionURLs(wsURL)
+	require.NoError(tb, err, "SetAllConnectionURLs must not error")
 
 	// For testing we never want to use the default subscriptions; Tests of GenerateSubscriptions should be exercising it directly
 	b.Features.Subscriptions = subscription.List{}
 	// Exchanges which don't support subscription conf; Can be removed when all exchanges support sub conf
 	b.Websocket.GenerateSubs = func() (subscription.List, error) { return subscription.List{}, nil }
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/669594f9fb4f7b219da84136895bbeb155ca33ae — removed the redundant GenerateSubs override so MockWsInstance now relies on SetSubscriptionsNotRequired() as the single source of that behaviour.

internal/testing/exchange/exchange.go
```diff
@@ -115,15 +115,16 @@ func MockWsInstance[T any, PT interface {
...
+	b.Websocket.SetSubscriptionsNotRequired()
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/669594f9fb4f7b219da84136895bbeb155ca33ae — removed the redundant GenerateSubs assignment and the now-obsolete extra comment so this helper only uses SetSubscriptionsNotRequired() for that behaviour.

exchange/websocket/manager.go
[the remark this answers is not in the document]
shazbert: I think its better to at least log a warning as this should be set up correctly

exchange/websocket/manager.go
```diff
@@ -772,6 +772,67 @@ func (m *Manager) IsEnabled() bool {
...
+
+	m.m.Lock()
+	defer m.m.Unlock()
+
+	if !m.useMultiConnectionManagement {
+		m.runningURL = u
+		m.runningURLAuth = u
+		if m.Conn != nil {
+			m.Conn.SetURL(u)
```
shazbert: This looks like an error case (and below) as we have already potentially established a connection. We should ensure this is called prior to a connect call.

exchange/websocket/manager.go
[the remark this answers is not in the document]
shazbert: I think logging is fine here.

exchange/websocket/manager_test.go
```diff
@@ -447,6 +447,66 @@ func TestCreateConnectAndSubscribe(t *testing.T) {
...
+
+	subs, err := singleConn.GenerateSubs()
+	require.NoError(t, err, "GenerateSubs must not error after subscriptions are disabled")
+	assert.Empty(t, subs, "GenerateSubs should return no subscriptions after subscriptions are disabled")
+
+	multiConn := NewManager()
+	multiConn.useMultiConnectionManagement = true
+	multiConn.connectionManager = []*websocket{
+		{setup: &ConnectionSetup{}},
+		{setup: &ConnectionSetup{SubscriptionsNotRequired: true}},
```
shazbert, line 469: should add in a nil setup case

exchange/websocket/manager_test.go
```diff
@@ -447,6 +447,66 @@ func TestCreateConnectAndSubscribe(t *testing.T) {
...
+	assert.Equal(t, "ws://mock.example.com/ws", singleConn.runningURL, "runningURL should be updated for single-connection managers")
+	assert.Equal(t, "ws://mock.example.com/ws", singleConn.runningURLAuth, "runningURLAuth should be updated for single-connection managers")
+	assert.Equal(t, "ws://mock.example.com/ws", singleConn.Conn.GetURL(), "Conn URL should be updated for single-connection managers")
+	assert.Equal(t, "ws://mock.example.com/ws", singleConn.AuthConn.GetURL(), "AuthConn URL should be updated for single-connection managers")
+
+	multiConn := NewManager()
+	multiConn.useMultiConnectionManagement = true
+	multiConn.connectionManager = []*websocket{
+		{setup: &ConnectionSetup{URL: "ws://first.example.com"}, connections: []Connection{&connection{URL: "ws://first.example.com"}}},
+		{setup: &ConnectionSetup{URL: "ws://second.example.com"}, connections: []Connection{&connection{URL: "ws://second.example.com"}}},
```
shazbert: Should add in a nil setup case

exchange/websocket/manager.go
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/ba823c24164bc82317e94eee52e1c90a8110814d — when this helper encounters a nil connection setup it now logs a warning before creating an empty setup, so the defensive guard stays visible instead of silent.

exchange/websocket/manager.go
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/ba823c24164bc82317e94eee52e1c90a8110814d — same treatment here: a nil setup now logs a warning before the empty setup is created.

exchange/websocket/manager.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/ba823c24164bc82317e94eee52e1c90a8110814d — I tightened SetAllConnectionURLs so it now errors once Connect has started, and I removed the multi-connection live-connection URL mutation so this stays a pre-connect helper.

exchange/websocket/manager_test.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/ba823c24164bc82317e94eee52e1c90a8110814d — added a nil setup case to TestSetSubscriptionsNotRequired and assert it initialises the missing setup before flipping SubscriptionsNotRequired.

exchange/websocket/manager_test.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/ba823c24164bc82317e94eee52e1c90a8110814d — added a nil setup case here as well, and the test now focuses on pre-connect setup URL updates rather than mutating an existing live connection.

exchange/websocket/manager.go
```diff
@@ -772,6 +772,74 @@ func (m *Manager) IsEnabled() bool {
 	return m.enabled.Load()
 }
 
+// SetSubscriptionsNotRequired configures the manager to connect without
+// generating default subscriptions. This is primarily used by test helpers
+// that need request-response websocket connectivity without stream bootstraps.
+// It is intended as a pre-connect test-mode mutation; tests that later need
+// the manager's normal default-subscription bootstrap should use a fresh
+// exchange/manager instance.
+func (m *Manager) SetSubscriptionsNotRequired() {
```
gloriousCode: I think these testonly functions should be part of their own file. Its wasting space in an already convoluted package/file. 
While splitting it can steer it to begin even more convoluted, at least by virtue of the filename users will know not to call them under normal operations. The comment is weak with "This is primarily used by test helpers", it is _only_ for testing purposes that these exist, so I think you should make that comment stronger

I can still be persuaded to not split it out into its own file, but please at least improve the comments

thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/9332623e3c082782363803403b74e46d6f40c8ff — I moved the test-only hooks out of `manager.go` into `exchange/websocket/manager_testhelpers.go` and tightened the comments to make it explicit they’re only for cross-package test harnesses / pre-connect test setup.

gloriousCode: You're alright Jarvis, no need to unplug you yet

thrasher-:
@codex please review

shazbert (approved):
ch ACK!

PR #2208, opened 2026-03-11
Deribit: add block trade and block RFQ APIs

### LINKED ISSUE #2081: Deribit: Add new block trades and block RFQ API endpoints
Block trade API endpoints: https://docs.deribit.com/#block-trade
Block RFQ endpoints: https://docs.deribit.com/#block-rfq

thrasher-:
Fixes #2081.

Adds the Deribit block trade and block RFQ API coverage requested in #2081, with REST + websocket parity, docs-aligned request/response types, and direct test coverage for the new validation paths.

Highlights
- add block trade endpoints for verify/execute/invalidate, pending request review, trade history, broker trade requests/history, and simulation
- add block RFQ endpoints for create/add/edit/cancel/accept flows plus RFQ/quote/maker/user/trade retrieval
- align request payloads with the current Deribit OpenAPI, including:
  - dedicated create_block_rfq leg request shape
  - accept_block_rfq parameter shape
  - add/edit_block_rfq_quote parameter shape
  - broker/client trade allocation validation
- tighten response typing for broker trade and block RFQ payloads where the docs/examples require different shapes from request models
- add direct validator coverage, offline parameter-validation tests, and focused unmarshal coverage for the new response types
- add rate-limit entries for the new Deribit endpoints

Validation
- `go test ./exchanges/deribit -count=1`
- `golangci-lint run ./exchanges/deribit`
- `codespell --dictionary=-,./contrib/spellcheck/codespell_custom_dictionary.txt exchanges/deribit/deribit.go exchanges/deribit/deribit_ws_endpoints.go exchanges/deribit/deribit_types.go exchanges/deribit/deribit_test.go exchanges/deribit/ratelimit.go`
- `git diff --check`

### DIFF
--- a/exchanges/deribit/deribit.go
+++ b/exchanges/deribit/deribit.go
@@ -2588,7 +2588,7 @@ func validateBlockRFQTradeAllocations(tradeAllocations []BlockRFQTradeAllocation
 
 func validateBlockRFQHedge(hedge *BlockRFQHedgeLeg) (*BlockRFQHedgeLeg, error) {
 	if hedge == nil {
-		return nil, nil
+		return nil, common.ErrNilPointer
 	}
 	checkedHedge := *hedge
 	if checkedHedge.InstrumentName == "" {
@@ -2630,9 +2630,12 @@ func (e *Exchange) CreateBlockRFQ(ctx context.Context, req *CreateBlockRFQReques
 	if err != nil {
 		return nil, err
 	}
-	checkedHedge, err := validateBlockRFQHedge(req.Hedge)
-	if err != nil {
-		return nil, err
+	var checkedHedge *BlockRFQHedgeLeg
+	if req.Hedge != nil {
+		checkedHedge, err = validateBlockRFQHedge(req.Hedge)
+		if err != nil {
+			return nil, err
+		}
 	}
 	params := url.Values{}
 	err = setJSONParam(params, "legs", checkedLegs)
@@ -2683,9 +2686,12 @@ func (e *Exchange) AddBlockRFQQuote(ctx context.Context, req *AddBlockRFQQuoteRe
 	if err != nil {
 		return nil, err
 	}
-	checkedHedge, err := validateBlockRFQHedge(req.Hedge)
-	if err != nil {
-		return nil, err
+	var checkedHedge *BlockRFQHedgeLeg
+	if req.Hedge != nil {
+		checkedHedge, err = validateBlockRFQHedge(req.Hedge)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if req.Amount < 0 {
 		return nil, errInvalidAmount
@@ -2741,9 +2747,12 @@ func (e *Exchange) EditBlockRFQQuote(ctx context.Context, req *EditBlockRFQQuote
 	if err != nil {
 		return nil, err
 	}
-	checkedHedge, err := validateBlockRFQHedge(req.Hedge)
-	if err != nil {
-		return nil, err
+	var checkedHedge *BlockRFQHedgeLeg
+	if req.Hedge != nil {
+		checkedHedge, err = validateBlockRFQHedge(req.Hedge)
+		if err != nil {
+			return nil, err
+		}
 	}
 	if req.Amount < 0 {
 		return nil, errInvalidAmount
@@ -2859,9 +2868,12 @@ func (e *Exchange) AcceptBlockRFQ(ctx context.Context, req *AcceptBlockRFQReques
 	if err != nil {
 		return nil, err
 	}
-	checkedHedge, err := validateBlockRFQHedge(req.Hedge)
-	if err != nil {
-		return nil, err
+	var checkedHedge *BlockRFQHedgeLeg
+	if req.Hedge != nil {
+		checkedHedge, err = validateBlockRFQHedge(req.Hedge)
+		if err != nil {
+			return nil, err
+		}
 	}
 	params := url.Values{}
 	params.Set("block_rfq_id", strconv.FormatUint(req.BlockRFQID, 10))
--- a/exchanges/deribit/deribit.go
+++ b/exchanges/deribit/deribit.go
@@ -159,15 +159,6 @@ const (
 	getCombos       = "public/get_combos"
 	createCombos    = "private/create_combo"
 
-	// Block Trades Endpoints
-	executeBlockTrades             = "private/execute_block_trade"
-	getBlockTrades                 = "private/get_block_trade"
-	getLastBlockTradesByCurrency   = "private/get_last_block_trades_by_currency"
-	invalidateBlockTradesSignature = "private/invalidate_block_trade_signature"
-	movePositions                  = "private/move_positions"
-	simulateBlockPosition          = "private/simulate_block_trade"
-	verifyBlockTrades              = "private/verify_block_trade"
-
 	// roles
 
 	roleMaker = "maker"
@@ -2319,7 +2310,7 @@ func (e *Exchange) ExecuteBlockTrade(ctx context.Context, timestampMS time.Time,
 	params.Set("counterparty_signature", signature)
 	params.Set("timestamp", strconv.FormatInt(timestampMS.UnixMilli(), 10))
 	var resp []BlockTradeResponse
-	return resp, e.SendHTTPAuthRequest(ctx, exchange.RestFutures, matchingEPL, http.MethodGet, executeBlockTrades, params, &resp)
+	return resp, e.SendHTTPAuthRequest(ctx, exchange.RestFutures, matchingEPL, http.MethodGet, "private/execute_block_trade", params, &resp)
 }
 
 // VerifyBlockTrade verifies and creates block trade signature
@@ -2366,7 +2357,7 @@ func (e *Exchange) VerifyBlockTrade(ctx context.Context, timestampMS time.Time, 
 	resp := &struct {
 		Signature string `json:"signature"`
 	}{}
-	return resp.Signature, e.SendHTTPAuthRequest(ctx, exchange.RestFutures, matchingEPL, http.MethodGet, verifyBlockTrades, params, resp)
+	return resp.Signature, e.SendHTTPAuthRequest(ctx, exchange.RestFutures, matchingEPL, http.MethodGet, "private/verify_block_trade", params, resp)
 }
 
 // InvalidateBlockTradeSignature user at any time (before the private/execute_block_trade is called) can invalidate its own signature effectively cancelling block trade
@@ -2377,7 +2368,7 @@ func (e *Exchange) InvalidateBlockTradeSignature(ctx context.Context, signature 
 	params := url.Values{}
 	params.Set("signature", signature)
 	var resp string
-	err := e.SendHTTPAuthRequest(ctx, exchange.RestSpot, nonMatchingEPL, http.MethodGet, invalidateBlockTradesSignature, params, &resp)
+	err := e.SendHTTPAuthRequest(ctx, exchange.RestSpot, blockTradeWriteEPL, http.MethodGet, "private/invalidate_block_trade_signature", params, &resp)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


thrasher-:
@codex please review

exchanges/deribit/deribit_types.go
```diff
@@ -952,6 +957,370 @@ type BlockTradeData struct {
...
+}
+
+// BlockRFQData represents one Block RFQ entry.
+type BlockRFQData struct {
+	CreationTimestamp     types.Time           `json:"creation_timestamp"`
+	ExpirationTimestamp   types.Time           `json:"expiration_timestamp"`
+	BlockRFQID            uint64               `json:"block_rfq_id"`
+	Role                  string               `json:"role"`
+	State                 string               `json:"state"`
+	TakerRating           string               `json:"taker_rating"`
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/77e6b3ae1ae1dc003ee89c495cff91439dba3c30 — I switched the RFQ taker-rating fields to types.Number so they tolerate both numeric and string payloads instead of forcing a single representation.

exchanges/deribit/deribit.go
```diff
@@ -2395,7 +2386,580 @@ func (e *Exchange) GetUserBlockTrade(ctx context.Context, id string) ([]BlockTra
...
+	copy(checkedLegs, legs)
+	for i := range checkedLegs {
+		if checkedLegs[i].InstrumentName == "" {
+			return nil, fmt.Errorf("%w, empty string", errInvalidInstrumentName)
+		}
+		if checkedLegs[i].Amount <= 0 {
+			return nil, errInvalidAmount
+		}
+		normalisedDirection, err := validateBlockRFQDirection(checkedLegs[i].Direction)
+		if err != nil {
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/77e6b3ae1ae1dc003ee89c495cff91439dba3c30 — create RFQ leg validation now rejects empty directions explicitly, and I added direct coverage for the empty-direction case.

exchanges/deribit/deribit.go
```diff
@@ -2319,7 +2310,7 @@ func (e *Exchange) ExecuteBlockTrade(ctx context.Context, timestampMS time.Time,
 	params.Set("counterparty_signature", signature)
 	params.Set("timestamp", strconv.FormatInt(timestampMS.UnixMilli(), 10))
 	var resp []BlockTradeResponse
-	return resp, e.SendHTTPAuthRequest(ctx, exchange.RestFutures, matchingEPL, http.MethodGet, executeBlockTrades, params, &resp)
+	return resp, e.SendHTTPAuthRequest(ctx, exchange.RestFutures, matchingEPL, http.MethodGet, "private/execute_block_trade", params, &resp)
```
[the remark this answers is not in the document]
thrasher-: Jarvis: I left this as-is. This file already follows the existing pattern of keeping endpoint paths inline, so I avoided style-only churn here.

exchanges/deribit/deribit.go
```diff
@@ -2395,7 +2386,580 @@ func (e *Exchange) GetUserBlockTrade(ctx context.Context, id string) ([]BlockTra
...
+func (e *Exchange) ApproveBlockTrade(ctx context.Context, timestampMS time.Time, tradeNonce, role string) error {
+	if tradeNonce == "" {
+		return errMissingNonce
+	}
+	if timestampMS.IsZero() {
+		return errZeroTimestamp
+	}
+	if role != roleMaker && role != roleTaker {
+		return errInvalidTradeRole
+	}
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/77e6b3ae1ae1dc003ee89c495cff91439dba3c30 — I extracted the shared approve/reject validation into validatePendingBlockTradeAction and reused it across the REST and websocket approve/reject paths.

exchanges/deribit/deribit.go
```diff
@@ -2395,7 +2386,580 @@ func (e *Exchange) GetUserBlockTrade(ctx context.Context, id string) ([]BlockTra
...
+		if err != nil {
+			return nil, err
+		}
+	}
+	if checkedHedge != nil {
+		err = setJSONParam(params, "hedge", checkedHedge)
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
thrasher-: Jarvis: I left this as-is. The current error handling is correct, and I avoided a style-only reshuffle in this PR.

exchanges/deribit/deribit_test.go
```diff
@@ -3033,6 +3033,670 @@ func TestWSRetrieveUserBlockTrade(t *testing.T) {
...
+}
+
+func TestApproveBlockTrade(t *testing.T) {
+	t.Parallel()
+	err := e.ApproveBlockTrade(t.Context(), time.Now(), "", roleMaker)
+	require.ErrorIs(t, err, errMissingNonce)
+	err = e.ApproveBlockTrade(t.Context(), time.Time{}, "nonce", roleMaker)
+	require.ErrorIs(t, err, errZeroTimestamp)
+	err = e.ApproveBlockTrade(t.Context(), time.Now(), "nonce", "")
+	require.ErrorIs(t, err, errInvalidTradeRole)
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/77e6b3ae1ae1dc003ee89c495cff91439dba3c30 — I consolidated the approve/reject validation coverage into table-driven method tests and added direct coverage for the extracted validation helper.

exchanges/deribit/deribit_ws_endpoints.go
```diff
@@ -2237,7 +2676,7 @@ func (e *Exchange) WsSimulateBlockTrade(ctx context.Context, role string, trades
 		Trades: trades,
 	}
 	var resp bool
-	return resp, e.SendWSRequest(ctx, matchingEPL, simulateBlockPosition, input, resp, true)
+	return resp, e.SendWSRequest(ctx, matchingEPL, "private/simulate_block_trade", input, resp, true)
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/23e754a422a4259bf6ef5162ba192c2697554961 — `WsSimulateBlockTrade` now passes `&resp` into `SendWSRequest`, matching the REST implementation and allowing the JSON-RPC result to populate correctly. I also tightened the happy-path tests to assert the returned bool value instead of just checking it was non-nil.

exchanges/deribit/deribit_types.go
```diff
@@ -952,6 +957,370 @@ type BlockTradeData struct {
...
+	if err := json.Unmarshal(data, &stringValue); err == nil {
+		*b = BrokerTradeUserID(stringValue)
+		return nil
+	}
+	var integerValue uint64
+	if err := json.Unmarshal(data, &integerValue); err == nil {
+		*b = BrokerTradeUserID(strconv.FormatUint(integerValue, 10))
+		return nil
+	}
+	return errors.New("invalid broker trade user_id")
```
shazbert: Can you please wrap the incoming data in the error as well. And use a sentinel error.

exchanges/deribit/deribit_types.go
```diff
@@ -952,6 +957,370 @@ type BlockTradeData struct {
...
+	Users  []BlockRFQUserIdentity `json:"users"`
+}
+
+// BlockRFQTradeData is one public trade tape entry.
+type BlockRFQTradeData struct {
+	ID          uint64             `json:"id"`
+	Timestamp   types.Time         `json:"timestamp"`
+	Direction   string             `json:"direction"`
+	Amount      float64            `json:"amount"`
+	Price       float64            `json:"price,omitempty"`
```
shazbert: remove omitempty as this is a response structure

exchanges/deribit/deribit_types.go
```diff
@@ -952,6 +957,370 @@ type BlockTradeData struct {
...
+type BlockRFQQuoteCancelResponse struct {
+	CancelCount uint64          `json:"cancel_count"`
+	Quotes      []BlockRFQQuote `json:"quotes"`
+}
+
+// UnmarshalJSON deserialises block RFQ quote cancellation response into a BlockRFQQuoteCancelResponse instance.
+func (a *BlockRFQQuoteCancelResponse) UnmarshalJSON(data []byte) error {
+	trimmed := bytes.TrimSpace(data)
+	if len(trimmed) == 0 {
+		return errors.New("empty block rfq quote cancel response")
```
shazbert: use sentinel errors, please update/or improve coding guidelines to require sentinel errors for assert/require errors.Is code coverage.

exchanges/deribit/deribit_test.go
```diff
@@ -3033,6 +3033,689 @@ func TestWSRetrieveUserBlockTrade(t *testing.T) {
...
+
+func TestWSEditBlockRFQQuote(t *testing.T) {
+	t.Parallel()
+	_, err := e.WSEditBlockRFQQuote(t.Context(), nil)
+	require.ErrorIs(t, err, errNoArgumentPassed)
+	_, err = e.WSEditBlockRFQQuote(t.Context(), &EditBlockRFQQuoteRequest{BlockRFQQuoteID: 0, Amount: 1, Price: 10})
+	require.ErrorIs(t, err, errMissingBlockRFQQuoteIdentifier)
+	_, err = e.WSEditBlockRFQQuote(t.Context(), &EditBlockRFQQuoteRequest{BlockRFQQuoteID: 1, Amount: 1, Price: -1})
+	require.ErrorIs(t, err, errInvalidPrice)
+	// Identify via block_rfq_id + label instead of quote ID
```
shazbert: All commentary can be added to require/assert messages to keep tests concise and meaning errors will be produce, please replicate across this PR. Update coding guidelines to reflect.

exchanges/deribit/deribit.go
```diff
@@ -2395,7 +2386,584 @@ func (e *Exchange) GetUserBlockTrade(ctx context.Context, id string) ([]BlockTra
...
+			return errInvalidAmount
+		}
+	}
+	return nil
+}
+
+func validateBlockRFQHedge(hedge *BlockRFQHedgeLeg) (*BlockRFQHedgeLeg, error) {
+	if hedge == nil {
+		return nil, nil
+	}
```
shazbert, line 2595: either return an error or let this panic as the caller should populate.

exchanges/deribit/deribit.go
```diff
@@ -2395,7 +2386,584 @@ func (e *Exchange) GetUserBlockTrade(ctx context.Context, id string) ([]BlockTra
...
+		params.Set("end_id", req.EndID)
+	}
+	var resp *BrokerTradeHistoryResponse
+	return resp, e.SendHTTPAuthRequest(ctx, exchange.RestFutures, blockTradeReadEPL, http.MethodGet, "private/get_broker_trades", params, &resp)
+}
+
+func validateBlockRFQDirection(direction string) (string, error) {
+	if direction == "" {
+		return "", nil
+	}
```
shazbert, line 2514: This is an error the burden of proof should be at the caller not here. Please update coding guidelines that the burden of proof is on the caller, they can do the if check != "" and do this call.

exchanges/deribit/deribit.go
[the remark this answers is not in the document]
shazbert: Please update coding guidelines to prefer inline instead of const declarations for single use functions.

shazbert (changes requested):
💃 🚀

exchanges/deribit/deribit.go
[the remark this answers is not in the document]
thrasher-: Jarvis: This guidance already exists in https://github.com/thrasher-corp/gocryptotrader/blob/master/docs/CODING_GUIDELINES.md#L19 ("Inline endpoint paths directly in the method implementation. Avoid defining them as constants elsewhere."). This branch already follows that rule here, so I left the code unchanged.

exchanges/deribit/deribit_types.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/73bac9b7157b270164403bee894501fb3e91d6d5 — BrokerTradeUserID now returns a sentinel error and wraps the incoming payload in the error message for easier debugging and errors.Is coverage.

exchanges/deribit/deribit_types.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/73bac9b7157b270164403bee894501fb3e91d6d5 — I removed omitempty from BlockRFQTradeData.Price since this is response-side data and zero values should remain explicit.

exchanges/deribit/deribit_types.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/73bac9b7157b270164403bee894501fb3e91d6d5 — BlockRFQQuoteCancelResponse now uses a sentinel error for the empty-input path, and I added direct ErrorIs coverage for the custom unmarshal error paths. I left any broader coding-guidelines changes out of this PR.

exchanges/deribit/deribit_test.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/73bac9b7157b270164403bee894501fb3e91d6d5 — I moved the block_rfq_id + label context into the assertion messages in the touched REST and websocket tests, but kept the change local instead of sweeping the whole PR for style-only churn.

exchanges/deribit/deribit.go
thrasher-: Jarvis: I left this unchanged. hedge is optional here, so nil is a valid omission path and returning nil, nil preserves that contract; panicking would also conflict with https://github.com/thrasher-corp/gocryptotrader/blob/master/docs/CODING_GUIDELINES.md#L93.

exchanges/deribit/deribit.go
thrasher-: Jarvis: I left this unchanged. validateBlockRFQDirection("") is the shared optional-field contract for these RFQ flows, while required callers already reject empty directions before reaching it, so removing the empty case here would just duplicate validation at each caller.

exchanges/deribit/deribit_types.go
shazbert: @thrasher- regarding sentinel enforcement in coding guidelines. My opinion is that we enforce it now. No broader changes need to occur.

exchanges/deribit/deribit_test.go
shazbert: @thrasher- regarding coding guidelines update prefer to enforce now. No broader sweeping changes need to occur.

exchanges/deribit/deribit.go
shazbert: validateBlockRFQDirection("") should not be called in the first place, please update coding guidelines sweeping changes need not occur.

exchanges/deribit/deribit.go
shazbert: function should not be called if param is nil, burden of proof should be at the caller. Please return error.

thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/125f875cd034414902c71d950d3f27a47db61981 — validateBlockRFQHedge now returns common.ErrNilPointer for nil input, and I moved the optional hedge check to the REST and websocket callers so nil hedge omission remains supported without silently accepting nil inside the validator.

exchanges/deribit/deribit_types.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/0821d705d000873f57c3d00c2c18967f9b6df030 — I updated docs/CODING_GUIDELINES.md to explicitly prefer package-level sentinel errors for validation/parsing/custom-unmarshal paths when stable errors.Is matching is needed, to wrap sentinels with %w when adding context, and to add direct ErrorIs coverage for sentinel paths.

exchanges/deribit/deribit.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/5b608bfbf92bdc970f707be530fb86964e9a0ffe — validateBlockRFQDirection now rejects empty input, and the optional top-level Add/Accept Block RFQ request paths only call it when direction is actually provided. Empty direction is now treated as caller error in the validator-backed paths (legs/hedge), while omitted optional top-level direction fields remain supported without silently treating "" as a valid direction.

shazbert (approved):
ch ACK.

gloriousCode:
I verified the websocket request payload for `private/invalidate_block_trade_signature` and it currently sends:

```json
{"jsonrpc":"2.0","id":"test-d6ad-75cc-b1f0-test","method":"private/invalidate_block_trade_signature","params":{"signature":["verified_signature_string"]}}
```

That `signature` value is being encoded as an array (because `url.Values` is used), rather than a string as shown in Deribit docs.

Suggestion: use either:
- `map[string]string{"signature": signature}`
- or `map[string]any{"signature": signature}`

so the params serialise as:

```json
"params": {"signature": "verified_signature_string"}
```

This applies only to WS

exchanges/deribit/deribit_ws_endpoints.go
gloriousCode: This function has changed, at least in the api docs https://docs.deribit.com/api-reference/block-trade/private-get_block_trade
It does not return a naked (🍑) slice but `"trades": [{...}],"id":"1337","timestamp":1337`

gloriousCode: Teh response signature here has also changed to be
`trades,id,timestamp` instead of just a nekkid (🍑 ) slice

gloriousCode (changes requested):
This applies to both rest and ws

exchanges/deribit/deribit_ws_endpoints.go
thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/766fe7c3dcf74b5edea1e3faaffcf30a6746b7f2 — `GetUserBlockTrade` and `WSRetrieveUserBlockTrade` now decode the current Deribit response wrapper (`id`, `timestamp`, `trades`) into `BlockTradeCollection` instead of treating it as a bare slice. I also updated the local fixture coverage to match the documented payload shape.

thrasher-: Jarvis: Done in https://github.com/thrasher-/gocryptotrader/commit/766fe7c3dcf74b5edea1e3faaffcf30a6746b7f2 — `ExecuteBlockTrade` and `WSExecuteBlockTrade` now decode the current Deribit response wrapper into `ExecutedBlockTradeCollection`, so the block-trade metadata and trade list are populated correctly instead of being treated as a bare slice.

exchanges/deribit/deribit_ws_endpoints.go
```diff
@@ -2130,7 +2130,7 @@ func (e *Exchange) WsInvalidateBlockTradeSignature(ctx context.Context, signatur
 	params := url.Values{}
 	params.Set("signature", signature)
 	var resp string
-	err := e.SendWSRequest(ctx, nonMatchingEPL, invalidateBlockTradesSignature, params, &resp, true)
+	err := e.SendWSRequest(ctx, blockTradeWriteEPL, "private/invalidate_block_trade_signature", params, &resp, true)
```
gloriousCode: This fell through the cracks, so here it is a second time 🌈 

I verified the websocket request payload for private/invalidate_block_trade_signature and it currently sends:

{"jsonrpc":"2.0","id":"test-d6ad-75cc-b1f0-test","method":"private/invalidate_block_trade_signature","params":{"signature":["verified_signature_string"]}}

That signature value is being encoded as an array (because url.Values is used), rather than a string as shown in Deribit docs.

Suggestion: use either:

    map[string]string{"signature": signature}
    or map[string]any{"signature": signature}

so the params serialise as:

"params": {"signature": "verified_signature_string"}

This applies only to WS

exchanges/deribit/deribit_ws_endpoints.go
```diff
@@ -2141,12 +2141,459 @@ func (e *Exchange) WsInvalidateBlockTradeSignature(ctx context.Context, signatur
...
+		return nil, errNoArgumentPassed
+	}
+	checkedLegs, err := validateCreateBlockRFQLegs(req.Legs)
+	if err != nil {
+		return nil, err
+	}
+	err = validateBlockRFQTradeAllocations(req.TradeAllocations)
+	if err != nil {
+		return nil, err
+	}
```
gloriousCode, line 2269: can if-short but I'm okay without

exchanges/deribit/deribit_test.go
```diff
@@ -3006,13 +3006,15 @@ func TestWSExecuteBlockTrade(t *testing.T) {
...
+const getUserBlocTradeResponseJSON = `{ "id": "61", "timestamp": 1565089523720, "trades": [ { "trade_seq": 37, "trade_id": "92437", "timestamp": 1565089523719, "tick_direction": 3, "state": "filled", "price": 0.0001, "order_type": "limit", "order_id": "343062", "matching_id": null, "liquidity": "T", "iv": 0, "instrument_name": "BTC-9AUG19-10250-C", "index_price": 11738, "fee_currency": "BTC", "fee": 0.00025, "direction": "sell", "block_trade_id": "61", "amount": 10 }, { "trade_seq": 25350, "trade_id": "92435", "timestamp": 1565089523719, "tick_direction": 3, "state": "filled", "price": 11590, "order_type": "limit", "order_id": "343058", "matching_id": null, "liquidity": "T", "instrument_name": "BTC-PERPETUAL", "index_price": 11737.98, "fee_currency": "BTC", "fee": 0.00000164, "direction": "buy", "block_trade_id": "61", "amount": 190 } ] }`
 
 func TestGetUserBlocTrade(t *testing.T) {
```
shazbert: `Bloc` -> `Block`? and update const or change everything to `Bloc`

shazbert (changes requested):
One minor thingerino.

exchanges/deribit/deribit_test.go
thrasher-: Jarvis: Done, good catch. Renamed the typoed `Bloc` identifiers to `Block` for consistency with the block trade naming used elsewhere in the tests and endpoint.

shazbert (changes requested):
Changes look good. 

suggestion: There is a test failure in deribit `TestWSRetrieveCombos` it seems intermittent: https://github.com/thrasher-corp/gocryptotrader/actions/runs/24703395412/job/72251285464?pr=2208#step:11:219

thrasher-:
Jarvis: Good catch. I updated `TestWSRetrieveCombos` to use a fixed valid combo currency (`BTC`) instead of the live-selected future combo pair base, which could intermittently send an invalid currency to Deribits `get_combos` websocket endpoint.

shazbert (approved):
Changes ACK! Thanks

gloriousCode (changes requested):

exchanges/deribit/deribit_ws_endpoints.go
thrasher-: Jarvis: Good catch. I updated the websocket invalidate call to send `signature` as a plain string instead of using `url.Values`, so the JSON-RPC params now match Deribits documented payload shape for `private/invalidate_block_trade_signature`.

thrasher-:
Jarvis: Left the optional if-short style suggestion as-is for now since it was non-blocking and you said you were okay without. Happy to shorten it as a follow-up if you’d prefer.

exchanges/deribit/deribit_ws_endpoints.go
thrasher-: Jarvis: Left this as-is for now since its style-only and you said you were okay without. Happy to shorten it as a follow-up if youd prefer.

gloriousCode (approved):
tACK! Thanks JarJar

PR #2218, opened 2026-03-22
exmo: replace provider live tests with deterministic mocks

thrasher-:
## Summary
- Replace the live EXMO provider endpoint tests with deterministic `httptest` responses
- Change `CryptoPaymentProvider.Min` and `Max` to `types.Number` so EXMO responses decode correctly when values are returned as either strings or raw numbers
- Add direct assertions for provider decoding and transfer-chain filtering from enabled deposit networks only

## Why
GitHub-hosted CI runs from US-hosted runners, and EXMO’s provider endpoint is geoblocked there. That caused the live tests to receive empty results in CI and fail for reasons unrelated to the EXMO implementation. This keeps the coverage while removing the region-dependent behaviour.

## Testing
- `go test ./exchanges/exmo -run 'TestGetCryptoPaymentProvidersList|TestGetAvailableTransferChains'`

### DIFF
--- a/exchanges/exmo/exmo_test.go
+++ b/exchanges/exmo/exmo_test.go
@@ -2,6 +2,8 @@ package exmo
 
 import (
 	"log"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"testing"
 	"time"
@@ -18,6 +20,7 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/exchanges/ticker"
 	testexch "github.com/thrasher-corp/gocryptotrader/internal/testing/exchange"
 	"github.com/thrasher-corp/gocryptotrader/portfolio/withdraw"
+	"github.com/thrasher-corp/gocryptotrader/types"
 )
 
 const (
@@ -29,6 +32,10 @@ const (
 var (
 	e        *Exchange
 	testPair = currency.NewBTCUSD().Format(currency.PairFormat{Uppercase: true, Delimiter: "_"})
+
+	// useMockCryptoPaymentProviderTests keeps provider endpoint tests deterministic in CI.
+	// Flip it locally to exercise the live EXMO provider endpoint until EXMO gets a full mock/live test split.
+	useMockCryptoPaymentProviderTests = true
 )
 
 func TestMain(m *testing.M) {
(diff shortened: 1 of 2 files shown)


thrasher-:
@codex please review

exchanges/exmo/exmo_test.go
```diff
@@ -464,18 +467,77 @@ func TestUpdateTickers(t *testing.T) {
...
+			]
+		}`))
+		assert.NoError(t, writeErr, "Writing response should not error")
+	}))
+	defer server.Close()
+
+	err = ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL)
+	require.NoError(t, err, "SetRunningURL must not error")
+
+	resp, err := ex.GetCryptoPaymentProvidersList(t.Context())
```
shazbert, line 514: Suggestion: Mocking now obfuscates exmo api updates and problems. We can find and abstract this function:

```go
func isCITest() bool {
	return os.Getenv("CI") == "true"
}
```

So if we detect its a CI env we can mock but locally it can run a rest request

shazbert (approved):
Changes ACK. 

One suggestion.

exchanges/exmo/exmo_test.go
```diff
@@ -464,18 +467,77 @@ func TestUpdateTickers(t *testing.T) {
 
 func TestGetCryptoPaymentProvidersList(t *testing.T) {
 	t.Parallel()
-	_, err := e.GetCryptoPaymentProvidersList(t.Context())
-	if err != nil {
-		t.Fatal(err)
+
+	ex := new(Exchange)
```
gloriousCode, line 475: I'd rather this actually go with standard `http.json`, `exmo_live_test.go` and `exmo_mock_test.go`
This feels too reactionary to a singular issue.
At least by sticking to that, we already have a CI build for live tests (even though it will be blocked) which also addresses @shazbert's issue

Way less noisy for test design too

exchanges/exmo/exmo_test.go
thrasher-: Jarvis: Good point. I’ve gone with a local bool toggle instead of branching on `CI`, since the failure mode here was region/geoblocking rather than CI itself. Default stays deterministic so CI and blocked environments don’t flap, but it can be flipped locally to exercise the live EXMO provider endpoint.

exchanges/exmo/exmo_test.go
thrasher-: Jarvis: Fair point. I agree the standard `http.json` + live/mock split makes sense for a full EXMO testing pass, but I didn’t want to half-introduce that structure in a narrow fix that only touches the provider parsing path. For now I’ve added a local bool toggle so the tests remain deterministic by default while still allowing live endpoint coverage locally, and I’d prefer to leave the full EXMO mock/live conversion for a broader upgrade pass.

PR #2240, opened 2026-04-28
Okx, Kucoin: Fix live API tests

thrasher-:
## Summary
- Update KuCoin `TestGetSymbols` to use the active documented `USDS` market instead of `ETF`, which currently returns no symbols.
- Keep coverage for both unfiltered and market-filtered KuCoin symbol requests, and assert the filtered result is non-empty and narrower than the full symbol list.
- Update OKX `TestGetOpenInterestData` so `uly`, `instFamily`, and `instId` are derived from the same live option instrument record.
- Preserve OKX `_UM` instrument family/ID values when the selected instrument uses them, avoiding mismatched request parameters that trigger `51002`.

## Test
- `go test ./exchanges/kucoin -run '^TestGetSymbols$' -count=1`
- `go test ./exchanges/okx -run '^TestGetOpenInterestData$' -count=1`
- `go test ./exchanges/okx -run 'TestGetTickers|TestGetOpenInterestData' -count=1`
- `go test ./exchanges/kucoin -count=1`
- `go test ./exchanges/okx -count=1`
- `codespell exchanges/kucoin/kucoin_test.go exchanges/okx/okx_test.go`
- `make misc_checks`
- `make lint`

### DIFF
--- a/exchanges/kucoin/kucoin_test.go
+++ b/exchanges/kucoin/kucoin_test.go
@@ -73,13 +73,15 @@ func TestMain(m *testing.M) {
 
 func TestGetSymbols(t *testing.T) {
 	t.Parallel()
-	symbols, err := e.GetSymbols(t.Context(), "")
+	allSymbols, err := e.GetSymbols(t.Context(), "")
 	assert.NoError(t, err)
-	assert.NotEmpty(t, symbols)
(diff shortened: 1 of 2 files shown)


shazbert (approved):
ch ACK

PR #2248, opened 2026-05-07
gateio: move isolated margin endpoints into dedicated files, expand support

shazbert:
# PR Description

- Move isolated margin REST methods into margin_isolated.go
- Move isolated margin response/request structs into margin_isolated_types.go
- Add focused isolated margin tests in margin_isolated_test.go
- Remove duplicated isolated/uni margin helpers and tests from gateio.go and gateio_test.go
- Update account balance sync to use GetIsolatedMarginAccountList
- Add Gate frontend API URL support for isolated margin pool loan data
- Add validation around USDT transferable requests and isolated margin edge cases
- Add endpoint-specific EPLs for isolated margin endpoints
- Split unified interest records from isolated margin interest-record rate limits
- Extend Gate.io EPL coverage in TestRateLimits

dependency: #2207  

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,343 @@
...
+	if leverage <= 0 {
+		return fmt.Errorf("%w, leverage must be greater than 0", errInvalidLeverage)
+	}
+	payload := struct {
+		Leverage float64 `json:"leverage"`
+		Pair     string  `json:"currency_pair"`
+	}{
+		Leverage: leverage,
+		Pair:     market.String(),
+	}
```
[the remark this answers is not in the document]
shazbert: its not optional turns out will remove comment

exchanges/gateio/margin_isolated_types.go
```diff
@@ -0,0 +1,140 @@
...
+	Message   string     `json:"message"`
+	Data      struct {
+		Total       int          `json:"total"`
+		List        []Loan       `json:"list"`
+		VIPSettings []VIPSetting `json:"vip_settings"`
+	} `json:"data"`
+}
+
+// Loan represents an individual loan entry
+type Loan struct {
```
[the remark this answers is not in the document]
shazbert: externally it looks like gateio.Loan so nope, if other people want me to change it I can

exchanges/gateio/margin_isolated_types.go
```diff
@@ -0,0 +1,140 @@
...
+	StockTotalLendAvailableFiat types.Number  `json:"stock_total_lend_available_fiat"`
+	StockUserMaxBorrowAmount    types.Number  `json:"stock_user_max_borrow_amount"`
+	Leverage                    types.Number  `json:"leverage"`
+}
+
+// VIPSetting represents the VIP level settings for margin trading, including various rates and limits.
+type VIPSetting struct {
+	VIPLevel             uint8        `json:"vip_level"`
+	LendDownRate         types.Number `json:"lend_down_rate"`
+	BorrowUpRate         types.Number `json:"borrow_up_rate"`
```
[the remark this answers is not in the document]
shazbert: same here

exchanges/gateio/gateio.go
```diff
@@ -1316,29 +1319,26 @@ func (e *Exchange) ConvertSmallBalances(ctx context.Context, currs ...currency.C
...
 
-// GetEstimatedInterestRate retrieves estimated interest rate for provided currencies
-func (e *Exchange) GetEstimatedInterestRate(ctx context.Context, currencies []currency.Code) (map[string]types.Number, error) {
-	if len(currencies) == 0 {
-		return nil, currency.ErrCurrencyCodesEmpty
+// QueryInterestDeductionRecords retrieves unified interest deduction records.
+func (e *Exchange) QueryInterestDeductionRecords(ctx context.Context, ccy currency.Code, page, limit int64, from, to time.Time, loanType string) ([]LoanInterestDeductionRecord, error) {
+	params := url.Values{}
+	if !ccy.IsEmpty() {
+		params.Set("currency", ccy.String())
```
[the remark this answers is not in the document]
shazbert: Nope

exchanges/request/request_test.go
```diff
@@ -177,6 +177,17 @@ func TestCheckRequest(t *testing.T) {
...
+	ctx = WithHeaders(ctx, http.Header{
+		"Content-Type": {"context override"},
+		"User-Agent":   {"context agent"},
+		"X-Values":     {"one", "two"},
+	})
+	req, err = check.validateRequest(ctx, r)
+	require.NoError(t, err, "validateRequest must not error")
+	assert.Equal(t, "context override", req.Header.Get("Content-Type"), "context header should override item header")
+	assert.Equal(t, "context agent", req.UserAgent(), "context header should override requester user agent")
+	assert.Equal(t, []string{"one", "two"}, req.Header.Values("X-Values"), "context header values should be preserved")
```
gloriousCode, line 296: Add scenario where something is preserved via not being overridden

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,326 @@
...
+	if arg.Currency.IsEmpty() {
+		return currency.ErrCurrencyCodeEmpty
+	}
+	if arg.Type != "borrow" && arg.Type != "repay" {
+		return errors.New("invalid isolated margin loan type: must be \"borrow\" or \"repay\"")
+	}
+	if arg.Amount <= 0 {
+		return fmt.Errorf("%w, amount must be greater than 0", errInvalidAmount)
+	}
+	return e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, marginCreateUniLoanEPL, http.MethodPost, "margin/uni/loans", nil, arg, nil)
```
gloriousCode, line 208: This duplicates `UniLoanBorrowOrRepay` / `UniLoanBorrowRepayParam` from #2207: both methods POST `margin/uni/loans`, but this method uses `marginCreateUniLoanEPL` while the existing method uses `marginUniLoanBorrowRepayEPL`, so one remote endpoint gets two independent rate-limit buckets. The existing method is also documented as operating a unified margin account even though Gate exposes actual unified loans under `/unified/loans`; `/margin/uni/loans` belongs to the isolated-margin API. Please reconcile the methods, request types, documentation, and EPLs. The same duplication exists between `GetIsolatedMarginInterestDeductionRecords` and `GetUniLoanInterestRecords`.

Gate API documentation: [isolated-margin borrow or repay](https://www.gate.com/docs/developers/apiv4/en/#borrow-or-repay-2), [unified-account borrow or repay](https://www.gate.com/docs/developers/apiv4/en/#borrow-or-repay).

exchanges/gateio/margin_isolated_types.go
```diff
@@ -0,0 +1,142 @@
...
+	MaximumPermissibleLeverage types.Number `json:"leverage"`
+}
+
+// IsolatedBorrowRepayRequest represents the request parameters for the isolated margin borrow-or-repay
+// endpoint (POST margin/uni/loans).
+type IsolatedBorrowRepayRequest struct {
+	CurrencyPair currency.Pair `json:"currency_pair"`
+	Currency     currency.Code `json:"currency"`
+	Type         string        `json:"type"` // Type is either "borrow" or "repay".
+	Amount       types.Number  `json:"amount"`
```
gloriousCode, line 70: Gate currently supports the optional `repaid_all` field on `CreateUniLoan`; when true, it requests full repayment and overrides `amount`. This new public request type cannot express that operation, so full repayment is unavailable through `IsolatedMarginBorrowOrRepay`. Please add the field and focused coverage if this method is intended to support the full borrow/repay endpoint.

Gate API documentation: [borrow or repay](https://www.gate.com/docs/developers/apiv4/en/#borrow-or-repay-2).

gloriousCode (changes requested):
DOO DOO DOO DOO DOO DOO DOO

exchanges/gateio/margin_isolated_types.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2248/commits/ed3c00a873bdf66d660ec31ff7d9fd41e5853eb6

thrasher-:
@codex please review

gloriousCode (approved):
tACK you shazbert

exchanges/gateio/gateio_wrapper.go
```diff
@@ -791,6 +789,50 @@ func (e *Exchange) UpdateAccountBalances(ctx context.Context, a asset.Item) (acc
 	return subAccts, e.Accounts.Save(ctx, subAccts, true)
 }
 
+func setIsolatedMarginAccountBalances(balances *accounts.CurrencyBalances, response []MarginAccountItem) error {
```
thrasher-, line 793: Heimdall: **Blocker** — one isolated market with an empty currency discards the whole `asset.Margin` balance update.

**Fix:** skip entries where `Currency` is empty rather than returning the error. Don't revert to `Set` — moving to `Add` is right and fixes the per-currency clobbering @Copilot flagged.

**Why:** `addIsolatedMarginAccountBalance` propagates `CurrencyBalances.Add`, which rejects an empty code ([balance.go:61](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchange/accounts/balance.go#L61)). The triggering response is well-formed and matches Gate's documented schema — `base`/`quote` and their `currency` are all `[optional]`, and `account_type: "inactive"` markets come back for anything not activated:

```
[{"currency_pair":"FOO_USDT","account_type":"inactive"}]
  -> err=<nil>, baseCcy="", quoteCcy=""
```

That's a clean decode straight into the abort. It happens mid-loop, so everything already collected goes with it:

```
require.ErrorIs(err, currency.ErrCurrencyCodeEmpty)          // passes
balances collected before abort: map[BTC:{...} USDT:{...}]   // discarded
```

The old `Set` path tolerated it, so this regresses a path that works today.

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,338 @@
...
+// SetUserMarketLeverageMultiplier sets the user's market leverage multiplier for isolated margin accounts.
+// Success returns 204 No Content with empty body.
+func (e *Exchange) SetUserMarketLeverageMultiplier(ctx context.Context, market currency.Pair, leverage float64) error {
+	if market.IsEmpty() {
+		return currency.ErrCurrencyPairEmpty
+	}
+	if leverage <= 0 {
+		return fmt.Errorf("%w, leverage must be greater than 0", errInvalidLeverage)
+	}
+	payload := struct {
```
thrasher-, line 309: Heimdall: **Should fix** — `SetUserMarketLeverageMultiplier` sends `leverage` as a bare JSON number, so the endpoint most likely never works.

**Fix:** retype `Leverage` to `types.Number`, which marshals to a quoted string. While you're there, the guidelines want a named `XRequest` struct passed by pointer rather than the anonymous inline struct passed by value.

**Why:** all three generated clients agree the field is a required string:

```
gateapi-go      Leverage string `json:"leverage"`
gateapi-python  'leverage': 'str'
gateapi-java    private String leverage;
```

This PR already sends amounts as strings via `types.Number` in `IsolatedBorrowRepayRequest`, so it's inconsistent with its own neighbour as well as the spec. I couldn't confirm the rejection live — no credentials — so if you've seen a 200 from this as-is, say so and I'll drop it.

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,338 @@
...
+		if limit > marginPoolLoanPageLimit {
+			return nil, fmt.Errorf("%w: maximum %d", errInvalidLimit, marginPoolLoanPageLimit)
+		}
+		params.Set("limit", strconv.FormatUint(limit, 10))
+	}
+
+	path := common.EncodeURLValues("spot_loan/margin/margin_loan_info", params)
+
+	var resp *IsolatedMarginPoolLoanResponse
+	return resp, e.SendHTTPRequest(ctx, exchange.EdgeCase1, publicIsolatedMarginPoolLoansEPL, path, &resp)
```
thrasher-: Heimdall: **Should fix** — `GetIsolatedMarginPoolLoans` returns 403 for every caller except the test.

**Fix:** set the headers inside the method — a browser `User-Agent` plus `Accept-Encoding` is enough — and widen the test's `require.ErrorContains(t, err, "504")` to cover 403.

**Why:** the workaround lives only in [margin_isolated_test.go:273](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_test.go#L273), so anything calling the exported method through the engine or as a library gets:

```
err=GateIO unsuccessful HTTP status code: 403 raw response: 403
```

It's an Akamai fingerprint score rather than a fixed header set, which is why the 11-header block is both more than needed and not the actual requirement:

```
UA + Accept-Encoding                200 200 200
UA + Accept-Language                403 403 403
UA + Accept                         403 403 403
GCT default request                 403 (6/6)
full 11-header set, no Accept-Enc   200 200 200
```

So the method is unusable as shipped, and CI fails on the 403 it will actually hit rather than the 504 the GoDoc anticipates. Scoring is tuned server-side, so whatever subset works today can stop working without notice — worth the test tolerating both.

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,338 @@
...
+
+// GetIsolatedMarginMarketLeverageTiers retrieves the market leverage tiers for isolated margin accounts.
+func (e *Exchange) GetIsolatedMarginMarketLeverageTiers(ctx context.Context, pair currency.Pair) ([]IsolatedMarginLendingTier, error) {
+	if pair.IsEmpty() {
+		return nil, currency.ErrCurrencyPairEmpty
+	}
+	params := url.Values{}
+	params.Set("currency_pair", pair.String())
+	var response []IsolatedMarginLendingTier
+	return response, e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, marginMarketLoanMarginTiersEPL, http.MethodGet, "margin/loan_margin_tiers", params, nil, &response)
```
thrasher-: Heimdall: **Should fix** — `GetIsolatedMarginMarketLeverageTiers` authenticates a public endpoint, so it needs credentials it doesn't need and burns the private bucket.

**Fix:** swap to `SendHTTPRequest` with a public EPL, and drop the `SkipTestIfCredentialsUnset` gate so the test actually runs in CI.

**Why:** `GET /margin/loan_margin_tiers` is documented "No authorization required", and live:

```
$ curl -s '.../api/v4/margin/loan_margin_tiers?currency_pair=BTC_USDT'
[{"upper_limit":"0","mmr":"0.02","leverage":"20"}, ...]

$ curl -s '.../api/v4/margin/user/loan_margin_tiers?currency_pair=BTC_USDT'
{"label":"MISSING_REQUIRED_HEADER","message":"Missing required header: Timestamp"}
```

The user variant directly above it is correctly private — only the market one is mis-scoped.

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,338 @@
...
+		params.Set("page", strconv.FormatUint(page, 10))
+	}
+	if limit > 0 {
+		if limit > marginPoolLoanPageLimit {
+			return nil, fmt.Errorf("%w: maximum %d", errInvalidLimit, marginPoolLoanPageLimit)
+		}
+		params.Set("limit", strconv.FormatUint(limit, 10))
+	}
+
+	path := common.EncodeURLValues("spot_loan/margin/margin_loan_info", params)
```
thrasher-, line 349: Heimdall: **Should fix** — application errors from the frontend endpoint come back as a successful empty result.

**Fix:** return an error when `Code != 200` before handing the response back.

**Why:** the envelope's `code`/`message` are unmarshalled at [margin_isolated_types.go:107](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_types.go#L107) and never read. The API answers HTTP 200 and puts the failure in the body:

```
$ curl '...margin_loan_info?page=18446744073709551615&limit=100'
{"code":-1,"message":"Type mismatch: 18446744073709551615"}
```

Currently reported as success with zero loans, indistinguishable from a genuinely empty page.

exchanges/gateio/ratelimiter.go
```diff
@@ -112,6 +116,11 @@ const (
 	marginGetAutoRepaySettingsEPL
 	marginGetMaxTransferEPL
 	marginGetMaxBorrowEPL
+	marginUniBorrowableEPL
```
thrasher-, line 115: Heimdall: **Should fix** — three EPL constants are now dead.

**Fix:** delete [marginAccountListEPL](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/ratelimiter.go#L101), [marginGetMaxBorrowEPL](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/ratelimiter.go#L118) and [marginEstimateRateEPL](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/ratelimiter.go#L135) along with their `packageRateLimits` entries.

**Why:** each had exactly one caller at the baseline; this PR added renamed replacements (`marginUserAccountListEPL`, `marginUniBorrowableEPL`, `marginUniEstimateRateEPL`) without removing the originals. `TestRateLimits` only detects *missing* rate limits, never unused ones, so nothing in CI will surface these.

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,338 @@
...
+	if currencyPair.IsPopulated() {
+		params.Set("currency_pair", currencyPair.String())
+	}
+	if !ccy.IsEmpty() {
+		params.Set("currency", ccy.String())
+	}
+	if page > 0 {
+		params.Set("page", strconv.FormatUint(page, 10))
+	}
+	if limit > 0 {
```
thrasher-, line 248: Heimdall: **Should fix** — the documented limit constraints are applied inconsistently: we reject requests Gate accepts and accept ones it rejects.

**Fix:** move the cap onto `GetIsolatedMarginLoans`, `GetIsolatedMarginLoanRecords` and `QueryInterestDeductionRecords`, whose docs all state `maximum: 100`, and drop it from `GetIsolatedMarginPoolLoans`.

**Why:** `/margin/uni/interest_records` is the one endpoint in this group with no documented maximum, and it's the only one we cap. Meanwhile the frontend endpoint serves whatever you ask for:

```
limit=100 -> code 200, list 100
limit=200 -> code 200, list 200
limit=500 -> code 200, list 500
```

Given the GoDoc warns about 504s on large responses, capping the one endpoint where a caller might want *fewer* round trips is backwards.

exchanges/gateio/margin_isolated_types.go
```diff
@@ -0,0 +1,142 @@
...
+	RepaidAll    bool          `json:"repaid_all,omitempty"` // If true, the amount is ignored and the entire outstanding loan is repaid.
+}
+
+// IsolatedMarginLoanResponse represents a record of borrowing or repaying in an isolated margin account.
+type IsolatedMarginLoanResponse struct {
+	Type         string        `json:"type"` // "borrow" or "repay" only returned in record response
```
thrasher-, line 76: Heimdall: **Should fix** — the shared struct's `type` field carries two different vocabularies and the comment documents only one.

**Fix:** either split the struct per endpoint, or reword the comment to cover both.

**Why:** the comment says `"borrow" or "repay" only returned in record response`, but `GET /margin/uni/loans` does return `type` — as `platform`/`margin`, the loan source, per Gate's `UniLoan` model. Only `/margin/uni/loan_records` uses `borrow`/`repay`. Both fit a Go string so nothing fails at decode; it just surfaces later as a wrong filter downstream.

exchanges/gateio/margin_isolated_test.go
```diff
@@ -0,0 +1,293 @@
+package gateio
```
thrasher-, line 1: Heimdall: **Should fix** — one sentinel in the new code has no offline assertion, and the one untested error path is where the blocker lives.

**Fix:** four small tests, no network needed for any of them.
1. `GetIsolatedMarginMaxTransferableAmount(ctx, currency.EMPTYCODE, ...)` -> `currency.ErrCurrencyCodeEmpty` ([margin_isolated.go:95](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated.go#L95)). The test at `:69` only exercises the USDT-pair rule, so this branch has never run.
2. `setIsolatedMarginAccountBalances` with an entry whose currency is empty. That single test would have caught the blocker.
3. `setCrossMarginAccountBalances(&balances, nil)` — the nil guard at [gateio_wrapper.go:819](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/gateio_wrapper.go#L819) is untested.
4. A table test on `setUnixTimeRangeParams`.

**Why:** the sentinel audit is otherwise in good shape — `errInvalidLimit`, `errInvalidIsolatedMarginLoanType`, `errAmountOverriddenByRepaidAll`, `errInvalidRepaidAllOperation`, `errTooManyCurrencyCodes`, `errInvalidLeverage`, `currency.ErrCurrencyCodesEmpty` and `common.ErrStartAfterEnd` all have direct `ErrorIs` coverage, and `IsolatedMarginBorrowOrRepay` asserts all seven of its own. Which is what makes the two misses stand out rather than blend in.

The balance helper is the one that matters:

```
setIsolatedMarginAccountBalances   66.7%   <- both error returns never execute
addIsolatedMarginAccountBalance   100.0%   <- misleading
```

That 100% is a single return statement, so it reads identically whether `Add` succeeds or returns `ErrCurrencyCodeEmpty`. Statement coverage actively hides the blocker here, which is why I'd want the error case asserted explicitly rather than trusting the number.

`setUnixTimeRangeParams` reads 100.0%, but only three cases are asserted — both-set, from-only, `ErrStartAfterEnd`. Not covered: to-only, both-zero, `ErrStartEqualsEnd`, `ErrStartAfterTimeNow`, and `time.Unix(0,0)` -> `ErrDateUnset`. It now gates 17 call sites and 16 of them never exercise the new `return nil, err`, so a table test on the helper is the cheap way to pin the behaviour down.

Last one: there are no JSON fixtures for any new or changed response type. `MarginAccountItem.CurrencyPair` went `string` -> `currency.Pair` and `Risk` went `string` -> `types.Number`; the whole `IsolatedMarginPoolLoanResponse` tree is new and fed by an undocumented endpoint. I probed the changed types and they behave — absent `risk`/`mmr` decode to 0, empty `currency_pair` to `EMPTYPAIR` — but nothing locks that in, and a delimiterless pair fails the entire response rather than the one entry:

```
"currency_pair":"BTCUSDT"  -> error creating currency pair from BTCUSDT ...
```

Same pattern as `TestCandlestickUnmarshalJSON` and `TestUnmarshalJSONOrderbookLevels`, so there's precedent in the package.

exchanges/gateio/margin_isolated_types.go
```diff
@@ -0,0 +1,142 @@
...
+	Message   string     `json:"message"`
+	Data      struct {
+		Total       int          `json:"total"`
+		List        []Loan       `json:"list"`
+		VIPSettings []VIPSetting `json:"vip_settings"`
+	} `json:"data"`
+}
+
+// Loan represents an individual loan entry
+type Loan struct {
```
thrasher-, line 128: Heimdall: **Minor** — `Loan` silently drops one of the two fiat fields.

**Fix:** add `MoneyTotalLendAvailableFiat types.Number` with the `money_total_lend_available_fiat` tag.

**Why:** live responses carry both sides; we map only the stock one:

```
money_total_lend_available    money_total_lend_available_fiat   <- unmapped
stock_total_lend_available    stock_total_lend_available_fiat   <- mapped
```

Not caught today because `require.NotEmpty(t, got)` asserts on the envelope pointer rather than `Data.List`, so nothing on this type would notice a dropped field.

exchanges/gateio/gateio_wrapper.go
```diff
@@ -791,6 +789,50 @@ func (e *Exchange) UpdateAccountBalances(ctx context.Context, a asset.Item) (acc
...
+			return err
+		}
+		if err := addIsolatedMarginAccountBalance(balances, response[i].Quote); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func addIsolatedMarginAccountBalance(balances *accounts.CurrencyBalances, balance AccountBalanceInformation) error {
```
thrasher-, line 805: Heimdall: **Minor** — isolated and cross disagree on what counts as borrowed, and the justification lives in a comment rather than in the code.

**Fix:** nothing needed if the observation holds — but if you're confident, subtracting `Interest` when it's non-zero costs nothing and makes the code self-defending.

**Why:** cross subtracts `borrowed + interest` (`:818`), isolated subtracts only `borrowed`. The comment says Gate applies isolated interest deductions to `Available` and never populates `Interest`, which is plausible, but the schema defines `interest` as "Unpaid interest" and we still unmarshal it — so the day Gate starts populating it we overstate free collateral with nothing flagging it. Can't verify without credentials, so this rests on your observation.

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,338 @@
...
+	return response, e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, marginAutoRepayEPL, http.MethodPost, "margin/auto_repay", params, nil, &response)
+}
+
+// GetIsolatedMarginMaxTransferableAmount get the max transferable amount for a specific margin currency.
+func (e *Exchange) GetIsolatedMarginMaxTransferableAmount(ctx context.Context, ccy currency.Code, currencyPair currency.Pair) (*MaxTransferAndLoanAmount, error) {
+	if ccy.IsEmpty() {
+		return nil, currency.ErrCurrencyCodeEmpty
+	}
+	params := url.Values{}
+	if currencyPair.IsEmpty() && ccy.Equal(currency.USDT) {
```
thrasher-: Heimdall: **Minor** — the USDT rule isn't in the spec, so it reads as a bug rather than a deliberate guard.

**Fix:** nothing needed if it came from a live rejection — just say so in the comment.

**Why:** Gate lists `currency_pair` as unconditionally optional on `GET /margin/transferable`, with no USDT carve-out in any generated client. If the API really does reject `currency=USDT` without a pair then the guard is a favour to callers, but as written the next person to read it will assume we invented the restriction and delete it.

exchanges/gateio/gateio.go
```diff
@@ -1327,80 +1341,26 @@ func (e *Exchange) ConvertSmallBalances(ctx context.Context, currs ...currency.C
...
-	}
-	return e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, marginUniLoanBorrowRepayEPL, http.MethodPost, gateioMarginUniLoans, nil, arg, nil)
-}
-
-// GetUniLoanInterestRecords retrieves interest deduction records for unified
-// margin loans. Optionally filters by currency pair or by currency code.
-// Pagination: page starts at 1, limit defaults to 100 (max 100).
-func (e *Exchange) GetUniLoanInterestRecords(ctx context.Context, pair currency.Pair, ccy currency.Code, page, limit uint64, from, to time.Time) ([]UniLoanInterestRecord, error) {
+// QueryInterestDeductionRecords retrieves unified interest deduction records.
+func (e *Exchange) QueryInterestDeductionRecords(ctx context.Context, ccy currency.Code, page, limit int64, from, to time.Time, loanType string) ([]LoanInterestDeductionRecord, error) {
```
thrasher-: Heimdall: **Nit** — `QueryInterestDeductionRecords` uses `int64` for pagination and takes an undocumented enum.

**Fix:** `uint64` for `page`/`limit`, and document `loanType` on the method or validate it.

**Why:** `int64` lets negatives through, and they're then silently dropped by the `> 0` guards rather than rejected — `uint64` makes that unrepresentable, which is why the guidelines default to it. On `loanType`, `/unified/interest_records` takes `platform`/`margin` (default `margin`), but the sibling `GetIsolatedMarginLoanRecords` takes `borrow`/`repay` under the same parameter name — two adjacent methods, same name, different vocabularies.

exchanges/gateio/gateio.go
```diff
@@ -1128,10 +1130,22 @@ func (e *Exchange) TransferCurrency(ctx context.Context, arg *TransferCurrencyPa
...
+		return crossMarginAccount
+	case asset.Options:
+		return optionsAccount
+	default:
+		return acc.String()
 	}
-	return acc.String()
+}
+
+func isSpotOrderAccount(account string) bool {
```
thrasher-, line 1153: Heimdall: **Nit** — `isSpotOrderAccount` is case-sensitive where the code it replaced wasn't; +1 to @Copilot's still-open comment from 2026-05-08.

**Fix:** normalise to lowercase before comparing, or call the tightening out deliberately — `TestIsSpotOrderAccount` currently locks in `"SPOT" -> false`.

**Why:** the baseline used `strings.EqualFold` in both `PlaceSpotOrder` and `CreateBatchOrders`. Lowercase-only does match Gate's wire enums, so the new behaviour is arguably more correct — it's just an undeclared break for anyone using this as a library. Both call sites validate `Account` before building a request, so it's assertable offline if you want it pinned. Separately, Gate's current `Order` model documents only `spot`, `margin` and `unified` — `cross_margin` is gone from the enum — but that pre-dates this PR and belongs in its own issue.

exchanges/gateio/margin_isolated_types.go
```diff
@@ -0,0 +1,142 @@
+package gateio
+
+import (
+	"github.com/thrasher-corp/gocryptotrader/currency"
+	"github.com/thrasher-corp/gocryptotrader/types"
+)
+
+// IsolatedMarginAccountBalanceChangeInfo represents margin account balance
+type IsolatedMarginAccountBalanceChangeInfo struct {
```
thrasher-, line 11: Heimdall: **Nit** — small stuff, none of it blocking.

- [margin_isolated_types.go:9](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_types.go#L9) — `IsolatedMarginAccountBalanceChangeInfo` is missing a `Type string` field with the `type` tag, which Gate's `MarginAccountBook` returns.
- [margin_isolated.go:44](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated.go#L44) — docs say "if `currency` is specified, `currency_pair` must also be specified"; not enforced, and the `type` filter isn't exposed.
- [gateio_wrapper.go:175](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/gateio_wrapper.go#L175) — nothing asserts the new `exchange.EdgeCase1: frontEndURL` entry. One `GetURL` assertion after `SetDefaults` covers it without any network.
- Parameter ordering drifts across the new siblings: `(ccy, pair)` at [:151](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated.go#L151), `(loanType, ccy, pair)` at [:199](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated.go#L199), `(pair, ccy)` at [:221](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated.go#L221).
- Bare `require.NoError(t, err)` with no message at [margin_isolated_test.go:120](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_test.go#L120), [:130](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_test.go#L130), [:198](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_test.go#L198), [:211](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_test.go#L211).
- [ratelimiter_test.go:11](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/ratelimiter_test.go#L11) — hard-codes `perpetualUpdateRiskEPL` as the bound, so anything appended after it is silently uncovered. Strictly better than what it replaced; worth a note on the constant so the next person moves it.
- [request.go:137](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/request/request.go#L137) — `headersFromContext` is applied after `Item.Headers`, so a context header can clobber gateio's `KEY`/`SIGN`/`Timestamp`. Opt-in, so no action needed, but worth a line on `WithHeaders`.
- [setUnixTimeRangeParams](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/gateio.go#L193) changes behaviour on ~20 endpoints outside this PR's scope — when both bounds are set, `from == to` now returns `ErrStartEqualsEnd` and a future `from` returns `ErrStartAfterTimeNow`. Good cleanup, worth a line in the description.
- Gate's [migration table](https://www.gate.com/docs/developers/apiv4/en/) deprecates `/margin/currency_pairs` and `/margin/currency_pairs/{pair}` in favour of `/margin/uni/currency_pairs`. We add the uni variants but keep the old methods at [gateio.go:1366](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/gateio.go#L1366) and add a new EPL (`publicCurrencyPairDetailMarginEPL`) for the deprecated detail endpoint.

thrasher- (changes requested):
Heimdall: Good refactor, and it quietly fixes a real bug — `UpdateAccountBalances` was routing `asset.CrossMargin` through the *isolated* endpoint, so cross balances have been wrong all along. The `setUnixTimeRangeParams` dedupe and `TestRateLimits` now covering every EPL are both worth having on their own.

One regression I'd want fixed before merge: the `Set` -> `Add` change (correct in itself) means a single isolated market with an empty currency now discards the entire `asset.Margin` balance update, and I can reproduce the triggering response from Gate's documented schema.

Three more stop new endpoints working at all — `leverage` goes out as a JSON number where Gate wants a string, `GetIsolatedMarginPoolLoans` 403s for every caller that isn't the test, and `GetIsolatedMarginMarketLeverageTiers` authenticates a public endpoint. Happy for those as follow-ups since none of them regress anything that works today.

On tests I'm only interested in what's assertable without hitting the exchange — details in the comment on [margin_isolated_test.go](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated_test.go). It's a short list: one missing sentinel, one untested error path (which is the blocker), and no JSON fixtures for the new response types. `IsolatedMarginBorrowOrRepay` with all seven sentinels asserted is the standard the rest of the file should meet.

**To unblock:**
1. Skip empty-currency entries instead of aborting the balance update — [gateio_wrapper.go:792](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/gateio_wrapper.go#L792)
2. Cover that error path, plus the missing `ErrCurrencyCodeEmpty` sentinel on `GetIsolatedMarginMaxTransferableAmount` — [margin_isolated.go:95](https://github.com/thrasher-corp/gocryptotrader/blob/9cbfcd2faf0ffedb26b795b10d75391c4183526f/exchanges/gateio/margin_isolated.go#L95)

exchanges/gateio/margin_isolated.go
shazbert: This worked with a standard float but it works with a string as well with no issues so will change to that to conform to spec.

exchanges/gateio/margin_isolated.go
shazbert: I don't think for this edge case is good enough to set inside the method itself because if they change their finger printing then it fails for the user anyway. Then if we want to set headers via context externally using this library then the context header setting in the method will always override what will be provided. 
I didn't want to handle the 403 case because I wanted it to error live if they did and that would alert that it needs changing. This is not an officially supported endpoint.

exchanges/gateio/margin_isolated_types.go
shazbert: It doesn't return that:
```
GateIO attempt 1 request path: https://api.gateio.ws/api/v4/margin/uni/loans
GateIO request header [Timestamp]: [1786495277]
GateIO request header [Accept]: [application/json]
GateIO request header [X-Gate-Size-Decimal]: [1]
GateIO request header [Content-Type]: [application/json]
GateIO request type: GET
GateIO request body:
GateIO response header [X-In-Time]: [1786495279487696]
GateIO response header [Date]: [Wed, 12 Aug 2026 00:41:19 GMT]
GateIO response header [Server]: [openresty]
GateIO response header [X-Request-Id]: []
GateIO response header [X-Gate-Ratelimit-Requests-Remain]: [149]
GateIO response header [X-Gate-Trace-Id]: [6a7bc12f112cd9463d6531ff17131aac]
GateIO response header [Content-Type]: [application/json]
GateIO response header [X-Gate-User-Verified]: [3]
GateIO response header [X-Gate-Ratelimit-Limit]: [150]
GateIO response header [X-Gate-Ratelimit-Reset-Timestamp]: [1786495279]
GateIO response header [X-Out-Time]: [1786495279506585]
GateIO raw response: [{"currency":"COTI","currency_pair":"COTI_USDT","amount":"14948.650967","create_time":1786122805306,"update_time":1786446777998}]
```

exchanges/gateio/gateio_wrapper.go
shazbert: The proposed defensive subtraction risks double-counting, Gate could populate it while still deducting it from available. If the change occurs I will change it then. The account system also needs an upgrade to actually populate isolated positions correctly in a meaningful way.

exchanges/gateio/margin_isolated.go
shazbert: Will add commentary:
GateIO unsuccessful HTTP status code: 400 raw response: {"label":"INVALID_PARAM_VALUE","message":"currency pair missing for USDT request"}

exchanges/gateio/margin_isolated_types.go
shazbert: IsolatedMarginAccountBalanceChangeInfo  - 👍 

> docs say "if currency is specified, currency_pair must also be specified

untrue, if a single currency is supplied then then it returns its isolated margin position balance change history and if USDT is passed in it returns all associated data. Added account type filter. 👍 

> nothing asserts the new exchange.EdgeCase1: frontEndURL entry. One GetURL assertion after SetDefaults covers it without any network.

It assets it in the test, or what do you mean?

> Parameter ordering drifts across the new siblings

Done 👍 

> Bare require.NoError(t, err) with no message at [margin_isolated_test.go:120]

Done 👍 

>  hard-codes perpetualUpdateRiskEPL as the bound

commentary added

> headersFromContext is applied after Item.Headers

commentary added

> [setUnixTimeRangeParams]changes behaviour on ~20 endpoints outside this PR's scope

commentary added

> Gate's [migration table] deprecates /margin/currency_pairs and /margin/currency_pairs/{pair}

Removed endpoints 👍 

Will push these changes soon.

exchanges/gateio/gateio_wrapper.go
```diff
@@ -433,16 +434,13 @@ func (e *Exchange) FetchTradablePairs(ctx context.Context, a asset.Item) (curren
...
 		if err != nil {
 			return nil, err
 		}
 		pairs := make([]currency.Pair, 0, len(tradables))
 		for x := range tradables {
-			if tradables[x].Status == 0 || tradables[x].BaseMinimumBorrowAmount.Float64() == 0 { // Pairs with min_base_amount == 0 are effectively dead and skipped.
-				continue
-			}
-			pairs = append(pairs, tradables[x].ID)
+			pairs = append(pairs, tradables[x].Pair)
```
thrasher-, line 447: Heimdall: **Blocker** — dropping the pair filter exposes 524 disabled isolated-margin markets as tradable, and it fails `TestUpdateOrderExecutionLimits` on today's data.

**Fix:** add a `Status` field (tag `status`), and optionally `DelistedTime` (tag `delisted_time`), to `IsolatedMarginLendingMarket` at [margin_isolated_types.go:41](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated_types.go#L41), then filter `status != "enabled"` at [gateio_wrapper.go:443](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L443) and [gateio_wrapper.go:2041](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L2041).

**Why:** the baseline filtered `Status == 0 || BaseMinimumBorrowAmount == 0`. `margin/uni/currency_pairs` returns both `status` and `delisted_time` and we model neither. Live, 17 Aug:

```
Uni total:               1275
Uni enabled:              751
Uni disabled:             524
Legacy after old filter:  751
enabled/legacy set diff:    0     <- status=="enabled" is an exact drop-in
```

This was latent when CI last ran green here (`MARGIN Total: 753` on 12 Aug); Gate has since populated the disabled rows. Today:

```
branch:  Updating available pairs [MARGIN] - Total: 1275.
         364 unsupported margin pairs found, no execution limits loaded for: map[TON_USDT:0x... ]
         order limit not found for "GateIO" "margin" "1DOLLARUSDT"
         --- FAIL: TestUpdateOrderExecutionLimits/margin
master:  Updating available pairs [MARGIN] - Total: 751.
         --- PASS: TestUpdateOrderExecutionLimits
```

The quieter half is worse than the failure: of the 524 disabled, 160 also exist in spot, so they receive full margin execution limits and present as healthy tradable markets. Only the 364 absent from spot reach the warning — which prints pointer addresses, since `unsupported` is a `map[currency.Pair]*IsolatedMarginLendingMarket` formatted with `%v` at [gateio_wrapper.go:2074](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L2074). Worth switching that to values while you're in there.

exchanges/gateio/margin_isolated.go
```diff
@@ -0,0 +1,358 @@
...
+
+// GetIsolatedMarginLendingMarkets retrieves isolated margin lending markets
+func (e *Exchange) GetIsolatedMarginLendingMarkets(ctx context.Context) ([]IsolatedMarginLendingMarket, error) {
+	var lendingMarkets []IsolatedMarginLendingMarket
+	return lendingMarkets, e.SendHTTPRequest(ctx, exchange.RestSpot, publicUniCurrencyPairsMarginEPL, "margin/uni/currency_pairs", &lendingMarkets)
+}
+
+// GetIsolatedMarginLendingMarketDetails retrieves isolated margin lending market detail given the currency pair.
+func (e *Exchange) GetIsolatedMarginLendingMarketDetails(ctx context.Context, pair currency.Pair) (*IsolatedMarginLendingMarket, error) {
+	if pair.IsEmpty() {
```
thrasher-, line 128: Heimdall: **Nit** — `IsEmpty()` isn't the complement of `IsPopulated()`, so a required-pair guard passes a half-populated pair straight through.

Nothing in GCT hits this — none of the seven methods has a caller outside its own GoDoc, and none is on `IBotExchange`, so it's library surface only. Raising it because the input is easy to produce by accident and the fix is one word per site.

**Fix:** guard on `!pair.IsPopulated()` where a pair is required — [margin_isolated.go:125](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L125), [margin_isolated.go:186](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L186), [margin_isolated.go:263](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L263), [margin_isolated.go:275](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L275), [margin_isolated.go:286](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L286), [margin_isolated.go:300](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L300) — and at [margin_isolated.go:106](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L106) for the USDT rule.

**Why:** `Pair.IsEmpty` at [pair_methods.go:134](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/currency/pair_methods.go#L134) needs both sides empty; `Pair.IsPopulated` at [pair_methods.go:161](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/currency/pair_methods.go#L161) needs both set. A half pair is neither, and the constructors hand one back without complaint:

```
NewPairFromString("BTC")     -> base="BTC" quote="" err=<nil>
NewPairFromString("BTC_")    -> base="BTC" quote="" err=<nil>
NewPairFromStrings("BTC","") -> base="BTC" quote="" err=<nil>
```

A consumer then gets a silently malformed request instead of an error:

```
GET  /api/v4/margin/uni/currency_pairs/BTC
GET  /api/v4/margin/uni/borrowable?currency=BTC&currency_pair=BTC
POST /api/v4/margin/leverage/user_market_setting  {"leverage":"2","currency_pair":"BTC"}
```

[margin_isolated.go:106](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L106) fails the other way — a half pair with `ccy == USDT` clears the USDT guard and then never sets `currency_pair`, which is the exact request that guard exists to stop.

thrasher- (changes requested):
Heimdall: Strong turnaround since the last pass — everything from it is in bar two. `SetUserMarketLeverageMultiplier` sends a string now, `GetIsolatedMarginMarketLeverageTiers` is public, the three dead EPLs are gone, `QueryInterestDeductionRecords` is `uint64` with a validated `loanType`, `isSpotOrderAccount` folds case, `Loan` picks up the missing fiat field, and three of the four offline tests are in — the `setUnixTimeRangeParams` table went further than I asked for.

One correction I owe you before the rest: I said `/margin/uni/interest_records` had no documented maximum. Wrong — the Go SDK markdown just omits the number. The spec value is 1000 and the Python SDK enforces it, so that cap wants raising to 1000, not removing. Details in the thread.

**The new blocker isn't your code changing — it's Gate's data.** `FetchTradablePairs(asset.Margin)` dropped the dead-pair filter when it moved to `margin/uni/currency_pairs`, and that endpoint carries a `status` field we don't model. When CI last ran green here the endpoint returned 753 rows; today it returns 1275, of which 524 are `status: "disabled"`. `TestUpdateOrderExecutionLimits/margin` fails on the branch and passes on master against that same live data. The quieter half is worse than the failure: 160 of those disabled markets also exist in spot, so they load full execution limits and present as healthy tradable pairs.

**The 403 has stopped being hypothetical.** All seven test jobs on the latest run are red and `TestGetIsolatedMarginPoolLoans` is the only failure on every one. Your call to let it error live rather than paper over it was right — it has errored live. My own header matrix from 11 Aug no longer reproduces either, which is the argument for not gating CI on an Akamai score.

**And the balance blocker is still half-open** — the `inactive` skip only catches rows where both sides are empty.

**To unblock:**
1. Filter `status != "enabled"` in both margin branches — [gateio_wrapper.go:443](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L443)
2. Stop `TestGetIsolatedMarginPoolLoans` gating CI on a live Akamai decision — [margin_isolated_test.go:283](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated_test.go#L283)
3. Skip empty-currency entries per side rather than aborting the whole update — [gateio_wrapper.go:791](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L791)

Everything else is follow-up: the limit cap at 1000, the last missing sentinel test, the `XRequest` struct, and a pair-guard nit that only reaches library consumers.

exchanges/gateio/gateio_wrapper.go
thrasher-: Heimdall: **Still blocking** — the guard added in `heimdal: nits` only covers one of three shapes.

**Fix:** move the check per-side into `addIsolatedMarginAccountBalance`, rather than per-row.

**Why:** [gateio_wrapper.go:791](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L791) skips only when `AccountType == "inactive"` **and** both sides are empty. Two shapes still abort:

```
non-inactive row, empty currency:  err=currency code is empty, 2 accumulated balances discarded
inactive row, one side populated:  err=currency code is empty, 1 accumulated balance discarded
```

`UpdateAccountBalances` returns `nil, err` at [gateio_wrapper.go:745](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L745) and never reaches `Accounts.Save` at [gateio_wrapper.go:786](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L786), so the partial map goes with it. Both shapes decode cleanly from Gate's documented schema — `base`/`quote` and their `currency` are all `[optional]`.

exchanges/gateio/margin_isolated.go
thrasher-: Heimdall: **Correction to my own comment, and still open.** I had the maximum wrong.

`heimdal: nits` added the cap to the three endpoints documenting `maximum: 100`, which is right. The cap on `/margin/uni/interest_records` is still there — but "no documented maximum" was wrong of me. The Go SDK markdown just omits the number; the spec value is **1000**, which the Python SDK enforces:

```
list_uni_loan_interest_records       limit <= 1000    gateapi-python/gate_api/api/margin_uni_api.py:838
list_uni_loans                       limit <= 100     gateapi-python/gate_api/api/margin_uni_api.py:458
list_uni_loan_records                limit <= 100     gateapi-python/gate_api/api/margin_uni_api.py:702
list_unified_loan_interest_records   limit <= 100     gateapi-python/gate_api/api/unified_api.py:1074
```

**Fix:** raise [margin_isolated.go:247](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L247) from 100 to 1000 rather than removing it — as shipped we reject the 101–1000 range the API accepts. Disregard my earlier "drop it".

exchanges/gateio/margin_isolated_test.go
thrasher-: Heimdall: **3 of 4 landed** — one left.

Done: the `setIsolatedMarginAccountBalances` empty-currency case, `setCrossMarginAccountBalances(nil)`, and the `setUnixTimeRangeParams` table — eight cases including the `time.Unix(0,0)` and `ErrStartAfterTimeNow` edges, which is more than I asked for.

Still missing is (1). [margin_isolated_test.go:74](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated_test.go#L74) only ever passes `currency.USDT`, so [margin_isolated.go:102](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L102) has still never executed:

```
margin_isolated.go:102.19,104.3  count=0   # empty currency branch
margin_isolated.go:106.56,108.3  count=1   # empty pair branch
```

One line, no network.

exchanges/gateio/margin_isolated.go
thrasher-: Heimdall: **Reopening** — the live alert you wanted has fired, and it's blocking merge.

All seven test jobs on the latest run are red, and this is the only failure on every one:

```
--- FAIL: TestGetIsolatedMarginPoolLoans (0.13s)
    margin_isolated_test.go:283
    Error: "...403 raw response: <HTML>...Access Denied..." does not contain "Type mismatch"
```

I'll drop the "set the headers inside the method" half — your reasoning holds. But the matrix has already shifted since I measured it, which is the part that matters:

```
                       11 Aug (mine)   17 Aug (rerun)   CI runners
UA + Accept-Encoding        200             -               -
UA + Accept                 403            200             403
UA only                     403            403              -
```

Same code, same headers, different answer by host and by week. That's the Akamai score moving, exactly as we both read it — so the call is fine opportunistically, it just can't be a hard CI dependency.

**Fix:** cover the decode and `Code != 200` paths with an `httptest` fixture, and either drop the live call or skip on 403/504. The `require.ErrorContains(t, err, "Type mismatch")` at [margin_isolated_test.go:283](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated_test.go#L283) is pinning Gate's validator string for a max-uint64 page, which isn't ours to hold stable either.

exchanges/gateio/margin_isolated_types.go
thrasher-: Heimdall: > It assets it in the test, or what do you mean?

It doesn't — that's the point. `git grep EdgeCase1 -- exchanges/gateio/` returns exactly two hits, both production code: [gateio_wrapper.go:178](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L178) registers it, [margin_isolated.go:349](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L349) consumes it. Nothing in `_test.go` touches it.

So the only coverage is indirect, via `TestGetIsolatedMarginPoolLoans` — the live-network test currently failing CI. A wrong or missing URL surfaces there as a network error rather than a config assertion. One `GetURL(exchange.EdgeCase1)` check after `SetDefaults` pins it offline.

Everything else in this batch is confirmed done — thanks.

exchanges/gateio/margin_isolated.go
thrasher-: Heimdall: `types.Number` half is in, thanks. Second half still open — [margin_isolated.go:306](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L306) is an anonymous struct passed **by value**. Guidelines want a named `XRequest` passed by pointer, and `IsolatedBorrowRepayRequest` at [margin_isolated_types.go:56](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated_types.go#L56) is the pattern. Non-blocking.

exchanges/gateio/margin_isolated.go
shazbert: > Fix: guard on !pair.IsPopulated() where a pair is required — [margin_isolated.go:125](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L125), [margin_isolated.go:186](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L186), [margin_isolated.go:263](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L263), [margin_isolated.go:275](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L275), [margin_isolated.go:286](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L286), [margin_isolated.go:300](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L300) — and at [margin_isolated.go:106](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/margin_isolated.go#L106) for the USDT rule.

These all 400 from the endpoint. But I will make a change for GetIsolatedMarginMaxTransferableAmount

thrasher-:
Heimdall: Re-checked at `36f98fd`, already up to date with master (`ae29d488e`). The pair filter landed and `TestUpdateOrderExecutionLimits` passes again — margin available pairs are back to 751 from 1275. Two things came back with it though, and one of them is red in CI.

**A scratch comment got committed and it breaks lint.** [margin_isolated_types.go:40](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/margin_isolated_types.go#L40) carries `// "status":"enabled","delisted_time":0}` directly above the type's doc comment. The `lint` job is red where it was green last round, and it reproduces locally as the only issue in the package:

```
margin_isolated_types.go:40:1: exported: comment on exported type IsolatedMarginLendingMarket
should be of the form "IsolatedMarginLendingMarket ..." (revive)
```

**The filter went into `FetchTradablePairs` but not `UpdateOrderExecutionLimits`.** [gateio_wrapper.go:2042](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/gateio_wrapper.go#L2042) still builds `supported` from the unfiltered `marginPairs`, so the two halves now disagree about which markets exist. Live today:

```
uni total 1275   enabled 751   filtered-out 524
filtered-out but present in spot -> still get margin limits: 160
filtered-out and absent from spot -> still warn:             364
MinMaxLevel entries built: 911  (751 enabled + 160 disabled)
```

So the 364-pair warning still fires on every call — it's sorted and legible now, which is a real improvement, but it's still listing markets we deliberately excluded — and 160 disabled markets carry full execution limits that `FetchTradablePairs` says aren't tradable.

**One question on the filter itself.** Excluding on any non-zero `DelistedTime` is stricter than how this same file treats spot, which excludes on `TradeStatus == "untradable"` at [gateio_wrapper.go:2012](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/gateio_wrapper.go#L2012) and passes `DelistingTime` through as `Delisted` metadata at [gateio_wrapper.go:2029](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/gateio_wrapper.go#L2029). A market with a *scheduled future* delisting is dropped while it's still tradable, and `Status` presumably flips to `disabled` at the actual delist anyway. No live pair currently carries a non-zero `delisted_time` and the fixture only covers a past one (`1700000000`), so the future case is untested either way — happy to be told Gate sets it only on completed delistings.

**Still open from the last pass:** the balance-update abort at [gateio_wrapper.go:792](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/gateio_wrapper.go#L792) is unchanged, and `TestGetIsolatedMarginPoolLoans` is still 403ing in CI at [margin_isolated_test.go:292](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/margin_isolated_test.go#L292). Follow-ups also unchanged: the 100→1000 cap, the anonymous leverage struct, and the `ErrCurrencyCodeEmpty` branch — the new sub-cases in `TestGetIsolatedMarginMaxTransferableAmount` assert `ErrCurrencyPairEmpty`, so [margin_isolated.go:102](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/margin_isolated.go#L102) still hasn't executed.

Two nits while you're in there: the assertion message at [gateio_wrapper_test.go:194](https://github.com/thrasher-corp/gocryptotrader/blob/36f98fd99db4d73746f7ce5dcbf6e51493c3063b/exchanges/gateio/gateio_wrapper_test.go#L194) still says "must return all pairs", which is now the opposite of what it tests; and `sort.Slice` pulls in a fresh `sort` import when `currency.Pairs(...).Strings()` plus `slices.Sort` would do it with the imports already there.

**To unblock:** the stray comment, the filter at the second call site, and the two carried-over blockers. Everything else is follow-up.

shazbert:
regarding [gateio_wrapper.go:791](https://github.com/thrasher-corp/gocryptotrader/blob/4509694997a96637b3696841b9ef4d05ec15c5d0/exchanges/gateio/gateio_wrapper.go#L791)

None of the actual data returned drops the currency, even when offline. I do want it to return an error if things change instead of silently processing when trading.

thrasher-:
Heimdall: Re-checked at `98b4680`. Nine of ten landed, lint is green, and the pool-loan test is passing on every runner — the only red left across the seven jobs is `TestGetOpenInterestAndVolumeStrike` in okx, which isn't yours. **No blockers from me.**

On the header block: I argued that was the wrong lever and that the CI dependency should go. You went the other way and it works — got past Akamai on all seven. The fragility point stands for the future, but my practical objection didn't survive contact.

On [margin_isolated.go:128](https://github.com/thrasher-corp/gocryptotrader/blob/98b4680bee502c96cffe3bf628051c9f1bbb2e22/exchanges/gateio/margin_isolated.go#L128) — conceded, and your split is better than my comment. I lumped all seven together; six *send* a malformed pair and get a clean rejection, which is fine:

```
GET /api/v4/margin/uni/currency_pairs/BTC      -> 400 INVALID_CURRENCY_PAIR
GET /api/v4/margin/loan_margin_tiers?...=BTC   -> 400 INVALID_CURRENCY_PAIR
```

Only `GetIsolatedMarginMaxTransferableAmount` *silently dropped* the constraint and sent a valid-looking request with no `currency_pair` at all. That was the one worth fixing and it's the one you fixed.

On the balance abort — taking your call. Failing loud beats partial balances understating borrowed exposure mid-trade, and I can't check the data myself without credentials.

One loose thread it leaves, though. If the response never drops the currency, then the guard at [gateio_wrapper.go:794](https://github.com/thrasher-corp/gocryptotrader/blob/98b4680bee502c96cffe3bf628051c9f1bbb2e22/exchanges/gateio/gateio_wrapper.go#L794) can never fire — `AccountType == "inactive" && Base.Currency.IsEmpty() && Quote.Currency.IsEmpty()` only matches rows that don't exist. Either it's dead and should go, or the case does occur and the abort is reachable. Both are fine; the code and the observation just currently disagree.

**One thing I should have raised earlier and didn't.** [setUnixTimeRangeParams](https://github.com/thrasher-corp/gocryptotrader/blob/98b4680bee502c96cffe3bf628051c9f1bbb2e22/exchanges/gateio/gateio.go#L197) now rejects `from == to` with `ErrStartEqualsEnd` across ~18 endpoints, and Gate serves that request:

```
GET /api/v4/spot/candlesticks?currency_pair=BTC_USDT&interval=1h&from=<now>&to=<now>
-> HTTP 200  [["1786928400","12898296.39...","63166.6","63241.2","62897.8","62914.6",...]]
```

Equal bounds are a legitimate way to ask for a single candle. `GetCandlesticks` and its siblings now refuse it, and [gateio_test.go:107](https://github.com/thrasher-corp/gocryptotrader/blob/98b4680bee502c96cffe3bf628051c9f1bbb2e22/exchanges/gateio/gateio_test.go#L107) locks the behaviour in. Not a blocker, but it's a live regression on endpoints outside this PR's remit.

Last nit: the filter predicate is now duplicated verbatim at [gateio_wrapper.go:443](https://github.com/thrasher-corp/gocryptotrader/blob/98b4680bee502c96cffe3bf628051c9f1bbb2e22/exchanges/gateio/gateio_wrapper.go#L443) and [gateio_wrapper.go:2044](https://github.com/thrasher-corp/gocryptotrader/blob/98b4680bee502c96cffe3bf628051c9f1bbb2e22/exchanges/gateio/gateio_wrapper.go#L2044). The bug this round was exactly "filter at one site, not the other" — an `IsTradable()` predicate on the type makes that unrepeatable. Same line: the mixed `||`/`&&` reads on precedence alone, and `time.Now()` is evaluated twice per iteration when one hoisted value would do.

Happy to approve once you've decided on the `:794` guard either way.

thrasher-:
Heimdall: Re-checked at `428d739`. The predicate extraction is exactly right — [margin_isolated_types.go:53](https://github.com/thrasher-corp/gocryptotrader/blob/428d739cfe24068cbd793bb2a6ebdd82f20e6ea9/exchanges/gateio/margin_isolated_types.go#L53) collapses all three of my nits into one change, and `IsTradable` is now the only thing either call site knows about, so the two can't drift apart again. `now` hoisted out of both loops, precedence made explicit, and the table test covers the `delisting now` boundary — more than I asked for.

I checked the refactor is behaviour-preserving: the old skip condition `Status != "enabled" || (!IsZero() && Before(now))` negates exactly to the new keep condition, and both sites (`:444`, `:2046`) use it with no third site missed.

**Lint is red again, though.** Same shape as last round — one issue, one line:

```
exchanges/gateio/margin_isolated_types.go:53:7: hugeParam: m is heavy (112 bytes);
consider passing it by pointer (gocritic)
```

A pointer receiver is a clean drop-in — both call sites are indexed slice elements and therefore addressable, so nothing else changes. I confirmed it locally:

```
func (m *IsolatedMarginLendingMarket) IsTradable(...)
-> golangci-lint: 0 issues, tests pass
```

That's two pushes in a row where lint caught something CI-only. `make lint` before pushing would turn both into a ten-second loop instead of a round trip.

**Still waiting on the `:794` question** — it's been through two pushes without an answer either way. If the response never drops the currency, the guard at [gateio_wrapper.go:795](https://github.com/thrasher-corp/gocryptotrader/blob/428d739cfe24068cbd793bb2a6ebdd82f20e6ea9/exchanges/gateio/gateio_wrapper.go#L795) can't fire and should go; if it can fire, the abort below it is reachable and your fail-loud reasoning applies to a case that actually happens. Either answer closes it — I just can't tell which from the code.

Unchanged and fine as follow-ups: the equal-bounds rejection at [gateio.go:197](https://github.com/thrasher-corp/gocryptotrader/blob/428d739cfe24068cbd793bb2a6ebdd82f20e6ea9/exchanges/gateio/gateio.go#L197), and the anonymous leverage struct at [margin_isolated.go:309](https://github.com/thrasher-corp/gocryptotrader/blob/428d739cfe24068cbd793bb2a6ebdd82f20e6ea9/exchanges/gateio/margin_isolated.go#L309).

Everything else is green — `exchanges/gateio` and `exchanges/request` both pass locally with `-race`, and the five red test jobs are all the unrelated okx failure.

shazbert:
The inline struct is fine to keep if there are only a limited amount of params

thrasher- (approved):
Heimdall: Approving. Everything raised across five rounds is resolved — the `hugeParam` receiver, the `:794` guard removed with a test pinning the fail-loud behaviour, and the equal-bounds regression fixed.

The equal-bounds fix is the careful one. `StartEndTimeCheck` returns `ErrStartEqualsEnd` *before* it reaches its own future-start check, so simply tolerating that error would have silently dropped the future guard — restoring it explicitly is right. Verified against the live API rather than just the table:

```
equal bounds  -> err=<nil>  candles=1
equal+future  -> err=start date is after current time
```

Lint is green and `exchanges/gateio` and `exchanges/request` both pass with `-race`. The two red jobs are `TestGetSupportedFlashSwapCurrencies` hitting a 429 from Gate on the shared CI IP — ambient, not yours.

One optional nit, non-blocking: the doc comment at [gateio.go:193](https://github.com/thrasher-corp/gocryptotrader/blob/9716fb2ca6f261cf5ff66f4472fea3a32b996852/exchanges/gateio/gateio.go#L193) got *less* specific in this commit just as the logic got subtler. A line noting that equal bounds are permitted except in the future would save the next reader unpicking the double negative.

Issue #1925, opened 2025-05-27
LBank: Update endpoints to v2 and other improvements


thrasher-:
The following improvements can be made:

- Most endpoints need updating to v2
- A generic error capture mechanism can be implemented in the SendX funcs like how we do it in other exchanges
- String concatenation of paths instead of `fmt.Sprintf`
- Websocket support

PARMESHWARPANWAR:
Hi @thrasher- , I'd like to work on this issue. I'm planning to start with updating the fmt.Sprintf path concatenations and the v2 endpoint updates. Any specific endpoints you'd recommend starting with?

PARMESHWARPANWAR:
Hi @thrasher- , I've opened a PR to address this issue. I've replaced all fmt.Sprintf path constructions with string concatenation following the same pattern used in bitstamp and huobi. Also fixed two incorrect endpoint constants found during the process. All tests pass.

PR #2256, opened 2026-05-19
exchanges/lbank: replace fmt.Sprintf with string concatenation for pa…

PARMESHWARPANWAR:
# PR Description

Replace `fmt.Sprintf` usage for building API endpoint paths with simple
string concatenation, consistent with other exchanges in the codebase
such as bitstamp and huobi.

#1925 (partially - string concatenation task complete)

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] go test ./exchanges/lbank/...

### DIFF
--- a/exchanges/lbank/lbank.go
+++ b/exchanges/lbank/lbank.go
@@ -170,7 +170,7 @@ func (e *Exchange) GetKlines(ctx context.Context, symbol, size, klineType string
 // GetUserInfo gets users account info
 func (e *Exchange) GetUserInfo(ctx context.Context) (InfoFinalResponse, error) {
 	var resp InfoFinalResponse
-	path := "/v" + lbankAPIVersion1 + "/" + lbankUserInfo + "?"
+	path := "/v" + lbankAPIVersion1 + "/" + lbankUserInfo
 	err := e.SendAuthHTTPRequest(ctx, http.MethodPost, path, nil, &resp)
 	if err != nil {
 		return resp, err
@@ -201,7 +201,7 @@ func (e *Exchange) CreateOrder(ctx context.Context, pair, side string, amount, p
 	params.Set("type", strings.ToLower(side))
 	params.Set("price", strconv.FormatFloat(price, 'f', -1, 64))
 	params.Set("amount", strconv.FormatFloat(amount, 'f', -1, 64))
-	path := "/v" + lbankAPIVersion1 + "/" + lbankPlaceOrder + "?"
+	path := "/v" + lbankAPIVersion1 + "/" + lbankPlaceOrder
 	err := e.SendAuthHTTPRequest(ctx, http.MethodPost, path, params, &resp)
 	if err != nil {
 		return resp, err
@@ -241,7 +241,7 @@ func (e *Exchange) QueryOrder(ctx context.Context, pair, orderIDs string) (Query
 	params := url.Values{}
 	params.Set("symbol", pair)
 	params.Set("order_id", orderIDs)
-	path := "/v" + lbankAPIVersion1 + "/" + lbankQueryOrder + "?"
+	path := "/v" + lbankAPIVersion1 + "/" + lbankQueryOrder
 	err := e.SendAuthHTTPRequest(ctx, http.MethodPost, path, params, &tempResp)
 	if err != nil {
 		return resp, err
@@ -277,7 +277,7 @@ func (e *Exchange) QueryOrderHistory(ctx context.Context, pair, pageNumber, page
 	params.Set("symbol", pair)
 	params.Set("current_page", pageNumber)
 	params.Set("page_length", pageLength)
-	path := "/v" + lbankAPIVersion1 + "/" + lbankQueryHistoryOrder + "?"
+	path := "/v" + lbankAPIVersion1 + "/" + lbankQueryHistoryOrder
 	err := e.SendAuthHTTPRequest(ctx, http.MethodPost, path, params, &tempResp)
 	if err != nil {
 		return resp, err
@@ -305,7 +305,7 @@ func (e *Exchange) QueryOrderHistory(ctx context.Context, pair, pageNumber, page
 // GetPairInfo finds information about all trading pairs
 func (e *Exchange) GetPairInfo(ctx context.Context) ([]PairInfoResponse, error) {
 	var resp []PairInfoResponse
-	path := "/v" + lbankAPIVersion1 + "/" + lbankPairInfo + "?"
+	path := "/v" + lbankAPIVersion1 + "/" + lbankPairInfo
 	return resp, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &resp)
 }
 
@@ -315,7 +315,7 @@ func (e *Exchange) OrderTransactionDetails(ctx context.Context, symbol, orderID 
 	params := url.Values{}
 	params.Set("symbol", symbol)
 	params.Set("order_id", orderID)
-	path := "/v" + lbankAPIVersion1 + "/" + lbankOrderTransactionDetails + "?"
+	path := "/v" + lbankAPIVersion1 + "/" + lbankOrderTransactionDetails
 	err := e.SendAuthHTTPRequest(ctx, http.MethodPost, path, params, &resp)
 	if err != nil {
 		return resp, err
@@ -339,7 +339,7 @@ func (e *Exchange) TransactionHistory(ctx context.Context, symbol, transactionTy
 	params.Set("from", from)
 	params.Set("direct", direct)
 	params.Set("size", size)
-	path := "/v" + lbankAPIVersion1 + "/" + lbankPastTransactions + "?"
+	path := "/v" + lbankAPIVersion1 + "/" + lbankPastTransactions
 	err := e.SendAuthHTTPRequest(ctx, http.MethodPost, path, params, &resp)
 	if err != nil {
 		return resp, err
--- a/exchanges/lbank/lbank.go
+++ b/exchanges/lbank/lbank.go
@@ -79,14 +79,14 @@ func (e *Exchange) GetTicker(ctx context.Context, symbol string) (*TickerRespons
 	var t *TickerResponse
 	params := url.Values{}
 	params.Set("symbol", symbol)
-	path := fmt.Sprintf("/v%s/%s?%s", lbankAPIVersion1, lbankTicker, params.Encode())
+	path := "/v" + lbankAPIVersion1 + "/" + lbankTicker + "?" + params.Encode()
 	return t, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &t)
 }
 
 // GetTimestamp returns a timestamp
 func (e *Exchange) GetTimestamp(ctx context.Context) (time.Time, error) {
 	var resp TimestampResponse
-	path := fmt.Sprintf("/v%s/%s", lbankAPIVersion2, lbankTimestamp)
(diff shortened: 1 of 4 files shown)


exchanges/lbank/lbank.go
```diff
@@ -79,14 +79,14 @@ func (e *Exchange) GetTicker(ctx context.Context, symbol string) (*TickerRespons
 	var t *TickerResponse
 	params := url.Values{}
 	params.Set("symbol", symbol)
-	path := fmt.Sprintf("/v%s/%s?%s", lbankAPIVersion1, lbankTicker, params.Encode())
+	path := "/v" + lbankAPIVersion1 + "/" + lbankTicker + "?" + params.Encode()
```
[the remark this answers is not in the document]
PARMESHWARPANWAR: Thanks for the review! I followed the pattern used in bitstamp and huobi exchanges which use constants rather than inlining. Happy to change if the maintainer prefers the inlined approach.

gloriousCode:
Hello, this does not seem to fix https://github.com/thrasher-corp/gocryptotrader/issues/1925 like suggested. There are no fixes for:
- Most endpoints need updating to v2
- A generic error capture mechanism can be implemented in the SendX funcs like how we do it in other exchanges
- Websocket support

This also does not appear to be present in changes, but is present in PR description:

>Also fix RemoveOrder using incorrect lbankPlaceOrder constant instead
of lbankCancelOrder, and QueryOrder using incorrect lbankPlaceOrder
instead of lbankQueryOrder.

Are these things you intend to still commit?

Thank you for opening a pull request!

PARMESHWARPANWAR:
Hi @gloriousCode, I've updated the PR description to accurately reflect only the changes made — fmt.Sprintf to string concatenation. The RemoveOrder/QueryOrder references in the previous description were incorrect and have been removed .
I'd like to continue working on the remaining tasks from #1925 — v2 endpoint updates and generic error capture in SendX funcs. Could you point me to another exchange that implements the error capture pattern so I can match the style?
For WebSocket support, I was thinking of raising a separate PR as it will likely involve larger changes — new types, event handling, subscription management — and keeping it separate would make both PRs easier to review. That said, if you'd prefer it in this PR, I'm happy to include it here. Just let me know!

PARMESHWARPANWAR:
@gloriousCode 

Additional changes in the 33e04e1 commit:

Updated all public endpoints to use LBank API v2
Added generic error capture in SendHTTPRequest by capturing raw response and unwrapping v2 envelope {"result","msg","data","error_code"}
Fixed MarketDepthResponse type to match v2 flat structure
Fixed GetTimestamp to handle v2 plain int64 millisecond response
Fixed GetTicker to handle v2 array response
Fixed USD2RMBRate to handle v2 plain string response

PARMESHWARPANWAR:
@gloriousCode 

Public endpoints have been updated to v2 with generic error capture in SendHTTPRequest. Authenticated endpoints (user_info, create_order, cancel_order, etc.) are still on v1 — I'd like to update these too but need to verify the v2 auth response format matches the same envelope {"result","msg","data","error_code"}. Could you confirm if v2 auth endpoints use the same response wrapper? I can then apply the same pattern to SendAuthHTTPRequest and update all remaining endpoints.

PARMESHWARPANWAR:
@gloriousCode    0d789a0

Verified all v2 auth endpoints return 10003 (auth error, not 404) — v2 exists for all. Updated all authenticated endpoints to v2 and removed the now-unused lbankAPIVersion1 constant. All local tests pass.

gloriousCode:
Thanks for your replies and further updates! Just let us know when you'd like us to review it. Totally fine to have the websocket be its own PR

PARMESHWARPANWAR:
> Thanks for your replies and further updates! Just let us know when you'd like us to review it. Totally fine to have the websocket be its own PR

Thanks @gloriousCode! I believe the current changes are ready for review. Will open a separate PR for WebSocket support once this is merged.

exchanges/lbank/lbank.go
```diff
@@ -76,37 +76,41 @@ var (
 // GetTicker returns a ticker for the specified symbol
 // symbol: eth_btc
 func (e *Exchange) GetTicker(ctx context.Context, symbol string) (*TickerResponse, error) {
-	var t *TickerResponse
+	var t []TickerResponse // change to slice
```
cranktakular: I think this comment should be removed.

exchanges/lbank/lbank.go
```diff
@@ -76,37 +76,41 @@ var (
...
-	path := fmt.Sprintf("/v%s/%s?%s", lbankAPIVersion1, lbankTicker, params.Encode())
-	return t, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &t)
+	path := "/v" + lbankAPIVersion2 + "/" + lbankTicker + "?" + params.Encode()
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &t); err != nil {
+		return nil, err
+	}
+	if len(t) == 0 {
+		return nil, errors.New("no ticker data returned")
+	}
+	return &t[0], nil // return first element
```
cranktakular: Comments should not be used for code that is self-explanatory, such as is the case here.

exchanges/lbank/lbank.go
```diff
@@ -76,37 +76,41 @@ var (
...
+	var t []TickerResponse // change to slice
 	params := url.Values{}
 	params.Set("symbol", symbol)
-	path := fmt.Sprintf("/v%s/%s?%s", lbankAPIVersion1, lbankTicker, params.Encode())
-	return t, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &t)
+	path := "/v" + lbankAPIVersion2 + "/" + lbankTicker + "?" + params.Encode()
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &t); err != nil {
+		return nil, err
+	}
+	if len(t) == 0 {
```
cranktakular: Optional: Consider erroring if len(t) != 1

You only return the first element anyway, so if the exchange returns more data, either that response is malformed, or this endpoint has been updated to return multiple ticker responses, meriting further investigation and integration.

exchanges/lbank/lbank.go
```diff
@@ -76,37 +76,41 @@ var (
...
-	path := fmt.Sprintf("/v%s/%s", lbankAPIVersion2, lbankTimestamp)
-	err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &resp)
-	if err != nil {
+	var ms int64
+	path := "/v" + lbankAPIVersion2 + "/" + lbankTimestamp
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &ms); err != nil {
 		return time.Time{}, err
 	}
-	return resp.Timestamp.Time(), nil
+	return time.UnixMilli(ms).UTC(), nil
```
cranktakular: This function [should have continued using](https://github.com/thrasher-corp/gocryptotrader/blob/master/docs/CODING_GUIDELINES.md#type-usage) types.Time, rather than swapping to int64 as you did.

However, you don't need to fully revert to having a separate types in lbank_types.go just for this. You can just make the var ms into types.Time

exchanges/lbank/lbank.go
```diff
@@ -490,19 +493,34 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path st
...
+	err = e.SendPayload(ctx, request.Unset, func() (*request.Item, error) {
 		return item, nil
 	}, request.UnauthenticatedRequest)
+	if err != nil {
+		return err
+	}
+
+	var v2Resp V2Response
+	if err := json.Unmarshal(tempResp, &v2Resp); err == nil && v2Resp.Data != nil {
+		if v2Resp.Result == "false" {
```
cranktakular, line 517: This handling is improper.

The exchange can return errors while leaving the v2Resp.Data field empty, i.e. `{"result":"false","msg":"currency pair nonsupport","error_code":10008,"ts":1780974433481}`

I believe the first check should just be for err == nil, then this v2Resp.Result == "false" error handling should occur, and only then should an if v2Resp.Data != nil check protect the unmarshalling of v2Resp.Data

cranktakular (changes requested):
Thanks for the contribution! But I have some changes I must request.

cranktakular:
Also, rather than doing `"/other/" + pathstuff + "?" + params.Encode()`, it's probably better to do `common.EncodeURLValues("/other/" + pathstuff, params)`

exchanges/lbank/lbank.go
```diff
@@ -388,18 +392,18 @@ func (e *Exchange) GetOpenOrders(ctx context.Context, pair, pageNumber, pageLeng
 }
 
 // USD2RMBRate finds USD-CNY Rate
-func (e *Exchange) USD2RMBRate(ctx context.Context) (ExchangeRateResponse, error) {
-	var resp ExchangeRateResponse
-	path := fmt.Sprintf("/v%s/%s", lbankAPIVersion1, lbankUSD2CNYRate)
-	return resp, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &resp)
+func (e *Exchange) USD2RMBRate(ctx context.Context) (string, error) {
```
shazbert:
```diff
diff --git a/exchanges/lbank/lbank.go b/exchanges/lbank/lbank.go
index 48bd4197e..753b19525 100644
--- a/exchanges/lbank/lbank.go
+++ b/exchanges/lbank/lbank.go
@@ -27,6 +27,7 @@ import (
        exchange "github.com/thrasher-corp/gocryptotrader/exchanges"
        "github.com/thrasher-corp/gocryptotrader/exchanges/order"
        "github.com/thrasher-corp/gocryptotrader/exchanges/request"
+       "github.com/thrasher-corp/gocryptotrader/types"
 )

 // Exchange implements exchange.IBotExchange and contains additional specific api methods for interacting with Lbank
@@ -392,10 +393,10 @@ func (e *Exchange) GetOpenOrders(ctx context.Context, pair, pageNumber, pageLeng
 }

 // USD2RMBRate finds USD-CNY Rate
-func (e *Exchange) USD2RMBRate(ctx context.Context) (string, error) {
-       var rate string
+func (e *Exchange) USD2RMBRate(ctx context.Context) (float64, error) {
+       var rate types.Number
        path := "/v" + lbankAPIVersion2 + "/" + lbankUSD2CNYRate
-       return rate, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &rate)
+       return rate.Float64(), e.SendHTTPRequest(ctx, exchange.RestSpot, path, &rate)
 }

 // GetWithdrawConfig gets information about withdrawals
```

shazbert (changes requested):
One thing from me. 🚀

PARMESHWARPANWAR:
@cranktakular @shazbert all feedback addressed: 58dc5d5

Removed self-explanatory comments from GetTicker
Updated GetTicker to error if len(t) != 1
Fixed GetTimestamp to use types.Time directly
Fixed V2Response handling — check result == "false" first, then data != nil
Updated all paths to use common.EncodeURLValues
Updated USD2RMBRate to return float64 using types.Number
Removed unused lbankAPIVersion1 constant

shazbert (approved):
Changes ACK! Thanks.

gloriousCode (approved):
tACK! Thank you for your changes!

thrasher- (approved):
Thanks for this PR!

cranktakular (approved):
Looks good! I have tested and can ACKnowledge the validity of these changes.

PR #2255, opened 2026-05-19
OKX: add new ratelimit system + websocket ordering

gloriousCode:
# PR Description

This is all the recent improvements I've made on OKX. ~Tried to make orderbook checksums more efficient.~ It also comes with some rate limit changes since OKX has subaccount and instrument level limits for ordering.
Some of this may look familiar, and its because a lot is taken from my other PRs. This one is more important.

OKX no longer provides orderbook checksums, so this removes the checksum work and uses `seqId`/`prevSeqId` for continuity and recovery instead.

Also includes:

- Instrument, option-family and subaccount trade limits across REST and websocket
- Decimal-safe order fields using `types.Number`
- Correct reduce-only hedge order sides
- Websocket batch cancellation
- Fractional spot and margin amendments
- Tests covering the new orderbook, rate-limit and order behaviour

Includes work from @shazbert with the original authorship preserved.

PR #2259, opened 2026-05-27
Kraken: model real rate limit with per-endpoint costs

valli0x:
The previous limiter was hard-coded at 1 request per second with burst=1 for every endpoint, which is far more restrictive than Kraken's actual policy (counter max 20 with -0.5/sec decay on the Verified tier) and treats every endpoint as having the same cost. As a result:

- AddOrder/CancelOrder contended with the same bucket as background reconciliation queries, so a burst of history reads could delay trading.
- TradesHistory/Ledgers/ClosedOrders (real cost +4 on Kraken's counter) were charged as +1 locally, drifting the emulated counter below the real one.
- No burst capacity at all — every request had to wait a full second.

This change introduces a Kraken-specific limiter built on golang.org/x/time/rate that mirrors the documented policy:

- Two underlying rate.Limiter instances: a general counter (rate 0.5/sec, burst 16 with a 20% safety margin under the real max of 20) and a separate trading counter for AddOrder/CancelOrder (rate 1.0/sec, burst 60), reflecting Kraken's per-pair trading limit which is independent of the general counter.
- Kraken-specific EndpointLimit constants (krakenLimitPublic, krakenLimitBalance, krakenLimitHistory, krakenLimitTrading, krakenLimitWithdraw, krakenLimitDefault) offset by +100 to avoid collision with request.Unset/Auth/UnAuth.
- A methodToLimit map that routes each REST method name to the right endpoint limit, with weights matching Kraken's documented costs (+4 for history endpoints, +1 for balance/info/withdraw, trading on its own limiter).
- SendHTTPRequest now passes krakenLimitPublic and SendAuthenticatedHTTPRequest looks up the limit by method name instead of always passing request.Unset.

Reference: https://support.kraken.com/articles/206548367

# PR Description

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/kraken/kraken.go
+++ b/exchanges/kraken/kraken.go
@@ -49,7 +49,7 @@ func (e *Exchange) GetCurrentServerTime(ctx context.Context) (*TimeResponse, err
 	path := fmt.Sprintf("/%s/public/%s", krakenAPIVersion, krakenServerTime)
 
 	var result TimeResponse
-	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &result); err != nil {
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &result); err != nil {
 		return nil, err
 	}
 
@@ -81,7 +81,7 @@ func (e *Exchange) SeedAssets(ctx context.Context) error {
 func (e *Exchange) GetAssets(ctx context.Context) (map[string]*Asset, error) {
 	path := fmt.Sprintf("/%s/public/%s", krakenAPIVersion, krakenAssets)
 	var result map[string]*Asset
-	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &result); err != nil {
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &result); err != nil {
 		return nil, err
 	}
 	return result, nil
@@ -105,7 +105,7 @@ func (e *Exchange) GetAssetPairs(ctx context.Context, assetPairs []string, info 
 		}
 		params.Set("info", info)
 	}
-	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path+params.Encode(), &result); err != nil {
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path+params.Encode(), &result); err != nil {
 		return nil, err
 	}
 	return result, nil
@@ -122,7 +122,7 @@ func (e *Exchange) GetTicker(ctx context.Context, symbol currency.Pair) (*Ticker
 
 	var data map[string]*TickerResponse
 	path := fmt.Sprintf("/%s/public/%s?%s", krakenAPIVersion, krakenTicker, values.Encode())
-	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &data); err != nil {
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &data); err != nil {
 		return nil, err
 	}
 
@@ -155,7 +155,7 @@ func (e *Exchange) GetTickers(ctx context.Context, pairList string) (map[string]
 	var result map[string]*TickerResponse
 	path := fmt.Sprintf("/%s/public/%s?%s", krakenAPIVersion, krakenTicker, values.Encode())
 
-	err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &result)
+	err := e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &result)
 	if err != nil {
 		return nil, err
 	}
@@ -196,7 +196,7 @@ func (e *Exchange) GetOHLC(ctx context.Context, symbol currency.Pair, interval s
 	path := fmt.Sprintf("/%s/public/%s?%s", krakenAPIVersion, krakenOHLC, values.Encode())
 
 	result := make(map[string]any)
-	err = e.SendHTTPRequest(ctx, exchange.RestSpot, path, &result)
+	err = e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &result)
 	if err != nil {
 		return nil, err
 	}
@@ -266,7 +266,7 @@ func (e *Exchange) GetDepth(ctx context.Context, symbol currency.Pair) (*Orderbo
 	}
 
 	result := make(map[string]*orderbookStructure)
-	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &result); err != nil {
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &result); err != nil {
 		return nil, err
 	}
 
@@ -308,7 +308,7 @@ func (e *Exchange) GetTrades(ctx context.Context, symbol currency.Pair, since ti
 
 	path := common.EncodeURLValues("/"+krakenAPIVersion+"/public/"+krakenTrades, values)
 	var resp *RecentTradesResponse
-	return resp, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &resp)
+	return resp, e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &resp)
 }
 
 // GetSpread returns the full spread on Kraken
@@ -324,13 +324,13 @@ func (e *Exchange) GetSpread(ctx context.Context, symbol currency.Pair, since ti
 	}
 	var peanutButter *SpreadResponse
 	path := common.EncodeURLValues("/"+krakenAPIVersion+"/public/"+krakenSpread, values)
-	return peanutButter, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &peanutButter)
+	return peanutButter, e.SendHTTPRequest(ctx, exchange.RestSpot, krakenLimitPublic, path, &peanutButter)
 }
 
 // GetBalance returns your balance associated with your keys
 func (e *Exchange) GetBalance(ctx context.Context) (map[string]Balance, error) {
 	var result map[string]Balance
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenBalance, url.Values{}, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenBalance, url.Values{}, &result); err != nil {
 		return nil, err
 	}
 
@@ -345,7 +345,7 @@ func (e *Exchange) GetWithdrawInfo(ctx context.Context, withdrawalAsset, withdra
 	params.Set("amount", strconv.FormatFloat(amount, 'f', -1, 64))
 
 	var result *WithdrawInformation
-	return result, e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenWithdrawInfo, params, &result)
+	return result, e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenWithdrawInfo, params, &result)
 }
 
 // Withdraw withdraws funds
@@ -356,7 +356,7 @@ func (e *Exchange) Withdraw(ctx context.Context, withdrawalAsset, withdrawalKey 
 	params.Set("amount", strconv.FormatFloat(amount, 'f', -1, 64))
 
 	var referenceID string
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenWithdraw, params, &referenceID); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitWithdraw, krakenWithdraw, params, &referenceID); err != nil {
 		return referenceID, err
 	}
 
@@ -369,7 +369,7 @@ func (e *Exchange) GetDepositMethods(ctx context.Context, a string) ([]DepositMe
 	params.Set("asset", a)
 
 	var result []DepositMethods
-	err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenDepositMethods, params, &result)
+	err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenDepositMethods, params, &result)
 	if err != nil {
 		return nil, err
 	}
@@ -392,7 +392,7 @@ func (e *Exchange) GetTradeBalance(ctx context.Context, args ...TradeBalanceOpti
 	}
 
 	var result TradeBalanceInfo
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenTradeBalance, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenTradeBalance, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -412,7 +412,7 @@ func (e *Exchange) GetOpenOrders(ctx context.Context, args OrderInfoOptions) (*O
 	}
 
 	var result OpenOrders
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenOpenOrders, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenOpenOrders, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -448,7 +448,7 @@ func (e *Exchange) GetClosedOrders(ctx context.Context, args GetClosedOrdersOpti
 	}
 
 	var result ClosedOrders
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenClosedOrders, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitHistory, krakenClosedOrders, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -474,7 +474,7 @@ func (e *Exchange) QueryOrdersInfo(ctx context.Context, args OrderInfoOptions, t
 	}
 
 	var result map[string]OrderInfo
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenQueryOrders, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitHistory, krakenQueryOrders, params, &result); err != nil {
 		return result, err
 	}
 
@@ -508,7 +508,7 @@ func (e *Exchange) GetTradesHistory(ctx context.Context, args ...GetTradesHistor
 	}
 
 	var result TradesHistory
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenTradeHistory, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitHistory, krakenTradeHistory, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -530,7 +530,7 @@ func (e *Exchange) QueryTrades(ctx context.Context, trades bool, txid string, tx
 	}
 
 	var result map[string]TradeInfo
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenQueryTrades, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitHistory, krakenQueryTrades, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -550,7 +550,7 @@ func (e *Exchange) OpenPositions(ctx context.Context, docalcs bool, txids ...str
 	}
 
 	var result map[string]Position
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenOpenPositions, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenOpenPositions, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -588,7 +588,7 @@ func (e *Exchange) GetLedgers(ctx context.Context, args ...GetLedgersOptions) (*
 	}
 
 	var result Ledgers
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLedgers, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitHistory, krakenLedgers, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -606,7 +606,7 @@ func (e *Exchange) QueryLedgers(ctx context.Context, id string, ids ...string) (
 	}
 
 	var result map[string]LedgerInfo
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenQueryLedgers, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitHistory, krakenQueryLedgers, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -633,7 +633,7 @@ func (e *Exchange) GetTradeVolume(ctx context.Context, feeinfo bool, symbol ...c
 	}
 
 	var result *TradeVolumeResponse
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenTradeVolume, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenTradeVolume, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -698,7 +698,7 @@ func (e *Exchange) AddOrder(ctx context.Context, symbol currency.Pair, side, ord
 	}
 
 	var result AddOrderResponse
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenOrderPlace, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitTrading, krakenOrderPlace, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -712,7 +712,7 @@ func (e *Exchange) CancelExistingOrder(ctx context.Context, txid string) (*Cance
 	}
 
 	var result CancelOrderResponse
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenOrderCancel, values, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitTrading, krakenOrderCancel, values, &result); err != nil {
 		return nil, err
 	}
 
@@ -720,7 +720,7 @@ func (e *Exchange) CancelExistingOrder(ctx context.Context, txid string) (*Cance
 }
 
 // SendHTTPRequest sends an unauthenticated HTTP requests
-func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path string, result any) error {
+func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, epl request.EndpointLimit, path string, result any) error {
 	endpoint, err := e.API.Endpoints.GetURL(ep)
 	if err != nil {
 		return err
@@ -737,7 +737,7 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path st
 		HTTPMockDataSliceLimit: e.HTTPMockDataSliceLimit,
 	}
 
-	err = e.SendPayload(ctx, request.Unset, func() (*request.Item, error) {
+	err = e.SendPayload(ctx, epl, func() (*request.Item, error) {
 		return item, nil
 	}, request.UnauthenticatedRequest)
 	if err != nil {
@@ -769,7 +769,7 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path st
 }
 
 // SendAuthenticatedHTTPRequest sends an authenticated HTTP request
-func (e *Exchange) SendAuthenticatedHTTPRequest(ctx context.Context, ep exchange.URL, method string, params url.Values, result any) error {
+func (e *Exchange) SendAuthenticatedHTTPRequest(ctx context.Context, ep exchange.URL, epl request.EndpointLimit, method string, params url.Values, result any) error {
 	creds, err := e.GetCredentials(ctx)
 	if err != nil {
 		return err
@@ -780,7 +780,7 @@ func (e *Exchange) SendAuthenticatedHTTPRequest(ctx context.Context, ep exchange
 	}
 
 	interim := json.RawMessage{}
-	err = e.SendPayload(ctx, request.Unset, func() (*request.Item, error) {
+	err = e.SendPayload(ctx, epl, func() (*request.Item, error) {
 		nonce := e.Requester.GetNonce(nonce.UnixNano).String()
 		params.Set("nonce", nonce)
 		encoded := params.Encode()
@@ -912,7 +912,7 @@ func (e *Exchange) GetCryptoDepositAddress(ctx context.Context, method, code str
 	}
 
 	var result []DepositAddress
-	err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenDepositAddresses, values, &result)
+	err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenDepositAddresses, values, &result)
 	if err != nil {
 		return nil, err
 	}
@@ -932,7 +932,7 @@ func (e *Exchange) WithdrawStatus(ctx context.Context, c currency.Code, method s
 	}
 
 	var result []WithdrawStatusResponse
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenWithdrawStatus, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenWithdrawStatus, params, &result); err != nil {
 		return nil, err
 	}
 
@@ -946,7 +946,7 @@ func (e *Exchange) WithdrawCancel(ctx context.Context, c currency.Code, refID st
 	params.Set("refid", refID)
 
 	var result bool
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenWithdrawCancel, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitWithdraw, krakenWithdrawCancel, params, &result); err != nil {
 		return result, err
 	}
 
@@ -956,7 +956,7 @@ func (e *Exchange) WithdrawCancel(ctx context.Context, c currency.Code, refID st
 // GetWebsocketToken returns a websocket token
 func (e *Exchange) GetWebsocketToken(ctx context.Context) (string, error) {
 	var response WsTokenResponse
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenWebsocketToken, url.Values{}, &response); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenWebsocketToken, url.Values{}, &response); err != nil {
 		return "", err
 	}
 	return response.Token, nil
--- a/exchanges/kraken/kraken.go
+++ b/exchanges/kraken/kraken.go
@@ -330,7 +330,7 @@ func (e *Exchange) GetSpread(ctx context.Context, symbol currency.Pair, since ti
 // GetBalance returns your balance associated with your keys
 func (e *Exchange) GetBalance(ctx context.Context) (map[string]Balance, error) {
 	var result map[string]Balance
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitBalance, krakenBalance, url.Values{}, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitAuth, krakenBalance, url.Values{}, &result); err != nil {
 		return nil, err
 	}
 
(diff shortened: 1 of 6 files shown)


gloriousCode (changes requested):
Thank you for your contribution! Looking over it, its a benefit to have these updated, thank you.

I have a few things to raise. I have a diff gist to help guide you on some of the things: https://gist.github.com/gloriousCode/76076a8e376c23ccecd5f583b6c18e46

- [Public Futures endpoints incur no ratelimiting](https://docs.kraken.com/api/docs/guides/futures-rate-limits/). Here's Kraken's wording on it:

> Public endpoints do not have a cost and therefore do not count against any rate limiting budget

- I don't think its necessary to add the 80% buffer for rate limiting. We take it as either its correct or incorrect. It will limit regardless, so we may as well limit to its actual limit. If you still disagree, I think 80% is way too conservative

- The rate limits for authenticated order endpoints for Kraken are also _by currency pair_. We don't currently have infrastructure for supporting that. So I think its good to accept your rate limits at present and we can improve them in a future PR. [This PR here ](https://github.com/thrasher-corp/gocryptotrader/pull/2255) attempts to add an additional layer of rate limiting since okx has limits on instruments too.

- Your rate limit styling is inconsistent with our others. So in the diff I update it to better fit in

valli0x:
Applied your diff in c25a38f0:
- Removed the 80% safety margin; using the documented max counter of 20 directly
- Added `krakenLimitFuturesPublic` with cost 0 since public Kraken endpoints don't count against any budget
- Refactored `SendHTTPRequest`/`SendAuthenticatedHTTPRequest` to take an `EndpointLimit` parameter so each call site declares its real cost, matching the styling in your other exchanges
- Kept the trading limiter as a single shared limiter; per-pair trading limits can be a follow-up when the infrastructure is in place (as you suggested)
- Added `ratelimit_test.go` covering construction and per-endpoint pool routing

Squashed everything into a single commit for clarity. Let me know if anything else needs adjustment.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,165 @@
...
+
+		// Kraken-specific endpoint costs
+		krakenLimitDefault:  withWeight(general, 1),
+		krakenLimitPublic:   withWeight(general, 1),
+		krakenLimitBalance:  withWeight(general, 1),
+		krakenLimitHistory:  withWeight(general, 4),
+		krakenLimitTrading:  withWeight(orders, 1),
+		krakenLimitWithdraw: withWeight(general, 1),
+	}
+}
```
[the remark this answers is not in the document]
valli0x: Fixed in c25a38f0 — separate `public` limiter added, krakenLimitPublic routes through it.

exchanges/kraken/kraken_ratelimit.go
```diff
@@ -0,0 +1,165 @@
...
+	// Verified tier (default). Override at build/config time for Pro tier.
+	krakenSpotMaxCounter  = 16  // real max 20, 80% safety margin
+	krakenSpotDecayPerSec = 0.5 // tokens per second (Verified tier)
+
+	// Trading endpoints (AddOrder/CancelOrder) — Kraken tracks these per
+	// currency pair under a separate counter. The exact policy is based on
+	// order book interaction; conservative defaults below.
+	krakenSpotOrderMaxBurst = 60
+	krakenSpotOrderRate     = 1.0 // tokens per second
+)
```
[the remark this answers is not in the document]
valli0x: Fixed in c25a38f0 — promoted to exported package-level vars (KrakenSpotMaxCounter, KrakenSpotDecayPerSec, etc.).

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+//	Withdraw, WithdrawCancel                                            +1
+//
+// We emulate each system with a separate golang.org/x/time/rate.Limiter
+// where:
+//
+//	Kraken counter max  ↔  rate.Limiter burst
+//	Kraken decay /sec   ↔  rate.Limiter refill rate
+//
+// Tunables. These are package-level vars rather than consts so they can be
+// overridden at process start-up (for example to switch the private counter
```
cranktakular: While I can see some merit in this idea, it looks like there's currently no way for them to be overridden yet, so they may as well be consts now. If someone implements a way for such values to be changed at startup they'd need to rewrite code anyway, so it's fine for them to need to rewrite this part.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+// integration tests). They are not safe to mutate after the Requester has
+// been constructed.
+var (
+	// Private general counter (Verified tier defaults).
+	KrakenSpotMaxCounter  = 20  // documented maximum counter
+	KrakenSpotDecayPerSec = 0.5 // tokens per second (Verified tier)
+
+	// Private trading counter (AddOrder, CancelOrder).
+	KrakenSpotOrderMaxBurst = 60
+	KrakenSpotOrderRate     = 1.0 // tokens per second
```
cranktakular: I don't think this is accurate; the [more detailed section](https://support.kraken.com/articles/360045239571-trading-rate-limits) gives noticeably different numbers than this, and that part seems more up to date.

Also, you should probably make it clearer that this endpoint doesn't just apply to adding and cancelling orders. In case someone adds functions for editing and amending orders in the future.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+	KrakenPublicRate     = 1.0 // tokens per second
+)
+
+// Kraken-specific EndpointLimit values. These extend the global Unset/Auth/
+// UnAuth constants from the request package and let each endpoint declare its
+// real cost so the limiter charges the correct number of tokens per call.
+const (
+	// krakenLimitDefault is used for unrecognised endpoints; weight 1 keeps
+	// it conservative.
+	krakenLimitDefault request.EndpointLimit = iota + 100
```
cranktakular: Why iota + 100 and not just iota?

Also, you don't need this default. It's not used in any actual files, and anyone who is implementing a new function without knowing the appropriate rate should find out a suitable one to assign.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+	private := newKrakenSpotPrivateLimiter()
+	orders := newKrakenSpotOrderLimiter()
+	public := newKrakenPublicLimiter()
+
+	return request.RateLimitDefinitions{
+		// Backward compatibility — anything still passing the global
+		// constants gets the private limiter at weight 1.
+		request.Unset:  request.GetRateLimiterWithWeight(private, 1),
+		request.Auth:   request.GetRateLimiterWithWeight(private, 1),
+		request.UnAuth: request.GetRateLimiterWithWeight(public, 1),
```
cranktakular: This is unnecessary. Signals aren't being received from older versions of the program, and current ones should always be setting Kraken-specific endpoint limits.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+	// modest steady-state with healthy burst suits both bots and
+	// occasional consumers.
+	KrakenPublicMaxBurst = 15
+	KrakenPublicRate     = 1.0 // tokens per second
+)
+
+// Kraken-specific EndpointLimit values. These extend the global Unset/Auth/
+// UnAuth constants from the request package and let each endpoint declare its
+// real cost so the limiter charges the correct number of tokens per call.
+const (
```
cranktakular, line 12: The comments at the start of, and within, this block of constants are far too verbose.

Big explainers are for functions or values that are exported. Parts which are unexported [should only have comments for unclear behaviour](https://github.com/thrasher-corp/gocryptotrader/blob/master/docs/CODING_GUIDELINES.md#comments).

You do not need to explain how they affect the general counter (that's made clear by the code itself), or which endpoints use that limit (endpoints which should use it, will simply use it).

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
+package kraken
+
+import (
+	"github.com/thrasher-corp/gocryptotrader/exchanges/request"
+	"golang.org/x/time/rate"
+)
+
+// Kraken applies two independent REST rate-limit systems:
```
cranktakular: The way this comment is placed within the file, it is treated as a general comment that should show up for every single variable declared in the block below it. This [breaks convention](https://go.dev/doc/comment#const) and negatively impacts readability.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+//     (balance, info, history, withdraw). Different endpoint groups carry
+//     different weights matching Kraken's documented costs.
+//   - orders:  used only by AddOrder/CancelOrder.
+//   - public:  used only by unauthenticated REST endpoints. Independent of
+//     the private counter because Kraken's public limit is enforced per
+//     IP, not per API key.
+//
+// The legacy request.Unset/Auth/UnAuth keys are kept and routed to
+// `private` for backward compatibility — anything that still passes them
+// will behave as before, just with the new tier-correct parameters.
```
cranktakular: This comment marks the fourth time in this file the basics of the exchange's rate limiting system were explained in comments. That's unnecessary.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+		request.UnAuth: request.GetRateLimiterWithWeight(public, 1),
+
+		// Kraken-specific endpoint costs.
+		krakenLimitDefault:       request.GetRateLimiterWithWeight(private, 1),
+		krakenLimitPublic:        request.GetRateLimiterWithWeight(public, 1),
+		krakenLimitFuturesPublic: request.NewRateLimitWithWeight(0, 0, 1),
+		krakenLimitBalance:       request.GetRateLimiterWithWeight(private, 1),
+		krakenLimitHistory:       request.GetRateLimiterWithWeight(private, 4),
+		krakenLimitTrading:       request.GetRateLimiterWithWeight(orders, 1),
+		krakenLimitWithdraw:      request.GetRateLimiterWithWeight(private, 1),
```
cranktakular: This is exactly the same as krakenLimitBalance, and they aren't even distinguished in the documentation; these should be merged.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+// will behave as before, just with the new tier-correct parameters.
+func buildKrakenRateLimits() request.RateLimitDefinitions {
+	private := newKrakenSpotPrivateLimiter()
+	orders := newKrakenSpotOrderLimiter()
+	public := newKrakenPublicLimiter()
+
+	return request.RateLimitDefinitions{
+		// Backward compatibility — anything still passing the global
+		// constants gets the private limiter at weight 1.
+		request.Unset:  request.GetRateLimiterWithWeight(private, 1),
```
cranktakular: This PR does not account for the [rate limits of private futures endpoints](https://docs.kraken.com/api/docs/guides/futures-rate-limits/). With the current implementation, they would instead take away from authenticated spot requests.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,160 @@
...
+	KrakenSpotDecayPerSec = 0.5 // tokens per second (Verified tier)
+
+	// Private trading counter (AddOrder, CancelOrder).
+	KrakenSpotOrderMaxBurst = 60
+	KrakenSpotOrderRate     = 1.0 // tokens per second
+
+	// Public API counter (per IP). Conservative defaults; Kraken
+	// documents a generous public rate that varies by endpoint, but a
+	// modest steady-state with healthy burst suits both bots and
+	// occasional consumers.
```
cranktakular: This comment doesn't actually say much of relevance.

For something more to the point and pertinent, I'd just point out that a precise burst limit wasn't stated, but that 15 has been tested and found safe.

exchanges/kraken/ratelimit_test.go
```diff
@@ -0,0 +1,79 @@
...
+	assert.ErrorIs(t, requester.InitiateRateLimit(request.WithDelayNotAllowed(t.Context()), krakenLimitBalance), request.ErrDelayNotAllowed, "balance limit should share the private limiter")
+	assert.NoError(t, requester.InitiateRateLimit(request.WithDelayNotAllowed(t.Context()), krakenLimitTrading), "trading limit should use a separate limiter")
+	assert.NoError(t, requester.InitiateRateLimit(request.WithDelayNotAllowed(t.Context()), krakenLimitPublic), "public limit should use a separate limiter")
+	assert.NoError(t, requester.InitiateRateLimit(request.WithDelayNotAllowed(t.Context()), krakenLimitFuturesPublic), "futures public limit should not consume a rate budget")
+
+	err = requester.InitiateRateLimit(request.WithDelayNotAllowed(t.Context()), request.UnAuth)
+	if errors.Is(err, request.ErrDelayNotAllowed) {
+		t.Fatal("unauthenticated limit must not share the exhausted private limiter")
+	}
+	require.NoError(t, err, "unauthenticated limit must use the public limiter")
```
cranktakular: These two error checks are largely redundant. I don't think the slightly different failure reason deserves a separate check, when that failure would be caught by the latter check anyway.

Also, the former is using an outdated structure. If that's seen as more important, it should just be require.NotErrorIs()

exchanges/kraken/kraken.go
```diff
@@ -474,7 +474,7 @@ func (e *Exchange) QueryOrdersInfo(ctx context.Context, args OrderInfoOptions, t
 	}
 
 	var result map[string]OrderInfo
-	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenQueryOrders, params, &result); err != nil {
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitHistory, krakenQueryOrders, params, &result); err != nil {
```
cranktakular: I'm a bit unsure on whether this is meant to be an account history endpoint or not. You might want to contact Kraken support to get a more proper list. Otherwise, I'm not interested in fighting over this one, so you can leave it as-is.

cranktakular (changes requested):
Thanks for the contribution! But there are some things I would like to see changed before this is merged.

cranktakular:
@gloriousCode 

>     * I don't think its necessary to add the 80% buffer for rate limiting. We take it as either its correct or incorrect. It will limit regardless, so we may as well limit to its actual limit. If you still disagree, I think 80% is way too conservative

The potential issue is that it seems like, if the limit is exceeded, the key may be punitively locked out for a few seconds. If we're already at 20, sending a request that occupies 1, instead of the exchange rejecting it until 2 seconds have passed, it might block all requests for 5 seconds, continually extending that if we send any requests in that time.

> If the rate limits are reached, additional calls will be restricted for a few seconds (or possibly longer if calls continue to be made while the rate limits are active).

It might be worthwhile to be careful, in case they receive requests in an unfortunate cluster.

Or that assumption of punitive locking could be wrong. Who knows? If you still hold your view after considering this, I'll be happy to defer to that.

thrasher-:
@codex please review

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+	krakenLimitTrading
+	krakenLimitCancel
+)
+
+const (
+	// Spot private counter, Verified tier: max 20, decays 0.5/sec.
+	krakenSpotMaxCounter  = 20
+	krakenSpotDecayPerSec = 0.5
+
+	// Spot trading engine counter, Starter tier: max 60, decays 1/sec.
```
cranktakular, line 27: This implementation of the spot trade rate limit uses the starter tier as its basis, but the spot private rate limit uses the verified tier as its basis. 

For spot trading, the verified tier has a burst of 125 and decay of 2.34/sec ([source](https://docs.kraken.com/exchange/guides/general/ratelimits)). And for spot private, the starter tier has a burst of 15 and decay of 0.33/sec ([source](https://docs.kraken.com/exchange/guides/rest/ratelimits)).

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+	krakenPublicMaxBurst = 15
+	krakenPublicRate     = 1.0
+)
+
+// buildKrakenRateLimits returns limiters matching Kraken's documented
+// rate-limit behaviour: a spot private counter per API key with account
+// history endpoints costing double, a separate trading engine counter with
+// age-weighted cancels, a per-IP public limit, a futures cost pool of 500
+// units per 10 seconds (order management endpoints cost up to 10, so 50
+// requests per 10 seconds is a safe floor), and uncosted public futures
```
cranktakular, line 50: I don't think that is too safe. There are many endpoints with a heavier cost, including ones we already implement; [Initiate withdrawal to Spot wallet](https://docs.kraken.com/api-reference/transfers/initiate-withdrawal-to-spot-wallet) (implemented under the function `FuturesWithdrawToSpotWallet`) has a weight of 100.

Incorporating most weights would only require a moderate expansion of the current setup, but unfortunately, the ability to batch place futures orders (through `FuturesBatchOrder` and [Batch order management](https://docs.kraken.com/api-reference/order-management/batch-order-management)) has a dynamic weight based on how many orders were placed. I can tell this dynamic weight would be a massive pain to manage. If you can, that would be great, but if not, just do that easier stuff and we can handle it later.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+	private := rate.NewLimiter(rate.Limit(krakenSpotDecayPerSec), krakenSpotMaxCounter)
+	trading := rate.NewLimiter(rate.Limit(krakenSpotOrderRate), krakenSpotOrderMaxBurst)
+	public := rate.NewLimiter(rate.Limit(krakenPublicRate), krakenPublicMaxBurst)
+
+	return request.RateLimitDefinitions{
+		krakenLimitPublic:        request.GetRateLimiterWithWeight(public, 1),
+		krakenLimitFuturesPublic: request.NewRateLimitWithWeight(0, 0, 1),
+		krakenLimitFuturesAuth:   request.NewRateLimitWithWeight(10*time.Second, 50, 1),
+		krakenLimitAuth:          request.GetRateLimiterWithWeight(private, 1),
+		krakenLimitHistory:       request.GetRateLimiterWithWeight(private, 2),
```
cranktakular, line 64: Looks like the documentation says the weight of this is 2, while their support article says the weight is 4.

We'll try to contact Kraken's support about this to get some clarification.

exchanges/kraken/kraken.go
cranktakular: Looks like I was wrong to doubt; [this support article](https://support.kraken.com/articles/206548367-what-are-the-api-rate-limits-) specifically lists this as one of the endpoints with a higher weight, so you will want to revert this unless something else comes up.

exchanges/kraken/ratelimit_test.go
```diff
@@ -0,0 +1,52 @@
...
+	assert.ErrorIs(t, requester.InitiateRateLimit(ctx, krakenLimitAuth), request.ErrDelayNotAllowed, "auth should share the exhausted private counter")
+
+	assert.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitTrading), "trading should use a separate counter")
+	assert.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitPublic), "public should use a separate limiter")
+	assert.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitFuturesPublic), "futures public should not consume any budget")
+	assert.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitFuturesAuth), "futures auth should use a separate limiter")
+
+	// One trading unit is already consumed above; 7 worst-case cancels
+	// consume another 56 of the 60 burst, and the next cannot fit
+	for range 7 {
```
cranktakular, line 48: Optional: Might make it more maintainable for the future, if any of these values end up changing.
```suggestion
	for range (krakenSpotOrderMaxBurst - 1) / krakenCancelWeight {
```
If implemented, may want to tweak the comment accordingly.

cranktakular (changes requested):
Great work! Still have a few changes I want.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+// rate-limit behaviour: a spot private counter per API key with account
+// history endpoints costing double, a separate trading engine counter with
+// age-weighted cancels, a per-IP public limit, a futures cost pool of 500
+// units per 10 seconds (order management endpoints cost up to 10, so 50
+// requests per 10 seconds is a safe floor), and uncosted public futures
+// endpoints.
+//
+// Reference: https://support.kraken.com/articles/206548367
+func buildKrakenRateLimits() request.RateLimitDefinitions {
+	private := rate.NewLimiter(rate.Limit(krakenSpotDecayPerSec), krakenSpotMaxCounter)
```
thrasher-, line 55: Heimdall: **Blocker** — giving the private counter a burst lets authenticated requests arrive at Kraken out of nonce order, which Kraken rejects.

**Fix:** serialise authenticated dispatch, or hold the nonce lock across the round trip. The nonce is taken inside `newRequest()` ([kraken.go:784](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/kraken.go#L784)) via `Requester.GetNonce` → `timedLock.LockForDuration()`, and `DefaultMutexLockTimeout` is 50 ms ([request_types.go:15](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/request/request_types.go#L15)). Any round trip longer than that releases the lock mid-flight and the next goroutine takes a higher nonce while the earlier request is still on the wire.

**Why:** master's `NewBasicRateLimit(time.Second, 1, 1)` has a burst of 1, so authenticated requests were released one per second and could not overlap. This burst of 20 releases them together. Kraken requires an always-increasing nonce and returns `EAPI:Invalid nonce` when one arrives late — [Why am I getting Invalid nonce errors?](https://support.kraken.com/au/articles/360001148063-why-am-i-getting-invalid-nonce-errors-)

Reproduced with a transport that delays only the first request by 150 ms — ordinary latency variance, well under the ~330 ms round trip I measured to `api.kraken.com` — recording nonce order as each request lands. Four concurrent `SendPayload` calls, three runs:

```
master NewBasicRateLimit(1s,1,1)   arrivals=4  out-of-order=0
PR     krakenLimitAuth (burst 20)  arrivals=4  out-of-order=1
master NewBasicRateLimit(1s,1,1)   arrivals=4  out-of-order=0
PR     krakenLimitAuth (burst 20)  arrivals=4  out-of-order=1
master NewBasicRateLimit(1s,1,1)   arrivals=4  out-of-order=0
PR     krakenLimitAuth (burst 20)  arrivals=4  out-of-order=1
```

3/3 on the branch, 0/3 on master. `krakenLimitTrading` at [line 56](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L56) has a burst of 60 and the same exposure. The burst itself is the right model of Kraken's counter — it's the nonce path underneath that isn't ready for concurrent authenticated requests.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+	trading := rate.NewLimiter(rate.Limit(krakenSpotOrderRate), krakenSpotOrderMaxBurst)
+	public := rate.NewLimiter(rate.Limit(krakenPublicRate), krakenPublicMaxBurst)
+
+	return request.RateLimitDefinitions{
+		krakenLimitPublic:        request.GetRateLimiterWithWeight(public, 1),
+		krakenLimitFuturesPublic: request.NewRateLimitWithWeight(0, 0, 1),
```
thrasher-, line 61: Heimdall: **Should fix** — this isn't "cost 0", it's no rate limiting at all.

**Fix:** give it a generous but finite ceiling, e.g. `request.GetRateLimiterWithWeight(rate.NewLimiter(100, 100), 1)`.

**Why:** `request.NewRateLimit` returns `rate.NewLimiter(rate.Inf, 1)` for a non-positive interval or action count ([limit.go:55-64](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/request/limit.go#L55-L64)), so `NewRateLimitWithWeight(0, 0, 1)` is a no-op. Measured: 5000 consecutive `InitiateRateLimit` calls under `request.WithDelayNotAllowed`, every one allowed with zero delay, against a burst of 15 for [`krakenLimitPublic`](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L60) on the line above.

That covers all seven futures public methods — orderbook, charts, trades, instruments, tickers, ticker-by-symbol and trade history — each capped at 1/sec on master today. `DefaultSyncerWorkers` is 15 ([config_types.go:43](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/config/config_types.go#L43)), so `UpdateOrderbook`/`UpdateTicker` over futures pairs will fan out against no ceiling.

I could not prove Kraken pushes back, and I'd rather say so: [Derivatives rate limits](https://docs.kraken.com/api/docs/guides/futures-rate-limits) does state that public endpoints "do not have a cost and therefore do not count against any rate limiting budget", and 40 requests at 10-way concurrency (~24 req/s) to `/derivatives/api/v3/tickers/PF_XBTUSD` all returned 200. So this is a missing safety valve rather than a demonstrated defect — but it's the only limiter in the file without one, and `rate.Inf` leaves a paginating caller or a misconfigured sync interval with nothing between it and Kraken's edge.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+		krakenLimitFuturesAuth:   request.NewRateLimitWithWeight(10*time.Second, 50, 1),
```
thrasher-, line 62: Heimdall: **Minor** — this is the one limiter in the file with no burst, and it looks unintentional rather than chosen.

**Fix:** `request.GetRateLimiterWithWeight(rate.NewLimiter(5, 50), 1)` if a burst is wanted; otherwise a word in the comment saying the smoothing is deliberate.

**Why:** `NewRateLimit` hard-codes the burst — "Burst rate is kept as one as this is not supported for out-bound requests" ([limit.go:53-64](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/request/limit.go#L53-L64)) — so this resolves to `rate.NewLimiter(5, 1)`. Measured:

```
krakenLimitFuturesAuth   burst 1   sustained 200ms
```

The sustained throughput is exactly what the GoDoc intends (5/s = 50 per 10s), so this isn't wrong — but every other limiter here is built with an explicit `rate.NewLimiter(rate, burst)` and gets a burst, and this one silently doesn't because it's the only one routed through `NewRateLimitWithWeight`. Worth making the difference deliberate either way.

exchanges/kraken/ratelimit_test.go
```diff
@@ -0,0 +1,52 @@
...
+	// maximum fills it exactly
+	for range krakenSpotMaxCounter / 2 {
+		require.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitHistory), "history requests must be allowed up to the counter maximum")
+	}
+	assert.ErrorIs(t, requester.InitiateRateLimit(ctx, krakenLimitHistory), request.ErrDelayNotAllowed, "history should be limited once the private counter is exhausted")
+	assert.ErrorIs(t, requester.InitiateRateLimit(ctx, krakenLimitAuth), request.ErrDelayNotAllowed, "auth should share the exhausted private counter")
+
+	assert.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitTrading), "trading should use a separate counter")
+	assert.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitPublic), "public should use a separate limiter")
+	assert.NoError(t, requester.InitiateRateLimit(ctx, krakenLimitFuturesPublic), "futures public should not consume any budget")
```
thrasher-, line 43: Heimdall: **Minor** — four of the seven limits aren't actually pinned by this test.

**Fix:** assert the shape rather than "the first call is allowed" — drain `krakenPublicMaxBurst` public calls and check the next is throttled, and count the burst for the two futures limits instead of a single `NoError`.

**Why:** `assert.NoError` on one call passes for any limiter with a token free. Mutating `ratelimit.go` and re-running `TestBuildKrakenRateLimits`:

```
mutation                                            result
krakenLimitFuturesPublic -> 1 request per hour      SURVIVES
krakenLimitFuturesAuth   -> 1 request per hour      SURVIVES
krakenPublicMaxBurst 15 -> 1                        SURVIVES
krakenLimitPublic -> the trading limiter            SURVIVES
krakenLimitHistory weight 2 -> 1                    caught
krakenLimitCancel weight 8 -> 1                     caught
krakenLimitFuturesPublic -> the private limiter     caught
krakenSpotOrderMaxBurst 60 -> 9                     caught
```

The public-to-trading swap is worth calling out: it only appears caught because the substitution leaves `public` declared and unused. Delete that line too and the mutation compiles and the test passes — the sequence spends `1 + 1 + 7×8 = 58` trading units, still inside the burst of 60. So the four assertions on lines 41-44 pin the private counter and nothing else.

exchanges/kraken/ratelimit_test.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	requester, err := request.New("krakenRateLimits", http.DefaultClient, request.WithLimiter(rateLimits))
+	require.NoError(t, err, "request.New must not error")
+
+	ctx := request.WithDelayNotAllowed(t.Context())
+
+	// History requests cost 2 on the private counter, so half the counter
+	// maximum fills it exactly
+	for range krakenSpotMaxCounter / 2 {
```
thrasher-, line 35: Heimdall: **Nit** — `krakenSpotMaxCounter / 2` only drains the counter exactly when the maximum is even.

**Fix:** drive the loop off the weight, the way [cranktakular suggested](https://github.com/thrasher-corp/gocryptotrader/pull/2259#discussion_r3756077954) for the cancel loop below — lift the history weight to a named const and use `for range krakenSpotMaxCounter / krakenHistoryWeight`.

**Why:** if the counter maximum is corrected to the Starter tier's 15, this runs 7 times and leaves one token, and [line 39](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit_test.go#L39) ("auth should share the exhausted private counter") then only passes by accident — the rejected weight-2 call on line 38 burns the leftover token rather than returning it. Traced with `rate.Limiter.TokensAt`, `krakenSpotMaxCounter = 15`:

```
after 7 history calls:           1.0000
after the rejected history call: 0.0000
```

That token loss is a pre-existing quirk of `cancelAll` in `exchanges/request`, not something this PR introduced — but it is currently load-bearing for that assertion.

exchanges/kraken/ratelimit_test.go
```diff
@@ -0,0 +1,52 @@
...
+		krakenLimitFuturesAuth,
+		krakenLimitAuth,
+		krakenLimitHistory,
+		krakenLimitTrading,
+		krakenLimitCancel,
+	} {
+		assert.NotNilf(t, rateLimits[limit], "rate limit should exist for endpoint %d", limit)
+	}
+
+	requester, err := request.New("krakenRateLimits", http.DefaultClient, request.WithLimiter(rateLimits))
```
thrasher-, line 28: Heimdall: **Nit** — `http.DefaultClient` makes this test single-shot.

**Fix:** `&http.Client{}`, as [binance/ratelimit_test.go:35](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/binance/ratelimit_test.go#L35) does.

**Why:** `request.New` registers the client in a package-global tracker, so a second registration of the same client fails:

```
$ go test ./exchanges/kraken/ -run '^TestBuildKrakenRateLimits$' -count 2
    ratelimit_test.go:29:
        Error: Received unexpected error:
               cannot set up a new requester for krakenRateLimits: cannot reuse http client
        Messages: request.New must not error
```

binance, okx, poloniex and kucoin all do the same thing in their rate-limit tests and all fail the same way, so this is copying an existing pattern rather than introducing anything — but it also permanently claims `http.DefaultClient` for the whole kraken test binary.

exchanges/kraken/kraken.go
```diff
@@ -712,15 +712,15 @@ func (e *Exchange) CancelExistingOrder(ctx context.Context, txid string) (*Cance
...
+	if err := e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, krakenLimitCancel, krakenOrderCancel, values, &result); err != nil {
 		return nil, err
 	}
 
 	return &result, nil
 }
 
 // SendHTTPRequest sends an unauthenticated HTTP requests
-func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path string, result any) error {
+func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, epl request.EndpointLimit, path string, result any) error {
```
thrasher-, line 723: Heimdall: **Minor** — this is a breaking change for library consumers, worth a deliberate nod rather than a silent one.

**Why:** `SendHTTPRequest` and `SendAuthenticatedHTTPRequest` are exported, and GoCryptoTrader is consumed as a library as well as a binary. A caller built against the old signatures fails on this branch:

```
base legacy caller build exit=0
head legacy caller build exit=1
    not enough arguments in call to k.SendHTTPRequest
    not enough arguments in call to k.SendAuthenticatedHTTPRequest
```

In-repo callers are all updated, and the new shape matches gateio, poloniex, deribit and kucoin, so this is aligning with the house style rather than diverging — I'd just want it stated in the PR description so it isn't a surprise in the release notes.

thrasher- (changes requested):
Heimdall: The mapping is complete — 9 spot public, 20 spot private, 7 futures public and the single futures-auth path each declare a limit, and no REST call site is left on `request.Unset`. The signature change matches how gateio, poloniex, deribit and kucoin already thread `epl request.EndpointLimit`. The trading-counter constants are right, including the age-tapered cancel: Kraken's [Spot Trading Limits](https://docs.kraken.com/exchange/guides/general/ratelimits) gives Cancel Order +8/+6/+5/+4/+2/+1 by resting time and Starter 60 @ -1/s, which is exactly what [ratelimit.go:27-37](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L27-L37) encodes.

Measured burst and steady state per limit against master's single `NewBasicRateLimit(time.Second, 1, 1)`:

```
limit                       burst   sustained     master
krakenLimitPublic              15        1.0s   1 / 1.0s
krakenLimitFuturesPublic     none        none   1 / 1.0s
krakenLimitFuturesAuth          1      200 ms   1 / 1.0s
krakenLimitAuth                20        2.0s   1 / 1.0s
krakenLimitHistory             10        4.0s   1 / 1.0s
krakenLimitTrading             60        1.0s   1 / 1.0s
krakenLimitCancel               7        8.0s   1 / 1.0s
```

Evidence for each finding is on the line it applies to.

**To unblock:**

1. [ratelimit.go:55](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L55) — the private counter's burst of 20 lets authenticated requests overlap, and with any latency variance above the 50 ms nonce lock they reach Kraken out of nonce order. Reproduced 3/3 on the branch, 0/3 on master. `krakenLimitTrading`'s burst of 60 has the same exposure.

Not gating: [ratelimit.go:61](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L61) futures public has no ceiling at all rather than "cost 0"; [ratelimit.go:62](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L62) is the only limiter with no burst; the exported signature change breaks external callers; and three test nits.

Still open from the 3 and 11 August rounds and untouched since [`cc89f78`](https://github.com/thrasher-corp/gocryptotrader/commit/cc89f78a747572404230985fbcdf98578891d5cb) on 29 July: the futures per-endpoint weights, the history 2-vs-4 discrepancy, reverting `QueryOrdersInfo`, `ClosedOrders`, the tier mix, and the cancel-loop constant. I've replied to the Copilot `iota` thread — that one is a false positive.

The branch conflicts with master on one line, [kraken_futures.go:74](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/kraken_futures.go#L74) (master moved it to `common.EncodeURLValues`). Resolved locally the merge builds, lints clean and `go test ./exchanges/kraken/... -race -count 1` passes. The PR description still describes the first commit — the +100 offset, `methodToLimit`, `krakenLimitDefault`, burst 16, history weight 4 — none of which survives; worth a refresh before this is squashed.

The failing `ubuntu-24.04-arm` job is ambient: `gateio_test.go:1032`, nothing to do with kraken.

Happy to approve once the nonce path is sorted.

exchanges/kraken/ratelimit.go
```diff
@@ -0,0 +1,68 @@
...
+// limits applied to each.
+const (
+	krakenLimitPublic request.EndpointLimit = iota
+	krakenLimitFuturesPublic
+	krakenLimitFuturesAuth
+	krakenLimitAuth
+	krakenLimitHistory
+	krakenLimitTrading
+	krakenLimitCancel
+)
```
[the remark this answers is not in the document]
thrasher-: Heimdall: Checked this one — it doesn't bite, and the plain `iota` is what [cranktakular asked for](https://github.com/thrasher-corp/gocryptotrader/pull/2259#discussion_r3378173512).

The collision is real numerically but unreachable. [binance/ratelimit.go:31](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/binance/ratelimit.go#L31) uses the identical `request.EndpointLimit = iota` pattern starting at 0, so it is the house style. The only `request.Unset` uses left in the package are the five in `kraken_websocket.go`, and Kraken's `websocket.ManagerSetup` at [kraken_wrapper.go:197](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/kraken_wrapper.go#L197) never sets `RateLimitDefinitions`, so [connection.go:200](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchange/websocket/connection.go#L200) falls through to the per-connection `RateLimit` and never consults this map. Nothing outside the kraken package passes `request.Unset` to Kraken's requester either. No change wanted.

exchanges/kraken/ratelimit.go
thrasher-: Heimdall: Confirming your reading, with the numbers it costs.

Kraken's [Spot Trading Limits](https://docs.kraken.com/exchange/guides/general/ratelimits) gives thresholds Starter 60 / Intermediate 125 / Pro 180 and decay -1 / -2.34 / -3.75, and [Spot REST Rate Limits](https://docs.kraken.com/api/docs/guides/spot-rest-ratelimits) gives counters Starter 15 @ -0.33 / Intermediate 20 @ -0.5 / Pro 20 @ -1. So [ratelimit.go:24](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L24) is Intermediate and [ratelimit.go:31](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/ratelimit.go#L31) is Starter, as you said — each correct in isolation, but not the same account.

Measured, that mix costs a Starter-consistent 8.0s per sustained cancel. [kraken_wrapper.go:832](https://github.com/thrasher-corp/gocryptotrader/blob/cc89f78a747572404230985fbcdf98578891d5cb/exchanges/kraken/kraken_wrapper.go#L832) loops `CancelExistingOrder` per open order when the authenticated websocket is down:

```
open orders    Starter (as coded)    Intermediate      Pro
         25                 2m20s             32s       5s
         50                 5m40s           1m58s      59s
        100                12m20s           4m48s    2m45s
```

Going Starter throughout is the safe choice and I'd support it; it's worth knowing that it's a REST-fallback `CancelAllOrders` measured in minutes. Kraken's counter is also per pair, so one shared limiter multiplies this by the number of pairs holding orders — the follow-up you already agreed with, which matters more at Starter than it looked at Intermediate.

PR #2260, opened 2026-05-28
types: add PreciseNumber for lossless decimal JSON values

valli0x:
The existing Number type stores a float64 and reconstructs Decimal() via decimal.NewFromFloat, which silently loses precision past ~15 significant digits. For order-book mid-prices and ticker fields used only in approximate computation that is acceptable, but it makes Number unsuitable for accounting, reconciliation, settlement, or any path where the exact decimal value from the wire must be reproduced.

This adds a sibling type, PreciseNumber, with the same API surface (UnmarshalJSON/MarshalJSON/Float64/Int64/Decimal/String) but a struct representation that preserves the original numeric string alongside a parsed float64 cache.

Behaviour:

  - UnmarshalJSON accepts both quoted strings and bare numbers, the same as Number.
  - MarshalJSON emits the preserved digit string verbatim, so a parse/serialize round-trip is bit-exact rather than reformatted through float64.
  - Float64() returns the cached float for cheap comparisons and sorting (e.g. order-book level alignment).
  - Decimal() parses the original string directly via decimal.NewFromString, avoiding the float round-trip.
  - Int64() parses integer values directly and truncates fractional values toward zero, matching Number.Int64's semantics but without the float round-trip on values that originated as integers (timestamps in seconds, trade IDs).
  - The zero value is the number 0 and marshals to "" to match Number.MarshalJSON's existing contract.

Adoption is opt-in: exchange wrappers can replace types.Number with types.PreciseNumber in struct fields where the exact wire value matters, while leaving market-data fields on types.Number where the cheaper representation is preferable.

A round-trip test pins the bit-exact behaviour against a 17-digit value beyond float64's reliable round-trip range; a head-to-head test demonstrates the precision gap vs Number.Decimal on the same input.

# PR Description

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/types/precise_number.go
+++ b/types/precise_number.go
@@ -32,22 +32,19 @@ type PreciseNumber struct {
 	val float64 // parsed cache for fast Float64()
 }
 
-// NewPreciseNumberFromString constructs a PreciseNumber from a base-10 decimal
-// string without precision loss. Empty input, hexadecimal, scientific notation,
-// NaN and Inf are rejected.
+// NewPreciseNumberFromString constructs a PreciseNumber from a numeric string.
+// It accepts the same inputs as decimal.NewFromString (so scientific notation
+// is allowed, but Go-literal underscores and hexadecimal are not), and
+// additionally rejects values too large to be represented by a float64.
 func NewPreciseNumberFromString(s string) (PreciseNumber, error) {
-	if s == "" {
-		return PreciseNumber{}, fmt.Errorf("%w: empty string", errInvalidPreciseNumberValue)
-	}
 	return parsePreciseNumber([]byte(s))
 }
 
-// parsePreciseNumber validates s as a base-10 decimal and returns a
+// parsePreciseNumber validates data via decimal.NewFromString and returns a
 // PreciseNumber holding the original string and a derived float64 cache.
 //
-// decimal.NewFromString is used as the source of truth: it rejects
-// hexadecimal, scientific notation, NaN and Inf, all of which strconv.ParseFloat
-// would happily accept and which are never legitimate exchange wire values.
+// Values whose magnitude overflows float64 (InexactFloat64 returns Inf) are
+// rejected so the cached float is always finite.
 func parsePreciseNumber(data []byte) (PreciseNumber, error) {
 	s := string(data)
 	d, err := decimal.NewFromString(s)
@@ -55,8 +52,8 @@ func parsePreciseNumber(data []byte) (PreciseNumber, error) {
 		return PreciseNumber{}, fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
 	}
 	val := d.InexactFloat64()
-	if math.IsNaN(val) || math.IsInf(val, 0) {
-		return PreciseNumber{}, fmt.Errorf("%w: non-finite %s", errInvalidPreciseNumberValue, data)
+	if math.IsInf(val, 0) {
+		return PreciseNumber{}, fmt.Errorf("%w: magnitude overflows float64: %s", errInvalidPreciseNumberValue, data)
 	}
 	return PreciseNumber{raw: s, val: val}, nil
 }
@@ -75,14 +72,14 @@ func (p *PreciseNumber) UnmarshalJSON(data []byte) error {
 	case 'n': // null
 		*p = PreciseNumber{}
 		return nil
-	case 't', 'f':
-		return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
 	case '"':
 		if len(data) < 2 || data[len(data)-1] != '"' {
 			return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
 		}
 		data = data[1 : len(data)-1]
 	default:
+		// Bare tokens (true, false, garbage) fail here: the first byte is
+		// neither '-' nor a digit. decimal.NewFromString validates the rest.
 		if c != '-' && (c < '0' || c > '9') {
 			return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
 		}
@@ -108,11 +105,7 @@ func (p PreciseNumber) MarshalJSON() ([]byte, error) {
 	if p.raw == "" {
 		return []byte(`""`), nil
 	}
-	out := make([]byte, 0, len(p.raw)+2)
-	out = append(out, '"')
-	out = append(out, p.raw...)
-	out = append(out, '"')
-	return out, nil
+	return []byte(`"` + p.raw + `"`), nil
 }
 
 // Float64 returns the cached float64 form. Suitable for sorting, comparisons
@@ -124,6 +117,10 @@ func (p PreciseNumber) Float64() float64 {
 // Int64 returns the integer value parsed from the original string, truncated
 // toward zero. For values that originated as integers (trade IDs, integer
 // timestamps) this returns the exact value with no float round-trip.
+//
+// If the underlying value is outside the int64 range, the result is the
+// truncated low-order bits and is not meaningful — callers handling
+// potentially large values should use Decimal instead.
 //
 // raw is always a valid base-10 decimal here because every entry path runs
 // it through [parsePreciseNumber]; the zero value short-circuits to 0.
--- a/types/precise_number.go
+++ b/types/precise_number.go
@@ -3,7 +3,7 @@ package types
 import (
 	"errors"
 	"fmt"
-	"strconv"
+	"math"
 
 	"github.com/shopspring/decimal"
 )
@@ -32,22 +32,39 @@ type PreciseNumber struct {
 	val float64 // parsed cache for fast Float64()
 }
 
-// NewPreciseNumberFromString constructs a PreciseNumber from a decimal string
-// without any precision loss. The string must be a valid base-10 number.
+// NewPreciseNumberFromString constructs a PreciseNumber from a base-10 decimal
+// string without precision loss. Empty input, hexadecimal, scientific notation,
+// NaN and Inf are rejected.
 func NewPreciseNumberFromString(s string) (PreciseNumber, error) {
 	if s == "" {
-		return PreciseNumber{}, nil
+		return PreciseNumber{}, fmt.Errorf("%w: empty string", errInvalidPreciseNumberValue)
 	}
-	val, err := strconv.ParseFloat(s, 64)
+	return parsePreciseNumber([]byte(s))
+}
+
+// parsePreciseNumber validates s as a base-10 decimal and returns a
+// PreciseNumber holding the original string and a derived float64 cache.
+//
+// decimal.NewFromString is used as the source of truth: it rejects
+// hexadecimal, scientific notation, NaN and Inf, all of which strconv.ParseFloat
+// would happily accept and which are never legitimate exchange wire values.
+func parsePreciseNumber(data []byte) (PreciseNumber, error) {
+	s := string(data)
+	d, err := decimal.NewFromString(s)
 	if err != nil {
-		return PreciseNumber{}, fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, s)
+		return PreciseNumber{}, fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
+	}
+	val, _ := d.Float64()
+	if math.IsNaN(val) || math.IsInf(val, 0) {
+		return PreciseNumber{}, fmt.Errorf("%w: non-finite %s", errInvalidPreciseNumberValue, data)
 	}
 	return PreciseNumber{raw: s, val: val}, nil
 }
 
 // UnmarshalJSON implements json.Unmarshaler. It accepts both JSON strings
 // ("1337.37") and JSON numbers (1337.37) and preserves the original text
-// verbatim for downstream Decimal()/Int64() calls.
+// verbatim for downstream Decimal()/Int64() calls. null, an empty string and
+// a missing/zero-length payload all parse to the zero value.
 func (p *PreciseNumber) UnmarshalJSON(data []byte) error {
 	if len(data) == 0 {
 		*p = PreciseNumber{}
@@ -76,14 +93,11 @@ func (p *PreciseNumber) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
-	s := string(data)
-	val, err := strconv.ParseFloat(s, 64)
+	parsed, err := parsePreciseNumber(data)
 	if err != nil {
-		return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
+		return err
 	}
-
-	p.raw = s
-	p.val = val
+	*p = parsed
 	return nil
 }
 
@@ -101,31 +115,24 @@ func (p PreciseNumber) MarshalJSON() ([]byte, error) {
 	return out, nil
 }
 
-// Float64 returns the cached float64 form. Suitable for sorting, comparisons,
+// Float64 returns the cached float64 form. Suitable for sorting, comparisons
 // and approximate computation. For exact arithmetic use Decimal.
 func (p PreciseNumber) Float64() float64 {
 	return p.val
 }
 
-// Int64 returns the integer value parsed from the original string. If the
-// original value contains a fractional part the fractional part is truncated
-// toward zero (matching [Number.Int64]). For values that originated as
-// integers (trade IDs, integer timestamps) this returns the exact value with
-// no float round-trip.
+// Int64 returns the integer value parsed from the original string, truncated
+// toward zero. For values that originated as integers (trade IDs, integer
+// timestamps) this returns the exact value with no float round-trip.
+//
+// raw is always a valid base-10 decimal here because every entry path runs
+// it through [parsePreciseNumber]; the zero value short-circuits to 0.
 func (p PreciseNumber) Int64() int64 {
 	if p.raw == "" {
 		return 0
 	}
-	if i, err := strconv.ParseInt(p.raw, 10, 64); err == nil {
-		return i
-	}
-	// Fractional value; fall back to truncating via Decimal to avoid the
-	// float-rounding behaviour exhibited by `int64(float64(p.val))` near
-	// representation boundaries.
-	if d, err := decimal.NewFromString(p.raw); err == nil {
-		return d.Truncate(0).IntPart()
-	}
-	return int64(p.val)
+	d, _ := decimal.NewFromString(p.raw)
+	return d.IntPart()
 }
 
 // Decimal returns the exact decimal value parsed from the original string,
@@ -134,10 +141,8 @@ func (p PreciseNumber) Decimal() decimal.Decimal {
 	if p.raw == "" {
 		return decimal.Zero
 	}
-	if d, err := decimal.NewFromString(p.raw); err == nil {
-		return d
-	}
-	return decimal.NewFromFloat(p.val)
+	d, _ := decimal.NewFromString(p.raw)
+	return d
 }
 
 // String returns the original numeric string, or "0" when the value is the
--- a/types/precise_number.go
+++ b/types/precise_number.go
@@ -76,12 +76,13 @@ func (p *PreciseNumber) UnmarshalJSON(data []byte) error {
 		return nil
 	}
 
-	val, err := strconv.ParseFloat(string(data), 64)
+	s := string(data)
+	val, err := strconv.ParseFloat(s, 64)
 	if err != nil {
 		return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
 	}
 
-	p.raw = string(data)
+	p.raw = s
 	p.val = val
 	return nil
 }
@@ -148,7 +149,13 @@ func (p PreciseNumber) String() string {
 	return p.raw
 }
 
-// IsZero reports whether the value is the zero value.
+// IsZero reports whether the value is the uninitialized zero value.
+//
+// Note: this returns true only for the uninitialized struct, not for an
+// explicit "0" parsed from the wire. The distinction matters when
+// json.Marshal honours the IsZero() method for `omitempty` (Go 1.24+):
+// an explicit "0" must round-trip through JSON for accounting fields,
+// whereas an unset field is correctly omitted.
 func (p PreciseNumber) IsZero() bool {
-	return p.raw == "" || p.val == 0
+	return p.raw == ""
 }
--- a/types/precise_number.go
+++ b/types/precise_number.go
@@ -1,0 +1,154 @@
+package types
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+
+	"github.com/shopspring/decimal"
+)
+
+var errInvalidPreciseNumberValue = errors.New("invalid value for PreciseNumber type")
+
+// PreciseNumber is a numeric type for JSON values that need lossless decimal
+// representation alongside a fast floating-point form.
+//
+// It stores the original numeric string as received from the wire and parses
+// a float64 cache for cheap comparisons and sorting (e.g. order book level
+// alignment). Decimal() and Int64() then return the value parsed from the
+// preserved string rather than from the float, avoiding the precision loss
+// that occurs in the existing [Number] type when [Number.Decimal] converts
+// from float64.
+//
+// PreciseNumber is the right choice for fields that flow into accounting,
+// reconciliation, settlement, or anything where the exact decimal value
+// from the exchange must be reproduced. For market-data fields where the
+// value is only used for sorting and rough computation, [Number] remains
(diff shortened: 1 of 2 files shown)


types/precise_number.go
```diff
@@ -0,0 +1,154 @@
...
+	if p.raw == "" {
+		return "0"
+	}
+	return p.raw
+}
+
+// IsZero reports whether the value is the zero value.
+func (p PreciseNumber) IsZero() bool {
+	return p.raw == "" || p.val == 0
+}
```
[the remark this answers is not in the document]
valli0x: Fixed in 8220481d — IsZero now returns true only for uninitialized zero value (raw == ""), preventing omitempty data loss for "0".

types/precise_number.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+
+	val, err := strconv.ParseFloat(string(data), 64)
+	if err != nil {
+		return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
+	}
+
+	p.raw = string(data)
+	p.val = val
+	return nil
```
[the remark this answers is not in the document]
valli0x: Fixed in 8220481d — single string allocation reused for ParseFloat and raw.

types/precise_number_test.go
```diff
@@ -0,0 +1,190 @@
...
+	t.Parallel()
+
+	p, err := NewPreciseNumberFromString("1.5")
+	assert.NoError(t, err)
+	assert.Equal(t, "1.5", p.String())
+	assert.Equal(t, 1.5, p.Float64())
+
+	zero, err := NewPreciseNumberFromString("")
+	assert.NoError(t, err)
+	assert.True(t, zero.IsZero())
```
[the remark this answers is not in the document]
valli0x: Fixed in 8220481d — added assertion that NewPreciseNumberFromString("0") returns IsZero() == false

gloriousCode:
Very nice, thank you for this. I've had it on my mind to make types.Number immutable and have the original string in there for awhile. Making it a seperate type is good and minimises the amount of changes needed to support it. It's a much faster and accurate way of handling high decimals in orderbooks, order sizing etc

Will review soon.

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+	raw string  // original numeric string from the source; "" means zero
+	val float64 // parsed cache for fast Float64()
+}
+
+// NewPreciseNumberFromString constructs a PreciseNumber from a decimal string
+// without any precision loss. The string must be a valid base-10 number.
+func NewPreciseNumberFromString(s string) (PreciseNumber, error) {
+	if s == "" {
+		return PreciseNumber{}, nil
+	}
```
shazbert: This looks like it should be an error. As the function shouldn't be called with zero string value.

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+	val, err := strconv.ParseFloat(s, 64)
+	if err != nil {
+		return PreciseNumber{}, fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, s)
+	}
+	return PreciseNumber{raw: s, val: val}, nil
```
shazbert, line 58: Suggestion: Should reject non-finite float values after parsing ie `math.IsNan` and `math.IsInf`

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+	val float64 // parsed cache for fast Float64()
+}
+
+// NewPreciseNumberFromString constructs a PreciseNumber from a decimal string
+// without any precision loss. The string must be a valid base-10 number.
+func NewPreciseNumberFromString(s string) (PreciseNumber, error) {
+	if s == "" {
+		return PreciseNumber{}, nil
+	}
+	val, err := strconv.ParseFloat(s, 64)
```
shazbert: suggestion: parsefloat accepts hex and can't guarantee base=10 so can pad it with a decimal.NewFromString above maybe?

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+		return 0
+	}
+	if i, err := strconv.ParseInt(p.raw, 10, 64); err == nil {
+		return i
+	}
+	// Fractional value; fall back to truncating via Decimal to avoid the
+	// float-rounding behaviour exhibited by `int64(float64(p.val))` near
+	// representation boundaries.
+	if d, err := decimal.NewFromString(p.raw); err == nil {
+		return d.Truncate(0).IntPart()
```
shazbert:
```suggestion
		return d.IntPart()
```

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+func (p PreciseNumber) Float64() float64 {
+	return p.val
+}
+
+// Int64 returns the integer value parsed from the original string. If the
+// original value contains a fractional part the fractional part is truncated
+// toward zero (matching [Number.Int64]). For values that originated as
+// integers (trade IDs, integer timestamps) this returns the exact value with
+// no float round-trip.
+func (p PreciseNumber) Int64() int64 {
```
shazbert, line 127: I am thinking in future (not now) we return this value with an error as well. Along with decimal. Because we need to ensure that we are using the most precise value available.

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+	return PreciseNumber{raw: s, val: val}, nil
+}
+
+// UnmarshalJSON implements json.Unmarshaler. It accepts both JSON strings
+// ("1337.37") and JSON numbers (1337.37) and preserves the original text
+// verbatim for downstream Decimal()/Int64() calls.
+func (p *PreciseNumber) UnmarshalJSON(data []byte) error {
+	if len(data) == 0 {
+		*p = PreciseNumber{}
+		return nil
```
shazbert, line 68: This needs coverage

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+	case 't', 'f':
+		return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
+	case '"':
+		if len(data) < 2 || data[len(data)-1] != '"' {
+			return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
+		}
+		data = data[1 : len(data)-1]
+	default:
+		if c != '-' && (c < '0' || c > '9') {
+			return fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
```
shazbert, line 84: This needs coverage

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+	if i, err := strconv.ParseInt(p.raw, 10, 64); err == nil {
+		return i
+	}
+	// Fractional value; fall back to truncating via Decimal to avoid the
+	// float-rounding behaviour exhibited by `int64(float64(p.val))` near
+	// representation boundaries.
+	if d, err := decimal.NewFromString(p.raw); err == nil {
+		return d.Truncate(0).IntPart()
+	}
+	return int64(p.val)
```
shazbert: this needs coverage

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+// Decimal returns the exact decimal value parsed from the original string,
+// avoiding the float64 round-trip that loses precision in [Number.Decimal].
+func (p PreciseNumber) Decimal() decimal.Decimal {
+	if p.raw == "" {
+		return decimal.Zero
+	}
+	if d, err := decimal.NewFromString(p.raw); err == nil {
+		return d
+	}
+	return decimal.NewFromFloat(p.val)
```
shazbert: This needs coverage

types/precise_number.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// String returns the original numeric string, or "0" when the value is the
+// zero value.
+func (p PreciseNumber) String() string {
+	if p.raw == "" {
+		return "0"
```
shazbert, line 149: This needs coverage

types/precise_number_test.go
```diff
@@ -0,0 +1,197 @@
...
+// TestPreciseNumberDecimal verifies Decimal() returns the exact decimal value
+// parsed from the original string, without going through float64.
+func TestPreciseNumberDecimal(t *testing.T) {
+	t.Parallel()
+
+	var p PreciseNumber
+	err := p.UnmarshalJSON([]byte(`"71428.12345678"`))
+	assert.NoError(t, err)
+
+	expected, _ := decimal.NewFromString("71428.12345678")
```
shazbert: Could you please check all errors in tests as well. Can just use require.NoError()

types/precise_number_test.go
```diff
@@ -0,0 +1,197 @@
...
+	const wireValue = "71428.12345678901234"
+
+	var n Number
+	require := assert.New(t)
+	require.NoError(n.UnmarshalJSON([]byte(`"` + wireValue + `"`)))
+
+	var p PreciseNumber
+	require.NoError(p.UnmarshalJSON([]byte(`"` + wireValue + `"`)))
+
+	expected, _ := decimal.NewFromString(wireValue)
```
shazbert: here as well

types/precise_number_test.go
```diff
@@ -0,0 +1,197 @@
...
+	assert.ErrorIs(t, err, errInvalidPreciseNumberValue)
+}
+
+// BenchmarkPreciseNumberUnmarshalJSON measures the cost of UnmarshalJSON for
+// a typical exchange string value.
+func BenchmarkPreciseNumberUnmarshalJSON(b *testing.B) {
+	var p PreciseNumber
+	for b.Loop() {
+		_ = p.UnmarshalJSON([]byte(`"0.04200074"`))
+	}
```
shazbert, line 242:
```suggestion
	for b.Loop() {
		if err := p.UnmarshalJSON([]byte(`"0.04200074"`)); err != nil {
			require.NoError(b, err)
		}
	}
```

types/precise_number_test.go
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// BenchmarkPreciseNumberDecimal measures the cost of Decimal() so we can
+// compare against [BenchmarkNumberDecimalConversion] in number_test.go.
+func BenchmarkPreciseNumberDecimal(b *testing.B) {
+	p, _ := NewPreciseNumberFromString("0.04200074")
+	for b.Loop() {
+		_ = p.Decimal()
+	}
```
shazbert:
```suggestion
	p, err := NewPreciseNumberFromString("0.04200074")
	require.NoError(b, err)
	for b.Loop() {
		if !p.Decimal().IsPositive() {
			b.Fatal("unexpected non-positive decimal")
		}
	}
```

shazbert (changes requested):
Cool PR! Just some things I noticed on first pass.

types/precise_number.go
```diff
@@ -0,0 +1,166 @@
...
+// decimal.NewFromString is used as the source of truth: it rejects
+// hexadecimal, scientific notation, NaN and Inf, all of which strconv.ParseFloat
+// would happily accept and which are never legitimate exchange wire values.
+func parsePreciseNumber(data []byte) (PreciseNumber, error) {
+	s := string(data)
+	d, err := decimal.NewFromString(s)
+	if err != nil {
+		return PreciseNumber{}, fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
+	}
+	val, _ := d.Float64()
```
shazbert:
```suggestion
	val := d.InexactFloat64()
```

If you are not using the error 🤷

types/precise_number_test.go
```diff
@@ -0,0 +1,260 @@
...
+	require.NoError(t, p.UnmarshalJSON([]byte(`1337.37`)))
+	assert.Equal(t, 1337.37, p.Float64())
+	assert.Equal(t, "1337.37", p.String())
+
+	require.NoError(t, p.UnmarshalJSON([]byte(`-7.25`)), "leading minus accepted")
+	assert.Equal(t, "-7.25", p.String())
+
+	// Each entry exercises a distinct invalid-input branch in UnmarshalJSON.
+	for _, in := range []string{
+		`"MEOW"`,   // quoted non-numeric — rejected by parsePreciseNumber
```
shazbert: 🐱 🚀

types/precise_number_test.go
```diff
@@ -0,0 +1,260 @@
...
+
+// TestPreciseNumberRoundTrip verifies the marshal/unmarshal cycle preserves
+// the exact original digit string — the key advantage over Number which
+// reformats through float64 and loses trailing precision.
+func TestPreciseNumberRoundTrip(t *testing.T) {
+	t.Parallel()
+
+	// 17 significant digits — beyond what float64 reliably round-trips
+	// through ParseFloat/FormatFloat
+	const high = "71428.12345678901234"
```
shazbert: suggestion: you could sanity check this but up to you if you can be bothered. 

Also this can be a package level const for tests and shared

types/precise_number_test.go
```diff
@@ -0,0 +1,260 @@
...
+	assert.Equal(t, 1.5, p.Float64())
+
+	// Numerically zero but explicitly set: must round-trip through JSON
+	// (i.e. must not be IsZero) so omitempty doesn't drop accounting "0".
+	notZero, err := NewPreciseNumberFromString("0")
+	require.NoError(t, err)
+	assert.False(t, notZero.IsZero(),
+		"explicit \"0\" must not be IsZero; otherwise omitempty silently drops accounting fields")
+
+	// Empty constructor input is now an error rather than a silent zero
```
shazbert: commentary across this test file can be in the assert message param

types/precise_number_test.go
```diff
@@ -0,0 +1,260 @@
...
+func BenchmarkPreciseNumberDecimal(b *testing.B) {
+	p, err := NewPreciseNumberFromString("0.04200074")
+	require.NoError(b, err)
+	for b.Loop() {
+		if !p.Decimal().IsPositive() {
+			b.Fatal("unexpected non-positive decimal")
+		}
+	}
+}
+
```
shazbert: classic linter issue

shazbert (changes requested):
Thanks! Just some minor things then looks good to go! 💃

types/precise_number.go
valli0x: Fixed in 63f17d26 — empty string now returns an error. Callers wanting the zero value can use PreciseNumber{} directly.

types/precise_number.go
```diff
@@ -0,0 +1,166 @@
...
+//
+// Zero value behaves as the number 0.
+type PreciseNumber struct {
+	raw string  // original numeric string from the source; "" means zero
+	val float64 // parsed cache for fast Float64()
+}
+
+// NewPreciseNumberFromString constructs a PreciseNumber from a base-10 decimal
+// string without precision loss. Empty input, hexadecimal, scientific notation,
+// NaN and Inf are rejected.
```
cranktakular: This is not fully true. As written right now, it does accept scientific notation, and rejects some notation accepted in Go literals (i.e. the inclusion of underscores).

Keep the claim to what is strictly true; it only accepts inputs decimal.NewFromString() accepts, except for values which are too large to be represented by a float64, which are rejected.

types/precise_number.go
```diff
@@ -0,0 +1,166 @@
...
+func NewPreciseNumberFromString(s string) (PreciseNumber, error) {
+	if s == "" {
+		return PreciseNumber{}, fmt.Errorf("%w: empty string", errInvalidPreciseNumberValue)
```
cranktakular: This is kind of unnecessary; decimal.NewFromString() rejects empty strings anyway.

types/precise_number.go
```diff
@@ -0,0 +1,166 @@
...
+// hexadecimal, scientific notation, NaN and Inf, all of which strconv.ParseFloat
+// would happily accept and which are never legitimate exchange wire values.
+func parsePreciseNumber(data []byte) (PreciseNumber, error) {
+	s := string(data)
+	d, err := decimal.NewFromString(s)
+	if err != nil {
+		return PreciseNumber{}, fmt.Errorf("%w: %s", errInvalidPreciseNumberValue, data)
+	}
+	val := d.InexactFloat64()
+	if math.IsNaN(val) || math.IsInf(val, 0) {
```
cranktakular: Why is this NaN check included? decimal.NewFromString() doesn't accept NaN values, and as far as I can tell, InexactFloat64() doesn't return NaN values.

types/precise_number.go
```diff
@@ -0,0 +1,166 @@
...
+// string as a JSON string. The zero value marshals to `""` to match the
+// existing [Number.MarshalJSON] behaviour.
+func (p PreciseNumber) MarshalJSON() ([]byte, error) {
+	if p.raw == "" {
+		return []byte(`""`), nil
+	}
+	out := make([]byte, 0, len(p.raw)+2)
+	out = append(out, '"')
+	out = append(out, p.raw...)
+	out = append(out, '"')
```
cranktakular: You don't need all these lines, you can just do

> out := []byte(`"` + p.raw + `"`)

That's mangled a bit by github's markdown implementation, but hopefully you can still see what I'm getting at.

types/precise_number.go
```diff
@@ -0,0 +1,166 @@
...
+func (p PreciseNumber) Float64() float64 {
+	return p.val
+}
+
+// Int64 returns the integer value parsed from the original string, truncated
+// toward zero. For values that originated as integers (trade IDs, integer
+// timestamps) this returns the exact value with no float round-trip.
+//
+// raw is always a valid base-10 decimal here because every entry path runs
+// it through [parsePreciseNumber]; the zero value short-circuits to 0.
```
cranktakular, line 126: I think it might be worth flagging that if the original value is out of the int64 range, that this could return unexpected values.

types/precise_number_test.go
```diff
@@ -0,0 +1,266 @@
...
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/shopspring/decimal"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// highPrecisionValue is a decimal string with 17 significant digits — beyond
```
cranktakular: This doesn't have 17 significant digits, it has 19.

types/precise_number_test.go
```diff
@@ -0,0 +1,266 @@
...
+// exceeds float64's round-trip precision, so the tests that rely on it are
+// exercising the property they claim to. If a float64 round-trip reproduced
+// the string exactly, those tests would pass trivially and prove nothing.
+func TestHighPrecisionValueSanity(t *testing.T) {
+	t.Parallel()
+	f, err := strconv.ParseFloat(highPrecisionValue, 64)
+	require.NoError(t, err)
+	roundTripped := strconv.FormatFloat(f, 'f', -1, 64)
+	assert.NotEqual(t, highPrecisionValue, roundTripped,
+		"highPrecisionValue must lose precision through float64 for the precision tests to be meaningful")
```
cranktakular: This, and many other messages in this test file, don't follow our [standards on require/assert messages](https://github.com/thrasher-corp/gocryptotrader/blob/master/docs/CODING_GUIDELINES.md#assertion-usage).

types/precise_number.go
```diff
@@ -0,0 +1,166 @@
...
+	if len(data) == 0 {
+		*p = PreciseNumber{}
+		return nil
+	}
+
+	switch c := data[0]; c {
+	case 'n': // null
+		*p = PreciseNumber{}
+		return nil
+	case 't', 'f':
```
cranktakular: I know that we have the same setup in number.go, but I don't think I reviewed that. And now that I see it here, I think this check should be removed. Such cases are already handled, with the exact same error, by the character check in the default case.

cranktakular (changes requested):
Thanks for the contribution!

I ran out of time to check the whole test file, but I should get to it next week. And in the meantime, I have some changes I'd like to see made.

types/precise_number_test.go
```diff
@@ -0,0 +1,266 @@
...
+	require.NoError(t, p.UnmarshalJSON([]byte(`"`+highPrecisionValue+`"`)))
+
+	data, err := p.MarshalJSON()
+	require.NoError(t, err)
+	assert.Equal(t, `"`+highPrecisionValue+`"`, string(data),
+		"PreciseNumber must round-trip the original digit string")
+}
+
+// TestPreciseNumberDecimal verifies Decimal() returns the exact decimal value
+// parsed from the original string, without going through float64.
```
cranktakular, line 107: This test does not properly exercise that. If I alter the logic to go through float64, this still passes, since it doesn't use a sufficiently high precision value.

cranktakular, line 107: This test, as-is, doesn't actually verify that; the value provided is not long enough to exceed float64's precision, so if the logic is replaced to go through float64, it won't leave an error.

Easiest fix is just to slot in highPrecisionValue here.

types/precise_number_test.go
```diff
@@ -0,0 +1,266 @@
...
+
+	p, err := NewPreciseNumberFromString("1.5")
+	require.NoError(t, err)
+	assert.Equal(t, "1.5", p.String())
+	assert.Equal(t, 1.5, p.Float64())
+
+	notZero, err := NewPreciseNumberFromString("0")
+	require.NoError(t, err)
+	assert.False(t, notZero.IsZero(),
+		"explicitly set \"0\" must not be IsZero, otherwise omitempty silently drops accounting fields")
```
cranktakular: Seems a bit redundant to test this both here, and in TestPreciseNumberIsZero.

types/precise_number_test.go
```diff
@@ -0,0 +1,266 @@
...
+	var p PreciseNumber
+	for b.Loop() {
+		if err := p.UnmarshalJSON([]byte(`"0.04200074"`)); err != nil {
+			require.NoError(b, err)
+		}
+	}
+}
+
+// BenchmarkPreciseNumberDecimal measures the cost of Decimal() so we can
+// compare against [BenchmarkNumberDecimalConversion] in number_test.go.
```
cranktakular: BenchmarkNumberDecimalConversion does not exist, so that part of the comment should be removed.

On top of that, these benchmarks should include their results on your machine in a comment above them. See the benchmarks in number_test.go for an example.

cranktakular (changes requested):
Finished looking through this PR, but there are a few more changes I must request.

types/precise_number_test.go
```diff
@@ -0,0 +1,255 @@
...
+// uninitialized struct, never for an explicit "0" — the property that
+// protects accounting fields from omitempty data loss.
+func TestPreciseNumberIsZero(t *testing.T) {
+	t.Parallel()
+
+	assert.True(t, PreciseNumber{}.IsZero(), "uninitialized value should be IsZero")
+
+	explicit, err := NewPreciseNumberFromString("0")
+	require.NoError(t, err, "NewPreciseNumberFromString must not error")
+	assert.False(t, explicit.IsZero(), "explicitly set \"0\" should not be IsZero")
```
cranktakular: Message wording feels a little strange. Consider rewording to `IsZero should return false for explicitly set \"0\"`

cranktakular (approved):
Looks great! There's one minor wording change I'd want, but aside from that I have tested and acknowledged the changes as acceptable.

shazbert (approved):
Nothing more from me. Thanks!

gloriousCode (changes requested):
Lots of words for what amounts to a comment change, but thought I'd just explain myself

types/precise_number.go
```diff
@@ -0,0 +1,163 @@
...
+		return "0"
+	}
+	return p.raw
+}
+
+// IsZero reports whether the value is the uninitialized zero value.
+//
+// Note: this returns true only for the uninitialized struct, not for an
+// explicit "0" parsed from the wire. The distinction matters when
+// json.Marshal honours the IsZero() method for `omitempty` (Go 1.24+):
```
gloriousCode: IsZero isnt called for omitempty

But we unmarshal data rather than marshal fields with PreciseNumber.

So actually I'm not worried, but it would be good to not state this

gloriousCode (commented):
Im happy with this, but if you could just update a comment we're good

types/precise_number.go
shazbert: @valli0x Bump

thrasher-:
@codex please review

types/precise_number.go
thrasher-: Fixed in 46eee4e8c — same claim was in the test doc, got that too.

types/precise_number_test.go
thrasher-: Reworded in 46eee4e8c — went with "for an explicit \"0\"", same idea. Did the other three messages in that test while I was there.

thrasher- (approved):
Thank you for this epic PR @valli0x !

shazbert (approved):
changes ACK!

PR #2270, opened 2026-06-17
exchanges/OKX, GateIO: Fix premium kline params and order limits

thrasher-:
## Summary
- Fix GateIO `PremiumIndexKLine` time parameter handling so supplied `from`/`to` values are sent instead of zero-time Unix values.
- Preserve omitted `from`/`to` bounds so callers control whether GateIO receives explicit bounds, `limit`, or neither.
- Simplify `TestPremiumIndexKLine` coverage into live request and validation subtests without the local HTTP mock.
- Skip inactive or empty OKX instruments when loading order execution limits, matching live tradable instrument handling and avoiding preopen empty `instId` rows.

## Tests
- `go test ./exchanges/gateio -run 'TestPremiumIndexKLine' -count=1`
- `go test ./exchanges/gateio -count=1`
- `go test ./exchanges/okx -run 'TestUpdateOrderExecutionLimits|TestLoadInstrumentOrderExecutionLimitsSkipsInactiveAndEmptyInstruments' -count=1`
- `go test ./exchanges/okx -count=1`
- `golangci-lint run ./exchanges/gateio`
- `golangci-lint run ./exchanges/okx`
- `make lint`
- `codespell`
- `./scripts/misc_checks.sh`
- `git diff --check`

### DIFF
--- a/exchanges/gateio/gateio.go
+++ b/exchanges/gateio/gateio.go
@@ -1937,24 +1937,35 @@ func (e *Exchange) PremiumIndexKLine(ctx context.Context, settleCurrency currenc
 	if contract.IsEmpty() {
 		return nil, currency.ErrCurrencyPairEmpty
 	}
+	intervalString, err := getIntervalString(interval)
+	if err != nil {
+		return nil, err
+	}
+	from, to = defaultPremiumIndexKLineTimeRange(from, to, time.Now(), interval)
 	params := url.Values{}
 	params.Set("contract", contract.String())
-	if from.IsZero() {
+	if !from.IsZero() {
 		params.Set("from", strconv.FormatInt(from.Unix(), 10))
 	}
-	if to.IsZero() {
+	if !to.IsZero() {
 		params.Set("to", strconv.FormatInt(to.Unix(), 10))
 	}
 	if limit > 0 {
 		params.Set("limit", strconv.FormatInt(limit, 10))
 	}
-	intervalString, err := getIntervalString(interval)
-	if err != nil {
-		return nil, err
-	}
 	params.Set("interval", intervalString)
 	var resp []FuturesPremiumIndexKLineResponse
 	return resp, e.SendHTTPRequest(ctx, exchange.RestSpot, publicPremiumIndexEPL, common.EncodeURLValues(futuresPath+settleCurrency.Item.Lower+"/premium_index", params), &resp)
+}
+
+func defaultPremiumIndexKLineTimeRange(from, to, now time.Time, interval kline.Interval) (adjustedFrom, adjustedTo time.Time) {
+	if to.IsZero() {
+		to = now.UTC()
+	}
+	if from.IsZero() {
+		from = to.Add(-2 * interval.Duration())
+	}
+	return from, to
 }
 
 // GetFuturesTickers retrieves futures ticker information for a specific settle and contract info.
--- a/exchanges/gateio/gateio.go
+++ b/exchanges/gateio/gateio.go
@@ -1941,7 +1941,6 @@ func (e *Exchange) PremiumIndexKLine(ctx context.Context, settleCurrency currenc
 	if err != nil {
 		return nil, err
 	}
-	from, to = defaultPremiumIndexKLineTimeRange(from, to, time.Now(), interval)
 	params := url.Values{}
 	params.Set("contract", contract.String())
 	if !from.IsZero() {
@@ -1956,16 +1955,6 @@ func (e *Exchange) PremiumIndexKLine(ctx context.Context, settleCurrency currenc
 	params.Set("interval", intervalString)
 	var resp []FuturesPremiumIndexKLineResponse
 	return resp, e.SendHTTPRequest(ctx, exchange.RestSpot, publicPremiumIndexEPL, common.EncodeURLValues(futuresPath+settleCurrency.Item.Lower+"/premium_index", params), &resp)
-}
-
-func defaultPremiumIndexKLineTimeRange(from, to, now time.Time, interval kline.Interval) (adjustedFrom, adjustedTo time.Time) {
-	if to.IsZero() {
-		to = now.UTC()
-	}
-	if from.IsZero() {
-		from = to.Add(-2 * interval.Duration())
-	}
-	return from, to
 }
 
 // GetFuturesTickers retrieves futures ticker information for a specific settle and contract info.
--- a/exchanges/gateio/gateio_test.go
+++ b/exchanges/gateio/gateio_test.go
@@ -5,6 +5,9 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
 	"os"
 	"slices"
 	"strconv"
@@ -22,6 +25,7 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/encoding/json"
 	"github.com/thrasher-corp/gocryptotrader/exchange/accounts"
 	"github.com/thrasher-corp/gocryptotrader/exchange/websocket"
+	exchange "github.com/thrasher-corp/gocryptotrader/exchanges"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/fundingrate"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/futures"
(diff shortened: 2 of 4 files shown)


exchanges/gateio/gateio.go
```diff
@@ -1937,26 +1937,37 @@ func (e *Exchange) PremiumIndexKLine(ctx context.Context, settleCurrency currenc
 	if contract.IsEmpty() {
 		return nil, currency.ErrCurrencyPairEmpty
 	}
+	intervalString, err := getIntervalString(interval)
+	if err != nil {
+		return nil, err
+	}
+	from, to = defaultPremiumIndexKLineTimeRange(from, to, time.Now(), interval)
```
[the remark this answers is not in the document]
thrasher-: Jarvis: Thanks, good catch. Addressed in https://github.com/thrasher-corp/gocryptotrader/commit/9013a6dfd6bbb578af2b139ff9f2232ea69dede8 by removing automatic `from`/`to` defaulting. `PremiumIndexKLine` now preserves caller-supplied params exactly, and tests cover omitted bounds, limit-only, explicit bounds, and validation errors.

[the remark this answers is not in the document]
poyatabrizi: Jervis, you are a good lad.

exchanges/gateio/gateio_test.go
```diff
@@ -967,12 +971,96 @@ func TestGetFuturesCandlesticks(t *testing.T) {
...
+	to := time.Now().UTC()
+	from := to.Add(-2 * kline.OneWeek.Duration())
+	_, err := e.PremiumIndexKLine(t.Context(), currency.BTC, getPair(t, asset.CoinMarginedFutures), from, to, 0, kline.OneWeek)
 	assert.NoError(t, err, "PremiumIndexKLine should not error for CoinMarginedFutures")
-	_, err = e.PremiumIndexKLine(t.Context(), currency.USDT, getPair(t, asset.USDTMarginedFutures), time.Time{}, time.Time{}, 0, kline.OneWeek)
+	_, err = e.PremiumIndexKLine(t.Context(), currency.USDT, getPair(t, asset.USDTMarginedFutures), from, to, 0, kline.OneWeek)
 	assert.NoError(t, err, "PremiumIndexKLine should not error for USDTMarginedFutures")
 }
 
+func TestPremiumIndexKLineRequestTimeRange(t *testing.T) {
```
gloriousCode: Thank you for the function fix!

However
1) The function being tested is not very complicated and doesn't warrant a large regression test with a mock server
2) There has been no suggestion that the API endpoint is prone to failure enough to warrant a mock server and it can be dropped
3) I don't think this needed to be its own separate test and should have been a t.Run scenario

I think this is much more appropriate
```go
func TestPremiumIndexKLine(t *testing.T) {
	t.Parallel()
	to := time.Now().UTC()
	from := to.Add(-2 * kline.OneWeek.Duration())
	e := new(Exchange)

	t.Run("live request", func(t *testing.T) {
		t.Parallel()
		contract := getPair(t, asset.USDTMarginedFutures)
		_, err := e.PremiumIndexKLine(t.Context(), currency.BTC, getPair(t, asset.CoinMarginedFutures), from, to, 0, kline.OneWeek)
		assert.NoError(t, err, "PremiumIndexKLine should not error for CoinMarginedFutures")
		_, err = e.PremiumIndexKLine(t.Context(), currency.USDT, contract, from, to, 0, kline.OneWeek)
		assert.NoError(t, err, "PremiumIndexKLine should not error for USDTMarginedFutures")
		_, err = e.PremiumIndexKLine(t.Context(), currency.USDT, contract, time.Time{}, time.Time{}, 1, kline.OneWeek)
		assert.NoError(t, err, "PremiumIndexKLine should not error with a limit set")
		_, err = e.PremiumIndexKLine(t.Context(), currency.USDT, contract, from, to, 1, kline.OneWeek)
		assert.NoError(t, err, "PremiumIndexKLine should not error with supplied bounds and limit")
	})

	t.Run("validation", func(t *testing.T) {
		t.Parallel()
		contract := currency.NewBTCUSDT()
		_, err := e.PremiumIndexKLine(t.Context(), currency.Code{}, contract, time.Time{}, time.Time{}, 0, kline.OneWeek)
		assert.ErrorIs(t, err, errEmptyOrInvalidSettlementCurrency, "empty settlement currency should return expected error")
		_, err = e.PremiumIndexKLine(t.Context(), currency.USDT, currency.Pair{}, time.Time{}, time.Time{}, 0, kline.OneWeek)
		assert.ErrorIs(t, err, currency.ErrCurrencyPairEmpty, "empty contract should return expected error")
		_, err = e.PremiumIndexKLine(t.Context(), currency.USDT, contract, time.Time{}, time.Time{}, 0, kline.FiveDay)
		assert.ErrorIs(t, err, kline.ErrUnsupportedInterval, "unsupported interval should return expected error")
	})
}

```
It could be made into a test table, but its more ugly and this is a simple function test

gloriousCode (changes requested):

exchanges/gateio/gateio_test.go
thrasher-: Jarvis: Thanks, agreed. Simplified in the latest push by removing the local HTTP mock and folding the useful cases into `TestPremiumIndexKLine` live request and validation subtests.

gloriousCode: Thank you Jarvis. Please place an order for 32767 Simon Says burgers from Grill'd and we can celebrate

gloriousCode (approved):
tACK! 🍔

thrasher-:
Jarvis: CI exposed a separate OKX live-data failure after the GateIO change: OKX can return preopen futures entries with an empty `instId` and a future `listTime`, which are valid upcoming instrument-family previews but not tradable instruments. I added https://github.com/thrasher-corp/gocryptotrader/commit/3edb1add4f91792cc9325c5296abd5ebe03ab1f2 so `UpdateOrderExecutionLimits` only loads live instruments with populated pairs, plus direct regression coverage. Fresh PR CI is green across the matrix.

exchanges/okx/okx_test.go
```diff
@@ -3396,6 +3397,48 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
 	})
 }
 
+func TestLoadInstrumentOrderExecutionLimitsSkipsInactiveAndEmptyInstruments(t *testing.T) {
```
shazbert: Took me 1000 years to read this.

shazbert (approved):
Legend. Nothing from me.

exchanges/okx/okx_test.go
```diff
@@ -3396,6 +3397,48 @@ func TestUpdateOrderExecutionLimits(t *testing.T) {
...
+	exch := &Exchange{}
+	exch.SetDefaults()
+	exch.Name = "OkxTestLoadInstrumentOrderExecutionLimitsSkipsInactiveAndEmptyInstruments"
```
gloriousCode: Drop setting exch.Name

exchanges/okx/okx_wrapper.go
```diff
@@ -380,6 +370,24 @@ func (e *Exchange) UpdateOrderExecutionLimits(ctx context.Context, a asset.Item)
...
+			continue
+		}
+		l = append(l, limits.MinMaxLevel{
+			Key:                    key.NewExchangeAssetPair(e.Name, a, insts[i].InstrumentID),
+			PriceStepIncrementSize: insts[i].TickSize.Float64(),
+			MinimumBaseAmount:      insts[i].MinimumOrderSize.Float64(),
+		})
+	}
+	if len(l) == 0 {
+		return common.ErrNoResponse
```
gloriousCode: This error only makes sense as an initial check at the top of the function. To throw this when you have received data, but not data we want is not "no response". I think keep this Len check here but change it to `common.ErrInvalidResponse`

exchanges/okx/okx_test.go
thrasher-: Jarvis: Thanks, agreed. Tidied this in https://github.com/thrasher-corp/gocryptotrader/commit/ad68d2d5dad0ceecf6a14de4c4b862adae68c328 by shortening the test name/assertion text while keeping the isolated exchange name and the same live/inactive/empty-instrument coverage.

shazbert (approved):
ch ACK!

gloriousCode (changes requested):
Minor changes, thank you

exchanges/okx/okx_wrapper.go
thrasher-: Jarvis: Thanks, agreed. Updated in https://github.com/thrasher-corp/gocryptotrader/commit/13d4bf6d4084bb77706338ea76dec99e7ad2cc50 so empty instrument responses return common.ErrNoResponse, while non-empty responses that are fully filtered return common.ErrInvalidResponse; added direct coverage for both paths. Follow-up cleanup in https://github.com/thrasher-corp/gocryptotrader/commit/f94816190d742aea373d6a71083bbcf69d81df02 uses t.Name() for the test exchange name so the global order-limit store stays isolated without a bespoke hardcoded string.

gloriousCode (approved):
yay

PR #2271, opened 2026-06-18
exchange: Replace per-exchange UpdateTradablePairs with Base implementation and add test

shazbert:
# PR Description

UpdateTradablePairs is removed from wrappers, and is standardised and adds some parallelism. Also added option for assets, so only specific assets can be fetched and updated externally. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/exchanges/exchange.go
+++ b/exchanges/exchange.go
@@ -1829,7 +1829,7 @@ func Bootstrap(ctx context.Context, b IBotExchange) error {
 	}
 
 	if b.GetEnabledFeatures().AutoPairUpdates {
-		if err := b.UpdateTradablePairs(ctx); err != nil {
+		if err := b.UpdateTradablePairs(ctx, b); err != nil {
 			return fmt.Errorf("failed to update tradable pairs: %w", err)
 		}
 	}
@@ -1882,7 +1882,7 @@ func GetDefaultConfig(ctx context.Context, exch IBotExchange) (*config.Exchange,
 	}
 
 	if b.Features.Supports.RESTCapabilities.AutoPairUpdates {
-		err = exch.UpdateTradablePairs(ctx)
+		err = exch.UpdateTradablePairs(ctx, exch)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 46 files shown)


exchanges/exchange.go
```diff
@@ -1987,3 +1987,24 @@ func (b *Base) MessageSequence() int64 {
 func (b *Base) SubscribeAccountBalances() (dispatch.Pipe, error) {
 	return b.Accounts.Subscribe()
 }
+
+// UpdateTradablePairs updates the tradable pairs for the specified assets, or all assets if none are specified.
+func (b *Base) UpdateTradablePairs(ctx context.Context, exch IBotExchange, assets ...asset.Item) error {
```
cranktakular, line 1992: Seems risky/unnecessary to have a Base method and an exch parameter, which are usually the same but could theoretically be different.

I haven't thought about it much, but is there any way you could refactor it to remove one of those two?

exchanges/exchange_test.go
```diff
@@ -2566,6 +2566,74 @@ func TestGetDefaultConfig(t *testing.T) {
...
+	t.Run("honours explicitly specified assets", func(t *testing.T) {
+		t.Parallel()
+		f := newExch()
+		f.pairs = currency.Pairs{btcusdPair}
+
+		require.NoError(t, f.Base.UpdateTradablePairs(t.Context(), f, asset.Spot))
+
+		avail, err := f.GetAvailablePairs(asset.Spot)
+		require.NoError(t, err)
+		assert.Equal(t, currency.Pairs{btcusdPair}, avail, "available pairs should be updated for the specified asset")
```
cranktakular, line 2624: Wouldn't a better test for this be, if the exchange had both Spot and Futures enabled, only one of the two was passed into UpdateTradablePairs, and you checked that the other was not updated?

Since right now this test exercises nothing about assets that would distinguish it from the previous test.

cranktakular (changes requested):
Good stuff! Yet there are a few changes I'd like to see.

Also don't forget to fix the linter issue!

exchanges/exchange.go
gloriousCode, line 1992: I don't really like this move to include the interface as a parameter here. Sure this PR removes a couple of hundred lines from code but now the code is more convoluted. I might be convinced with an explanation.

`Base` cannot call methods from the concrete exchange that embeds it, so this works by passing the parent exchange back into Base. That makes the public interface harder to understand and easier to misuse, because the receiver and the `exch` argument can technically be different exchanges.

It also seems to flatten some exchange-specific behaviour that may not be accidental. Existing implementations differ on `GetAssetTypes(true)` vs `false`, whether they call `EnsureOnePairEnabled`, and Deribit already had custom parallel handling. Adding parallelism everywhere is also a behaviour change, not just deduplication.

I think I would rather keep the normal `UpdateTradablePairs(ctx)`

thrasher-:
@codex please review

shazbert:
MAJOR REJECTION BY @gloriousCode

PR #2275, opened 2026-06-22
exchange: add CredentialsManagement interface and simplify SetCredentials

shazbert:
# PR Description

Set credentials receives a pointer. 
Got rid of a bunch of redundant funcs for credentials management
Added to i bot exchange interface so its good lookin. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


exchanges/binance/binance_test.go
```diff
@@ -37,12 +38,16 @@ import (
...
 // Please supply your own keys here for due diligence testing
 const (
-	apiKey                  = ""
-	apiSecret               = ""
 	canManipulateRealOrders = false
 	useTestNet              = false
 )
 
+// Please supply your own credentials here to do authenticated endpoint testing
+var apiCredentials = &accounts.Credentials{
```
gloriousCode: `apiCredentials` is only used by `binance_live_test.go`, and that file has the `mock_test_off` build tag. Normal `make lint` does not build that file, so this package-level variable is seen as unused.

Please move `apiCredentials` into `binance_live_test.go`, or otherwise make sure it is used by code that is built during the default lint run.

exchanges/gemini/gemini_test.go
```diff
@@ -24,11 +25,12 @@ import (
...
 
 // Please enter sandbox API keys & assigned roles for better testing procedures
-const (
-	apiKey                  = ""
-	apiSecret               = ""
-	canManipulateRealOrders = false
-)
+const canManipulateRealOrders = false
+
+var apiCredentials = &accounts.Credentials{
```
gloriousCode: `apiCredentials` is only used by `gemini_live_test.go`, which is excluded unless the `mock_test_off` build tag is enabled. During normal `make lint`, this variable is therefore unused.

Please move the credential fixture into `gemini_live_test.go`, or add a default-built use for it.

docs/EXCHANGE_API.md
```diff
@@ -53,12 +53,14 @@ supplied meet the requirements to make an authenticated request.
...
     b.SetDefaults()
 
-    // Set default keys 
-    b.API.SetKey("your_key") 
-    b.API.SetSecret("your_secret") 
-    b.API.SetClientID("your_clientid")
-    b.API.SetPEMKey("your_PEM_key")
-    b.API.SetSubAccount("your_specific_subaccount")
+    // Set default credentials
+    b.SetCredentials(&accounts.Credentials{
```
gloriousCode, line 57: This example now uses `accounts.Credentials` for `SetCredentials`, but the context override example below still refers to `exchange.DeployCredentialsToContext` and `exchange.Credentials`.

Those symbols live in `exchange/accounts`, so this example should use `accounts.DeployCredentialsToContext(ctx, &accounts.Credentials{...})` instead. The imports should also reflect that.

exchanges/bitstamp/bitstamp_test.go
```diff
@@ -23,12 +24,14 @@ import (
...
-const (
-	apiKey                  = ""
-	apiSecret               = ""
-	customerID              = "" // This is the customer id you use to log in
-	canManipulateRealOrders = false
-)
+const canManipulateRealOrders = false
+
+// Please supply your own credentials here to do authenticated endpoint testing
+var apiCredentials = &accounts.Credentials{
```
gloriousCode: `apiCredentials` is only referenced from `bitstamp_live_test.go`, but that file is guarded by the `mock_test_off` build tag. Normal linting does not build it, so this variable becomes unused.

Please move the fixture into the live-test file, or keep the variable in code that is built by default.

exchanges/credentials.go
```diff
@@ -179,18 +144,21 @@ func (b *Base) VerifyAPICredentials(creds *accounts.Credentials) error {
 	return nil
 }
 
-// SetCredentials is a method that sets the current API keys for the exchange
-func (b *Base) SetCredentials(apiKey, apiSecret, clientID, subaccount, pemKey, oneTimePassword string) {
+// SetCredentials sets the exchange's default API credentials, copying the
+// supplied credentials so later caller mutation has no effect. When the
+// exchange requires a base64-decoded secret, the secret is decoded here and
+// authenticated support is disabled if decoding fails.
+func (b *Base) SetCredentials(creds *accounts.Credentials) {
```
gloriousCode, line 151: The exchange test template still generates the old `SetCredentials(apiKey, apiSecret, "", "", "", "")` call.

Since `SetCredentials` now accepts `*accounts.Credentials`, `cmd/exchange_template/test.tmpl` also needs to be updated. Otherwise, newly generated exchange tests will not compile.

exchanges/huobi/huobi_test.go
```diff
@@ -18,6 +18,7 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/common/key"
 	"github.com/thrasher-corp/gocryptotrader/core"
 	"github.com/thrasher-corp/gocryptotrader/currency"
+	"github.com/thrasher-corp/gocryptotrader/exchange/accounts"
```
gloriousCode, line 21: Adding the `accounts` import creates a name shadowing issue in `TestSubmitOrder`, where the local variable is also named `accounts`.

`make lint` reports this as `gocritic/importShadow`. Please rename the local variable, for example to `exchangeAccounts`.

gloriousCode (changes requested):
Looks good

exchanges/credentials.go
```diff
@@ -179,18 +144,21 @@ func (b *Base) VerifyAPICredentials(creds *accounts.Credentials) error {
...
-
-	if b.API.CredentialsValidator.RequiresBase64DecodeSecret {
-		result, err := base64.StdEncoding.DecodeString(apiSecret)
+	if creds == nil {
+		b.API.credentials = accounts.Credentials{}
+		return
+	}
+	b.API.credentials = *creds
+
+	if b.API.CredentialsValidator.RequiresBase64DecodeSecret && !b.API.credentials.SecretBase64Decoded {
```
gloriousCode: `SecretBase64Decoded` is derived state for `Secret`, but this assignment copies both values from the caller before the condition below trusts the flag.

A failing sequence is:

1. Kraken or BTC Markets decodes its configured secret and stores `SecretBase64Decoded: true`.
2. A caller retrieves that value through `GetDefaultCredentials()`.
3. The caller replaces `Secret` with a newly rotated base64 secret and passes the struct to `SetCredentials`; the old `true` flag remains.
4. This condition skips decoding, so requests are signed using the encoded text. `VerifyAPICredentials` trusts the same flag and can still report the credentials as valid.

The old string-based setter decoded every newly supplied secret, so this is a regression introduced by accepting the complete credential struct.

gloriousCode (changes requested):
This is mostly minor and one could argue that this *shouldn't* happen. But lots of things shouldn't happen and they do shazbert, THEY DO

exchanges/credentials.go
shazbert: https://github.com/thrasher-corp/gocryptotrader/pull/2275/commits/c63e86fb525082cf8b0cddfd2ec4e6c2b4391a63

gloriousCode (approved):
tACK! Thank you!

thrasher- (approved):
Changes look good!

PR #2288, opened 2026-07-16
websocket: Prevent connect test teardown race

thrasher-:
## Summary

- Prevent raw websocket test managers from starting zero-duration traffic monitors.
- Initialise all three affected successful-`Connect` fixtures with a production-valid traffic timeout.
- Keep production websocket behaviour unchanged.

## Root cause

The tests construct `Manager` instances without calling `Setup`, leaving `trafficTimeout` at zero. `Connect` starts the traffic monitor before completing subscription setup, so `time.After(0)` can trigger an asynchronous shutdown as soon as the manager becomes connected. If shutdown wins the scheduling race, it clears the subscription stores before the assertions run.

## Impact

This removes a scheduling-dependent websocket test failure observed on macOS CI and protects the two adjacent fixtures with the same lifecycle pattern.

## Validation

- `go test ./exchange/websocket -run 'TestConnect(TracksOnExistingConnectionBeforeNewConnection|ReducesTrackedSubscriptionsBeforeBatching|PreBatchTrackedSubscriptionsAutoRecordState)$' -race -count=100`
- `go test ./exchange/websocket -race -count=1`
- `make lint`
- `codespell exchange/websocket/subscriptions_test.go`
- `make misc_checks`

### DIFF
--- a/exchange/websocket/subscriptions_test.go
+++ b/exchange/websocket/subscriptions_test.go
@@ -1416,6 +1416,7 @@ func TestConnectTracksOnExistingConnectionBeforeNewConnection(t *testing.T) {
 	m.exchangeName = "test"
 	m.useMultiConnectionManagement = true
 	m.MaxSubscriptionsPerConnection = 1
+	m.trafficTimeout = time.Minute
 	m.setEnabled(true)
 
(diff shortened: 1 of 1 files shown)


shazbert (approved):
LGTM

gloriousCode (approved):
This is fine 🔥 🐶 ☕ 🔥

PR #2293, opened 2026-07-22
currency/coinmarketcap: fix api response decoding and plan access

thrasher-:
## Summary

- Correct CoinMarketCap V3 quote-array and ID-keyed cryptocurrency and exchange response decoding while rejecting malformed or ambiguous quote collections.
- Model all documented V3 latest fields, including nullable supply and TVL values and the complete per-conversion quote metrics.
- Preserve every latest exchange quote and the existing Binance compatibility view; decode numeric price-conversion IDs and numeric or quoted API error codes.
- Align plan access and live request throttling with current tiers: 50, 300, 600, 750, 1,200, and a 1,600 requests/minute Enterprise floor.
- Add a reversible config v13 migration for legacy plan values, reject invalid plans without partial setup mutation, and preserve the underlying limiter error chain.

## Type of change

- [x] Bug fix
- [x] Breaking change

## Compatibility

- Exported `coinmarketcap.Hobbyist` and `coinmarketcap.Standard` constants are removed; callers must use `Builder` and `Growth`. Direct `SetAccountPlan` calls with the legacy names are rejected.
- Config compatibility is retained: `config.DefaultUnsetAccountPlan` remains `"accountPlan"`, while v13 migrates `hobbyist` to `builder`, `standard` to `growth`, and legacy unset placeholders to `basic`; downgrade restores the legacy names.
- `PriceConversion.ID` changes from `string` to `int64` to match the API response.
- The exported V3 latest models now follow the API contract: `CryptocurrencyLatestListings.MaxSupply` is nullable, `Quote` uses `CryptocurrencyLatestQuoteMap`, and `CryptocurrencyLatestQuotes` uses the same `CryptocurrencyLatestListings` element type. Consumers of the previous field types must update.
- Selecting a plan now reconfigures the live request limiter; Enterprise uses CoinMarketCap's documented 1,600 requests/minute floor.
- Existing legacy object-map quote payloads and `ExchangeLatestQuotes.Binance` remain supported.

## Testing

- `go test ./config ./config/versions/v13 ./currency/coinmarketcap ./exchanges/request -count=1`
- `go test ./config ./config/versions/v13 ./currency/coinmarketcap ./exchanges/request -race -count=1`
- `go test ./config ./config/versions/v13 ./currency/coinmarketcap ./exchanges/request -count=1 -tags=sonic_on`
- `go test ./config ./config/versions/v13 ./currency/coinmarketcap ./exchanges/request -race -count=1 -tags=sonic_on`
- `make lint`
- `make misc_checks`
- `codespell .`

### DIFF
--- a/currency/coinmarketcap/coinmarketcap.go
+++ b/currency/coinmarketcap/coinmarketcap.go
@@ -28,6 +28,9 @@ var (
 	errRateLimiterNotSet     = errors.New("rate limiter not set")
 )
 
+// requestNew allows direct coverage of SetDefaults' defensive requester error path.
+var requestNew = request.New
+
 // NewFromSettings returns a new coin market cap instance with supplied settings
 func NewFromSettings(cfg Settings) (*Coinmarketcap, error) {
 	c := &Coinmarketcap{}
@@ -44,18 +47,13 @@ func (c *Coinmarketcap) SetDefaults() {
 	c.Enabled = false
 	c.Verbose = false
 	c.APIUrl = baseURL
-	// SetDefaults cannot return the requester construction error; newRequester logs it.
-	c.Requester, _ = newRequester(c.Name,
+	var err error
+	c.Requester, err = requestNew(c.Name,
 		common.NewHTTPClientWithTimeout(defaultTimeOut),
-		getRateLimits())
-}
-
-func newRequester(name string, client *http.Client, limits request.RateLimitDefinitions) (*request.Requester, error) {
-	requester, err := request.New(name, client, request.WithLimiter(limits))
+		request.WithLimiter(getRateLimits()))
 	if err != nil {
 		log.Errorln(log.Global, err)
 	}
-	return requester, err
 }
 
 // Setup sets user configuration
--- a/currency/coinmarketcap/coinmarketcap.go
+++ b/currency/coinmarketcap/coinmarketcap.go
@@ -24,6 +24,8 @@ var (
 	errAPIResponse           = errors.New("coinmarketcap api response error")
 	errEndpointNotAvailable  = errors.New("endpoint is not yet available")
 	errFunctionUseNotAllowed = errors.New("function use not allowed, higher plan needed")
+	errInvalidAccountPlan    = errors.New("invalid account plan")
+	errRateLimiterNotSet     = errors.New("rate limiter not set")
 )
 
 // NewFromSettings returns a new coin market cap instance with supplied settings
@@ -59,10 +61,13 @@ func (c *Coinmarketcap) Setup(conf Settings) error {
 		return nil
 	}
 
+	if err := c.SetAccountPlan(conf.AccountPlan); err != nil {
+		return err
+	}
 	c.Enabled = true
 	c.Verbose = conf.Verbose
 	c.APIkey = conf.APIKey
-	return c.SetAccountPlan(conf.AccountPlan)
+	return nil
 }
 
 // GetCryptocurrencyInfo returns all static metadata for one or more
@@ -181,7 +186,7 @@ func (c *Coinmarketcap) GetCryptocurrencyLatestMarketPairs(currencyID, start, li
 		Status Status                          `json:"status"`
 	}{}
 
-	err := c.CheckAccountPlan(Standard)
+	err := c.CheckAccountPlan(Growth)
 	if err != nil {
 		return resp.Data, err
 	}
@@ -222,7 +227,7 @@ func (c *Coinmarketcap) GetCryptocurrencyOHLCHistorical(currencyID int64, tStart
 		Status Status                       `json:"status"`
 	}{}
 
-	err := c.CheckAccountPlan(Standard)
+	err := c.CheckAccountPlan(Startup)
 	if err != nil {
 		return resp.Data, err
 	}
@@ -324,13 +329,13 @@ func (c *Coinmarketcap) GetCryptocurrencyLatestQuotes(currencyID ...int64) (Cryp
 // tEnd - refers to the end of the time block if zero will default to time.Now()
 func (c *Coinmarketcap) GetCryptocurrencyHistoricalQuotes(currencyID int64, tStart, tEnd time.Time) (CryptocurrencyHistoricalQuotes, error) {
 	resp := struct {
-		Data   CryptocurrencyHistoricalQuotes `json:"data"`
-		Status Status                         `json:"status"`
+		Data   map[string]CryptocurrencyHistoricalQuotes `json:"data"`
+		Status Status                                    `json:"status"`
 	}{}
 
-	err := c.CheckAccountPlan(Standard)
+	err := c.CheckAccountPlan(Basic)
 	if err != nil {
-		return resp.Data, err
+		return CryptocurrencyHistoricalQuotes{}, err
 	}
 
 	val := url.Values{}
@@ -343,14 +348,18 @@ func (c *Coinmarketcap) GetCryptocurrencyHistoricalQuotes(currencyID int64, tSta
 
 	err = c.SendHTTPRequest(http.MethodGet, endpointGetMarketQuotesHistorical, val, &resp)
 	if err != nil {
-		return resp.Data, err
+		return CryptocurrencyHistoricalQuotes{}, err
 	}
 
 	if resp.Status.ErrorCode != 0 {
-		return resp.Data, fmt.Errorf("%w: %s", errAPIResponse, resp.Status.ErrorMessage)
+		return CryptocurrencyHistoricalQuotes{}, fmt.Errorf("%w: %s", errAPIResponse, resp.Status.ErrorMessage)
 	}
 
-	return resp.Data, nil
+	result, found := resp.Data[strconv.FormatInt(currencyID, 10)]
+	if !found {
+		return CryptocurrencyHistoricalQuotes{}, fmt.Errorf("%w: cryptocurrency ID %d", common.ErrNoResponse, currencyID)
+	}
+	return result, nil
 }
 
 // GetExchangeInfo returns all static metadata for one or more exchanges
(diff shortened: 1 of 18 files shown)


thrasher-:
@codex please review

currency/coinmarketcap/coinmarketcap_types.go
```diff
@@ -434,15 +446,19 @@ func (c *APIErrorCode) UnmarshalJSON(data []byte) error {
 
 // UnmarshalJSON handles quote payloads that may be either an object map or array.
 func (q *QuoteMap) UnmarshalJSON(data []byte) error {
-	if err := json.Unmarshal(data, (*map[string]Currency)(q)); err == nil {
+	var quotes map[string]Currency
+	if err := json.Unmarshal(data, &quotes); err == nil {
+		*q = quotes
 		return nil
```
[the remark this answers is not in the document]
thrasher-: Alice: Thanks for flagging this. The named/unnamed map assignments here are valid Go: `QuoteMap` and `map[string]Currency` have identical underlying types, and one side is unnamed. At the current head, the legacy `QuoteMap` decoder preserves API-provided keys from object maps and arrays of object maps; it does not derive keys from `symbol`. V3 quote arrays are handled by `CryptocurrencyLatestQuoteMap`, which rejects missing, empty, whitespace-padded, and duplicate symbols, with direct coverage in `TestCryptocurrencyLatestQuoteMapUnmarshal`.

currency/coinmarketcap/coinmarketcap.go
```diff
@@ -343,14 +343,14 @@ func (c *Coinmarketcap) GetCryptocurrencyHistoricalQuotes(currencyID int64, tSta
...
+		return CryptocurrencyHistoricalQuotes{}, err
 	}
 
 	if resp.Status.ErrorCode != 0 {
-		return resp.Data, fmt.Errorf("%w: %s", errAPIResponse, resp.Status.ErrorMessage)
+		return CryptocurrencyHistoricalQuotes{}, fmt.Errorf("%w: %s", errAPIResponse, resp.Status.ErrorMessage)
 	}
 
-	return resp.Data, nil
+	return resp.Data[strconv.FormatInt(currencyID, 10)], nil
```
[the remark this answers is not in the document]
thrasher-: Alice: Thanks for flagging this. At the current head, both historical methods decode `data` as an ID-keyed map, explicitly check for the requested ID, and return a wrapped `common.ErrNoResponse` when it is absent. `TestGetCryptocurrencyHistoricalQuotesRejectsMissingResult` and `TestGetExchangeHistoricalQuotesRejectsMissingResult` directly cover both paths.

currency/coinmarketcap/coinmarketcap.go
```diff
@@ -696,22 +727,38 @@ func (c *Coinmarketcap) CheckAccountPlan(minAllowable uint8) error {
...
 	default:
-		log.Warnf(log.Currency, "account plan %s not found, defaulting to basic", s)
-		c.Plan = Basic
+		return fmt.Errorf("%w: %q", errInvalidAccountPlan, s)
+	}
+
+	if c.Requester != nil {
+		limiter := c.Requester.GetRateLimiterDefinitions()[request.Unset]
+		if err := limiter.SetRateLimit(rateInterval, requestRate); err != nil {
+			return fmt.Errorf("%w for account plan %q: %v", errRateLimiterNotSet, planName, err)
```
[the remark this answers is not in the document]
thrasher-: Alice: Thanks for catching this. Fixed in https://github.com/thrasher-corp/gocryptotrader/commit/00dd6879f60b2ad90dcbc145bc844c6b85f1ca99. `SetAccountPlan` now wraps the underlying limiter error with `%w`, so `errors.Is` matches both `errRateLimiterNotSet` and `common.ErrNilPointer`. `TestSetAccountPlanWithoutRateLimiter` directly asserts both sentinels and verifies that the plan remains unchanged on error.

gloriousCode (approved):
tACK!

currency/coinmarketcap/coinmarketcap_types.go
```diff
@@ -160,24 +168,40 @@ type CryptocurrencyHistoricalListings struct {
...
+	NumMarketPairs                int64                        `json:"num_market_pairs"`
+	CmcRank                       int64                        `json:"cmc_rank"`
+	LastUpdated                   time.Time                    `json:"last_updated"`
+	TVLRatio                      *float64                     `json:"tvl_ratio"`
+	SelfReportedCirculatingSupply *float64                     `json:"self_reported_circulating_supply"`
+	SelfReportedMarketCap         *float64                     `json:"self_reported_market_cap"`
+	UnlockedCirculatingSupply     *float64                     `json:"unlocked_circulating_supply"`
+	UnlockedMarketCap             *float64                     `json:"unlocked_market_cap"`
+	// MintedMarketCap is documented on the listings/latest response.
+	MintedMarketCap *float64 `json:"minted_market_cap"`
```
shazbert: These fields types don't need to be *float64 and can just be float64

shazbert (changes requested):
Just minor thingo

currency/coinmarketcap/coinmarketcap_types.go
thrasher-: Alice: Addressed in https://github.com/thrasher-/gocryptotrader/commit/980f8bb9888f52ed5a5d353b14e75575753125d0 the seven requested supply and TVL fields now use `float64`, with direct JSON-decoding coverage confirming `null` values decode to zero.

currency/coinmarketcap/coinmarketcap_types.go
```diff
@@ -240,24 +264,39 @@ type CryptocurrencyOHLCLatest map[string]struct {
...
+	PercentChange24H       float64   `json:"percent_change_24h"`
+	PercentChange7D        float64   `json:"percent_change_7d"`
+	PercentChange30D       float64   `json:"percent_change_30d"`
+	PercentChange60D       float64   `json:"percent_change_60d"`
+	PercentChange90D       float64   `json:"percent_change_90d"`
+	MarketCap              float64   `json:"market_cap"`
+	MarketCapDominance     float64   `json:"market_cap_dominance"`
+	FullyDilutedMarketCap  float64   `json:"fully_diluted_market_cap"`
+	MintedMarketCap        float64   `json:"minted_market_cap"`
+	TVL                    *float64  `json:"tvl"`
```
shazbert: This doesn't need to be a pointer

exchanges/request/limit.go
```diff
@@ -82,6 +82,18 @@ func GetRateLimiterWithWeight(l *rate.Limiter, weight Weight) *RateLimiterWithWe
...
+	if err := common.NilGuard(r); err != nil {
+		return err
+	}
+
+	r.m.Lock()
+	r.limiter = NewRateLimit(interval, actions)
+	r.m.Unlock()
+	return nil
+}
+
```
shazbert: Instead of having this why not have a endpoint rate limit based of the tiers defined in the package?

shazbert (changes requested):
Thanks for the changes found some more than looks good!

exchanges/request/limit.go
thrasher-: Alice: Addressed in https://github.com/thrasher-/gocryptotrader/commit/7f35b8bcd62f6913cdb5f08c87c99ad2f50fe117

CoinMarketCap now owns six typed endpoint rate-limit definitions, creates fresh limiter state per client, and selects the configured account tier for each request. `SetAccountPlan` validates the selected definition without mutating or resetting limiter state, and the shared mutable rate-limit setter added by this PR has been removed.

currency/coinmarketcap/coinmarketcap.go
```diff
@@ -44,14 +44,18 @@ func (c *Coinmarketcap) SetDefaults() {
 	c.Enabled = false
 	c.Verbose = false
 	c.APIUrl = baseURL
-	var err error
-	c.Requester, err = request.New(c.Name,
+	// SetDefaults cannot return the requester construction error; newRequester logs it.
+	c.Requester, _ = newRequester(c.Name,
```
shazbert: handle error and log it here. If SetDefaults changes in future we can then RM and then pass it to the caller.

shazbert (changes requested):
one thing left over from me and then looks good to go. 🚀

currency/coinmarketcap/coinmarketcap.go
thrasher-: Alice: Addressed in https://github.com/thrasher-corp/gocryptotrader/commit/47f5717383d18a858c61ffea286567186021c8f4 — SetDefaults now handles and logs requester construction errors inline, the single-use helper has been removed, and direct coverage exercises the defensive error path.

currency/coinmarketcap/coinmarketcap.go
```diff
@@ -28,6 +28,9 @@ var (
 	errRateLimiterNotSet     = errors.New("rate limiter not set")
 )
 
+// requestNew allows direct coverage of SetDefaults' defensive requester error path.
+var requestNew = request.New
```
shazbert: Can this be removed? globals that are only used in tests by overriding them, make things more complex and obtuse.

shazbert (changes requested):
Just one thing

currency/coinmarketcap/coinmarketcap.go
thrasher-: Alice: Addressed in https://github.com/thrasher-corp/gocryptotrader/commit/3cee0c47085114151d8ef8c34553d1ea6be0a337 — the test-only global and its synthetic override test have been removed. SetDefaults now calls request.New directly while retaining its inline error handling and logging.

shazbert (approved):
ch ACK. Thank youuuuu.

PR #2297, opened 2026-07-25
orderbook: reduce insert update allocations

thrasher-:
## Summary

- replace the nested middle-insert append with one slice extension, an overlapping-safe copy, and assignment
- snapshot each incoming level before mutating the destination so aliased update slices retain the existing read-before-write behaviour
- add direct coverage for empty, head, middle, tail, collision, capacity-growth, capacity-reuse, and aliased-input paths
- group orderbook benchmarks into subject-specific files without changing existing benchmark bodies
- document that the insert-only path can be removed when BitMEX support is removed

## Performance

The five workloads introduced by this PR were measured between the PR base (Before) and the measured production state (After) with identical benchmark bodies; the current head only adds benchmark-result comments. Each row uses 20 counterbalanced fresh-process observations per revision on Go 1.26.5/Linux amd64 with `GOMAXPROCS=1`, a pinned CPU, PGO disabled, allocation reporting, and one-second adaptive timing.

| Scope | Workload | Time/op, Before → After | B/op | allocs/op | Result |
| --- | --- | ---: | ---: | ---: | --- |
| Primary | `MiddleFreshCopy` | 11.156 µs → 7.631 µs | 32,640 → 21,760 | 2 → 1 | -31.60%, p=0.000 |
| Primary | `MiddleSpareCapacity` | 4,643.0 ns → 462.4 ns | 10,880 → 0 | 1 → 0 | -90.04%, p=0.000 |
| Primary | `ProcessUpdateInsertDelete` | 5,021.0 ns → 688.1 ns | 10,880 → 0 | 1 → 0 | -86.30%, p=0.000 |
| Exploratory guard | `TailSpareCapacity` | 565.1 ns → 566.1 ns | 0 → 0 | 0 → 0 | No significant difference at n=20, p=0.774 |
| Exploratory guard | `Collision` | 625.9 ns → 626.2 ns | 104 → 104 | 3 → 3 | No significant difference at n=20, p=0.899 |

Benchstat's displayed `p=0.000` values are rounded, not literal zero. The 16 existing benchmark bodies that were only moved into subject-specific files are unchanged and are not remeasured or claimed here.

These are serial, single-core, 256-level microbenchmarks and do not establish whole-application throughput, concurrent scaling, exchange/websocket latency, or production tail latency. `MiddleSpareCapacity` includes the benchmark's state-restoration copy; `ProcessUpdateInsertDelete` includes locking, action dispatch, mutation, and error checks for a balanced public insert/delete lifecycle.

## Validation

- `go test ./exchanges/orderbook -run '^TestLevelsInsertUpdates$|^TestInsertUpdates(Bid|Ask)$' -count=1000`
- `go test ./exchanges/orderbook -count=1`
- `go test -race ./exchanges/orderbook -count=1`
- `go test ./... -count=1`
- `go test -race ./... -count=1`
- `go test ./exchanges/orderbook -run '^$' -bench '^(BenchmarkInsertUpdates|BenchmarkProcessUpdateInsertDelete)$' -benchtime=1x -count=1`
- `make lint`
- `codespell .`
- `make misc_checks`

The exact base also reproduces existing `-count=10` failures in `TestGetOrderbook` and `TestBookGetDepth` because package-global registry state persists between in-process repetitions; the changed-path stress run passes 1,000 repetitions.


exchanges/orderbook/levels.go
```diff
@@ -213,21 +213,24 @@ updates:
 func (l *Levels) insertUpdates(updts Levels, comp comparison) error {
 updates:
 	for x := range updts {
+		update := updts[x]
```
shazbert:
```suggestion
for _, update  := range updts {
```

shazbert (changes requested):
Can you also write a note that this function can be removed when Bitmex is removed as a TODO. This will happen very soon.

thrasher-:
@codex please review

thrasher-:
Iris: Added the BitMEX-removal TODO above `insertUpdates` as requested.

exchanges/orderbook/levels.go
thrasher-: Iris: Updated the loop to range directly over each `update` value as suggested.

shazbert (approved):
Changes ACK. Thanks.

gloriousCode (approved):
Nice, [tACK](https://www.youtube.com/watch?v=_J9NpHKrKMw)!

shazbert (approved):
ch ack

PR #2298, opened 2026-07-26
types: reduce timestamp unmarshal allocations

thrasher-:
## Summary

- normalise supported timestamp bytes in a bounded local buffer instead of materialising and concatenating intermediate strings
- return safely for direct nil or empty byte input without changing the receiver, while preserving existing null, zero, quoted/unquoted, and error behaviour
- extend direct and decoder coverage across supported timestamp widths, buffer boundaries, signs, and deterministic errors without treating dotted calendar dates as a supported format
- benchmark four representative quoted timestamp shapes at both the direct method and project JSON decoder boundaries

## Performance

Measured the four retained benchmark cases between the PR base (Before) and current head (After) with an identical harness and 20 counterbalanced fresh-process observations per revision on Go 1.26.5/Linux amd64 with `GOMAXPROCS=1`, a pinned CPU, PGO disabled, and one second per row:

| Path | Before time/op | After time/op | Change | B/op | allocs/op |
| --- | ---: | ---: | ---: | ---: | ---: |
| direct/seconds | 55.67 ns ±2% | 32.10 ns ±2% | -42.34% (`p=0.000`) | 16 → 0 | 1 → 0 |
| decoder/seconds | 229.2 ns ±1% | 209.1 ns ±1% | -8.75% (`p=0.000`) | 160 → 144 | 2 → 1 |
| direct/padded_width | 97.26 ns ±2% | 40.50 ns ±2% | -58.36% (`p=0.000`) | 32 → 0 | 2 → 0 |
| decoder/padded_width | 280.3 ns ±1% | 225.5 ns ±1% | -19.55% (`p=0.000`) | 176 → 144 | 3 → 1 |
| direct/fractional | 119.55 ns ±1% | 56.73 ns ±1% | -52.54% (`p=0.000`) | 48 → 0 | 2 → 0 |
| decoder/fractional | 335.3 ns ±2% | 279.4 ns ±1% | -16.70% (`p=0.000`) | 192 → 144 | 3 → 1 |
| direct/date | 102.7 ns ±1% | 102.9 ns ±1% | no significant difference detected at n=20 (`p=0.542`) | 16 → 8 | 1 → 1 |
| decoder/date | 281.2 ns ±2% | 285.7 ns ±1% | +1.58% (`p=0.027`, exploratory) | 160 → 152 | 2 → 2 |

The `±` values are Benchstat's 95% confidence intervals for each median, and each displayed `p=0.000` is rounded rather than zero. The direct and decoder fractional rows were the preselected timing claims and both remain significant after Holm correction. The other timing rows are descriptive; allocation values were identical across all 20 observations per revision at the printed precision.

- Relevance: measures the changed timestamp parser directly and through the project decoder, including no-padding, padding, fractional normalisation, and calendar-date paths.
- Related checks: the date path retained its allocation counts while reducing B/op; its direct timing had no significant difference, while the small decoder timing increase is exploratory.
- Limitations: results are from one serial microbenchmark host and do not establish exchange/websocket throughput, multi-core scaling, or production tail latency.

## Validation

- `go test ./types -run '^(TestUnmarshalJSON|TestTimeUnmarshalJSONEmptyInput)$' -count=1000`
- `go test ./types -count=10`
- `go test -race ./types -count=1`
- `go test -tags=sonic_on ./types -count=10`
- `go test -race -tags=sonic_on ./types -count=1`
- `go test ./types -run '^$' -bench '^BenchmarkUnmarshalJSON$' -benchtime=1x -count=1`
- `make lint`
- `codespell .`
- `make misc_checks`

### DIFF
--- a/types/time.go
+++ b/types/time.go
@@ -23,6 +23,9 @@ var ErrInvalidTimestampFormat = errors.New("invalid timestamp format")
 func (t *Time) UnmarshalJSON(data []byte) error {
 	timestamp := data
 
+	if len(timestamp) == 0 {
+		return nil
+	}
 	if timestamp[0] == '"' {
 		timestamp = timestamp[1 : len(timestamp)-1]
 	}
--- a/types/time.go
+++ b/types/time.go
@@ -1,10 +1,10 @@
 package types
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"strconv"
-	"strings"
 	"time"
 
 	"github.com/thrasher-corp/gocryptotrader/encoding/json"
@@ -21,44 +21,68 @@ var ErrInvalidTimestampFormat = errors.New("invalid timestamp format")
 
 // UnmarshalJSON deserialises json and timestamp information.
 func (t *Time) UnmarshalJSON(data []byte) error {
-	s := string(data)
+	timestamp := data
 
-	if s[0] == '"' {
-		s = s[1 : len(s)-1]
+	if timestamp[0] == '"' {
+		timestamp = timestamp[1 : len(timestamp)-1]
 	}
 
-	if s == "" || s[0] == 'n' || s == "0" {
+	if len(timestamp) == 0 || timestamp[0] == 'n' || (len(timestamp) == 1 && timestamp[0] == '0') {
 		return nil
 	}
 
-	if target := strings.Index(s, "."); target != -1 {
-		s = s[:target] + s[target+1:]
-
-		if strings.Trim(s, "0") == "" {
+	target := bytes.IndexByte(timestamp, '.')
+	length := len(timestamp)
+	if target != -1 {
+		length--
+	}
+	padding := 0
+	switch length {
+	case 12, 15, 18: // Expects a string of length 10 (seconds), 13 (milliseconds), 16 (microseconds), or 19 (nanoseconds) representing a Unix timestamp
+		padding = 1
+	case 11, 14, 17:
+		padding = 2
+	}
+	if target != -1 || padding != 0 {
+		var normalised [19]byte // A nanosecond Unix timestamp is the largest supported representation.
+		length += padding
+		var destination []byte
+		if length <= len(normalised) {
+			destination = normalised[:length]
+		} else {
+			destination = make([]byte, length)
+		}
+		if target == -1 {
+			copy(destination, timestamp)
+		} else {
+			copy(destination, timestamp[:target])
+			copy(destination[target:], timestamp[target+1:])
+		}
+		for x := length - padding; x < length; x++ {
+			destination[x] = '0'
+		}
+		timestamp = destination
+		if target != -1 && len(bytes.Trim(timestamp, "0")) == 0 {
 			return nil
 		}
 	}
 
-	switch len(s) {
-	case 8:
-		parsed, err := time.Parse("20060102", s)
+	if len(timestamp) == 8 {
+		value := string(timestamp)
+		parsed, err := time.Parse("20060102", value)
 		if err != nil {
-			return fmt.Errorf("%w error parsing %q into date: %w", ErrInvalidTimestampFormat, s, err)
+			return fmt.Errorf("%w error parsing %q into date: %w", ErrInvalidTimestampFormat, value, err)
 		}
 		*t = Time(parsed)
 		return nil
-	case 12, 15, 18: // Expects a string of length 10 (seconds), 13 (milliseconds), 16 (microseconds), or 19 (nanoseconds) representing a Unix timestamp
-		s += "0"
-	case 11, 14, 17:
-		s += "00"
 	}
 
-	unixTS, err := strconv.ParseInt(s, 10, 64)
+	unixTS, err := strconv.ParseInt(string(timestamp), 10, 64)
 	if err != nil {
 		return fmt.Errorf("error parsing unix timestamp: %w", err)
 	}
 
-	switch len(s) {
+	switch len(timestamp) {
 	case 10:
 		*t = Time(time.Unix(unixTS, 0))
 	case 13:
--- a/types/time_test.go
+++ b/types/time_test.go
@@ -69,6 +69,12 @@ func TestUnmarshalJSON(t *testing.T) {
 	}
 }
 
+// Reproduce from the repository root (Go 1.26.1, linux/amd64 WSL2,
+// Intel i7-6700K, warm Go build cache):
+// GOMAXPROCS=1 go test ./types -run '^$' -bench '^BenchmarkUnmarshalJSON$' -benchmem -benchtime=1s -count=15 -cpu=1
+// 4005054         304.1 ns/op       168 B/op          2 allocs/op (current; median time/op of 15 samples)
+// 6152384         195.5 ns/op       168 B/op          2 allocs/op (older historical recording; conditions
+// and statistic not retained, so it is not directly comparable)
 func BenchmarkUnmarshalJSON(b *testing.B) {
 	var testTime Time
 	for b.Loop() {
--- a/types/time_test.go
+++ b/types/time_test.go
@@ -69,12 +69,6 @@ func TestUnmarshalJSON(t *testing.T) {
 	}
 }
 
-// Reproduce from the repository root (Go 1.26.1, linux/amd64 WSL2,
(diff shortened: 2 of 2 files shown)


types/time.go
```diff
@@ -21,44 +21,68 @@ var ErrInvalidTimestampFormat = errors.New("invalid timestamp format")
 
 // UnmarshalJSON deserialises json and timestamp information.
 func (t *Time) UnmarshalJSON(data []byte) error {
-	s := string(data)
+	timestamp := data
 
-	if s[0] == '"' {
-		s = s[1 : len(s)-1]
+	if timestamp[0] == '"' {
```
shazbert, line 29: In the test could you please add this case:

{"", time.Time{}, nil},

I think the len check needs to be done first.

shazbert (changes requested):
one thing from me 🚀

thrasher-:
@codex please review

thrasher-:
Iris: Thanks for flagging this. Both behaviours are inherited unchanged from `master`: the decoder’s `n`-prefix shortcut accepts values such as `"notatime"`, while `null`, `""`, `"0"`, and zero-valued fractions return without clearing an existing receiver. This PR is intended to remove allocations while preserving accepted-input and receiver semantics. A test expecting a seeded receiver to be cleared would therefore introduce a new contract and fail on the base revision; Go’s `time.Time.UnmarshalJSON` also treats JSON `null` as a no-op. I suggest handling exact-null validation and any reset policy as a separate correctness change with explicit compatibility tests. The direct empty-`[]byte` panic is distinct and can be fixed here with a length guard without changing valid-input behaviour.

types/time_test.go
```diff
@@ -26,17 +26,28 @@ func TestUnmarshalJSON(t *testing.T) {
 		{`"0.00000"`, time.Time{}, nil},
 		{`"0.0.0.0"`, time.Time{}, strconv.ErrSyntax},
 		{`"0.1"`, time.Time{}, ErrInvalidTimestampFormat},
+		{`"202003.25"`, time.Date(2020, 3, 25, 0, 0, 0, 0, time.UTC), nil},
```
gloriousCode: Is this really acceptable? What kind of scenario is this a proper date format?

I can accept not actioning fixing this within this PR, but I don't think you should enshrine it as an acceptable pathway either

gloriousCode (changes requested):
After the existing things are addressed, it should be good

I updated the benchmark locally to be more thorough to better demonstrate improvements:
```

func BenchmarkUnmarshalJSON(b *testing.B) {
	for _, tc := range []struct {
		name  string
		input string
	}{
		{"seconds", `"1628736847"`},
		{"padded_width", `"16287368473"`},
		{"fractional", `"1691122380942.173000"`},
		{"date", `"20200325"`},
	} {
		data := []byte(tc.input)
		b.Run("direct/"+tc.name, func(b *testing.B) {
			var testTime Time
			b.ReportAllocs()
			for b.Loop() {
				if err := testTime.UnmarshalJSON(data); err != nil {
					b.Fatal(err)
				}
			}
		})
		b.Run("decoder/"+tc.name, func(b *testing.B) {
			var testTime Time
			b.ReportAllocs()
			for b.Loop() {
				if err := json.Unmarshal(data, &testTime); err != nil {
					b.Fatal(err)
				}
			}
		})
	}
}

```

Path | Master | PR | Change | Memory
-- | -- | -- | -- | --
testTime.Unmarshal seconds | 26.69 ns | 17.61 ns | -34.0% | 16 B/1 alloc → 0 B/0
json.Unmarshal(data, &testTime) seconds | 95.40 ns | 87.84 ns | -7.9% | 160 B/2 → 144 B/1
testTime.Unmarshal  padded | 43.57 ns | 21.38 ns | -50.9% | 32 B/2 → 0 B/0
json.Unmarshal(data, &testTime)  padded | 114.55 ns | 93.00 ns | -18.8% | 176 B/3 → 144 B/1
testTime.Unmarshal  fractional | 52.35 ns | 28.23 ns | -46.1% | 48 B/2 → 0 B/0
json.Unmarshal(data, &testTime)  fractional | 141.2 ns | 117.5 ns | -16.8% | 192 B/3 → 144 B/1
testTime.Unmarshal  date | 40.21 ns | 39.30 ns | -2.3% | 16 B/1 → 8 B/1
json.Unmarshal(data, &testTime)  date | 107.2 ns | 108.6 ns | No significant change | 160 B/2 → 152 B/2

types/time.go
thrasher-: Iris: Addressed in [e37c6d9e](https://github.com/thrasher-corp/gocryptotrader/commit/e37c6d9e0f31e4cbd09d258ae528fd97193c6f22). `Time.UnmarshalJSON` now checks the input length before indexing, and the new regression test directly covers both nil and empty byte slices against a seeded non-zero receiver. I kept these cases separate from the main table because `json.Unmarshal` correctly rejects nil and empty JSON documents, whereas the direct method can safely return without changing the receiver.

types/time.go
```diff
@@ -21,44 +21,68 @@ var ErrInvalidTimestampFormat = errors.New("invalid timestamp format")
...
+		timestamp = timestamp[1 : len(timestamp)-1]
 	}
 
-	if s == "" || s[0] == 'n' || s == "0" {
+	if len(timestamp) == 0 || timestamp[0] == 'n' || (len(timestamp) == 1 && timestamp[0] == '0') {
```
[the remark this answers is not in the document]
thrasher-: Iris: This is inherited behaviour from `master`, not introduced by this allocation refactor. The base implementation also accepts stripped values beginning with `n` and leaves an existing receiver unchanged for `null`, empty or zero values, and all-zero fractions. Tightening null validation or changing receiver-reset semantics would be a separate compatibility change, so I have left those behaviours unchanged here. The distinct direct nil and empty-byte panic is now guarded.

types/time_test.go
```diff
@@ -46,15 +57,24 @@ func TestUnmarshalJSON(t *testing.T) {
...
-			require.ErrorIsf(t, err, tc.expError, "Unmarshal must not error for input %q", tc.input)
-			assert.Equal(t, tc.want, testTime.Time())
+			err := testTime.UnmarshalJSON([]byte(tc.input))
+			require.ErrorIsf(t, err, tc.expError, "UnmarshalJSON must return the expected error for input %q", tc.input)
+			assert.Equalf(t, tc.want, testTime.Time(), "UnmarshalJSON should set Time correctly for input %q", tc.input)
+
+			var decoded Time
+			err = json.Unmarshal([]byte(tc.input), &decoded)
+			require.ErrorIsf(t, err, tc.expError, "json.Unmarshal must return the expected error for input %q", tc.input)
+			assert.Equalf(t, tc.want, decoded.Time(), "json.Unmarshal should set Time correctly for input %q", tc.input)
```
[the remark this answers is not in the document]
thrasher-: Iris: Clearing the receiver in these cases would differ from `master`: `null`, quoted empty or zero values, and all-zero fractions currently return successfully without assigning to `*t`. Seeding these rows and expecting zero would therefore add a new behavioural contract to this allocation-only PR. I added a seeded non-zero receiver test for the newly guarded direct nil and empty-byte path, while leaving null and zero reset semantics for a separate compatibility-focused change.

types/time_test.go
thrasher-: Iris: Agreed. I removed the `"202003.25"` test case in [e37c6d9e](https://github.com/thrasher-corp/gocryptotrader/commit/e37c6d9e0f31e4cbd09d258ae528fd97193c6f22), so this PR no longer presents that inherited parser quirk as a supported date format. The parser behaviour itself remains unchanged; changing accepted formats should be handled separately with an explicit compatibility decision.

thrasher-:
Iris: Thanks — I adopted the four-case direct and decoder benchmark matrix and added result validation outside the timed loop. Across 20 counterbalanced fresh-process observations per revision, the direct seconds, padded-width, and fractional paths improved by 42.34–58.36% and fell from one or two allocations to zero. Their decoder paths improved by 8.75–19.55% and removed one or two allocations. Direct date timing showed no significant difference; decoder date timing increased by an exploratory 1.58%, while both date paths reduced bytes per operation.

shazbert (approved):
changes ACK

PR #2299, opened 2026-07-26
exchanges/coinut: fix trade history pagination

thrasher-:
# PR Description

Coinut's REST trade-history request dropped valid pagination offsets above 100 and conditioned `limit` on `start`. This could prevent callers from paginating beyond the first 100 trades or supplying a limit independently.

This change includes each non-negative pagination argument independently. It adds a local HTTP regression test covering omitted pagination, limit-only requests, start-only requests beyond 100, and combined start/limit requests.

No linked issue.

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [x] golangci-lint run
- [x] go test ./exchanges/coinut -race -count=1
- [x] go test ./exchanges/coinut -run '^TestGetTradeHistoryPagination$' -race -count=20
- [x] make misc_checks
- [x] codespell

The focused regression test fails against the base for start values above 100 and passes after the change. The full repository race suite was not run; GitHub Actions has not run.

## Checklist

Documentation changes and regeneration are not required for this request-parameter correction. Local package tests pass, but GitHub Actions has not run. There are no dependent downstream changes.

### DIFF
--- a/exchanges/coinut/coinut_test.go
+++ b/exchanges/coinut/coinut_test.go
@@ -143,7 +143,14 @@ func TestGetTradeHistoryPagination(t *testing.T) {
 			require.NoError(t, err, "GetTradeHistory must not error")
 			assert.Equal(t, TradeHistory{Trades: []OrderFilledResponse{}}, result, "Trade history should match the response")
 
-			request := <-requestC
+			timer := time.NewTimer(time.Second)
+			defer timer.Stop()
+			var request requestResult
+			select {
+			case request = <-requestC:
+			case <-timer.C:
+				t.Fatal("GetTradeHistory must send a request to the mock server")
+			}
 			require.NoError(t, request.err, "Decoding the request must not error")
 			delete(request.payload, "nonce")
 			expected := map[string]any{
--- a/exchanges/coinut/coinut_test.go
+++ b/exchanges/coinut/coinut_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"os"
+	"slices"
 	"sort"
 	"strconv"
 	"strings"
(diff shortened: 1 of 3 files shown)


exchanges/coinut/coinut_test.go
```diff
@@ -101,6 +103,64 @@ func TestSeedInstruments(t *testing.T) {
...
+func TestGetTradeHistoryPagination(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		name  string
+		start int64
+		limit int64
+	}{
+		{name: "omit pagination", start: -1, limit: -1},
+		{name: "limit only", start: -1, limit: 2},
```
shazbert: I think limit and start can be uints and just != 0 check can be done, it can then fail on the exchange side.

shazbert (changes requested):
Legend.

thrasher-:
@codex please review

exchanges/coinut/coinut_test.go
thrasher-: Alice: Updated `start` and `limit` to `uint64`, switched their inclusion checks to `!= 0`, and now use zero for omission in the REST wrapper. Added exact `math.MaxUint64` request coverage as well.

shazbert (approved):
Thanks approved

exchanges/coinut/coinut_test.go
```diff
@@ -250,19 +324,343 @@ func TestGetActiveOrders(t *testing.T) {
...
+					"trades":       trades,
+				})
+				if err != nil {
+					return err
+				}
+				return conn.WriteMessage(gws.TextMessage, response)
+			}))
+			t.Cleanup(server.Close)
+
+			ex := new(Exchange)
```
gloriousCode, line 632: New t.Run, new t.Parallel

exchanges/coinut/coinut_wrapper.go
```diff
@@ -918,7 +918,7 @@ func (e *Exchange) GetOrderHistory(ctx context.Context, req *order.MultiOrderReq
 
 		for x := range instrumentsToUse {
 			var orders TradeHistory
-			orders, err = e.GetTradeHistory(ctx, instrumentsToUse[x], -1, -1)
+			orders, err = e.GetTradeHistory(ctx, instrumentsToUse[x], 0, 0)
```
gloriousCode, line 921: req does have a startdate, but outside of scope fix

exchanges/coinut/coinut_test.go
```diff
@@ -101,6 +109,72 @@ func TestSeedInstruments(t *testing.T) {
...
+			require.NoError(t, err, "GetTradeHistory must not error")
+			assert.Equal(t, TradeHistory{Trades: []OrderFilledResponse{}}, result, "GetTradeHistory should return the correct trade history")
+
+			timer := time.NewTimer(time.Second)
+			defer timer.Stop()
+			var request requestResult
+			select {
+			case request = <-requestC:
+			case <-timer.C:
+				t.Fatal("GetTradeHistory must send a request to the mock server")
```
gloriousCode: Remember you can use `time.After(time.Second)` with less space taken up

exchanges/coinut/coinut_test.go
```diff
@@ -101,6 +109,72 @@ func TestSeedInstruments(t *testing.T) {
...
+		{name: "omit pagination"},
+		{name: "limit only", limit: 2},
+		{name: "start only", start: 101},
+		{name: "start and limit", start: 101, limit: 2},
+		{name: "maximum start", start: math.MaxUint64},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			type requestResult struct {
```
gloriousCode: If you want to go CRAZY alice, you could t.Run("mocked" and t.Run("live" scenarios with additional t.Parallels

exchanges/coinut/coinut_test.go
```diff
@@ -250,19 +324,343 @@ func TestGetActiveOrders(t *testing.T) {
...
+			enabledPairs:           currency.Pairs{currency.NewBTCUSD()},
+			response:               invalidSideResponse,
+			wantErrorIs:            order.ErrSideIsInvalid,
+			wantRequestCount:       1,
+			wantTradeInstrumentIDs: []float64{123},
+			wantOmittedPagination:  true,
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			var (
```
gloriousCode, line 458: t.Parallel ?
If its due to server stuff, declare server outside this and t.Parallel here?

exchanges/coinut/coinut_test.go
```diff
@@ -250,19 +324,343 @@ func TestGetActiveOrders(t *testing.T) {
...
+		{
+			name:        "websocket invalid order side",
+			mode:        "invalid side",
+			instruments: map[string]int64{"BTCUSD": 123},
+			wantErrorIs: order.ErrSideIsInvalid,
+			wantStarts:  []int64{0},
+		},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			var (
```
gloriousCode, line 572: t.Parallel?

gloriousCode (approved):
I approve, but I think it can still be better

exchanges/coinut/coinut_test.go
thrasher-: Alice: I tested the parallel form, but it is not safe with Coinut's current WebSocket lifecycle. WsConnect writes to a package-global channels map, causing concurrent Exchange instances to race, and the fixture must disable the manager before shutdown to prevent reconnects after its mock server closes. Hoisting the server would introduce shared mutable fixture state and would not solve either issue. Those production and lifecycle fixes should be handled separately, so I’m keeping these cases sequential in this PR.

exchanges/coinut/coinut_test.go
thrasher-: Alice: I tested the parallel form, but it is not safe with Coinut's current WebSocket lifecycle. WsConnect writes to a package-global channels map, causing concurrent Exchange instances to race, and the fixture must disable the manager before shutdown to prevent reconnects after its mock server closes. Hoisting the server would introduce shared mutable fixture state and would not solve either issue. Those production and lifecycle fixes should be handled separately, so I’m keeping these cases sequential in this PR.

exchanges/coinut/coinut_test.go
thrasher-: Alice: I tested the parallel form, but it is not safe with Coinut's current WebSocket lifecycle. WsConnect writes to a package-global channels map, causing concurrent Exchange instances to race, and the fixture must disable the manager before shutdown to prevent reconnects after its mock server closes. Hoisting the server would introduce shared mutable fixture state and would not solve either issue. Those production and lifecycle fixes should be handled separately, so I’m keeping these cases sequential in this PR.

exchanges/coinut/coinut_test.go
```diff
@@ -101,6 +103,64 @@ func TestSeedInstruments(t *testing.T) {
...
+			ex.SkipAuthCheck = true
+			require.NoError(t, ex.SetHTTPClient(server.Client()), "Setting the HTTP client must not error")
+			require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL), "Setting the REST endpoint must not error")
+
+			result, err := ex.GetTradeHistory(t.Context(), 123, tc.start, tc.limit)
+			require.NoError(t, err, "GetTradeHistory must not error")
+			assert.Equal(t, TradeHistory{Trades: []OrderFilledResponse{}}, result, "Trade history should match the response")
+
+			request := <-requestC
+			require.NoError(t, request.err, "Decoding the request must not error")
```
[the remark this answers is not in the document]
thrasher-: Alice: Addressed in https://github.com/thrasher-corp/gocryptotrader/commit/7caad7c6e83a1b687c4216081d32e87d2005dfe8 with a one-second bounded wait, so a missing fixture request fails deterministically instead of blocking.

exchanges/coinut/coinut_test.go
thrasher-: Alice: Applied in https://github.com/thrasher-corp/gocryptotrader/commit/d95442c9d32cd18cd0c58d7fdf3a986b14b25923: the timeout now uses `time.After(time.Second)`. Thanks.

exchanges/coinut/coinut_test.go
thrasher-: Alice: I’ve kept this fully mocked so it can assert the exact outbound `start`/`limit` payload. A live case cannot inspect that request and would only add network and credential dependencies.

exchanges/coinut/coinut_wrapper.go
thrasher-: Alice: Agreed this is separate. Here, `start` is the exchange pagination offset; `StartTime` and `EndTime` are applied to returned orders by `req.Filter`, while REST history pagination is a pre-existing concern outside this PR.

shazbert (approved):
ch ACK!

exchanges/coinut/coinut_test.go
```diff
@@ -101,6 +109,72 @@ func TestSeedInstruments(t *testing.T) {
...
+			require.NoError(t, ex.SetHTTPClient(server.Client()), "Setting the HTTP client must not error")
+			require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL), "Setting the REST endpoint must not error")
+
+			result, err := ex.GetTradeHistory(t.Context(), 123, tc.start, tc.limit)
+			require.NoError(t, err, "GetTradeHistory must not error")
+			assert.Equal(t, TradeHistory{Trades: []OrderFilledResponse{}}, result, "GetTradeHistory should return the correct trade history")
+
+			timer := time.NewTimer(time.Second)
+			defer timer.Stop()
+			var request requestResult
```
gloriousCode: rename var due to package collision

exchanges/coinut/coinut_test.go
gloriousCode: Not wanting to test due to API keys despite our testing files all supporting that is not a choice I would normally accept. The only reason(s) I'm not going to block you on it is that the rest of coinut is untested (and that coinut times out when I tested it)

gloriousCode (changes requested):

exchanges/coinut/coinut_test.go
thrasher-: Alice: Renamed the local variable to `capturedRequest` in https://github.com/thrasher-corp/gocryptotrader/commit/6fc42729d1130ff3614d410dc7dee825a5197626 to avoid the package-name collision.

exchanges/coinut/coinut_test.go
thrasher-: Alice: Added parallel `mocked` and credential-gated `live` scenarios in https://github.com/thrasher-corp/gocryptotrader/commit/78f8f16a6c0276795e3c700440e0c339f5ac185f. The mocked matrix still asserts the exact outbound pagination payload, while the live case uses `start` 101 and `limit` 2 and checks that the returned page honours the requested limit.

gloriousCode (approved):
tACK Thanks @thrasher- for making those changes

PR #2302, opened 2026-07-27
currency: reduce pair join allocations

thrasher-:
## Summary

- build comma-separated currency-pair lists directly into one pre-sized buffer
- preserve empty, single, formatted, and partially empty pair behaviour
- add direct branch coverage and a reproducible seven-pair benchmark

## Performance

Measured `Pairs.Join` for a deterministic seven-pair workload with an identical harness, one pinned logical CPU, `GOMAXPROCS=1`, and 10 alternating observations per revision:

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| time/op | 477.8 ns ±14% | 135.9 ns ±8% | -71.55% (p=0.000) |
| B/op | 232 B/op | 64 B/op | -72.41% |
| allocs/op | 9 allocs/op | 1 alloc/op | -88.89% |

The ± values are Benchstat confidence intervals for the medians.

- Relevance: this isolates the production `Pairs.Join` operation used to build comma-separated pair lists.
- Related checks: `BenchmarkPairsString`, `BenchmarkPairsFormat`, and `BenchmarkRemovePairsByFilter` had no significant timing difference detected at n=10.
- Limitations: single-host microbenchmark of one deterministic workload; it does not establish aggregate call frequency, end-to-end throughput, or tail latency.

## Validation

- `go test ./currency -run '^TestPairsJoin$' -count=100`
- `go test ./currency -count=1`
- `go test -race ./currency -count=1`
- `make lint`
- `codespell`
- `make misc_checks`

### DIFF
--- a/currency/pairs.go
+++ b/currency/pairs.go
@@ -63,13 +63,14 @@ func (p Pairs) Join() string {
 	}
 
 	// Build directly to avoid allocating an intermediate string for each pair.
-	size := len(p) - 1
+	// len(p) - 1 accounts for the commas inserted between pairs.
+	outputLen := len(p) - 1
 	for i := range p {
-		size += len(p[i].Base.String()) + len(p[i].Delimiter) + len(p[i].Quote.String())
+		outputLen += len(p[i].Base.String()) + len(p[i].Delimiter) + len(p[i].Quote.String())
 	}
 
 	var joined strings.Builder
-	joined.Grow(size)
+	joined.Grow(outputLen)
 	for i := range p {
 		if i > 0 {
 			joined.WriteByte(',')
--- a/currency/pairs.go
+++ b/currency/pairs.go
@@ -55,7 +55,30 @@ func (p Pairs) Strings() []string {
 
 // Join returns a comma separated list of currency pairs
 func (p Pairs) Join() string {
-	return strings.Join(p.Strings(), ",")
+	switch len(p) {
+	case 0:
+		return ""
+	case 1:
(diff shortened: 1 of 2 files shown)


thrasher-:
@codex please review

currency/pairs.go
```diff
@@ -55,7 +55,30 @@ func (p Pairs) Strings() []string {
...
-	return strings.Join(p.Strings(), ",")
+	switch len(p) {
+	case 0:
+		return ""
+	case 1:
+		return p[0].String()
+	}
+
+	// Build directly to avoid allocating an intermediate string for each pair.
+	size := len(p) - 1
```
shazbert: suggestion to change size to capacity and comment that the `len(p) - 1` is the inserted commas between.

shazbert (changes requested):
👍

currency/pairs.go
thrasher-: Iris: Thanks. Making the separator count explicit is useful. I suggest renaming `size` to `outputLen` and commenting that `len(p) - 1` accounts for the commas between pairs. `outputLen` is more precise than `capacity` here because the calculation is the exact final byte length passed to `strings.Builder.Grow`.

gloriousCode: Yep, once this is done I'll approve too

shazbert (approved):
Thanks, Changes ACK!

gloriousCode (approved):
`return "t,A,C,K,!"`

PR #2305, opened 2026-07-29
cmd/benchcheck: Add new centralised benchmark tool 

### LINKED ISSUE #2304: benchmarks: Automate current record vs a PR in a time series view

thrasher-:
# PR Description

Adds a new benchcheck tool to record benchmark result series over time, plus flags whenever it creeps upwards. It stores it in a branch daily for now since GHA cache expires after 7 days of non-use and are immutable. It also rids all comments and enforces it in misc_checks so that we all use the new centralised tool

## Type of change

## How has this been tested

- [x] go test ./... -race
- [x] golangci-lint run

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,25 @@ DRIVER ?= psql
 RACE_FLAG := $(if $(NO_RACE_TEST),,-race)
 CONFIG_FLAG = $(if $(CONFIG),-config $(CONFIG),)
 
-.PHONY: all lint lint_docker misc_checks check test build install fmt gofumpt update_deps
+.PHONY: all lint lint_docker misc_checks check test build install fmt gofumpt update_deps bench bench_update bench_trend check_bench_pkgs bench_pkg
+
+# Edit benchmarks/packages.txt to change which packages are benchmarked.
+BENCH_PKGS = $(shell go run ./cmd/benchcheck -list)
+
+# BENCH_FLAGS must stay identical between bench and bench_update, so recorded and compared values
+# are produced the same way.
+#
+# BENCH_SAMPLES must stay odd: benchcheck compares the median, and an even count has no middle
+# sample. -cpu and -p pin what a scheduler-sensitive benchmark measures, so CI and a developer's
+# machine agree; keep -cpu single valued, or `-cpu 1,4` files two populations under one key.
+BENCH_SAMPLES = 7
+BENCH_FLAGS = -run '^$$' -bench . -benchmem -benchtime 100ms -count $(BENCH_SAMPLES) -cpu 4 -p 1 -timeout 20m
+
+# BENCH_TREND_FLAGS feed the scheduled ns/op history, where tight confidence intervals do matter and
+# a long runtime is affordable. Never use these for the PR gate; they take over ten minutes.
+BENCH_TREND_SAMPLES = 15
+BENCH_TREND_FLAGS = -run '^$$' -bench . -benchmem -benchtime 1s -count $(BENCH_TREND_SAMPLES) -cpu 4 -p 1 -timeout 60m
+BENCH_SERIES = benchmarks/series.jsonl
 
 all: check build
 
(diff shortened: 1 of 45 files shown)


thrasher-:
@codex please review

Makefile
```diff
@@ -11,7 +11,33 @@ DRIVER ?= psql
...
+
+# BENCH_FLAGS gates allocations on every PR and must stay identical between bench and bench_update
+# so recorded and compared values are produced the same way. It is deliberately cheap: allocs/op is
+# per-iteration and reproducible, so a longer benchtime or a higher count buys the gate nothing.
+#
+# -cpu 4 pins GOMAXPROCS inside each test binary so a benchmark measures the same parallelism on CI
+# as on a developer's machine. It does not pin -p, the number of package binaries run concurrently,
+# which still defaults to the host's core count, so a scheduler-sensitive benchmark could still vary
+# between machines; -p 1 closes that gap by running one package binary at a time.
+BENCH_FLAGS = -run '^$$' -bench . -benchmem -benchtime 100ms -count 6 -cpu 4 -p 1 -timeout 20m
```
gloriousCode: You have a very basic median calculator function. In that sense, choosing 7 or 5 is better than choosing a non-median result!!!!!!!!!

Please use odd number samples and reject any benchmark that does not produce the same number of measurements. Suggested diff: [seven-sample validation](https://gist.github.com/gloriousCode/050763534e13789b01ae261ec1aae470).

benchmarks/packages.txt
```diff
@@ -0,0 +1,54 @@
...
+
+# Benchmarked and checked against the baseline, but not gated:
+config                     # near-deterministic; entry carries a small allocs tolerance
+exchanges/gateio           # seeded, but the package's benchmarks are not yet audited for hermeticity
+exchanges/okx              # as above
+
+# Not benchmarked at all, and why. Note that a baseline "ignore" only skips comparison - it does not
+# stop the benchmark running - so a benchmark that is unsafe to execute must be excluded here.
+#
+#   exchanges/alert           - BenchmarkWait calls Wait(nil) and never reads the reply, so each
```
gloriousCode: `BenchmarkWait` discards the reply, so every iteration leaves `hold` blocked. The history run can eventually exhaust memory.

Consume a closed-kick reply before starting the next iteration, then gate this package. Suggested diff: [benchmark exclusion fixes](https://gist.github.com/gloriousCode/7c4f7ca258bef99c8613c247039fba20).

Expected: every iteration completes its waiter and both benchmark durations run safely.

benchmarks/packages.txt
```diff
@@ -0,0 +1,54 @@
...
+#                               iteration leaks the goroutine that alert.hold documents as leakable.
+#                               Harmless at 100ms, but the scheduled 1s x 15 run accumulates millions
+#                               and can exhaust memory. Excluding the package also drops
+#                               BenchmarkAlert, which is otherwise fine
+#   exchange/websocket/buffer - every benchmark fails with "channel buffer is full: failed to relay
```
gloriousCode: These benchmarks never drain `DataHandler`, so they fail once the relay fills. The buffered cases also leave the buffer limit at zero, so they do not exercise batching.

Use an isolated relay, set a non-zero buffered limit, and drain each batch. Suggested diff: [benchmark exclusion fixes](https://gist.github.com/gloriousCode/7c4f7ca258bef99c8613c247039fba20).

Expected: the benchmarks run to completion and the buffered cases measure actual batching.

benchmarks/packages.txt
```diff
@@ -0,0 +1,54 @@
...
+#                               <*orderbook.Depth>" once it runs enough iterations. They pass at
+#                               -benchtime 10x and fail at 100ms, alone or together, so the relay
+#                               channel simply overruns rather than this being shared state
+#   exchanges/binance         - BenchmarkWsHandleData is hermetic and its output is clean, but its
```
gloriousCode: This aggregate benchmark does not take one repeatable path: depth state changes, relay draining is scheduler-dependent, and execution reports can scan enabled assets in map order.

Split the fixture paths, isolate spot state, reset depth state, and drain synchronously. Suggested diff: [benchmark exclusion fixes](https://gist.github.com/gloriousCode/7c4f7ca258bef99c8613c247039fba20).

Expected: every websocket path has a stable allocation budget and identifies its own regression.

benchmarks/packages.txt
```diff
@@ -0,0 +1,54 @@
...
+#                               counts are not reproducible: allocs/op moves between 198 and 209 and
+#                               B/op by around 20% even at a fixed iteration count, so no tolerance
+#                               gates anything useful. Would need wsHandleData itself to take the
+#                               same path every iteration
+#   exchanges/kucoin          - its TestMain calls getFirstTradablePairOfAssets unconditionally, so
```
gloriousCode: `TestMain` always refreshes pairs before this benchmark, so the allocation gate depends on the live KuCoin API.

Skip pair discovery when `-run '^$'` selects no tests, then gate the package. Suggested diff: [benchmark exclusion fixes](https://gist.github.com/gloriousCode/7c4f7ca258bef99c8613c247039fba20).

Expected: benchmark-only runs are hermetic; normal tests keep their existing pair setup.

benchmarks/packages.txt
```diff
@@ -0,0 +1,54 @@
...
+#                               the package fetches tradable pairs from the live KuCoin API even
+#                               under -run '^$'. A rate limit or outage would fail the whole gate
+#                               job, for one benchmark that allocates nothing
+#   log                       - the logger writes to stdout during measurement, which interleaves
```
gloriousCode: The enabled logger benchmarks write asynchronously to stdout and do not wait for completion. That can corrupt benchmark result lines or overflow the queue.

Give each benchmark an isolated test writer and wait for every queued write. Suggested diff: [benchmark exclusion fixes](https://gist.github.com/gloriousCode/7c4f7ca258bef99c8613c247039fba20).

Expected: logger output never reaches the benchmark stream and every measured write completes.

gloriousCode (changes requested):
Its cool and all. I just think that the introduction of a benchmark app deserves fixing broken benchmarks instead of ignoring them. The median func just dealing with odd numbers is preferable imo to making up a new number by averaging both middle numbers, or returning a non-median number like current.

PR #2306, opened 2026-07-29
exchange/websocket/buffer: reduce pending update lookup overhead

thrasher-:
## Summary

- reuse the stable orderbook holder for consecutive pending updates to the same pair and asset while reading its live update ID on every iteration
- install a newly loaded snapshot holder in one exclusive map critical section so concurrent first snapshots cannot replace a cached holder
- add deterministic backlog benchmarks and direct coverage for sequencing, errors, buffering, invalidation, mixed orderbooks, and concurrent holder creation

## Performance

The deterministic offline benchmark used Go 1.26.5 on Linux/amd64 WSL2 with an Intel Core i7-6700K, `GOMAXPROCS=1`, one pinned benchmark CPU, one-second samples, and 20 counterbalanced fresh-process observations per revision. The byte-identical fixture measures `LoadSnapshot`, `applyPendingUpdates`, and relay draining.

| Pending updates | Before | After | Change |
| ---: | ---: | ---: | ---: |
| 1 | 528.8 ns/op | 535.4 ns/op | +1.25% (p=0.007) |
| 2 | 822.2 ns/op | 797.9 ns/op | -2.97% (p=0.000) |
| 8 | 2.551 µs/op | 2.327 µs/op | -8.80% (p=0.000) |
| 64 | 18.58 µs/op | 16.66 µs/op | -10.30% (p=0.000) |
| 256 | 74.13 µs/op | 65.87 µs/op | -11.14% (p=0.000) |

Bytes and allocations were unchanged in every case. The 64-update workload remained at 3120 B/op and 65 allocs/op; the 256-update workload remained at 12336 B/op and 257 allocs/op.

The statistically detected 1.25% single-update slowdown is disclosed as a small guard regression. The primary 64- and 256-update backlogs retain double-digit latency improvements without allocation movement. The baseline CPU profile attributed 16.96% cumulative time to the repeated `Orderbook.LastUpdateID` wrapper lookup; that wrapper no longer appeared in the candidate profile's top entries.

## Validation

- direct tests cover both changed functions at 100% statement coverage, including concurrent first snapshots and all reachable error paths
- focused tests passed 100 repetitions and 10 race-detector repetitions
- package and affected caller tests passed repeated and race-detector runs
- the complete repository race-detector suite passed
- the ordinary repository suite had one GateIO live request return HTTP 504; that exact test passed immediately on focused retry
- formatting, lint, spelling, and miscellaneous repository policy checks passed

## Limitations

- this microbenchmark measures a deterministic orderbook resynchronisation backlog, not end-to-end websocket throughput or network latency
- CPU governor, external power state, and thermal sensors were unavailable in the WSL2 environment
- the one-update workload is 1.25% slower even though the declared primary backlog workloads improve materially

### DIFF
--- a/exchange/websocket/buffer/buffer.go
+++ b/exchange/websocket/buffer/buffer.go
@@ -53,20 +53,25 @@ func (o *Orderbook) LoadSnapshot(book *orderbook.Book) error {
 	}
 
 	bookKey := key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}
-	o.m.Lock()
+	o.m.RLock()
 	holder, ok := o.ob[bookKey]
+	o.m.RUnlock()
 	if !ok {
-		// Associate orderbook pointer with local exchange depth map
-		depth, err := orderbook.DeployDepth(book.Exchange, book.Pair, book.Asset)
-		if err != nil {
-			o.m.Unlock()
-			return err
+		o.m.Lock()
+		holder, ok = o.ob[bookKey]
+		if !ok {
+			// Associate orderbook pointer with local exchange depth map
+			depth, err := orderbook.DeployDepth(book.Exchange, book.Pair, book.Asset)
+			if err != nil {
+				o.m.Unlock()
+				return err
+			}
+			depth.AssignOptions(book)
+			holder = &orderbookHolder{ob: depth, buffer: make([]orderbook.Update, 0, o.obBufferLimit)}
+			o.ob[bookKey] = holder
 		}
-		depth.AssignOptions(book)
-		holder = &orderbookHolder{ob: depth, buffer: make([]orderbook.Update, 0, o.obBufferLimit)}
-		o.ob[bookKey] = holder
+		o.m.Unlock()
 	}
-	o.m.Unlock()
 
 	book.RestSnapshot = false
 	if err := holder.ob.LoadSnapshot(book); err != nil {
@@ -86,7 +91,11 @@ func (o *Orderbook) Update(u *orderbook.Update) error {
 	if !ok {
 		return fmt.Errorf("%w for Exchange %s CurrencyPair: %s AssetType: %s", orderbook.ErrDepthNotFound, o.exchangeName, u.Pair, u.Asset)
 	}
+	return o.updateHolder(holder, u)
+}
 
+// updateHolder avoids repeating the map lookup when the caller already has the holder for the update key.
+func (o *Orderbook) updateHolder(holder *orderbookHolder, u *orderbook.Update) error {
 	if o.bufferEnabled {
 		if processed, err := o.processBufferUpdate(holder, u); err != nil || !processed {
 			return err
--- a/exchange/websocket/buffer/buffer.go
+++ b/exchange/websocket/buffer/buffer.go
@@ -52,11 +52,10 @@ func (o *Orderbook) LoadSnapshot(book *orderbook.Book) error {
 		return err
 	}
 
-	o.m.RLock()
-	holder, ok := o.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}]
-	o.m.RUnlock()
+	bookKey := key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}
+	o.m.Lock()
+	holder, ok := o.ob[bookKey]
 	if !ok {
-		o.m.Lock()
 		// Associate orderbook pointer with local exchange depth map
 		depth, err := orderbook.DeployDepth(book.Exchange, book.Pair, book.Asset)
 		if err != nil {
@@ -65,9 +64,9 @@ func (o *Orderbook) LoadSnapshot(book *orderbook.Book) error {
 		}
 		depth.AssignOptions(book)
 		holder = &orderbookHolder{ob: depth, buffer: make([]orderbook.Update, 0, o.obBufferLimit)}
-		o.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}] = holder
-		o.m.Unlock()
+		o.ob[bookKey] = holder
 	}
+	o.m.Unlock()
 
 	book.RestSnapshot = false
 	if err := holder.ob.LoadSnapshot(book); err != nil {
--- a/exchange/websocket/buffer/buffer_test.go
+++ b/exchange/websocket/buffer/buffer_test.go
@@ -425,6 +425,86 @@ func TestRunUpdateWithoutAnyUpdates(t *testing.T) {
 	require.ErrorIs(t, err, orderbook.ErrEmptyUpdate)
 }
 
+func TestUpdateHolder(t *testing.T) {
+	t.Parallel()
+
+	for _, tc := range []struct {
+		name          string
+		bufferEnabled bool
+		bufferLimit   int
+		allowEmpty    bool
+		updateID      int64
+		relayFull     bool
+		wantUpdateID  int64
+		wantBufferLen int
+		wantRelayLen  int
+		wantErr       error
+		wantErrText   string
+		wantLastIDErr error
+	}{
+		{name: "unbuffered", allowEmpty: true, updateID: 69421, wantUpdateID: 69421, wantRelayLen: 1},
+		{name: "unbuffered invalid", wantErr: orderbook.ErrEmptyUpdate, wantLastIDErr: orderbook.ErrOrderbookInvalid},
+		{name: "buffered pending", bufferEnabled: true, bufferLimit: 2, allowEmpty: true, updateID: 69421, wantUpdateID: 69420, wantBufferLen: 1},
+		{name: "buffered applied", bufferEnabled: true, bufferLimit: 1, allowEmpty: true, updateID: 69421, wantUpdateID: 69421, wantRelayLen: 1},
+		{name: "buffered invalid", bufferEnabled: true, bufferLimit: 1, wantErr: orderbook.ErrEmptyUpdate, wantLastIDErr: orderbook.ErrOrderbookInvalid},
+		{name: "relay full", allowEmpty: true, updateID: 69421, relayFull: true, wantUpdateID: 69421, wantRelayLen: 1, wantErrText: "channel buffer is full"},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+
+			cp, err := getExclusivePair()
+			require.NoError(t, err, "getExclusivePair must not error")
+			relay := stream.NewRelay(1)
+			obl := &Orderbook{exchangeName: exchangeName, ob: make(map[key.PairAsset]*orderbookHolder), dataHandler: relay}
+			err = obl.LoadSnapshot(&orderbook.Book{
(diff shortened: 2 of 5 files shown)


thrasher-:
@codex please review

exchange/websocket/buffer/ob_update_manager.go
```diff
@@ -241,8 +241,29 @@ func (m *UpdateManager) syncOrderbook(ctx context.Context, cache *updateCache, p
...
 func (m *UpdateManager) applyPendingUpdates(cache *updateCache) error {
 	var updated bool
+	var lastKey key.PairAsset
+	var lastHolder *orderbookHolder
 	for _, data := range cache.updates {
-		bookLastUpdateID, err := m.ob.LastUpdateID(data.update.Pair, data.update.Asset)
+		currentKey := key.PairAsset{Base: data.update.Pair.Base.Item, Quote: data.update.Pair.Quote.Item, Asset: data.update.Asset}
+		// Read the live update ID on every iteration because buffering or skipped
+		// updates can leave it unchanged, while avoiding repeated map lookups.
+		if lastHolder == nil || currentKey != lastKey {
```
shazbert: If the pending update keys are not the same then that is an error because something really weird is going on. Goblins probably.

shazbert (changes requested):
Only one thing I saw

exchange/websocket/buffer/ob_update_manager.go
thrasher-: Iris: Agreed — goblins should fail loudly here. An `updateCache` is scoped to one pair/asset, so https://github.com/thrasher-corp/gocryptotrader/commit/09b8fa6e4157416f7d1a09f7d30c151f61ea913d updates `applyPendingUpdates` to return a wrapped `errPendingUpdateKeyMismatch` with the expected and actual keys instead of switching holders. Direct tests cover both pair and asset mismatches.

exchange/websocket/buffer/buffer_test.go
```diff
@@ -473,6 +474,49 @@ func TestLoadSnapshot(t *testing.T) {
...
+				Pair:         cp,
+				Asset:        asset.Spot,
+				LastUpdated:  time.Unix(int64(i+1), 0),
+				LastUpdateID: int64(i + 1),
+			})
+			obl.m.RLock()
+			holders[i] = obl.ob[bookKey]
+			obl.m.RUnlock()
+		})
+	}
```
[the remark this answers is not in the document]
thrasher-: Iris: Thanks for flagging this. This module targets Go 1.26. Since Go 1.22, variables declared by a `range` clause are scoped per iteration, and `for i := range snapshotCount` uses those semantics. Each goroutine therefore captures a distinct `i` and writes to its own `errs[i]`/`holders[i]` slot. Adding `i := i` would be redundant here, so no code change is needed. See the Go 1.22 language changes: https://go.dev/doc/go1.22#language

exchange/websocket/buffer/buffer.go
```diff
@@ -52,11 +52,10 @@ func (o *Orderbook) LoadSnapshot(book *orderbook.Book) error {
 		return err
 	}
 
-	o.m.RLock()
-	holder, ok := o.ob[key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}]
-	o.m.RUnlock()
+	bookKey := key.PairAsset{Base: book.Pair.Base.Item, Quote: book.Pair.Quote.Item, Asset: book.Asset}
+	o.m.Lock()
```
gloriousCode: Could we retain the shared read path and only take the exclusive lock when the holder is absent?

The exclusive lock prevents concurrent first snapshots from creating different holders, but it also serialises every existing-holder lookup. In the isolated benchmark this path is 4.95% slower than master, appearing slower in 19/20 paired runs.

Rechecking after acquiring the exclusive lock provides the same creation guarantee: if another goroutine created the holder while we were waiting, we reuse it. The latest readable attempt keeps this part unchanged.

[Exact patch against reviewed commit `09b8fa6e4`](https://gist.github.com/gloriousCode/5318a5c8112f9bbedb1f1c5803c00416)

| Existing-snapshot path | Median | vs master | Paired evidence |
|---|---:|---:|---:|
| Master | 104.95 ns | — | Baseline |
| PR #2306 | 110.15 ns | **+4.95% slower** | Slower in 19/20 runs |
| Revised attempt | **105.55 ns** | +0.57% | Statistically indistinguishable from master |

The revised path recovers 4.60 ns, or approximately **88% of PR #2306's existing-snapshot regression**, while preserving stable concurrent holder creation.

Expected result: concurrent first snapshots still retain and publish one holder, while existing snapshots remain on the shared-lock fast path.

exchange/websocket/buffer/ob_update_manager.go
```diff
@@ -241,8 +242,31 @@ func (m *UpdateManager) syncOrderbook(ctx context.Context, cache *updateCache, p
...
+			if !ok {
+				return fmt.Errorf("%s %w: %s.%s", m.ob.exchangeName, orderbook.ErrDepthNotFound, data.update.Asset, data.update.Pair)
+			}
+			pendingKey = currentKey
+		} else if currentKey != pendingKey {
+			return fmt.Errorf("%w: expected %s.%s, got %s.%s", errPendingUpdateKeyMismatch, pendingKey.Asset, pendingKey.Pair(), currentKey.Asset, currentKey.Pair())
+		}
+		// Read the live update ID on every iteration because buffering or skipped
+		// updates can leave it unchanged, while avoiding repeated map lookups.
+		bookLastUpdateID, err := holder.ob.LastUpdateID()
```
gloriousCode, line 274: Could we separate queue validation from update application, then reuse the validated holder?

The current loop discovers the holder on its first iteration, checks later keys as it goes, and calls `m.ob.Update`, which locks and searches the map again. That interleaves three responsibilities and can publish earlier updates before discovering a mismatched key later in the queue.

The revised flow is easier to follow:

1. Validate the first update and every queued key.
2. Look up the holder once.
3. Apply updates through that holder.

A malformed mixed-key queue therefore fails before any mutation or publication, while the normal path avoids the repeated map lookup.

[Exact production and test patch against reviewed commit `09b8fa6e4`](https://gist.github.com/gloriousCode/7b41be3003b6b747c3eb85e85e8c2238)

The corresponding test pins the new fail-before-mutation guarantee:

The dedicated helper coverage and fail-before-mutation assertions are included in the [same exact patch](https://gist.github.com/gloriousCode/7b41be3003b6b747c3eb85e85e8c2238).

Eight fresh-process observations per version against PR #2306's exact hot paths produced:

| Pending updates | PR #2306 | Revised attempt | Improvement |
|---:|---:|---:|---:|
| 64 | 6.524 µs | **5.793 µs** | **11.21% faster** |
| 256 | 25.841 µs | **22.822 µs** | **11.68% faster** |

Bytes and allocations were unchanged. In a separate immediate A/B against the first holder-reuse attempt, this linear version was also 1.3% faster at 64 updates and 3.1% faster at 256 updates.

Expected result: the public `Update` behaviour remains unchanged, malformed queues fail before publishing partial state, and the valid pending-update path performs one map lookup for the whole queue.

gloriousCode (changes requested):
This is a nice PR. But I think it could go further!

The holder-caching direction is good, and the additional sequencing/concurrency coverage is valuable. I think two follow-up changes would preserve those correctness improvements while avoiding the remaining lookup overhead and the existing-snapshot regression:

1. Keep the shared read-lock path for holders that already exist, taking the exclusive lock only after a miss and rechecking before creation.
2. Reuse the holder that `applyPendingUpdates` already validated instead of asking `Orderbook.Update` to find it again.

Twenty counterbalanced fresh-process observations on Go 1.26.3 produced:

| Workload | Master | PR #2306 | Proposed | Proposed vs #2306 |
|---:|---:|---:|---:|---:|
| 1 update | 227.7 ns | 230.4 ns | **222.8 ns** | **-3.30%** |
| 64 updates | 7.975 µs | 7.412 µs | **6.729 µs** | **-9.21%** |
| 256 updates | 31.556 µs | 29.202 µs | **26.710 µs** | **-8.53%** |
| Existing snapshot | 104.95 ns | 110.15 ns | **105.55 ns** | **-4.18%** |

Bytes and allocations were unchanged. The proposed version also reached 100% buffer-package statement coverage, passed ten race repetitions, and passed repository lint and miscellaneous checks.

This PR finds the holder once for the ID check, but asks the map for the same holder again when applying the update. It also locks the whole holder map for writing when most snapshots only need to read it. The suggested changes avoid both costs while preserving the same safety checks.

exchange/websocket/buffer/buffer.go
thrasher-: Iris: Implemented in [e9ac751e5fe91ddb9355168f9071a8fa3fc8c867](https://github.com/thrasher-corp/gocryptotrader/commit/e9ac751e5fe91ddb9355168f9071a8fa3fc8c867). `LoadSnapshot` now checks under `RLock`, returns immediately for an existing holder, then takes `Lock` and rechecks before creating after a miss. The concurrent holder-stability test still passes. Across 20 counterbalanced PR-head/follow-up runs, the existing-holder path improved from 249.3 ns/op to 241.8 ns/op (-3.03%, p<0.001), with bytes and allocations unchanged.

exchange/websocket/buffer/ob_update_manager.go
thrasher-: Iris: Implemented in [e9ac751e5fe91ddb9355168f9071a8fa3fc8c867](https://github.com/thrasher-corp/gocryptotrader/commit/e9ac751e5fe91ddb9355168f9071a8fa3fc8c867). `applyPendingUpdates` now validates the entire queue before any lookup or mutation, looks up the holder once, and applies each update through that validated holder. Direct pair- and asset-mismatch tests confirm there is no partial holder mutation or relay publication. Across 20 counterbalanced PR-head/follow-up runs, 64 updates improved from 16.89 µs/op to 15.21 µs/op (-9.92%) and 256 from 66.99 µs/op to 60.64 µs/op (-9.48%), both p<0.001, with bytes and allocations unchanged.

exchange/websocket/buffer/ob_update_manager.go
```diff
@@ -240,9 +241,36 @@ func (m *UpdateManager) syncOrderbook(ctx context.Context, cache *updateCache, p
 // applyPendingUpdates applies all pending updates to the orderbook
 // assumes lock already active on cache
 func (m *UpdateManager) applyPendingUpdates(cache *updateCache) error {
+	if len(cache.updates) == 0 {
+		return errPendingUpdatesNotApplied
```
shazbert: Maybe this error should be `errUpdatesNotSupplied`?

shazbert (changes requested):
One thing from me; other than this I can ack it.

exchange/websocket/buffer/ob_update_manager.go
thrasher-: Iris: Implemented in [2403312b5ee899e656cc0428af446fbc0e1ec731](https://github.com/thrasher-corp/gocryptotrader/commit/2403312b5ee899e656cc0428af446fbc0e1ec731). An empty pending-update queue now returns `errUpdatesNotSupplied`; the distinct non-empty/all-skipped path continues to return `errPendingUpdatesNotApplied`, with direct sentinel coverage for both.

shazbert (approved):
ch ACK! Thank you very much 🤖

gloriousCode (approved):
tACK!

PR #2311, opened 2026-08-02
websocket: update parseBinaryResponse, remove DEFLATE

evgmalkov:
# PR Description

`parseBinaryResponse` currently assumes that any binary websocket frame which is not GZIP is a raw
DEFLATE stream, and pipes it through `flate.NewReader` unconditionally. That assumption does not hold
for every venue — some push **uncompressed** binary payloads (protobuf, for example). For those
exchanges the frame is corrupted inside the shared connection code before the exchange handler ever
sees it, and the failure surfaces as an unparsable message rather than as a decoding problem.

This adds an optional `ConnectionSetup.BinaryMessageDecoder`, letting an exchange take over binary
frame handling for its own connections:

```go
BinaryMessageDecoder func([]byte) ([]byte, error)
```

When the field is `nil` the existing code path runs verbatim, so exchanges that do not opt in are
unaffected — the change is additive and the default behaviour is unchanged.

The alternative — teaching the shared decoder about individual venues, or relaxing its heuristics —
would change behaviour for every exchange in order to accommodate one, which seemed worse than
handing the venue-specific case back to the venue.

Fixes # (no issue filed)

## Type of change

## How has this been tested

- [x] `go test ./... -race` (run for `./exchange/...` and `./exchanges/order/...`; full-tree runs
      require exchange credentials we do not have)
- [x] `gofmt -l` clean on the changed files
- [x] `go vet ./exchange/websocket/`
- [x] `TestParseBinaryResponseWithDecoder` — decoder is used when set, its error is propagated, and
      the default DEFLATE path still applies when no decoder is supplied
- [x] `TestCreateConnectionFromSetupPassesBinaryMessageDecoder` — the field reaches the connection,
      and stays nil when not supplied

The change has been running against a live MEXC spot/futures websocket feed in a downstream fork,
where the venue sends uncompressed protobuf frames: with the decoder attached the feed parses
correctly, and the shared default is left untouched for every other exchange.

## Checklist

- [ ] I have made corresponding changes to the documentation and regenerated documentation via the
      documentation tool — no user-facing documentation covers this internal hook; happy to add some
      if you would like it documented
- [x] New and existing unit tests pass locally with my changes

### DIFF
--- a/exchange/websocket/connection.go
+++ b/exchange/websocket/connection.go
@@ -2,7 +2,6 @@ package websocket
 
 import (
 	"bytes"
-	"compress/flate"
 	"compress/gzip"
 	"context"
 	"errors"
@@ -318,21 +317,18 @@ func (c *connection) ReadMessage() Response {
 
 // parseBinaryResponse parses a websocket binary response into a usable byte array
 func (c *connection) parseBinaryResponse(resp []byte) ([]byte, error) {
-	var reader io.ReadCloser
-	var err error
-	if len(resp) >= 2 && resp[0] == 31 && resp[1] == 139 { // Detect GZIP
-		reader, err = gzip.NewReader(bytes.NewReader(resp))
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		reader = flate.NewReader(bytes.NewReader(resp))
+	if len(resp) < 2 || resp[0] != 0x1f || resp[1] != 0x8b {
+		return resp, nil // non-GZIP response, return as-is
 	}
-	standardMessage, err := io.ReadAll(reader)
+	reader, err := gzip.NewReader(bytes.NewReader(resp))
 	if err != nil {
 		return nil, err
 	}
-	return standardMessage, reader.Close()
+	msg, readErr := io.ReadAll(reader)
+	if err := errors.Join(readErr, reader.Close()); err != nil {
+		return nil, err
+	}
+	return msg, nil
 }
 
 // Shutdown shuts down and closes specific connection
--- a/exchange/websocket/connection.go
+++ b/exchange/websocket/connection.go
@@ -324,11 +324,11 @@ func parseBinaryResponse(resp []byte) ([]byte, error) {
 	if err != nil {
 		return nil, err
 	}
-	msg, readErr := io.ReadAll(reader)
-	if err := errors.Join(readErr, reader.Close()); err != nil {
+	msg, err := io.ReadAll(reader)
+	if err != nil {
 		return nil, err
 	}
-	return msg, nil
+	return msg, reader.Close()
 }
 
 // Shutdown shuts down and closes specific connection
--- a/exchange/websocket/connection.go
+++ b/exchange/websocket/connection.go
@@ -315,7 +315,7 @@ func (c *connection) ReadMessage() Response {
 	return Response{Raw: standardMessage, Type: mType}
 }
 
-// parseBinaryResponse parses a websocket binary response into a usable byte array
+// parseBinaryResponse decompresses GZIP frames and returns all other payloads unchanged
 func parseBinaryResponse(resp []byte) ([]byte, error) {
 	if len(resp) < 2 || resp[0] != 0x1f || resp[1] != 0x8b {
 		return resp, nil // non-GZIP response, return as-is
--- a/exchange/websocket/connection.go
+++ b/exchange/websocket/connection.go
@@ -303,7 +303,7 @@ func (c *connection) ReadMessage() Response {
 	case gws.TextMessage:
 		standardMessage = resp
 	case gws.BinaryMessage:
-		standardMessage, err = c.parseBinaryResponse(resp)
+		standardMessage, err = parseBinaryResponse(resp)
 		if err != nil {
 			log.Errorf(log.WebsocketMgr, "%v %v: Parse binary response error: %v", c.ExchangeName, removeURLQueryString(c.URL), err)
 			return Response{Raw: []byte(``)} // Non-nil response to avoid the reader returning on this case.
@@ -316,7 +316,7 @@ func (c *connection) ReadMessage() Response {
 }
 
 // parseBinaryResponse parses a websocket binary response into a usable byte array
-func (c *connection) parseBinaryResponse(resp []byte) ([]byte, error) {
+func parseBinaryResponse(resp []byte) ([]byte, error) {
 	if len(resp) < 2 || resp[0] != 0x1f || resp[1] != 0x8b {
 		return resp, nil // non-GZIP response, return as-is
 	}
--- a/exchange/websocket/connection.go
+++ b/exchange/websocket/connection.go
@@ -325,10 +325,13 @@ func parseBinaryResponse(resp []byte) ([]byte, error) {
 		return nil, err
 	}
 	msg, err := io.ReadAll(reader)
+	if closeErr := reader.Close(); err == nil {
+		err = closeErr
+	}
 	if err != nil {
 		return nil, err
 	}
-	return msg, reader.Close()
+	return msg, nil
 }
 
 // Shutdown shuts down and closes specific connection
--- a/exchange/websocket/manager_test.go
+++ b/exchange/websocket/manager_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"net/http/httptest"
 	"strconv"
@@ -1248,29 +1249,81 @@ func TestParseBinaryResponse(t *testing.T) {
 	require.NoError(t, err, "captured MEXC frame must decode")
 
 	testCases := []struct {
-		name    string
-		input   []byte
-		expect  []byte
-		wantErr bool
+		name   string
+		input  []byte
+		expect []byte
+		err    error
 	}{
 		{name: "gzip", input: gzipBuffer.Bytes(), expect: []byte("hello")},
-		{name: "uncompressed MEXC protobuf", input: mexcFrame, expect: mexcFrame},
+		{name: "uncompressed MEXC protobuf", input: mexcFrame, expect: bytes.Clone(mexcFrame)},
 		{name: "unrecognised binary", input: []byte{0x01, 0x02, 0x03}, expect: []byte{0x01, 0x02, 0x03}},
-		{name: "empty"},
-		{name: "corrupt gzip header", input: []byte{0x1f, 0x8b, 0x08}, wantErr: true},
-		{name: "corrupt gzip payload", input: corruptGZIPPayload, wantErr: true},
+		{name: "nil"},
+		{name: "empty frame", input: []byte{}, expect: []byte{}},
+		{name: "single byte", input: []byte{0x1f}, expect: []byte{0x1f}},
+		{name: "partial gzip magic", input: []byte{0x1f, 0x00}, expect: []byte{0x1f, 0x00}},
+		{name: "corrupt gzip header", input: []byte{0x1f, 0x8b, 0x08}, err: io.ErrUnexpectedEOF},
+		{name: "corrupt gzip payload", input: corruptGZIPPayload, err: gzip.ErrChecksum},
 	}
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
 			resp, err := parseBinaryResponse(tc.input)
-			if tc.wantErr {
-				assert.Error(t, err, "parseBinaryResponse should return an error")
+			if tc.err != nil {
+				assert.ErrorIs(t, err, tc.err, "parseBinaryResponse should return the expected error")
+				assert.Nil(t, resp, "parseBinaryResponse should not return a payload with an error")
 				return
 			}
 			require.NoError(t, err, "parseBinaryResponse must not error")
 			assert.Equal(t, tc.expect, resp, "parseBinaryResponse should return the expected response")
+		})
+	}
+}
+
+// TestReadMessageBinary ensures binary frames reach the caller as the exchange handler expects
+func TestReadMessageBinary(t *testing.T) {
+	t.Parallel()
+
+	mock := httptest.NewServer(mockws.CurryWsMockUpgrader(t, func(_ testing.TB, p []byte, c *gws.Conn) error {
+		return c.WriteMessage(gws.BinaryMessage, p)
+	}))
+	defer mock.Close()
+
+	wc := &connection{
+		ExchangeName:     "test",
+		URL:              "ws" + mock.URL[len("http"):],
+		ResponseMaxLimit: time.Second * 5,
+		Match:            NewMatch(),
+	}
+	require.NoError(t, wc.Dial(t.Context(), &gws.Dialer{}, http.Header{}, nil), "Dial must not error")
+	defer func() { assert.NoError(t, wc.Connection.Close(), "Close should not error") }()
+
+	var gzipBuffer bytes.Buffer
+	g := gzip.NewWriter(&gzipBuffer)
+	_, err := g.Write([]byte("hello"))
+	require.NoError(t, err, "gzip.Write must not error")
+	require.NoError(t, g.Close(), "gzip.Close must not error")
+
+	mexcFrame, err := hex.DecodeString("0a3473706f74407075626c69632e61676772652e626f6f6b5469636b657" +
+		"22e76332e6170692e7062403130306d73404b4153555344541a074b41535553445430d4d882c2fc33da1335" +
+		"0a08302e3032363239331204313430371a08302e30323633353722053237372e352a0b3133303438393134" +
+		"3133393090d482c2fc33")
+	require.NoError(t, err, "captured MEXC frame must decode")
+
+	for _, tc := range []struct {
+		name   string
+		send   []byte
+		expect []byte
+	}{
+		{name: "gzip is decompressed", send: gzipBuffer.Bytes(), expect: []byte("hello")},
+		{name: "uncompressed protobuf is preserved", send: mexcFrame, expect: mexcFrame},
+		{name: "malformed gzip yields an empty non-nil response", send: []byte{0x1f, 0x8b, 0x08}, expect: []byte("")},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			require.NoError(t, wc.Connection.WriteMessage(gws.BinaryMessage, tc.send), "WriteMessage must not error")
+			resp := wc.ReadMessage()
+			require.NotNil(t, resp.Raw, "Raw must not be nil; a nil response stops the reader")
+			assert.Equal(t, tc.expect, resp.Raw, "ReadMessage should return the expected payload")
 		})
 	}
 }
--- a/exchange/websocket/manager_test.go
+++ b/exchange/websocket/manager_test.go
@@ -1261,11 +1261,10 @@ func TestParseBinaryResponse(t *testing.T) {
 		{name: "corrupt gzip payload", input: corruptGZIPPayload, wantErr: true},
 	}
 
-	wc := &connection{}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
-			resp, err := wc.parseBinaryResponse(tc.input)
+			resp, err := parseBinaryResponse(tc.input)
 			if tc.wantErr {
 				assert.Error(t, err, "parseBinaryResponse should return an error")
 				return
--- a/exchange/websocket/manager_test.go
+++ b/exchange/websocket/manager_test.go
@@ -2,9 +2,9 @@ package websocket
 
 import (
 	"bytes"
-	"compress/flate"
 	"compress/gzip"
 	"context"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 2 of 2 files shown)


thrasher-:
@codex please review

exchange/websocket/connection.go
```diff
@@ -318,6 +324,12 @@ func (c *connection) ReadMessage() Response {
 
 // parseBinaryResponse parses a websocket binary response into a usable byte array
 func (c *connection) parseBinaryResponse(resp []byte) ([]byte, error) {
+	// An exchange whose binary frames are not covered by the default assumption below
+	// supplies its own decoder via ConnectionSetup.BinaryMessageDecoder. When unset the
+	// behaviour is unchanged, so exchanges that do not opt in are unaffected.
+	if c.binaryMessageDecoder != nil {
+		return c.binaryMessageDecoder(resp)
+	}
```
shazbert: Can shift this to the field in the struct. Also can drop the `An exchange` as it can be used for things separate to an exchange. And drop ` When unset the behaviour is unchanged, so exchanges that do not opt in are unaffected.` because that is not needed. 
```suggestion
	if c.binaryMessageDecoder != nil {
		return c.binaryMessageDecoder(resp)
	}
```

exchange/websocket/connection.go
```diff
@@ -134,6 +139,7 @@ type connection struct {
 	ResponseMaxLimit     time.Duration
 	Traffic              chan struct{}
 	readMessageErrors    chan error
+	binaryMessageDecoder func([]byte) ([]byte, error)
```
shazbert: suggestion: Can update name with something like `optionalBinaryMessageDecoder` so that its more obvious what its doing when reading the code.

exchange/websocket/manager_test.go
```diff
@@ -1265,6 +1265,50 @@ func TestParseBinaryResponse(t *testing.T) {
 	assert.ErrorContains(t, err, "unexpected EOF", "parseBinaryResponse should error on empty input")
 }
 
+func TestParseBinaryResponseWithDecoder(t *testing.T) {
```
shazbert: Would be good to have the example frame where it fails on the default path and then once the option decoder gets assigned then gets correctly processed.

shazbert (changes requested):
Awesome, one thing to consider is that this is applied to all connection to an exchange, in some cases some asset types or auth/private endpoints are different and may have different encoding. But we can always defer that when we get to a problem. Thanks for this PR. Will check out the updated MEXC. 🚀

evgmalkov:
Thanks for the review — all four points are addressed in `20014cd7`.

- **Call the decoder directly / trim the comment.** Applied as suggested. The remaining comment only
  states what the default assumes and where it does not hold, without repeating the nil case.
- **`optionalBinaryMessageDecoder`.** Renamed, so the optionality is visible at the call site rather
  than only in the doc comment.
- **Test with a real frame.** `TestParseBinaryResponseWithDecoder` now starts from an uncompressed
  protobuf-like frame: it fails on the default path, is handled once a decoder is supplied, the
  decoder's own error is propagated, and the default DEFLATE behaviour still applies when no decoder
  is set.
- **`SetupNewConnection` emptiness check** (raised by the bot). Good catch — a setup carrying only
  `BinaryMessageDecoder` was rejected as empty, so the decoder never survived the public setup path.
  The field is now part of that check, and the passthrough is covered from the public entry point,
  not just the constructor.

On your wider point about different asset types and private endpoints potentially needing different
encoding — that is exactly the case downstream: MEXC spot and futures encode differently, and the
futures side has further problems of its own. Happy to leave it deferred as you suggested; the hook
is per-connection, so a venue that needs to distinguish them can already do so.

Regarding the MEXC PR (#2312): I am narrowing it to spot only. The futures code has defects I can
reproduce but cannot yet verify fixes for — the environment this was developed in runs without API
keys by policy, so the private paths cannot be exercised. Rather than ask you to review code I have
not been able to prove, futures will follow as a separate PR once I can test it properly.

exchange/websocket/manager_test.go
evgmalkov: Done in `347d7d67` — and thanks for pushing on this, the first version used a frame I had constructed by hand, which was weaker than it looked.

The test now uses a frame captured live from MEXC's public spot websocket (`spot@public.aggre.bookTicker.v3.api.pb@100ms@KASUSDT`, no credentials needed): 126 bytes of uncompressed protobuf, so neither GZIP nor a valid DEFLATE stream.

On the default path it fails with:

```
flate: corrupt input before offset 9
```

after emitting 6 bytes of garbage — which is exactly the case this hook exists for. With the decoder assigned it is returned intact, the decoder's own error is propagated, and clearing the decoder restores the DEFLATE behaviour on a genuinely deflated payload.

Worth noting for the record: I checked whether `flate` might silently return garbage *without* an error here, since that would have made the assertion meaningless. It does error, so the test is doing real work.

exchange/websocket/connection.go
```diff
@@ -318,6 +323,9 @@ func (c *connection) ReadMessage() Response {
 
 // parseBinaryResponse parses a websocket binary response into a usable byte array
 func (c *connection) parseBinaryResponse(resp []byte) ([]byte, error) {
```
gloriousCode: I'm not a huge fan of appending another custom hook to the websocket, especially for a single case. To avoid this, I wanted to see how we could adapt current code to support MEXC.

The problem occurs because the existing non-GZIP path treats every binary message as a raw DEFLATE stream using flate.NewReader. That behaviour was [introduced for OKEx in 2018](https://github.com/thrasher-corp/gocryptotrader/commit/c3201141671619076d28d0d7d10ea0161c7439b8). Kraken, Bittrex (RIP) and Huobi have also used it in the past. Every exchange known to have required raw DEFLATE either no longer requires it or is gone.

So the issue you're facing doesn't even need to occur anymore and we don't need to do these changes. I instead propose:
```go
// parseBinaryResponse decompresses GZIP frames and preserves unrecognised binary payloads.
func (c *connection) parseBinaryResponse(resp []byte) ([]byte, error) {
	if len(resp) < 2 || resp[0] != 0x1f || resp[1] != 0x8b {
		return resp, nil // non-GZIP response, return as-is
	}
	reader, err := gzip.NewReader(bytes.NewReader(resp))
	if err != nil {
		return nil, err
	}
	msg, readErr := io.ReadAll(reader)
	if err := errors.Join(readErr, reader.Close()); err != nil {
		return nil, err
	}
	return msg, nil
}
```

Here's tests:
```go
func TestParseBinaryResponse(t *testing.T) {
	t.Parallel()

	var gzipBuffer bytes.Buffer
	g := gzip.NewWriter(&gzipBuffer)
	_, err := g.Write([]byte("hello"))
	require.NoError(t, err, "gzip.Write must not error")
	require.NoError(t, g.Close(), "gzip.Close must not error")
	corruptGZIPPayload := bytes.Clone(gzipBuffer.Bytes())
	corruptGZIPPayload[len(corruptGZIPPayload)-1] ^= 0xff

	// Captured from MEXC's public spot websocket for
	// spot@public.aggre.bookTicker.v3.api.pb@100ms@KASUSDT.
	mexcFrame, err := hex.DecodeString("0a3473706f74407075626c69632e61676772652e626f6f6b5469636b657" +
		"22e76332e6170692e7062403130306d73404b4153555344541a074b41535553445430d4d882c2fc33da1335" +
		"0a08302e3032363239331204313430371a08302e30323633353722053237372e352a0b3133303438393134" +
		"3133393090d482c2fc33")
	require.NoError(t, err, "captured MEXC frame must decode")

	testCases := []struct {
		name    string
		input   []byte
		expect  []byte
		wantErr bool
	}{
		{name: "gzip", input: gzipBuffer.Bytes(), expect: []byte("hello")},
		{name: "uncompressed MEXC protobuf", input: mexcFrame, expect: mexcFrame},
		{name: "unrecognised binary", input: []byte{0x01, 0x02, 0x03}, expect: []byte{0x01, 0x02, 0x03}},
		{name: "empty"},
		{name: "corrupt gzip header", input: []byte{0x1f, 0x8b, 0x08}, wantErr: true},
		{name: "corrupt gzip payload", input: corruptGZIPPayload, wantErr: true},
	}

	wc := &connection{}
	for _, tc := range testCases {
			t.Parallel()
			resp, err := wc.parseBinaryResponse(tc.input)
			if tc.wantErr {
				assert.Error(t, err, "parseBinaryResponse should return an error")
				return
			}
			require.NoError(t, err, "parseBinaryResponse must not error")
			assert.Equal(t, tc.expect, resp, "parseBinaryResponse should return the expected response")
		})
	}
}
```

This allows for the flow to be like:
```
- receive message
- parseBinaryResponse
   - GZIP: decompress
   - otherwise: return unchanged
- MEXC WsHandleData
- proto.Unmarshal(respRaw, result)
- process the decoded ticker/orderbook data
```

So now MEXC's message still works on your PR with proto.unmarshal and now we don't need a hook

The remaining question is whether MEXC actually sends zlib-wrapped application payloads. The MEXC PR adds zlib support, but I could not find documentation or a captured frame demonstrating that behaviour. Its synthetic zlib unit test proves that the decoder supports zlib but doesn't mean it uses it - so I didn't add support in my function

gloriousCode (changes requested):

evgmalkov:
You're right, and your approach is better than mine. Rewritten in `d5cd92f3` — the hook is gone, `parseBinaryResponse` and its test are yours as posted.

I checked your claim rather than taking it on trust, since removing the fallback affects every exchange rather than just the one I care about:

- `compress/flate` appears in exactly one place in the repository — this function. No exchange package imports it.
- No package under `exchanges/` imports `compress/*` at all, and none handles binary frames itself, so nothing depends on the raw-DEFLATE branch indirectly either.

That matches what you said about OKEx, Kraken, Bittrex and Huobi, so the branch has no remaining consumer. Dropping it is a smaller change than the one I proposed and removes the problem instead of routing around it.

On zlib: you're right to be sceptical, and I should have flagged it myself. The zlib support in the MEXC PR is justified structurally — the low nibble of the first byte being 8 is the zlib CM field — not by a captured frame. I have not observed MEXC sending a zlib-wrapped payload, and the unit test is synthetic, so it demonstrates that the decoder handles zlib rather than that the venue emits it. Your function is correct to leave it out; I'll drop the zlib branch from #2312 unless a real frame turns up.

One note for whoever reviews the removal: with the fallback gone, an exchange that does send headerless DEFLATE would now silently receive compressed bytes rather than erroring. That was already the behaviour for anything that isn't a valid DEFLATE stream, and there is no such exchange in the tree today, but it is the one behavioural difference worth naming explicitly.

exchange/websocket/connection.go
```diff
@@ -318,21 +317,18 @@ func (c *connection) ReadMessage() Response {
...
```
shazbert: The receiver is unused so can drop `(c *connection)` also a suggestion would be to change the name of the function also to something like: `decompressBinaryMessage`

shazbert (changes requested):
One thing from me. Then looks good to go!

exchange/websocket/connection.go
gloriousCode: I think the name should stay. Decompressing only sometimes happens

evgmalkov:
Dropped the receiver in 107fc4ff — you're right, the function reads nothing from the connection. The test's `connection` value went with it, since it existed only to make the call.

On the rename I've left the name as `parseBinaryResponse`, siding with @gloriousCode: the function returns the frame untouched when it isn't GZIP, so decompressing is one of its outcomes rather than what it does. `decompressBinaryMessage` would promise more than the body delivers.

Happy to rename if you'd both rather have it the other way.

shazbert (approved):
ch ACK. Thanks!

exchange/websocket/connection.go
```diff
@@ -317,22 +316,19 @@ func (c *connection) ReadMessage() Response {
...
-	var reader io.ReadCloser
-	var err error
-	if len(resp) >= 2 && resp[0] == 31 && resp[1] == 139 { // Detect GZIP
-		reader, err = gzip.NewReader(bytes.NewReader(resp))
-		if err != nil {
-			return nil, err
-		}
-	} else {
-		reader = flate.NewReader(bytes.NewReader(resp))
+func parseBinaryResponse(resp []byte) ([]byte, error) {
```
thrasher-: Heimdall: **Minor** — the godoc line no longer describes what the function does now that the DEFLATE branch is gone.

**Fix:**
```suggestion
// parseBinaryResponse decompresses GZIP frames and returns all other payloads unchanged
func parseBinaryResponse(resp []byte) ([]byte, error) {
```

**Why:** "parses a websocket binary response into a usable byte array" was written when every binary frame was decompressed one way or the other. The passthrough is the new contract and it is the part a future reader needs told — the inline `// non-GZIP response, return as-is` says it at the branch, but the doc line is what shows up in godoc and in editor hovers. This is close to @gloriousCode's wording from the thread that produced this implementation, tightened only so it does not promise more than the magic-byte sniff delivers: a payload that starts `1f 8b` but is not valid GZIP errors rather than being passed through.

exchange/websocket/manager_test.go
```diff
@@ -1231,38 +1231,48 @@ func TestSetupPingHandler(t *testing.T) {
...
+			t.Parallel()
+			resp, err := parseBinaryResponse(tc.input)
+			if tc.wantErr {
+				assert.Error(t, err, "parseBinaryResponse should return an error")
+				return
+			}
+			require.NoError(t, err, "parseBinaryResponse must not error")
+			assert.Equal(t, tc.expect, resp, "parseBinaryResponse should return the expected response")
+		})
+	}
```
thrasher-, line 1280: Heimdall: **Minor** — the `empty` case exercises a nil slice rather than an empty frame, the two error cases assert only that *an* error came back, and two of the three ways the new predicate falls through to the passthrough have no case.

**Fix:** (needs `"io"` in the import block, so not a clean one-click apply)
```suggestion
	testCases := []struct {
		name   string
		input  []byte
		expect []byte
		err    error
	}{
		{name: "gzip", input: gzipBuffer.Bytes(), expect: []byte("hello")},
		{name: "uncompressed MEXC protobuf", input: mexcFrame, expect: bytes.Clone(mexcFrame)},
		{name: "unrecognised binary", input: []byte{0x01, 0x02, 0x03}, expect: []byte{0x01, 0x02, 0x03}},
		{name: "nil"},
		{name: "empty frame", input: []byte{}, expect: []byte{}},
		{name: "single byte", input: []byte{0x1f}, expect: []byte{0x1f}},
		{name: "partial gzip magic", input: []byte{0x1f, 0x00}, expect: []byte{0x1f, 0x00}},
		{name: "corrupt gzip header", input: []byte{0x1f, 0x8b, 0x08}, err: io.ErrUnexpectedEOF},
		{name: "corrupt gzip payload", input: corruptGZIPPayload, err: gzip.ErrChecksum},
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			resp, err := parseBinaryResponse(tc.input)
			if tc.err != nil {
				assert.ErrorIs(t, err, tc.err, "parseBinaryResponse should return the expected error")
				assert.Nil(t, resp, "parseBinaryResponse should not return a payload with an error")
				return
			}
			require.NoError(t, err, "parseBinaryResponse must not error")
			assert.Equal(t, tc.expect, resp, "parseBinaryResponse should return the expected response")
		})
	}
```

**Why:** four small things, all in the one table.

`{name: "empty"}` leaves `input` and `expect` nil, so it pins the nil path rather than an empty frame. A real zero-length binary frame arrives as `[]byte{}` — gorilla builds it with `io.ReadAll`, which starts from `make([]byte, 0, 512)` and so never yields nil. Checked against a server writing both `WriteMessage(BinaryMessage, []byte{})` and `WriteMessage(BinaryMessage, nil)`:

```
gorilla frame 0: type=2 err=<nil> p==nil? false len=0
gorilla frame 1: type=2 err=<nil> p==nil? false len=0
```

That distinction carries weight, because [`manager.go:1030`](https://github.com/thrasher-corp/gocryptotrader/blob/107fc4ff329c3ed7836806587f9539a8a0dafea6/exchange/websocket/manager.go#L1030) reads `resp.Raw == nil` as "connection has been closed" and returns. The passthrough cannot hand back nil for a live frame today, and the added case is what keeps that true — testify treats `[]byte(nil)` and `[]byte{}` as unequal, so the two are genuinely separate cases.

`single byte` and `partial gzip magic` cover the `len(resp) < 2` and `resp[1] != 0x8b` legs of the predicate; only `resp[0] != 0x1f` has a case today. A later refactor of the sniff — `bytes.HasPrefix`, or reordering the disjuncts — could index a one-byte frame out of range with the suite still green.

On the error cases, `assert.Error` passes for any error at all, including one a refactor introduces for an unrelated reason. The identities are stable — `io.ErrUnexpectedEOF` for the truncated header, `gzip.ErrChecksum` reachable through the `errors.Join` for the flipped trailer byte — and `ErrorIs` is what AGENTS.md asks for here. The `assert.Nil` pins the other half of that contract: an unverified payload must not escape alongside a checksum failure.

Last one: `expect: mexcFrame` aliases the input, and the function returns its argument, so that case would still pass if a future version mutated the buffer in place. `bytes.Clone` makes the oracle independent.

Verified on `107fc4ff3`: `go test ./exchange/websocket/ -run TestParseBinaryResponse -race` gives 9/9 subtests passing, `gofmt` clean, and `parseBinaryResponse` stays at 100% statement coverage.

exchange/websocket/manager_test.go
```diff
@@ -1231,38 +1231,48 @@ func TestSetupPingHandler(t *testing.T) {
...
 }
```
thrasher-, line 1281: Heimdall: **Blocker** — nothing in the suite drives a binary frame through `ReadMessage`, so the behaviour this PR changes is only covered one level down.

**Fix:** append after `TestParseBinaryResponse`. Every import it needs is already in this file, so it applies as-is.
```suggestion
}

// TestReadMessageBinary ensures binary frames reach the caller as the exchange handler expects
func TestReadMessageBinary(t *testing.T) {
	t.Parallel()

	mock := httptest.NewServer(mockws.CurryWsMockUpgrader(t, func(_ testing.TB, p []byte, c *gws.Conn) error {
		return c.WriteMessage(gws.BinaryMessage, p)
	}))
	defer mock.Close()

	wc := &connection{
		ExchangeName:     "test",
		URL:              "ws" + mock.URL[len("http"):],
		ResponseMaxLimit: time.Second * 5,
		Match:            NewMatch(),
	}
	require.NoError(t, wc.Dial(t.Context(), &gws.Dialer{}, http.Header{}, nil), "Dial must not error")
	defer func() { assert.NoError(t, wc.Connection.Close(), "Close should not error") }()

	var gzipBuffer bytes.Buffer
	g := gzip.NewWriter(&gzipBuffer)
	_, err := g.Write([]byte("hello"))
	require.NoError(t, err, "gzip.Write must not error")
	require.NoError(t, g.Close(), "gzip.Close must not error")

	mexcFrame, err := hex.DecodeString("0a3473706f74407075626c69632e61676772652e626f6f6b5469636b657" +
		"22e76332e6170692e7062403130306d73404b4153555344541a074b41535553445430d4d882c2fc33da1335" +
		"0a08302e3032363239331204313430371a08302e30323633353722053237372e352a0b3133303438393134" +
		"3133393090d482c2fc33")
	require.NoError(t, err, "captured MEXC frame must decode")

	for _, tc := range []struct {
		name   string
		send   []byte
		expect []byte
	}{
		{name: "gzip is decompressed", send: gzipBuffer.Bytes(), expect: []byte("hello")},
		{name: "uncompressed protobuf is preserved", send: mexcFrame, expect: mexcFrame},
		{name: "malformed gzip yields an empty non-nil response", send: []byte{0x1f, 0x8b, 0x08}, expect: []byte("")},
	} {
		t.Run(tc.name, func(t *testing.T) {
			require.NoError(t, wc.Connection.WriteMessage(gws.BinaryMessage, tc.send), "WriteMessage must not error")
			resp := wc.ReadMessage()
			require.NotNil(t, resp.Raw, "Raw must not be nil; a nil response stops the reader")
			assert.Equal(t, tc.expect, resp.Raw, "ReadMessage should return the expected payload")
		})
	}
}
```

**Why:** `grep -rn BinaryMessage` over the repo returns exactly one hit — [`connection.go:305`](https://github.com/thrasher-corp/gocryptotrader/blob/107fc4ff329c3ed7836806587f9539a8a0dafea6/exchange/websocket/connection.go#L305) — so no test has ever pushed a binary frame through the reader. `TestParseBinaryResponse` covers the decoder, but not the wiring the exchanges actually see: that the passthrough bytes come back on `Response.Raw`, and that a GZIP failure still produces a **non-nil** empty `Raw` rather than the nil that [`manager.go:1030`](https://github.com/thrasher-corp/gocryptotrader/blob/107fc4ff329c3ed7836806587f9539a8a0dafea6/exchange/websocket/manager.go#L1030) reads as "connection has been closed".

It is not a coverage ritual — the test has teeth. Against this branch all three subtests pass; against `upstream/master` the same file fails on exactly the case the PR exists for:

```
--- FAIL: TestReadMessageBinary
    --- PASS: TestReadMessageBinary/gzip_is_decompressed
    --- FAIL: TestReadMessageBinary/uncompressed_protobuf_is_preserved
    --- PASS: TestReadMessageBinary/malformed_gzip_yields_an_empty_non-nil_response
```

`ReadMessage` statement coverage goes 64.7% → 88.2%, and it runs clean under `-race`. It does repeat the captured frame; if you would rather not, hoist `mexcFrame` to a package-level fixture shared with `TestParseBinaryResponse`.

thrasher- (changes requested):
Heimdall: The right call — a per-connection hook would have left the shared path carrying a venue-specific exception, where deleting a heuristic nothing uses is the smaller change. And a captured frame is a far stronger fixture than a constructed one.

Whether anything still needs that branch is measurable, so I measured it. Across all 20 websocket-enabled exchanges in the tree, only Huobi sends binary, and every frame is GZIP — across `/ws`, `/feed`, `swap-ws`, `linear-swap-ws` and `api.hbdm.com/ws`, with `/ws/v2` in text, matching HTX's docs. Nothing needs raw DEFLATE. Live Huobi frames decode end to end on this branch; master drops the captured frame with `flate: corrupt input before offset 9`. **The change itself is safe.**

It also settles @gloriousCode's zlib question: MEXC spot pushes uncompressed protobuf on every `.pb` channel, and futures stays text JSON even with `"gzip": true` and `"compress": true` (`compress` there is depth merging). MEXC needs the passthrough and nothing else — no capture or doc supports zlib.

What I am holding on is the coverage. No test drives a binary frame through [`ReadMessage`](https://github.com/thrasher-corp/gocryptotrader/blob/107fc4ff329c3ed7836806587f9539a8a0dafea6/exchange/websocket/connection.go#L270), so the behaviour this PR changes is only pinned a level down. The test I suggest inline fails on master and passes here, which is the point: it holds the new contract in place. Two minors alongside it.

CI: the reds are live-API tests elsewhere — `TestFetchOrderbook` and `TestUpdateOrderExecutionLimits` fail on master's own CI too ([run 32461201993](https://github.com/thrasher-corp/gocryptotrader/actions/runs/32461201993)). Note @gloriousCode's 4 Aug change request is still open, so a re-run alone will not clear the gate.

**To unblock:**
1. Add the direct `ReadMessage` binary test ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2311#discussion_r3859320747))

evgmalkov:
Done in 1473d41f1. Added the ReadMessage binary test and applied the two minor test changes plus the godoc fix.

Before pushing I ran the new test against the old DEFLATE path to make sure it actually catches the regression: the uncompressed protobuf case fails there (the frame comes back empty) and passes on this branch. gofmt and go vet are clean, and the package tests pass under -race.

thrasher- (approved):
Thanks for making those changes @evgmalkov !

exchange/websocket/connection.go
```diff
@@ -316,23 +315,20 @@ func (c *connection) ReadMessage() Response {
...
+	reader, err := gzip.NewReader(bytes.NewReader(resp))
 	if err != nil {
 		return nil, err
 	}
-	return standardMessage, reader.Close()
+	msg, readErr := io.ReadAll(reader)
+	if err := errors.Join(readErr, reader.Close()); err != nil {
+		return nil, err
+	}
+	return msg, nil
```
thrasher-:
```suggestion
	msg, err := io.ReadAll(reader)
	if err != nil {
		return nil, err
	}
	return msg, reader.Close()
```

Old way was more simple and won't duplicate errors on a bad deflate offset

thrasher- (changes requested):
Sorry, one last change and then this LGTG

exchange/websocket/connection.go
evgmalkov: Applied in 8b2ac97cb. The test table stays green: corrupt gzip payload still comes back as a nil body with gzip.ErrChecksum, since ReadAll surfaces the checksum error before Close. go vet clean and the package tests pass under -race.

thrasher- (approved):
Thanks again @evgmalkov !

exchange/websocket/connection.go
```diff
@@ -315,20 +315,20 @@ func (c *connection) ReadMessage() Response {
...
 		return nil, err
 	}
-	msg, readErr := io.ReadAll(reader)
-	if err := errors.Join(readErr, reader.Close()); err != nil {
+	msg, err := io.ReadAll(reader)
+	if err != nil {
 		return nil, err
 	}
-	return msg, nil
+	return msg, reader.Close()
```
gloriousCode: Main reason why I suggested the errors.Join is because if io.ReadAll fails, reader.Close is never called and is a potential leak. Was a cleaner way of not calling it twice or adding a defer
image

gloriousCode (approved):
Thanks for listening and changing according to my code sample 🌈 
Since the reader.Close() leak was pre-existing, I won't block you on it
tACK

exchange/websocket/connection.go
```diff
@@ -316,23 +315,20 @@ func (c *connection) ReadMessage() Response {
...
+	if err != nil {
+		return nil, err
 	}
-	standardMessage, err := io.ReadAll(reader)
+	msg, err := io.ReadAll(reader)
 	if err != nil {
 		return nil, err
 	}
-	return standardMessage, reader.Close()
+	return msg, reader.Close()
```
thrasher-:
```suggestion
	msg, err := io.ReadAll(reader)
	if closeErr := reader.Close(); err == nil {
		err = closeErr
	}
	if err != nil {
		return nil, err
	}
	return msg, nil
```

Well this was quite funny. https://cs.opensource.google/go/go/+/refs/tags/go1.27.0:src/compress/gzip/gunzip.go;l=287

This states it doesn't actually close the underlying reader. Whereas `gzip.NewReader` mentions the caller must call `Close`. After doing a bit of digging, we've never experienced a leak since this code was introduced because `gzip.Close` just calls `func (z *Reader) Close() error { return z.decompressor.Close() }` which in turn calls  `if f.err == io.EOF { return nil }; return f.err`. As such, it does not release buffers, close the underlying bytes.Reader, stop goroutines or release an OS resource. It would be garbage collected once the function exits. However, it's still good practice and this would all change if we have to implement other compression algo which may not behave like this so it's good to get this sorted now.

This solution eliminates the `errors.Join` error duplication bug and should make GoLand static analysis warnings happy

thrasher- (changes requested):

exchange/websocket/connection.go
evgmalkov: Applied in b9c64a085. Close now runs on every path with the read error kept ahead of the close error. Thanks for digging into the gzip Close behaviour, and thanks to @gloriousCode for catching the skipped Close.

Tests stay green: corrupt gzip payload still returns a nil body with gzip.ErrChecksum (Close returns nil once the checksum error has already surfaced from ReadAll, so the read error stands), and go vet is clean.

gloriousCode (approved):
Thank you!

exchange/websocket/connection.go
thrasher-: Thanks @evgmalkov !

PR #2320, opened 2026-08-04
common/timeperiods: speed up data range matching

thrasher-:
## Summary

- replace the nested period/comparison scan in `FindTimeRangesContainingData` with one truncation per comparison and a binary search over generated periods
- remove the redundant post-match sort because generated periods remain in ascending order
- preserve range boundaries, duplicate comparison handling, location-equivalent timestamps, long ranges, and calculator reuse behaviour
- add deterministic benchmarks and direct tests for the matching paths and boundaries

## Benchmarks

Primary range-matching comparison:

`GOMAXPROCS=1 go test ./common/timeperiods -run '^$' -bench '^BenchmarkFindTimeRangesContainingData$' -benchmem -count=15 -benchtime=300ms`

Go 1.26.5, Linux/amd64, Intel Core i7-6700K:

| Case | Baseline | Initial PR head | Change |
| --- | ---: | ---: | ---: |
| empty | 50.27 us/op | 49.11 us/op | -2.31% |
| sparse | 7,772.49 us/op | 94.63 us/op | -98.78% |
| dense | 21,578.6 us/op | 169.3 us/op | -99.22% |
| out-of-range | 8,606.91 us/op | 64.50 us/op | -99.25% |

Each comparison used 15 samples with p <= 0.007. Bytes/op and allocs/op were unchanged in this initial comparison.

Redundant-sort follow-up:

The initial PR head and the same code without `Sort(false)` were compiled and run in alternating order with `GOMAXPROCS=1`, CPU affinity fixed to one core, `-test.benchtime=500ms`, and 10 samples per state.

| Case | Initial PR head | Final PR head | Timing | Allocs/op |
| --- | ---: | ---: | ---: | ---: |
| empty | 105.31 us/op | 96.00 us/op | -8.84% (p=0.035) | 16 -> 13 |
| sparse | 204.5 us/op | 196.8 us/op | ~ (p=0.739) | 25 -> 22 |
| dense | 346.8 us/op | 303.7 us/op | ~ (p=0.280) | 16 -> 13 |
| out-of-range | 141.8 us/op | 131.0 us/op | ~ (p=0.579) | 16 -> 13 |

The timing geomean improved by 8.23%. Every workload saved exactly 3 allocs/op and 104 B/op; the allocation and byte reductions were stable across all samples (p < 0.001).

## Test plan

- `go test ./common/timeperiods -count=20`
- `go test -race ./common/timeperiods -count=1`
- `go test ./common/timeperiods -count=1 -coverprofile=coverage.out`
- `go vet ./common/timeperiods`
- `golangci-lint run ./...`
- `codespell`
- `make misc_checks`

## Limitations

These results are from a deterministic package microbenchmark using a 24-hour range divided into one-minute periods. They do not measure end-to-end RPC, database, or exchange latency. WSL2 timing variance was high in the redundant-sort comparison, so only the empty timing case reached individual significance; the allocation and byte reductions were stable across every sample.

### DIFF
--- a/common/timeperiods/timeperiods.go
+++ b/common/timeperiods/timeperiods.go
@@ -33,7 +33,6 @@ func FindTimeRangesContainingData(start, end time.Time, period time.Duration, co
 	t.comparisonTimes = comparisonTimes
 
 	t.setTimePeriodExists()
-	t.Sort(false)
 	t.calculateRanges()
 
 	return t.TimeRanges, nil
--- a/common/timeperiods/timeperiods.go
+++ b/common/timeperiods/timeperiods.go
@@ -130,6 +130,8 @@ func (t *TimePeriodCalculator) setTimePeriodExists() {
 		return
 	}
 	newPeriods := t.TimePeriods[periodOffset:]
+	// Check every period retained from earlier calls because the same
+	// timestamp may appear in more than one previously appended range.
 	for i := range t.comparisonTimes {
 		comparisonTime := t.comparisonTimes[i].Truncate(t.periodDuration)
 		for j := range periodOffset {
@@ -137,7 +139,9 @@ func (t *TimePeriodCalculator) setTimePeriodExists() {
 				t.TimePeriods[j].dataInRange = true
 			}
 		}
-		// calculatePeriods appends in ascending order; binary search also avoids duration overflow across long ranges.
+		// The range appended by calculatePeriods contains one entry per
+		// interval in ascending time order. Find the first period at or after
+		// comparisonTime, then mark it only when the timestamps match exactly.
 		periodIndex := sort.Search(len(newPeriods), func(j int) bool {
 			return !newPeriods[j].Time.Before(comparisonTime)
 		})
--- a/common/timeperiods/timeperiods.go
+++ b/common/timeperiods/timeperiods.go
@@ -125,13 +125,25 @@ func (t *TimePeriodCalculator) calculatePeriods() {
 // against calculated TimePeriods to determine whether
 // there is existing data within the time period
 func (t *TimePeriodCalculator) setTimePeriodExists() {
+	periodOffset := len(t.TimePeriods)
 	t.calculatePeriods()
-	for i := range t.TimePeriods {
-		for j := range t.comparisonTimes {
-			if t.comparisonTimes[j].Truncate(t.periodDuration).Equal(t.TimePeriods[i].Time) {
-				t.TimePeriods[i].dataInRange = true
-				break
+	if len(t.TimePeriods) == 0 {
+		return
+	}
+	newPeriods := t.TimePeriods[periodOffset:]
+	for i := range t.comparisonTimes {
+		comparisonTime := t.comparisonTimes[i].Truncate(t.periodDuration)
+		for j := range periodOffset {
+			if t.TimePeriods[j].Time.Equal(comparisonTime) {
+				t.TimePeriods[j].dataInRange = true
 			}
+		}
+		// calculatePeriods appends in ascending order; binary search also avoids duration overflow across long ranges.
+		periodIndex := sort.Search(len(newPeriods), func(j int) bool {
+			return !newPeriods[j].Time.Before(comparisonTime)
+		})
+		if periodIndex < len(newPeriods) && newPeriods[periodIndex].Time.Equal(comparisonTime) {
+			newPeriods[periodIndex].dataInRange = true
 		}
 	}
 }
--- a/common/timeperiods/timeperiods_bench_test.go
+++ b/common/timeperiods/timeperiods_bench_test.go
@@ -1,0 +1,38 @@
+package timeperiods
+
+import (
+	"testing"
+	"time"
+)
+
+func BenchmarkFindTimeRangesContainingData(b *testing.B) {
+	start := time.Date(2026, time.January, 1, 0, 0, 0, 0, time.UTC)
+	end := start.Add(24 * time.Hour)
+	sparseComparisonTimes := make([]time.Time, 0, 24*12)
+	for offset := time.Duration(0); offset < end.Sub(start); offset += 5 * time.Minute {
+		sparseComparisonTimes = append(sparseComparisonTimes, start.Add(offset+30*time.Second))
+	}
+	denseComparisonTimes := make([]time.Time, 0, 24*60)
+	for offset := time.Duration(0); offset < end.Sub(start); offset += time.Minute {
+		denseComparisonTimes = append(denseComparisonTimes, start.Add(offset+30*time.Second))
+	}
+	outOfRangeComparisonTimes := make([]time.Time, len(sparseComparisonTimes))
+	for i := range sparseComparisonTimes {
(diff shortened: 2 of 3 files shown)


thrasher-:
@codex please review

common/timeperiods/timeperiods.go
```diff
@@ -125,14 +124,26 @@ func (t *TimePeriodCalculator) calculatePeriods() {
...
-		for j := range t.comparisonTimes {
-			if t.comparisonTimes[j].Truncate(t.periodDuration).Equal(t.TimePeriods[i].Time) {
-				t.TimePeriods[i].dataInRange = true
-				break
+	if len(t.TimePeriods) == 0 {
+		return
+	}
+	newPeriods := t.TimePeriods[periodOffset:]
+	for i := range t.comparisonTimes {
+		comparisonTime := t.comparisonTimes[i].Truncate(t.periodDuration)
```
shazbert, line 136: This doesn't smell right; we shouldn't be truncating a slice of comparison times by the period duration they should already be aligned at this point right? If alignment is off by period duration then it should be an error.

common/timeperiods/timeperiods.go
```diff
@@ -125,14 +124,26 @@ func (t *TimePeriodCalculator) calculatePeriods() {
...
+	periodOffset := len(t.TimePeriods)
 	t.calculatePeriods()
-	for i := range t.TimePeriods {
-		for j := range t.comparisonTimes {
-			if t.comparisonTimes[j].Truncate(t.periodDuration).Equal(t.TimePeriods[i].Time) {
-				t.TimePeriods[i].dataInRange = true
-				break
+	if len(t.TimePeriods) == 0 {
+		return
+	}
```
shazbert, line 131: This looks like an error as it should contain time periods?

common/timeperiods/timeperiods.go
```diff
@@ -125,14 +124,26 @@ func (t *TimePeriodCalculator) calculatePeriods() {
...
+	newPeriods := t.TimePeriods[periodOffset:]
+	for i := range t.comparisonTimes {
+		comparisonTime := t.comparisonTimes[i].Truncate(t.periodDuration)
+		for j := range periodOffset {
+			if t.TimePeriods[j].Time.Equal(comparisonTime) {
+				t.TimePeriods[j].dataInRange = true
```
shazbert, line 139: shouldn't there be a `break` in this if statement?

shazbert (changes requested):
I'm confused and I need more coffee; but I think a bunch of things could be refactored here to return errors, `TimePeriodCalculator` type could be removed and just have pure function with returns.

common/timeperiods/timeperiods.go
thrasher-: Iris: The truncation is intentional. `comparisonTimes` are observations that may fall anywhere within a period, not pre-aligned period boundaries; the trade-based caller passes raw trade timestamps. The previous implementation also truncated each comparison before matching—this change only hoists that work out of the per-period scan—and the tests cover sub-period timestamps being assigned to their correct bucket. Treating those timestamps as errors would reject valid data and change existing behaviour.

common/timeperiods/timeperiods.go
thrasher-: Iris: This guard does not cause the empty result. `calculatePeriods` can append nothing for zero-length or reversed ranges, or when truncation collapses the interval, and the previous implementation returned an empty result for those inputs as well. The guard therefore preserves existing behaviour; returning an error instead would be a separate public-contract change. The explicit guard is redundant because `sort.Search(0, ...)` is safe, but that is independent of whether an empty result is valid.

common/timeperiods/timeperiods.go
thrasher-: Iris: No `break` here: when the calculator is reused, `calculatePeriods` appends another set, so the existing prefix can contain multiple periods with the same timestamp. The previous nested loop marked every matching `TimePeriod`; breaking would mark only the first duplicate and change that behaviour. The newly appended suffix has unique ascending timestamps and is handled separately by the binary search.

thrasher-:
Iris: Agreed that this would be simpler as pure functions. The public entry points already return validation errors, but zero-length, reversed, and truncate-to-zero ranges currently produce empty output, so turning those cases into errors is a behaviour change. `TimePeriodCalculator` and `Sort` are also exported, so removing them is a public API change even though this repository has no call sites outside the package. I’d keep this performance change behaviour-preserving and handle that API and validation refactor separately.

shazbert (approved):
Nothing more from me 👍

common/timeperiods/timeperiods.go
```diff
@@ -125,14 +124,26 @@ func (t *TimePeriodCalculator) calculatePeriods() {
...
-	for i := range t.TimePeriods {
-		for j := range t.comparisonTimes {
-			if t.comparisonTimes[j].Truncate(t.periodDuration).Equal(t.TimePeriods[i].Time) {
-				t.TimePeriods[i].dataInRange = true
-				break
+	if len(t.TimePeriods) == 0 {
+		return
+	}
+	newPeriods := t.TimePeriods[periodOffset:]
+	for i := range t.comparisonTimes {
```
gloriousCode, line 135:
```suggestion
	// Check every period retained from earlier calls because the same
	// timestamp may appear in more than one previously appended range.
	for i := range t.comparisonTimes {
```

common/timeperiods/timeperiods.go
```diff
@@ -125,14 +124,26 @@ func (t *TimePeriodCalculator) calculatePeriods() {
...
+		comparisonTime := t.comparisonTimes[i].Truncate(t.periodDuration)
+		for j := range periodOffset {
+			if t.TimePeriods[j].Time.Equal(comparisonTime) {
+				t.TimePeriods[j].dataInRange = true
 			}
 		}
+		// calculatePeriods appends in ascending order; binary search also avoids duration overflow across long ranges.
```
gloriousCode:
```suggestion
		// The range appended by calculatePeriods contains one entry per
		// interval in ascending time order. Find the first period at or after
		// comparisonTime, then mark it only when the timestamps match exactly.
```

gloriousCode (approved):
tACK! I've added commentary suggestions because I needed more than what was presented to understand and I'll likely forget by the time I need to look again. But it is optional

common/timeperiods/timeperiods.go
thrasher-: Iris: Applied in [af2d0ef49cc63e9c056f1b10f3ee09843b7795f5](https://github.com/thrasher-/gocryptotrader/commit/af2d0ef49cc63e9c056f1b10f3ee09843b7795f5). The added comment explains why every retained period is checked: reuse can append the same timestamp more than once, so stopping after the first match could leave later duplicates unmarked.

common/timeperiods/timeperiods.go
thrasher-: Iris: Applied in [af2d0ef49cc63e9c056f1b10f3ee09843b7795f5](https://github.com/thrasher-/gocryptotrader/commit/af2d0ef49cc63e9c056f1b10f3ee09843b7795f5). The added comment explains that newly appended periods are in ascending order, so the binary search finds the first period at or after the comparison time and marks it only on an exact timestamp match.

PR #2323, opened 2026-08-06
KuCoin: harden private state, tickers, and order submission

gloriousCode:
## Summary

- update ticker caches from all available KuCoin markets, including disabled but available pairs, while retaining per-ticker error aggregation
- classify current spot and margin account families consistently across REST and WebSocket updates, aggregate split balances, and preserve unclassified funding events as raw data
- submit margin orders through the supported high-frequency endpoint with corrected JSON encoding and independent auto-borrow and auto-repay handling
- query the futures account position mode and populate the required marginMode and positionSide fields for one-way and hedge-mode orders
- validate nil and invalid order inputs with stable wrapped errors

## Testing

- go test ./exchanges/kucoin -race -count 1
- make gofumpt
- make lint
- make misc_checks

### DIFF
--- a/exchanges/kucoin/kucoin.go
+++ b/exchanges/kucoin/kucoin.go
@@ -12,6 +12,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/thrasher-corp/gocryptotrader/common"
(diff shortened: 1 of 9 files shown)


thrasher-:
@codex please review

PR #2295, opened 2026-07-23
BitMEX: Remove exchange implementation

thrasher-:
## Summary

- Remove the BitMEX exchange implementation and active support references.
- Add config version ~13~ 14 to remove legacy BitMEX configuration and credentials during upgrade.
- Preserve historical migration fixtures so older configs continue to upgrade safely.

## Context

BitMEX announced that exchange operations will cease on 23 September 2026 at 04:00 UTC:
https://x.com/BitMEX/status/2080201602456301580

## Testing

- `go test ./config/versions/... ./config ./engine ./exchanges ./exchanges/orderbook ./cmd/documentation ./cmd/exchange_wrapper_issues -race -count=1`
- `make lint`
- `make misc_checks`
- `codespell`
- `go test ./... -race -count=1` encountered transient failures in `TestGetOpenInterest`, `TestGetSupportedFlashSwapCurrencies`, and `TestProcessOrderbook`; each passed when rerun in isolation.

PR #2325, opened 2026-08-06
BitMEX: Remove exchange implementation

thrasher-:
## Summary

- remove the BitMEX exchange implementation and active support references
- add config version 14 to remove legacy BitMEX configuration and credentials while preserving the CoinMarketCap v13 migration already on `master`
- remove the BitMEX-only insertion path from the shared orderbook implementation while retaining `InsertAction` as a deprecated compatibility constant that is rejected at runtime
- preserve historical migration fixtures so older configs continue to upgrade safely

## Context

BitMEX announced that exchange operations will cease on 23 September 2026 at 04:00 UTC:
https://x.com/BitMEX/status/2080201602456301580

This replaces #2295 with the change rebased onto current `master`; the original published branch cannot be updated to the rebased history without rewriting it.

## Testing

- `go test ./config/versions/... ./config ./engine ./exchanges ./exchanges/orderbook ./cmd/documentation ./cmd/exchange_wrapper_issues -race -count=1`
- `go test ./... -race -count=1`
- `make lint`
- `make misc_checks`
- `codespell`
- `go mod tidy -diff`


thrasher-:
@codex please review

thrasher-:
Testing functionality, will merge in the existing one

PR #2137, opened 2025-12-19
Fix NTP: implement RFC 5905 offset calculation and verify time before warning

romanornr:
# PR implement RFC 5905 offset calculation and verify time before warning

Before I got 
```
[DEBUG] | DATABASE | 19/12/2025 22:06:26 | Database manager starting...
[DEBUG] | CONNECTION | 19/12/2025 22:06:26 | Connection manager starting...
[DEBUG] | LOG | 19/12/2025 22:06:26 | Internet connectivity found
[DEBUG] | CONNECTION | 19/12/2025 22:06:26 | Connection manager started.
Your system time is out of sync, this may cause issues with trading
How would you like to show future notifications? (a)lert at startup / (w)arn periodically / (d)isable
```

No, the issue is not my system/setup

```
❯ timedatectl timesync-status

       Server: 162.159.200.1 (time.cloudflare.com)
Poll interval: 34min 8s (min: 32s; max 34min 8s)
         Leap: normal
      Version: 4
      Stratum: 3
    Reference: AE20804
    Precision: 1us (-26)
Root distance: 2.852ms (max: 5s)
       Offset: -383us
        Delay: 10.140ms
       Jitter: 1.091ms
 Packet count: 15
    Frequency: -18.014ppm
```

```
My PC (perfectly synced)           NTP Server
        |                                       |
   T1 = 10:00:00.000 ─── You send request ───→  |
        |                                       | T2 = 10:00:00.050 (server receives)
        |                                       | T3 = 10:00:00.050 (server sends response)
        |  ←── Response packet arrives ─────────|
   T4 = 10:00:00.100 (you receive it)           |
```

The 4 timestamps
```
| Name | When | Which Clock | Where is it? |
|------|------|-------------|--------------|
| **T1** | Client sends request | Client | **Not recorded** in current code! |
| **T2** | Server receives request | Server | `RxTimeSec` / `RxTimeFrac` |
| **T3** | Server sends response | Server | `TxTimeSec` / `TxTimeFrac` |
| **T4** | Client receives response | Client | **Not recorded** in current code! |
```

Original code
```go
NTPTime := T3              // 10:00:00.050 (server's transmit time)
currentTime := time.Now()  // 10:00:00.100 (this is T4)
diff := NTPTime.Sub(currentTime)  // T3 - T4 = -50ms
```

The issue
`T3 - T4` gives us `-(network_delay) + offset`
If my clock is perfect (offset = 0), I still get `-50ms` due to network

[...]

PR #2327, opened 2026-08-06
engine: fix NTP check counting network delay as clock error

romanornr:
## Summary

GoCryptoTrader told me my system clock was wrong. It was correct.

The cause is a measurement error in the NTP client on `master`. The code compares two timestamps
that are not comparable, so it reads network delay as clock error. On a normal internet connection,
the network delay alone can use the complete default tolerance.

This pull request replaces the hand-written SNTP code with a validated measurement. It uses the
standard offset formula from RFC 5905, it reports how accurate each measurement is, and it warns
only when the measurement supports a conclusion.

You do not need to read any other pull request to review this one. Everything is explained below.

---

## Background: what the NTP client does in GoCryptoTrader

Exchange APIs reject signed requests when the timestamp in the request is outside a receive window.
The window is often a few seconds. If the local clock drifts, orders fail. The failure message from
the exchange is usually not clear about the cause.

GoCryptoTrader therefore checks the local clock against public NTP servers. The subsystem is called
`ntp_timekeeper`. It is advisory only: **it never changes the system clock.** It writes a log
message, and at one setting it asks the user how to be told in future.

The configuration is small:

```json
"ntpclient": {
  "enabled": 0,
  "pool": ["pool.ntp.org:123"],
  "allowedDifference": 50000000,
  "allowedNegativeDifference": 50000000
}
```

| Field | Meaning |
|---|---|
| `enabled` | `-1` off, `0` check at startup and prompt, `1` warn periodically |
| `pool` | NTP servers, tried in order |
| `allowedDifference` | How far the clock may be **behind**, in nanoseconds (50 ms) |
| `allowedNegativeDifference` | How far the clock may be **ahead**, in nanoseconds (50 ms) |

---

## The problem I hit

At startup, GoCryptoTrader showed me this:

```
Your system time is out of sync, this may cause issues with trading
How would you like to show future notifications? (a)lert at startup / (w)arn periodically / (d)isable
```

My clock was fine. `systemd-timesyncd` reported an error of **−383 microseconds**:

```
       Server: 162.159.200.1 (time.cloudflare.com)
Root distance: 2.852ms (max: 5s)
       Offset: -383us
        Delay: 10.140ms
```

So I looked at the code on `master`.

### Why the code on `master` is wrong

To compare your clock with a server clock, you need the time at the server **now**. The reply from
the server is always late, because the packet needs time to travel.

An NTP exchange produces four timestamps:

| Name | Event | Clock that records it |
|---|---|---|
| **T1** | The client sends the request | Client |
| **T2** | The server receives the request | Server |
| **T3** | The server sends the reply | Server |
| **T4** | The client receives the reply | Client |

`master` uses two of the four. It reads the server transmit time (T3) and compares it with the local
time at reception (T4):

```go
// engine/ntp_manager.go on master
NTPTime, err := m.FetchNTPTime()   // T3: time at the server when it sent the reply
currentTime := time.Now()          // T4: time at the client when the reply arrived
diff := NTPTime.Sub(currentTime)   // T3 − T4
```

`T3 − T4` is not the clock error. It is:

```
T3 − T4  =  clock offset  −  one-way delay (server → client)
```

The network delay is always inside the result. A correct clock therefore looks slow, by the size of
the delay.

### The numbers show why this fires so easily

The default tolerance is 50 ms in each direction. I measured `pool.ntp.org` from a home connection:
the round-trip time is 35 ms to 55 ms, so the one-way delay is roughly 18 ms to 28 ms. Slower paths
give more.

| True clock error | One-way delay | Result on `master` | Warning shown? |
|---:|---:|---:|---|
| 0 ms | 20 ms | −20 ms | No |
| 0 ms | 50 ms | −50 ms | **Yes, but the clock is perfect** |
| 0 ms | 60 ms | −60 ms | **Yes, but the clock is perfect** |

The distance to the server alone can use the whole tolerance.

### Two more problems in the same code

**It asks before it measures.** At `enabled: 0`, `master` shows the prompt every time. The prompt
does not depend on the clock:

```go
// engine/engine.go on master
if bot.Config.NTPClient.Level == 0 {
    responseMessage, err := bot.Config.SetNTPCheck(os.Stdin)   // no measurement first
```

**A total failure looks like success.** When no server answers, `master` returns the local time:

```go
log.Warnln(log.TimeMgr, "No valid NTP servers found, using current system time")
return time.Now().UTC()
```

The caller then compares local time with local time. The difference is always zero. So a complete
network failure and a perfect clock produce the same result.

That last point explains why the bug survived so long. The check reported success no matter what
happened.

---

## The approach

The three problems above share one cause: the code measures the clock, judges the clock, talks to
the network, and prompts the user, all in the same functions. Nothing can be tested without a
network, so nothing was.

This change splits that into four stages with one direction of flow:

```
  configured pools (in order)
            │
            ▼
  queryClock ──────► github.com/beevik/ntp ──────► reply validation
            │
            ▼
  clockObservation { correction, uncertainty, round-trip, source }
            │
            ▼
  evaluateClock   ← pure function: no network, no logging, no config
            │
            ▼
  In sync | Ahead | Behind | Inconclusive
            │
            ├────────────► startup policy   (measure, then maybe prompt)
            └────────────► periodic loop    (measure, then log)
```

Four decisions define the design:

**1. The protocol is not ours to implement.** `queryClock` is the only function that speaks NTP, and
it delegates to a maintained library. Library types stop at the adapter; nothing else in the engine
knows the dependency exists.

**2. The measurement is a value, not a time.** The adapter returns one local type:

```go
type clockObservation struct {
    source          string
    clockCorrection time.Duration   // server − local; positive means the local clock is behind
    uncertainty     time.Duration   // root distance
    roundTrip       time.Duration
}
```

`master` returned a `time.Time`, which throws away everything except the answer. Carrying the
uncertainty is what makes the fourth result possible.

**3. The decision is a pure function.** `evaluateClock` takes an observation and two limits and
returns one of four states. It performs no I/O, writes no logs, and reads no configuration. Every
boundary case is therefore testable in microseconds, with no network and no clock.

**4. Startup and the periodic loop share one check.** Both call the same
measure-then-evaluate helper, so the sign convention and the threshold comparison exist in exactly
one place. On `master` the comparison was written twice, in two different forms.

### How each stage now handles failure

| Stage | On failure |
|---|---|
| One server | Try the next configured server |
| All servers | Return an error. Never substitute local time. |
| Startup measurement | Write one warning, continue starting, do not prompt |
| Engine shutdown during a check | Stop, and report the cancellation, not a clock verdict |

---

### 1. It uses all four timestamps

RFC 5905 defines the clock offset as:

```
offset = ((T2 − T1) + (T3 − T4)) / 2
```

The two delays cancel each other when the path is symmetric. What remains is the clock difference.

### 2. It reports how accurate each measurement is

You cannot know the offset exactly. You can only know it within a limit. The limit has two parts:

- **Path asymmetry.** If the two directions have different delays, the error can be as large as half
  the round-trip time.
- **Server error.** The server has its own error against its reference clock.

NTP calls the total the **root distance**. This change keeps that value and uses it as the radius of
an interval:

```
low  = offset − root distance
high = offset + root distance
```

### 3. It decides from the interval, not from one number

The code compares the complete interval with the tolerance window:

| Result | Condition | Action |
|---|---|---|
| **In sync** | The interval is completely inside the window | Log at debug level |
| **Behind** | The interval is completely above the window | Warn, and prompt at `enabled: 0` |
| **Ahead** | The interval is completely below the window | Warn, and prompt at `enabled: 0` |
| **Inconclusive** | The interval crosses a limit of the window | Warn, but **never** prompt |

`Inconclusive` is the important addition. The code on `master` has no way to say *"I cannot tell"*.
It always produces an answer, even when the measurement does not support one. An `Inconclusive`
result also tells the user what to do: configure a nearer NTP server, or one with a lower delay.

### 4. It measures before it prompts

The prompt appears only after a conclusive `Ahead` or `Behind` result. A failed measurement writes
one warning, and startup continues.

---

## `master` compared with this pull request

| Item | `master` today | This pull request |
|---|---|---|
| Timestamps used | 2 of 4 (T3, T4) | 4 of 4 |
| Network delay | Counted as clock error | Removed by the formula |
| Accuracy of result | Not known | Root distance, reported with every result |
| Possible results | Warn, or nothing | In sync, Ahead, Behind, Inconclusive |
| Prompt at `enabled: 0` | Always, before measuring | Only after a conclusive result |
| If every server fails | Returns local time as if it worked | Returns an error, warns once, continues |
| Packet handling | Hand-written struct and bit arithmetic | `github.com/beevik/ntp` v1.5.0 |
| Reply validation | None | Mode, stratum, leap indicator, originate echo, kiss-o'-death |
| Wire version | NTPv3 (`0x1B`) | NTPv4 (library default) |
| Server selection | First that answers | First that answers **and** passes validation |

---

## Why a library

The code on `master` contains a partial SNTP client: a packet struct, epoch conversion, and
fixed-point arithmetic. It performs no validation of the reply.

`github.com/beevik/ntp` v1.5.0 is the standard Go client. It supplies the offset formula, the
round-trip time, the root distance, and the reply checks. The version is pinned exactly.

Library types do not leave the adapter. Only two functions name them. The rest of the engine uses one
local type:

```go
type clockObservation struct {
    source          string
    clockCorrection time.Duration   // server − local; positive means the local clock is behind
    uncertainty     time.Duration   // root distance
    roundTrip       time.Duration
}
```

The adapter also rejects a reply when its own values disagree:

| Check | Reason |
|---|---|
| `RTT < 0` | Impossible; the reply is not trustworthy |
| `RootDistance < 0` | Impossible |
| `RootDistance < RTT/2` | The stated accuracy is better than physics allows |

A rejected reply causes a move to the next configured server.

---

## What is deleted

- `ntpPacket`, and the hand-written epoch and fixed-point conversion
- `checkTimeInPools`, including the fallback to local time
- `FetchNTPTime`, which has no caller in the code base
- The retry loop in `ntpManager.Start` and its `retryLimit` field
- The `loggingEnabled` field and parameter
- Per-server log messages that repeat the error the caller already reports

Production code is **`+304/−158`** against `master`: a net increase of 146 lines that removes a
protocol implementation and adds validation, uncertainty, and four distinct results.

---

## Compatibility

Nothing in the public surface changes:

| Surface | Status |
|---|---|
| `ntp_timekeeper` subsystem name | Unchanged |
| `SetSubsystem`, gRPC `EnableSubsystem`, `gctcli` | Unchanged |
| Config JSON shape and field names | Unchanged |
| `enabled` values `-1`, `0`, `1` | Unchanged |
| `config.SetNTPCheck` and its prompt text | Unchanged |
| Default pool `pool.ntp.org:123` | Unchanged |
| Default tolerance ±50 ms | Unchanged |
| Error values such as `errNTPManagerDisabled` | Unchanged |

The `enabled` numbers now have names in code (`NTPClientDisabled`, `NTPClientStartup`,
`NTPClientPeriodic`). The numbers themselves do not change, because they are written to the
configuration file.

---

## Existing behaviour you should know about

These items are **not** new, and this pull request does not change them. They are written down here
because they are easy to miss.

1. **Periodic mode does not start by itself.** `Engine.Start` builds the manager but never starts it.
   Only `gctcli enablesubsystem ntp_timekeeper` starts it, so `enabled: 1` is inactive in a normal
   run. This also explains how the wrong formula stayed in the code: almost nothing ran it. This
   needs a separate issue.

2. **If you answer `w` at the prompt, you get no periodic warnings.** The answer updates the
   configuration and the manager, but it does not start the manager, for the reason above.

3. **`Stop` does not wait for the periodic goroutine.** A fast `Stop` then `Start` can race. This is
   an existing defect and needs a separate issue.

4. **The default pool name stays `pool.ntp.org`.** The NTP Pool project asks software vendors to use
   a vendor zone. Requesting one is a project decision, not part of this change.

5. **The check interval changes from 30 seconds to 30 minutes.** 30 seconds is 120 requests per hour
   to the public pool. `systemd-timesyncd` uses about 34 minutes for comparison. Because of item 1,
   this affects only a manager that a user starts. This is one line; remove it if you prefer.

6. **One server at a time, in order.** The code asks the configured servers in order and uses the
   first valid reply. It does not query servers at the same time and does not compare their answers.
   It makes no claim of consensus.

---

## Tests

All tests are hermetic. No test contacts a public NTP server.

| Area | What it protects |
|---|---|
| Classifier table | The interval logic, using **different** ahead and behind limits, so a swap of signs or limits fails |
| Boundary rows | Exact limits, and one nanosecond outside each limit |
| Saturation | A very large offset cannot overflow into the opposite result |
| Startup policy | In sync and Inconclusive do not prompt; Ahead and Behind do; a failed measurement does not prompt |
| Context handling | A cancelled parent stops startup; an expired measurement budget does not |
| Ordered fallback | The first valid server wins; later servers are not asked |
| Adapter rejects | Each invalid-metadata rejection returns its own error value |
| **Production path** | `setupNTPManager` builds the real measurement chain, proved through a local UDP server |
| Subsystem | `ntp_timekeeper` enable and disable still work |

The production-path test is worth a note. Tests that replace a component prove the parts, not the
assembly. An earlier version of that test only checked that a function was present. A stub that
always reported "in sync" would have passed the complete suite, and the original failure would have
returned in silence. The test now runs a real exchange against a local UDP server and checks the
sign of the result.

---

## Timing

Each attempt against one server is bounded twice: 5 seconds to connect, then 5 seconds to read.
These are the same limits as `master`.

| Path | Bound | What the bound covers |
|---|---|---|
| Startup | 10 second budget | Stops further servers and any work before the next query |
| Periodic | servers × 10 seconds | The same per-server limits, applied in turn |

**The startup budget is not a hard stop on an exchange already in progress.** The library sets its
own read deadline after the connection opens, so one reply that is already being awaited can finish
up to 5 seconds after the budget ends. The budget stops the *next* server from being tried; it does
not interrupt the current read.

This is a known limit of the library, which has no context-aware query method. The alternative is a
wrapper around the connection. For a feature that only writes a log line, that cost is not
justified, so the behaviour is documented instead of worked around.

---

## Verification

```console
go test ./config -race -shuffle=on -count=20 -run '^(TestCheckNTPConfig|TestSetNTPCheck|TestCheckConfig.*NTP)'
go test ./engine -race -shuffle=on -count=50 -run '^(Test.*NTP|Test.*Clock)'
go test ./... -race -count 1
make check
go mod tidy -diff
```

All pass.

---

## Review guidance

The change is 17 files. Most lines are tests and generated documentation.

**Read the production code first. It is `+304/−158` of the `+1376/−372` total:**

| File | Content |
|---|---|
| `engine/ntp_clock.go` | New. Adapter, ordered fallback, and the pure classifier |
| `engine/ntp_manager.go` | Setup, lifecycle, the shared check, startup policy, reporting |
| `engine/ntp_manager_types.go` | Fields removed, measurement seam added |
| `engine/engine.go` | One call replaces the old block |
| `config/config.go` | Normalisation is now unconditional; both tolerances treat `<= 0` the same |

One configuration defect is fixed at the same time. `CheckNTPConfig` normalised `allowedDifference`
only when it was `0`, but normalised `allowedNegativeDifference` when it was `<= 0`. A negative
`allowedDifference` therefore survived and inverted the valid window. `CheckConfig` also skipped
normalisation completely at `enabled: 0`, which is the setting that needs it most.

---

## Housekeeping

This supersedes #2137, an earlier attempt of mine that applied the correct formula but kept the
hand-written protocol code and extended it. Nothing from that pull request is needed to review this
one. Please close it in favour of this change.

### DIFF
--- a/cmd/documentation/config_templates/config_readme.tmpl
+++ b/cmd/documentation/config_templates/config_readme.tmpl
@@ -227,21 +227,43 @@ See the section `exchange.features.enabled.subscriptions` for configuring subscr
 ```
 
 
-## Configure Network Time Server 
+## Configure Network Time Server
 
-+ To configure and enable a NTP server you need to set the "enabled" field to one of 3 values -1 is disabled 0 is enabled and alert at start up 1 is enabled and warn at start up
-servers are configured by the pool array and attempted first to last allowedDifference and allowedNegativeDifference are how far ahead and behind is acceptable for the time to be out in nanoseconds
+The `enabled` field controls when NTP clock checks run:
+
+These settings take effect only when the engine's NTP client feature is enabled; the command-line
+application enables that feature by default.
+
+- `-1` disables checks.
+- `0` checks during engine startup. The alert/warn/disable prompt is shown only when the complete
+  measured uncertainty interval proves that the local clock is ahead or behind the configured
+  tolerance window.
+- `1` checks every 30 minutes after the `ntp_timekeeper` subsystem is explicitly started. Normal
+  engine startup constructs this subsystem but does not start it.
+
+Pools are attempted in configured order, and the first valid response is used. The result is one of
+four states: in sync, ahead, behind, or inconclusive. An inconclusive result warns without prompting;
+configure a closer or lower-latency NTP server to reduce measurement uncertainty.
+The interval is a diagnostic bound based on the server's reported NTP metadata, not authenticated
+proof of UTC.
+
+`allowedDifference` is the permitted local-clock-behind magnitude and
+`allowedNegativeDifference` is the permitted local-clock-ahead magnitude, both expressed as Go
+`time.Duration` nanoseconds. Missing, zero, or negative values default to 50 milliseconds. An empty
+pool list defaults to `pool.ntp.org:123`.
 
 ```js
- "ntpclient": {
-  "enabled": 0,
-  "pool": [
-   "pool.ntp.org:123"
-  ],
-  "allowedDifference": 0,
-  "allowedNegativeDifference": 0
- },
- ```
+{
+  "ntpclient": {
+    "enabled": 0,
+    "pool": [
+      "pool.ntp.org:123"
+    ],
+    "allowedDifference": 0,
+    "allowedNegativeDifference": 0
+  }
+}
+```
 
 {{template "donations" .}}
 {{end}}
--- a/config/README.md
+++ b/config/README.md
@@ -244,21 +244,43 @@ See the section `exchange.features.enabled.subscriptions` for configuring subscr
 ```
 
 
-## Configure Network Time Server 
+## Configure Network Time Server
 
-+ To configure and enable a NTP server you need to set the "enabled" field to one of 3 values -1 is disabled 0 is enabled and alert at start up 1 is enabled and warn at start up
-servers are configured by the pool array and attempted first to last allowedDifference and allowedNegativeDifference are how far ahead and behind is acceptable for the time to be out in nanoseconds
+The `enabled` field controls when NTP clock checks run:
+
+These settings take effect only when the engine's NTP client feature is enabled; the command-line
+application enables that feature by default.
+
+- `-1` disables checks.
+- `0` checks during engine startup. The alert/warn/disable prompt is shown only when the complete
+  measured uncertainty interval proves that the local clock is ahead or behind the configured
+  tolerance window.
+- `1` checks every 30 minutes after the `ntp_timekeeper` subsystem is explicitly started. Normal
+  engine startup constructs this subsystem but does not start it.
+
+Pools are attempted in configured order, and the first valid response is used. The result is one of
+four states: in sync, ahead, behind, or inconclusive. An inconclusive result warns without prompting;
+configure a closer or lower-latency NTP server to reduce measurement uncertainty.
+The interval is a diagnostic bound based on the server's reported NTP metadata, not authenticated
+proof of UTC.
+
+`allowedDifference` is the permitted local-clock-behind magnitude and
+`allowedNegativeDifference` is the permitted local-clock-ahead magnitude, both expressed as Go
+`time.Duration` nanoseconds. Missing, zero, or negative values default to 50 milliseconds. An empty
+pool list defaults to `pool.ntp.org:123`.
 
 ```js
- "ntpclient": {
-  "enabled": 0,
-  "pool": [
-   "pool.ntp.org:123"
-  ],
-  "allowedDifference": 0,
-  "allowedNegativeDifference": 0
- },
- ```
+{
+  "ntpclient": {
+    "enabled": 0,
+    "pool": [
+      "pool.ntp.org:123"
+    ],
+    "allowedDifference": 0,
+    "allowedNegativeDifference": 0
+  }
+}
+```
 
 ## Donations
 
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -12,6 +12,7 @@ import (
 	"slices"
 	"strings"
 	"testing"
+	"testing/iotest"
 	"time"
 
 	"github.com/stretchr/testify/assert"
@@ -1879,11 +1880,10 @@ func TestCheckRemoteControlConfig(t *testing.T) {
 	assert.True(t, c.RemoteControl.GRPC.GRPCProxyEnabled, "gRPCProxyEnabled should be true when gRPC is enabled")
 }
 
-func TestCheckConfig(t *testing.T) {
-	t.Parallel()
+func newConfigForCheckConfig() *Config {
 	cp1 := currency.NewPair(currency.DOGE, currency.XRP)
 	cp2 := currency.NewPair(currency.DOGE, currency.USD)
-	cfg := &Config{
+	return &Config{
 		Exchanges: []Exchange{
 			{
 				Name:    testFakeExchangeName,
@@ -1909,8 +1909,34 @@ func TestCheckConfig(t *testing.T) {
 			},
 		},
 	}
-	if err := cfg.CheckConfig(); err != nil {
-		t.Fatal(err)
+}
+
+func TestCheckConfigNTP(t *testing.T) {
+	t.Parallel()
+	tests := []struct {
+		name  string
+		level int
+	}{
+		{name: "disabled", level: NTPClientDisabled},
+		{name: "startup", level: NTPClientStartup},
+		{name: "periodic", level: NTPClientPeriodic},
+		{name: "unknown", level: 42},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Parallel()
+			cfg := newConfigForCheckConfig()
+			cfg.NTPClient.Level = tt.level
+
+			require.NoError(t, cfg.CheckConfig(), "CheckConfig must not error")
+			assert.Equal(t, tt.level, cfg.NTPClient.Level, "NTP Level should remain unchanged")
+			require.NotNil(t, cfg.NTPClient.AllowedDifference, "AllowedDifference must be normalized")
+			require.NotNil(t, cfg.NTPClient.AllowedNegativeDifference, "AllowedNegativeDifference must be normalized")
+			assert.Equal(t, 50*time.Millisecond, *cfg.NTPClient.AllowedDifference, "AllowedDifference should use the default")
+			assert.Equal(t, 50*time.Millisecond, *cfg.NTPClient.AllowedNegativeDifference, "AllowedNegativeDifference should use the default")
+			assert.Equal(t, []string{"pool.ntp.org:123"}, cfg.NTPClient.Pool, "NTP pools should use the exact default")
+		})
 	}
 }
 
@@ -1976,33 +2002,48 @@ func TestCheckLoggerConfig(t *testing.T) {
 	}
 }
 
-func TestDisableNTPCheck(t *testing.T) {
-	t.Parallel()
+func TestSetNTPCheck(t *testing.T) {
+	tests := []struct {
+		name          string
+		input         string
+		expectedLevel int
+		expectedText  string
+	}{
+		{
+			name:          "alert at startup",
+			input:         "a\n",
+			expectedLevel: 0,
+			expectedText:  "Time sync has been set to alert",
+		},
+		{
+			name:          "disable",
+			input:         "d\n",
+			expectedLevel: -1,
+			expectedText:  "Future notifications for out of time sync has been disabled",
+		},
+		{
+			name:          "invalid then warn periodically",
+			input:         "invalid\nw\n",
+			expectedLevel: 1,
+			expectedText:  "Time sync has been set to warn only",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			c := Config{NTPClient: NTPClientConfig{Level: 42}}
+			input := strings.NewReader(tt.input)
+			response, err := c.SetNTPCheck(iotest.OneByteReader(input))
+			require.NoError(t, err, "SetNTPCheck must not error")
+			assert.Equal(t, tt.expectedText, response, "SetNTPCheck response should remain unchanged")
+			assert.Equal(t, tt.expectedLevel, c.NTPClient.Level, "SetNTPCheck should set the expected Level")
+			assert.Zero(t, input.Len(), "SetNTPCheck should consume the complete input")
+		})
+	}
 
 	var c Config
-
-	warn, err := c.SetNTPCheck(strings.NewReader("w\n"))
-	if err != nil {
-		t.Fatalf("to create ntpclient failed reason: %v", err)
-	}
-
-	if warn != "Time sync has been set to warn only" {
-		t.Errorf("failed expected %v got %v", "Time sync has been set to warn only", warn)
-	}
-	alert, _ := c.SetNTPCheck(strings.NewReader("a\n"))
-	if alert != "Time sync has been set to alert" {
-		t.Errorf("failed expected %v got %v", "Time sync has been set to alert", alert)
-	}
-
-	disable, _ := c.SetNTPCheck(strings.NewReader("d\n"))
-	if disable != "Future notifications for out of time sync has been disabled" {
-		t.Errorf("failed expected %v got %v", "Future notifications for out of time sync has been disabled", disable)
-	}
-
-	_, err = c.SetNTPCheck(strings.NewReader(" "))
-	if err.Error() != "EOF" {
-		t.Errorf("failed expected EOF got: %v", err)
-	}
+	_, err := c.SetNTPCheck(strings.NewReader(" "))
+	require.ErrorIs(t, err, io.EOF, "SetNTPCheck must propagate EOF")
 }
 
 func TestCheckGCTScriptConfig(t *testing.T) {
@@ -2051,23 +2092,76 @@ func TestCheckDatabaseConfig(t *testing.T) {
 
 func TestCheckNTPConfig(t *testing.T) {
 	t.Parallel()
-	cfg := &Config{
-		NTPClient: NTPClientConfig{},
+	durationPointer := func(value time.Duration) *time.Duration {
+		return &value
+	}
+	validConfig := func() NTPClientConfig {
+		return NTPClientConfig{
+			AllowedDifference:         durationPointer(75 * time.Millisecond),
+			AllowedNegativeDifference: durationPointer(25 * time.Millisecond),
+			Pool:                      []string{"configured.example:123"},
+		}
+	}
+	check := func(t *testing.T, input NTPClientConfig, wantAllowed, wantNegative time.Duration, wantPools []string) {
+		t.Helper()
+		cfg := &Config{NTPClient: input}
+		cfg.CheckNTPConfig()
+		require.NotNil(t, cfg.NTPClient.AllowedDifference, "AllowedDifference must be populated")
+		require.NotNil(t, cfg.NTPClient.AllowedNegativeDifference, "AllowedNegativeDifference must be populated")
+		assert.Equal(t, wantAllowed, *cfg.NTPClient.AllowedDifference, "AllowedDifference should be normalized independently")
+		assert.Equal(t, wantNegative, *cfg.NTPClient.AllowedNegativeDifference, "AllowedNegativeDifference should be normalized independently")
+		assert.Equal(t, wantPools, cfg.NTPClient.Pool, "NTP pool order should be preserved")
+	}
+	invalidValues := []struct {
+		name  string
+		value *time.Duration
+	}{
+		{name: "nil"},
+		{name: "zero", value: durationPointer(0)},
+		{name: "negative", value: durationPointer(-time.Millisecond)},
+	}
+	fields := []struct {
+		name         string
+		setInvalid   func(*NTPClientConfig, *time.Duration)
+		wantAllowed  time.Duration
+		wantNegative time.Duration
+	}{
+		{
+			name:         "allowed difference",
+			setInvalid:   func(c *NTPClientConfig, value *time.Duration) { c.AllowedDifference = value },
+			wantAllowed:  50 * time.Millisecond,
+			wantNegative: 25 * time.Millisecond,
+		},
+		{
+			name:         "allowed negative difference",
+			setInvalid:   func(c *NTPClientConfig, value *time.Duration) { c.AllowedNegativeDifference = value },
+			wantAllowed:  75 * time.Millisecond,
+			wantNegative: 50 * time.Millisecond,
+		},
+	}
+	for _, field := range fields {
+		for _, invalid := range invalidValues {
+			t.Run(invalid.name+" "+field.name, func(t *testing.T) {
+				t.Parallel()
+				input := validConfig()
+				field.setInvalid(&input, invalid.value)
+				check(t, input, field.wantAllowed, field.wantNegative, input.Pool)
+			})
+		}
 	}
 
-	cfg.CheckNTPConfig()
-
-	if cfg.NTPClient.Pool[0] != "pool.ntp.org:123" {
-		t.Error("ntpclient with no valid pool should default to pool.ntp.org")
-	}
-
-	if cfg.NTPClient.AllowedDifference == nil {
-		t.Error("ntpclient with nil alloweddifference should default to sane value")
-	}
-
-	if cfg.NTPClient.AllowedNegativeDifference == nil {
-		t.Error("ntpclient with nil allowednegativedifference should default to sane value")
-	}
+	t.Run("default pool", func(t *testing.T) {
+		t.Parallel()
+		input := validConfig()
+		input.Pool = nil
+		check(t, input, 75*time.Millisecond, 25*time.Millisecond, []string{"pool.ntp.org:123"})
+	})
+	t.Run("valid asymmetric values and ordered pools", func(t *testing.T) {
+		t.Parallel()
+		input := validConfig()
+		input.Pool = []string{"first.example:123", "second.example:123"}
+		check(t, input, 75*time.Millisecond, 25*time.Millisecond, input.Pool)
+	})
 }
 
 func TestCheckCurrencyConfigValues(t *testing.T) {
--- a/engine/helpers_test.go
+++ b/engine/helpers_test.go
@@ -21,7 +21,6 @@ import (
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/thrasher-corp/gocryptotrader/common/convert"
 	"github.com/thrasher-corp/gocryptotrader/common/file"
 	"github.com/thrasher-corp/gocryptotrader/communications"
 	"github.com/thrasher-corp/gocryptotrader/config"
@@ -35,7 +34,6 @@ import (
 	"github.com/thrasher-corp/gocryptotrader/exchanges/protocol"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/stats"
 	"github.com/thrasher-corp/gocryptotrader/gctscript/vm"
-	"github.com/thrasher-corp/gocryptotrader/log"
 )
 
 var testExchange = "Bitstamp"
@@ -148,7 +146,7 @@ func TestSetSubsystem(t *testing.T) { //nolint // TO-DO: Fix race t.Parallel() u
 		},
 		{
 			Subsystem:    NTPManagerName,
-			Engine:       &Engine{Config: &config.Config{Logging: log.Config{Enabled: convert.BoolPtr(false)}}},
+			Engine:       &Engine{Config: &config.Config{}},
 			EnableError:  errNilNTPConfigValues,
 			DisableError: ErrNilSubsystem,
 		},
@@ -206,6 +204,33 @@ func TestSetSubsystem(t *testing.T) { //nolint // TO-DO: Fix race t.Parallel() u
 			require.ErrorIs(t, err, tt.DisableError)
 		})
 	}
+}
+
+func TestSetSubsystemNTPManagerRoundTrip(t *testing.T) {
+	cfg := &config.Config{NTPClient: config.NTPClientConfig{Level: config.NTPClientPeriodic}}
+	cfg.CheckNTPConfig()
+	bot := &Engine{Config: cfg}
+
+	status, ok := bot.GetSubsystemsStatus()["ntp_timekeeper"]
+	require.True(t, ok, "subsystem status must retain the ntp_timekeeper key")
+	assert.False(t, status, "NTP manager should initially be stopped")
+
+	require.NoError(t, bot.SetSubsystem(NTPManagerName, true), "NTP manager subsystem must start")
+	registerNTPManagerCleanup(t, bot.ntpManager)
+	assert.True(t, bot.GetSubsystemsStatus()["ntp_timekeeper"], "ntp_timekeeper status should report running")
+
+	require.NoError(t, bot.SetSubsystem(NTPManagerName, false), "NTP manager subsystem must stop")
+	assert.False(t, bot.GetSubsystemsStatus()["ntp_timekeeper"], "ntp_timekeeper status should report stopped")
+}
+
+func TestSetSubsystemNTPManagerRejectsNonPeriodicLevel(t *testing.T) {
+	cfg := &config.Config{NTPClient: config.NTPClientConfig{Level: config.NTPClientStartup}}
+	cfg.CheckNTPConfig()
+	bot := &Engine{Config: cfg}
+
+	err := bot.SetSubsystem("ntp_timekeeper", true)
+	assert.ErrorIs(t, err, errNTPManagerDisabled, "non-periodic subsystem enable should return errNTPManagerDisabled")
+	assert.False(t, bot.GetSubsystemsStatus()["ntp_timekeeper"], "non-periodic NTP manager should remain stopped")
 }
 
 func TestSetSubsystemEnableDeniedAfterShutdownRequest(t *testing.T) {
--- a/engine/ntp_clock.go
+++ b/engine/ntp_clock.go
@@ -1,0 +1,167 @@
+package engine
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"net"
+	"time"
+
+	"github.com/beevik/ntp"
+)
+
+const (
+	ntpDialTimeout  = 5 * time.Second
+	ntpIOTimeout    = 5 * time.Second
+	maximumDuration = time.Duration(1<<63 - 1)
+	minimumDuration = -maximumDuration - 1
+)
+
+var (
+	errNoNTPPoolsConfigured    = errors.New("no NTP pools configured")
+	errNoValidNTPServer        = errors.New("no valid NTP server")
+	errNegativeNTPRoundTrip    = errors.New("negative NTP round-trip time")
+	errNegativeNTPRootDistance = errors.New("negative NTP root distance")
+	errNTPRootDistanceBelowRTT = errors.New("NTP root distance below half round-trip time")
+)
+
+// clockCorrection is server time minus local time: a positive value means the local clock is behind.
+// uncertainty is the NTP root distance. It bounds how far clockCorrection can be from the true offset.
+type clockObservation struct {
+	source          string
+	clockCorrection time.Duration
+	uncertainty     time.Duration
+	roundTrip       time.Duration
+}
+
+type queryClockFunc func(context.Context, string) (clockObservation, error)
+
+type clockMeasurer struct {
+	queryOne queryClockFunc
+}
+
+// Servers are asked in order and the first valid reply wins.
+// Selection never depends on the answer: a server is not skipped because its result is inconvenient.
+func (m clockMeasurer) measure(ctx context.Context, pools []string) (clockObservation, error) {
+	if len(pools) == 0 {
+		return clockObservation{}, errNoNTPPoolsConfigured
+	}
+
+	failures := make([]error, 0, len(pools))
+	for i := range pools {
+		if err := ctx.Err(); err != nil {
+			return clockObservation{}, err
+		}
+
+		observation, err := m.queryOne(ctx, pools[i])
+		// Drop even a successful observation after cancellation so shutdown
+		// cannot reach the startup prompt.
+		if contextErr := ctx.Err(); contextErr != nil {
+			return clockObservation{}, contextErr
+		}
+		if err == nil {
+			return observation, nil
+		}
+		failures = append(failures, fmt.Errorf("NTP server %q: %w", pools[i], err))
+	}
+
+	return clockObservation{}, fmt.Errorf("%w: %w", errNoValidNTPServer, errors.Join(failures...))
+}
+
+func queryClock(ctx context.Context, source string) (clockObservation, error) {
+	if err := ctx.Err(); err != nil {
+		return clockObservation{}, err
+	}
+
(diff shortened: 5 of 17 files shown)


thrasher-:
@codex please review

gloriousCode:
Thank you for sitting down, putting on your writer's hat and writing 2600 words on your 1440 line change PR without any help

config/config_test.go
```diff
@@ -2051,23 +2092,76 @@ func TestCheckDatabaseConfig(t *testing.T) {
...
-
-	if cfg.NTPClient.Pool[0] != "pool.ntp.org:123" {
-		t.Error("ntpclient with no valid pool should default to pool.ntp.org")
+	durationPointer := func(value time.Duration) *time.Duration {
+		return &value
+	}
+	validConfig := func() NTPClientConfig {
+		return NTPClientConfig{
+			AllowedDifference:         durationPointer(75 * time.Millisecond),
+			AllowedNegativeDifference: durationPointer(25 * time.Millisecond),
```
shazbert, line 2101: suggestion: This field types may not need to be pointers

config/config_test.go
```diff
@@ -2051,23 +2092,76 @@ func TestCheckDatabaseConfig(t *testing.T) {
...
+		name         string
+		setInvalid   func(*NTPClientConfig, *time.Duration)
+		wantAllowed  time.Duration
+		wantNegative time.Duration
+	}{
+		{
+			name:         "allowed difference",
+			setInvalid:   func(c *NTPClientConfig, value *time.Duration) { c.AllowedDifference = value },
+			wantAllowed:  50 * time.Millisecond,
+			wantNegative: 25 * time.Millisecond,
```
shazbert, line 2133: suggestion: We could just have a single time window instead of forward and backwards bounding. I don't know if this adjustability is needed. 🤷

engine/helpers_test.go
```diff
@@ -208,6 +206,33 @@ func TestSetSubsystem(t *testing.T) { //nolint // TO-DO: Fix race t.Parallel() u
 	}
 }
 
+func TestSetSubsystemNTPManagerRoundTrip(t *testing.T) {
+	cfg := &config.Config{NTPClient: config.NTPClientConfig{Level: config.NTPClientPeriodic}}
```
shazbert, line 210: t.Parallel()

engine/helpers_test.go
```diff
@@ -208,6 +206,33 @@ func TestSetSubsystem(t *testing.T) { //nolint // TO-DO: Fix race t.Parallel() u
...
+	require.NoError(t, bot.SetSubsystem(NTPManagerName, true), "NTP manager subsystem must start")
+	registerNTPManagerCleanup(t, bot.ntpManager)
+	assert.True(t, bot.GetSubsystemsStatus()["ntp_timekeeper"], "ntp_timekeeper status should report running")
+
+	require.NoError(t, bot.SetSubsystem(NTPManagerName, false), "NTP manager subsystem must stop")
+	assert.False(t, bot.GetSubsystemsStatus()["ntp_timekeeper"], "ntp_timekeeper status should report stopped")
+}
+
+func TestSetSubsystemNTPManagerRejectsNonPeriodicLevel(t *testing.T) {
+	cfg := &config.Config{NTPClient: config.NTPClientConfig{Level: config.NTPClientStartup}}
```
shazbert, line 227: t.Parallel()

engine/helpers_test.go
```diff
@@ -208,6 +206,33 @@ func TestSetSubsystem(t *testing.T) { //nolint // TO-DO: Fix race t.Parallel() u
...
+	cfg.CheckNTPConfig()
+	bot := &Engine{Config: cfg}
+
+	err := bot.SetSubsystem("ntp_timekeeper", true)
+	assert.ErrorIs(t, err, errNTPManagerDisabled, "non-periodic subsystem enable should return errNTPManagerDisabled")
+	assert.False(t, bot.GetSubsystemsStatus()["ntp_timekeeper"], "non-periodic NTP manager should remain stopped")
```
shazbert, line 233: Can you please assert true so that the key is maintained or not?

engine/ntp_clock.go
```diff
@@ -0,0 +1,167 @@
...
+		if err := ctx.Err(); err != nil {
+			return clockObservation{}, err
+		}
+
+		observation, err := m.queryOne(ctx, pools[i])
+		// Drop even a successful observation after cancellation so shutdown
+		// cannot reach the startup prompt.
+		if contextErr := ctx.Err(); contextErr != nil {
+			return clockObservation{}, contextErr
+		}
```
shazbert, line 61: This can be done and this block removed: https://github.com/thrasher-corp/gocryptotrader/pull/2327/changes#r3732053224

engine/ntp_clock.go
```diff
@@ -0,0 +1,167 @@
...
+)
+
+// evaluateClock compares the complete uncertainty interval with the tolerance window.
+// A single value would report network delay as clock error, so a result is conclusive
+// only when the whole interval falls outside the window.
+func evaluateClock(
+	observation clockObservation,
+	allowedDifference,
+	allowedNegativeDifference time.Duration,
+) clockSyncState {
```
shazbert, line 129: we can one liner these params

engine/ntp_clock_test.go
```diff
@@ -0,0 +1,391 @@
...
+	"net"
+	"testing"
+	"time"
+
+	"github.com/beevik/ntp"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestClockMeasurerEmptyPools(t *testing.T) {
```
shazbert, line 18: can called t.Parallel on these tests.

engine/ntp_manager.go
```diff
@@ -94,111 +72,130 @@ func (m *ntpManager) Stop() error {
...
 }
 
-// run continuously checks the internet connection at intervals
+// run periodically checks the configured time sources.
 func (m *ntpManager) run() {
-	t := time.NewTicker(m.checkInterval)
-	defer func() {
-		t.Stop()
-	}()
+	ticker := time.NewTicker(m.checkInterval)
```
shazbert, line 77: ticker is a package conflic, could you please rename?

engine/ntp_manager.go
```diff
@@ -94,111 +72,130 @@ func (m *ntpManager) Stop() error {
...
-		return time.Time{}, fmt.Errorf("NTP manager %w", ErrSubSystemNotStarted)
+func (m *ntpManager) checkClock(ctx context.Context) (clockObservation, clockSyncState, error) {
+	observation, err := m.measure(ctx, m.pools)
+	if err != nil {
+		return clockObservation{}, clockSyncInconclusive, fmt.Errorf("NTP clock measurement failed: %w", err)
 	}
-	return m.checkTimeInPools(), nil
+	return observation,
+		evaluateClock(observation, m.allowedDifference, m.allowedNegativeDifference),
+		nil
```
shazbert, line 99: can one line this

engine/ntp_manager.go
```diff
@@ -94,111 +72,130 @@ func (m *ntpManager) Stop() error {
...
 }
```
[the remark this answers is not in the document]
shazbert: I don't think this is necessary, but up to you.

engine/ntp_manager_test.go
```diff
@@ -1,167 +1,431 @@
...
+		if m.IsRunning() {
+			require.NoError(t, m.Stop(), "started manager cleanup must stop")
+		}
+	})
+}
 
-	if m.IsRunning() {
-		t.Error("expected false")
-	}
+func TestSetupNTPManager(t *testing.T) {
```
shazbert, line 40: t.Parrallel across these testerinos

shazbert (changes requested):
Here's a quick review, there is a new dependency, do we need that dependency? Things are looking good.

engine/ntp_clock.go
```diff
@@ -0,0 +1,167 @@
...
+	dialer := &net.Dialer{Timeout: ntpDialTimeout}
+	response, err := ntp.QueryWithOptions(source, ntp.QueryOptions{
+		Timeout: ntpIOTimeout,
+		Dialer: func(_, remoteAddress string) (net.Conn, error) {
+			return dialer.DialContext(ctx, "udp", remoteAddress)
+		},
+	})
+	if err != nil {
+		return clockObservation{}, fmt.Errorf("query: %w", err)
+	}
```
thrasher-, line 85: Heimdall: **Should fix** — the startup budget and cancellation stop at the dial; an in-flight read runs to the library's own 5 s deadline regardless.

**Fix:** derive the library timeout from the context deadline and close the socket when the context is cancelled:

```suggestion
	// The library sets its own read deadline after dialing, so derive it from the
	// context deadline and close the socket on cancellation; otherwise an in-flight
	// read outlives both the startup budget and engine shutdown.
	timeout := ntpIOTimeout
	if deadline, ok := ctx.Deadline(); ok {
		timeout = min(timeout, time.Until(deadline))
		if timeout <= 0 {
			return clockObservation{}, context.DeadlineExceeded
		}
	}
	dialer := &net.Dialer{Timeout: ntpDialTimeout}
	stopCloseOnCancel := func() bool { return false }
	response, err := ntp.QueryWithOptions(source, ntp.QueryOptions{
		Timeout: timeout,
		Dialer: func(_, remoteAddress string) (net.Conn, error) {
			conn, err := dialer.DialContext(ctx, "udp", remoteAddress)
			if err != nil {
				return nil, err
			}
			stopCloseOnCancel = context.AfterFunc(ctx, func() { _ = conn.Close() })
			return conn, nil
		},
	})
	stopCloseOnCancel()
	if err != nil {
		if ctxErr := ctx.Err(); ctxErr != nil {
			return clockObservation{}, ctxErr
		}
		return clockObservation{}, fmt.Errorf("query: %w", err)
	}
```

The `<= 0` guard matters because the library treats `Timeout == 0` as "use the 5 s default" ([`ntp.go#L482`](https://github.com/beevik/ntp/blob/v1.5.0/ntp.go#L482)). `stopCloseOnCancel` is assigned inside the dialer, which the library calls synchronously before returning, and releasing it after the query means nothing stays registered on the context; on the periodic loop's `context.Background()` the hook never fires at all. The `ctx.Err()` check on the error path is what lets callers see `context.Canceled` or `context.DeadlineExceeded` rather than the socket's "use of closed network connection" once the hook has fired. A loopback test for "cancel during read" next to `TestQueryClockLoopbackNegativeCorrectionSign` would pin it.

**Why:** `dialer.DialContext` only observes `ctx` while resolving and creating the UDP socket; the library then calls `con.SetDeadline(time.Now().Add(opt.Timeout))` and blocks in `con.Read` ([`ntp.go#L531-L584`](https://github.com/beevik/ntp/blob/v1.5.0/ntp.go#L531-L584)). Measured on this head with `clockMeasurer{queryOne: queryClock}` against local UDP listeners that never reply:

```
2 silent servers, 7s budget:    elapsed=10.006s  err=context deadline exceeded
cancel at 0.5s during read:     elapsed=5.002s   err=query: read udp 127.0.0.1:60104->127.0.0.1:60709: i/o timeout
```

With the suggestion applied:

```
2 silent servers, 7s budget:    elapsed=7.001s   err=context deadline exceeded
cancel at 0.5s during read:     elapsed=501ms    err=context canceled
measure() with cancel at 0.3s:  elapsed=301ms    err=context canceled
background ctx, silent server:  elapsed=5.003s   err=query: read udp ...: i/o timeout   (periodic path unchanged)
```

So today the advertised 10 s startup budget is really "10 s plus one read", and `RequestShutdown` during the startup check waits out the read. The PR body documents this as a library limit; it is not one, the library just needs the deadline and a close hook. This is also what @shazbert asked for on [`ntp_clock.go:61`](https://github.com/thrasher-corp/gocryptotrader/blob/25d27fc80f5e3b5282dd2b156cc14153369162da/engine/ntp_clock.go#L61). The suggestion passes gofumpt, lint, and `go test ./engine ./config -race` as posted.

engine/ntp_manager.go
```diff
@@ -94,111 +72,130 @@ func (m *ntpManager) Stop() error {
...
+		observation.uncertainty,
+		observation.roundTrip,
+		observation.source,
+		m.allowedNegativeDifference,
+		m.allowedDifference,
+	)
+	if state == clockSyncInconclusive {
+		message += "; no conclusion was possible; configure a closer or lower-latency NTP server"
+	}
+	return message, state != clockSyncInSync
```
thrasher-, line 187: Heimdall: **Should fix** — a correct clock on the shipped default config still gets a WARN telling the user to change their NTP servers.

**Fix:** report an inconclusive result whose point estimate is inside the window without the warning, and keep the warning when the point estimate is outside it. Apply together with the test update in the paired suggestion ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760372)); `TestClockObservationReport` pins the current behaviour with a 45 ms correction inside the 50 ms window.

```suggestion
	if state != clockSyncInconclusive {
		return message, state != clockSyncInSync
	}
	message += "; no conclusion was possible; configure a closer or lower-latency NTP server"
	// A point estimate inside the window with an interval that crosses it is far more
	// likely a distant server than a bad clock, so it is reported without a warning.
	inside := -m.allowedNegativeDifference <= observation.clockCorrection && observation.clockCorrection <= m.allowedDifference
	return message, !inside
```

**Why:** the radius is the root distance, and for the shipped `pool.ntp.org:123` that depends entirely on which pool member DNS hands out. Live replies from here (Australia, home connection), same clock, minutes apart:

```
source                 rtt        root distance
pool.ntp.org:123       4.3ms      3.9ms
pool.ntp.org:123       4.0ms      4.6ms
pool.ntp.org:123       29.8ms     14.9ms
pool.ntp.org:123       9.1ms      61.3ms     inconclusive for ANY clock at ±50ms
pool.ntp.org:123       251.1ms    249.9ms    inconclusive for ANY clock at ±50ms
0.au.pool.ntp.org:123  3.7ms      46.7ms     root dispersion 42.2ms; inconclusive once |offset| > 3.3ms
0.au.pool.ntp.org:123  4.1ms      48.5ms     root dispersion 46.4ms; inconclusive once |offset| > 1.5ms
time.google.com:123    99.0ms     49.6ms     inconclusive once |offset| > 0.4ms
```

Re-evaluating the replies as a healthy clock (subtracting the median measured offset, residuals all within ±4 ms) with the ±50 ms defaults gave `inconclusive` on 2 of 10 default-pool replies and 2 of 12 overall. That is one startup in five logging

```
NTP manager: state=inconclusive correction=-4.171ms ... root-distance=±61.25ms ... ; no conclusion was possible; configure a closer or lower-latency NTP server
```

at warn level for a clock that is 4 ms off. It is far better than master's false "out of sync", but it is still the feature crying wolf on the default path, and the advice it gives (reconfigure servers) is wrong for a user whose clock is fine. Root distance is the right radius for a conclusive ahead/behind call (RFC 5905 §4 calls the synchronization distance "the maximum error due to all causes", and the library's `(RootDelay + RTT)/2 + RootDispersion` is the one-shot SNTP form of it); it is just too conservative to justify a warning on its own when the point estimate is 4 ms. Both suggestions together pass gofumpt, lint, and `go test ./engine ./config -race`.

engine/ntp_manager_test.go
```diff
@@ -1,167 +1,431 @@
...
-	err = m.Start()
-	assert.NoError(t, err)
-
-	tt, err := m.FetchNTPTime()
-	assert.NoError(t, err)
-
-	if tt.IsZero() {
-		t.Error("expected time")
+	message, warning := m.clockObservationReport(observation, clockSyncInconclusive)
+	assert.True(t, warning, "Inconclusive should use warning severity")
```
thrasher-, line 202: Heimdall: **Should fix** — companion to the report-severity change ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760363)); the two suggestions apply as a pair.

**Fix:** the 45 ms correction in this test sits inside the 20/50 ms window, so with the severity change it must not warn, and a point estimate outside the window should still warn:

```suggestion
	message, warning := m.clockObservationReport(observation, clockSyncInconclusive)
	assert.False(t, warning, "Inconclusive with the point estimate inside the window should not use warning severity")
	_, warning = m.clockObservationReport(clockObservation{clockCorrection: 55 * time.Millisecond, uncertainty: 10 * time.Millisecond}, clockSyncInconclusive)
	assert.True(t, warning, "Inconclusive with the point estimate outside the window should use warning severity")
```

**Why:** this assertion is the only thing pinning "inconclusive always warns", and the paired suggestion changes exactly that case. The rest of the test (message contents, in-sync at debug, ahead/behind at warn) is unaffected.

engine/ntp_manager.go
```diff
@@ -94,111 +72,130 @@ func (m *ntpManager) Stop() error {
...
 
-		err = con.Close()
-		if err != nil {
-			log.Errorln(log.TimeMgr, err)
-		}
-		return time.Unix(int64(secs), nanos)
+	responseMessage, err := bot.Config.SetNTPCheck(input)
+	if err != nil {
+		return fmt.Errorf("unable to set NTP check: %w", err)
+	}
```
thrasher-, line 154: Heimdall: **Minor** — a headless deployment at the default `enabled: 0` now starts fine until the day its clock drifts, then refuses to start.

**Fix:** consider treating `io.EOF` from the prompt as "no operator present": log the measured observation at warn and continue, keeping the abort for other read errors. Not posted as a suggestion because it is a policy decision for the maintainers, needs an `errors` import, and flips `TestHandleStartupNTPPolicyFailures/prompt_EOF`; the shape would be:

```go
	responseMessage, err := bot.Config.SetNTPCheck(input)
	if err != nil {
		if errors.Is(err, io.EOF) {
			log.Warnln(log.TimeMgr, "NTP manager: no interactive input available; leaving the NTP setting unchanged")
			return nil
		}
		return fmt.Errorf("unable to set NTP check: %w", err)
	}
```

**Why:** `SetNTPCheck` returns `io.EOF` when stdin is closed (systemd, docker without `-i`), and `main.go` treats a `Start` error as fatal ([`main.go:151-160`](https://github.com/thrasher-corp/gocryptotrader/blob/25d27fc80f5e3b5282dd2b156cc14153369162da/main.go#L151-L160)). On master the prompt was unconditional at level 0, so headless-at-level-0 always failed at startup and nobody could run that way. With this change the same configuration works for months and then fails precisely when the clock is conclusively wrong, which is the moment the user most needs the process up and the warning in the log.

engine/ntp_clock.go
```diff
@@ -0,0 +1,167 @@
...
+		return clockObservation{}, errNegativeNTPRoundTrip
+	}
+	if response.RootDistance < 0 {
+		return clockObservation{}, errNegativeNTPRootDistance
+	}
+	// Path asymmetry can put all of the delay in one direction, so half the round-trip
+	// time is the smallest honest uncertainty. A smaller value means the reply contradicts itself.
+	if response.RootDistance < response.RTT/2 {
+		return clockObservation{}, errNTPRootDistanceBelowRTT
+	}
```
thrasher-, line 103: Heimdall: **Minor** — these three rejections cannot fire with beevik/ntp v1.5.0.

**Fix:** none required; either keep them as a guard against a future library change and say so in the comment, or drop them with their three test cases.

**Why:** the library clamps a negative round trip to zero ([`rtt()`](https://github.com/beevik/ntp/blob/v1.5.0/ntp.go#L753-L761)), decodes `RootDelay` and `RootDispersion` from unsigned Q16.16 so both are ≥ 0, and computes `RootDistance = (RTT + RootDelay)/2 + RootDispersion` ([`rootDistance()`](https://github.com/beevik/ntp/blob/v1.5.0/ntp.go#L800-L815)), which is ≥ `RTT/2` by construction. The PR body describes them as rejecting "a reply when its own values disagree", but the value being checked is derived by the library from the same inputs, so it cannot disagree with them.

engine/ntp_clock.go
thrasher-, line 61: Heimdall: **Minor** — this discard is still needed after the context fix, so I would keep it.

**Fix:** keep the block, or move the check to `handleStartupNTPPolicy` as a `parent.Err()` test immediately before prompting.

**Why:** once the socket is closed on cancellation ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760346)), a cancellation that lands after `con.Read` has already returned still yields a valid observation from `queryOne`. Without this check `measure` would hand it to the startup policy, which could then block on stdin during shutdown. The context fix shortens that window and this check narrows it further, but neither closes it: `SetNTPCheck` is not cancellation-aware and holds the config mutex while it waits on stdin (pre-existing, master prompted the same way), so a cancellation that lands between this check and the prompt still reaches it. Closing it fully means a cancellation-aware prompt, which is beyond this PR.

engine/ntp_manager.go
```diff
@@ -94,111 +72,130 @@ func (m *ntpManager) Stop() error {
...
-			log.Warnf(log.TimeMgr, "Unable to read. Error: %s\n", err)
-			err = con.Close()
-			if err != nil {
-				log.Errorln(log.TimeMgr, err)
-			}
-			continue
+	ctx, cancel := context.WithTimeout(parent, budget)
+	defer cancel()
+
+	observation, state, err := bot.ntpManager.checkClock(ctx)
```
thrasher-, line 137: Heimdall: **Minor** — when UDP/123 is silently dropped, startup now pauses 5 s per pool member with nothing in the log until the failure line.

**Fix:** say what is happening before the query:

```suggestion
	log.Infoln(log.TimeMgr, "NTP manager: checking the system clock against the configured NTP servers")
	observation, state, err := bot.ntpManager.checkClock(ctx)
```

**Why:** master's `Engine.Start` never measured (it prompted, then constructed a manager it never started), so this latency is new. A silent server costs exactly one read timeout (measured `5.002s` against a local UDP listener that never replies), so the shipped one-server pool adds up to 5 s behind a firewall that drops UDP, two members add up to 10 s, capped by the budget plus one in-flight read; a slow resolver adds its own time on top, up to the 5 s dial timeout per member. A refused or unroutable port fails fast, so this only bites on drop-without-reject networks. Verified with gofumpt, lint and the race tests.

cmd/documentation/config_templates/config_readme.tmpl
```diff
@@ -227,21 +227,43 @@ See the section `exchange.features.enabled.subscriptions` for configuring subscr
...
-## Configure Network Time Server 
+## Configure Network Time Server
 
-+ To configure and enable a NTP server you need to set the "enabled" field to one of 3 values -1 is disabled 0 is enabled and alert at start up 1 is enabled and warn at start up
-servers are configured by the pool array and attempted first to last allowedDifference and allowedNegativeDifference are how far ahead and behind is acceptable for the time to be out in nanoseconds
+The `enabled` field controls when NTP clock checks run:
+
+These settings take effect only when the engine's NTP client feature is enabled; the command-line
+application enables that feature by default.
+
```
thrasher-, line 236: Heimdall: **Minor** — the colon on "controls when NTP clock checks run:" leads into the "These settings take effect only when…" paragraph rather than the list it introduces.

**Fix:** put the feature-flag paragraph first:

```suggestion
These settings take effect only when the engine's NTP client feature is enabled; the command-line
application enables that feature by default.

The `enabled` field controls when NTP clock checks run:

```

The same change to the generated `config/README.md` is in the paired suggestion ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760416)); with both applied the doc generator reproduces the README with no drift.

config/README.md
```diff
@@ -244,21 +244,43 @@ See the section `exchange.features.enabled.subscriptions` for configuring subscr
...
-## Configure Network Time Server 
+## Configure Network Time Server
 
-+ To configure and enable a NTP server you need to set the "enabled" field to one of 3 values -1 is disabled 0 is enabled and alert at start up 1 is enabled and warn at start up
-servers are configured by the pool array and attempted first to last allowedDifference and allowedNegativeDifference are how far ahead and behind is acceptable for the time to be out in nanoseconds
+The `enabled` field controls when NTP clock checks run:
+
+These settings take effect only when the engine's NTP client feature is enabled; the command-line
+application enables that feature by default.
+
```
thrasher-, line 253: Heimdall: **Minor** — generated counterpart of the template fix ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760404)); apply both or regenerate.

```suggestion
These settings take effect only when the engine's NTP client feature is enabled; the command-line
application enables that feature by default.

The `enabled` field controls when NTP clock checks run:

```

thrasher- (changes requested):
Heimdall: This fixes a real measurement bug, and the fix is right. I ran the two formulas side by side on the same live replies: master's `T3 − T4` differed from the PR's four-timestamp offset by exactly half the round-trip every time (−1.5 ms at 2.9 ms RTT, −49.2 ms at 98 ms RTT), while the PR's offset agreed across five server families and RTTs from 3 ms to 251 ms to within ±4 ms. The pure `evaluateClock` classifier, the hermetic loopback server in the tests, and the "measure before prompting" ordering are all improvements, and the write-up is unusually careful about what is and is not being claimed.

Nothing here is a blocker. Two things I would fix before merge, both measured rather than inferred:

- **The startup budget and cancellation do not reach an in-flight read.** Two silent servers with a 7 s budget returned after 10.006 s; a cancel at 0.5 s returned after 5.002 s. The PR body documents this as a library limitation, but it is fixable in `queryClock` in about ten lines, and it is a one-click suggestion ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760346)). This is also what @shazbert asked for on [`ntp_clock.go:61`](https://github.com/thrasher-corp/gocryptotrader/blob/25d27fc80f5e3b5282dd2b156cc14153369162da/engine/ntp_clock.go#L61).
- **A correct clock still gets a WARN on the default config.** Root distance from `pool.ntp.org` here ranged from 2.7 ms to 250 ms depending on which pool member DNS returned; with the ±50 ms default, a clock that is fine produced `state=inconclusive` on 2 of 12 replies and the log tells the user to reconfigure their NTP servers. Suggestion plus its paired test update ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760363), [test](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760372)).

Minor, no action required to approve: the headless/EOF failure mode ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760383)), the three unreachable metadata checks ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760385)), the post-cancellation discard being still needed after the context fix ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760391)), the silent 5 s pause when UDP/123 is dropped ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760393)), a pre-existing quirk I cannot anchor inline because the lines are untouched ([`SetNTPCheck`](https://github.com/thrasher-corp/gocryptotrader/blob/25d27fc80f5e3b5282dd2b156cc14153369162da/config/config.go#L1281-L1284) drops a final answer with no trailing newline: `ReadString` returns `"d", io.EOF` and the EOF wins), and a doc-template ordering nit ([template](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760404), [generated README](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760416)).

**On the dependency** (@shazbert's question): `github.com/beevik/ntp` v1.5.0 is BSD-2-Clause, is the [current tag](https://github.com/beevik/ntp/releases/tag/v1.5.0) (tagged 2025-09-29, repo last pushed 2026-02-22), and its requirements (`golang.org/x/net`, `golang.org/x/sys`, testify) are already in the module graph, so it adds no new transitive modules and `go mod tidy -diff` is clean. What it buys over the deleted hand-written packet code is the offset/RTT/root-distance arithmetic and the reply validation (mode, stratum, kiss-o'-death, origin echo, freshness), which is exactly the part master got wrong. I think it is worth it, but that is the maintainers' call.

**Also verified:** every suggestion below was applied to the head as posted and passes gofumpt, golangci-lint with no findings beyond master's, `go test ./engine ./config -race`, `scripts/misc_checks.sh` and the doc generator; `go test ./engine ./config -race` passes on the untouched head and on master (`9ced76445`); the only lint delta against master under golangci-lint 2.13.2 is two `goconst` hits on test strings in `ntp_clock_test.go` (CI's pinned linter passes); `scripts/misc_checks.sh` passes; the doc generator reproduces `engine/ntp_manager.md` and `config/README.md` with no drift; adding `t.Parallel()` to every test in `ntp_clock_test.go`, `ntp_manager_test.go`, the two new `helpers_test.go` tests and `TestSetNTPCheck` is race-clean across `-count 3 -shuffle=on`, so the outstanding `t.Parallel` requests are safe to apply. The 30-minute interval is inside the [NTP Pool vendor guidance](https://www.ntppool.org/vendors.html) ("4-5 times an hour" at most); master's 30 s was not. That same page says applications "must absolutely not use the default pool.ntp.org zone names" as a shipped default. The PR body flags this as a separate project decision, and it is, but note that this PR is what turns the shipped default from a prompt into an automatic query of `pool.ntp.org` on every start (master never queried at startup), so the vendor-zone request should be filed as an issue now rather than left open-ended.

**CI:** the macOS arm64 failure is `exchanges/huobi` live tests (`TestGetRecentTrades`, `TestFQueryHisOpenInterest`, `TestFLastTradeData`) timing out, unrelated to this change. `codecov/project` reports −1.19% against base `39319b1` with 405 files of indirect changes and 27k more lines in the head report than the base, so it is the stale base report rather than this change (patch coverage is 92.9%).

**To unblock:**
1. Bound the read by the context deadline and close the socket on cancellation ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760346))
2. Log an inconclusive result whose point estimate is inside the window without the warning ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760363) with its [test](https://github.com/thrasher-corp/gocryptotrader/pull/2327#discussion_r3909760372))

PR #1477, opened 2024-02-13
exchanges/request: abstract and consolidate rate limiting code to request package

shazbert:
# PR Description

Pulled from initial draft #1151 

* Allows context cancellation to occur while routine is being rate limited for any outbound REST request
* Reduces wrapper code size and bespoke implementations
* Fixes a couple of endpoints defaulting to unauth rates [BUG]

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2330, opened 2026-08-14
bybit: correct rate-limit definitions

shazbert:
## Summary

Corrects Bybit rate-limit definitions where endpoint quotas were also used as per-request weights.

`RateLimiterWithWeight` consumes one reservation for each unit of weight. Configuring both `actions` and `weight` to the documented request quota caused many endpoints to consume their entire allowance per call, reducing throughput to roughly one request per second and potentially delaying the first request.

## Changes

- Sets affected Bybit endpoint weights to one reservation per request.
- Retains documented endpoint quotas in the limiter action rate.
- Corrects these endpoint limits:
  - Account fee rate: 5 requests per second.
  - Create sub-member: 1 request per second.
  - Create sub-API key: 1 request per second.
- Adds a nil-safe `Weight()` accessor to `RateLimiterWithWeight`.
- Adds generalized coverage preventing fixed-cost endpoints from duplicating their request quota as weight.
- Excludes batch endpoints from the fixed weight invariant because Bybit charges them according to the number of orders in each request.
- Adds a TODO to replace the IP-wide `defaultEPL` fallback with endpoint-specific per-UID limits.

## Type of change

## How has this been tested

```console
go test ./exchanges/bybit ./exchanges/request -race -count=1
go test ./exchanges/bybit -run 'Test(GetWSRateLimitEPLByCategory|RateLimits)$' -count=2
golangci-lint run ./exchanges/bybit/... ./exchanges/request/...
```

## Checklist


thrasher-:
@codex please review

exchanges/bybit/ratelimit_test.go
```diff
@@ -33,3 +33,8 @@ func TestGetWSRateLimitEPLByCategory(t *testing.T) {
 		})
 	}
 }
+
+func TestPositionSetLeverageRateLimit(t *testing.T) {
```
thrasher-: Heimdall: **Nit** — two small things, and a suggestion that would make this test carry far more weight.

No `t.Parallel()` — `TestGetWSRateLimitEPLByCategory` directly above uses it at both levels.

It also fails under `-count 2`, because it consumes the single burst token from the package-level `rateLimits` map and the limiter only refills at 10/s:

```
-count 1 -> ok
-count 2 -> FAIL: delay not allowed (ratelimit_test.go:39)
```

CI runs `-count 1` so this doesn't break the build, and it's somewhat intrinsic — `weight` is unexported with no accessor, so a test outside `request` can only probe behaviourally. Worth knowing rather than necessarily fixing: any future test that exercises `postPositionSetLeverageEPL` will perturb this one.

**The suggestion:** generalise it. A table over the whole `rateLimits` map asserting that no EPL requires a delay on its first request catches every instance of this bug in one test, in the spirit of gateio's `TestRateLimits`. Run against this branch it reports:

```
EPLs that DELAY on their very first request: 43 of 65
```

One test that would have caught this class of bug, and will catch the next one.

thrasher- (approved):
Heimdall: Correct fix, verified. Approving — the finding below is about the other 39 instances, not this one.

`NewRateLimit` hard-codes `burst = 1` and `RateLimit()` loops `ReserveN(tn, 1)` *weight* times, so `weight` is a per-request cost against a bucket that only ever holds one token. Measured directly:

```
(1s, 10, 10)  old  ->  first call: ErrDelayNotAllowed (~900ms)   sustained ~1.02 req/s
(1s, 10,  1)  new  ->  first call: nil                           sustained ~12.48 req/s
```

Matches the throughput claim in the description exactly. The test earns its place too — I reverted `ratelimit.go` while keeping it and it fails:

```
--- FAIL: TestPositionSetLeverageRateLimit
    delay not allowed
    first position leverage request must not require a rate limit delay
```

**The same defect is still in 39 other entries.** Setting `weight == actions` means each request costs the entire per-second budget, collapsing the endpoint to ~1 req/s and making the `actions` value inert:

```
(time.Second, 10, 10) x 18   -> ~1.02 req/s
(time.Second,  5,  5) x  8   -> ~1.04 req/s
(time.Second, 50, 50) x  6   -> ~1.00 req/s
(time.Second, 20, 20) x  5   -> ~1.01 req/s
(time.Second,  2,  2) x  2   -> ~1.00 req/s
(time.Second, 20, 50) x  3   -> ~0.40 req/s   <- weight exceeds actions
```

The costliest leftovers sit in the same block as the one being fixed — `createOrderEPL`, `amendOrderEPL`, `cancelOrderEPL` and `createBatchOrderEPL` at [ratelimit.go:84-93](https://github.com/thrasher-corp/gocryptotrader/blob/2cd5ea37aa3a7704a6aaf2dbeff40252a88d6ce5/exchanges/bybit/ratelimit.go#L84-L93) are all throttled to ~1 req/s instead of 10-20/s. Order placement is a bigger throughput win than set-leverage.

It predates the rate-limit consolidation rather than being introduced by it. Before #1477 the code read `limiter, tokens = r.CreateOrderRate, 10` where `CreateOrderRate` was already `NewRateLimit(time.Second, 10)` — the documented rate written twice, once as the limiter rate and once as the reservation count. #1477 migrated that faithfully into `NewRateLimitWithWeight(time.Second, 10, 10)`.

The correct convention is already in the same file four times over: `defaultEPL` (weight 1, carrying the docs link), `cancelAllEPL` via `NewWeightedRateLimitByDuration`, the `time.Minute, N, 1` block at [ratelimit.go:105-122](https://github.com/thrasher-corp/gocryptotrader/blob/2cd5ea37aa3a7704a6aaf2dbeff40252a88d6ce5/exchanges/bybit/ratelimit.go#L105-L122), and the `wsOrder*EPL` block at [ratelimit.go:144-147](https://github.com/thrasher-corp/gocryptotrader/blob/2cd5ea37aa3a7704a6aaf2dbeff40252a88d6ce5/exchanges/bybit/ratelimit.go#L144-L147). Both conventions sit side by side and only one can be right.

Happy for the rest as a follow-up — flagging it because a one-line fix in a 43-entry problem is easy to lose. Test nits are in the comment on the test file.

CI: the failing jobs are `exchanges/huobi` timeouts, unrelated. Locally lint is clean and the bybit suite passes with `-race`.

thrasher- (changes requested):
Heimdall: Re-checked at `9f10dee`. The sweep is thorough and the test is the right shape — 43 entries converted, `Weight()` added to the request package with a nil guard, and `TestRateLimits` replacing the single-endpoint test. I mutation-tested it: reverting one entry to weight 10 fails it, and it passes `-count 2` where the old one didn't.

**I approved this too early and I'm withdrawing that.** I approved when it was a one-line change. It's now a 43-entry sweep, and going back to Bybit's live rate-limit table turns up two things the one-liner didn't have.

**Three endpoints are now configured faster than Bybit documents.** Weight was the throttle holding these under; with it uniformly 1, `actions` governs for the first time and these run at their configured rate:

| EPL | Endpoint | Documented | Configured |
|---|---|---|---|
| [`getAccountFeeEPL`](https://github.com/thrasher-corp/gocryptotrader/blob/9f10dee97febd98fab243f6733998014edcb88de/exchanges/bybit/ratelimit.go#L104) | `/v5/account/fee-rate` | 5/s | **10/s** |
| [`userCreateSubMemberEPL`](https://github.com/thrasher-corp/gocryptotrader/blob/9f10dee97febd98fab243f6733998014edcb88de/exchanges/bybit/ratelimit.go#L123) | `/v5/user/create-sub-member` | 1/s | **5/s** |
| [`userCreateSubAPIKeyEPL`](https://github.com/thrasher-corp/gocryptotrader/blob/9f10dee97febd98fab243f6733998014edcb88de/exchanges/bybit/ratelimit.go#L124) | `/v5/user/create-sub-api` | 1/s | **5/s** |

`getAccountFeeEPL` was comfortably under at ~1 req/s before and is now double the limit. The two user endpoints were marginally over already and are now 5x. Bybit answers with `retCode 10006, "Too many visits!"`, and the IP guidance is a ten-minute automatic ban.

**The new test enforces the wrong invariant for batch endpoints.** From the [batch instructions](https://bybit-exchange.github.io/docs/v5/rate-limit#instructions-for-batch-endpoints):

> the counting consumption will be consumed according to the actual number of orders. (Number of consumption = number of requests * number of orders included in a single request) [...] if a batch order request is made once and contains 5 orders, then the request limit will consume 5

`PlaceBatchOrder` accepts 1-10 orders and charges [`createBatchOrderEPL`](https://github.com/thrasher-corp/gocryptotrader/blob/9f10dee97febd98fab243f6733998014edcb88de/exchanges/bybit/bybit.go#L639) once. With [ratelimit.go:91-93](https://github.com/thrasher-corp/gocryptotrader/blob/9f10dee97febd98fab243f6733998014edcb88de/exchanges/bybit/ratelimit.go#L91-L93) pinned at 1, a ten-order batch consumes 1 unit locally and 10 at Bybit — and [the assertion at ratelimit_test.go:43](https://github.com/thrasher-corp/gocryptotrader/blob/9f10dee97febd98fab243f6733998014edcb88de/exchanges/bybit/ratelimit_test.go#L43) now locks that in for all three batch EPLs.

Weight is exactly the mechanism for this, so the batch endpoints are precisely where it shouldn't be 1. A static weight can't express "N per request" either, so either they carry the worst case of 10, or the test carves them out rather than asserting blanket equality.

**Out of scope but worth an issue.** [`defaultEPL`](https://github.com/thrasher-corp/gocryptotrader/blob/9f10dee97febd98fab243f6733998014edcb88de/exchanges/bybit/ratelimit.go#L83) is `(5s, 600, 1)` which is about 120/s, and that models the **IP** limit rather than a per-UID endpoint limit — yet it is the fallback for roughly 36 call sites documented anywhere from 1/s to 50/s. That is untouched by this PR and predates it, so I would keep it out of scope here, but it is a bigger over-rate surface than anything above.

Fourteen other endpoints are configured *slower* than documented — `/v5/order/realtime`, `/v5/order/history`, `/v5/execution/list`, `/v5/position/list`, `/v5/position/closed-pnl` and `/v5/account/wallet-balance` are all at 10/s against a documented 50/s. Harmless, just throughput left on the table, and fine as a follow-up.

Everything else is green: lint clean, `exchanges/bybit` and `exchanges/request` both pass with `-race`, and the red CI jobs are the unrelated okx failure.

Happy to re-approve once the three rates are corrected and the batch endpoints are handled.

thrasher- (approved):
Heimdall: Approving — withdrawing my changes request.

All three rates now match the documented limits exactly: `getAccountFeeEPL` 5/s, `userCreateSubMemberEPL` and `userCreateSubAPIKeyEPL` 1/s. I re-ran the full mapping against Bybit's live table and no PR-scoped endpoint exceeds its documented limit.

The batch carve-out works as intended — mutating a batch weight no longer trips the test while a non-batch entry still does. Worth noting it removes the enforcement rather than the exposure: the three batch entries are still weight 1, so a ten-order batch consumes 1 unit locally against 10 at Bybit. Fine as an interim since a static weight can't express "N per request" either way, but a `TODO` on the entries themselves would help — the reasoning currently lives in the test while the misleading numbers live in the table.

Good call on the `defaultEPL` TODO too; that one's worth its own issue given it's the fallback for ~36 call sites documented anywhere from 1/s to 50/s.

Lint is green and bybit passes with `-race`. The red jobs are `TestFetchOrderbook` in gateio and a codecov threshold — neither yours.

exchanges/request/limit.go
```diff
@@ -82,6 +82,14 @@ func GetRateLimiterWithWeight(l *rate.Limiter, weight Weight) *RateLimiterWithWe
 	return &RateLimiterWithWeight{limiter: l, weight: weight}
 }
 
+// Weight returns the number of reservations consumed by each request.
+func (r *RateLimiterWithWeight) Weight() Weight {
+	if r == nil {
```
gloriousCode, line 87: I am shocked you don't want it to panic

gloriousCode (approved):
ch ACK

PR #2350, opened 2026-08-26
test: repair failing Coinbase and OKX live tests

roberttidball:
Two live tests currently fail on `master`, which turns CI red for every PR branched from it. This PR fixes only those two tests and touches nothing else, so it should be quick to review and unblocks other open PRs.

### `TestGetAllProducts` (Coinbase) — deterministic failure

The test requests `limit=30000` and Coinbase now rejects anything above 1000:

```
--- FAIL: TestGetAllProducts
    Coinbase unsuccessful HTTP status code: 400
    {"error":"INVALID_ARGUMENT","error_details":"invalid limit: limit must be <= 1000, got 30000"}
```

Every production caller passes `0`, so the cap only ever affected the test. It now passes `1000`, which matches the max page size already used elsewhere in the package by `GetProductBookV3` and `iterativeGetAllOrders`.

Reproducible on a clean checkout of master:

```
FAIL	github.com/thrasher-corp/gocryptotrader/exchanges/coinbase	4.568s
```

### `TestGetOpenInterestAndVolumeStrike` (OKX) — intermittent failure

The test selects the first instrument with a non-zero expiry time. Depending on what the endpoint returns at that moment, that can be an already-expired or not-yet-listed option, so the subsequent call fails. It is genuinely intermittent — it passed for me locally while failing in CI on the same day.

It now requires an instrument that is `live`, already listed, and whose expiry is still in the future.

### Verification

Both tests pass against the live endpoints with these changes:

```
--- PASS: TestGetOpenInterestAndVolumeStrike (0.66s)
ok  	github.com/thrasher-corp/gocryptotrader/exchanges/okx	4.878s
ok  	github.com/thrasher-corp/gocryptotrader/exchanges/coinbase	4.611s
```

`go vet` is clean on both packages. The Coinbase change is also present in #2294; if you would prefer to take it there instead, I am happy to drop it from this PR.

### DIFF
--- a/exchanges/coinbase/coinbase_test.go
+++ b/exchanges/coinbase/coinbase_test.go
@@ -708,7 +708,7 @@ func TestGetHistoricKlines(t *testing.T) {
 func TestGetAllProducts(t *testing.T) {
 	t.Parallel()
 	testPairs := []string{testPairFiat.String(), "ETH-USD"}
-	resp, err := e.GetAllProducts(t.Context(), 30000, 1, "SPOT", "PERPETUAL", "STATUS_ALL", "PRODUCTS_SORT_ORDER_UNDEFINED", testPairs, true, true, false)
+	resp, err := e.GetAllProducts(t.Context(), 1000, 1, "SPOT", "PERPETUAL", "STATUS_ALL", "PRODUCTS_SORT_ORDER_UNDEFINED", testPairs, true, true, false)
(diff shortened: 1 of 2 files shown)


thrasher-:
@codex please review

thrasher-:
Closing as this will be done in https://github.com/thrasher-corp/gocryptotrader/pull/2294#issuecomment-5420365537

PR #2355, opened 2026-08-31
engine: fall back across assets when validating API credentials

shazbert:
# PR Description

Try other enabled account types before disabling authentication globally when an asset-specific account is unavailable.

Problem arose in Gateio when its checking an asset which hadn't been loaded with funds and errored. 

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

### DIFF
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"net"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -937,7 +938,8 @@ func validateAPICredentials(ctx context.Context, exchangeName string, enabledAss
 	for _, a := range assets {
 		if err := validate(ctx, a); err != nil {
 			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
-			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
+			// A transport failure means the venue is unreachable, so no other account can answer either.
+			if _, ok := errors.AsType[net.Error](err); ok || errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
 				break
 			}
 			continue
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -893,7 +893,7 @@ func (bot *Engine) LoadExchange(name string) error {
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
 		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
 		if err := validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials); err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed; disabling authenticated support", b.Name)
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed, disabling authenticated support: %v", b.Name, err)
 			b.API.AuthenticatedSupport = false
 			b.API.AuthenticatedWebsocketSupport = false
 			if b.Websocket != nil {
@@ -906,10 +906,6 @@ func (bot *Engine) LoadExchange(name string) error {
 }
 
 func validateAPICredentials(ctx context.Context, exchangeName string, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
-	if len(enabledAssets) == 0 {
-		return fmt.Errorf("%s: %w", exchangeName, asset.ErrNotEnabled)
-	}
-
 	// Spot covers the widest set of GCT functionality, followed by futures;
 	// other account types are fallbacks, and the first successful validation wins.
 	assets := make(asset.Items, 0, len(enabledAssets))
@@ -925,6 +921,9 @@ func validateAPICredentials(ctx context.Context, exchangeName string, enabledAss
 		if a != asset.Spot && !a.IsFutures() {
 			assets = append(assets, a)
 		}
+	}
+	if len(assets) == 0 {
+		return fmt.Errorf("%s: %w", exchangeName, asset.ErrNotEnabled)
 	}
 
 	var errs error
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -892,24 +892,8 @@ func (bot *Engine) LoadExchange(name string) error {
 	b := exch.GetBase()
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
 		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
-		var preferredAsset asset.Item
-		switch {
-		case enabledAssets.Contains(asset.Spot): // prioritise validating credentials with spot due to wide usage across GCT
-			preferredAsset = asset.Spot
-		default:
-			for _, a := range enabledAssets { // second priority to futures if spot isn't available
-				if a.IsFutures() {
-					preferredAsset = a
-					break
-				}
-			}
-			if preferredAsset == 0 {
-				preferredAsset = enabledAssets[0] // last resort pick first available
-			}
-		}
-
-		if err := exch.ValidateAPICredentials(ctx, preferredAsset); err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v for %s", b.Name, err, preferredAsset)
+		if err := validateAPICredentials(ctx, enabledAssets, exch.ValidateAPICredentials); err != nil {
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", b.Name, err)
 			b.API.AuthenticatedSupport = false
 			b.API.AuthenticatedWebsocketSupport = false
 			if b.Websocket != nil {
@@ -919,6 +903,33 @@ func (bot *Engine) LoadExchange(name string) error {
 	}
 
 	return exchange.Bootstrap(ctx, exch)
+}
+
+func validateAPICredentials(ctx context.Context, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
+	assets := make(asset.Items, 0, len(enabledAssets))
+	if enabledAssets.Contains(asset.Spot) {
+		assets = append(assets, asset.Spot)
+	}
+	for _, a := range enabledAssets {
+		if a != asset.Spot && a.IsFutures() {
+			assets = append(assets, a)
+		}
+	}
+	for _, a := range enabledAssets {
+		if a != asset.Spot && !a.IsFutures() {
+			assets = append(assets, a)
+		}
+	}
+
+	var errs error
+	for _, a := range assets {
+		if err := validate(ctx, a); err != nil {
+			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
+			continue
+		}
+		return nil
+	}
+	return errs
 }
 
 func (bot *Engine) dryRunParamInteraction(param string) {
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -892,9 +892,9 @@ func (bot *Engine) LoadExchange(name string) error {
 	b := exch.GetBase()
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
 		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
-		// Credentials that are unusable fail identically for every account type,
-		// so establish that once rather than once per asset.
-		_, err := b.GetCredentials(ctx)
+		// Structurally unusable credentials fail identically for every account
+		// type, so establish that once rather than once per asset.
+		_, err := exch.GetCredentials(ctx)
 		if err == nil {
 			err = validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials)
 		}
@@ -943,7 +943,7 @@ func validateAPICredentials(ctx context.Context, exchangeName string, enabledAss
 			continue
 		}
 		if errs != nil { // the caller only logs errs when every account fails, so surface the recovered ones here
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credentials validated on %s after earlier failures: %v", exchangeName, a, errs)
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Authenticated support retained on %s after earlier failures: %v", exchangeName, a, errs)
 		}
 		return nil
 	}
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -906,6 +906,10 @@ func (bot *Engine) LoadExchange(name string) error {
 }
 
 func validateAPICredentials(ctx context.Context, exchangeName string, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
+	if len(enabledAssets) == 0 {
+		return fmt.Errorf("%s: %w", exchangeName, asset.ErrNotEnabled)
+	}
+
 	// Spot covers the widest set of GCT functionality, followed by futures;
 	// other account types are fallbacks, and the first successful validation wins.
 	assets := make(asset.Items, 0, len(enabledAssets))
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -893,7 +893,7 @@ func (bot *Engine) LoadExchange(name string) error {
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
 		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
 		if err := validateAPICredentials(ctx, enabledAssets, exch.ValidateAPICredentials); err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", b.Name, err)
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed; disabling authenticated support", b.Name)
 			b.API.AuthenticatedSupport = false
 			b.API.AuthenticatedWebsocketSupport = false
 			if b.Websocket != nil {
@@ -906,6 +906,8 @@ func (bot *Engine) LoadExchange(name string) error {
 }
 
 func validateAPICredentials(ctx context.Context, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
+	// Spot covers the widest set of GCT functionality, followed by futures;
+	// other account types are fallbacks, and the first successful validation wins.
 	assets := make(asset.Items, 0, len(enabledAssets))
 	if enabledAssets.Contains(asset.Spot) {
 		assets = append(assets, asset.Spot)
@@ -924,7 +926,12 @@ func validateAPICredentials(ctx context.Context, enabledAssets asset.Items, vali
 	var errs error
 	for _, a := range assets {
 		if err := validate(ctx, a); err != nil {
-			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
+			assetErr := fmt.Errorf("%s: %w", a, err)
+			gctlog.Warnf(gctlog.ExchangeSys, "Error validating credentials: %v", assetErr)
+			errs = common.AppendError(errs, assetErr)
+			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
+				break
+			}
 			continue
 		}
 		return nil
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -892,7 +892,7 @@ func (bot *Engine) LoadExchange(name string) error {
 	b := exch.GetBase()
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
 		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
-		if err := validateAPICredentials(ctx, enabledAssets, exch.ValidateAPICredentials); err != nil {
+		if err := validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials); err != nil {
 			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed; disabling authenticated support", b.Name)
 			b.API.AuthenticatedSupport = false
 			b.API.AuthenticatedWebsocketSupport = false
@@ -905,7 +905,7 @@ func (bot *Engine) LoadExchange(name string) error {
 	return exchange.Bootstrap(ctx, exch)
 }
 
-func validateAPICredentials(ctx context.Context, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
+func validateAPICredentials(ctx context.Context, exchangeName string, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
 	// Spot covers the widest set of GCT functionality, followed by futures;
 	// other account types are fallbacks, and the first successful validation wins.
 	assets := make(asset.Items, 0, len(enabledAssets))
@@ -927,7 +927,7 @@ func validateAPICredentials(ctx context.Context, enabledAssets asset.Items, vali
 	for _, a := range assets {
 		if err := validate(ctx, a); err != nil {
 			assetErr := fmt.Errorf("%s: %w", a, err)
-			gctlog.Warnf(gctlog.ExchangeSys, "Error validating credentials: %v", assetErr)
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", exchangeName, assetErr)
 			errs = common.AppendError(errs, assetErr)
 			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
 				break
--- a/engine/engine.go
+++ b/engine/engine.go
@@ -892,7 +892,13 @@ func (bot *Engine) LoadExchange(name string) error {
 	b := exch.GetBase()
 	if b.API.AuthenticatedSupport || b.API.AuthenticatedWebsocketSupport {
 		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
-		if err := validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials); err != nil {
+		// Credentials that are unusable fail identically for every account type,
+		// so establish that once rather than once per asset.
+		_, err := b.GetCredentials(ctx)
+		if err == nil {
+			err = validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials)
+		}
+		if err != nil {
 			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed, disabling authenticated support: %v", b.Name, err)
 			b.API.AuthenticatedSupport = false
 			b.API.AuthenticatedWebsocketSupport = false
@@ -908,6 +914,7 @@ func (bot *Engine) LoadExchange(name string) error {
 func validateAPICredentials(ctx context.Context, exchangeName string, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
 	// Spot covers the widest set of GCT functionality, followed by futures;
 	// other account types are fallbacks, and the first successful validation wins.
+	// This assumes validation is asset-specific; global validators repeat the same request.
 	assets := make(asset.Items, 0, len(enabledAssets))
 	if enabledAssets.Contains(asset.Spot) {
 		assets = append(assets, asset.Spot)
@@ -929,13 +936,14 @@ func validateAPICredentials(ctx context.Context, exchangeName string, enabledAss
 	var errs error
 	for _, a := range assets {
 		if err := validate(ctx, a); err != nil {
-			assetErr := fmt.Errorf("%s: %w", a, err)
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", exchangeName, assetErr)
-			errs = common.AppendError(errs, assetErr)
+			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
 			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
 				break
 			}
 			continue
+		}
+		if errs != nil { // the caller only logs errs when every account fails, so surface the recovered ones here
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credentials validated on %s after earlier failures: %v", exchangeName, a, errs)
 		}
 		return nil
 	}
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -451,6 +451,7 @@ func TestValidateAPICredentials(t *testing.T) {
 			return nil
 		})
 		require.ErrorIs(t, err, asset.ErrNotEnabled, "validation must fail when no assets are enabled")
+		assert.ErrorContains(t, err, testExchange, "error should name the exchange it came from")
 		assert.False(t, called, "validation should not run without an enabled asset")
 	})
 
@@ -498,6 +499,9 @@ func TestValidateAPICredentials(t *testing.T) {
 		asset.Margin,
 		asset.Index,
 	}, seen, "assets should be tried spot first, then futures, then the rest")
+	for _, a := range seen {
+		assert.ErrorContainsf(t, err, a.String()+": "+errDeliveryAccountMissing.Error(), "aggregated error should name %s and why it failed", a)
+	}
 
 	for _, assetIndependentErr := range []error{
 		exchange.ErrCredentialsAreEmpty,
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -464,6 +464,41 @@ func TestValidateAPICredentials(t *testing.T) {
 		return errDeliveryAccountMissing
 	})
 	require.ErrorIs(t, err, errDeliveryAccountMissing, "validation must return an error when every enabled account fails")
+
+	var seen asset.Items
+	err = validateAPICredentials(t.Context(), asset.Items{
+		asset.Options,
+		asset.USDTMarginedFutures,
+		asset.Margin,
+		asset.Spot,
+		asset.DeliveryFutures,
+		asset.Index,
+	}, func(_ context.Context, a asset.Item) error {
+		seen = append(seen, a)
+		return errDeliveryAccountMissing
+	})
+	require.ErrorIs(t, err, errDeliveryAccountMissing, "mixed asset validation must return the aggregated error")
+	assert.Equal(t, asset.Items{
+		asset.Spot,
+		asset.USDTMarginedFutures,
+		asset.DeliveryFutures,
+		asset.Options,
+		asset.Margin,
+		asset.Index,
+	}, seen, "assets should be tried spot first, then futures, then the rest")
+
+	for _, assetIndependentErr := range []error{
+		exchange.ErrCredentialsAreEmpty,
+		exchange.ErrAuthenticationSupportNotEnabled,
+	} {
+		validated = nil
+		err = validateAPICredentials(t.Context(), asset.Items{asset.Spot, asset.USDTMarginedFutures}, func(_ context.Context, a asset.Item) error {
+			validated = append(validated, a)
+			return assetIndependentErr
+		})
+		require.ErrorIs(t, err, assetIndependentErr, "asset-independent credential failure must be returned")
+		assert.Equal(t, asset.Items{asset.Spot}, validated, "asset-independent credential failure should stop further validation")
+	}
 }
 
 func TestFlagSetWith(t *testing.T) {
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"context"
+	"errors"
 	"os"
 	"slices"
 	"strings"
@@ -12,6 +13,7 @@ import (
 	"github.com/stretchr/testify/require"
 	"github.com/thrasher-corp/gocryptotrader/config"
 	exchange "github.com/thrasher-corp/gocryptotrader/exchanges"
+	"github.com/thrasher-corp/gocryptotrader/exchanges/asset"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/bitfinex"
 	"github.com/thrasher-corp/gocryptotrader/exchanges/bitstamp"
 )
@@ -437,6 +439,31 @@ func TestDryRunParamInteraction(t *testing.T) {
 		!exchCfg.Verbose {
 		t.Error("dryrun should be true and verbose should be true")
 	}
+}
+
+func TestValidateAPICredentials(t *testing.T) {
+	t.Parallel()
+
+	errDeliveryAccountMissing := errors.New("delivery account missing")
+	var validated asset.Items
+	err := validateAPICredentials(t.Context(), asset.Items{
+		asset.DeliveryFutures,
+		asset.USDTMarginedFutures,
+	}, func(_ context.Context, a asset.Item) error {
+		validated = append(validated, a)
+		if a == asset.DeliveryFutures {
+			return errDeliveryAccountMissing
+		}
+		return nil
+	})
+	require.NoError(t, err, "validation must succeed when another enabled account is available")
+	assert.Equal(t, asset.Items{asset.DeliveryFutures, asset.USDTMarginedFutures}, validated,
+		"validation should try another enabled futures account after an account-specific failure")
+
+	err = validateAPICredentials(t.Context(), asset.Items{asset.DeliveryFutures}, func(context.Context, asset.Item) error {
+		return errDeliveryAccountMissing
+	})
+	require.ErrorIs(t, err, errDeliveryAccountMissing, "validation must return an error when every enabled account fails")
 }
 
 func TestFlagSetWith(t *testing.T) {
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -3,6 +3,8 @@ package engine
 import (
 	"context"
 	"errors"
+	"fmt"
+	"net/url"
 	"os"
 	"slices"
 	"strings"
@@ -503,17 +505,18 @@ func TestValidateAPICredentials(t *testing.T) {
 		assert.ErrorContainsf(t, err, a.String()+": "+errDeliveryAccountMissing.Error(), "aggregated error should name %s and why it failed", a)
 	}
 
-	for _, assetIndependentErr := range []error{
+	for _, venueWideErr := range []error{
 		exchange.ErrCredentialsAreEmpty,
 		exchange.ErrAuthenticationSupportNotEnabled,
+		&url.Error{Op: "Get", URL: "https://example.com/accounts", Err: context.DeadlineExceeded},
 	} {
 		validated = nil
 		err = validateAPICredentials(t.Context(), testExchange, asset.Items{asset.Spot, asset.USDTMarginedFutures}, func(_ context.Context, a asset.Item) error {
 			validated = append(validated, a)
-			return assetIndependentErr
+			return fmt.Errorf("%w, authenticated request failed", venueWideErr)
 		})
-		require.ErrorIs(t, err, assetIndependentErr, "asset-independent credential failure must be returned")
-		assert.Equal(t, asset.Items{asset.Spot}, validated, "asset-independent credential failure should stop further validation")
+		require.ErrorIs(t, err, venueWideErr, "venue-wide failure must be returned")
+		assert.Equal(t, asset.Items{asset.Spot}, validated, "venue-wide failure should stop further validation")
 	}
 }
 
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -481,6 +481,7 @@ func TestValidateAPICredentials(t *testing.T) {
 		asset.USDTMarginedFutures,
 		asset.Margin,
 		asset.Spot,
+		asset.PerpetualSwap,
 		asset.DeliveryFutures,
 		asset.Index,
 	}, func(_ context.Context, a asset.Item) error {
@@ -491,6 +492,7 @@ func TestValidateAPICredentials(t *testing.T) {
 	assert.Equal(t, asset.Items{
 		asset.Spot,
 		asset.USDTMarginedFutures,
+		asset.PerpetualSwap,
 		asset.DeliveryFutures,
 		asset.Options,
 		asset.Margin,
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -443,6 +443,16 @@ func TestDryRunParamInteraction(t *testing.T) {
 
 func TestValidateAPICredentials(t *testing.T) {
 	t.Parallel()
+	t.Run("no enabled assets", func(t *testing.T) {
+		t.Parallel()
+		called := false
+		err := validateAPICredentials(t.Context(), testExchange, nil, func(context.Context, asset.Item) error {
+			called = true
+			return nil
+		})
+		require.ErrorIs(t, err, asset.ErrNotEnabled, "validation must fail when no assets are enabled")
+		assert.False(t, called, "validation should not run without an enabled asset")
+	})
 
 	errDeliveryAccountMissing := errors.New("delivery account missing")
 	var validated asset.Items
--- a/engine/engine_test.go
+++ b/engine/engine_test.go
@@ -446,7 +446,7 @@ func TestValidateAPICredentials(t *testing.T) {
 
 	errDeliveryAccountMissing := errors.New("delivery account missing")
 	var validated asset.Items
-	err := validateAPICredentials(t.Context(), asset.Items{
+	err := validateAPICredentials(t.Context(), testExchange, asset.Items{
 		asset.DeliveryFutures,
 		asset.USDTMarginedFutures,
 	}, func(_ context.Context, a asset.Item) error {
@@ -460,13 +460,13 @@ func TestValidateAPICredentials(t *testing.T) {
 	assert.Equal(t, asset.Items{asset.DeliveryFutures, asset.USDTMarginedFutures}, validated,
 		"validation should try another enabled futures account after an account-specific failure")
 
-	err = validateAPICredentials(t.Context(), asset.Items{asset.DeliveryFutures}, func(context.Context, asset.Item) error {
+	err = validateAPICredentials(t.Context(), testExchange, asset.Items{asset.DeliveryFutures}, func(context.Context, asset.Item) error {
 		return errDeliveryAccountMissing
 	})
 	require.ErrorIs(t, err, errDeliveryAccountMissing, "validation must return an error when every enabled account fails")
 
 	var seen asset.Items
-	err = validateAPICredentials(t.Context(), asset.Items{
+	err = validateAPICredentials(t.Context(), testExchange, asset.Items{
 		asset.Options,
 		asset.USDTMarginedFutures,
 		asset.Margin,
@@ -492,7 +492,7 @@ func TestValidateAPICredentials(t *testing.T) {
 		exchange.ErrAuthenticationSupportNotEnabled,
 	} {
 		validated = nil
-		err = validateAPICredentials(t.Context(), asset.Items{asset.Spot, asset.USDTMarginedFutures}, func(_ context.Context, a asset.Item) error {
+		err = validateAPICredentials(t.Context(), testExchange, asset.Items{asset.Spot, asset.USDTMarginedFutures}, func(_ context.Context, a asset.Item) error {
 			validated = append(validated, a)
 			return assetIndependentErr
 		})


thrasher-:
@codex please review

engine/engine_test.go
```diff
@@ -439,6 +441,31 @@ func TestDryRunParamInteraction(t *testing.T) {
 	}
 }
 
+func TestValidateAPICredentials(t *testing.T) {
+	t.Parallel()
+
+	errDeliveryAccountMissing := errors.New("delivery account missing")
+	var validated asset.Items
+	err := validateAPICredentials(t.Context(), asset.Items{
```
thrasher-: Heimdall: **Should fix** — this test passes against three separate mutations of the ordering logic it covers.

**Fix:** add a mixed-asset case at [engine_test.go:449](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine_test.go#L449) so the tiering is actually asserted:

```go
in := asset.Items{asset.Options, asset.USDTMarginedFutures, asset.Margin, asset.Spot, asset.DeliveryFutures, asset.Index}
var seen asset.Items
_ = validateAPICredentials(t.Context(), in, func(_ context.Context, a asset.Item) error {
	seen = append(seen, a)
	return errDeliveryAccountMissing
})
require.Equal(t, asset.Items{
	asset.Spot,
	asset.USDTMarginedFutures, asset.DeliveryFutures,
	asset.Options, asset.Margin, asset.Index,
}, seen, "assets must be tried spot first, then futures, then the rest")
```

**Why:** both existing inputs are futures-only, so nothing constrains the tiering. Mutating [engine.go:909-921](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine.go#L909-L921) and re-running:

```
mutation                                  TestValidateAPICredentials   with the case above
drop the spot prioritisation block                 PASS                       FAIL
drop the non-futures tier                          PASS                       FAIL
assets := enabledAssets (no ordering)              PASS                       FAIL
```

The second row is the one that matters: dropping the non-futures tier means margin, cross-margin, options, index and binary are never validated at all — and for an exchange with neither spot nor futures enabled, `assets` comes out empty, the loop at [engine.go:925](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine.go#L925) never runs, and the function returns `nil` — auth left enabled having validated nothing. CI stays green through all three.

engine/engine.go
```diff
@@ -921,6 +905,33 @@ func (bot *Engine) LoadExchange(name string) error {
...
+		}
+	}
+	for _, a := range enabledAssets {
+		if a != asset.Spot && !a.IsFutures() {
+			assets = append(assets, a)
+		}
+	}
+
+	var errs error
+	for _, a := range assets {
```
thrasher-, line 938: Heimdall: **Should fix** — a failure preceding the first success is discarded entirely, and errors that can't be asset-specific are retried against every enabled asset.

**Fix:** at [engine.go:925](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine.go#L925), log each per-asset failure as it occurs so a broken account stays visible after a later asset succeeds; and break early on errors that aren't asset-scoped (`exchange.ErrCredentialsAreEmpty` and friends) instead of repeating them once per asset.

**Why:** two measured effects.

Silent swallow — spot fails, futures succeeds, the function returns `nil` and nothing is logged. This is the GateIO case in the description: the account that errored is now exactly the one the operator can no longer see. On [`4e73967`](https://github.com/thrasher-corp/gocryptotrader/commit/4e73967aea642f3d92dcd0b2eb00576f2075c4f2) that failure disabled auth and produced a warning — too aggressive, but visible.

Amplification — the real Binance wrapper against a local server returning `401 {"code":-2015,...}`, identical conditions both ways:

```
enabled assets: 4 (spot, margin, coinmarginedfutures, usdtmarginedfutures)
HTTP requests    -- 4e73967: 1     | 6c1ee4f: 4
warning length   -- 4e73967: 157   | 6c1ee4f: 690 chars
```

GateIO ships 7 enabled assets in `testdata/configtest.json`, so a revoked key there means 7 authenticated 401s at start-up and the venue's error body repeated 7x in one warning line.

engine/engine.go
```diff
@@ -921,6 +905,33 @@ func (bot *Engine) LoadExchange(name string) error {
 	return exchange.Bootstrap(ctx, exch)
 }
 
+func validateAPICredentials(ctx context.Context, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
```
thrasher-: Heimdall: **Nit** — the rationale for the priority order went with the old code and wasn't replaced.

**Fix:** a line on [engine.go:908](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine.go#L908) covering why spot leads, why futures is second, and that the first success wins.

**Why:** [`4e73967`](https://github.com/thrasher-corp/gocryptotrader/commit/4e73967aea642f3d92dcd0b2eb00576f2075c4f2) carried `// prioritise validating credentials with spot due to wide usage across GCT` and `// second priority to futures if spot isn't available`. Both were removed, the new function has no comments, and `git grep` finds that reasoning nowhere in the tree now. The ordering is the entire point of the function and the one thing a reader can't infer from the code.

thrasher- (approved):
Heimdall: Approving. The fall-back ordering is correct: fed all 20 of `asset.Supported()`, every asset is tried exactly once in the intended tiers (spot -> futures -> rest), the loop short-circuits on the first success, and every aggregated error stays `errors.Is`-matchable and names its asset. It also removes a latent panic — the `enabledAssets[0]` fallback on [`4e73967`](https://github.com/thrasher-corp/gocryptotrader/commit/4e73967aea642f3d92dcd0b2eb00576f2075c4f2) panics `index out of range [0] with length 0` on an empty list (unreachable today, `config.go:906-912` force-enables an asset first). `go test ./engine/ -race` passes, gofumpt is clean on both files, and `make misc_checks` passes in full.

Three comments inline, none of them conditions on this merge: coverage of the tiering at [engine_test.go:449](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine_test.go#L449), the silent swallow and request amplification at [engine.go:925](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine.go#L925), and the dropped rationale comments at [engine.go:908](https://github.com/thrasher-corp/gocryptotrader/blob/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc/engine/engine.go#L908). Evidence is on the lines.

CI red here is ambient, not this PR: failures are confined to `backtester/...` with `dial tcp: lookup api.binance.us on 127.0.0.53:53: no such host`, and master's own run on the identical base SHA is equally red ([33360888477](https://github.com/thrasher-corp/gocryptotrader/actions/runs/33360888477) vs [33364366464](https://github.com/thrasher-corp/gocryptotrader/actions/runs/33364366464)).

Take the three or leave them — nothing here blocks.

engine/engine.go
```diff
@@ -892,24 +892,8 @@ func (bot *Engine) LoadExchange(name string) error {
...
-			}
-			if preferredAsset == 0 {
-				preferredAsset = enabledAssets[0] // last resort pick first available
-			}
-		}
-
-		if err := exch.ValidateAPICredentials(ctx, preferredAsset); err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v for %s", b.Name, err, preferredAsset)
+		if err := validateAPICredentials(ctx, enabledAssets, exch.ValidateAPICredentials); err != nil {
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed; disabling authenticated support", b.Name)
```
thrasher-: Heimdall: **Should fix** — a credential warning can no longer be tied to the exchange that produced it.

**Fix:** pass the exchange name into `validateAPICredentials` and include it in the per-asset warning at [engine.go:930](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L930). Restoring `%v` on this line would also give the aggregated error somewhere to land — as written, the value built by `common.AppendError` is only ever nil-checked in production.

**Why:** the name and the error are now on separate lines, and in the case this PR exists to fix they never appear together at all.

| | exchange | error | asset |
|---|---|---|---|
| [`4e73967`](https://github.com/thrasher-corp/gocryptotrader/commit/4e73967aea642f3d92dcd0b2eb00576f2075c4f2) `engine.go:912` | yes | yes | yes |
| [engine.go:896](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L896) (this line) | yes | **no** | no |
| [engine.go:930](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L930) | **no** | yes | yes |

When spot fails and a later asset succeeds — the GateIO case in the description — `validateAPICredentials` returns nil, so this line never runs and only the unattributed one does. `SetupExchanges` starts a goroutine per exchange at [engine.go:1023](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L1023), so they interleave. Three exchanges, spot failing, futures succeeding:

```
[WARN] |  01/09/2026 15:06:46  | Error validating credentials: spot: spot account not funded
[WARN] |  01/09/2026 15:06:46  | Error validating credentials: spot: spot account not funded
[WARN] |  01/09/2026 15:06:46  | Error validating credentials: spot: spot account not funded
```

Structured output is no better — `loggerWorker` ([logger_multiwriter.go:41](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/log/logger_multiwriter.go#L41)) emits `sublogger` (always `EXCHANGE`) and `botname` (the bot instance), and this call site passes no fields:

```json
{"botname":"test-bot","message":"Error validating credentials: spot: spot account not funded","severity":"warn","sublogger":"EXCHANGE","timestamp":1788239219608}
```

The error text doesn't reliably carry the venue either: Binance spot with a subaccount configured returns a bare `common.ErrNotYetImplemented` at [binance_wrapper.go:565](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/exchanges/binance/binance_wrapper.go#L565), and futures then validates fine.

thrasher- (approved):
Heimdall: Approving again — all three from the last pass are in, and I've resolved those threads. The ordering test now catches all three mutations that escaped it on [`6c1ee4f`](https://github.com/thrasher-corp/gocryptotrader/commit/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc) — dropping the spot block, dropping the non-futures tier, and `assets := enabledAssets` — plus the new asset-independent break; all 20 of `asset.Supported()` are tried exactly once in tier order; per-asset failures are now logged at [engine.go:930](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L930) so a failure before a later success no longer disappears; and the priority rationale is back at [engine.go:909-910](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L909-L910).

`go test ./engine/ -race` passes in 24.7s and `make misc_checks` is clean. Lint is unchanged by this PR — the three hits inside the changed files ([engine.go:476](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L476), [engine.go:663](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L663), [engine_test.go:651](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine_test.go#L651)) reproduce identically on [`4e73967`](https://github.com/thrasher-corp/gocryptotrader/commit/4e73967aea642f3d92dcd0b2eb00576f2075c4f2).

One comment, with the evidence on the line: the new per-asset warning can't be tied to the exchange that produced it — [engine.go:896](https://github.com/thrasher-corp/gocryptotrader/blob/ca349a292ad2af2dd1060ced2b2cee04821456b0/engine/engine.go#L896). Not a condition on the merge.

Take it or leave it — nothing here blocks.

thrasher- (approved):
Heimdall: Approving on [`1aec341b`](https://github.com/thrasher-corp/gocryptotrader/commit/1aec341bb2c0f3953a47e1d5722b305ccc37f4b8). The attribution fix is in — [engine.go:930](https://github.com/thrasher-corp/gocryptotrader/blob/1aec341bb2c0f3953a47e1d5722b305ccc37f4b8/engine/engine.go#L930) now carries the exchange name, so the case that motivated it — spot fails, a later asset succeeds, and the outer warning at [engine.go:896](https://github.com/thrasher-corp/gocryptotrader/blob/1aec341bb2c0f3953a47e1d5722b305ccc37f4b8/engine/engine.go#L896) never fires — is attributable. Three exchanges validating concurrently:

```
[WARN] | Okx: Error validating credentials: spot: spot account not funded
[WARN] | Binance: Error validating credentials: spot: spot account not funded
[WARN] | GateIO: Error validating credentials: spot: spot account not funded
```

All five call sites updated, and the ordering coverage survived the signature change — dropping the spot block, dropping the non-futures tier, `assets := enabledAssets` and removing the asset-independent break all still fail the test. `go test ./engine/ -race` passes in 24.1s, `make misc_checks` is clean, and lint is unchanged from [`4e73967`](https://github.com/thrasher-corp/gocryptotrader/commit/4e73967aea642f3d92dcd0b2eb00576f2075c4f2).

`codecov/project` red is ambient, not this PR: `codecov/patch` passes at 92.86%, [engine.go](https://github.com/thrasher-corp/gocryptotrader/blob/1aec341bb2c0f3953a47e1d5722b305ccc37f4b8/engine/engine.go) itself is +2.34%, and the -0.01% comes from 22 files this PR doesn't touch. Against the same base, [`ca349a2`](https://github.com/thrasher-corp/gocryptotrader/commit/ca349a292ad2af2dd1060ced2b2cee04821456b0) scored +0.02% and [`1aec341b`](https://github.com/thrasher-corp/gocryptotrader/commit/1aec341bb2c0f3953a47e1d5722b305ccc37f4b8) -0.01% — two commits differing only by threading a string and adding a `%s` verb, so that swing is run-to-run variance rather than a real delta. `codecov.yml` sets no `project.threshold`, so zero tolerance turns noise red; the base report is also 4 commits behind master.

Nothing outstanding — resolving the last thread.

engine/engine.go
```diff
@@ -921,6 +905,40 @@ func (bot *Engine) LoadExchange(name string) error {
...
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", exchangeName, assetErr)
+			errs = common.AppendError(errs, assetErr)
+			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
+				break
+			}
+			continue
+		}
+		return nil
+	}
+	return errs
```
gloriousCode, line 952: Found a small buggeroo - reachable through this RPC sequence:

1. Call `SetExchangeAsset` with `enable=false` for the final enabled asset. That RPC updates both the live pair manager and the stored exchange config at [`rpcserver.go:2836-2840`](https://github.com/thrasher-corp/gocryptotrader/blob/1aec341bb2c0f3953a47e1d5722b305ccc37f4b8/engine/rpcserver.go#L2836-L2840).
2. Call `DisableExchange`.
3. Call `EnableExchange`. `LoadExchange` uses that stored config without rerunning the initial config validation, so `GetAssetTypes(true)` supplies an empty slice here.

This implementation skips the loop and returns nil because `errs` was never populated. `LoadExchange` therefore keeps REST and WebSocket authenticated-support flags enabled even though no credential validation ran. 

```suggestion
if len(assets) == 0 {
	return fmt.Errorf("%s: %w", exchangeName, asset.ErrNotEnabled)
}
return errs
```

Add the empty-input regression as a `no enabled assets` subtest inside the existing [`TestValidateAPICredentials`](https://github.com/thrasher-corp/gocryptotrader/blob/1aec341bb2c0f3953a47e1d5722b305ccc37f4b8/engine/engine_test.go#L444) in `engine/engine_test.go`:

```go
t.Run("no enabled assets", func(t *testing.T) {
	t.Parallel()
	called := false
	err := validateAPICredentials(t.Context(), testExchange, nil, func(context.Context, asset.Item) error {
		called = true
		return nil
	})
	require.ErrorIs(t, err, asset.ErrNotEnabled, "validation must fail when no assets are enabled")
	assert.False(t, called, "validation should not run without an enabled asset")
})
```

I would call this not very high priority

gloriousCode (changes requested):
There is a bug that's pretty simple to fix. I think the chances of this ever being an issue is low, but given the easiness to fix, I'm requesting it. I'll fold pretty quickly though. DONT TEST ME

engine/engine.go
```diff
@@ -921,6 +905,44 @@ func (bot *Engine) LoadExchange(name string) error {
...
+			assets = append(assets, a)
+		}
+	}
+	for _, a := range enabledAssets {
+		if a != asset.Spot && !a.IsFutures() {
+			assets = append(assets, a)
+		}
+	}
+
+	var errs error
```
thrasher-: Heimdall: **Minor** — the guard is placed before the tiering, so it proves the input was non-empty rather than that anything will actually be validated.

**Fix:** move it after the tier build:

```suggestion
	// Spot covers the widest set of GCT functionality, followed by futures;
	// other account types are fallbacks, and the first successful validation wins.
	assets := make(asset.Items, 0, len(enabledAssets))
	if enabledAssets.Contains(asset.Spot) {
		assets = append(assets, asset.Spot)
	}
	for _, a := range enabledAssets {
		if a != asset.Spot && a.IsFutures() {
			assets = append(assets, a)
		}
	}
	for _, a := range enabledAssets {
		if a != asset.Spot && !a.IsFutures() {
			assets = append(assets, a)
		}
	}

	if len(assets) == 0 {
		return fmt.Errorf("%s: %w", exchangeName, asset.ErrNotEnabled)
	}

	var errs error
```

**Why:** the two placements are equivalent today — every element of `enabledAssets` is spot, or futures, or neither, so `len(assets) == 0` exactly when `len(enabledAssets) == 0`. They stop being equivalent the moment the partition stops being total, and the failure mode is silent. Narrowing the `IsFutures()` check at [engine.go:920](https://github.com/thrasher-corp/gocryptotrader/blob/58136939f4e0dadd40edec6dad742c6594e12d54/engine/engine.go#L920) to a two-value check leaves `asset.PerpetualSwap` in neither tier:

```
input {asset.PerpetualSwap}          validate calls   returned
  58136939                                 1          perpetualswap: boom
  narrowed IsFutures()                     0          <nil>          <- auth stays on, nothing validated
  narrowed + this suggestion               0          probe: asset type not enabled
```

`TestValidateAPICredentials` passes in the middle row. With the guard after the build it fails closed instead, and the suggestion above is a straight swap: `go test ./engine/ -race` passes in 44.6s and no test change is required. It adds no formatting delta either — `gofumpt -d engine/engine.go` reports the same three hunks before and after, all of them pre-existing and none in this PR's diff.

engine/engine_test.go
```diff
@@ -439,6 +441,76 @@ func TestDryRunParamInteraction(t *testing.T) {
...
+	})
+	require.ErrorIs(t, err, errDeliveryAccountMissing, "mixed asset validation must return the aggregated error")
+	assert.Equal(t, asset.Items{
+		asset.Spot,
+		asset.USDTMarginedFutures,
+		asset.DeliveryFutures,
+		asset.Options,
+		asset.Margin,
+		asset.Index,
+	}, seen, "assets should be tried spot first, then futures, then the rest")
```
thrasher-, line 503: Heimdall: **Minor** — every futures value in this case is `DeliveryFutures` or `USDTMarginedFutures`, so the test pins those two rather than the futures tier.

**Fix:** add a third futures class to the input and to the expected order:

```suggestion
	err = validateAPICredentials(t.Context(), testExchange, asset.Items{
		asset.Options,
		asset.USDTMarginedFutures,
		asset.Margin,
		asset.Spot,
		asset.PerpetualSwap,
		asset.DeliveryFutures,
		asset.Index,
	}, func(_ context.Context, a asset.Item) error {
		seen = append(seen, a)
		return errDeliveryAccountMissing
	})
	require.ErrorIs(t, err, errDeliveryAccountMissing, "mixed asset validation must return the aggregated error")
	assert.Equal(t, asset.Items{
		asset.Spot,
		asset.USDTMarginedFutures,
		asset.PerpetualSwap,
		asset.DeliveryFutures,
		asset.Options,
		asset.Margin,
		asset.Index,
	}, seen, "assets should be tried spot first, then futures, then the rest")
```

**Why:** replacing `a.IsFutures()` at [engine.go:920](https://github.com/thrasher-corp/gocryptotrader/blob/58136939f4e0dadd40edec6dad742c6594e12d54/engine/engine.go#L920) with `a == asset.DeliveryFutures || a == asset.USDTMarginedFutures` is a real behavioural change — it drops ten of the twelve `IsFutures()` classifications out of both tiers — and the suite stays green:

```
mutant: IsFutures() -> two-value check
  TestValidateAPICredentials on 58136939        ok
  TestValidateAPICredentials with this change   FAIL  assets should be tried spot first, then futures, then the rest
```

This is the same concern as the guard-placement comment, from the test side; either one alone closes the hole. Verified: passes on the current head, fails on the mutant, `go test ./engine/ -race` green.

engine/engine.go
```diff
@@ -892,24 +892,8 @@ func (bot *Engine) LoadExchange(name string) error {
...
-			}
-			if preferredAsset == 0 {
-				preferredAsset = enabledAssets[0] // last resort pick first available
-			}
-		}
-
-		if err := exch.ValidateAPICredentials(ctx, preferredAsset); err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v for %s", b.Name, err, preferredAsset)
+		if err := validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials); err != nil {
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed; disabling authenticated support", b.Name)
```
thrasher-: Heimdall: **Minor** — on the path this round added, the reason never reaches the log.

**Fix:** `gctlog.Warnf(gctlog.ExchangeSys, "%s: Credential validation failed, disabling authenticated support: %v", b.Name, err)`.

**Why:** every other failure has already been logged per-asset at [engine.go:934](https://github.com/thrasher-corp/gocryptotrader/blob/58136939f4e0dadd40edec6dad742c6594e12d54/engine/engine.go#L934), so this line adds nothing there. The empty-asset return at [engine.go:910](https://github.com/thrasher-corp/gocryptotrader/blob/58136939f4e0dadd40edec6dad742c6594e12d54/engine/engine.go#L910) is the one case that never goes through that loop, and it is exactly the case where the error is the whole diagnosis. Running the RPC sequence on this head, the complete output is:

```
[INFO] | Bitstamp exchange unloaded successfully.
[WARN] | Bitstamp rate limiting has been turned off
[WARN] | Bitstamp: Credential validation failed; disabling authenticated support
```

`asset type not enabled` is built, returned, and discarded. Same line I raised on [`ca349a2`](https://github.com/thrasher-corp/gocryptotrader/commit/ca349a292ad2af2dd1060ced2b2cee04821456b0) — the exchange-name half of that got fixed and this half didn't; I'm raising it once more only because the new guard gave it a concrete case.

thrasher- (approved):
Heimdall: Approving on [`58136939`](https://github.com/thrasher-corp/gocryptotrader/commit/58136939f4e0dadd40edec6dad742c6594e12d54). The empty-asset guard is the whole of this round and it earns its place — I ran gloriousCode's exact RPC sequence against both trees, and on master it doesn't misbehave, it takes the bot down:

```
SetExchangeAsset(enable=false) x every enabled asset -> DisableExchange -> EnableExchange

master   panic: runtime error: index out of range [0] with length 0
           engine.(*Engine).LoadExchange        engine/engine.go:907
           engine.(*RPCServer).EnableExchange   engine/rpcserver.go:366
58136939 [WARN] Bitstamp: Credential validation failed; disabling authenticated support
         AuthenticatedSupport=false, AuthenticatedWebsocketSupport=false
         EnableExchange -> failed to update tradable pairs: currency pairs is empty
```

Neither [`SetExchangeAsset`](https://github.com/thrasher-corp/gocryptotrader/blob/58136939f4e0dadd40edec6dad742c6594e12d54/engine/rpcserver.go#L2836) nor [`SetAssetEnabled`](https://github.com/thrasher-corp/gocryptotrader/blob/58136939f4e0dadd40edec6dad742c6594e12d54/currency/manager.go#L403) guards the last asset, and `EnableExchange` re-enters `LoadExchange` without the cold-start repair at [`config.go:899`](https://github.com/thrasher-corp/gocryptotrader/blob/58136939f4e0dadd40edec6dad742c6594e12d54/config/config.go#L899). To be clear about scope, it's the panic that's fixed and not the invariant: `EnableExchange` still fails, just in `Bootstrap` rather than by taking the process with it.

The rest holds on this head: all 20 of `asset.Supported()` tried exactly once in tier order, first success short-circuits, `validateAPICredentials` at 100% statement coverage, `go test ./engine/ -race` green in 42.6s, `make misc_checks` clean, and `golangci-lint run ./engine/...` identical to master at 30 issues either side.

Three minor comments inline. Two are one concern from opposite sides — the tier partition is total today and nothing enforces that it stays total; the third is a log line.

CI red is ambient: all seven failing jobs fail on the same live-exchange test, `okx.TestGetOpenInterestAndVolumeStrike` (`51021 "Contract to be listed"`), the only `--- FAIL` in [the run](https://github.com/thrasher-corp/gocryptotrader/actions/runs/33565742453), and this PR touches only the two engine files. Master is no better — red on the PR's own base [`4e73967a`](https://github.com/thrasher-corp/gocryptotrader/commit/4e73967aea642f3d92dcd0b2eb00576f2075c4f2) ([33360888477](https://github.com/thrasher-corp/gocryptotrader/actions/runs/33360888477)) and again on `8606b772` ([33468559590](https://github.com/thrasher-corp/gocryptotrader/actions/runs/33468559590)).

engine/engine.go
```diff
@@ -892,24 +892,8 @@ func (bot *Engine) LoadExchange(name string) error {
...
-				}
-			}
-			if preferredAsset == 0 {
-				preferredAsset = enabledAssets[0] // last resort pick first available
-			}
-		}
-
-		if err := exch.ValidateAPICredentials(ctx, preferredAsset); err != nil {
-			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v for %s", b.Name, err, preferredAsset)
+		if err := validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials); err != nil {
```
thrasher-: Heimdall: **Should fix** — the early break catches the credential error that almost never fires and misses the ones the shipped config actually produces.

**Fix:** establish once that the credentials are usable at all, before entering the per-asset loop:

```suggestion
		enabledAssets := b.CurrencyPairs.GetAssetTypes(true)
		// Credentials that are unusable fail identically for every account type,
		// so establish that once rather than once per asset.
		_, err := b.GetCredentials(ctx)
		if err == nil {
			err = validateAPICredentials(ctx, b.Name, enabledAssets, exch.ValidateAPICredentials)
		}
		if err != nil {
```

**Why:** `VerifyAPICredentials` produces six local, asset-independent failures — [`credentials.go:113`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/credentials.go#L113), [`117`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/credentials.go#L117), [`122`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/credentials.go#L122), [`127`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/credentials.go#L127), [`132`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/credentials.go#L132), [`138`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/credentials.go#L138) — and the break at [`engine.go:935`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/engine/engine.go#L935) recognises one of them. `ErrCredentialsAreEmpty` needs *every* field blank. The shipped `"Key"`/`"Secret"` are `config.DefaultAPIKey`/`DefaultAPISecret`, so `creds.IsEmpty()` is false and you get the unexported `errRequiresAPIKey` — not in the set. Same for a key pasted without its secret, or Bitstamp's missing `clientID`.

Through real `LoadExchange` with `authenticatedSupport: true` over the credentials `testdata/configtest.json` ships:

```
                              master 9ced764   2a0a2ed   this suggestion
GateIO   validate calls             1             7             0
         [WARN] lines               1             8             1
Deribit  HTTP requests              1             5             0
```

Deribit is the one that costs traffic rather than cycles: `UpdateAccountBalances` calls public `get_currencies` at [`deribit_wrapper.go:339`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/deribit/deribit_wrapper.go#L339) before the credential lookup, so each of its five enabled assets is a round trip that was always going to fail locally.

One behaviour change to know about, and I think it's the right one. [`btcmarkets_wrapper.go:843`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/btcmarkets/btcmarkets_wrapper.go#L843) is the only validator of 24 that swallows base-local credential errors, so a placeholder key with a base64-valid secret leaves BTC Markets authenticated support **on** today, with credentials that can never sign a request. Measured on this head:

```
b.GetCredentials(ctx)             -> BTC Markets requires API key but default/empty one set
ValidateAPICredentials(ctx, spot) -> <nil>
```

The suggestion turns that into auth disabled with the reason logged. Keep the existing break as defence in depth — it costs nothing and still guards future callers.

Verified: applies clean, `go test ./engine/ -race -count 1` passes in 47.4s with **no test change required**, `make misc_checks` clean, gofumpt delta unchanged.

engine/engine.go
```diff
@@ -921,6 +905,43 @@ func (bot *Engine) LoadExchange(name string) error {
 	return exchange.Bootstrap(ctx, exch)
 }
 
+func validateAPICredentials(ctx context.Context, exchangeName string, enabledAssets asset.Items, validate func(context.Context, asset.Item) error) error {
```
thrasher-, line 915: Heimdall: **Should fix** — the helper assumes each enabled asset is a distinct validation target. For seven exchanges the retried request is byte-identical, so the fallback can only add load.

**Ask:** either bound the loop, or say in the comment at [`engine.go:909-910`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/engine/engine.go#L909-L910) that the helper assumes per-asset validation and that exchanges validating globally will repeat the same call. I'd rather not prescribe which — it depends on whether you'd sooner fix the validators.

**Why:** BTSE, Bitfinex, BitMEX, Coinbase, Deribit, Okx and Poloniex pass `assetType` only as the label on the resulting `SubAccount`; the request itself doesn't vary. Okx issues the same [`AccountBalance(ctx, currency.EMPTYCODE)`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/okx/okx_wrapper.go#L671) every time. Against a local server returning `401 {"code":"50111","msg":"Invalid API Key"}`:

```
Okx, 6 enabled assets, revoked key
  master 9ced764   1 x GET /account/balance
  2a0a2ed          6 x GET /account/balance    identical request, identical 401
```

This is narrower than the amplification I raised on [`6c1ee4f`](https://github.com/thrasher-corp/gocryptotrader/commit/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc) and then resolved, and I'm not reopening that. A 401 on GateIO's `/futures/usdt/accounts` genuinely can differ from `/spot/accounts` — that's the case this PR exists to serve, and I measured 7 distinct paths there. What's different here is that the second request cannot return anything the first didn't.

Not yours, but it is the sharpest illustration and worth its own issue: Deribit's validator can never succeed. `UpdateAccountBalances` builds [`accounts.NewSubAccount(asset.All, "")`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchanges/deribit/deribit_wrapper.go#L343) and `Save` rejects it at [`accounts.go:244`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/exchange/accounts/accounts.go#L244) because `asset.All.IsValid()` is false. With well-formed credentials and a venue answering 200 to everything:

```
ValidateAPICredentials(spot) -> error loading [all] SubAccount holdings: unsupported asset type
whole-flow HTTP requests        master 2   |   2a0a2ed 10
```

Pre-existing — master does it once — but this PR does it five times.

engine/engine.go
```diff
@@ -921,6 +905,43 @@ func (bot *Engine) LoadExchange(name string) error {
...
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Error validating credentials: %v", exchangeName, assetErr)
+			errs = common.AppendError(errs, assetErr)
+			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
+				break
+			}
+			continue
+		}
+		return nil
+	}
+	return errs
```
thrasher-, line 952: Heimdall: **Minor** — every credential failure now reaches the log twice. This one's mine: I asked for the `%v` at [`engine.go:896`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/engine/engine.go#L896) on [`58136939`](https://github.com/thrasher-corp/gocryptotrader/commit/58136939f4e0dadd40edec6dad742c6594e12d54) for the empty-asset case and didn't account for the overlap with the per-asset line.

**Fix:** log only where the caller can't — the partial-success path, which returns nil and so never reaches the caller's warning:

```suggestion
	var errs error
	for _, a := range assets {
		if err := validate(ctx, a); err != nil {
			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
				break
			}
			continue
		}
		if errs != nil { // the caller only logs errs when every account fails, so surface the recovered ones here
			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credentials validated on %s after earlier failures: %v", exchangeName, a, errs)
		}
		return nil
	}
	return errs
```

**Why:** on the total-failure path every byte of the per-asset lines reappears inside the aggregate. GateIO, seven enabled assets, all 401:

```
master 9ced764   1 [WARN]
2a0a2ed          8 [WARN]   7 per-asset + a 690-char aggregate repeating all 7
suggestion       1 [WARN]
```

Single enabled asset with blank credentials is the same shape: master 1 line, `2a0a2ed` 2 lines with identical content.

It also improves the case this PR exists for. Spot fails, a futures account validates:

```
2a0a2ed     GateIO: Error validating credentials: spot: ...
suggestion  GateIO: Credentials validated on usdtmarginedfutures after earlier failures: spot: ...
```

One thing it costs, so you can weigh it: today a spot failure is visible immediately, whereas this defers the line until the run resolves. If a later asset hangs on an HTTP timeout you'd see nothing until it does. If that matters, keeping the per-asset line at `Debugf` gets both.

Verified: applies clean, `go test ./engine/ -race -count 1` passes in 53.5s with **no test change required**, `make misc_checks` clean, gofumpt delta unchanged.

engine/engine_test.go
```diff
@@ -439,6 +441,78 @@ func TestDryRunParamInteraction(t *testing.T) {
...
+	require.ErrorIs(t, err, errDeliveryAccountMissing, "mixed asset validation must return the aggregated error")
+	assert.Equal(t, asset.Items{
+		asset.Spot,
+		asset.USDTMarginedFutures,
+		asset.PerpetualSwap,
+		asset.DeliveryFutures,
+		asset.Options,
+		asset.Margin,
+		asset.Index,
+	}, seen, "assets should be tried spot first, then futures, then the rest")
```
thrasher-, line 503: Heimdall: **Minor** — the ordering is well covered now, but nothing asserts what the returned error actually says, and that text is the whole of what the operator sees on the caller's warning.

**Fix:**

```suggestion
	}, seen, "assets should be tried spot first, then futures, then the rest")
	for _, a := range seen {
		assert.ErrorContainsf(t, err, a.String()+": "+errDeliveryAccountMissing.Error(), "aggregated error should name %s and why it failed", a)
	}
```

and one line alongside [`engine_test.go:453`](https://github.com/thrasher-corp/gocryptotrader/blob/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86/engine/engine_test.go#L453):

```go
		assert.ErrorContains(t, err, testExchange, "error should name the exchange it came from")
```

**Why:** every assertion in this test is `errors.Is`, so three separate degradations of the error content all leave the suite green:

```
mutant                                                     TestValidateAPICredentials
errs = assetErr             (last failure wins, no aggregation)      PASS
assetErr := err             (drop the "<asset>: " prefix)            PASS
return asset.ErrNotEnabled  (drop the exchange name)                 PASS
```

All three were deliberate additions from the last three rounds, and all three fail with the four lines above. Verified on this head: mutants caught, `go test ./engine/ -race -count 1` passes in 42.2s, `make misc_checks` clean including the require/assert wording and `f`-variant rules.

thrasher- (approved):
Heimdall: Approving on [`2a0a2ed`](https://github.com/thrasher-corp/gocryptotrader/commit/2a0a2ed2f2217ab5c5eea01c577ed05bc8e69d86). All three from the last pass landed exactly as suggested — the guard moved after the tier build so it fails closed if the partition ever stops being total, `asset.PerpetualSwap` pins the futures tier rather than two members of it, and the caller's warning finally carries the reason for the one case that never goes through the loop. gloriousCode's empty-asset guard is in and holding. CI is green across the board, including `lint`, `miscellaneous checks` and both codecov gates.

Nothing blocking. Four comments, and two of them are consequences of my own earlier asks.

The one worth your time is the early break. It recognises `ErrCredentialsAreEmpty`, which needs every credential field blank — but the placeholders GCT ships are `"Key"`/`"Secret"`, so `IsEmpty()` is false and you get `errRequiresAPIKey` instead, which isn't in the set. Neither is a key pasted without its secret. Flipping `authenticatedSupport: true` over the shipped credentials, through real `LoadExchange`:

```
                              master 9ced764   2a0a2ed
GateIO   validate calls             1             7
         [WARN] lines               1             8
Deribit  HTTP requests              1             5
```

Deribit's five are real network round trips, because its `UpdateAccountBalances` makes a public call before the credential lookup. Doing the credential check once up front takes all of those to zero ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2355#discussion_r3910121151)).

The other three:

- The helper assumes each asset is a distinct validation target. For BTSE, Bitfinex, BitMEX, Coinbase, Deribit, Okx and Poloniex the asset only labels the resulting `SubAccount` — Okx sends the same `GET /account/balance` six times against a revoked key where master sent it once. Narrower than the amplification I raised and resolved on [`6c1ee4f`](https://github.com/thrasher-corp/gocryptotrader/commit/6c1ee4fd3a69d91d74e5a0092bb49abee9e890fc); I'm asking rather than prescribing, since the fix depends on your intent ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2355#discussion_r3910121161)).
- Every failure is logged twice, seven times over on GateIO. That's my `%v` from last round meeting the per-asset line from the round before; there's a shape that keeps both purposes and emits one line ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2355#discussion_r3910121168)).
- Nothing asserts the returned error's content, and three mutations of it pass ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2355#discussion_r3910121176)).

I deliberately didn't raise the 401 amplification on venues that really do validate per asset — GateIO hits seven distinct paths, and probing them is the point of this PR — nor the map-ordering nondeterminism in the futures tier, which predates you.

Verified locally on both trees: `go test ./engine/ -race -count 1` passes on `2a0a2ed` in 43.3s and with each suggestion applied (47.4s / 53.5s / 42.2s), `make misc_checks` clean throughout, and `golangci-lint run ./engine/...` reports the same 30 issues on `2a0a2ed` as on [`9ced764`](https://github.com/thrasher-corp/gocryptotrader/commit/9ced764459526432eaf4e03944f73f411c81d93b) with none inside this diff. gofumpt flags `engine.go` identically before and after — three pre-existing hunks, none of them yours — and `engine_test.go` is clean.

Happy to approve as is; take the four at whatever weight you think they deserve.

engine/engine.go
```diff
@@ -892,24 +892,14 @@ func (bot *Engine) LoadExchange(name string) error {
...
-					preferredAsset = a
-					break
-				}
-			}
-			if preferredAsset == 0 {
-				preferredAsset = enabledAssets[0] // last resort pick first available
-			}
+		// Credentials that are unusable fail identically for every account type,
+		// so establish that once rather than once per asset.
+		_, err := b.GetCredentials(ctx)
```
thrasher-: Heimdall: **Should fix** — the preflight reads credentials through `Base` rather than the interface, skipping any custom implementation's own lookup.

**Fix:**

```suggestion
		// Structurally unusable credentials fail identically for every account
		// type, so establish that once rather than once per asset.
		_, err := exch.GetCredentials(ctx)
```

**Why:** `GetCredentials` is on the interface ([`interfaces.go:157`](https://github.com/thrasher-corp/gocryptotrader/blob/fbcc583af1a57b73bd6f5d450d612125d2f7cc26/exchanges/interfaces.go#L157)) and the engine loads exchanges through [`CustomExchangeBuilder`](https://github.com/thrasher-corp/gocryptotrader/blob/fbcc583af1a57b73bd6f5d450d612125d2f7cc26/engine/exchange_manager.go#L28), so a custom implementation can back credentials with something other than `exchange.Base`. Reading through `b.` bypasses that override, disabling authentication on empty `Base` credentials while the exchange's own lookup and validator both succeed. [`engine.go:899`](https://github.com/thrasher-corp/gocryptotrader/blob/fbcc583af1a57b73bd6f5d450d612125d2f7cc26/engine/engine.go#L899) already dispatches through `exch`; this brings the pair into line. My `b.` from last round, not something you introduced.

Built-in exchanges are unaffected — all resolve to the `Base` method. Re-measured after the swap: GateIO placeholder credentials still 0 validate calls, Deribit still 0 HTTP requests.

The comment is in the same suggestion only because it's the same two lines: "unusable" reads wider than what the preflight catches, which is *structural* unusability — a revoked or permission-limited key passes it and continues to per-asset validation by design.

This also settles the break at [`engine.go:940`](https://github.com/thrasher-corp/gocryptotrader/blob/fbcc583af1a57b73bd6f5d450d612125d2f7cc26/engine/engine.go#L940). I'd assumed the preflight made it dead code; it doesn't, because a custom `GetCredentials` can satisfy the preflight and the loop can still surface either sentinel. Keeping it was right.

Verified: applies clean, `go test ./engine/ -race -count 1` passes with no test change, `make misc_checks` clean, gofumpt unchanged.

engine/engine.go
```diff
@@ -921,6 +911,45 @@ func (bot *Engine) LoadExchange(name string) error {
...
+		if err := validate(ctx, a); err != nil {
+			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
+			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
+				break
+			}
+			continue
+		}
+		if errs != nil { // the caller only logs errs when every account fails, so surface the recovered ones here
+			gctlog.Warnf(gctlog.ExchangeSys, "%s: Credentials validated on %s after earlier failures: %v", exchangeName, a, errs)
+		}
```
thrasher-, line 949: Heimdall: **Should fix** — this line can report a validation that never happened.

**Fix:**

```suggestion
		if errs != nil { // the caller only logs errs when every account fails, so surface the recovered ones here
			gctlog.Warnf(gctlog.ExchangeSys, "%s: Authenticated support retained on %s after earlier failures: %v", exchangeName, a, errs)
		}
```

**Why:** [`CheckTransientError`](https://github.com/thrasher-corp/gocryptotrader/blob/fbcc583af1a57b73bd6f5d450d612125d2f7cc26/exchanges/exchange.go#L957) converts a `net.Error` to `nil` so a network blip doesn't disable authentication, and every wrapper returns through it. `nil` from `validate` therefore means *either* "this account validated" *or* "this account was unreachable" — the loop can't separate them, and the message assumes the first.

Measured, spot returning a genuine 401 and the next account timing out:

```
CheckTransientError(&url.Error{...i/o timeout}) = <nil>

[WARN] Probe net error captured, will not disable authentication Get "https://api.example/x": dial: i/o timeout
[WARN] GateIO: Credentials validated on usdtmarginedfutures after earlier failures: spot: ... 401 ... INVALID_KEY
        ^ usdtmarginedfutures validated nothing; the request never completed
```

"Authenticated support retained on" is true either way and doesn't assert the venue answered. The wording is mine from last round.

I looked at this last round and set it aside as matching the existing fail-open design — too generous, since master only ever tried one account, so a genuine 401 always disabled authentication. I'd failed to reproduce it against Okx because my probe dropped connections repeatedly, exhausting the requester's retry policy at [`request.go:306`](https://github.com/thrasher-corp/gocryptotrader/blob/fbcc583af1a57b73bd6f5d450d612125d2f7cc26/exchanges/request/request.go#L306), which wraps the error past `CheckTransientError`'s type assertion. A single clean `*url.Error` fires it. Separating a genuine success from a swallowed transient properly needs a contract change across all 23 validators, which isn't this PR's job.

Verified: applies clean, `go test ./engine/ -race -count 1` passes with no test change, `make misc_checks` clean, gofumpt unchanged.

thrasher- (approved):
Heimdall: Approving on [`fbcc583a`](https://github.com/thrasher-corp/gocryptotrader/commit/fbcc583af1a57b73bd6f5d450d612125d2f7cc26). It stops one unfunded account taking authentication down for a whole exchange, and it now costs *less* than master in the case operators actually hit — a key that isn't set up yet:

```
authenticatedSupport: true over the shipped credentials
                          master 9ced764   fbcc583
GateIO  validate calls          1             0
Deribit HTTP requests           1             0
```

Three things worth naming:

- The `enabledAssets[0]` fallback is gone, and with it a panic. gloriousCode's condition through the real `LoadExchange` — every asset disabled, authentication on — kills master with `index out of range [0] with length 0`; this head disables authentication with a logged reason and carries on.
- Tiering is exhaustive: all 20 of `asset.Supported()` tried exactly once, none missed, none duplicated, tier order held.
- `validateAPICredentials` is at 100% coverage and the test defends its contract — six mutations, five caught (aggregation, asset tag, exchange name, futures tier, the break). The survivor is the new log line, and no engine test asserts log output.

A revoked key still probes all seven GateIO accounts, which is the point — those are distinct endpoints and a key can hold spot permissions without futures — but the failure is now reported in one warning instead of eight, and the partial-success path names the account that carried it.

Two refinements, both to suggestions I made rather than anything that went wrong here:

- The preflight reads through `Base`, not the interface. `GetCredentials` is on `IBotExchange` and the engine loads via `CustomExchangeBuilder`, so a custom exchange would get its override skipped ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2355#discussion_r3911248977)).
- The success line can claim more than it knows. `CheckTransientError` turns a `net.Error` into `nil`, so a spot 401 followed by a timeout logs *Credentials validated on usdtmarginedfutures*, which validated nothing ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2355#discussion_r3911248987)).

On the second: I looked at transient masking last round and set it aside as matching the existing fail-open design. Too generous — on master only one account was ever tried, so a genuine 401 always disabled authentication. The wording is all I'd ask for; separating a real success from a swallowed transient needs a contract change across all 23 validators.

Verified on this head: `go test ./engine/ -race -count 1` passes (42-51s across runs), as it does with either suggestion applied, neither needing a test change. `make misc_checks` clean, gofumpt flags `engine.go` with master's three pre-existing hunks and `engine_test.go` clean, 15/15 CI green. Correcting my last footer: the `golangci-lint` count of 30 was stale — a correctly rebuilt v2.9.0 reports 47 on both trees, all `typecheck` noise on an untouched file. Parity holds, the number didn't.

The untested authenticated branch of `LoadExchange` is inherited — zero on master too, since all 23 exchanges in `testdata/configtest.json` ship `authenticatedSupport: false`. Worth a future pass, not this one.

gloriousCode's `CHANGES_REQUESTED` is still the blocking decision; the guard they asked for landed in [`58136939`](https://github.com/thrasher-corp/gocryptotrader/commit/58136939f4e0dadd40edec6dad742c6594e12d54) and I re-confirmed it holds.

engine/engine.go
```diff
@@ -921,6 +911,45 @@ func (bot *Engine) LoadExchange(name string) error {
...
+	}
+	if len(assets) == 0 {
+		return fmt.Errorf("%s: %w", exchangeName, asset.ErrNotEnabled)
+	}
+
+	var errs error
+	for _, a := range assets {
+		if err := validate(ctx, a); err != nil {
+			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
+			if errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
```
thrasher-: Heimdall: **Should fix** — a transport failure is treated as asset-specific, so an unreachable venue is retried once per enabled account.

**Fix:**

```suggestion
	var errs error
	var netErr net.Error
	for _, a := range assets {
		if err := validate(ctx, a); err != nil {
			errs = common.AppendError(errs, fmt.Errorf("%s: %w", a, err))
			// A transport failure means the venue itself is unreachable, so no other account can answer either
			if errors.As(err, &netErr) || errors.Is(err, exchange.ErrCredentialsAreEmpty) || errors.Is(err, exchange.ErrAuthenticationSupportNotEnabled) {
```

Not a one-click apply — it also wants `"net"` in the import block at [`engine.go:7`](https://github.com/thrasher-corp/gocryptotrader/blob/ba6f625fb397c50a6bae033ccaed70ded01bf7c3/engine/engine.go#L7), and a case for the new branch. The existing loop at [`engine_test.go:506`](https://github.com/thrasher-corp/gocryptotrader/blob/ba6f625fb397c50a6bae033ccaed70ded01bf7c3/engine/engine_test.go#L506) takes one with no new structure, given `"fmt"` and `"net/url"` alongside [`engine_test.go:5`](https://github.com/thrasher-corp/gocryptotrader/blob/ba6f625fb397c50a6bae033ccaed70ded01bf7c3/engine/engine_test.go#L5):

```go
	for _, venueWideErr := range []error{
		exchange.ErrCredentialsAreEmpty,
		exchange.ErrAuthenticationSupportNotEnabled,
		&url.Error{Op: "Get", URL: "https://example.com/accounts", Err: context.DeadlineExceeded},
	} {
		validated = nil
		err = validateAPICredentials(t.Context(), testExchange, asset.Items{asset.Spot, asset.USDTMarginedFutures}, func(_ context.Context, a asset.Item) error {
			validated = append(validated, a)
			return fmt.Errorf("%w, authenticated request failed", venueWideErr)
		})
		require.ErrorIs(t, err, venueWideErr, "venue-wide failure must be returned")
		assert.Equal(t, asset.Items{asset.Spot}, validated, "venue-wide failure should stop further validation")
	}
```

**Why:** GateIO's seven enabled assets against a listener that accepts connections and never answers, `HTTPTimeout` 400ms — and it scales with enabled assets against the shipped 15s default rather than my 400ms:

```
                     validate calls   LoadExchange
master 9ced764             1             4.81s
ba6f625f                   7            18.08s
with this fix              1             4.75s
```

The requester hands back `failed to retry request exceeds maximum retry attempts: err: Get "…/spot/accounts": context deadline exceeded (Client.Timeout exceeded while awaiting headers)`. `errors.As(err, &netErr)` is true through that wrap; `ctx.Err()` is nil, so a context check does not catch it.

It doesn't weaken the case this PR exists for — a revoked-key 401 arrives as a decoded venue error, so every account is still tried:

```
venue 401            5 / 5 accounts tried
transport failure    1 / 5
```

`net.Error` is also the predicate [`CheckTransientError`](https://github.com/thrasher-corp/gocryptotrader/blob/ba6f625fb397c50a6bae033ccaed70ded01bf7c3/exchanges/exchange.go#L957) already uses for "the venue, not the credentials". `errors.Is(err, context.DeadlineExceeded)` covers the measured case without the new import if you'd rather keep it narrow; I prefer the broader one because connection-refused and TLS failures are the same class.

Verified: applies clean, eight full `go test ./engine/ -race -count 1` runs green, the new test case fails without the `engine.go` change and passes with it, `make misc_checks` clean, gofumpt delta unchanged.

thrasher- (commented):
Heimdall: everything from the last four rounds is in and verified — the preflight dispatches through the interface now, and the retained-support warning no longer claims a validation that may not have happened.

One new thing, and it's the last from me: the fallback treats a transport failure as asset-specific, so an unreachable venue is retried once per enabled account. GateIO's seven assets against a listener that accepts and never answers, 400ms HTTP timeout — 1 validate call and 4.81s on master, 7 calls and 18.08s here. Same end state, but [`SetupExchanges` waits on all of it](https://github.com/thrasher-corp/gocryptotrader/blob/ba6f625fb397c50a6bae033ccaed70ded01bf7c3/engine/engine.go#L1048), so a venue outage holds start-up about four times longer. Tested two-line fix inline; a revoked-key 401 is a decoded venue error rather than a transport failure, so the fallback still walks every account.

Also flagged, not blocking: nothing exercises the wiring inside `LoadExchange` — deleting the whole branch leaves `go test ./engine/ -race` green. Pre-existing, master is identical, and I don't have a remedy I'd stand behind: the test I wrote kills all five mutations but coincided with six live-API failures across eighteen full-suite runs, against none in thirteen without it.

CI's two red jobs are `exchanges/huobi` live-API tests — a response-shape change in `TestGetLiquidationOrders` and a rate limit in `TestGetHistoricCandles`. Unrelated; both pass locally on master and on this head. `make misc_checks` is clean and `gofumpt` reports only the three pre-existing hunks.

**To unblock:**
1. Break out of the fallback on a transport failure ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2355#discussion_r3920303045))

thrasher- (approved):
Heimdall: applied and verified. `errors.AsType[net.Error]` is a better form than the `errors.As` I suggested, and it lands in Go 1.26, so it clears both `go.mod` and the CI pin.

An unreachable venue is back to one attempt — GateIO's seven assets against a listener that accepts and never answers, 400ms HTTP timeout: 1 validate call and 4.72s, against 4.81s on master and 18.08s on `ba6f625f`. The fallback is untouched where it matters: a revoked-key 401 is still a decoded venue error and still walks all seven accounts, and reverting the break still fails the new test case. Nothing in the tree implements `net.Error`, so the predicate can only catch genuine transport failures.

`go test ./engine/ -race -count 1` green over four runs, `golangci-lint run ./engine/...` 0 issues, `make misc_checks` clean, gofumpt delta unchanged. CI's two red jobs are live-venue tests in `exchanges/bitmex` and `exchanges/bithumb`, neither reachable from a diff that touches only `engine/`; both pass locally on master and on this head.

LGTM.

PR #2357, opened 2026-09-01
exchanges/order: fix UpdateOrderFromDetail re-deriving RemainingAmount from trades

Tyagiquamar:
## Problem

`Detail.UpdateOrderFromDetail` re-derived `RemainingAmount` by subtracting each merged trade's `Amount` inside the trade-merge loop:

```go
for ... {
    d.Trades = append(d.Trades, m.Trades[x])
    updated = true
    m.RemainingAmount -= m.Trades[x].Amount   // <-- double-counts fills
}
```

`RemainingAmount` is an authoritative value reported by the exchange (`Amount - ExecutedAmount`). Subtracting trade amounts from it again double-counts fills already reflected in the exchange's own figure, producing a too-low (potentially negative) remaining amount and incorrect order state.

## Fix

Use the exchange-reported `RemainingAmount` as-is; never re-derive it from trades.

## Test

`TestUpdateOrderFromDetailRemainingAmountInvariant` covers the invariant that merging trades must not double-subtract from `RemainingAmount`.

`go test ./exchanges/order/ -run TestUpdateOrderFromDetailRemainingAmountInvariant -count=1 -v`

### DIFF
--- a/exchanges/order/orders.go
+++ b/exchanges/order/orders.go
@@ -297,9 +297,8 @@ func (d *Detail) UpdateOrderFromDetail(m *Detail) error {
 			updated = true
 		}
 	}
-	// RemainingAmount is an authoritative value reported by the exchange
-	// (Amount - ExecutedAmount); it must be used as-is and never re-derived by
-	// subtracting trade amounts, which would double-count fills.
+	// Do not derive RemainingAmount from Trades: trade lists may be cumulative
+	// or from a different snapshot. A supplied non-zero value is preserved.
 	if m.RemainingAmount > 0 && m.RemainingAmount != d.RemainingAmount {
 		d.RemainingAmount = m.RemainingAmount
 		updated = true
--- a/exchanges/order/orders.go
+++ b/exchanges/order/orders.go
@@ -296,8 +296,10 @@ func (d *Detail) UpdateOrderFromDetail(m *Detail) error {
 			d.Trades = append(d.Trades, m.Trades[x])
 			updated = true
 		}
-		m.RemainingAmount -= m.Trades[x].Amount
 	}
+	// RemainingAmount is an authoritative value reported by the exchange
+	// (Amount - ExecutedAmount); it must be used as-is and never re-derived by
+	// subtracting trade amounts, which would double-count fills.
 	if m.RemainingAmount > 0 && m.RemainingAmount != d.RemainingAmount {
 		d.RemainingAmount = m.RemainingAmount
 		updated = true
--- a/exchanges/order/orders_test.go
+++ b/exchanges/order/orders_test.go
@@ -1156,13 +1156,13 @@ func TestUpdateOrderFromDetailRemainingAmountInvariant(t *testing.T) {
 	}
 
 	od := &Detail{OrderID: "abc", Amount: 10}
-	require.NoError(t, od.UpdateOrderFromDetail(om))
+	require.NoError(t, od.UpdateOrderFromDetail(om), "UpdateOrderFromDetail must not error")
 
-	assert.Equal(t, 10.0, od.Amount)
-	assert.Equal(t, 4.0, od.ExecutedAmount)
-	assert.Equal(t, 6.0, od.RemainingAmount, "RemainingAmount is the authoritative remaining, not reduced by trade amounts")
-	assert.Equal(t, od.Amount, od.ExecutedAmount+od.RemainingAmount, "ExecutedAmount + RemainingAmount equals Amount")
-	assert.Equal(t, 6.0, om.RemainingAmount, "incoming detail is not mutated")
+	assert.Equal(t, 10.0, od.Amount, "Amount should be unchanged")
+	assert.Equal(t, 4.0, od.ExecutedAmount, "ExecutedAmount should come from the incoming detail")
+	assert.Equal(t, 6.0, od.RemainingAmount, "RemainingAmount should be the authoritative remaining, not reduced by trade amounts")
+	assert.Equal(t, od.Amount, od.ExecutedAmount+od.RemainingAmount, "ExecutedAmount + RemainingAmount should equal Amount")
+	assert.Equal(t, 6.0, om.RemainingAmount, "incoming detail should not be mutated")
 }
 
 func TestClassificationError_Error(t *testing.T) {
--- a/exchanges/order/orders_test.go
+++ b/exchanges/order/orders_test.go
@@ -1139,11 +1139,10 @@ func TestUpdateOrderFromDetail(t *testing.T) {
 }
 
 // TestUpdateOrderFromDetailRemainingAmountInvariant ensures that when an
-// incoming detail carries an authoritative RemainingAmount (Amount -
-// ExecutedAmount, as every exchange populates it) alongside the fill(s) that
-// produced it, the stored RemainingAmount is trusted rather than re-derived by
-// subtracting trade amounts. Re-deriving double-counts fills and violates the
-// invariant ExecutedAmount + RemainingAmount == Amount.
+// incoming detail carries a RemainingAmount alongside the fill(s) that
+// produced it, the stored RemainingAmount is trusted rather than re-derived
+// by subtracting trade amounts. Re-deriving double-counts fills and violates
+// the invariant ExecutedAmount + RemainingAmount == Amount.
 func TestUpdateOrderFromDetailRemainingAmountInvariant(t *testing.T) {
 	t.Parallel()
 
@@ -1163,6 +1162,44 @@ func TestUpdateOrderFromDetailRemainingAmountInvariant(t *testing.T) {
 	assert.Equal(t, 6.0, od.RemainingAmount, "RemainingAmount should be the authoritative remaining, not reduced by trade amounts")
 	assert.Equal(t, od.Amount, od.ExecutedAmount+od.RemainingAmount, "ExecutedAmount + RemainingAmount should equal Amount")
 	assert.Equal(t, 6.0, om.RemainingAmount, "incoming detail should not be mutated")
+
+	// A fill at least as large as the reported remainder drove the old
+	// subtraction negative, which failed the guard in UpdateOrderFromDetail and
+	// dropped the update entirely, leaving the stored value stale rather than
+	// merely wrong.
+	od = &Detail{OrderID: "def", Amount: 10, RemainingAmount: 10}
+	om = &Detail{
+		OrderID:         "def",
+		Amount:          10,
+		ExecutedAmount:  8,
+		RemainingAmount: 2,
+		Trades:          []TradeHistory{{TID: "2", Price: 100, Amount: 8}},
+	}
+	require.NoError(t, od.UpdateOrderFromDetail(om), "UpdateOrderFromDetail must not error")
+	assert.Equal(t, 2.0, od.RemainingAmount, "RemainingAmount should be updated when the fill is larger than the reported remainder")
+}
+
+// TestUpdateOrderFromDetailTradesOnly pins the behaviour for feeds that report
+// fills without a remainder (bitmex execution, kraken ownTrades, huobi
+// trade.clearing). The engine merges such a detail into a copy of the stored
+// order first, so this method sees the stored remainder alongside the new
+// trade; that remainder is now left alone rather than reduced by the trade
+// amount. Those exchanges get their remainder from a companion order feed or
+// the REST poll instead.
+func TestUpdateOrderFromDetailTradesOnly(t *testing.T) {
+	t.Parallel()
+
+	od := &Detail{OrderID: "ghi", Amount: 10, RemainingAmount: 10}
+	om := &Detail{
+		OrderID:         "ghi",
+		Amount:          10,
+		RemainingAmount: 10,
+		Trades:          []TradeHistory{{TID: "1", Price: 100, Amount: 4}},
+	}
+	require.NoError(t, od.UpdateOrderFromDetail(om), "UpdateOrderFromDetail must not error")
+
+	require.Len(t, od.Trades, 1, "trade must be merged")
+	assert.Equal(t, 10.0, od.RemainingAmount, "RemainingAmount should be left as reported, not reduced by the merged trade")
 }
 
 func TestClassificationError_Error(t *testing.T) {
--- a/exchanges/order/orders_test.go
+++ b/exchanges/order/orders_test.go
@@ -1138,6 +1138,33 @@ func TestUpdateOrderFromDetail(t *testing.T) {
 	assert.NotEqual(t, id, od.InternalOrderID, "Should not be able to update the internal order ID after initialisation")
 }
 
+// TestUpdateOrderFromDetailRemainingAmountInvariant ensures that when an
+// incoming detail carries an authoritative RemainingAmount (Amount -
+// ExecutedAmount, as every exchange populates it) alongside the fill(s) that
+// produced it, the stored RemainingAmount is trusted rather than re-derived by
+// subtracting trade amounts. Re-deriving double-counts fills and violates the
+// invariant ExecutedAmount + RemainingAmount == Amount.
+func TestUpdateOrderFromDetailRemainingAmountInvariant(t *testing.T) {
+	t.Parallel()
+
+	om := &Detail{
+		OrderID:         "abc",
+		Amount:          10,
+		ExecutedAmount:  4,
+		RemainingAmount: 6, // authoritative: 10 - 4
+		Trades:          []TradeHistory{{TID: "1", Price: 100, Amount: 4}},
+	}
+
+	od := &Detail{OrderID: "abc", Amount: 10}
+	require.NoError(t, od.UpdateOrderFromDetail(om))
+
+	assert.Equal(t, 10.0, od.Amount)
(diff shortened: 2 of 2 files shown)


thrasher-:
@codex please review

thrasher- (commented):
Heimdall: Good fix, and the test earns its place — it fails on master with three assertions and passes here. I replayed it through a real `OrderManager` on HitBTC's payload shape: master decrements the exchange's remainder twice per websocket report, fails the `> 0` guard and drops the update, leaving an order reading as fully open after 40% of it has filled. On this branch it tracks correctly. The same double-count hits the REST path. Mechanism and measurements are inline.

One thing I'd want addressed before approving: the deletion also removes the only thing that decremented orders on feeds reporting fills *without* a remainder — bitmex, kraken `ownTrades`, huobi `trade.clearing`, all on by default. I'm not asking for a code change, but it should be pinned in a test rather than left silent.

`go test ./exchanges/order/... ./exchanges/futures/... ./engine/... -race -count 1` passes on this branch and on master, and `golangci-lint` gives the same two pre-existing `goconst` findings on both, on lines you haven't touched. The red `ubuntu-latest/386` job is `backtester/engine` losing an ephemeral port in `TestStartRPCServerProxyFailureStopsListener` — unrelated to this change.

**To unblock:**
1. Narrow the comment so it stops asserting a contract the code doesn't keep ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2357#discussion_r3903895601))
2. Cover the fill-exceeds-remainder case, the branch behind the worst symptom ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2357#discussion_r3903895623))
3. Pin the trades-only behaviour with a test ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2357#discussion_r3903895631))

Happy to approve once those are in. I'll raise the wider `RemainingAmount` contract separately — zero-presence semantics, deriving a remainder for trades-only feeds, and poloniex mixing base and quote units across `ExecutedAmount`/`RemainingAmount` — so none of that lands on you here.

exchanges/order/orders.go
```diff
@@ -296,8 +296,10 @@ func (d *Detail) UpdateOrderFromDetail(m *Detail) error {
 			d.Trades = append(d.Trades, m.Trades[x])
 			updated = true
 		}
-		m.RemainingAmount -= m.Trades[x].Amount
 	}
+	// RemainingAmount is an authoritative value reported by the exchange
+	// (Amount - ExecutedAmount); it must be used as-is and never re-derived by
+	// subtracting trade amounts, which would double-count fills.
```
thrasher-: Heimdall: **Should fix** — the comment asserts a contract the code doesn't keep, and that isn't true of the callers either.

**Fix:**
```suggestion
	// Do not derive RemainingAmount from Trades: trade lists may be cumulative
	// or from a different snapshot. A supplied non-zero value is preserved.
```

**Why:** three things in the current wording don't hold.

"Must be used as-is" is contradicted by the `m.RemainingAmount > 0` guard on the very next line — a terminal report where the exchange says the remainder is `0` is discarded. Measured on this branch:

```
report 1: exchange said remaining=6  executed=4  -> stored remaining=6 executed=4  invariant holds? true
report 2: exchange said remaining=0  executed=10 -> stored remaining=6 executed=10 invariant holds? false
after cancel:                                       stored remaining=6 status=CANCELLED (exchange said remaining=0)
```

`(Amount - ExecutedAmount)` isn't how it's populated everywhere: [btcmarkets](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/btcmarkets/btcmarkets_websocket.go#L274) sets `RemainingAmount` from `OpenVolume` and never sets `ExecutedAmount` at all, and [poloniex](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/poloniex/poloniex_wrapper.go#L1116) computes the remainder in base units while its `ExecutedAmount` is in quote.

And "authoritative" is too strong for a struct that's [explicitly partially populated](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/order/order_types.go#L225) — [coinbase](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/coinbase/coinbase_wrapper.go#L576-L581) and [btse](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/btse/btse_wrapper.go#L743) fetch the order snapshot and its fills in separate calls, so a fill landing between them yields a remainder that genuinely predates a trade in the same `Detail`.

The replacement keeps the landmine marker — which is the part worth having — without promising the rest.

exchanges/order/orders_test.go
```diff
@@ -1138,6 +1138,33 @@ func TestUpdateOrderFromDetail(t *testing.T) {
...
+	}
+
+	od := &Detail{OrderID: "abc", Amount: 10}
+	require.NoError(t, od.UpdateOrderFromDetail(om), "UpdateOrderFromDetail must not error")
+
+	assert.Equal(t, 10.0, od.Amount, "Amount should be unchanged")
+	assert.Equal(t, 4.0, od.ExecutedAmount, "ExecutedAmount should come from the incoming detail")
+	assert.Equal(t, 6.0, od.RemainingAmount, "RemainingAmount should be the authoritative remaining, not reduced by trade amounts")
+	assert.Equal(t, od.Amount, od.ExecutedAmount+od.RemainingAmount, "ExecutedAmount + RemainingAmount should equal Amount")
+	assert.Equal(t, 6.0, om.RemainingAmount, "incoming detail should not be mutated")
```
thrasher-, line 1164: Heimdall: **Minor** — the branch behind the worst engine symptom isn't covered.

**Fix:**
```suggestion
	assert.Equal(t, 6.0, om.RemainingAmount, "incoming detail should not be mutated")

	// A fill at least as large as the reported remainder drove the old
	// subtraction negative, which failed the guard in UpdateOrderFromDetail and
	// dropped the update entirely, leaving the stored value stale rather than
	// merely wrong.
	od = &Detail{OrderID: "def", Amount: 10, RemainingAmount: 10}
	om = &Detail{
		OrderID:         "def",
		Amount:          10,
		ExecutedAmount:  8,
		RemainingAmount: 2,
		Trades:          []TradeHistory{{TID: "2", Price: 100, Amount: 8}},
	}
	require.NoError(t, od.UpdateOrderFromDetail(om), "UpdateOrderFromDetail must not error")
	assert.Equal(t, 2.0, od.RemainingAmount, "RemainingAmount should be updated when the fill is larger than the reported remainder")
```

**Why:** your case has `sum(trades) = 4 < RemainingAmount = 6`, so master lands on a *wrong* value (2). Once the fill reaches the reported remainder the subtraction goes negative, the `> 0` guard fails, and the update is dropped — the stored figure doesn't move at all. That second signature is the one the engine actually hits, and it's what leaves an order reading as fully open after it has partly filled.

Applied, the amended test fails on master with four assertions instead of three — the new one reporting `expected: 2, actual: 10` — and passes here; `gofmt` clean, and `golangci-lint` unchanged — the new case needs its own order ID, otherwise `"abc"` reaches six occurrences and trips `goconst`.

Worth folding both into a table if you'd rather; the second case is the part I care about.

exchanges/order/orders.go
```diff
@@ -296,8 +296,10 @@ func (d *Detail) UpdateOrderFromDetail(m *Detail) error {
 			d.Trades = append(d.Trades, m.Trades[x])
 			updated = true
 		}
-		m.RemainingAmount -= m.Trades[x].Amount
 	}
```
thrasher-, line 299: Heimdall: **Should fix** — this line was doing two jobs, and only one of them was the bug. Worth a test rather than a code change.

**Why:** the subtraction was there by design, not by accident. [#446](https://github.com/thrasher-corp/gocryptotrader/commit/22ff441a772ba8f692333467934bdb3b51d063df) moved what had been an unconditional copy of `RemainingAmount` to *after* the trade loop, added the `> 0` guard, and put the subtraction inside the loop — in both merge methods. So it's worth a moment on what it was doing before it goes.

**Job one, which you're right to remove.** The engine merges twice per update: once into the copy from [`GetByExchangeAndID`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/engine/order_manager.go#L968), again inside [`updateExisting`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/engine/order_manager.go#L1001). Replaying [HitBTC's payload shape](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/hitbtc/hitbtc_websocket.go#L469-L479) through [that sequence](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/engine/websocketroutine_manager.go#L314-L323) on a real `OrderManager`:

```
MASTER  ws report 1: exchange said Remaining=6 Executed=4 -> order manager holds Amount=10 Executed=4 Remaining=10
PR      ws report 1: exchange said Remaining=6 Executed=4 -> order manager holds Amount=10 Executed=4 Remaining=6
```

6 is decremented to 2, then to -2, fails the guard, and the update is thrown away. Futures get a third pass — `updateExisting` hands the *stored* pointer to [`TrackNewOrder`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/engine/order_manager.go#L1008), which merges again at [`futures.go:774`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/futures/futures.go#L774). REST is sharper still, since `processOrders` upserts through here on its own ten-second loop and [`btse_wrapper.go:729-773`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/btse/btse_wrapper.go#L729-L773) attaches the *cumulative* fill list beside `RemainingAmount: Size - FilledSize`:

```
MASTER  size=10 filled=4: exchange said Remaining=6 -> stored Remaining=2   WRONG
MASTER  size=10 filled=6: exchange said Remaining=4 -> stored Remaining=10  WRONG
MASTER  size=10 filled=9: exchange said Remaining=1 -> stored Remaining=10  WRONG
PR      lands on 6, 4 and 1.
```

**Job two, which goes with it.** Feeds that report fills *without* a remainder aren't unaffected, as I first assumed: the first merge hands the copy's stored remainder to the second, so the subtraction landed there and was the only thing decrementing those orders.

```
                                     MASTER      PR        truth
single fill 4 of 10:                 6           10        6
single fill 10 of 10:                10          10        0
fills 4 -> 3 -> 2:                   6, 6, 6     10,10,10  6, 3, 1
```

That's [bitmex `execution`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/bitmex/bitmex_websocket.go#L212), [kraken `ownTrades`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/kraken/kraken_websocket.go#L262) and [huobi `trade.clearing`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/huobi/huobi_websocket.go#L413), all enabled by default. Kraken and huobi also subscribe to a companion order feed that sets `RemainingAmount` properly, so they self-correct in the same burst; bitmex's `order` branch [never sets it](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/exchanges/bitmex/bitmex_websocket.go#L254), so bitmex has nothing but the REST poll.

**I'm not asking you to restore it**, and the trade is clearly worth making. What it did was right only for the *first partial fill* of a multi-fill order — on a complete fill it was already wrong, and after the first fill it froze — and doing it properly needs presence information, since `0` means "unset" for every other field in this method. Nothing in the repo branches on the value either: the list RPCs recompute `OpenVolume` as `Amount - ExecutedAmount` ([`rpcserver.go:971`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/engine/rpcserver.go#L971)), so only [`GetOrder`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/engine/rpcserver.go#L1148), the two futures RPCs, the websocket log line and gctscript's [`amountremaining`](https://github.com/thrasher-corp/gocryptotrader/blob/630657286a5b12ea2d2792db6b5e244f69a0bef5/gctscript/modules/gct/exchange.go#L364) see it raw.

**Fix:** pin the new behaviour so the change is recorded rather than silent:

```go
// TestUpdateOrderFromDetailTradesOnly pins the behaviour for feeds that report
// fills without a remainder (bitmex execution, kraken ownTrades, huobi
// trade.clearing). The engine merges such a detail into a copy of the stored
// order first, so this method sees the stored remainder alongside the new
// trade; that remainder is now left alone rather than reduced by the trade
// amount. Those exchanges get their remainder from a companion order feed or
// the REST poll instead.
func TestUpdateOrderFromDetailTradesOnly(t *testing.T) {
	t.Parallel()

	od := &Detail{OrderID: "ghi", Amount: 10, RemainingAmount: 10}
	om := &Detail{
		OrderID:         "ghi",
		Amount:          10,
		RemainingAmount: 10,
		Trades:          []TradeHistory{{TID: "1", Price: 100, Amount: 4}},
	}
	require.NoError(t, od.UpdateOrderFromDetail(om), "UpdateOrderFromDetail must not error")

	require.Len(t, od.Trades, 1, "trade must be merged")
	assert.Equal(t, 10.0, od.RemainingAmount, "RemainingAmount should be left as reported, not reduced by the merged trade")
}
```

It fails on master with `expected: 10, actual: 6` and passes here, so it captures exactly what changed. It needs its own order ID rather than reusing `"abc"`, which would otherwise reach six occurrences and trip `goconst`.

exchanges/order/orders.go
Tyagiquamar: Done in 94b0817 — replaced with the suggested wording, and softened the matching test doc comment the same way.

exchanges/order/orders_test.go
Tyagiquamar: Done in 94b0817 — second case added as suggested with its own order ID ("def"); passes on this branch, gofmt clean.

exchanges/order/orders.go
Tyagiquamar: Done in 94b0817 — TestUpdateOrderFromDetailTradesOnly added as suggested with order ID "ghi"; passes on this branch. Full ./exchanges/order/ suite and go vet are green locally.

exchanges/order/orders_test.go
```diff
@@ -1138,6 +1138,33 @@ func TestUpdateOrderFromDetail(t *testing.T) {
...
+	om := &Detail{
+		OrderID:         "abc",
+		Amount:          10,
+		ExecutedAmount:  4,
+		RemainingAmount: 6, // authoritative: 10 - 4
+		Trades:          []TradeHistory{{TID: "1", Price: 100, Amount: 4}},
+	}
+
+	od := &Detail{OrderID: "abc", Amount: 10}
+	require.NoError(t, od.UpdateOrderFromDetail(om))
```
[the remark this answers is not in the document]
Tyagiquamar: Already addressed in 6306572 — require calls carry "must" messages and assert calls carry "should" messages per the repo convention.

shazbert (approved):
nice catch, nothing from me

thrasher- (approved):
Heimdall: All three addressed, verified against the control.

The comment is narrowed, the fill-exceeds-remainder case is in, and `TestUpdateOrderFromDetailTradesOnly` pins the split-stream behaviour so the change is recorded rather than silent. Both new cases are genuine regression tests: on master `TestUpdateOrderFromDetailRemainingAmountInvariant` now fails four assertions instead of three, and `TestUpdateOrderFromDetailTradesOnly` fails with `expected: 10, actual: 6`. Both pass here.

You also softened the test's doc comment, which I'd only mentioned in passing — appreciated.

`gofmt`, `go vet` and `golangci-lint` are clean (the same two pre-existing `goconst` findings, on lines you haven't touched), and `order`, `futures` and `engine` pass under `-race`.

The red `windows-latest` job is `gateio.TestGetSupportedFlashSwapCurrencyPairs` taking an HTTP 500 from Gate.io's own server — the same test passed on ubuntu amd64 in that run, and nothing here reaches gateio. The 386 job that was red last round now passes.

I'll open the follow-up on the wider `RemainingAmount` contract now, so none of it lands on you here.

gloriousCode (approved):
I think this was more a relic of FTX's days. While Kraken, Huobi and Bitmex are impacted by this change, it ultimately works out when final order update comes through. Whereas for all other exchanges, the RemainingAmount trade subtraction impacted accuracy

tACK

PR #2361, opened 2026-09-02
btse: fix funding history symbol query

### LINKED ISSUE #2360: btse: FetchFundingHistory constructs an invalid URL when symbol is provided
## Context

`FetchFundingHistory` appends the encoded query parameters directly to the funding history endpoint.

## Expected Behavior

Calling:

```go
e.FetchFundingHistory(ctx, "BTC")
```

should request:

```text
/futures/api/v2.1/funding_history?symbol=BTC
```

## Current Behavior

The method currently constructs the endpoint using:

```go
btseFuturesFunding + params.Encode()
```

Since `params.Encode()` returns `symbol=BTC` without a leading question mark, the resulting request path is:

```text
/futures/api/v2.1/funding_historysymbol=BTC
```

The existing test only passes an empty symbol, so the malformed URL is not covered.

## Failure Information (for bugs)

The symbol filter cannot be used because it becomes part of the URL path instead of a query parameter.

### Steps to Reproduce

1. Call `FetchFundingHistory(ctx, "BTC")`.
2. Inspect the outgoing request URL.
3. Observe that the request path ends in `funding_historysymbol=BTC` instead of using `?symbol=BTC`.

### Failure Logs

No failure logs are required. The malformed request can be reproduced with a local HTTP test server that inspects the request URI.

luangucun:
# PR Description

Fixes https://github.com/thrasher-corp/gocryptotrader/issues/2360

`FetchFundingHistory` previously appended `url.Values.Encode()` directly to the funding history endpoint. When a symbol was supplied, this produced `funding_historysymbol=BTC` instead of `funding_history?symbol=BTC`.

This change uses `common.EncodeURLValues`, as required by the project's path construction guidelines. The helper adds the query delimiter only when parameters are present.

A regression test uses a local HTTP server to verify both the funding history path and the encoded `symbol=BTC` query.

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [x] `go test ./exchanges/btse -race -count=1`
- [x] `go vet ./exchanges/btse`
- [x] `go test ./exchanges/btse -run TestFetchFundingHistoryWithSymbol -count=20`

## Checklist

### DIFF
--- a/exchanges/btse/btse_test.go
+++ b/exchanges/btse/btse_test.go
@@ -1,7 +1,10 @@
 package btse
 
 import (
+	"io"
 	"log"
+	"net/http"
+	"net/http/httptest"
 	"os"
 	"testing"
 	"time"
@@ -72,6 +75,28 @@ func TestUpdateTradablePairs(t *testing.T) {
 func TestFetchFundingHistory(t *testing.T) {
 	_, err := e.FetchFundingHistory(t.Context(), "")
 	assert.NoError(t, err, "FetchFundingHistory should not error")
+}
+
+func TestFetchFundingHistoryWithSymbol(t *testing.T) {
+	t.Parallel()
+
+	ex := new(Exchange)
+	require.NoError(t, testexch.Setup(ex), "Test exchange Setup must not error")
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method, "FetchFundingHistory should use GET")
+		assert.Equal(t, "/futures/api/v2.1/funding_history", r.URL.Path, "funding history request path should be correct")
+		assert.Equal(t, "symbol=BTC", r.URL.RawQuery, "funding history request should use the symbol query parameter")
+		_, err := io.WriteString(w, `{}`)
+		assert.NoError(t, err, "writing mock response should not error")
+	}))
+	defer server.Close()
+
+	require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestFutures.String(), server.URL), "SetRunningURL must not error")
+	require.NoError(t, ex.SetHTTPClient(server.Client()), "SetHTTPClient must not error")
+
+	_, err := ex.FetchFundingHistory(t.Context(), "BTC")
+	require.NoError(t, err, "FetchFundingHistory must not error for a symbol")
 }
 
 func TestGetMarketsSummary(t *testing.T) {
--- a/exchanges/btse/btse.go
+++ b/exchanges/btse/btse.go
@@ -66,7 +66,7 @@ func (e *Exchange) FetchFundingHistory(ctx context.Context, symbol string) (map[
 	if symbol != "" {
 		params.Set("symbol", symbol)
 	}
-	return resp, e.SendHTTPRequest(ctx, exchange.RestFutures, http.MethodGet, btseFuturesFunding+params.Encode(), &resp, false, queryFunc)
+	return resp, e.SendHTTPRequest(ctx, exchange.RestFutures, http.MethodGet, common.EncodeURLValues(btseFuturesFunding, params), &resp, false, queryFunc)
 }
 
 // GetRawMarketSummary returns an unfiltered list of market pairs


thrasher-:
@codex please review

luangucun:
I fixed an unit test error.

@codex please review

exchanges/btse/btse_test.go
```diff
@@ -74,6 +77,28 @@ func TestFetchFundingHistory(t *testing.T) {
...
+	require.NoError(t, testexch.Setup(ex), "Test exchange Setup must not error")
+
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, http.MethodGet, r.Method, "FetchFundingHistory should use GET")
+		assert.Equal(t, "/futures/api/v2.1/funding_history", r.URL.Path, "funding history request path should be correct")
+		assert.Equal(t, "symbol=BTC", r.URL.RawQuery, "funding history request should use the symbol query parameter")
+		_, err := io.WriteString(w, `{}`)
+		assert.NoError(t, err, "writing mock response should not error")
+	}))
+	defer server.Close()
```
thrasher-, line 93: Heimdall: **Minor** — `defer server.Close()` runs against the repo's convention for httptest servers.

**Fix:**

```suggestion
	t.Cleanup(server.Close)
```

**Why:** `t.Cleanup(<server>.Close)` appears at 21 sites in the tree against 4 for `defer <server>.Close()`, and three of those four are the older `bitmex_test.go`. The two closest templates to this test — [`exchanges/bitfinex/bitfinex_test.go:688`](https://github.com/thrasher-corp/gocryptotrader/blob/2c89d7be0ce174932ee740a4ebbb3eb2acc472c4/exchanges/bitfinex/bitfinex_test.go#L688) and [`exchanges/gateio/gateio_wrapper_test.go:128`](https://github.com/thrasher-corp/gocryptotrader/blob/2c89d7be0ce174932ee740a4ebbb3eb2acc472c4/exchanges/gateio/gateio_wrapper_test.go#L128) — both use `t.Cleanup` immediately before the same `SetHTTPClient(server.Client())` line you have here. I applied the suggestion: `gofumpt` clean, `go test ./exchanges/btse -run TestFetchFundingHistory -race -count=3` passes.

exchanges/btse/btse_test.go
```diff
@@ -74,6 +77,28 @@ func TestFetchFundingHistory(t *testing.T) {
 	assert.NoError(t, err, "FetchFundingHistory should not error")
 }
 
+func TestFetchFundingHistoryWithSymbol(t *testing.T) {
```
thrasher-, line 80: Heimdall: **Minor** — this leaves two top-level tests for one method, and the `symbol == ""` branch is still covered only by a live network call.

**Fix:** fold this into `TestFetchFundingHistory` and drive both branches through the mock server. Not a one-click suggestion, because it also rewrites [`exchanges/btse/btse_test.go:75`](https://github.com/thrasher-corp/gocryptotrader/blob/2c89d7be0ce174932ee740a4ebbb3eb2acc472c4/exchanges/btse/btse_test.go#L75), which is outside the diff:

```go
func TestFetchFundingHistory(t *testing.T) {
	t.Parallel()

	_, err := e.FetchFundingHistory(t.Context(), "")
	assert.NoError(t, err, "FetchFundingHistory should not error")

	for _, tt := range []struct{ symbol, exp string }{
		{symbol: "", exp: ""},
		{symbol: "BTC-PERP", exp: "symbol=BTC-PERP"},
	} {
		ex := new(Exchange)
		require.NoError(t, testexch.Setup(ex), "Test exchange Setup must not error")

		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
			assert.Equal(t, http.MethodGet, r.Method, "FetchFundingHistory should use GET")
			assert.Equal(t, "/futures/api/v2.1/funding_history", r.URL.Path, "FetchFundingHistory should request the funding history path")
			assert.Equalf(t, tt.exp, r.URL.RawQuery, "FetchFundingHistory should send the correct query for symbol %q", tt.symbol)
			_, err := io.WriteString(w, `{}`)
			assert.NoError(t, err, "Writing the mock response should not error")
		}))
		t.Cleanup(server.Close)

		require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestFutures.String(), server.URL), "SetRunningURL must not error")

		_, err = ex.FetchFundingHistory(t.Context(), tt.symbol)
		require.NoErrorf(t, err, "FetchFundingHistory must not error for symbol %q", tt.symbol)
	}
}
```

**Why:** The [coding guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/2c89d7be0ce174932ee740a4ebbb3eb2acc472c4/docs/CODING_GUIDELINES.md#L167) make test deduplication the default. More concretely, [`exchanges/btse/btse.go:66`](https://github.com/thrasher-corp/gocryptotrader/blob/2c89d7be0ce174932ee740a4ebbb3eb2acc472c4/exchanges/btse/btse.go#L66) guards with `if symbol != ""`, and that guard is the only thing keeping a stray `?symbol=` out of the URL — nothing pins it deterministically today. Drop it while "simplifying" and `common.EncodeURLValues` will happily emit `?symbol=`, with only the live test in a position to notice. I ran the version above on your head (`-race -count=3` passes, `gofumpt` clean); copied onto `upstream/master` it fails on the `BTC-PERP` row with the original defect, so it keeps the regression guard intact.

thrasher- (approved):
Heimdall: Good catch, and the test is the part worth singling out. Asserting on `r.URL.RawQuery` rather than `r.URL.Query().Get("symbol")` is exactly right for this defect — `Query()` parses the mangled path into an empty query and would have passed silently, whereas `RawQuery` pins the `?` delimiter itself. It is a genuine regression test: copied onto `upstream/master` it fails with precisely the reported symptom, and passes on your head.

```
# on upstream/master, with this PR's btse_test.go
$ go test ./exchanges/btse -run TestFetchFundingHistoryWithSymbol -race -count=1
    Error:  Not equal:
            expected: "/futures/api/v2.1/funding_history"
            actual  : "/futures/api/v2.1/funding_historysymbol=BTC"
    Error:  Not equal:
            expected: "symbol=BTC"
            actual  : ""
--- FAIL: TestFetchFundingHistoryWithSymbol (0.18s)
```

I swept all 108 non-test `url.Values.Encode()` call sites for siblings of this defect and there are none: the hits in `binance_cfutures.go`, `binance_ufutures.go`, `kraken.go`, `btcmarkets.go`, `huobi.go`, `bitmex.go` and `bitfinex.go` all terminate their endpoint constant with `?` (`"/dapi/v1/depth?"`, `"AssetPairs?"`, `"status/deriv?"`, and so on). BTSE's `FetchFundingHistory` was the only one, so the fix is complete as scoped. The [BTSE futures docs](https://btsecom.github.io/docs/futures/en/) confirm the contract: `GET /api/v2.1/funding_history` with an optional `symbol` parameter.

No blockers. Two minor points inline:

- `defer server.Close()` where the repo overwhelmingly uses `t.Cleanup` ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2361#discussion_r3919136784))
- two top-level tests for one method, with the `symbol == ""` branch still covered only by a live call ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2361#discussion_r3919136809))

And one nit too small for its own thread: the docs' example symbol is `BTC-PERP`; `"BTC"` is not a symbol BTSE futures accepts, so a real one would document the contract a little better.

Verified against `2c89d7be0ce174932ee740a4ebbb3eb2acc472c4`: `go test ./exchanges/btse -race -count=1` passes, the new test is stable at `-count=50`, and `gofumpt` is clean on both changed files (the `btse_websocket.go` it flags is equally unformatted on master). All 15 CI checks are green. I could not run `golangci-lint` locally — my binary is built against go1.26 and panics loading this tree — so I am leaning on the green CI lint job for that gate.

exchanges/btse/btse_test.go
```diff
@@ -74,6 +77,28 @@ func TestFetchFundingHistory(t *testing.T) {
...
+	}))
+	defer server.Close()
+
+	require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestFutures.String(), server.URL), "SetRunningURL must not error")
+	require.NoError(t, ex.SetHTTPClient(server.Client()), "SetHTTPClient must not error")
+
+	_, err := ex.FetchFundingHistory(t.Context(), "BTC")
+	require.NoError(t, err, "FetchFundingHistory must not error for a symbol")
+}
+
```
gloriousCode, line 101: Thank you for fixing the params issue. However, I'd like this test to be updated

I don't really like this test because its returning a valid response for a symbol that doesn't exist -Something that would be highlighted if it wasn't using a server returning fake data.

I don't think this needs a server at all and it also could just be a `t.Run` scenario, but that's optional

PR #2363, opened 2026-09-03
request: context rate-limit coordination barrier

shazbert:
## Summary

Adds context-based rate-limit barriers for coordinating the first rate-limit admission of concurrent HTTP and websocket requests.

Participants wait without reserving capacity. The final arrival checks participant cancellation and reserves capacity for the group under a shared accounting lock. The group is admitted only when every participant can proceed without a rate-limit delay; otherwise, admission is rejected and acquired reservations are cancelled.

## Usage

Create participant contexts with `NewRateLimitBarrierContexts(parent, count)` and run each participant concurrently.

Each request owner must defer `AbortRateLimitBarrier(participantCtx)` so a failure before reaching the first gate releases waiting peers.

Concurrent reuse of a participant before the barrier resolves is rejected. After acceptance, subsequent calls and retries use ordinary rate limiting without re-entering the barrier.

## Behavior

- Coordinates HTTP and websocket rate-limit admission.
- Supports participants without active rate limiters.
- Rejects the group when a participant aborts or cancellation is observed before acceptance.
- Distinguishes peer rejection (`ErrRateLimitBarrierRejected`) from insufficient immediate capacity (`ErrDelayNotAllowed`).
- Removes pending websocket response signatures when coordination fails.

## Scope and Limitations

- Coordinates the first rate-limit gate only, not atomic request execution or order fills.
- Additional gates, cancellation after admission, transport errors, and exchange rejection can still produce partial execution.
- Finite limiters created by `NewRateLimit` have burst 1: participating requests must use distinct limiter instances and weight 1.
- Shared limiters or larger weights require sufficient custom burst capacity. Infinite-rate limiters are exempt.
- Reservation cancellation uses existing accounting. Its pre-existing rounding-related rollback defect is outside this PR; exact capacity restoration is not guaranteed.

## Validation

- Request and websocket suites pass with `-race -count=2`.
- Regression tests detect removal of reservation locking, post-reservation cancellation checks, zero-weight peer release, and parked cancellation handling.
- Scoped lint and repository miscellaneous checks pass.
- Full repository race testing was attempted; failures occurred in unrelated Bitstamp and Huobi response-decoding tests.

### DIFF
--- a/exchange/websocket/connection.go
+++ b/exchange/websocket/connection.go
@@ -215,6 +215,8 @@ func (c *connection) writeToConn(ctx context.Context, epl request.EndpointLimit,
 		if err := rl.RateLimit(ctx); err != nil {
 			return fmt.Errorf("%s websocket connection: rate limit error: %w", c.ExchangeName, err)
 		}
+	} else if err := request.WaitForRateLimitBarrier(ctx); err != nil {
+		return fmt.Errorf("%s websocket connection: rate limit barrier error: %w", c.ExchangeName, err)
 	}
 	// This lock acts as a rolling gate to prevent WriteMessage panics. Acquire after rate limit check.
 	c.writeControl.Lock()
@@ -391,6 +393,7 @@ func (c *connection) SendMessageReturnResponsesWithInspector(ctx context.Context
 	start := time.Now()
 	err = c.SendRawMessage(ctx, epl, gws.TextMessage, outbound)
 	if err != nil {
+		c.Match.RemoveSignature(signature)
 		return nil, err
 	}
 
--- a/exchanges/request/context_helpers.go
+++ b/exchanges/request/context_helpers.go
@@ -2,9 +2,19 @@ package request
 
 import (
 	"context"
+	"errors"
 	"net/http"
+	"sync"
+	"sync/atomic"
 
 	"github.com/thrasher-corp/gocryptotrader/common"
+)
+
+var (
+	// ErrInvalidRateLimitBarrierParticipants is returned when a barrier cannot coordinate at least two requests.
+	ErrInvalidRateLimitBarrierParticipants = errors.New("rate limit barrier requires at least two participants")
+	// ErrRateLimitBarrierParticipantUsed is returned when a one-use participant context reaches a limiter more than once.
+	ErrRateLimitBarrierParticipantUsed = errors.New("rate limit barrier participant already used")
 )
 
 const contextVerboseFlag verbosity = "verbose"
@@ -56,6 +66,116 @@ func WithDelayNotAllowed(ctx context.Context) context.Context {
 func hasDelayNotAllowed(ctx context.Context) bool {
 	_, ok := ctx.Value(delayNotAllowedKey{}).(struct{})
 	return ok
+}
+
+type rateLimitBarrierKey struct{}
+
+type rateLimitBarrier struct {
+	done      chan struct{}
+	remaining uint
+	rejected  bool
+	closed    bool
+	mu        sync.Mutex
+}
+
+type rateLimitBarrierParticipant struct {
+	barrier *rateLimitBarrier
+	used    atomic.Bool
+}
+
+// NewRateLimitBarrierContexts returns distinct one-use contexts whose rate-limit calls proceed only when every participant can proceed immediately.
+// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants.
+func NewRateLimitBarrierContexts(ctx context.Context, participants uint) ([]context.Context, error) {
+	if participants < 2 {
+		return nil, ErrInvalidRateLimitBarrierParticipants
+	}
+	barrier := &rateLimitBarrier{done: make(chan struct{}), remaining: participants}
+	contexts := make([]context.Context, participants)
+	for i := range contexts {
+		contexts[i] = context.WithValue(ctx, rateLimitBarrierKey{}, &rateLimitBarrierParticipant{barrier: barrier}) //nolint:fatcontext // Participants are independent siblings of the same parent.
+	}
+	return contexts, nil
+}
+
+// AbortRateLimitBarrier rejects a participant that failed before reaching RateLimit. It is a no-op after the participant is consumed.
+func AbortRateLimitBarrier(ctx context.Context) {
+	participant, _ := ctx.Value(rateLimitBarrierKey{}).(*rateLimitBarrierParticipant)
+	if participant == nil || !participant.used.CompareAndSwap(false, true) {
+		return
+	}
+	participant.barrier.reject()
+}
+
+// WaitForRateLimitBarrier marks a request without an active limiter as immediately available and waits for its peers.
+func WaitForRateLimitBarrier(ctx context.Context) error {
+	participant := rateLimitBarrierParticipantFromContext(ctx)
+	if participant == nil {
+		return nil
+	}
+	return participant.wait(ctx, true)
+}
+
+func rateLimitBarrierParticipantFromContext(ctx context.Context) *rateLimitBarrierParticipant {
+	participant, _ := ctx.Value(rateLimitBarrierKey{}).(*rateLimitBarrierParticipant)
+	return participant
+}
+
+func (p *rateLimitBarrierParticipant) wait(ctx context.Context, immediate bool) error {
+	if !p.used.CompareAndSwap(false, true) {
+		return ErrRateLimitBarrierParticipantUsed
+	}
+	if !immediate {
+		p.barrier.reject()
+		return ErrDelayNotAllowed
+	}
+	if p.barrier.arrive() {
+		return nil
+	}
+	select {
+	case <-p.barrier.done:
+		return p.barrier.result()
+	case <-ctx.Done():
+		if p.barrier.reject() {
+			return ctx.Err()
+		}
+		return p.barrier.result()
+	}
+}
+
+func (b *rateLimitBarrier) arrive() bool {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.closed {
+		return !b.rejected
+	}
+	b.remaining--
+	if b.remaining == 0 {
+		b.closed = true
+		close(b.done)
+		return true
+	}
+	return false
+}
+
+func (b *rateLimitBarrier) reject() bool {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.closed {
+		return false
+	}
+	b.rejected = true
+	b.closed = true
+	close(b.done)
+	return true
+}
+
+func (b *rateLimitBarrier) result() error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.rejected {
+		return ErrDelayNotAllowed
+	}
+	return nil
 }
 
 type retryNotAllowedKey struct{}
--- a/exchanges/request/context_helpers.go
+++ b/exchanges/request/context_helpers.go
@@ -15,6 +15,8 @@ var (
 	ErrInvalidRateLimitBarrierParticipants = errors.New("rate limit barrier requires at least two participants")
 	// ErrRateLimitBarrierParticipantUsed is returned when a participant context reaches a limiter again before its barrier resolves.
 	ErrRateLimitBarrierParticipantUsed = errors.New("rate limit barrier participant already used")
+	// ErrRateLimitBarrierRejected is returned when a peer aborts or is cancelled before group admission.
+	ErrRateLimitBarrierRejected = errors.New("rate limit barrier rejected")
 )
 
 const contextVerboseFlag verbosity = "verbose"
@@ -74,7 +76,7 @@ type rateLimitBarrier struct {
 	done         chan struct{}
 	participants []*rateLimitBarrierParticipant
 	remaining    uint
-	rejected     bool
+	rejection    error
 	closed       bool
 	mu           sync.Mutex
 }
@@ -90,6 +92,10 @@ type rateLimitBarrierParticipant struct {
 // Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants. The final arrival
 // atomically reserves capacity for the group; shared limiters must have enough burst capacity for all participants, and each weight must fit its
 // limiter's burst. After acceptance, subsequent rate-limit calls using a participant context proceed normally, including retries.
+// Coordination covers only the first gate, not request execution: additional rate-limited calls under the same participant context can fail
+// independently and lose the group's all-or-nothing admission guarantee. Cancellation or transport errors after admission can also split execution.
+// NewRateLimit uses burst 1, so participants using its finite limiters must use distinct limiter instances and weight 1. Larger groups sharing
+// a limiter or larger weights require sufficient burst capacity, for example via GetRateLimiterWithWeight with a custom limiter.
 func NewRateLimitBarrierContexts(ctx context.Context, participants uint) ([]context.Context, error) {
 	if participants < 2 {
 		return nil, ErrInvalidRateLimitBarrierParticipants
@@ -167,7 +173,7 @@ func (b *rateLimitBarrier) arrive(ctx context.Context, participant *rateLimitBar
 	participant.limiter = limiter
 	b.remaining--
 	if b.remaining == 0 {
-		b.rejected = !b.admitLocked()
+		b.rejection = b.admitLocked()
 		b.closed = true
 		close(b.done)
 		return true
@@ -181,7 +187,7 @@ func (b *rateLimitBarrier) reject() bool {
 	if b.closed {
 		return false
 	}
-	b.rejected = true
+	b.rejection = ErrRateLimitBarrierRejected
 	b.closed = true
 	close(b.done)
 	return true
@@ -190,10 +196,7 @@ func (b *rateLimitBarrier) reject() bool {
 func (b *rateLimitBarrier) result() error {
 	b.mu.Lock()
 	defer b.mu.Unlock()
-	if b.rejected {
-		return ErrDelayNotAllowed
-	}
-	return nil
+	return b.rejection
 }
 
 func (b *rateLimitBarrier) resultFor(ctx context.Context) error {
@@ -210,10 +213,7 @@ func (b *rateLimitBarrier) closedResult() (bool, error) {
 	if !b.closed {
 		return false, nil
 	}
-	if b.rejected {
-		return true, ErrDelayNotAllowed
-	}
-	return true, nil
+	return true, b.rejection
 }
 
 func (b *rateLimitBarrier) reuseResult() error {
--- a/exchanges/request/context_helpers.go
+++ b/exchanges/request/context_helpers.go
@@ -71,30 +71,39 @@ func hasDelayNotAllowed(ctx context.Context) bool {
 type rateLimitBarrierKey struct{}
 
 type rateLimitBarrier struct {
-	done      chan struct{}
-	remaining uint
-	rejected  bool
-	closed    bool
-	mu        sync.Mutex
+	done         chan struct{}
+	participants []*rateLimitBarrierParticipant
+	remaining    uint
+	rejected     bool
+	closed       bool
+	mu           sync.Mutex
 }
 
 type rateLimitBarrierParticipant struct {
 	barrier *rateLimitBarrier
+	done    <-chan struct{}
+	limiter *RateLimiterWithWeight
 	used    atomic.Bool
 }
 
 // NewRateLimitBarrierContexts returns distinct contexts whose first rate-limit calls proceed only when every participant can proceed immediately.
-// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants. After acceptance,
-// subsequent rate-limit calls using a participant context proceed normally, including retries. Participants sharing a limiter are paced normally
-// after the barrier accepts them and therefore might not execute simultaneously.
+// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants. The final arrival
+// atomically reserves capacity for the group; shared limiters must have enough burst capacity for all participants, and each weight must fit its
+// limiter's burst. After acceptance, subsequent rate-limit calls using a participant context proceed normally, including retries.
 func NewRateLimitBarrierContexts(ctx context.Context, participants uint) ([]context.Context, error) {
 	if participants < 2 {
 		return nil, ErrInvalidRateLimitBarrierParticipants
 	}
-	barrier := &rateLimitBarrier{done: make(chan struct{}), remaining: participants}
+	barrier := &rateLimitBarrier{
+		done:         make(chan struct{}),
+		participants: make([]*rateLimitBarrierParticipant, participants),
+		remaining:    participants,
+	}
 	contexts := make([]context.Context, participants)
 	for i := range contexts {
-		contexts[i] = context.WithValue(ctx, rateLimitBarrierKey{}, &rateLimitBarrierParticipant{barrier: barrier}) //nolint:fatcontext // Participants are independent siblings of the same parent.
+		participant := &rateLimitBarrierParticipant{barrier: barrier}
+		barrier.participants[i] = participant
+		contexts[i] = context.WithValue(ctx, rateLimitBarrierKey{}, participant) //nolint:fatcontext // Participants are independent siblings of the same parent.
 	}
 	return contexts, nil
 }
@@ -114,7 +123,8 @@ func WaitForRateLimitBarrier(ctx context.Context) error {
 	if participant == nil {
 		return nil
 	}
-	return participant.wait(ctx, true)
+	_, err := participant.wait(ctx, nil)
+	return err
 }
 
 func rateLimitBarrierParticipantFromContext(ctx context.Context) *rateLimitBarrierParticipant {
@@ -122,43 +132,42 @@ func rateLimitBarrierParticipantFromContext(ctx context.Context) *rateLimitBarri
 	return participant
 }
 
-func (p *rateLimitBarrierParticipant) wait(ctx context.Context, immediate bool) error {
+func (p *rateLimitBarrierParticipant) wait(ctx context.Context, limiter *RateLimiterWithWeight) (bool, error) {
 	if !p.used.CompareAndSwap(false, true) {
-		return p.barrier.reuseResult()
+		return false, p.barrier.reuseResult()
 	}
 	if closed, err := p.barrier.closedResult(); closed {
-		return err
-	}
-	if !immediate {
-		p.barrier.reject()
-		return ErrDelayNotAllowed
+		return true, err
 	}
 	if err := ctx.Err(); err != nil {
 		p.barrier.reject()
-		return err
+		return true, err
 	}
-	if p.barrier.arrive() {
-		return nil
+	if p.barrier.arrive(ctx, p, limiter) {
+		return true, p.barrier.resultFor(ctx)
 	}
 	select {
 	case <-p.barrier.done:
-		return p.barrier.result()
+		return true, p.barrier.resultFor(ctx)
 	case <-ctx.Done():
 		if p.barrier.reject() {
-			return ctx.Err()
+			return true, ctx.Err()
 		}
-		return p.barrier.result()
+		return true, p.barrier.resultFor(ctx)
 	}
 }
 
-func (b *rateLimitBarrier) arrive() bool {
+func (b *rateLimitBarrier) arrive(ctx context.Context, participant *rateLimitBarrierParticipant, limiter *RateLimiterWithWeight) bool {
 	b.mu.Lock()
 	defer b.mu.Unlock()
 	if b.closed {
-		return !b.rejected
+		return true
 	}
+	participant.done = ctx.Done()
+	participant.limiter = limiter
 	b.remaining--
 	if b.remaining == 0 {
+		b.rejected = !b.admitLocked()
 		b.closed = true
 		close(b.done)
 		return true
@@ -185,6 +194,14 @@ func (b *rateLimitBarrier) result() error {
 		return ErrDelayNotAllowed
 	}
 	return nil
+}
+
+func (b *rateLimitBarrier) resultFor(ctx context.Context) error {
+	err := b.result()
+	if err != nil && ctx.Err() != nil {
+		return ctx.Err()
+	}
+	return err
 }
 
 func (b *rateLimitBarrier) closedResult() (bool, error) {
--- a/exchanges/request/context_helpers.go
+++ b/exchanges/request/context_helpers.go
@@ -13,7 +13,7 @@ import (
 var (
 	// ErrInvalidRateLimitBarrierParticipants is returned when a barrier cannot coordinate at least two requests.
 	ErrInvalidRateLimitBarrierParticipants = errors.New("rate limit barrier requires at least two participants")
-	// ErrRateLimitBarrierParticipantUsed is returned when a one-use participant context reaches a limiter more than once.
+	// ErrRateLimitBarrierParticipantUsed is returned when a participant context reaches a limiter again before its barrier resolves.
 	ErrRateLimitBarrierParticipantUsed = errors.New("rate limit barrier participant already used")
 )
 
@@ -83,8 +83,10 @@ type rateLimitBarrierParticipant struct {
 	used    atomic.Bool
 }
 
-// NewRateLimitBarrierContexts returns distinct one-use contexts whose rate-limit calls proceed only when every participant can proceed immediately.
-// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants.
+// NewRateLimitBarrierContexts returns distinct contexts whose first rate-limit calls proceed only when every participant can proceed immediately.
+// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants. After acceptance,
+// subsequent rate-limit calls using a participant context proceed normally, including retries. Participants sharing a limiter are paced normally
+// after the barrier accepts them and therefore might not execute simultaneously.
 func NewRateLimitBarrierContexts(ctx context.Context, participants uint) ([]context.Context, error) {
 	if participants < 2 {
 		return nil, ErrInvalidRateLimitBarrierParticipants
@@ -122,11 +124,18 @@ func rateLimitBarrierParticipantFromContext(ctx context.Context) *rateLimitBarri
 
 func (p *rateLimitBarrierParticipant) wait(ctx context.Context, immediate bool) error {
 	if !p.used.CompareAndSwap(false, true) {
-		return ErrRateLimitBarrierParticipantUsed
+		return p.barrier.reuseResult()
+	}
+	if closed, err := p.barrier.closedResult(); closed {
+		return err
 	}
 	if !immediate {
 		p.barrier.reject()
 		return ErrDelayNotAllowed
+	}
+	if err := ctx.Err(); err != nil {
+		p.barrier.reject()
+		return err
 	}
 	if p.barrier.arrive() {
 		return nil
@@ -176,6 +185,26 @@ func (b *rateLimitBarrier) result() error {
 		return ErrDelayNotAllowed
 	}
 	return nil
+}
+
+func (b *rateLimitBarrier) closedResult() (bool, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if !b.closed {
+		return false, nil
+	}
+	if b.rejected {
+		return true, ErrDelayNotAllowed
+	}
+	return true, nil
+}
+
+func (b *rateLimitBarrier) reuseResult() error {
+	closed, err := b.closedResult()
+	if !closed {
+		return ErrRateLimitBarrierParticipantUsed
+	}
+	return err
 }
 
 type retryNotAllowedKey struct{}
--- a/exchanges/request/limit.go
+++ b/exchanges/request/limit.go
@@ -102,7 +102,7 @@ func (r *Requester) InitiateRateLimit(ctx context.Context, e EndpointLimit) erro
 		return ErrRequestSystemIsNil
 	}
 	if atomic.LoadInt32(&r.disableRateLimiter) == 1 {
-		return nil
+		return WaitForRateLimitBarrier(ctx)
 	}
 	if err := common.NilGuard(r.limiter); err != nil {
 		return err
@@ -141,6 +141,17 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
 		reserved = append(reserved, r.limiter.ReserveN(tn, 1))
 	}
 	finalDelay := reserved[len(reserved)-1].DelayFrom(tn)
+	barrierParticipant := rateLimitBarrierParticipantFromContext(ctx)
+	if barrierParticipant != nil {
+		r.m.Unlock()
+		if err := barrierParticipant.wait(ctx, finalDelay == 0); err != nil {
+			r.m.Lock()
+			cancelAll(reserved, tn)
+			r.m.Unlock()
+			return err
+		}
+		return nil
+	}
 
 	if finalDelay == 0 {
 		r.m.Unlock()
--- a/exchanges/request/limit.go
+++ b/exchanges/request/limit.go
@@ -48,7 +48,10 @@ type RateLimitDefinitions map[any]*RateLimiterWithWeight
 type RateLimiterWithWeight struct {
 	limiter *rate.Limiter
 	weight  Weight
-	m       sync.Mutex
+	// Redundant under rateLimitReservationMu in production, but retained so
+	// TestRateLimitReservationLock and TestRateLimitBarrierCancellationDuringAdmission
+	// can pause admission deterministically to verify locking and cancellation safeguards.
+	m sync.Mutex
 }
 
 // NewRateLimit creates a new RateLimit based of time interval and how many actions allowed and breaks it down to an
--- a/exchanges/request/limit.go
+++ b/exchanges/request/limit.go
@@ -125,27 +125,19 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
 // RateLimit throttles a request based on weight, delaying the request.
 // Errors if no delay is permitted via the context and a delay is required.
 func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
-	return r.rateLimit(ctx, true)
-}
-
-func (r *RateLimiterWithWeight) rateLimit(ctx context.Context, coordinate bool) error {
 	if err := common.NilGuard(r); err != nil {
 		return err
 	}
 
 	if r.weight == 0 {
-		if coordinate {
-			AbortRateLimitBarrier(ctx)
-		}
+		AbortRateLimitBarrier(ctx)
 		return errInvalidWeight
 	}
 
-	if coordinate {
-		if participant := rateLimitBarrierParticipantFromContext(ctx); participant != nil {
-			admitted, err := participant.wait(ctx, r)
-			if err != nil || admitted {
-				return err
-			}
+	if participant := rateLimitBarrierParticipantFromContext(ctx); participant != nil {
+		admitted, err := participant.wait(ctx, r)
+		if err != nil || admitted {
+			return err
 		}
 	}
 
@@ -194,16 +186,16 @@ func (r *RateLimiterWithWeight) reserveLocked(at time.Time) ([]*rate.Reservation
 }
 
 // admitLocked reserves every participant's capacity as one transaction. The barrier lock must be held by the caller.
-func (b *rateLimitBarrier) admitLocked() bool {
+func (b *rateLimitBarrier) admitLocked() error {
 	rateLimitReservationMu.Lock()
 	defer rateLimitReservationMu.Unlock()
 
 	at := time.Now()
 	reservations := make([]*rate.Reservation, 0, len(b.participants))
 	for _, participant := range b.participants {
-		if participant.done == nil && !participant.used.Load() || contextDone(participant.done) {
+		if contextDone(participant.done) {
 			cancelAll(reservations, at)
-			return false
+			return ErrRateLimitBarrierRejected
 		}
 		if participant.limiter == nil {
 			continue
@@ -214,16 +206,16 @@ func (b *rateLimitBarrier) admitLocked() bool {
 		reservations = append(reservations, reserved...)
 		if delay != 0 {
 			cancelAll(reservations, at)
-			return false
+			return ErrDelayNotAllowed
 		}
 	}
 	for _, participant := range b.participants {
 		if contextDone(participant.done) {
 			cancelAll(reservations, at)
-			return false
+			return ErrRateLimitBarrierRejected
 		}
 	}
-	return true
+	return nil
 }
 
 func contextDone(done <-chan struct{}) bool {
--- a/exchanges/request/limit.go
+++ b/exchanges/request/limit.go
@@ -124,6 +124,10 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
 // RateLimit throttles a request based on weight, delaying the request.
 // Errors if no delay is permitted via the context and a delay is required.
 func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
+	return r.rateLimit(ctx, true)
+}
+
+func (r *RateLimiterWithWeight) rateLimit(ctx context.Context, coordinate bool) error {
 	if err := common.NilGuard(r); err != nil {
 		return err
 	}
@@ -141,16 +145,20 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
 		reserved = append(reserved, r.limiter.ReserveN(tn, 1))
 	}
 	finalDelay := reserved[len(reserved)-1].DelayFrom(tn)
-	barrierParticipant := rateLimitBarrierParticipantFromContext(ctx)
+	var barrierParticipant *rateLimitBarrierParticipant
+	if coordinate {
+		barrierParticipant = rateLimitBarrierParticipantFromContext(ctx)
+	}
 	if barrierParticipant != nil {
+		// Reservations only probe immediate availability. Restore them before
+		// parking so rejection cannot consume capacity and acceptance is charged
+		// when the request is ready to proceed.
+		cancelAll(reserved, tn)
 		r.m.Unlock()
 		if err := barrierParticipant.wait(ctx, finalDelay == 0); err != nil {
-			r.m.Lock()
-			cancelAll(reserved, tn)
-			r.m.Unlock()
 			return err
 		}
-		return nil
+		return r.rateLimit(ctx, false)
 	}
 
 	if finalDelay == 0 {
--- a/exchanges/request/limit.go
+++ b/exchanges/request/limit.go
@@ -19,7 +19,8 @@ var (
 	ErrRateLimiterAlreadyEnabled  = errors.New("rate limiter already enabled")
 	ErrDelayNotAllowed            = errors.New("delay not allowed")
 
-	errInvalidWeight = errors.New("weight must be equal-or-greater than 1")
+	errInvalidWeight       = errors.New("weight must be equal-or-greater than 1")
+	rateLimitReservationMu sync.Mutex
 )
 
 // RateLimitNotRequired is a no-op rate limiter.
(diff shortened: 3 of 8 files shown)


thrasher-:
@codex please review

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,116 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+	}
+	return participant.wait(ctx, true)
+}
+
+func rateLimitBarrierParticipantFromContext(ctx context.Context) *rateLimitBarrierParticipant {
+	participant, _ := ctx.Value(rateLimitBarrierKey{}).(*rateLimitBarrierParticipant)
+	return participant
+}
+
+func (p *rateLimitBarrierParticipant) wait(ctx context.Context, immediate bool) error {
```
thrasher-: Heimdall: **Blocker** — a participant context is consumed by the first rate-limit gate it passes, but three live request paths put the same context through a gate more than once. All three hard-fail with `ErrRateLimitBarrierParticipantUsed` *after* the group has already committed.

**Fix:** once the barrier has closed **successfully**, a later arrival on that participant should fall through to ordinary rate limiting rather than erroring — the group is already decided, and further calls on that context are just normal requests. Only reuse *before* the group resolves is a genuine programming error. [context_helpers.go:123](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/context_helpers.go#L123)

**Why:**

**1. The HTTP retry loop.** [request.go:157-165](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/request.go#L157-L165) calls `InitiateRateLimit` on every attempt, and `DefaultRetryPolicy` retries on a 429 **or any response carrying a `Retry-After` header** ([retry.go:21](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/retry.go#L21)). Two legs, leg B gets a single 429 then would have succeeded:

```
leg A (200): <nil>
leg B (429): failed to rate limit HTTP request: cannot rate limit request
             rate limit barrier participant already used for endpoint 0
```

Leg B never issues its second attempt. That is the one-legged position the barrier exists to prevent — A's order is on the book, B is permanently dead. A plain 200 carrying `Retry-After` does the same. (Scope: needs default retries — `WithRetryNotAllowed` or a custom policy suppresses it.)

**2. Deribit's websocket path puts one context through two gates** — `InitiateRateLimit` at [deribit_ws_endpoints.go:2732](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/deribit/deribit_ws_endpoints.go#L2732), then `SendMessageReturnResponse` -> `writeToConn`. Deterministic, in both connection configurations, and it fails before anything reaches the wire (reproduced with no underlying socket installed — reaching `WriteMessage` would have panicked):

```
no connection limiter (Deribit today):
  websocket connection: rate limit barrier error: rate limit barrier participant already used
connection limiter present:
  websocket connection: rate limit error: rate limit barrier participant already used
```

`sendWsPayload`'s own retry loop on Deribit error 10040 — their rate-limit code, [deribit_ws_endpoints.go:2749](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/deribit/deribit_ws_endpoints.go#L2749) — reproduces the same failure on the second pass through line 2732.

**3. Any wrapper making two rate-limited calls under one participant context** fails on the second; it never reaches transport:

```
first=<nil>
second=failed to rate limit HTTP request: ... participant already used
wire calls=[1, 1]
```

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,116 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+	if !p.used.CompareAndSwap(false, true) {
+		return ErrRateLimitBarrierParticipantUsed
+	}
+	if !immediate {
+		p.barrier.reject()
+		return ErrDelayNotAllowed
+	}
+	if p.barrier.arrive() {
```
thrasher-: Heimdall: **Blocker** — a cancelled group still sends, on **both** the HTTP and websocket paths. `wait` never consults `ctx` on the `arrive()` branch, so a participant whose context is already cancelled closes the barrier successfully and releases its parked peer. The PR description promises the opposite: "If any participant would be delayed, is cancelled, or aborts before reaching its limiter, the entire group is rejected."

**Fix:** check `ctx.Err()` before `arrive()` and reject rather than close when the arriving participant is already cancelled.

**Why:** cancelling a two-leg group while leg A is parked, with `cancel()` fired concurrently from its own goroutine — a timeout or an errgroup sibling, not an inline call:

```
HTTP  (recording RoundTripper, jittered cancel offset):  both legs sent  3996/4000
WS    (writeToConn callback counter):                    both legs wrote  299/300
```

For the websocket path the window is governed by whether the arriving leg yields before it reaches `arrive()`:

```
work between cancel and last arrival             both legs wrote
none                                                 300/300
mutex round-trip (Match.Set + RemoveSignature)       299/300
CPU burn (stands in for json.Marshal)                296/300
runtime.Gosched()                                      2/300
time.Sleep(50us) or longer                             0/300
```

Only an explicit yield closes it, and `SendMessageReturnResponsesWithInspector` does `json.Marshal` then `Match.Set` then `SendRawMessage` ([connection.go:383-397](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchange/websocket/connection.go#L383-L397)) — exactly the two rows sitting at ~99%. `writeToConn` has no context check of its own ([connection.go:214-224](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchange/websocket/connection.go#L214-L224)). Measured at GOMAXPROCS=64, so the parked peer had a free core and still lost: closing `Done` only makes it runnable, while the arriving leg runs straight into `arrive()`.

`doRequest`'s `ctx.Done()` check at the top of each attempt does **not** save the HTTP path — it only catches cancellation already visible at attempt entry. Once leg B is past it, the same race applies.

The `cancellation rejects peer and restores reservation` subtest passes because its cancel and the peer's arrival are separated by a scheduling yield.

Same mechanism as Codex's P1 ([thread](https://github.com/thrasher-corp/gocryptotrader/pull/2363#discussion_r3930035670)) — opening a fresh thread because the scope is wider than stated there (it is not websocket-only) and the frequency is measured.

exchanges/request/limit.go
```diff
@@ -141,6 +141,17 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
 		reserved = append(reserved, r.limiter.ReserveN(tn, 1))
 	}
 	finalDelay := reserved[len(reserved)-1].DelayFrom(tn)
+	barrierParticipant := rateLimitBarrierParticipantFromContext(ctx)
+	if barrierParticipant != nil {
+		r.m.Unlock()
+		if err := barrierParticipant.wait(ctx, finalDelay == 0); err != nil {
```
thrasher-: Heimdall: **Blocker** — a successful barrier bursts straight through the rate limit it is meant to coordinate. A participant reserves at `tn` ([limit.go:137-143](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L137-L143)), parks here for an unbounded time waiting for its peers, then returns at line 153 without revalidating. Its peer reserves against a limiter that has been refilling the whole time, so both act at once.

**Fix:** cancel the reservation before parking and re-take it once the group is accepted, so the token is spent when the request actually goes out rather than when the participant arrived.

**Why:** limiter of 1 per 100ms, burst 1 — measuring the gap between the two group members' actions:

```
no barrier (two sequential requests)      100ms apart   <- correct pacing
barrier, peer parked 120ms                 32us apart
barrier, peer parked 500ms                  9us apart
```

The longer a peer parks, the tighter the burst. This defeats the reason the limiter is constructed with burst 1 in the first place — [limit.go:140](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L140): *"This avoids needing burst capacity in the limiter, which would otherwise allow the rate limit to be exceeded over short periods."* Independent of the reimbursement issue on line 146; no competing request touches the limiter here.

exchanges/request/limit.go
```diff
@@ -141,6 +141,17 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
 		reserved = append(reserved, r.limiter.ReserveN(tn, 1))
 	}
 	finalDelay := reserved[len(reserved)-1].DelayFrom(tn)
+	barrierParticipant := rateLimitBarrierParticipantFromContext(ctx)
+	if barrierParticipant != nil {
+		r.m.Unlock()
```
thrasher-: Heimdall: **Should fix** — a rejected group's reservation is never reimbursed if anything else touches the same limiter while the barrier is parked.

**Fix:** cancelling before the park (see the comment on line 147) resolves this too. As written, releasing `r.m` at [limit.go:146](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L146) and re-acquiring it at [limit.go:148](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L148) breaks the invariant every other path in `RateLimit` holds: `r.m` stays held from `ReserveN` through `cancelAll`, so restoration is exact.

**Why:** `Reservation.CancelAt` restores nothing once a later reservation has advanced `lim.lastEvent` past the cancelled reservation's `timeToAct` — `restoreTokens` comes out <= 0 and it returns early ([x/time/rate@v0.15.0 rate.go:181-186](https://github.com/golang/time/blob/v0.15.0/rate/rate.go#L181-L186)). On a 5/sec burst-1 limiter, capacity sampled at a common start time:

```
start                                 1.0000
after parked participant reserved      0.0000
after competitor reserved             -1.0000
after group rejected + cancelAll      -1.0000   <- nothing reimbursed
control, one request only              0.0000
```

To be precise: the limiter is not depleted forever — current-time capacity refills normally — but the rejected group's token is never given back. The `one delayed rejects both and restores reservations` subtest passes only because nothing competes for the limiter while the barrier is open.

Same mechanism as Codex's P2 ([thread](https://github.com/thrasher-corp/gocryptotrader/pull/2363#discussion_r3930035674)) — opening a fresh thread to carry the measurements and to tie the fix to the reservation-lifecycle change on line 147.

exchanges/request/limit_test.go
```diff
@@ -235,6 +295,13 @@ func TestInitiateRateLimit(t *testing.T) {
 	atomic.StoreInt32(&r.disableRateLimiter, 1)
 	err = r.InitiateRateLimit(t.Context(), Unset)
 	assert.NoError(t, err, "should not error when rate limiter is disabled")
+	contexts, err := NewRateLimitBarrierContexts(t.Context(), 2)
```
thrasher-, line 382: Heimdall: **Should fix** — the two new integration branches have no effective regression coverage. I mutated each and ran all seven new/modified barrier tests; both mutants survived:

```
mutant: limit.go:105  disabled-limiter branch -> return nil (bypass coordination)
        ok  exchanges/request

mutant: request.go:163  limiter-less HTTP branch -> bypass WaitForRateLimitBarrier
        ok  exchanges/request
        ok  exchange/websocket
```

**Fix:** this test launches both goroutines and requires both return nil, but never asserts the first one actually *parks* — so it passes identically when coordination is removed. `TestWriteToConnRateLimitBarrierWithoutLimiter` gets this right with its `require.Never` on the first participant; the same shape here would kill the mutant. The limiter-less HTTP branch at [request.go:163](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/request.go#L163) has no integration test at all.

**Why it matters:** for contrast, the signature-leak test in this PR is genuinely effective — reverting only [connection.go:396](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchange/websocket/connection.go#L396) makes it fail with `signature collision`. These two do not clear that bar.

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,116 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+}
+
+type rateLimitBarrierParticipant struct {
+	barrier *rateLimitBarrier
+	used    atomic.Bool
+}
+
+// NewRateLimitBarrierContexts returns distinct one-use contexts whose rate-limit calls proceed only when every participant can proceed immediately.
+// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants.
+func NewRateLimitBarrierContexts(ctx context.Context, participants uint) ([]context.Context, error) {
```
thrasher-, line 93: Heimdall: **Minor** — worth documenting that participants sharing one rate limiter will normally reject each other.

**Why:** `NewRateLimit` hardcodes burst to 1 ([limit.go:63](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L63)), so two participants arriving together on a shared limiter make the second reservation delayed and the group is rejected:

```
shared limiter,  1/sec, burst 1, simultaneous:  err1=delay not allowed  err2=delay not allowed
shared limiter, 10/sec, burst 1, simultaneous:  err1=delay not allowed  err2=delay not allowed
```

It is not absolute — a shared 10/sec limiter passes if the second participant arrives after the token refills (+110ms: both `<nil>`), and a limiter built via `GetRateLimiterWithWeight` with burst 2 passes two concurrent participants. But simultaneity is the whole point of the barrier, so in the intended usage a shared burst-1 limiter rejects. The paired long/short in the PR description therefore wants the two legs on different limiter instances. Worth a line on the doc comment, otherwise the first caller to point both legs at one endpoint reads a 100% rejection rate as a bug.

thrasher- (changes requested):
Heimdall: Reviewed at [`729e7aa`](https://github.com/thrasher-corp/gocryptotrader/commit/729e7aa895e09d2698d4846977c6a5a80b9d68c0). The signature leak fix is a good one and properly tested: reverting only [connection.go:396](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchange/websocket/connection.go#L396) makes `TestSendMessageBarrierRejectionRemovesSignature` fail with `signature collision`, and `Match.Set` has exactly two production callers with no other unguarded path. The barrier state machine itself held up under stress — 5,000 iterations each of four-party success, four-party rejection, duplicate-participant races and abort-versus-arrival races under `-race`, with no lost wakeup, `remaining` underflow or double close. `go test ./exchanges/request/... ./exchange/websocket/... -race -count 2` passes, `make misc_checks` passes, and gofumpt is clean across the seven changed files.

Six comments inline; the evidence is on the lines themselves.

**To unblock:**
1. Participant contexts are single-use, but the HTTP retry loop, Deribit's `sendWsPayload` and any two-call wrapper put one context through a rate-limit gate twice — each fails with `ErrRateLimitBarrierParticipantUsed` after the group has committed, which can leave one leg filled and the other permanently dead. [context_helpers.go:123](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/context_helpers.go#L123)
2. A cancelled group still sends on both paths — 3996/4000 HTTP groups and 299/300 websocket groups sent both legs. [context_helpers.go:131](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/context_helpers.go#L131)
3. A successful barrier bursts past the rate limit it coordinates: two actions 9-32us apart on a limiter that paces them 100ms apart without the barrier. [limit.go:147](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L147)

Blockers 1 and 3 share a root cause — the participant reserves on arrival and never revalidates — so a single reservation-lifecycle change (cancel before parking, re-take on acceptance) closes 3 and the non-gating [limit.go:146](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L146) together.

Not gating: a rejected group's reservation is never reimbursed under contention ([limit.go:146](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit.go#L146)); the two new integration branches survive mutation ([limit_test.go:298](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/limit_test.go#L298)); and the shared-limiter constraint wants a line on the doc comment ([context_helpers.go:88](https://github.com/thrasher-corp/gocryptotrader/blob/729e7aa895e09d2698d4846977c6a5a80b9d68c0/exchanges/request/context_helpers.go#L88)).

The single red check is ambient — [job 100534401852](https://github.com/thrasher-corp/gocryptotrader/actions/runs/33719107557/job/100534401852) fails in `TestAllExchangeWrappers/yobit_wrapper_tests/UpdateOrderbook` with `read tcp ...: connection reset by peer` against yobit.net, while the simultaneous non-sonic wrapper job passed.

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,145 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+		return nil
+	}
+	select {
+	case <-p.barrier.done:
+		return p.barrier.result()
+	case <-ctx.Done():
+		if p.barrier.reject() {
+			return ctx.Err()
+		}
+		return p.barrier.result()
```
thrasher-: Heimdall: **Blocker** — cancellation is only checked on arrival, so a participant cancelled while parked still lets the group commit.

**Fix:** treat cancellation as fatal to the group at any point before acceptance, not just at the gate. A re-check after waking will not do it on its own — by then the peer has been released.

**Why:** the guard at [`:135`](https://github.com/thrasher-corp/gocryptotrader/blob/19ed142158f3421317208848aa798d4a4ab88cd0/exchanges/request/context_helpers.go#L135) fixed the pre-cancelled case I measured last time — 0/2000 now. But a participant cancelled *after* parking races `arrive()`, and acceptance usually wins: `reject()` returns false because the barrier is already closed, so line 150 falls through to `result()`, which is nil.

```
participant parked, then cancelled, then final arrival
  -> both legs accepted: 495/500
```

This reaches the websocket writer too, which goes from the barrier straight into the write lock. It pre-dates this commit rather than being a regression from it — but it is the guarantee the PR body advertises.

exchanges/request/limit.go
```diff
@@ -141,6 +145,21 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
...
+		// Reservations only probe immediate availability. Restore them before
+		// parking so rejection cannot consume capacity and acceptance is charged
+		// when the request is ready to proceed.
+		cancelAll(reserved, tn)
+		r.m.Unlock()
+		if err := barrierParticipant.wait(ctx, finalDelay == 0); err != nil {
+			return err
+		}
+		return r.rateLimit(ctx, false)
+	}
```
thrasher-: Heimdall: **Blocker** — a committed group can still split under an ordinary deadline.

**Fix:** acceptance has to be the final rate-limit decision; the group's capacity should be held from admission rather than re-taken per participant afterwards.

**Why:** cancelling before parking was the right call for the burst, but line 161 re-reserves *after* the barrier released the peer. If anything consumed that limiter during the park, the re-reservation is delayed and the deadline check at [`:175`](https://github.com/thrasher-corp/gocryptotrader/blob/19ed142158f3421317208848aa798d4a4ab88cd0/exchanges/request/limit.go#L175) refuses it. Two distinct 200ms limiters, one interfering request, a plain `context.WithTimeout(100ms)`:

```
barrier closed=true accepted=true
participant0 err = rate limit delay of 200ms will exceed deadline: context deadline exceeded
participant1 err = <nil>
parent ctx err   = <nil>          (70ms still on the deadline)
```

One leg sends, the other is refused, parent context still live. No `WithDelayNotAllowed` needed — I had assumed that was required, and it is not.

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,145 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+
+type rateLimitBarrierParticipant struct {
+	barrier *rateLimitBarrier
+	used    atomic.Bool
+}
+
+// NewRateLimitBarrierContexts returns distinct contexts whose first rate-limit calls proceed only when every participant can proceed immediately.
+// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants. After acceptance,
+// subsequent rate-limit calls using a participant context proceed normally, including retries. Participants sharing a limiter are paced normally
+// after the barrier accepts them and therefore might not execute simultaneously.
```
thrasher-: Heimdall: **Should fix** — the doc covers shared limiters, but not the two cases that actually bite.

**Fix:** note that an accepted participant can still be delayed or refused if its limiter is consumed while it waits, and that weighted participants may never be admissible.

**Why:** `NewRateLimit` hardcodes burst to 1, so a weight-2 participant is never immediately available — `delay not allowed` both at t=0 and after an hour of idle time. That qualifies "coordinates weighted rate-limit reservations" in the PR description.

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,145 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+func (b *rateLimitBarrier) result() error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.rejected {
+		return ErrDelayNotAllowed
+	}
+	return nil
+}
+
+func (b *rateLimitBarrier) closedResult() (bool, error) {
```
thrasher-, line 207: Heimdall: **Minor** — no checked-in test calls `rateLimit`, `closedResult` or `reuseResult` directly.

**Why:** they are reached only through `RateLimit`, so the terminal-state logic has no focused coverage. The two mutation kills I checked exercise the integration branches, not these.

thrasher- (changes requested):
Heimdall: re-checked at `19ed14215`. Reuse, the shared-limiter burst and both mutation gaps are
fixed — verified by running each rather than reading them.

Two blockers remain, both the same shape: the group commits and one leg then doesn't run. Detail is
on the lines.

They share a root — acceptance is not the final rate-limit decision. What closes both together is one
group admission at the last arrival: register each participant's context, limiter and weight,
re-check every context, reserve the whole group under a single accounting lock, then either roll back
and reject or keep the reservations and accept, with nobody re-reserving afterwards.

CI is 16/16 green on this head, so my last review's ambient-failure note is moot. The PR body still
describes participant contexts as single-use, which the code comment now correctly contradicts.

**To unblock:**
1. Reject a cancelled participant at any point before acceptance, not only on arrival ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2363#discussion_r3945842451))
2. Make acceptance the final rate-limit decision ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2363#discussion_r3945842455))

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,162 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+	barrier *rateLimitBarrier
+	done    <-chan struct{}
+	limiter *RateLimiterWithWeight
+	used    atomic.Bool
+}
+
+// NewRateLimitBarrierContexts returns distinct contexts whose first rate-limit calls proceed only when every participant can proceed immediately.
+// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants. The final arrival
+// atomically reserves capacity for the group; shared limiters must have enough burst capacity for all participants, and each weight must fit its
+// limiter's burst. After acceptance, subsequent rate-limit calls using a participant context proceed normally, including retries.
```
thrasher-, line 94: Heimdall: **Should fix** — the all-or-nothing guarantee stops at the first rate-limit gate. This sentence states the mechanism; it is worth stating the consequence, because a path with two gates can still send one leg and refuse the other.

**Fix:** say plainly that the barrier covers the first gate only, and that a caller wrapping two rate-limited calls in one participant context loses the guarantee.

**Why:** driven through the real writer — `InitiateRateLimit` (barrier gate) then `writeToConn` (connection-limiter gate), one participant context, a 100ms deadline, leg 1's connection limiter already used by unrelated traffic:

```
100 groups: writes=100   (all-or-nothing would be 200, or 0), legs refused=100
sample:  websocket connection: rate limit error:
         rate limit delay of 184.489078ms will exceed deadline: context deadline exceeded
```

Same shape as [the deadline blocker](https://github.com/thrasher-corp/gocryptotrader/pull/2363#discussion_r3945842455) from last round, relocated to gate two.

**Not reachable in-tree today.** The only double gate is Deribit's `sendWsPayload` — `InitiateRateLimit` at [deribit_ws_endpoints.go:2732](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/deribit/deribit_ws_endpoints.go#L2732) then `SendMessageReturnResponse` -> `writeToConn` — and Deribit configures no connection limiter ([deribit_wrapper.go:192-196](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/deribit/deribit_wrapper.go#L192-L196)), so its second gate takes the `WaitForRateLimitBarrier` branch and is a no-op after acceptance. It goes live the moment Deribit gets one — which its own `// TODO: Refactor to use rate limiting system` at [deribit_ws_endpoints.go:2718](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/deribit/deribit_ws_endpoints.go#L2718) invites — or when a downstream caller puts two rate-limited calls under one participant context.

The HTTP retry loop is fine, and your fix for it is the right one: attempt 2's gate comes after attempt 1 has already gone out, so both legs have sent.

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,162 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+type rateLimitBarrierParticipant struct {
+	barrier *rateLimitBarrier
+	done    <-chan struct{}
+	limiter *RateLimiterWithWeight
+	used    atomic.Bool
+}
+
+// NewRateLimitBarrierContexts returns distinct contexts whose first rate-limit calls proceed only when every participant can proceed immediately.
+// Each request owner must defer AbortRateLimitBarrier so a failure before reaching the limiter releases the other participants. The final arrival
+// atomically reserves capacity for the group; shared limiters must have enough burst capacity for all participants, and each weight must fit its
```
thrasher-, line 93: Heimdall: **Should fix** — both burst preconditions on this line are accurate, and no limiter in this repo can meet either.

**Fix:** note that `NewRateLimit` fixes burst at 1 and is the only limiter constructor under `exchanges/`, so in practice participants must sit on distinct limiter instances and every participant weight must be 1.

**Why:** [limit.go:64](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit.go#L64) hardcodes `rate.NewLimiter(rate.Limit(rps), 1)`, and there is no other `rate.NewLimiter` call anywhere under `exchanges/`:

```
265 endpoint definitions with weight > 1  (binance, binanceus, kucoin)  -> never admissible
two participants sharing one limiter instance                          -> never admissible
```

Binance's `spotDefaultLimiter` alone backs dozens of endpoints, so pairing any two Binance spot endpoints rejects 100%. The doc is right about what *would* work — it just cannot be built through the public constructors:

```
shared burst=2, two weight-1 participants:   <nil> / <nil>            tokens after=0.001
weight=2 on a burst=2 limiter:               <nil> / <nil>            tokens after=0.001
NewRateLimitWithWeight(time.Second, 10, 2):  delay not allowed / delay not allowed
```

This also qualifies "coordinates weighted rate-limit reservations atomically" in the description. Follows on from [the doc thread](https://github.com/thrasher-corp/gocryptotrader/pull/2363#discussion_r3945842460) last round.

exchanges/request/limit.go
```diff
@@ -168,6 +184,57 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
...
+	for range r.weight {
+		// This avoids needing burst capacity in the limiter, which would otherwise allow the rate limit to be exceeded over short periods
+		reserved = append(reserved, r.limiter.ReserveN(at, 1))
+	}
+	return reserved, reserved[len(reserved)-1].DelayFrom(at)
+}
+
+// admitLocked reserves every participant's capacity as one transaction. The barrier lock must be held by the caller.
+func (b *rateLimitBarrier) admitLocked() bool {
+	rateLimitReservationMu.Lock()
```
thrasher-, line 193: Heimdall: **Should fix** — four mutants survive the full suite, including deleting the mutex on this line.

**Fix:** the four rows below. The last two are cheap deterministic tests.

**Why:** `go test ./exchanges/request/ ./exchange/websocket/ -count 1` after each edit:

```
mutant                                                             result
delete rateLimitReservationMu entirely (both lock sites)           SURVIVED
delete the post-reservation re-check loop, limit.go:220-225        SURVIVED
delete AbortRateLimitBarrier(ctx) from the weight==0 branch, :138  SURVIVED
delete the `case <-ctx.Done()` arm of the park select, :152        SURVIVED
never reject on delay, limit.go:215                                killed x4
reuseResult always errors                                          killed x4
arrive forgets participant.limiter                                 killed x5
admitLocked always accepts                                         killed x5
skip cancelAll when delay != 0                                     killed x3
arrive forgets ctx.Done / resultFor drops ctx.Err / drop closedResult  killed
```

The mutex is the mechanism the redesign rests on and nothing covers it: `contention before final arrival rejects group atomically` at [limit_test.go:199](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit_test.go#L199) runs its competitor sequentially before the final arrival, so it passes with the lock gone.

In fairness, **I could not construct an experiment that observes the difference either** — 0/5000 torn admissions with the mutex and 0/5000 without, across two probe shapes. So this is a coverage gap, not a demonstrated defect. If you hold the window to be unobservable by construction, say so and I'll drop it.

The other two are behaviourally correct today; I confirmed both:

```
zero-weight leg = weight must be equal-or-greater than 1
  parked peer released with: delay not allowed        <- peer parks instead, with :138 gone

cancelled parked participant, no final arrival ever = context canceled
  peer arriving afterwards = delay not allowed        <- hangs, with :152 gone
```

Every existing barrier test supplies a final arrival, which is why the park select's cancellation arm has no coverage.

exchanges/request/limit.go
```diff
@@ -124,41 +125,56 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
...
+		if participant := rateLimitBarrierParticipantFromContext(ctx); participant != nil {
+			admitted, err := participant.wait(ctx, r)
+			if err != nil || admitted {
+				return err
+			}
+		}
 	}
-	finalDelay := reserved[len(reserved)-1].DelayFrom(tn)
+
+	rateLimitReservationMu.Lock()
```
thrasher-, line 147: Heimdall: **Nit** — this lock is now process-wide and sits on every REST request and websocket write. What it costs:

**Why:** 64 goroutines over 64 distinct limiters, `rate.Inf` so the measurement is lock cost rather than pacing, `-benchtime 2s -count 3`, master vs this branch:

```
                   master                  this branch
weight 1     68.45 / 72.73 / 67.88    221.1 / 223.8 / 222.1   ns/op   3.3x
weight 250   13800 / 13023 / 12823    23836 / 24485 / 24623   ns/op   1.9x
```

Negligible against a real round trip — 154ns on a 50ms call — so I am not asking for a change. Flagging it because the serialisation is global rather than per-limiter, and because [`admitLocked`](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit.go#L196) holds it while reserving for every participant. If you wanted the concurrency back: an `RWMutex`, ordinary reservations taking `RLock` and `admitLocked` taking `Lock`, preserves group atomicity — per-limiter ordering is already held by `r.m`.

exchanges/request/limit.go
```diff
@@ -168,6 +184,57 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
...
+
+// admitLocked reserves every participant's capacity as one transaction. The barrier lock must be held by the caller.
+func (b *rateLimitBarrier) admitLocked() bool {
+	rateLimitReservationMu.Lock()
+	defer rateLimitReservationMu.Unlock()
+
+	at := time.Now()
+	reservations := make([]*rate.Reservation, 0, len(b.participants))
+	for _, participant := range b.participants {
+		if participant.done == nil && !participant.used.Load() || contextDone(participant.done) {
```
thrasher-: Heimdall: **Nit** — `participant.done == nil && !participant.used.Load()` is unreachable.

**Fix:** drop it, leaving `if contextDone(participant.done)`.

**Why:** `admitLocked` only runs from [context_helpers.go:170](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L170) at `remaining == 0`, and `remaining` only reaches 0 once every participant has come through [`wait`](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L135), which sets `used` before it calls `arrive`. So `!used.Load()` is false for every participant by the time this loop runs. Replacing the whole condition with `contextDone(participant.done)` passes the suite, consistent with that.

Same for the `ctx.Err()` guard at [context_helpers.go:142](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L142) — deleting it also passes, because `admitLocked`'s own `contextDone` check rejects the group and [`resultFor`](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L199) then prefers `ctx.Err()`, producing the identical `context.Canceled`. Harmless as belt-and-braces; noting it so neither reads as load-bearing.

exchanges/request/context_helpers.go
```diff
@@ -58,6 +68,162 @@ func hasDelayNotAllowed(ctx context.Context) bool {
...
+	b.closed = true
+	close(b.done)
+	return true
+}
+
+func (b *rateLimitBarrier) result() error {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	if b.rejected {
+		return ErrDelayNotAllowed
```
thrasher-: Heimdall: **Nit** — `ErrDelayNotAllowed` is doing two jobs here and callers cannot tell them apart.

**Fix:** a dedicated sentinel — `ErrRateLimitBarrierRejected` — for a group taken down by a peer, keeping `ErrDelayNotAllowed` for "a delay was required and the context forbade it".

**Why:** a participant whose peer called `AbortRateLimitBarrier` before reaching its limiter gets "delay not allowed", which is not what happened to it — it was never delayed. For the paired long/short in the description those are different operational conditions: "my sibling failed to build its order" is worth retrying as a pair; "the venue would have paced me" is worth backing off. Both currently arrive as the same error, from here and from [context_helpers.go:214](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L214).

exchanges/request/limit_test.go
```diff
@@ -103,6 +103,150 @@ func TestRateLimit_Concurrent(t *testing.T) {
...
+		weighted := NewRateLimitWithWeight(time.Second, 1, 2)
+		peer := NewRateLimitWithWeight(time.Second, 1, 1)
+		errs := make(chan error, 2)
+		go func() { errs <- weighted.RateLimit(contexts[0]) }()
+		go func() { errs <- peer.RateLimit(contexts[1]) }()
+		require.ErrorIs(t, <-errs, ErrDelayNotAllowed)
+		require.ErrorIs(t, <-errs, ErrDelayNotAllowed)
+	})
+
+	t.Run("rejected probe restores capacity before competitor", func(t *testing.T) {
```
thrasher-: Heimdall: **Nit** — this name and the message on [limit_test.go:241](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit_test.go#L241) describe the reservation lifecycle you removed.

**Fix:** there is no probe and nothing to restore before admission any more — a participant reserves nothing until [`admitLocked`](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit.go#L196) runs. Something like "parked participant holds no capacity" reads true. [limit_test.go:168](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit_test.go#L168) has the same problem — "cancellation rejects peer and restores reservation", where no reservation is ever taken.

**Why:** the assertions are correct and worth keeping; it is the names that will mislead the next reader working out the lifecycle. The property they actually test now:

```
shared limiter tokens:  start=1.0000   afterPark=1.0000   afterReject=1.0000
```

exchanges/request/limit.go
```diff
@@ -124,41 +125,56 @@ func (r *Requester) GetRateLimiterDefinitions() RateLimitDefinitions {
 // RateLimit throttles a request based on weight, delaying the request.
 // Errors if no delay is permitted via the context and a delay is required.
 func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
+	return r.rateLimit(ctx, true)
+}
+
+func (r *RateLimiterWithWeight) rateLimit(ctx context.Context, coordinate bool) error {
```
thrasher-: Heimdall: **Nit** — `coordinate` has no production caller.

**Why:** `RateLimit` is the only non-test caller and always passes `true` ([limit.go:128](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit.go#L128)); the three `false` sites are all in tests — [limit_test.go:133](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit_test.go#L133), [limit_test.go:135](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit_test.go#L135), [limit_test.go:208](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/limit_test.go#L208) — standing in for an unrelated request competing for the same limiter. That reads fine as intent, but it is production API shape carried for tests; an in-package test helper would keep the seam without the parameter. Happy to leave it if you prefer the explicit flag.

thrasher- (approved):
Heimdall: Approving at [`219ba8d`](https://github.com/thrasher-corp/gocryptotrader/commit/219ba8d92865ffe8387f118fa7b323e394d21229). Both blockers from the last round are fixed, and I verified each by running it rather than reading it: a participant cancelled while parked now takes the group down 500/500 (was 495/500 accepted), and your deadline repro — two 200ms limiters, one interfering request, plain `context.WithTimeout(100ms)` — is 200/200 both-rejected with no splits. Parking holds no reservation any more (`start=1.0 afterPark=1.0 afterReject=1.0`), and I saw zero splits across 1500 mixed 2-4-participant groups and 400 concurrent groups under background traffic, all under `-race`. Nine of the thirteen mutants I planted in the new admission logic were killed.

Eight comments inline; the evidence is on the lines themselves. Two are worth a look before merge — the guarantee stopping at the first rate-limit gate ([context_helpers.go:92](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L92)) and the burst preconditions no limiter in this repo can meet ([context_helpers.go:91](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L91)). The rest is one test-coverage note and five nits. None are conditions on the merge.

The PR body still says participant contexts are single-use and that reuse is rejected; the code and [context_helpers.go:92](https://github.com/thrasher-corp/gocryptotrader/blob/219ba8d92865ffe8387f118fa7b323e394d21229/exchanges/request/context_helpers.go#L92) now say the opposite. Same note as last round.

CI is 16/17. The red one is ambient — [job 101593915836](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34073070891/job/101593915836) fails in `TestAllExchangeWrappers/coinbase_wrapper_tests/UpdateOrderbook-futures-PT-TEST-DELIM-29dec26-cde` on a live `404 {"error":"NOT_FOUND","error_details":"no pricebook found"}`, and nothing here touches Coinbase.

Take the eight or leave them — nothing here blocks.

exchanges/request/limit.go
```diff
@@ -128,37 +129,44 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
...
+	if participant := rateLimitBarrierParticipantFromContext(ctx); participant != nil {
+		admitted, err := participant.wait(ctx, r)
+		if err != nil || admitted {
+			return err
+		}
 	}
-	finalDelay := reserved[len(reserved)-1].DelayFrom(tn)
+
+	rateLimitReservationMu.Lock()
+	r.m.Lock()
```
thrasher-, line 148: Heimdall: **Minor** — `r.m` no longer protects anything in production, and the two new tests now depend on it.

**Fix:** none needed today. Worth a comment on the field saying it exists only so a test can wedge admission, or the next person to simplify it will take two safeguards down with it.

**Why:** `RateLimiterWithWeight.m` has exactly two production acquisitions — this one and [limit.go:203](https://github.com/thrasher-corp/gocryptotrader/blob/29631f38406dafc8527dffbe7ea4f2125d95178a/exchanges/request/limit.go#L203) — and both are strictly nested inside `rateLimitReservationMu` ([limit.go:144](https://github.com/thrasher-corp/gocryptotrader/blob/29631f38406dafc8527dffbe7ea4f2125d95178a/exchanges/request/limit.go#L144), [limit.go:190](https://github.com/thrasher-corp/gocryptotrader/blob/29631f38406dafc8527dffbe7ea4f2125d95178a/exchanges/request/limit.go#L190)). Since that mutex is process-global, no two goroutines can ever contend for `r.m`. Deleting both production acquisitions and running the suite under `-race`:

```
--- FAIL: TestRateLimitReservationLock
    --- FAIL: TestRateLimitReservationLock/ordinary_reservation
    --- FAIL: TestRateLimitReservationLock/group_admission
--- FAIL: TestRateLimitBarrierCancellationDuringAdmission
ok      exchange/websocket
```

Only those two fail — the entire barrier suite and the websocket suite pass without it. They are the only things exercising it, and they hold it deliberately to stall admission and open their observation window.

The trap is that removing `r.m` is a reasonable follow-up given the global lock already serialises everything, and those two tests are the only coverage for the global mutex and the post-reservation re-check. A line on the field is enough.

exchanges/request/limit.go
```diff
@@ -168,6 +176,57 @@ func (r *RateLimiterWithWeight) RateLimit(ctx context.Context) error {
...
+		}
+		if participant.limiter == nil {
+			continue
+		}
+		participant.limiter.m.Lock()
+		reserved, delay := participant.limiter.reserveLocked(at)
+		participant.limiter.m.Unlock()
+		reservations = append(reservations, reserved...)
+		if delay != 0 {
+			cancelAll(reservations, at)
```
thrasher-, line 211: Heimdall: **Minor, and not yours** — `cancelAll` does not restore capacity exactly, so a rejected admission can leave the limiter worse off than it found it. This reproduces unchanged on master; flagging it because admission is a new caller.

**Fix:** nothing in this PR. It wants fixing in `cancelAll` — exact reservation accounting rather than trusting reverse-order `CancelAt` — which is its own change.

**Why:** `Reservation.CancelAt` computes `restoreTokens` by discounting later reservations, and only rewinds `lastEvent` when the derived previous event is not in the past ([x/time/rate@v0.15.0 rate.go:181-186](https://github.com/golang/time/blob/v0.15.0/rate/rate.go#L181-L186)). Cancelling in reverse order is necessary but not sufficient: one nanosecond of rounding breaks the rewind chain and every remaining reservation returns unreimbursed.

On master ([`bdb48ad`](https://github.com/thrasher-corp/gocryptotrader/commit/bdb48ad03243c68244521b58fc490fe7ad2a2f2d)), burst 1, weight 3, ordinary `RateLimit` — no barrier code present:

```
rate= 30/s  after=+1.000000000000  next weight-1 request: <nil>
rate= 60/s  after=-0.000000060000  next weight-1 request: delay not allowed
rate=100/s  after=+1.000000000000  next weight-1 request: <nil>
```

and at Binance's spot shape, refused by an ordinary context deadline rather than `WithDelayNotAllowed`:

```
tokens before heavy weight-250 request: +0.0020
refused with: rate limit delay of 2.499934025s will exceed deadline
tokens after  refusal:                  -211.9934      <- should be +0.0020
  unrelated light request 1-3: refused, delay 2.129s
```

`cancelAll` is byte-identical on master and this branch, and master reproduces every case, so this is latent rather than introduced. What is new here is the route: every limiter built through [NewRateLimit](https://github.com/thrasher-corp/gocryptotrader/blob/29631f38406dafc8527dffbe7ea4f2125d95178a/exchanges/request/limit.go#L56) has burst 1, so a weight>1 participant can never be admitted and always lands on this line. Happy to raise it separately if you would rather keep it off this PR.

thrasher- (approved):
Heimdall: Approving at [`29631f3`](https://github.com/thrasher-corp/gocryptotrader/commit/29631f38406dafc8527dffbe7ea4f2125d95178a). All four surviving mutants from the last round are now killed, each by the test written for it — I re-ran every one. `TestRateLimitReservationLock` is the pick of them: it pins the global accounting lock that the whole admission design rests on, and it was the hardest of the four to get at. The new `ErrRateLimitBarrierRejected` goes further than I asked, separating peer-abort from capacity at admission too, and every leg of a group agrees on which it got. The "finite" qualifier on the burst caveat is exactly right — `rate.Inf` limiters do admit shared and weight-5 participants, so the word is load-bearing rather than hedging. Both earlier blockers stay fixed on this head: 0/500 accepted with a cancelled participant, 0 splits in 200 deadline runs, 0 splits in 1500 mixed 2-5 participant groups.

Two comments inline, both Minor, both non-gating; the evidence is on the lines. One is that [limit.go:145](https://github.com/thrasher-corp/gocryptotrader/blob/29631f38406dafc8527dffbe7ea4f2125d95178a/exchanges/request/limit.go#L145) is now redundant and the new tests depend on it; the other is a pre-existing rollback defect at [limit.go:208](https://github.com/thrasher-corp/gocryptotrader/blob/29631f38406dafc8527dffbe7ea4f2125d95178a/exchanges/request/limit.go#L208) that this PR inherits rather than causes — it reproduces unchanged on master.

The PR body still says participant contexts are single-use and that reuse is rejected; [context_helpers.go:94](https://github.com/thrasher-corp/gocryptotrader/blob/29631f38406dafc8527dffbe7ea4f2125d95178a/exchanges/request/context_helpers.go#L94) says the opposite. Third round for that one.

CI is 16/17. The red is ambient — [job 101640043120](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34089532624/job/101640043120) fails `TestExecuteOrderBuySellSizeLimit` on a live `api.btcmarkets.net` connection reset, in a backtester package this PR does not touch.

Take the two or leave them — nothing here blocks.

exchanges/request/limit.go
shazbert: issue: https://github.com/thrasher-corp/gocryptotrader/issues/2371

exchange/websocket/connection.go
```diff
@@ -391,6 +393,7 @@ func (c *connection) SendMessageReturnResponsesWithInspector(ctx context.Context
 	start := time.Now()
 	err = c.SendRawMessage(ctx, epl, gws.TextMessage, outbound)
 	if err != nil {
+		c.Match.RemoveSignature(signature)
```
thrasher-, line 396: Heimdall: **Minor** — this can remove a *successor's* registration, not just your own.

**Fix:** nothing in this PR. [`Match.RemoveSignature`](https://github.com/thrasher-corp/gocryptotrader/blob/5d5d3d5b1e14f99c40f32ef72a32c3ebec6c6a5b/exchange/websocket/match.go#L79) deletes by key with no check that the entry is the one the caller registered, and the same hazard sits on the three existing call sites at [`connection.go:425`](https://github.com/thrasher-corp/gocryptotrader/blob/5d5d3d5b1e14f99c40f32ef72a32c3ebec6c6a5b/exchange/websocket/connection.go#L425), [`connection.go:429`](https://github.com/thrasher-corp/gocryptotrader/blob/5d5d3d5b1e14f99c40f32ef72a32c3ebec6c6a5b/exchange/websocket/connection.go#L429) and [`connection.go:432`](https://github.com/thrasher-corp/gocryptotrader/blob/5d5d3d5b1e14f99c40f32ef72a32c3ebec6c6a5b/exchange/websocket/connection.go#L432). Worth an issue against `Match` — an identity-checked removal, or a handle returned by `Set` — alongside #2371, rather than a change here.

**Why:** your entry is deleted the moment the peer's answers complete it. If a send reports failure *after* the peer received and answered the frame, and another request has registered the same signature in that window, this line closes the newcomer's channel. The victim gets no error — `waitForResponses` drains the closed channel `expected` times and returns successfully with empty payloads:

```
waitForResponses returned after 0s: err=<nil> resps=[][]uint8{[]uint8(nil), []uint8(nil)}
```

Driven through the real path with a `net.Conn` whose `Write` lands the bytes and then reports an error:

```
this head:  successor channel closed by the old send's cleanup (open=false), still matchable=false
a983549c6:  successor channel still open - cleanup did not touch it,         still matchable=true
```

**The honest caveat is reachability.** I had to force the write error with a hooked `net.Conn` returning `(len(p), err)`; I could not produce a real `WriteMessage` failure after the peer had answered, because the client path buffers a small text frame and flushes it once in `flushFrame`, so it is all-or-nothing. Every other error path in `writeToConn` returns before the write and can only ever remove its own entry — including the barrier rejection this line was added for. Not gating.

thrasher- (approved):
Heimdall: Approving at [`5d5d3d5`](https://github.com/thrasher-corp/gocryptotrader/commit/5d5d3d5b1e14f99c40f32ef72a32c3ebec6c6a5b). The only code change since my last pass is the field comment at [`limit.go:51`](https://github.com/thrasher-corp/gocryptotrader/blob/5d5d3d5b1e14f99c40f32ef72a32c3ebec6c6a5b/exchanges/request/limit.go#L51), and it is exactly right — I replaced both production acquisitions of `r.m` and precisely the two tests it names failed, nothing else in the request or websocket suites. Every finding from the previous five rounds is now closed, including the PR body, which after three rounds of me raising it now describes reuse the way the code actually behaves.

The master merge is the part of this head worth checking, and it is clean. It brought `golang.org/x/time` 0.15.0 → 0.16.0 with it — the dependency this PR's whole correctness argument rests on — and all four files in `rate/` are byte-identical between the two versions, so `ReserveN`, `CancelAt` and `DelayFrom` are unchanged.

I re-ran the property that matters rather than trusting the earlier rounds. 30,720 randomised groups under `-race` — sizes 2–5, shared and distinct limiters, limiter-free legs, unrelated background traffic on the same limiters, random aborts, per-leg deadlines and parent cancellation — gave **zero splits**. Both round-2 blockers stay fixed: 0/500 groups admitted with a participant cancelled while parked, 0/200 on your deadline repro. Parking still holds no reservation, no admitted group overdrafts its limiters, and nothing leaks across 300 resolved and 300 abandoned groups.

The check I had not done in earlier rounds is against a real exchange's limiter topology, and it is the one that settled it for me. Binance builds `spotOrderLimiter` and `usdMarginedFuturesOrdersLimiter` as separate `NewRateLimit` instances and gives `spotOrderRate` and `uFuturesOrdersDefaultRate` weight 1, so a paired spot/futures order meets your preconditions exactly — 300/300 admitted. The two shapes your doc comment warns about both refuse cleanly: two spot orders sharing one limiter 0/300, a weight-10 leg 0/300, zero splits in either. Where it cannot admit, it fails safe.

One comment inline, Minor and not gating: the new `RemoveSignature` on the send-failure path can remove a *successor's* registration. I have the mechanism measured and A/B'd against master, but I could not trigger it on a real transport, so treat the reachability as unproven — it reads to me as an issue against `Match` rather than a change here.

CI is 16/16 green. Locally `go test ./... -race` is 161 packages ok with no data races; the two Bitstamp failures reproduce identically on [`a983549`](https://github.com/thrasher-corp/gocryptotrader/commit/a983549c6bed925e6887ae3bada993405b3954ea) with none of your code present (`json: cannot unmarshal string into Go value of type int64` — their ticker started quoting a number). `gofumpt -l` is clean on all eight changed files and `make misc_checks` passes.

Nothing here blocks.

PR #2365, opened 2026-09-04
types/decimal: add IsInteger support

shazbert:
# PR Description

Adds `IsInteger` support to the udecimal-backed `decimal.Decimal`
implementation.

The default Shopspring backend already exposes this method through its type
alias. Adding the equivalent method to the udecimal implementation keeps the
facade API consistent across both build configurations.

`IsInteger` returns true when a value has no fractional component, including
zero and negative integers.

No additional dependencies are required.

## Type of change

## How has this been tested

Tested the complete decimal package with race detection under both backends:

- [x] `go test ./types/decimal -count=1 -race`
- [x] `go test ./types/decimal -count=1 -race -tags=udecimal_on`
- [x] `golangci-lint run ./types/decimal/...`

Added matching backend tests covering:

- Positive integers
- Negative integers
- Zero
- Positive fractions
- Negative fractions

## Checklist

- [x] I have added tests that prove the feature works
- [x] Relevant existing unit tests pass locally with both decimal backends

### DIFF
--- a/types/decimal/decimal_shopspring_test.go
+++ b/types/decimal/decimal_shopspring_test.go
@@ -53,3 +53,24 @@ func TestMustFromString(t *testing.T) {
 		"MustFromString should preserve shopspring parsing")
 	assert.Panics(t, func() { MustFromString("invalid") }, "MustFromString should panic for invalid input")
 }
+
+func TestDecimalIsInteger(t *testing.T) {
+	t.Parallel()
+	for _, tc := range []struct {
+		name     string
+		value    string
+		expected bool
+	}{
+		{name: "positive integer", value: "42", expected: true},
+		{name: "negative integer", value: "-42.000", expected: true},
+		{name: "zero", value: "0", expected: true},
+		{name: "positive fraction", value: "42.1", expected: false},
+		{name: "negative fraction", value: "-0.1", expected: false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tc.expected, MustFromString(tc.value).IsInteger(),
+				"IsInteger should identify values without a fractional component")
+		})
+	}
+}
--- a/types/decimal/decimal_udecimal.go
+++ b/types/decimal/decimal_udecimal.go
@@ -159,6 +159,11 @@ func (d Decimal) IsNegative() bool {
 	return d.value.IsNeg()
 }
 
+// IsInteger reports whether d has no fractional component.
+func (d Decimal) IsInteger() bool {
+	return d.Equal(d.Truncate(0))
+}
+
 // Round rounds d half away from zero to places fractional digits. Negative
 // places round digits in the integer component.
 func (d Decimal) Round(places int32) Decimal {
--- a/types/decimal/decimal_udecimal_test.go
+++ b/types/decimal/decimal_udecimal_test.go
@@ -218,6 +218,27 @@ func TestDecimalIsNegative(t *testing.T) {
 	assert.False(t, Zero.IsNegative(), "IsNegative should reject zero")
 }
 
+func TestDecimalIsInteger(t *testing.T) {
+	t.Parallel()
+	for _, tc := range []struct {
+		name     string
+		value    string
+		expected bool
+	}{
+		{name: "positive integer", value: "42", expected: true},
+		{name: "negative integer", value: "-42.000", expected: true},
+		{name: "zero", value: "0", expected: true},
+		{name: "positive fraction", value: "42.1", expected: false},
+		{name: "negative fraction", value: "-0.1", expected: false},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			assert.Equal(t, tc.expected, MustFromString(tc.value).IsInteger(),
+				"IsInteger should identify values without a fractional component")
+		})
+	}
+}
+
 func TestDecimalRound(t *testing.T) {
 	t.Parallel()
 	largeHalf := "5" + strings.Repeat("0", maxStringDigits)


thrasher-:
@codex please review

types/decimal/decimal_udecimal_test.go
```diff
@@ -218,6 +218,27 @@ func TestDecimalIsNegative(t *testing.T) {
 	assert.False(t, Zero.IsNegative(), "IsNegative should reject zero")
 }
 
+func TestDecimalIsInteger(t *testing.T) {
```
thrasher-: Heimdall: **Should fix** — every case here is built with `MustFromString`, which normalises the fractional part away on this backend, so no case actually has a stored non-zero scale. The result is a test that cannot fail for the most tempting wrong implementation.

`normalise` strips trailing zeros before udecimal ever constructs the value — udecimal itself would have kept them:

```
udecimal.Parse("-42.000") directly   -> prec=3
facade normalise("-42.000")          -> "-42"
facade MustFromString("-42.000")     -> prec=0
```

So the "negative integer" case is indistinguishable from `-42`, and swapping the body of [decimal_udecimal.go:164](https://github.com/thrasher-corp/gocryptotrader/blob/d55769c403f97842e9676c327abb6bd87d5e7824/types/decimal/decimal_udecimal.go#L164) for `return d.value.Prec() == 0` still passes:

```
return true                    KILLED
d.Equal(d.Truncate(1))         KILLED
!d.Equal(d.Truncate(0))        KILLED
d.value.Prec() == 0            SURVIVES   <-- wrong, but green
```

That matters because `Prec() == 0` is wrong precisely where this repo lives — values decoded from an exchange response keep their scale, because `UnmarshalJSON` takes udecimal's native path and skips `normalise`:

```
json.Unmarshal(`"-42.000"`)             -> scale=3   IsInteger=true   Prec()==0 would say false
json.Unmarshal(`"1.000000000000000000"`)-> scale=18  IsInteger=true   Prec()==0 would say false
MustFromString("-42.000")               -> scale=0   IsInteger=true   Prec()==0 would say true
```

Arithmetic results behave the same way: `0.5+0.5` -> scale 1, `1.5*2` -> scale 1, `2.50*4` -> scale 1, `3/1.5` -> scale 19. All are integers; all would be reported as non-integers by the naive check.

Nothing is broken today — I ran 20,000 randomly generated values (1–45 integer digits, 0–19 fractional, half with all-zero fractions) through both backends and the answers hash identically, `sha256 f71b8d1f…`, 10,599 integers each side. The implementation is right. The test just doesn't hold it in place.

**Fix:** add cases whose scale survives construction, and one past u128. In [decimal_udecimal_test.go:221](https://github.com/thrasher-corp/gocryptotrader/blob/d55769c403f97842e9676c327abb6bd87d5e7824/types/decimal/decimal_udecimal_test.go#L221), taking `value Decimal` instead of `value string`:

```go
{name: "computed integer", value: MustFromString("0.5").Add(MustFromString("0.5")), expected: true},
{name: "computed integer at max scale", value: MustFromString("3").Div(MustFromString("1.5")), expected: true},
{name: "beyond u128", value: MustFromString(strings.Repeat("9", 40) + "." + strings.Repeat("0", 19)), expected: true},
{name: "beyond u128 fraction", value: MustFromString(strings.Repeat("9", 40) + ".5"), expected: false},
```

The last two also close a coverage gap — the current cases never reach udecimal's big.Int fallback in `Trunc`, which is the branch `IsInteger` relies on for large values. I ran the four added cases (plus swapping `zero` to the package `Zero`, matching `TestDecimalIsZero` and its neighbours) and the whole thing lands:

```
                       Trunc coverage   Prec()==0 mutant
PR as-is                    66.7%           SURVIVES
with the four cases        100.0%           KILLED
                                            --- FAIL: computed_integer
                                            --- FAIL: computed_integer_at_max_scale
```

`TestDecimalRound` already uses `maxStringDigits` and `TestDecimalStringFixed` already uses `math.MinInt32`, so probing those boundaries is consistent with the file rather than new ceremony.

types/decimal/decimal_shopspring_test.go
```diff
@@ -53,3 +53,24 @@ func TestMustFromString(t *testing.T) {
 		"MustFromString should preserve shopspring parsing")
 	assert.Panics(t, func() { MustFromString("invalid") }, "MustFromString should panic for invalid input")
 }
+
+func TestDecimalIsInteger(t *testing.T) {
```
thrasher-: Heimdall: **Minor** — this test body is byte-identical to the one added at [decimal_udecimal_test.go:221](https://github.com/thrasher-corp/gocryptotrader/blob/d55769c403f97842e9676c327abb6bd87d5e7824/types/decimal/decimal_udecimal_test.go#L221). It only needs to exist once.

`decimal_benchmark_test.go` carries no build tag, so a tag-free test file compiles into both configurations. I moved the test into `types/decimal/decimal_test.go`, deleted both copies, and got the same 5 subtests running in each build with identical coverage:

```
                     PR as-is    single tag-free file
udecimal_on          97.6%       97.6%     (5/5 subtests PASS)
shopspring default  100.0%      100.0%     (5/5 subtests PASS)
```

**Fix:** keep one copy in a build-tag-free file; drop the duplicate. 21 lines and one future divergence saved.

Worth noting the copies aren't equivalent even though the source is: under shopspring `"-42.000"` keeps exponent -3 and does exercise the fractional-digit scan, while under udecimal `normalise` flattens it to scale 0 first. If you'd rather keep them separate for that reason, that's a fair call — but then the udecimal copy needs different values, which is the point of the other comment.

types/decimal/decimal_udecimal.go
```diff
@@ -159,6 +159,11 @@ func (d Decimal) IsNegative() bool {
 	return d.value.IsNeg()
 }
 
+// IsInteger reports whether d has no fractional component.
+func (d Decimal) IsInteger() bool {
```
thrasher-, line 163: Heimdall: **Minor** — nothing stops this happening again. There is no compile-time check that the two backends expose the same method set, which is why `IsInteger` could go missing here in the first place while the shopspring build stayed green.

A tag-free assertion catches it at `go vet` time. I added one listing the facade's method set and it compiles clean under both backends today; removing this method to simulate the pre-PR state fails the udecimal build immediately:

```
$ go vet -tags=udecimal_on ./types/decimal/
vet: types/decimal/parity_test.go:49:23: cannot use Decimal{} (value of struct type Decimal)
     as decimalFacade value in variable declaration: Decimal does not implement
     decimalFacade (missing method IsInteger)

$ go vet ./types/decimal/          # shopspring, unaffected — the alias supplies it
$
```

**Fix:** optional here, but a `types/decimal/parity_test.go` with no build tag:

```go
type decimalFacade interface {
	Add(Decimal) Decimal
	// ... the rest of the value-receiver surface
	IsInteger() bool
}

type decimalPtrFacade interface {
	UnmarshalJSON([]byte) error
	UnmarshalText([]byte) error
	UnmarshalBinary([]byte) error
	Scan(any) error
}

var (
	_ decimalFacade    = Decimal{}
	_ decimalPtrFacade = (*Decimal)(nil)
)
```

The existing `var _ *Decimal = (*shopspring.Decimal)(nil)` at [decimal_shopspring_test.go:14](https://github.com/thrasher-corp/gocryptotrader/blob/d55769c403f97842e9676c327abb6bd87d5e7824/types/decimal/decimal_shopspring_test.go#L14) only pins the alias; it says nothing about the udecimal side. Happy for this to be a follow-up rather than part of this PR.

types/decimal/decimal_udecimal.go
```diff
@@ -159,6 +159,11 @@ func (d Decimal) IsNegative() bool {
...
+	return d.Equal(d.Truncate(0))
```
thrasher-: Heimdall: **Nit** — a stored scale of zero already means integral, so short-circuiting on it skips the truncate-and-compare entirely:

```go
return d.value.Prec() == 0 || d.Equal(d.Truncate(0))
```

That is the common case — everything from `MustFromString`, `NewFromInt` and `MustFromFloat` arrives with scale 0, because `normalise` trims the trailing zeros — and it brings the udecimal backend level with shopspring instead of 5x behind it:

```
                            current    with fast path    shopspring
42                          12.87 ns       2.47 ns         2.69 ns
42.1                        17.51 ns      17.00 ns        68.05 ns
3/1.5 (scale 19)            20.45 ns      17.10 ns       293.80 ns
40 nines + ".5" (big.Int)  586.60 ns     558.60 ns       116.20 ns
```

Same answers — I re-ran the 20,000-value corpus and the 48 edge cases against both forms and the digests are unchanged (`sha256 f71b8d1f…`), with `-race` green.

Note this is the *correct* form of the shortcut. `Prec() == 0` on its own, without the `||`, is wrong — see the other comment.

thrasher- (approved):
Heimdall: Approving. `d.Equal(d.Truncate(0))` is correct. For values parsed from the same decimal string both backends agree on 20,000 generated values (1–45 integer digits, 0–19 fractional) plus 48 edge cases — identical SHA-256 — covering `Trunc`'s big.Int fallback, the 19-digit precision boundary, negative zero and the uninitialised `Decimal{}`. `-race` passes under all four tag combinations; `golangci-lint` is clean in both.

Four comments on the lines, evidence on each: test power on the udecimal case list, the duplicated test body, a compile-time parity guard, and a one-line fast path. None gate the merge.

All three red checks are ambient:

- **docker** — four yobit tests, `read: connection reset by peer` against live `yobit.net`.
- **`sonic=true, udecimal=true`** — a data race on `os.Stdin` between `config/config_encryption_test.go:427` and `config/config.go:1490`; no frame in `types/decimal`, and `./config/...` passes locally with the same tags. Worth its own issue.
- **codecov/project** — −0.10% aggregate; `codecov/patch` green, every added line covered.

types/decimal/decimal_udecimal_test.go
gloriousCode: This looks worthwhile

shazbert: Yeah agreed, added it to https://github.com/thrasher-corp/gocryptotrader/pull/2365/commits/7ad175ab8b133837d61076840854c3784069d458

thrasher- (approved):
Heimdall: re-checked at `7ad175ab` — all four points from the last pass are in, and verified by
running them rather than reading them.

The one that mattered: `return d.value.Prec() == 0` survived the old test and dies against the new
one. Every other wrong mutant dies too; only the equivalent `d.Equal(d.Truncate(0))` survives. Both
backends still agree exactly — 20,000 generated values plus the `UnmarshalJSON` path and the
uninitialised `Decimal{}`, identical hashes — all four tag combinations pass under `-race`, and the
fast path takes the common scale-0 case from 11.2 ns to 1.9 ns.

Worth knowing but not worth changing: the parity guard pins the 32 methods udecimal has, so none can
go missing again, but shopspring exposes 33 more — reaching for one of those still breaks only under
`udecimal_on`, where the CI job catches it.

Approving.

types/decimal/parity_test.go
```diff
@@ -0,0 +1,50 @@
...
+	UnmarshalJSON([]byte) error
+	UnmarshalText([]byte) error
+	UnmarshalBinary([]byte) error
+	Scan(any) error
+}
+
+var (
+	_ decimalFacade    = Decimal{}
+	_ decimalPtrFacade = (*Decimal)(nil)
+)
```
gloriousCode: I think if I were to suggest anything it would be:

1) rn file to decimal_contract_test.go - parity implies same results which we have been explicit is not the case between implementations.
2) rn interface to decimalAPI
3) change things to this:

```go
import (
	"database/sql"
	"database/sql/driver"
	"encoding"
	"encoding/json" //nolint:depguard // Compile-time interface checks only; the custom JSON package does not export these interfaces.
)

// decimalAPI checks the shared method signatures under either backend build.
type decimalAPI interface {
	Add(Decimal) Decimal
	Sub(Decimal) Decimal
	Mul(Decimal) Decimal
	Div(Decimal) Decimal
	Mod(Decimal) Decimal
	Pow(Decimal) Decimal
	Abs() Decimal
	Neg() Decimal
	Cmp(Decimal) int
	Compare(Decimal) int
	Equal(Decimal) bool
	GreaterThan(Decimal) bool
	GreaterThanOrEqual(Decimal) bool
	LessThan(Decimal) bool
	LessThanOrEqual(Decimal) bool
	IsZero() bool
	IsPositive() bool
	IsNegative() bool
	IsInteger() bool
	Round(int32) Decimal
	Truncate(int32) Decimal
	Floor() Decimal
	Ceil() Decimal
	String() string
	StringFixed(int32) string
	Float64() (float64, bool)
	InexactFloat64() float64
	IntPart() int64
}

var (
	_ decimalAPI                 = Decimal{}
	_ driver.Valuer              = Decimal{}
	_ encoding.BinaryMarshaler   = Decimal{}
	_ encoding.BinaryUnmarshaler = (*Decimal)(nil)
	_ encoding.TextMarshaler     = Decimal{}
	_ encoding.TextUnmarshaler   = (*Decimal)(nil)
	_ json.Marshaler             = Decimal{}
	_ json.Unmarshaler           = (*Decimal)(nil)
	_ sql.Scanner                = (*Decimal)(nil)
)

```

gloriousCode (changes requested):
I think only point 1 is important to me, the rest are optional

gloriousCode (approved):
yayayayayACk

thrasher- (approved):
Heimdall: Approving at [`9ead593e`](https://github.com/thrasher-corp/gocryptotrader/commit/9ead593ebdcc3cacbee6ea731a9b5c36425b54d0).

The contract rewrite loses nothing. Enumerating both method sets by reflection: `decimalAPI` plus the six standard-library interfaces pin 36 methods, the udecimal `*Decimal` method set is 36, and the difference is empty in both directions — exactly the surface the deleted `parity_test.go` pinned. It still bites, too: deleting `IsInteger` fails `go vet -tags=udecimal_on` at [decimal_contract_test.go:43](https://github.com/thrasher-corp/gocryptotrader/blob/9ead593ebdcc3cacbee6ea731a9b5c36425b54d0/types/decimal/decimal_contract_test.go#L43), deleting `MarshalJSON` at [decimal_contract_test.go:49](https://github.com/thrasher-corp/gocryptotrader/blob/9ead593ebdcc3cacbee6ea731a9b5c36425b54d0/types/decimal/decimal_contract_test.go#L49) via `json.Marshaler`, and `Round(int32)`→`Round(int64)` gives `wrong type for method Round` — all while the shopspring build stays clean.

The predicate is right, and the fast path changes no answers. 20,045 values across `NewFromString`, `UnmarshalJSON` and `UnmarshalText` hash identically with and without the `Prec() == 0` short-circuit, and identically across the two backends — `sha256 865b2cac…`, 10,736 integers each side. Every wrong mutant dies against the new tests (`true`, `false`, `Prec() == 0`, `Equal(Truncate(1))`, `!Equal(Truncate(0))`); only the correct form survives. [decimal_udecimal.go:166](https://github.com/thrasher-corp/gocryptotrader/blob/9ead593ebdcc3cacbee6ea731a9b5c36425b54d0/types/decimal/decimal_udecimal.go#L166) and `Truncate` both reach 100% coverage, all four tag combinations pass under `-race`, and the short-circuit takes the common scale-0 case from 13.8 ns to 2.6 ns.

The 44 `IsInteger` differences I found between the backends are all values the two sides genuinely hold differently — `Div` rounds at 16 places under shopspring against udecimal's 19 — and `IsZero`, `Equal` and `Cmp` already diverge on exactly those values. Nothing new is exposed.

No comments this round; nothing in the diff needs changing.

PR #2341, opened 2026-08-21
build/linters: Bump Go to v1.27 and golangci-lint to v2.13.2

thrasher-:
Bumps Go to v1.27, golangci-lint to v2.13.2 and sonic to v1.15.4, and moves the JSON wrapper to `encoding/json/v2`, with `v1Compat` pinning the 17 v1 behaviours the codebase relies on. v2 decodes a Gate.io order 1.6x faster than v1 on a fifth of the allocations.

Also fixes wire tags and ticker volume units across exchanges, deribit's and kraken's websocket tickers, tickers missing or misreporting their last price, symlink escapes in `archive.Zip` and `common.ChangePermission`, and a data race in `cmd/config_builder`, and adopts Go 1.27 idioms (`new(expr)`, `httptest.NewTestServer`, `testing/synctest`, `os.OpenRoot`).

Fixes #2338

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Breaking for library consumers: exchange types change field names, numeric types and nested types; `ticker.Price.Volume` becomes `BaseVolume`; `encoding/json`'s `Marshal` and friends become functions in the default build; and `common/convert` drops seven helpers, including `FloatFromString` and `BoolPtr`.

## How has this been tested

- [x] `go test ./... -race`, default and `sonic_on`
- [x] `golangci-lint run`, default and `udecimal_on`
- [x] `make misc_checks`, gofumpt, `go vet` and codespell
- [x] Regression tests for each fixed decode path, checked against live payloads where the endpoint is public

Follow-ups, all pre-existing: binance and binanceus drop `EXPIRED_IN_MATCH` order updates, gemini's REST ticker leaves `Last` at 0, and deribit's `quote`, kucoin's futures `tickerV2` and huobi's detail channels overwrite ticker fields they don't carry.

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented my code, particularly in hard-to-understand areas
- [x] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally and on Github Actions with my changes
- [x] Any dependent changes have been merged and published in downstream modules

PR #2183, opened 2026-02-18
Exchanges: Support available pairs, scope bulk order cancellation, localise ticker processing on WS

gloriousCode:
Local review-fix commit `a60288bd6` is prepared but has not been pushed. This description includes those pending changes.

Allows explicit operations on available exchange assets and pairs without requiring them to be enabled for background processing, while retaining enabled-only defaults for automatic fan-out and subscriptions.

### Changes
- Update exchange wrappers and symbol matching across the integrations to validate explicit requests against available assets/pairs.
- Preserve enabled-only background subscriptions, omitted-argument fan-out and asset-specific bulk-query exceptions, including Gate.io options, OKX spreads and Yobit request-size limits.
- Accept recognised available-pair websocket data; preserve intentional spot/margin duplication only for enabled assets. Correct Coinbase alias handling, KuCoin topic-specific matching and OKX incoming asset resolution.
- Move ticker processing into exchange handlers, add batch ticker processing and simplify engine websocket handling. Bithumb skips missing bulk ticker entries without losing later results; Gate.io iterates enabled standard-margin assets without allocating a slice per message.
- Standardise CancelAllOrders responses as pointers with the Add helper, propagate the contract through exchange interfaces and callers, and remove obsolete ClassificationError handling.
- Add explicit cancellation scope controls through CLI/RPC and regenerate the associated protobuf/Swagger definitions. Require a pair for unsafe fan-out paths; Kraken cancels only matching order IDs rather than invoking account-wide websocket cancellation.
- Remove dead cancellation-response guards while retaining meaningful partial-result handling.
- Correct OKX cancellation filtering to honour the requested order/client ID and side, include instrument IDs, and omit empty request entries.
- Retain accompanying engine shutdown/race fixes and expand regression coverage for available-but-disabled assets/pairs, ticker delivery and scoped cancellation.
- Merge upstream master at bdb48ad03243c68244521b58fc490fe7ad2a2f2d and retain upstream remaining-amount tests during conflict resolution.

### Latest review fixes
- Keep inbound symbol resolution within enabled assets, including Coinbase ticker aliases and OKX derivatives; retain available-pair handling within that scope.
- Report disabled KuCoin assets explicitly and allow Bitfinex margin/funding subscriptions when spot is disabled.
- Preserve partial BTCMarkets/Lbank cancellation results; require both supplied OKX order identifiers to match; remove the obsolete Kraken account-wide websocket cancellation path.
[...]

PR #2368, opened 2026-09-07
exchanges/GateIO: Add mark and index prices for derivatives

gloriousCode:
Adds mark and index prices to Gate.io futures, delivery futures and options tickers over REST and WebSocket, with mocked mapping tests and live API assertions.

Corrects the contract-statistics path in `GetOpenInterest` to return `open_interest_usd` rather than the raw contract count. A regression test checks that the latest statistics entry supplies the USD notional. This does not address the separate existing differences between statistics and contract endpoint calculations.

Fixes the orderbook snapshot test's unregistered WebSocket connection and waits for background resubscription to succeed, so a connection-not-found error can no longer pass silently.

### DIFF
--- a/exchanges/gateio/gateio_test.go
+++ b/exchanges/gateio/gateio_test.go
@@ -346,8 +346,17 @@ func TestUpdateTicker(t *testing.T) {
 	t.Run("all supported assets", func(t *testing.T) {
 		t.Parallel()
 		for _, a := range e.GetAssetTypes(false) {
-			_, err := e.UpdateTicker(t.Context(), getPair(t, a), a)
-			assert.NoErrorf(t, err, "UpdateTicker should not error for %s", a)
+			t.Run(a.String(), func(t *testing.T) {
+				t.Parallel()
+				got, err := e.UpdateTicker(t.Context(), getPair(t, a), a)
+				require.NoError(t, err, "UpdateTicker must not error")
+				switch a {
+				case asset.USDTMarginedFutures, asset.CoinMarginedFutures, asset.DeliveryFutures:
+					require.NotNil(t, got, "live ticker must not be nil")
+					assert.Positive(t, got.MarkPrice, "live ticker mark price should be positive")
+					assert.Positive(t, got.IndexPrice, "live ticker index price should be positive")
+				}
+			})
 		}
 	})
 
@@ -358,19 +367,19 @@ func TestUpdateTicker(t *testing.T) {
 		expectedPath string
 	}{
 		{
-			name:         "USDT margined futures mark and index prices",
+			name:         "mocked USDT margined futures mark and index prices",
 			asset:        asset.USDTMarginedFutures,
 			pair:         currency.NewPairWithDelimiter("BTC", "USDT", currency.UnderscoreDelimiter),
 			expectedPath: "/api/v4/futures/usdt/tickers",
 		},
 		{
-			name:         "coin margined futures mark and index prices",
+			name:         "mocked coin margined futures mark and index prices",
 			asset:        asset.CoinMarginedFutures,
 			pair:         currency.NewPairWithDelimiter("BTC", "USD", currency.UnderscoreDelimiter),
 			expectedPath: "/api/v4/futures/btc/tickers",
 		},
 		{
-			name:         "delivery futures mark and index prices",
+			name:         "mocked delivery futures mark and index prices",
 			asset:        asset.DeliveryFutures,
 			pair:         currency.NewPairWithDelimiter("BTC", "USDT_20261225", currency.UnderscoreDelimiter),
 			expectedPath: "/api/v4/delivery/usdt/tickers",
@@ -2163,8 +2172,25 @@ func TestUpdateTickers(t *testing.T) {
 	t.Run("all supported assets", func(t *testing.T) {
 		t.Parallel()
 		for _, a := range e.GetAssetTypes(false) {
-			err := e.UpdateTickers(t.Context(), a)
-			assert.NoErrorf(t, err, "UpdateTickers should not error for %s", a)
+			t.Run(a.String(), func(t *testing.T) {
+				t.Parallel()
+				switch a {
+				case asset.USDTMarginedFutures, asset.CoinMarginedFutures, asset.DeliveryFutures:
+					pair := getPair(t, a)
+					ex := new(Exchange)
+					require.NoError(t, testexch.Setup(ex), "Setup must not error")
+					// Isolate the cache so another test cannot supply the ticker being checked.
+					ex.Name = t.Name()
+					require.NoError(t, ex.UpdateTickers(t.Context(), a), "live UpdateTickers must not error")
+					got, err := ticker.GetTicker(ex.Name, pair, a)
+					require.NoError(t, err, "live UpdateTickers must cache the requested ticker")
+					require.NotNil(t, got, "live ticker must not be nil")
+					assert.Positive(t, got.MarkPrice, "live ticker mark price should be positive")
+					assert.Positive(t, got.IndexPrice, "live ticker index price should be positive")
+				default:
+					assert.NoError(t, e.UpdateTickers(t.Context(), a), "UpdateTickers should not error")
+				}
+			})
 		}
 	})
 
@@ -2175,19 +2201,19 @@ func TestUpdateTickers(t *testing.T) {
 		expectedPath string
 	}{
 		{
-			name:         "USDT margined futures mark and index prices",
+			name:         "mocked USDT margined futures mark and index prices",
 			asset:        asset.USDTMarginedFutures,
 			pair:         currency.NewPairWithDelimiter("BTC", "USDT", currency.UnderscoreDelimiter),
 			expectedPath: "/api/v4/futures/usdt/tickers",
 		},
 		{
-			name:         "coin margined futures mark and index prices",
+			name:         "mocked coin margined futures mark and index prices",
 			asset:        asset.CoinMarginedFutures,
 			pair:         currency.NewPairWithDelimiter("BTC", "USD", currency.UnderscoreDelimiter),
 			expectedPath: "/api/v4/futures/btc/tickers",
 		},
 		{
-			name:         "delivery futures mark and index prices",
+			name:         "mocked delivery futures mark and index prices",
 			asset:        asset.DeliveryFutures,
 			pair:         currency.NewPairWithDelimiter("BTC", "USDT_20261225", currency.UnderscoreDelimiter),
 			expectedPath: "/api/v4/delivery/usdt/tickers",
--- a/exchanges/gateio/gateio_test.go
+++ b/exchanges/gateio/gateio_test.go
@@ -351,7 +351,7 @@ func TestUpdateTicker(t *testing.T) {
 				got, err := e.UpdateTicker(t.Context(), getPair(t, a), a)
 				require.NoError(t, err, "UpdateTicker must not error")
 				switch a {
-				case asset.USDTMarginedFutures, asset.CoinMarginedFutures, asset.DeliveryFutures:
+				case asset.USDTMarginedFutures, asset.CoinMarginedFutures, asset.DeliveryFutures, asset.Options:
 					require.NotNil(t, got, "live ticker must not be nil")
 					assert.Positive(t, got.MarkPrice, "live ticker mark price should be positive")
 					assert.Positive(t, got.IndexPrice, "live ticker index price should be positive")
@@ -384,6 +384,12 @@ func TestUpdateTicker(t *testing.T) {
 			pair:         currency.NewPairWithDelimiter("BTC", "USDT_20261225", currency.UnderscoreDelimiter),
 			expectedPath: "/api/v4/delivery/usdt/tickers",
 		},
+		{
+			name:         "mocked options mark and index prices",
+			asset:        asset.Options,
+			pair:         currency.NewPairWithDelimiter("BTC", "USDT-20261225-100000-C", currency.UnderscoreDelimiter),
+			expectedPath: "/api/v4/options/tickers",
+		},
 	} {
 		t.Run(tc.name, func(t *testing.T) {
 			t.Parallel()
@@ -391,10 +397,19 @@ func TestUpdateTicker(t *testing.T) {
 			ex := new(Exchange)
 			require.NoError(t, testexch.Setup(ex), "Setup must not error")
 			ex.Name = t.Name()
+			if tc.asset == asset.Options {
+				require.NoError(t, ex.UpdatePairs(currency.Pairs{tc.pair}, tc.asset, true), "mocked options pair must be enabled")
+			}
 
 			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 				assert.Equal(t, http.MethodGet, r.Method)
 				assert.Equal(t, tc.expectedPath, r.URL.Path)
+				if tc.asset == asset.Options {
+					assert.Equal(t, "BTC_USDT", r.URL.Query().Get("underlying"), "options underlying should match")
+					_, err := fmt.Fprintf(w, `[{"name":%q,"last_price":"118.4","mark_price":"118.35","index_price":"100000.25"}]`, tc.pair.String())
+					assert.NoError(t, err, "mocked response should be written")
+					return
+				}
 				assert.Equal(t, tc.pair.String(), r.URL.Query().Get("contract"))
 				_, err := fmt.Fprintf(w, `[{"contract":%q,"last":"118.4","low_24h":"99.2","high_24h":"132.5","volume_24h_base":"5526","volume_24h_quote":"1665006","mark_price":"118.35","index_price":"118.36"}]`, tc.pair.String())
 				assert.NoError(t, err)
@@ -407,7 +422,11 @@ func TestUpdateTicker(t *testing.T) {
 			got, err := ex.UpdateTicker(t.Context(), tc.pair, tc.asset)
 			require.NoError(t, err)
 			assert.Equal(t, 118.35, got.MarkPrice)
-			assert.Equal(t, 118.36, got.IndexPrice)
+			if tc.asset == asset.Options {
+				assert.Equal(t, 100000.25, got.IndexPrice, "options index price should match")
+			} else {
+				assert.Equal(t, 118.36, got.IndexPrice)
+			}
 		})
 	}
 }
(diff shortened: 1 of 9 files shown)


thrasher-:
@codex please review

exchanges/gateio/gateio_test.go
```diff
@@ -338,9 +342,92 @@ func TestGetOrderInfo(t *testing.T) {
...
+		t.Parallel()
+		for _, a := range e.GetAssetTypes(false) {
+			t.Run(a.String(), func(t *testing.T) {
+				t.Parallel()
+				got, err := e.UpdateTicker(t.Context(), getPair(t, a), a)
+				require.NoError(t, err, "UpdateTicker must not error")
+				switch a {
+				case asset.USDTMarginedFutures, asset.CoinMarginedFutures, asset.DeliveryFutures, asset.Options:
+					require.NotNil(t, got, "live ticker must not be nil")
+					assert.Positive(t, got.MarkPrice, "live ticker mark price should be positive")
```
thrasher-: Heimdall: **Blocker** — a positive mark price is not a valid invariant for options, and this is why CI is red on two jobs.

**Fix:** drop `MarkPrice` from the positivity check for `asset.Options` at [gateio_test.go:356](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L356) — either split the `switch` so only the three futures assets assert a positive mark, or relax options to `assert.GreaterOrEqual(t, got.MarkPrice, 0.0, ...)`. The `IndexPrice` assertion is fine as-is and should stay.

**Why:** Gate legitimately returns `"mark_price":"0"` for out-of-the-money options approaching expiry. Live counts at the time of writing:

```
                        contracts   mark_price==0   index_price==0
options BTC_USDT              636              17                0
options ETH_USDT              600              15                0
futures/usdt                  983               0                0
futures/btc                     1               0                0
delivery/usdt                  30               0                0
```

Every one of the BTC zeros was a contract expiring that day. [getPair](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L3954) returns `enabledPairs[0]` without screening the mark price, and the enabled options contract rotates as contracts expire — the CI log for this run even shows `Checked and updated enabled pairs [OPTIONS] - Removed: 1`. So which contract gets asserted on is arbitrary and changes daily.

Calling the real `UpdateTicker` against a live zero-mark contract reproduces the failure exactly, with no API error involved:

```
LIVE BTC_USDT-20260907-86000-C -> MarkPrice=0 IndexPrice=79688.18 Last=0
    Error:    "0" is not positive
    Messages: live ticker mark price should be positive
--- FAIL
```

That is character-for-character the failure in [job 101586531395](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34070395844/job/101586531395) and [job 101586531648](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34070395844/job/101586531648). It passes locally only when the currently-enabled contract happens to have a live mark.

Worth stressing that you lose no coverage by dropping it. Deleting every `MarkPrice:` and `IndexPrice:` line from the three production files fails all eight mocked subtests, offline:

```
--- FAIL: TestUpdateTicker/mocked_{options,delivery,coin_margined,USDT_margined}_mark_and_index_prices
--- FAIL: TestUpdateTickers/mocked_{options,delivery,coin_margined,USDT_margined}_mark_and_index_prices
```

The mocks already pin mark and index for all four assets including options. The live assertion adds nothing except a dependency on market conditions.

exchanges/gateio/gateio_test.go
```diff
@@ -2100,9 +2187,104 @@ func TestFetchTradablePairs(t *testing.T) {
...
+					if a == asset.Options {
+						testexch.UpdatePairsOnce(t, ex)
+					}
+					// Isolate the cache so another test cannot supply the ticker being checked.
+					ex.Name = t.Name()
+					require.NoError(t, ex.UpdateTickers(t.Context(), a), "live UpdateTickers must not error")
+					got, err := ticker.GetTicker(ex.Name, pair, a)
+					require.NoError(t, err, "live UpdateTickers must cache the requested ticker")
+					require.NotNil(t, got, "live ticker must not be nil")
+					assert.Positive(t, got.MarkPrice, "live ticker mark price should be positive")
```
thrasher-: Heimdall: **Blocker** — same invalid options invariant as [gateio_test.go:356](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L356); this is the second of the two failing assertions.

**Fix:** apply the same change here at [gateio_test.go:2210](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L2210) — exclude `asset.Options` from the mark-price positivity check, keep `IndexPrice`.

**Why:** evidence is on the other comment. Both sites fail together in the same runs, on the same contract:

```
--- FAIL: TestUpdateTicker/all_supported_assets/options
        gateio_test.go:356: "0" is not positive
--- FAIL: TestUpdateTickers/all_supported_assets/options
        gateio_test.go:2210: "0" is not positive
```

Flagging separately so neither gets fixed in isolation.

exchanges/gateio/gateio_wrapper.go
```diff
@@ -2559,7 +2567,7 @@ func openInterestFromStats(stats []ContractStat) (float64, error) {
 			latest = stats[i]
 		}
 	}
-	return latest.OpenInterest.Float64(), nil
+	return latest.OpenInterestUSD.Float64(), nil
```
thrasher-, line 2570: Heimdall: **Should fix** — this is a real correctness fix, but it is not mentioned anywhere in the PR title or description, and it changes what `GetOpenInterest` returns.

**Fix:** call it out in the PR description. No code change needed for the switch itself — see below for the two things it exposes.

**Why:** the change at [gateio_wrapper.go:2570](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_wrapper.go#L2570) is correct. Both other implementations of the interface — [FuturesContract.openInterest](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_wrapper.go#L2540) and [DeliveryContract.openInterest](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_wrapper.go#L2554) — return a quote notional, so the stats path was previously returning a raw contract count into the same field. Measured through `GetOpenInterest` on `BTC_USDT`:

```
              one key (stats path)      two keys (contract path)
master           565,754,248 contracts     2,252,707,523 USD      <- incomparable units
this PR        4,504,608,871 USD           2,253,025,506 USD      <- units now agree
```

Two things worth knowing, neither of which blocks this PR:

**1. A ~2x scale gap survives.** `contract_stats` counts both sides of each position while `contracts.position_size` counts the long side only, so the same pair at the same instant still answers differently depending on whether the caller passed exactly one key (see [useOpenInterestStats](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_wrapper.go#L2573)). It is approximately rather than exactly 2x — the stats notional is derived from `mark_price` while the contracts path uses `index_price`, and the two endpoints sample at different times:

```
BTC_USDT  1.998x     ETH_USDT  2.000x     SOL_USDT  1.997x
```

**2. Pre-existing and much worse: inverse contracts.** `FuturesContract.openInterest()` skips the quanto multiplier when it is zero, and live `BTC_USD` has `quanto_multiplier: "0"`, so it multiplies a USD-denominated contract count by BTC's index price. Gate specifies 1 USD per `BTC_USD` contract. Real wrapper calls:

```
COIN-MARGINED BTC_USD one key  (stats path)     =          13,311,458
COIN-MARGINED BTC_USD two keys (contract path)  =     529,721,683,907
RATIO = 39,794x
```

We report $530bn of open interest on a market with $13m. That reproduces on master too, so it is out of scope here — but you are the one in this function, so flagging it. Happy to raise it as its own issue rather than let it ride on this PR.

exchanges/gateio/gateio_wrapper_test.go
```diff
@@ -86,16 +86,44 @@ func TestCancelAllOrders(t *testing.T) {
...
-	openInterest, err := openInterestFromStats([]ContractStat{
-		{Time: types.Time(time.Unix(100, 0)), OpenInterest: types.Number(2)},
-		{Time: types.Time(time.Unix(300, 0)), OpenInterest: types.Number(4)},
-		{Time: types.Time(time.Unix(200, 0)), OpenInterest: types.Number(3)},
-	})
-	require.NoError(t, err)
-	assert.Equal(t, 4.0, openInterest)
+			openInterest, err := openInterestFromStats(tc.stats)
+			if tc.errIs != nil {
+				require.ErrorIs(t, err, tc.errIs)
```
thrasher-: Heimdall: **Nit** — three assertions in the rewritten table lack messages.

**Fix:** add them at [gateio_wrapper_test.go:120](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_wrapper_test.go#L120), [gateio_wrapper_test.go:123](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_wrapper_test.go#L123) and [gateio_wrapper_test.go:124](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_wrapper_test.go#L124) — "must" for `require`, "should" for `assert`.

**Why:** the messages were absent in the code this replaced too, so it is carried over rather than introduced — but the block was rewritten here, which makes it the moment to add them. Codex flagged the same class at `gateio_test.go:410`; that one has since been fixed and every assertion in the new `gateio_test.go` blocks carries a message. These three were missed.

exchanges/gateio/gateio_test.go
```diff
@@ -2100,9 +2187,104 @@ func TestFetchTradablePairs(t *testing.T) {
...
+				}
+			})
+		}
+	})
+
+	for _, tc := range []struct {
+		name         string
+		asset        asset.Item
+		pair         currency.Pair
+		expectedPath string
```
thrasher-: Heimdall: **Nit** — this case table is byte-identical to the one in `TestUpdateTicker`.

**Fix:** hoist the table (and ideally the handler) to a package-level helper shared by both tests — [gateio_test.go:2223](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L2223) and [gateio_test.go:367](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L367).

**Why:** the two 27-line tables are identical, and the two `httptest` handlers differ only in the `contract` assertion and one volume field name. Deduplicating tests is the house default, and a shared table means the next asset added gets covered by both tests rather than one.

thrasher- (changes requested):
Heimdall: Plumbing is complete — all six `ticker.Price` sites that should carry mark and index do, the three spot sites are correctly untouched, and the REST options endpoint does return `index_price`, so the struct move is right. The mocks genuinely constrain it: stripping the production lines fails all eight mocked subtests. The orderbook change is a real regression test — it fails without the connection registration.

Five comments inline carry the evidence.

**To unblock:**

1. [gateio_test.go:356](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L356) — options mark price is legitimately `0` near expiry; drop it from the positivity check.
2. [gateio_test.go:2210](https://github.com/thrasher-corp/gocryptotrader/blob/ef1c8e1490945af38104a12ccf59d426e6e2194e/exchanges/gateio/gateio_test.go#L2210) — same fix, second site.

Those two are the current CI failures in [run 34070395844](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34070395844). The other three comments are suggestions and gate nothing.

Local `golangci-lint` noise in this package is ambient — identical on [`2690670`](https://github.com/thrasher-corp/gocryptotrader/commit/2690670f84f047c9fdebd9ac998bb084c5c52c17); CI lint passes.

Happy to approve once those two are in.

exchanges/gateio/gateio_websocket_test.go
```diff
@@ -228,7 +228,10 @@ func TestProcessOrderbookUpdateWithSnapshot(t *testing.T) {
 	subs, err := e.Features.Subscriptions.ExpandTemplates(e)
 	require.NoError(t, err)
 
-	conn := &FixtureConnection{}
+	baseConn, err := e.Websocket.CreateTestConnection(asset.Spot)
```
thrasher-, line 231: Heimdall: **Should fix** — registering the connection is the right call, but it silently drops the only coverage of the resubscribe-failure cleanup path.

**Fix:** add a failing-resubscribe case to `TestResubscribe` in `ws_ob_resub_manager_test.go`. That file isn't touched by this PR, so this can't be a one-click suggestion:

```go
	// An untracked connection makes the background resubscribe fail, which must still clear the entry
	err = m.Resubscribe(t.Context(), e, &FixtureConnection{}, qualifiedChannel, currency.NewBTCUSDT(), asset.Spot)
	require.NoError(t, err, "Resubscribe must not error when dispatching to an untracked connection")
	assert.Eventually(t,
		func() bool {
			return !m.IsResubscribing(currency.NewBTCUSDT(), asset.Spot)
		},
		time.Second,
		10*time.Millisecond,
		"a failed resubscribe should clear the resubscribing entry",
	)
```

**Why:** on master the unregistered connection made `ResubscribeToChannel` fail, which incidentally exercised [`ws_ob_resub_manager.go:50-51`](https://github.com/thrasher-corp/gocryptotrader/blob/29093b4d8ce35c44b001ff4eccfacb151d8cc53d/exchanges/gateio/ws_ob_resub_manager.go#L50) — the `CompletedResubscribe` call whose own comment reads "Ensure we clear the map entry on failure too". Registering the connection makes the resubscribe succeed, so nothing reaches that branch any more, and [`TestResubscribe`](https://github.com/thrasher-corp/gocryptotrader/blob/29093b4d8ce35c44b001ff4eccfacb151d8cc53d/exchanges/gateio/ws_ob_resub_manager_test.go#L85) only ever covered the success path. Package coverage profile, master vs this PR:

```
Resubscribe                        100.0%  ->  90.5%
ws_ob_resub_manager.go:50.4,52.1   master: 1   PR: 0
```

A/B: reverting just the registration puts that count back to 1, so the fix is the cause. That branch is what stops a pair being stuck in the resubscribing map forever after a failed resubscribe. With the case above the package returns to 47.0%, matching master; I ran it under `-race` along with `misc_checks`.

exchanges/gateio/gateio_wrapper.go
thrasher-, line 2570: Heimdall: **Minor** — pre-existing and nothing to do with this PR; flagging it here only because this is the line that makes the two open-interest paths comparable, and on coin-margined they still aren't.

**Fix:** none in this PR. It belongs on `FuturesContract.openInterest`.

**Why:** Gate reports `quanto_multiplier: "0"` for the BTC-settled `BTC_USD` contract, and [`gateio_wrapper.go:2542`](https://github.com/thrasher-corp/gocryptotrader/blob/29093b4d8ce35c44b001ff4eccfacb151d8cc53d/exchanges/gateio/gateio_wrapper.go#L2542) skips the multiply when it is zero, leaving a bare `position_size * index_price`. End-to-end through `GetOpenInterest`:

```
coin-margined ONE key  (stats path)    ->          13,269,654
coin-margined TWO keys (contract path) -> 525,510,449,012      <- 39,602x
```

$525B of claimed open interest against $13.3M actual, on a contract Gate specifies as 1 USD. Master returns the same two numbers, so your change is exactly neutral here — for coin-margined `open_interest_usd` equals `open_interest`, inverse contracts being $1 each.

Worth knowing for whoever picks it up: [#2341](https://github.com/thrasher-corp/gocryptotrader/pull/2341) already solves the identical problem on the volume side, where `futuresBaseVolume` uses a `max()` precisely because "coin-margined reports 0, which zeroes the product". The same reasoning hasn't been carried over to open interest, where the guard drops the multiply instead.

thrasher- (approved):
Heimdall: Both blockers are fixed, and load-bearing rather than incidentally green — pinned against a live zero-mark option (`BTC_USDT-20260908-87000-C`, mark 0, index 79293.33) the old assertion fails and the new one passes. The table hoist, the assertion messages and the description update all landed. Stripping the 12 mark/index assignments fails every mocked subtest, so the mapping is genuinely constrained, and Gate's SDK confirms [`open_interest_usd`](https://github.com/gateio/gateapi-go/blob/master/docs/ContractStat.md) is a quote-currency notional against `open_interest` in contracts.

One ask, inline: the orderbook fix drops the last coverage of the resubscribe-failure cleanup branch, which was only ever reached by the bug this PR fixes.

Also flagged inline, purely so it doesn't get lost: `GetOpenInterest` is ~39,600x out on coin-margined. It reproduces identically on master, so it is neither yours nor this PR's to carry.

Full `-race` suite and `misc_checks` pass, local lint noise is identical on master, and CI is green on [all 16 checks](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34177369048).

Approving.

exchanges/gateio/gateio_websocket_futures_test.go
```diff
@@ -5,14 +5,42 @@ import (
...
+
+	ex := new(Exchange)
+	require.NoError(t, testexch.Setup(ex), "Setup must not error")
+
+	payload := []byte(`{"time":1541659086,"channel":"futures.tickers","event":"update","result":[{"contract":"BTC_USDT","last":"118.4","mark_price":"118.35","index_price":"118.36","volume_24h_quote":"1665006","volume_24h_base":"5526","low_24h":"99.2","high_24h":"132.5"}]}`)
+	require.NoError(t, ex.processFuturesTickers(t.Context(), payload, asset.USDTMarginedFutures), "futures ticker processing must succeed")
+
+	select {
+	case msg := <-ex.Websocket.DataHandler.C:
+		got, ok := msg.Data.([]ticker.Price)
```
shazbert, line 31: Just noticed we are not processing the ticker into the ticker package like depth its in websocket routine manager, if we are using the library as an upstream module then engine is bypassed for these updates, maybe this can be an issue to enforce for websocket so its standardised to accounts and orderbook processing; for the tests specifically REST and websocket test that all return values also need to have a secondary package lookup so that it can be accessed and that the values match.

shazbert (approved):
Tested ACK. One potential issue on standardisation not blocking.

exchanges/gateio/gateio_websocket_futures_test.go
gloriousCode: I tried to change this ages ago https://github.com/thrasher-corp/gocryptotrader/pull/2183/changes#diff-847984cc298bdb52084df4fc0e517ec0edcefe6131e245d40556fa1f97c4aaedR124 in #2183 comments welcome

shazbert: @gloriousCode I forgot about that! Any chance we can pull that specific change out into its own PR? Before that I can create an issue for it, with some wants/needs. Or is it best to keep it consolidated into that PR and focus on merging that?

gloriousCode: I will aim for Heimdall approval of that PR,  then await your thoughts before doing additional splitting and making things

shazbert: 👍 ack

exchanges/gateio/gateio_wrapper.go
thrasher-: Fixed in #2341

PR #2369, opened 2026-09-07
yobit: ignore ticker response metadata

shazbert:
# PR Description

Fixes Yobit bulk ticker parsing when the API includes top-level response metadata alongside ticker entries.

Yobit may return:

- `success`, containing the request status.
- `error`, describing unavailable pairs within a bulk request.
- Valid pair-specific ticker objects.

The parser now ignores the known metadata fields while continuing to decode ticker entries. Unexpected malformed ticker entries still return an error rather than being silently discarded.

The live single-pair ticker test now skips when Yobit does not return the requested market.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Testing

- `go test ./exchanges/yobit -race -count=1`
- `go test -tags sonic_on ./exchanges/yobit -count=1`
- `go vet ./exchanges/yobit`
- `golangci-lint run --new-from-rev=upstream/master ./exchanges/yobit/...`


thrasher-:
@codex please review

exchanges/yobit/yobit.go
```diff
@@ -53,14 +54,24 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
+		if pair == "success" || pair == "error" {
+			continue
+		}
+		var ticker Ticker
+		if err := json.Unmarshal(entry, &ticker); err != nil {
+			return nil, fmt.Errorf("error decoding ticker for %s: %w", pair, err)
+		}
+		result[pair] = ticker
+	}
+	return result, nil
```
thrasher-, line 74: Heimdall: **Blocker** — skipping `error` unconditionally turns a rejected request into an empty success, and `UpdateTicker` then serves a stale price.

**Fix:**

```suggestion
	result := make(map[string]Ticker, len(raw))
	var apiError string
	for pair, entry := range raw {
		switch pair {
		case "success":
			continue
		case "error":
			if err := json.Unmarshal(entry, &apiError); err != nil {
				return nil, fmt.Errorf("error decoding error field: %w", err)
			}
			continue
		}
		var ticker Ticker
		if err := json.Unmarshal(entry, &ticker); err != nil {
			return nil, fmt.Errorf("error decoding ticker for %s: %w", pair, err)
		}
		result[pair] = ticker
	}
	// A reported error is only metadata about unavailable pairs when usable entries came back;
	// with none it is the request failing, and must not be reported as an empty success.
	if len(result) == 0 && apiError != "" {
		return nil, fmt.Errorf("%w: %s", errTickerRequestFailed, apiError)
	}
	return result, nil
```

Needs `var errTickerRequestFailed = errors.New("ticker request failed")` at package level, so it is not a clean one-click apply.

**Why:** on `{"success":0,"error":"Invalid pair name: btc_usd"}`, against a control on master:

```
new:    GetTicker -> 0 entries, err=<nil>    UpdateTickers -> nil
master: GetTicker -> error                   UpdateTickers -> error
```

The consequence is worse than a silent no-op. `UpdateTicker` falls back to the ticker cache, so after one good poll and then a failure:

```
first poll    -> last=80001
after failure -> last=80001, err=<nil>     <- cached price served as current
```

A consumer gets a plausible price of unbounded age with nothing to signal it. With the patch: partial responses still return their entries, `success:0` errors with the sentinel, and the stale read becomes `ticker=nil, err=ticker request failed: Invalid pair name: btc_usd`. Both of your new tests still pass under the default and `sonic_on` backends, with `go vet` and `gofumpt` clean.

Caveat I cannot close: yobit.net is unreachable from my machine, so that payload is constructed rather than captured. If the endpoint cannot produce it, this is defensive only.

exchanges/yobit/yobit_test.go
```diff
@@ -504,6 +544,9 @@ func TestGetHistoricTrades(t *testing.T) {
 func TestUpdateTicker(t *testing.T) {
 	t.Parallel()
 	_, err := e.UpdateTicker(t.Context(), testPair, asset.Spot)
+	if errors.Is(err, ticker.ErrTickerNotFound) {
+		t.Skipf("UpdateTicker should skip when pair %s is unavailable in live ticker feed", testPair)
+	}
```
thrasher-, line 549: Heimdall: **Should fix** — this skip is broader than the flake it targets, and hides the failure above.

**Fix:** skip on the transport failing, not on the ticker being absent — or assert the pair is genuinely delisted before skipping.

**Why:** `UpdateTicker` against a `success:0` response returns `ticker.ErrTickerNotFound`, which is exactly this condition. So after this PR the test can no longer fail for the parsing bug class the PR is about. Network failures still fail it — I confirmed that locally, where every live yobit test times out rather than skipping — so the skip is not inert, just too wide.

thrasher- (changes requested):
Heimdall: reviewed at `71e81969`. The fix works for what it targets — a partial response now parses
where master errors on it.

One blocker and one should-fix, both inline.

The blocker rests on something I could not check: yobit.net is unreachable from here, so my failure
payload is constructed from the shapes the code implies, not captured from the venue. If Yobit never
returns `success:0` on this endpoint, say so and I will drop it.

**To unblock:** propagate the API error when no usable entries came back ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3946046008))

Happy to approve once that is in, or once you confirm the payload cannot occur.

exchanges/yobit/yobit.go
```diff
@@ -53,14 +56,34 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
 // GetTicker returns a ticker for a specific currency
 func (e *Exchange) GetTicker(ctx context.Context, symbol string) (map[string]Ticker, error) {
-	type Response struct {
-		Data map[string]Ticker
-	}
-
-	response := Response{}
+	var raw map[string]json.RawMessage
 	path := fmt.Sprintf("/%s/%s/%s", apiPublicVersion, publicTicker, symbol)
+	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &raw); err != nil {
```
thrasher-, line 64: Heimdall: **Should fix** — one delisted pair still takes down the whole Yobit ticker feed. `ignore_invalid=1` is what fixes that, and it is what would make this PR's title true.

**Fix:** [yobit.go:59-61](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit.go#L59-L61) — send the parameter, and drop the `fmt.Sprintf` while you are in here:

```suggestion
	var raw map[string]json.RawMessage
	// ignore_invalid drops unknown or delisted pairs instead of rejecting the whole batch
	path := "/" + apiPublicVersion + "/" + publicTicker + "/" + symbol + "?ignore_invalid=1"
	if err := e.SendHTTPRequest(ctx, exchange.RestSpot, path, &raw); err != nil {
```

**Why:** yobit.net is reachable from here today, which closes the caveat I left on the blocker — I can capture the payload rather than construct it. `bcc_btc` is in `config_example.json`'s available list and gone from live `/info`; it is the exact delisted pair your fixture names:

```
$ curl -s ".../api/3/ticker/ltc_btc-eth_btc-btc_usd-dash_btc-bcc_btc"
{"success":0,"error":"Pair is off: bcc_btc"}

$ curl -s ".../api/3/ticker/ltc_btc-eth_btc-btc_usd-dash_btc-bcc_btc?ignore_invalid=1"
-> keys: ['btc_usd', 'dash_btc', 'eth_btc', 'ltc_btc']
```

One dead pair rejects the entire batch. And `error` never arrives *alongside* ticker entries: I sampled 1800 live pairs in ten 180-pair windows across `/info` and not one response carried a `success` or `error` key. So `{"success":1,"error":"Pair is off: bcc_btc","btc_usd":{…}}` in `TestGetTickerIgnoresMetadata` ([yobit_test.go:74](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit_test.go#L74)) is a shape the venue does not produce, and "`error`, describing unavailable pairs within a bulk request" in the PR description does not hold as written.

`UpdateTickers`, live, with those five pairs enabled:

```
master:                     err=json: cannot unmarshal number into ...yobit.Ticker   0 of 5 tickers
this PR:                    err=ticker request failed: Pair is off: bcc_btc          0 of 5 tickers
this PR + ignore_invalid=1: err=<nil>                                                4 of 5 tickers
```

So the PR turns an opaque failure into a legible one, which is worth having on its own — but the feed is still down for every pair because one of them is dead. `go test ./exchanges/yobit/... -count 1` stays green against the live venue with the parameter applied.

**Do not take it on its own, though.** `UpdateTickers` silently continues past pairs missing from the result ([yobit_wrapper.go:162-164](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit_wrapper.go#L162-L164)), and `ignore_invalid=1` makes a missing pair the normal case rather than an unreachable one. Cache two pairs, then serve a response that drops one:

```
first poll   -> eth_btc last=0.031
second poll  -> UpdateTickers=nil, eth_btc last=0.031   <- stale, indefinitely, with no error
```

That trades a loud outage for a silent stale price on the dead pair, which is the failure mode the blocker was about. `LastUpdated` still holds the earlier poll's time, so it is detectable, but nothing in the return signals it — worth logging the pairs the venue did not return.

The all-dead case does not need anything extra, though. With this and the guard on [yobit.go:83](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit.go#L83) both applied, live:

```
enabled pairs        UpdateTickers                                     tickers
valid + delisted     err=<nil>                                         4 of 5
all delisted         err=ticker request failed: Empty pair list        0 of 2
all valid            err=<nil>                                         2 of 2
```

Yobit answers an entirely invalid list with `{"success":0,"error":"Empty pair list"}`, so the sentinel still fires when nothing usable came back. The two changes compose; the suite is green under both JSON backends with both applied.

Same envelope hits the neighbours if you want them in scope: `GetDepth("bcc_btc")` → `json: cannot unmarshal number into Go value of type yobit.Orderbook`, `GetTrades("bcc_btc")` → `…[]yobit.Trade`. Both pre-existing, neither has a sentinel.

Fine as a follow-up if you would rather keep this PR to the parser — it does not gate.

exchanges/yobit/yobit.go
```diff
@@ -53,14 +56,34 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
+		var ticker Ticker
+		if err := json.Unmarshal(entry, &ticker); err != nil {
+			return nil, fmt.Errorf("error decoding ticker for %s: %w", pair, err)
+		}
+		result[pair] = ticker
+	}
+	if len(result) == 0 && apiError != "" {
+		return nil, fmt.Errorf("%w: %s", errTickerRequestFailed, apiError)
+	}
+	return result, nil
```
thrasher-, line 102: Heimdall: **Minor**, and it is my own suggestion I am correcting — the guard shipped here is the one I gave you on [the blocker thread](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3946046008), applied verbatim. It keys off the error string rather than off `success`, so four failure envelopes still return with a nil error: two as an empty success, two publishing a zero price over the cache.

**Fix:** decode `success` and fail on `0`, regardless of what `error` holds:

```suggestion
	result := make(map[string]Ticker, len(raw))
	var apiError string
	var failed bool
	for pair, entry := range raw {
		switch pair {
		case "success":
			var success uint8
			if err := json.Unmarshal(entry, &success); err != nil {
				return nil, fmt.Errorf("error decoding ticker success field: %w", err)
			}
			failed = success == 0
			continue
		case "error":
			if err := json.Unmarshal(entry, &apiError); err != nil {
				return nil, fmt.Errorf("error decoding ticker error field: %w", err)
			}
			continue
		}
		var ticker Ticker
		if err := json.Unmarshal(entry, &ticker); err != nil {
			return nil, fmt.Errorf("error decoding ticker for %s: %w", pair, err)
		}
		result[pair] = ticker
	}
	if failed {
		if apiError == "" {
			return nil, errTickerRequestFailed
		}
		return nil, fmt.Errorf("%w: %s", errTickerRequestFailed, apiError)
	}
	return result, nil
```

**Why:** poll once with good data, flip the handler, read back through `UpdateTicker`:

```
second response                                     master      this PR           with the above
{"success":0,"error":"Invalid pair name: btc_usd"}   error       error             error
{"success":0}                                        error       last=80001, nil   error
{"success":0,"error":""}                             error       last=80001, nil   error
{"success":0,"error":"Pair is off: x","x":null}      error       last=0,     nil   error
{"success":0,"error":"Pair is off: x","x":{}}        error       last=0,     nil   error
{}                                                   nil         nil               nil
```

Rows 2-5 are behaviour changes against master, all in the direction of the blocker you just closed. Rows 4 and 5 are the worse pair: `null` and `{}` decode to a zero-valued `Ticker`, `len(result)` becomes 1, the guard is suppressed, and `ProcessTicker` publishes `last=0` over the cached price. Row 6 is at parity with master; leaving it alone is correct.

`success` is the right key because the venue only emits it on failure — across every live response I captured today, a successful one carried no `success` and no `error` at all. Your `{"success":1,…}` fixture still passes, as do the other three tests and the live suite.

Caveat I cannot close: I could not get any of rows 2-5 out of the venue. Every failure envelope I captured carried a non-empty `error` string and no entries, so this is defensive only — say so and I will drop it.

exchanges/yobit/yobit_test.go
```diff
@@ -69,6 +71,76 @@ func TestGetTicker(t *testing.T) {
 	assert.NoError(t, err, "GetTicker should not error")
 }
 
+func TestGetTickerIgnoresMetadata(t *testing.T) {
```
thrasher-: Heimdall: **Nit** — the four new tests repeat the same eight lines of server/client/endpoint setup and differ only in the response body and the assertion.

**Fix:** one table-driven test. [coinut_test.go:113-176](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/coinut/coinut_test.go#L113-L176) is the in-repo precedent.

```go
func TestGetTickerResponseHandling(t *testing.T) {
	t.Parallel()
	for _, tc := range []struct {
		name      string
		body      string
		want      map[string]Ticker
		errIs     error
		errString string
	}{
		{
			name: "metadata alongside entries is ignored",
			body: `{"success":1,"error":"Pair is off: bcc_btc","btc_usd":{"last":80001}}`,
			want: map[string]Ticker{"btc_usd": {Last: 80001}},
		},
		{
			name:      "malformed ticker entry",
			body:      `{"btc_usd":1}`,
			errString: "error decoding ticker for btc_usd",
		},
		{
			name:  "failed request",
			body:  `{"success":0,"error":"Invalid pair name: btc_usd"}`,
			errIs: errTickerRequestFailed,
		},
		{
			name:      "malformed error metadata",
			body:      `{"success":0,"error":{}}`,
			errString: "error decoding ticker error field",
		},
	} {
			t.Parallel()
			ex := new(Exchange)
			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
				_, err := w.Write([]byte(tc.body))
				assert.NoError(t, err, "writing ticker response should not error")
			}))
			t.Cleanup(server.Close)
			require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL), "SetRunningURL must not error")

			result, err := ex.GetTicker(t.Context(), "btc_usd")
			switch {
			case tc.errIs != nil:
				assert.ErrorIs(t, err, tc.errIs, "GetTicker should return the expected error")
			case tc.errString != "":
				assert.ErrorContains(t, err, tc.errString, "GetTicker should return the expected error")
			default:
				require.NoError(t, err, "GetTicker must not error")
				assert.Equal(t, tc.want, result, "GetTicker should return only ticker entries")
			}
		})
	}
}
```

**Why:** same four cases, 14 fewer lines, `gofmt` clean, all four subtests pass, and the `want` map asserts the whole entry rather than just `Last`. Not worth a round trip on its own — fold it in if you touch the file again.

thrasher- (approved):
Heimdall: yobit.net is reachable from here today — it was not on the last pass — so here is an updated review grounded in captured payloads rather than constructed ones. That changes three things.

The caveat I left on the blocker is closed: `{"success":0,"error":"Invalid pair name: …"}` and `{"success":0,"error":"Pair is off: bcc_btc"}` both come back HTTP 200 from the live endpoint, so that finding was never defensive. The premise in the PR description does not survive the same treatment — `error` never arrives alongside ticker entries, in 1800 live pairs sampled across `/info`. And the venue has the parameter that actually fixes the failure this PR targets, which GCT never sends.

Both findings from [`71e8196`](https://github.com/thrasher-corp/gocryptotrader/commit/71e819691600cb24c2c654bd43d3e41940e53e78) are fixed at [`0c2c22e`](https://github.com/thrasher-corp/gocryptotrader/commit/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d) and verified, and I am resolving those threads — one with a correction, because the guard you applied is the one I gave you and it is not tight enough.

The sentinel propagates end to end: cache a price, then serve the rejection, and `UpdateTicker` returns `ticker request failed: …` where it previously handed back the cached `80001` with a nil error. The over-broad `ErrTickerNotFound` skip is gone and the test-file diff is now purely additive — strip the four added tests and it is byte-identical to master's. The regression test earns its keep: delete the three-line guard and `TestGetTickerRejectsFailedRequest` fails and nothing else does. Clean on `go vet`, `golangci-lint` (0 issues), `-race -count 1`, `-count 2`, and `-tags sonic_on`; the malformed-error assertion holds under both JSON backends despite their different decoder text.

Three comments, none of them gates: `ignore_invalid=1` ([yobit.go:61](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit.go#L61)) is what that failure actually needs, and what it must be paired with; the correction to my own guard ([yobit.go:83](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit.go#L83)), a narrower residual of the same class as the blocker; a test dedup nit ([yobit_test.go:74](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit_test.go#L74)).

Take them or leave them — nothing here blocks the merge.

exchanges/yobit/yobit.go
thrasher-: Heimdall: Fixed — and the caveat I left on this is now closed. But the guard I handed you is not tight enough, so read the new comment on the line before treating it as settled.

The caveat first. yobit.net is reachable from here today, so the payload is captured rather than constructed:

```
$ curl -s ".../api/3/ticker/btc_usd-notarealpair_zzz"
{"success":0,"error":"Invalid pair name: notarealpair_zzz"}

$ curl -s ".../api/3/ticker/ltc_btc-eth_btc-btc_usd-dash_btc-bcc_btc"
{"success":0,"error":"Pair is off: bcc_btc"}
```

Both HTTP 200, and `bcc_btc` is a genuinely delisted pair still sitting in `config_example.json`. So this was never defensive, and `UpdateTicker` now returns `ticker request failed: …` where it previously served the cached `80001`.

**Correction to my own suggestion.** `len(result) == 0 && apiError != ""` is the guard I gave you, and it keys off the error string rather than off `success`. Four envelopes still return with a nil error: `{"success":0}`, `{"success":0,"error":""}`, and a `success:0` carrying a `null` or `{}` entry — the last two publishing `last=0` over the cached price, because the zero-valued `Ticker` makes `len(result)` non-zero and suppresses the guard. Master rejects all four. A tighter version and the measurements are on [yobit.go:83](https://github.com/thrasher-corp/gocryptotrader/blob/0c2c22eb9e10565a52e47e0e7a45b87a5a29a84d/exchanges/yobit/yobit.go#L83); none of it gates the merge.

The blocker itself is closed, so I am resolving this.

exchanges/yobit/yobit.go
```diff
@@ -51,16 +54,46 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
 }
 
-// GetTicker returns a ticker for a specific currency
+// GetTicker returns tickers for the requested symbols, omitting invalid pairs.
+// An explicit API failure returns an error even if ticker entries are present.
 func (e *Exchange) GetTicker(ctx context.Context, symbol string) (map[string]Ticker, error) {
-	type Response struct {
-		Data map[string]Ticker
+	var raw map[string]json.RawMessage
+	path := "/" + apiPublicVersion + "/" + publicTicker + "/" + symbol + "?ignore_invalid=1"
```
thrasher-, line 63: Heimdall: **Minor** — `?ignore_invalid=1` is the load-bearing half of this change and arrives with nothing saying so. The line came from the suggestion on [the earlier thread](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3947056831), which carried this comment above it; the comment did not survive the trip.

**Fix:**

```suggestion
	// ignore_invalid drops unknown or delisted pairs instead of rejecting the whole batch
	path := "/" + apiPublicVersion + "/" + publicTicker + "/" + symbol + "?ignore_invalid=1"
```

**Why:** the tests already stop it being deleted — strip the parameter and all 19 subtests fail — but nothing tells a reader what it buys. It is the vendor's own word for the behaviour, from [yobit.net/en/api](https://yobit.net/en/api/): "Note! If we disable any pair from the list API will throw an error message. To enable the ignoration of such errors GET-parameter is available `ignore_invalid`." Live today it is the whole difference:

```
$ curl -s ".../api/3/ticker/ltc_btc-eth_btc-btc_usd-dash_btc-bcc_btc"
{"success":0,"error":"Pair is off: bcc_btc"}

$ curl -s ".../api/3/ticker/ltc_btc-eth_btc-btc_usd-dash_btc-bcc_btc?ignore_invalid=1"
-> keys: ['btc_usd', 'dash_btc', 'eth_btc', 'ltc_btc']
```

At the wrapper level, with those same five pairs enabled:

```
master:   err=json: cannot unmarshal number into Go struct field .success of type yobit.Ticker   0 of 5 cached
this PR:  err=no ticker found: Yobit spot omitted pairs [BCC_BTC]                                4 of 5 cached
```

And the value is read as a number, not a flag, which is easy to get wrong later. Against `btc_usd-bcc_btc` today:

```
absent, =0, =true, =yes, =(empty)   -> {"success":0,"error":"Pair is off: bcc_btc"}
=1, =2, =01, =-1, =1.0              -> {"btc_usd": {...}}
```

So `ignore_invalid=true` silently does nothing. `1` is the documented spelling and the right one; worth a line saying why it is there.

Applied: `gofumpt`, `go vet` and `golangci-lint` clean, suite green under the default and `sonic_on udecimal_on` backends.

exchanges/yobit/yobit_wrapper.go
```diff
@@ -179,6 +182,9 @@ func (e *Exchange) UpdateTickers(ctx context.Context, a asset.Item) error {
 			return err
 		}
 	}
+	if len(missingPairs) != 0 {
+		return fmt.Errorf("%w: %s %s omitted pairs %s", ticker.ErrTickerNotFound, e.Name, a, missingPairs)
```
thrasher-: Heimdall: **Minor** — this is the logging I asked for on [the earlier thread](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3947056831) and better than what I asked for, since an error cannot be scrolled past. One consequence worth naming: it also fails `UpdateTicker` for every *other* pair, including ones that refreshed cleanly in the same call.

**Why:** live, with the delisted `BCC_BTC` enabled next to four healthy pairs:

```
UpdateTickers          -> no ticker found: Yobit spot omitted pairs [BCC_BTC]
  BTC_USD  cache: last=82820          <- fresh, from this very poll
  ETH_BTC  cache: last=0.03143014
  LTC_BTC  cache: last=0.00070551
  DASH_BTC cache: last=0.00080805
  BCC_BTC  cache: no ticker found
UpdateTicker(BTC_USD)  -> ticker=nil, err=no ticker found: ... omitted pairs [BCC_BTC]
```

Not a regression — master errors here too, with a worse message and 0 of 5 cached — so this is the remaining half of the win rather than something you broke. It reaches users through GCTScript's `exchange.ticker(...)`, which is `UpdateTicker` ([`gctscript/wrappers/gct/exchange/exchange.go:62`](https://github.com/thrasher-corp/gocryptotrader/blob/efb279ebf10c05bafe2c3a46cb02a07d183fa8c6/gctscript/wrappers/gct/exchange/exchange.go#L62)); the sync manager and the gRPC server both read `GetCachedTicker` and are unaffected. Pairs refresh at bootstrap or on `gctcli pair update Yobit`, not on a timer, so a mid-run delisting holds this open until one of those.

Through the real syncer the trade is clearly worth it. Driving `SyncManager.syncTicker` over five agents from a cold cache, master errors on all five and caches nothing; this branch caches four and errors on two — the dead pair, and whichever pair happened to trigger the batch, which is the `if err == nil` gate on [`engine/sync_manager.go:575-577`](https://github.com/thrasher-corp/gocryptotrader/blob/efb279ebf10c05bafe2c3a46cb02a07d183fa8c6/engine/sync_manager.go#L575-L577) and not yours. Warm the cache first and the dead pair reports `syncErrors=0` while serving an hour-old price, because that branch reads `GetCachedTicker` with no freshness check — true on master too, and no wrapper can fix it.

The return is pinned by exactly one test: remove it and `TestUpdateTickersResponseHandling/partial_batch` is the only thing that fails.

**Fix, if you want it in this PR.** Not a clean apply — it spans both methods:

```go
// UpdateTickers updates the ticker for all currency pairs of a given asset type
// Returned tickers are cached before reporting any omitted pairs with ticker.ErrTickerNotFound.
func (e *Exchange) UpdateTickers(ctx context.Context, a asset.Item) error {
	_, err := e.updateTickers(ctx, a)
	return err
}

// updateTickers caches every ticker the venue returned and reports the enabled pairs it omitted
func (e *Exchange) updateTickers(ctx context.Context, a asset.Item) (currency.Pairs, error) {
	// ... body unchanged, each bare `return err` becomes `return nil, err`
	if len(missingPairs) != 0 {
		return missingPairs, fmt.Errorf("%w: %s %s omitted pairs %s", ticker.ErrTickerNotFound, e.Name, a, missingPairs)
	}
	return nil, nil
}

// UpdateTicker updates and returns the ticker for a currency pair
func (e *Exchange) UpdateTicker(ctx context.Context, p currency.Pair, a asset.Item) (*ticker.Price, error) {
	missingPairs, err := e.updateTickers(ctx, a)
	if err != nil && !errors.Is(err, ticker.ErrTickerNotFound) {
		return nil, err
	}
	if missingPairs.Contains(p, true) {
		return nil, fmt.Errorf("%w: %s %s %s", ticker.ErrTickerNotFound, e.Name, a, p)
	}
	return ticker.GetTicker(e.Name, p, a)
}
```

The obvious shortcut — swallow `ErrTickerNotFound` in `UpdateTicker` and stop there — is wrong, and your own [`yobit_test.go:181`](https://github.com/thrasher-corp/gocryptotrader/blob/efb279ebf10c05bafe2c3a46cb02a07d183fa8c6/exchanges/yobit/yobit_test.go#L181) already says so. I tried it:

```
--- FAIL: TestUpdateTickersResponseHandling/partial_batch
    Expected error with "no ticker found" in chain but got nil.
    Expected nil, but got: &ticker.Price{Last:0.031, ...}
```

That is the blocker you just closed, returning through the pair that actually *was* omitted. The version above keeps that assertion passing and puts `UpdateTicker(BTC_USD)` back to `last=82820, err=<nil>`. gofumpt, vet and lint clean, `-race -count 2` and `sonic_on` green, no test changes needed — though a subtest for the healthy-pair-beside-an-omitted-one case would be worth adding with it.

One hole it does not close, unchanged either way: `UpdateTicker` for a pair that is no longer enabled still reads straight from the cache, since it appears in neither the batch nor the omitted list.

Equally fine as a no — failing the whole call is the conservative reading of `UpdateTicker`, and it does not gate.

exchanges/yobit/yobit.go
```diff
@@ -51,16 +54,46 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
+			}
+			continue
+		}
+		var ticker Ticker
+		if err := json.Unmarshal(entry, &ticker); err != nil {
+			return nil, fmt.Errorf("error decoding ticker for %s: %w", pair, err)
+		}
+		result[pair] = ticker
+	}
+	if failed || len(result) == 0 && apiError != "" {
```
thrasher-, line 96: Heimdall: **Minor**, and it is the last corner of the hazard from [the guard thread](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3947056859) rather than a new one. The fix I gave you there keys on `success`, so it closed the four `success:0` envelopes and left the same shapes open when the failure metadata arrives *without* `success:0`. A `null` or `{}` entry still publishes `last=0` over a good cached price.

**Why:** warm a cache to 80001, flip the response, read the ticker back:

```
second response                                       master                     this PR
{"error":"Pair is off: x","btc_usd":null}             error, cache keeps 80001   nil err, cache -> 0
{"error":"Pair is off: x","btc_usd":{}}               error, cache keeps 80001   nil err, cache -> 0
{"success":1,"btc_usd":null}                          error, cache keeps 80001   nil err, cache -> 0
{"error":"Pair is off: x","btc_usd":{"last":80002}}   error, cache keeps 80001   nil err, cache -> 80002   <- the intent
{"btc_usd":null}                                      nil err, cache -> 0        same
{"success":null}                                      nil err, cache keeps 80001 ticker request failed
```

Rows 1-3 are behaviour changes against master, in the wrong direction. `null` and `{}` decode into a zero-valued `Ticker` without error, so `len(result)` becomes 1 and both arms of the guard are suppressed. Master only rejected them by accident — the string `error` and the numeric `success` failed to decode *as tickers* — and shedding that accident is exactly what this PR set out to do, so this is a consequence of the design rather than a slip. Row 5 shows the zero-price overwrite itself predates you; row 6 shows the `success` arm working.

The arm you did apply is pinned: revert `failed` and 9 subtests fail, covering the four envelopes I flagged last round plus a fifth you found on your own, `success:0` arriving with a populated ticker.

**How reachable: as far as I can measure, not at all.** I have never got the venue to emit `success` or `error` beside ticker entries — 1800 pairs in ten windows last round, plus every `ignore_invalid` shape today — and I have never seen a `null` or `{}` ticker entry. Same caveat I gave you last round, and it still holds: this is defensive only.

**So a question rather than a fix.** `if failed || apiError != ""` closes all three rows, on the same reasoning that made `success` the right key: the venue only sends `error` when nothing usable came back. The cost is that it also rejects `{"success":1,"error":"...","btc_usd":{"last":80001}}`, which is your `defensive metadata alongside entries` case — a shape the venue does not produce either. Whichever way you go, the row-1 body is worth a table entry so the choice is written down.

Your call which unreachable shape you would rather be wrong about, and it does not gate. If you would rather leave the parser tolerant, say so and I will drop it.

thrasher- (approved):
Heimdall: This is the version I wanted — `ignore_invalid=1`, the `success` guard, the one-off tests folded into tables, and `UpdateTickers` now naming the pairs the venue dropped instead of walking past them, which is more than I asked for. Measurements are on the three comments below; the short version is 4 of 5 tickers cached where master cached none.

One note with nowhere to anchor: the PR description has drifted from the branch on two points, and it becomes the squash message. It says `error` describes "unavailable pairs within a bulk request" arriving beside valid ticker objects — measured against the venue today, those never come together. And it still says the live single-pair test skips when the market is missing, which was true of `71e8196` but not here; you removed the skip, which was the right answer to that comment.

`gofumpt`, `go vet`, `golangci-lint run --new-from-rev=upstream/master` and `make misc_checks` clean. Green: `./exchanges/yobit/... -race -count 3`, `./engine/... ./exchanges/ticker/... -race`, the live `exchange_wrapper_standards` run, and both `sonic_on` and `sonic_on udecimal_on`.

**Three minors, none gating:**
1. Say what `ignore_invalid=1` is for ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3953823501))
2. `UpdateTicker` still fails for a pair that did refresh ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3953823507))
3. The last corner of the zero-price guard ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3953823511))

Approving, and resolving the three threads I left open last round — `ignore_invalid`, the `success` guard and the table-driven nit are all satisfied by this head.

exchanges/yobit/yobit_test.go
```diff
@@ -69,6 +73,143 @@ func TestGetTicker(t *testing.T) {
...
+		{name: "malformed ticker", body: `{"btc_usd":1}`, errString: "error decoding ticker for btc_usd"},
+		{name: "failed request", body: `{"success":0,"error":"Invalid pair name: btc_usd"}`, errIs: errTickerRequestFailed, errString: "Invalid pair name: btc_usd"},
+		{name: "malformed error", body: `{"success":0,"error":{}}`, errString: "error decoding ticker error field"},
+		{name: "malformed success", body: `{"success":"invalid"}`, errString: "error decoding ticker success field"},
+		{name: "failure without error", body: `{"success":0}`, errIs: errTickerRequestFailed},
+		{name: "failure with empty error", body: `{"success":0,"error":""}`, errIs: errTickerRequestFailed},
+		{name: "failure with null ticker", body: `{"success":0,"error":"Pair is off: btc_usd","btc_usd":null}`, errIs: errTickerRequestFailed},
+		{name: "failure with empty ticker", body: `{"success":0,"error":"Pair is off: btc_usd","btc_usd":{}}`, errIs: errTickerRequestFailed},
+		{name: "failure with populated ticker", body: `{"success":0,"btc_usd":{"last":80001}}`, errIs: errTickerRequestFailed},
+		{name: "error without success", body: `{"error":"Pair is off: btc_usd"}`, errIs: errTickerRequestFailed},
```
thrasher-, line 96: Heimdall: **Minor** — the new empty-ticker guard swallowed the only two cases that pinned the `failed` arm, so the check that closed the original blocker is no longer protected by any test.

**Fix:**

```suggestion
		{name: "error without success", body: `{"error":"Pair is off: btc_usd"}`, errIs: errTickerRequestFailed},
		{name: "failure metadata alongside a populated ticker", body: `{"success":0,"error":"Pair is off: eth_btc","btc_usd":{"last":80001}}`, errIs: errTickerRequestFailed, errString: "Pair is off: eth_btc"},
```

**Why:** `{"success":0,"error":"...","btc_usd":null}` and the `{}` variant used to reach [`yobit.go:95`](https://github.com/thrasher-corp/gocryptotrader/blob/76419faada0c40efc38b1ca37bbb1d6dcafa1daf/exchanges/yobit/yobit.go#L95) with `len(result) == 1`, because the placeholder decoded into a zero-valued `Ticker`. [`yobit.go:90-92`](https://github.com/thrasher-corp/gocryptotrader/blob/76419faada0c40efc38b1ca37bbb1d6dcafa1daf/exchanges/yobit/yobit.go#L90-L92) now rejects them first, so nothing exercises "failure metadata arriving with a non-empty result" any more. `failure with populated ticker` does not cover it either — it carries `success:0` but no `error` string.

A/B against an incorrect parser that treats that combination as success:

```go
if (failed || len(result) == 0 && apiError != "") && !(failed && apiError != "" && len(result) > 0) {
```

```
efb279ebf  (previous head)     4 subtests fail
76419faad  (this head)        31 of 31 pass      <- nothing catches it
76419faad  + the fixture       1 subtest fails
```

Coverage moved the other way across that same change — package 61.5% → 62.0%, `GetTicker` 96.2% → 96.8% — which is what makes this one easy to miss.

Applied: passes on the unmutated parser, `gofumpt` clean, `./exchanges/yobit/... -race` green.

exchanges/yobit/yobit.go
```diff
@@ -51,16 +54,51 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
+			if err := json.Unmarshal(entry, &apiError); err != nil {
+				return nil, fmt.Errorf("error decoding ticker error field: %w", err)
+			}
+			continue
+		}
+		var ticker Ticker
+		if err := json.Unmarshal(entry, &ticker); err != nil {
+			return nil, fmt.Errorf("error decoding ticker for %s: %w", pair, err)
+		}
+		if ticker == (Ticker{}) {
```
thrasher-, line 91: Heimdall: **Note, not an ask** — this returns rather than skipping, so one placeholder entry costs every *other* pair its update.

**Why:** warm a cache, then serve one null beside a healthy pair carrying a *newer* price:

```
after good poll : btc last=80001
after null entry: UpdateTickers err=ticker request failed: empty ticker for eth_btc
                  btc last=80001          <- the 80002 in that same response is discarded
                  UpdateTicker(BTC_USD) -> result=nil err=ticker request failed: empty ticker for eth_btc
```

Nothing stale is served as fresh — the cached timestamp does not move — so it is safe, just blunt: a good price already in hand is thrown away, and the feed stalls for every pair until the venue stops sending the bad entry. `continue` would let `missingPairs` degrade it per-pair instead, at the cost of eight test expectations; I ran that and the other 23 subtests stay green.

**Not asking for it, because the trigger is not observed.** I swept every live pair this round rather than a slice — 9033 pairs, 37 batches of 250, with `ignore_invalid=1`: zero entries decoding to an empty ticker, zero pairs with `updated == 0`, zero pairs omitted, and zero responses carrying `success` or `error` beside entries. Rejecting a malformed entry is a coherent policy and you have documented it on the function, so I am recording the trade-off rather than disputing the choice.

Worth knowing either way: no test sends a placeholder *beside a healthy pair* — every placeholder case sends the bad entry alone — so whichever behaviour you keep is currently unpinned.

thrasher- (approved):
Heimdall: All three minors are in, and your `UpdateTicker` condition is tighter than the one I handed you. Verified live — with the delisted `BCC_BTC` enabled beside four healthy pairs, `UpdateTicker(BTC_USD)` returns `last=82410, err=<nil>` where it used to fail.

One Minor, and it is a test rather than the code ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3955007670)). One note I am not asking you to act on ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2369#discussion_r3955007687)).

`gofumpt`, `go vet`, `golangci-lint run --new-from-rev=upstream/master` and `make misc_checks` clean; `-race`, `-shuffle`, `./engine/... ./exchanges/ticker/...` and the live `exchange_wrapper_standards` run green. Sonic checked under Go 1.26.5 for the native decoder — on Go 1.27 the `sonic_on` tag falls back to `encoding/json`.

Approving.

thrasher- (approved):
Heimdall: Re-reviewed at `0bb8da9a`. Test-only since `76419faa` — the non-test code is
byte-identical, and both comments from last round are answered with fixtures.

Both fixture groups verified by mutation, not by reading:

```
                                                 76419faa tests    0bb8da9a tests
loosened success/error guard (yobit.go:95)        31 pass 0 fail    33 pass 2 fail
skip placeholder beside healthy (yobit.go:90)      0 fail            2 fail
```

The second is more than I asked for — I said I was recording that trade-off rather than
disputing it, and you pinned the behaviour instead of changing it.

Clean: `gofumpt`, `go vet`, `golangci-lint` (0 issues), `make misc_checks`. Green on
`-race -count 2`, `-shuffle=on`, `./engine/... ./exchanges/ticker/...`, the live
`exchange_wrapper_standards` run and both sonic legs. Live venue re-probed; assumptions
hold, and the `all invalid` fixture is a captured payload.

**Non-gating, unchanged since my last pass:** the description becomes the squash message
and is stale on three points — `error` never arrives beside ticker entries at the venue;
`TestGetTicker` has no skip at this head; and it never mentions `ignore_invalid=1`, the
load-bearing change.

Approving, and resolving the five threads I left open.

exchanges/yobit/yobit.go
thrasher-: Heimdall: Pinned by tests rather than a behaviour change — the outcome I wanted when I
said I was not asking for the `continue`.

Against a mutation that skips a placeholder only when a healthy entry coexists: 0 subtests
fail on `76419faa`'s tests, 2 fail on this head (`healthy_ticker_beside_null_entry`,
`healthy_ticker_beside_empty_entry`). Reject-don't-skip is the pinned contract now.

exchanges/yobit/yobit.go
```diff
@@ -51,16 +54,51 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
-	response := Response{}
-	path := fmt.Sprintf("/%s/%s/%s", apiPublicVersion, publicTicker, symbol)
-
-	return response.Data, e.SendHTTPRequest(ctx, exchange.RestSpot, path, &response.Data)
+	result := make(map[string]Ticker, len(raw))
+	var apiError string
+	var failed bool
+	for pair, entry := range raw {
+		switch pair {
+		case "success":
```
gloriousCode: `pair` is not very accurate here for what it is. Its a status

exchanges/yobit/yobit.go
```diff
@@ -51,16 +54,51 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
+			var success uint8
+			if err := json.Unmarshal(entry, &success); err != nil {
+				return nil, fmt.Errorf("error decoding ticker success field: %w", err)
+			}
+			failed = success == 0
+			continue
```
gloriousCode: Given the if statement below of `if failed` we should break/return here immediately since you have no tolerance for failures in this code block

exchanges/yobit/yobit.go
```diff
@@ -51,16 +54,51 @@ func (e *Exchange) GetInfo(ctx context.Context) (Info, error) {
...
+		case "error":
+			if err := json.Unmarshal(entry, &apiError); err != nil {
+				return nil, fmt.Errorf("error decoding ticker error field: %w", err)
+			}
+			continue
```
gloriousCode, line 85:
```suggestion
        // these cases only appear when there is an error
        // standard responses do not contain these fields
		case "success":
			var success uint8
			if err := json.Unmarshal(entry, &success); err != nil {
				return nil, fmt.Errorf("error decoding ticker success field: %w", err)
			}
			failed = success == 0
			continue
		case "error":
			if err := json.Unmarshal(entry, &apiError); err != nil {
				return nil, fmt.Errorf("error decoding ticker error field: %w", err)
			}
			continue
```

gloriousCode (changes requested):
Only one suggestion because I was tripped up looking only at code

gloriousCode:
I wrecked your lint sorry

gloriousCode (approved):
tACK!

PR #2384, opened 2026-09-14
engine/event_manager: continue checking events until stopped

ginavalent:
# PR Description

Fix the event manager so that it continues checking events on every ticker interval until shutdown.

The previous implementation executed one `select` only. After the first ticker event, `eventManager.run()` returned and the goroutine stopped checking events, while the `started` flag remained true.

This change:

- wraps the ticker select in a loop;
- stops the ticker when the goroutine exits;
- adds a regression test proving that an event can trigger on a later
  checking interval after its condition becomes satisfied.

## Type of change

## How has this been tested

- [x] `go test ./engine -race -count=1`
- [x] `go test ./engine -run TestEventManagerRunChecksEventsUntilStopped -race -count=10`
- [x] `gofumpt -d engine/event_manager.go engine/event_manager_test.go`
- [ ] `golangci-lint run`
- [ ] `go test ./... -race`

The new regression test uses local ticker data and does not require exchange
credentials or network access.

## Checklist


thrasher-:
@codex please review

engine/event_manager_test.go
```diff
@@ -146,6 +155,45 @@ func TestEventManagerStop(t *testing.T) {
...
+		sleepDelay = 10 * time.Millisecond
+		threshold  = 200
+	)
+	exchangeName := newUniqueFakeExchangeName()
+	eventCh := make(chan base.Event, 1)
+	m, err := setupEventManager(&eventChannelCommsManager{events: eventCh}, &testExchangeManager{validExchange: exchangeName}, sleepDelay, false)
+	require.NoError(t, err, "setupEventManager must not error")
+	m.events = []Event{newPriceEvent(exchangeName, threshold)}
+
+	seedTicker(t, exchangeName, currency.NewBTCUSD(), asset.Spot, 100, 99, 101)
```
thrasher-: Heimdall: **Blocker** — these two new calls fail the `lint` job. `seedTicker` now has five callers, all passing `asset.Spot`, so `unparam` reports its `a` parameter at [event_manager_test.go:62](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L62).

**Fix:** drop the `a asset.Item` parameter, set `AssetType: asset.Spot` inside `seedTicker`, and remove the argument from all five calls: [event_manager_test.go:171](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L171), [event_manager_test.go:186](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L186), [event_manager_test.go:292](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L292), [event_manager_test.go:457](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L457) and [event_manager_test.go:545](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L545).

**Why:** `unparam` only reports "always receives" once a function has four call sites ([check.go:655](https://github.com/mvdan/unparam/blob/5beb8c8f8f15f2d0c9ebe817478178baa18d543f/check/check.go#L655)). The base has three calls, and these two make five. With CI's linter (golangci-lint v2.9.0 on go1.26.8) over `./engine/...`:

```
this PR          engine/event_manager_test.go:62:72: seedTicker - a always receives asset.Spot (1) (unparam)
                 1 issues:
this PR + fix    0 issues.   (also 0 with --build-tags=udecimal_on)
```

[Lint job 103926203871](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34828552460/job/103926203871) failed on exactly this issue; the base, [`23a1042`](https://github.com/thrasher-corp/gocryptotrader/commit/23a104278c423db9d7b54464fae96273e5ccce51), passed ([lint job 103115022331](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34551445075/job/103115022331)).

engine/event_manager_test.go
```diff
@@ -146,6 +155,45 @@ func TestEventManagerStop(t *testing.T) {
...
+	startTestEventManager(t, m)
+	defer func() {
+		require.NoError(t, m.Stop(), "Stop must not error")
+	}()
+
+	require.Never(t, func() bool {
```
thrasher-: Heimdall: **Should fix** — this `require.Never` makes the test flaky, and it can let the test pass against the unfixed `run()`.

The 50ms window stands in for "the first tick has happened", and it breaks in two ways:

1. `Never` returns when its timer fires, even while a callback it started is still running ([assertions.go:2171-2179](https://github.com/stretchr/testify/blob/959dbdacf1533e155162811ea90c90117a420463/assert/assertions.go#L2171-L2179)). That callback drains `eventCh`. If it resumes after the manager pushes the event, the event is gone and the `Eventually` at [event_manager_test.go:187](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L187) never sees it, because an executed event is not pushed again.
2. Nothing orders the first check before the price moves to 300. If the worker's first tick lands after the window, the base's one-shot `run()` sees 300 on its only tick and the test passes.

**Fix:** replace the window with a second event that triggers on the first tick, listed after the pending one, so receiving it proves the pending event was checked at 100:

```go
	// Each tick checks the pending event before the one that triggers straight away.
	m.events = []Event{newPriceEvent(exchangeName, threshold), newPriceEvent(exchangeName, 50)}
```

```go
	select {
	case <-eventCh: // the first tick has checked the pending event against the initial price
	case <-time.After(time.Second):
		require.FailNow(t, "first tick must trigger the event below the initial ticker price")
	}
```

That also retires the `should` message flagged on [event_manager_test.go:184](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L184).

**Why:**

```
                                                  this test        two-event version
250 parallel copies alongside ./engine, -cpu 1    1 of 1,500 fail  0 of 1,500 fail
  failure: event_manager_test.go:187: Condition never satisfied
  every instrumented failure (5 of 5): event pushed, then drained by a Never callback that outlived Never
Never callback held in flight until the push      3/3 fail         n/a (no Never)
base run(), first tick 100ms late                 5/5 pass         5/5 fail
this PR, -race, run alone                         500/500 pass     200/200 pass
```

engine/event_manager.go
```diff
@@ -71,17 +71,21 @@ func (m *eventManager) Stop() error {
...
+	for {
+		select {
+		case <-m.shutdown:
+			return
+		case <-t.C:
+			total, executed := m.getEventCounter()
+			if total > 0 && executed != total {
+				m.m.Lock()
+				for i := range m.events {
+					m.executeEvent(i)
```
thrasher-, line 86: Heimdall: **Minor** — with `executeEvent` now running on every tick, each pending event logs its unmet condition every 500ms, where the one-shot `run()` logged it once.

**Fix:** return a sentinel from `shouldProcessEvent` ([event_manager.go:303](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L303)) and skip it in `executeEvent` ([event_manager.go:99](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L99)), so genuine failures still log:

```go
errEventConditionNotMet = errors.New("event condition not met")
```

```go
if err := m.checkEventCondition(&m.events[i]); err != nil {
	if !errors.Is(err, errEventConditionNotMet) {
		log.Debugf(log.EventMgr, "Events: Failed to check event condition: %v", err)
	}
	return
}
```

`TestShouldProcessEvent` can then assert it with `ErrorIs`.

**Why:** DEBUG is on in the default logging config ([logger_setup.go:57](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/log/logger_setup.go#L57)), and the delay defaults to 500ms ([event_manager_types.go:28](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_types.go#L28)). One event with an unmet price condition, over 5s with the default logger:

```
this PR          9 × [DEBUG] Events: Failed to check event condition: does not meet conditions
base             1
this PR + fix    0    (event-manager tests pass under -race)
```

engine/event_manager_test.go
```diff
@@ -146,6 +155,45 @@ func TestEventManagerStop(t *testing.T) {
 	assert.ErrorIs(t, err, ErrNilSubsystem, "Stop should return nil subsystem error")
 }
 
+func TestEventManagerRunChecksEventsUntilStopped(t *testing.T) {
```
thrasher-, line 167: Heimdall: **Minor** — `Stop` returns without waiting for `run` ([event_manager.go:61-70](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L61-L70)), so nothing can prove the "until stopped" half of this name. With `case <-m.shutdown:` changed from `return` to `continue`, so the worker spins after `Stop`, all 18 tests in this file still pass.

**Fix:** have `Stop` wait for the worker, as `DatabaseConnectionManager.Stop` does ([database_connection.go:141](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/database_connection.go#L141)):

```go
wg sync.WaitGroup // new eventManager field

m.wg.Go(m.run) // Start

close(m.shutdown) // Stop
m.wg.Wait()
```

`CommunicationManager.PushEvent` never blocks ([communication_manager.go:91-95](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/communication_manager.go#L91-L95)), so the wait is at most one pass over the events. If you'd rather leave `Stop` alone, name the test for what it proves.

**Why:**

```
               return → continue in run()                 Stop then Start, PushEvent mid-tick
this PR        18/18 pass                                 2 run workers; -race: event_manager.go:78 vs :47
this PR + wg   panic: test timed out after 20s            1 run worker; -race clean
               (deferred Stop, event_manager_test.go:174)
```

The right-hand column is the restart case from the comment on [event_manager.go:78](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L78). Nothing in the engine restarts a stopped instance (`SetSubsystem` has no event-manager case), so it's test-only, but the same change closes it.

Related: six existing tests start a manager and never stop it. With the loop, their workers now tick every 500ms for the rest of the package run: 6 are still running after the event-manager tests finish, against 0 on the base.

thrasher- (changes requested):
Heimdall: The fix is right. `run()` now loops the way its sibling subsystems do (the same shape as [ntp_manager.go:98-115](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/ntp_manager.go#L98-L115)), and the new test is a genuine regression test: against the base's one-shot `run()` it fails at [event_manager_test.go:187](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L187). Requesting changes only for lint.

Evidence is on the lines themselves.

**To unblock:**
1. The `unparam` failure from the two new `seedTicker` calls — [event_manager_test.go:171](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L171).

Not gating: the flaky `Never` window at [event_manager_test.go:177](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L177), per-tick debug logging for pending events at [event_manager.go:85](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L85), and `Stop` not waiting for `run` at [event_manager_test.go:158](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager_test.go#L158).

Out of scope and pre-existing, both hidden while `run` exited after its first tick. Take them here or I'll raise separate issues:
- `Add` derives IDs from `len(m.events)+1` ([event_manager.go:133](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L133)), so `Remove` then `Add` reuses one: IDs `[0 2 3]`, `Remove(2)`, and the next `Add` returns 3 again.
- `Add` stores `item` as given, but `checkEventCondition` compares it with `PRICE` ([event_manager.go:186](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L186)). An event added as `price` falls through to the orderbook path, which succeeds when neither `CheckBids` nor `CheckAsks` is set, so it fires as soon as an orderbook exists: ticker at 100, condition `> 200`, fired.

The amd64 and docker test jobs failed on `connection reset by peer` from Coinut, Bitstamp, OKX and Coinbase; no event-manager test failed in either ([job 103926204763](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34828552512/job/103926204763), [job 103926204435](https://github.com/thrasher-corp/gocryptotrader/actions/runs/34828552512/job/103926204435)).

Happy to approve once the lint fix is in.

engine/event_manager.go
ginavalent: Thanks. Modified.

engine/event_manager_test.go
ginavalent: Thanks. Modified.

engine/event_manager_test.go
ginavalent: Please review it again.

engine/event_manager.go
```diff
@@ -71,17 +71,21 @@ func (m *eventManager) Stop() error {
...
-		total, executed := m.getEventCounter()
-		if total > 0 && executed != total {
-			m.m.Lock()
-			for i := range m.events {
-				m.executeEvent(i)
+	defer t.Stop()
+
+	for {
+		select {
+		case <-m.shutdown:
```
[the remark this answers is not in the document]
thrasher-: Heimdall: Closed by the `wg.Wait` in `Stop` ([event_manager.go:69](https://github.com/thrasher-corp/gocryptotrader/blob/48605d506be3a78f064a203768cbe8eafc5f9d2e/engine/event_manager.go#L69)). With `PushEvent` held mid-pass, `Stop` then `Start` now leaves one `run` worker and `-race` is clean. At 4d1e62f the same probe left two workers and a race between [event_manager.go:78](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L78) and [:47](https://github.com/thrasher-corp/gocryptotrader/blob/4d1e62f931d63d1b18915050453356d668fa4088/engine/event_manager.go#L47).

thrasher- (approved):
Heimdall: All four round-1 fixes hold. The rewritten test fails against the old one-shot `run()` ([event_manager_test.go:193](https://github.com/thrasher-corp/gocryptotrader/blob/48605d506be3a78f064a203768cbe8eafc5f9d2e/engine/event_manager_test.go#L193)), even with a late first tick, and `Stop` now waits for the worker ([event_manager.go:69](https://github.com/thrasher-corp/gocryptotrader/blob/48605d506be3a78f064a203768cbe8eafc5f9d2e/engine/event_manager.go#L69)). Thanks for taking every one of them. Approving.

shazbert (approved):
changes approved, thanks for opening this PR.

thrasher-:
Thanks for this PR @ginavalent !

PR #2390, opened 2026-09-17
gctscript: remove 

### LINKED ISSUE #2343: Remove gctscript
nobody uses it; maintainer, dev toil.

shazbert:
# PR Description

closes #2343


thrasher-:
@codex please review

engine/helpers.go
```diff
@@ -84,7 +83,6 @@ func (bot *Engine) GetSubsystemsStatus() map[string]bool {
 		SyncManagerName:               bot.Settings.EnableExchangeSyncManager,
 		grpcName:                      bot.Settings.EnableGRPC,
 		grpcProxyName:                 bot.Settings.EnableGRPCProxy,
-		vm.Name:                       bot.gctScriptManager.IsRunning(),
 		dispatch.Name:                 dispatch.IsRunning(),
```
thrasher-, line 86: Heimdall: **Blocker** — CI is red. `TestGetSubsystemsStatus` still expects 13 subsystems, but this map now has 12.

**Fix:** expect `12` at [helpers_test.go:102](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/engine/helpers_test.go#L102).

**Why:** all 8 failing test jobs in run [35175215377](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35175215377) fail on this assertion, and it reproduces locally:

```
--- FAIL: TestGetSubsystemsStatus (0.00s)
    Error: "map[communications:false currency_state_manager:false data_history_manager:false database:false dispatch:false exchange_syncer:false grpc:false grpc_proxy:false internet_monitor:false ntp_timekeeper:false orders:false portfolio:false]" should have 13 item(s), but has 12
```

It passes at [`ce7da89`](https://github.com/thrasher-corp/gocryptotrader/commit/ce7da894e85098ddb88ad4a627575f494b98c7ef), and passes here once it expects `12`.

go.mod
```diff
@@ -72,4 +72,5 @@ require (
 	golang.org/x/sys v0.48.0 // indirect
 	golang.org/x/xerrors v0.0.0-20220907171357-04be3eba64a2 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20260803160001-6ac0973c030d // indirect
+	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.6.2 // indirect
```
thrasher-, line 75: Heimdall: **Should fix** — `go.mod` still requires tengo, and `go mod tidy` strips this new line straight back out.

**Fix:** run `go mod tidy` to drop `github.com/d5/tengo/v2` ([go.mod:9](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/go.mod#L9)) and its `go.sum` lines. This line is what makes the unversioned `go install` in [gctrpc/README.md:38-42](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/gctrpc/README.md#L38-L42) work (it fails on master), so keep it as a tool dependency instead. Registering all four generators that README installs also keeps the command working after `make update_deps` vendors, and adds no new requirements:

```
go get -tool github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-grpc-gateway github.com/grpc-ecosystem/grpc-gateway/v2/protoc-gen-openapiv2 google.golang.org/protobuf/cmd/protoc-gen-go google.golang.org/grpc/cmd/protoc-gen-go-grpc
```

**Why:**

```
$ go mod tidy -diff                                 # 4f33e47: exit 1
-	github.com/d5/tengo/v2 v2.17.0
-	google.golang.org/grpc/cmd/protoc-gen-go-grpc v1.6.2 // indirect
$ go mod why -m github.com/d5/tengo/v2
(main module does not need module github.com/d5/tengo/v2)

README go install:
  ce7da89                                   exit 1  no required module provides package google.golang.org/grpc/cmd/protoc-gen-go-grpc
  4f33e47                                   exit 0
  4f33e47 + tidy                            exit 1  no required module provides package ...
  + tidy, protoc-gen-go-grpc as a tool      exit 0; exit 1 once vendored: import lookup disabled by -mod=vendor
  + tidy, all four as tools                 exit 0, vendored or not; go mod tidy -diff clean
```

log/logger_setup.go
```diff
@@ -245,7 +245,6 @@ func init() {
 	PortfolioMgr = registerNewSubLogger("PORTFOLIO")
 	SyncMgr = registerNewSubLogger("SYNC")
 	TimeMgr = registerNewSubLogger("TIMEKEEPER")
-	GCTScriptMgr = registerNewSubLogger("GCTSCRIPT")
 	WebsocketMgr = registerNewSubLogger("WEBSOCKET")
```
thrasher-, line 248: Heimdall: **Should fix** — with this registration gone, a saved `GCTSCRIPT` sublogger now stops the engine starting.

**Fix:** skip it in `SetupSubLoggers` ([logger_setup.go:134-148](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/log/logger_setup.go#L134-L148)), where the error is raised, rather than rewriting configs:

```go
	for x := range s {
		name := strings.ToUpper(s[x].Name)
		if _, found := SubLoggers[name]; !found && name == "GCTSCRIPT" {
			// GCTScript has been removed, and configs saved before then may still list its sub logger; that must not stop startup
			continue
		}
		output, err := getWriters(&s[x])
		if err != nil {
			return err
		}
		err = configureSubLogger(name, s[x].Level, output)
		if err != nil {
			return err
		}
	}
```

Please also keep the `gctscript` block in `testdata/configtestv0.json`. It's the real legacy input, and every config test passes with it restored.

**Why:** [engine.go:100](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/engine/engine.go#L100) passes the saved list to `SetupSubLoggers`, and an unregistered name is fatal at [logger_setup.go:121](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/log/logger_setup.go#L121). Loading the pre-removal `testdata/configtest.json` through `NewFromSettings`:

```
                                          ce7da89   4f33e47                           4f33e47 + skip
+ {"name":"GCTSCRIPT"} sublogger           starts    sub logger GCTSCRIPT not found    starts
+ {"name":"gctscript"} sublogger           starts    sub logger GCTSCRIPT not found    starts
  same, under a "subLoggers" key           starts    sub logger GCTSCRIPT not found    starts
  same, under a "Logging" key              starts    sub logger GCTSCRIPT not found    starts
  same, after an empty "subloggers" key    starts    sub logger GCTSCRIPT not found    starts
```

This is the one place every path meets, whichever decoder filled the list. Rewriting the config instead falls short:
- **A config-version migration on `logging.subloggers`** misses the last three rows.
- **Filtering in `readConfig`** misses configs decoded directly with `json.Unmarshal`.
- **Filtering in place in `CheckLoggerConfig`** corrupts a `Config` copied before `UpdateConfig`.

Saved configs keep the stale entry, which is harmless. A consumer that registers its own `GCTSCRIPT` sublogger is still configured, and unknown names still error. A test giving each spelling an invalid output fails without the skip. `GCTSCRIPT` never shipped in an example config or doc, so only hand-added entries hit this.

engine/rpcserver.go
```diff
@@ -20,8 +19,6 @@ import (
 	"github.com/pquerna/otp/totp"
 	"github.com/thrasher-corp/gct-ta/indicators"
 	"github.com/thrasher-corp/gocryptotrader/common"
-	"github.com/thrasher-corp/gocryptotrader/common/file"
-	"github.com/thrasher-corp/gocryptotrader/common/file/archive"
 	"github.com/thrasher-corp/gocryptotrader/common/key"
```
thrasher-, line 22: Heimdall: **Minor** — four more gctscript-only pieces are still here, and nothing uses them now.

**Fix:** delete:
- `common/file/archive`: `GCTScriptUpload`, removed from this file, was its only importer
- `common.GctExt` ([common.go:33-34](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/common/common.go#L33-L34)), "the extension for GCT Tengo script files"
- `database/repository/script`: `Event` was only called from `gctscript/vm/vm.go`. Keep the `script`/`script_execution` migrations and models, since dropping those tables would delete users' data
- `testdata/gctscript/`: with the VM tests gone, only `archive`'s own test still uses it ([zip_test.go:47](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/common/file/archive/zip_test.go#L47))

**Why:** at [`4f33e47`](https://github.com/thrasher-corp/gocryptotrader/commit/4f33e474ad1d645d2e764ca44f57f5979ec1c497), `git grep` across every file type finds no importer of either package, and no use of `GctExt` beyond its declaration. With all four deleted, build, `go vet` and `go test -race ./common/... ./database/... ./config/...` still pass.

cmd/gctcli/commands.go
```diff
@@ -37,7 +34,8 @@ func getInfo(c *cli.Context) error {
 	defer closeConn(conn, cancel)
 
 	client := gctrpc.NewGoCryptoTraderServiceClient(conn)
-	result, err := client.GetInfo(c.Context,
+	result, err := client.GetInfo(
```
thrasher-, line 37: Heimdall: **Nit** — 48 hunks across six files are pure reformatting like this one, and none of them relate to gctscript.

**Fix:** revert the 42 that only move a first argument onto its own line, so the diff covers just the removal. The 6 that put a closing paren on its own line (`engine.go`, `helpers.go`, `rpcserver.go`) can stay: gofumpt v0.11.0 asks for them under `-extra`, which `.golangci.yml` enables.

**Why:** CI's pinned golangci-lint v2.9.0, formatters included, passed on all six files at [`ce7da89`](https://github.com/thrasher-corp/gocryptotrader/commit/ce7da894e85098ddb88ad4a627575f494b98c7ef) ([lint job](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35165083656/job/105024261721)). gofumpt v0.11.0 asks for none of the 42, with or without `-extra`.

thrasher- (changes requested):
Heimdall: A clean cut: 11.5k lines of scripting engine gone. There are no new lint findings against master, and nothing in the docs, templates, workflows or example configs still mentions it. The regenerated `gctrpc` files are byte-identical to a fresh `buf generate` with the plugins CI installs. The evidence for each point is on its line.

**To unblock:**
1. [helpers_test.go:102](https://github.com/thrasher-corp/gocryptotrader/blob/4f33e474ad1d645d2e764ca44f57f5979ec1c497/engine/helpers_test.go#L102): expect 12 subsystems ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4034035254)). It's the only CI failure this PR causes.

These don't gate the merge:
- `go mod tidy`, with the README's proto generators kept as tool dependencies ([go.mod](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4034035262))
- skipping saved `GCTSCRIPT` sublogger entries so upgraded configs still start ([logger_setup.go](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4034035270))
- leftover gctscript-only code ([rpcserver.go](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4034035286))
- the unrelated reformatting ([commands.go](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4034035295))

The Windows job's Kraken `TestGetRecentTrades` failure (`EQuery:Unknown asset pair`) is ambient. It's a live-API test in a package this PR doesn't touch, and the same commit passed it on the other seven jobs.

Happy to approve once CI is green.

config/versions/v16/v16.go
```diff
@@ -0,0 +1,64 @@
...
+var legacyGCTScriptConfig = []byte(`{"enabled":false,"timeout":30000000000,"max_virtual_machines":10,"allow_imports":false,"auto_load":null,"verbose":false}`)
+
+// Version implements ConfigVersion to remove decommissioned GCTScript configuration.
+type Version struct{}
+
+// UpgradeConfig removes the GCTScript configuration and any obsolete GCTSCRIPT sublogger.
+func (*Version) UpgradeConfig(_ context.Context, config []byte) ([]byte, error) {
+	config = jsonparser.Delete(config, "gctscript")
+
+	subloggersJSON, valueType, _, err := jsonparser.Get(config, "logging", "subloggers")
```
thrasher-: Heimdall: **Should fix** — the sub logger filter and the decoder resolve `logging.subloggers` by different rules, so seven config shapes still stop the engine starting.

`jsonparser.Get` takes the exact key and the first occurrence that has it. `encoding/json` matches keys case-insensitively and lets the last occurrence win. Whenever the two disagree, the `GCTSCRIPT` entry survives the migration and reaches [logger_setup.go:138](https://github.com/thrasher-corp/gocryptotrader/blob/0fd916d690fb0ce64e4b410a009b8c93be28e0a0/log/logger_setup.go#L138), which is fatal at [engine.go:100](https://github.com/thrasher-corp/gocryptotrader/blob/0fd916d690fb0ce64e4b410a009b8c93be28e0a0/engine/engine.go#L100).

**Fix:** keep the migration, and add the skip in `SetupSubLoggers` as well ([logger_setup.go:133-142](https://github.com/thrasher-corp/gocryptotrader/blob/0fd916d690fb0ce64e4b410a009b8c93be28e0a0/log/logger_setup.go#L133-L142)). It is the one place every decode path meets, and the last two rows below are not reachable by making the lookup case-insensitive:

```go
	for x := range s {
		name := strings.ToUpper(s[x].Name)
		if _, found := SubLoggers[name]; !found && name == "GCTSCRIPT" {
			// GCTScript has been removed, and configs saved before then may still list its sub logger; that must not stop startup
			continue
		}
		output, err := getWriters(&s[x])
		if err != nil {
			return err
		}
		if err := configureSubLogger(name, s[x].Level, output); err != nil {
			return err
		}
	}
```

**Why:** master's `testdata/configtest.json` at version 15, plus one `GCTSCRIPT` sub logger entry, through `ReadConfigFromFile` then `SetupSubLoggers`:

```
                                                    c6e47c6e  0fd916d6                        + skip
logging / subloggers,  "GCTSCRIPT"                  starts    starts                          starts
logging / subloggers,  "gctscript"                  starts    starts                          starts
logging / subLoggers,  "GCTSCRIPT"                  starts    sub logger GCTSCRIPT not found  starts
Logging / subloggers,  "GCTSCRIPT"                  starts    sub logger GCTSCRIPT not found  starts
"subloggers":[] then "subLoggers":[GCTSCRIPT]       starts    sub logger GCTSCRIPT not found  starts
"subloggers":[] then "subloggers":[GCTSCRIPT]       starts    sub logger GCTSCRIPT not found  starts
two "logging" objects, the later one carrying it    starts    sub logger GCTSCRIPT not found  starts
```

Identical on both JSON backends (`encoding/json/v2` and `-tags sonic_on`). With the skip applied, `go test ./log/ ./config/... -race -count 1` stays green.

End to end on row three with the built binary:

```
2026/09/21 10:07:26 Running upgrade to config version 16
2026/09/21 10:07:26 Unable to initialise bot engine. Error: failed to setup sub loggers. sub logger GCTSCRIPT not found
exit status 1
```

It does not self-heal either way, and which way depends on `encryptConfig` ([config.go:1434](https://github.com/thrasher-corp/gocryptotrader/blob/0fd916d690fb0ce64e4b410a009b8c93be28e0a0/config/config.go#L1434)):

```
encryptConfig: 0   saved on the way through -> disk becomes "version": 16 with the key
                   normalised to "subloggers" and GCTSCRIPT still in it. Deploy returns at
                   target == current, so v16 never gets another run at the spelling it handles
encryptConfig: -1  not saved. start 1: exit=1 disk_version=15 v16_migrations_run=1
                   start 2: exit=1 disk_version=15 v16_migrations_run=1
```

Hand-edited configs are the only way in — nothing GCT writes uses those shapes, and `"subLoggers"` and `"Logging"` have never appeared as keys in this repo's history — so the population is small. It is four lines to close the class.

config_example.json
```diff
@@ -106,14 +106,6 @@
   "allowedDifference": 50000000,
   "allowedNegativeDifference": 50000000
  },
- "gctscript": {
```
thrasher-, line 109: Heimdall: **Should fix** — the block went but the version did not: the example is still `"version": 15` with 16 now registered, so every fresh install runs a no-op migration and rewrites the file on first start.

**Fix:** bump [config_example.json:3](https://github.com/thrasher-corp/gocryptotrader/blob/0fd916d690fb0ce64e4b410a009b8c93be28e0a0/config_example.json#L3) to `16`. [`3b37dd0`](https://github.com/thrasher-corp/gocryptotrader/commit/3b37dd00750d2a48189b9c4d0f2e7a58a16f49d9), which added v15, bumped `config_example.json` and `testdata/configtest.json` together; this one bumped only the fixture.

**Why:** copying the example and loading it through `ReadConfigFromFile` gives `Config.Version == 16` in memory — v16 running on a config that already has nothing for it to do — and `encryptConfig: 0` means it is then saved back. Nothing asserts the example's version, so CI is green either way. With the bump, `go test ./config/...` and `make misc_checks` both pass.

testdata/configtest.json
```diff
@@ -1,6 +1,6 @@
 {
  "name": "Skynet",
- "version": 15,
+ "version": 16,
```
thrasher-, line 3: Heimdall: **Minor** — the fixture now says version 16 but still carries the `gctscript` block at [configtest.json:110-117](https://github.com/thrasher-corp/gocryptotrader/blob/0fd916d690fb0ce64e4b410a009b8c93be28e0a0/testdata/configtest.json#L110-L117). Nothing that reaches version 16 through the migration keeps that key — `UpgradeConfig` deletes it.

**Fix:** drop those eight lines. `go test ./config/... ./engine/ ./common/file/archive/ -count 1` passes without them. [`3b37dd0`](https://github.com/thrasher-corp/gocryptotrader/commit/3b37dd00750d2a48189b9c4d0f2e7a58a16f49d9) removed the `Bitmex` entry from the fixtures in the same change that added v15.

Nothing breaks today — `Deploy` leaves an already-16 config alone and `Config` ignores the field — so this is fixture hygiene, not a defect. Keep the block in `testdata/configtestv0.json`: that one is a genuine v0 input and wants it.

config/versions/v16/v16_test.go
```diff
@@ -0,0 +1,49 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"github.com/thrasher-corp/gocryptotrader/config/versions"
+	v16 "github.com/thrasher-corp/gocryptotrader/config/versions/v16"
+)
+
+func TestUpgradeConfig(t *testing.T) {
```
thrasher-, line 12: Heimdall: **Minor** — the v16 tests are materially thinner than v15's, and the dimension they leave out is the one that hides the sub logger gap above.

**Fix:** match [v15_test.go](https://github.com/thrasher-corp/gocryptotrader/blob/0fd916d690fb0ce64e4b410a009b8c93be28e0a0/config/versions/v15/v15_test.go) — a table for `UpgradeConfig`, plus the error paths. Missing today:

- no `logging` key at all (the `KeyPathNotFoundError` return); the existing case covers only `subloggers: null`
- an entry with no `name`, correctly preserved: `[{"level":"INFO"},{"name":"GCTSCRIPT"}]` → `[{"level":"INFO"}]`
- the two error returns — `{"logging":{"subloggers":[{"name":"GCTSCRIPT"},]}}` gives `invalid character ']' looking for beginning of value` and the original config back; v15 has three such tests, v16 none
- a multi-hop `Deploy`; `TestRegisteredMigration` only does 15 → 16 → 15
- one case through `ReadConfigFromFile` with `logging.subloggers` spelled as a saved config spells it

thrasher- (approved):
Heimdall: The blocker is gone — `TestGetSubsystemsStatus` expects 12 and engine passes on all eight test jobs. `go mod tidy -diff` is clean, `git grep -il "gctscript\|tengo"` is down to v16 and the two fixtures, and the `gctrpc` diff is pure deletion against a master regenerated with the same pinned plugins, bar message renumbering.

Four comments, evidence on the lines, none gating: [v16.go:22](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4058670033) sub logger shapes that still stop startup; [config_example.json](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4058670044) left at version 15; [configtest.json:3](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4058670051) stale `gctscript` block; [v16_test.go](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4058670056) thin against v15's.

The three red jobs are ambient — btse, deribit and currencylayer live-API tests in untouched packages; master at the same base fails the same arm64 job on btse ([job 105502915780](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35314394628/job/105502915780)).

Release note when it lands: nine gRPC methods, two CLI flags, `kline.LoadFromGCTScriptCSV`, `common.GctExt` and the `gctscript` config key.

Take the four or leave them — nothing here blocks.

log/logger_setup.go
thrasher-: Heimdall: Partly. v16 strips the entry for the spellings a GCT-written config uses, and it matches `GCTSCRIPT` and `gctscript` alike — the common case is covered. Five shapes the loader still accepts get past it and stop the engine starting, including two that a case-insensitive lookup would not reach; evidence and the four-line close are on [v16.go:22](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4058670033). Leaving this open behind that one.

log/logger_test.go
```diff
@@ -110,6 +110,12 @@ func TestSetGlobalLogConfig(t *testing.T) {
 	require.NoError(t, err)
 }
 
+func TestSetupSubLoggersIgnoresRemovedGCTScript(t *testing.T) {
+	t.Parallel()
+	require.NoError(t, SetupSubLoggers([]SubLoggerConfig{{Name: "gctscript"}}), "SetupSubLoggers must ignore the removed GCTScript sublogger")
+	require.Error(t, SetupSubLoggers([]SubLoggerConfig{{Name: "unknown"}}), "SetupSubLoggers must reject other unknown subloggers")
```
thrasher-: Heimdall: **Minor** — the second assertion passes whether or not unknown sub loggers are rejected.

`SubLoggerConfig.Output` is empty, so `getWriters` ([logger_setup.go:43](https://github.com/thrasher-corp/gocryptotrader/blob/5f69eec0b9258134abc5225fa3c015c776d6f560/log/logger_setup.go#L43)) returns `unhandled output writer:` before `configureSubLogger` ever looks the name up. The assertion proves an empty output is rejected, not the thing its message claims.

**Fix:** give both entries an output.

```go
	require.NoError(t, SetupSubLoggers([]SubLoggerConfig{{Name: "gctscript", Output: "console"}}), "SetupSubLoggers must ignore the removed GCTScript sublogger")
	require.Error(t, SetupSubLoggers([]SubLoggerConfig{{Name: "unknown", Output: "console"}}), "SetupSubLoggers must reject other unknown subloggers")
```

**Why:** direct results on this HEAD, and the mutation that should fail the test:

```
Name:"unknown",   Output:""         -> unhandled output writer:
Name:"unknown",   Output:"console"  -> sub logger UNKNOWN not found
Name:"gctscript", Output:"console"  -> <nil>

configureSubLogger returns nil for an unregistered name:
  as committed        ok      github.com/thrasher-corp/gocryptotrader/log
  + Output "console"  FAIL    An error is expected but got nil
```

The first assertion is sound either way — the skip fires before `getWriters` — but it is worth giving it an output too, so the pair reads the same.

config/versions/v16/v16.go
```diff
@@ -0,0 +1,69 @@
...
+	if valueType != jsonparser.Array {
+		return config, nil
+	}
+
+	var subloggers []json.RawMessage
+	if err := json.Unmarshal(subloggersJSON, &subloggers); err != nil {
+		return original, err
+	}
+
+	filtered := subloggers[:0]
```
thrasher-: Heimdall: **Minor** — rewriting the array breaks a config that spells `subloggers` twice, and can silently change a sub logger that survives.

`encoding/json` merges duplicate arrays element-wise rather than replacing them, so a partially specified later array inherits fields from the first. Emptying the first array destroys what it was inheriting.

**Why:** master's `testdata/configtest.json` at version 15 with the `logging` block below, through `ReadConfigFromFile` then `SetupSubLoggers`:

```
                                          c6e47c6e (master)                 5f69eec0b
"subloggers":[GCTSCRIPT/console]
  + "subloggers":[{"level":"WARN"}]       {GCTSCRIPT WARN console} starts   {"" WARN ""} unhandled output writer
  + "subLoggers":[{"level":"WARN"}]       {GCTSCRIPT WARN console} starts   {"" WARN ""} unhandled output writer
[GCTSCRIPT/ERROR, DATABASE/INFO]
  + "subloggers":[{"name":"DATABASE"}]    {DATABASE ERROR} starts           {DATABASE INFO} starts
```

The third row is the one that worries me: a sub logger the user keeps moves from ERROR to INFO with no error raised. The `GCTSCRIPT` skip cannot catch the first two, because the name ends up empty rather than `GCTSCRIPT`.

**Fix — your call, two that I measured:**

- Drop the sub logger filtering from `UpgradeConfig` and keep only `jsonparser.Delete(config, "gctscript")`. The runtime skip already makes every shape safe. All eleven configs I have then start, and master's levels are preserved. The trade is a stale `GCTSCRIPT` entry left in the saved config, silently ignored.
- Keep the filter, but leave the array alone when `logging` holds more than one key that case-insensitively matches `subloggers`. That keeps the tidy-up for ordinary configs; I have not implemented it, so I am not claiming it works.

**This is not new in [`5f69eec`](https://github.com/thrasher-corp/gocryptotrader/commit/5f69eec0b9258134abc5225fa3c015c776d6f560) — it reproduces identically on [`0fd916d`](https://github.com/thrasher-corp/gocryptotrader/commit/0fd916d690fb0ce64e4b410a009b8c93be28e0a0), so I missed it last round.** My matrix varied the key spelling but never the contents of a second array. Related to the now-fixed thread on this file.

thrasher- (approved):
Heimdall: All four are in and verified. The skip at [logger_setup.go:133-138](https://github.com/thrasher-corp/gocryptotrader/blob/5f69eec0b9258134abc5225fa3c015c776d6f560/log/logger_setup.go#L133-L138) starts every one of the seven shapes that failed last round, on both JSON backends, and the binary starts on the config that exited 1 before. `config_example.json` is 16 and runs no migration on a fresh install; the fixture's `gctscript` block is gone; v16's tests now cover three distinct error returns and a 12→16 `Deploy`, taking the package from 80.0% to 91.4%. Returning the pristine `original` on error is a nice extra.

Two comments, neither gating: [logger_test.go:116](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4058790964) an assertion that passes either way; [v16.go:39](https://github.com/thrasher-corp/gocryptotrader/pull/2390#discussion_r4058790967) the array rewrite breaking a config that spells `subloggers` twice — which is mine, it was there last round and I missed it.

The three red jobs are ambient: gateio and deribit live-API tests, both passing on four other jobs of the same run. `codecov/project` at -0.09% is not this commit — every package it touches gained coverage — but I could not establish what did move it.

thrasher- (approved):
Heimdall: Both are in. [`UpgradeConfig`](https://github.com/thrasher-corp/gocryptotrader/blob/3f20003a25a01a8f8ffd9e15920aaba1718d4314/config/versions/v16/v16.go#L17-L19) is a single `jsonparser.Delete` now, and across eleven config shapes — every key spelling plus the duplicate-array cases — the decoded sub logger name, level and output are identical to [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367), on both JSON backends. The two shapes that died with `unhandled output writer:` last round start, and the `DATABASE` logger that silently moved ERROR→INFO keeps ERROR. v16 is at 100% coverage, upgrade/downgrade is lossless now that the entry survives, and a saved config round-trips idempotently.

The logger test's second assertion bites: making `configureSubLogger` accept unregistered names fails it, where before that mutation left it passing. `exact: true` on the duplicate-array case was the right call — `assert.JSONEq` returns true when one of two duplicate arrays is dropped, so it would not have caught a regression there.

CI is 16/16, `codecov/project` included once the uploads settled. Nothing open from me.

PR #2158, opened 2026-01-30
gateio: Update API fields due to widespread upgrade 

shazbert:
From article https://www.gate.com/announcements/article/48788 there is header context field switching as they upgrade their endpoints. 

- fix: update Gate.io websocket handling for size and amount conversions
- Added "X-Gate-Size-Decimal" header rest and websocket connection
- Updated futures and options processing to convert sizes and amounts to float64 where necessary.
- Ensured consistency in handling size and amount across various functions in the Gate.io exchange wrapper.

Closes #2173 

# PR Description

Please include a summary of the change, feature or issue which this pull request addresses. Please also include relevant motivation and context. List any dependencies that are required for this change.

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2392, opened 2026-09-17
fix:gateio order & okx instIdCode

dazi005:
# PR Description

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


thrasher-:
@codex please review

exchanges/gateio/gateio_types.go
```diff
@@ -1880,7 +1880,7 @@ type DeliveryOrderCreateParams struct {
 // FuturesOrder represents future order response
 type FuturesOrder struct {
 	ID                        int64         `json:"id"`
-	User                      string        `json:"user"`
+	User                      uint64        `json:"user"`
```
thrasher-: Heimdall: **Blocker** — `User uint64` rejects the `futures.orders` websocket push, which sends `user` as a quoted string.

**Fix:** use `types.Number`, which takes both shapes. The [coding guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/docs/CODING_GUIDELINES.md#L42) prescribe it when an API "may return either a string or a bare number". `SelfTradePreventionID` already uses it in the same struct for the same reason: `stp_id` is `0` over REST and `"0"` over the websocket. Then:

- At [gateio_websocket_futures.go:743](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/gateio/gateio_websocket_futures.go#L743), set `AccountID: resp.Result[x].User.String()`.
- Replace Gate's `"110xxxxx"` doc placeholder in the `futures.orders` fixtures with numeric strings, as the live capture has: [gateio_test.go:3951](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/gateio/gateio_test.go#L3951), [gateio_test.go:4084-4091](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/gateio/gateio_test.go#L4084-L4091) and [wsFutures.json:4](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/gateio/testdata/wsFutures.json#L4).

With exactly those changes, the whole `exchanges/gateio` package passes under `-race`, and the four tests below pass under `-tags sonic_on`. A small test that decodes `{"user":12870774}` and `{"user":"12870774"}` into `FuturesOrder` pins both shapes. I tried one: it fails on this branch's `uint64` and passes with `types.Number`.

**Why:** `FuturesOrder` is shared by 12 REST methods and by `processFuturesOrdersPushData` ([gateio_websocket_futures.go:699](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/gateio/gateio_websocket_futures.go#L699)), which handles every perpetual and delivery order update. The two sides disagree on the type:

- REST sends an integer: [model_futures_order.go:17](https://github.com/gateio/gateapi-go/blob/44355712305a5ea00490fae347322dc99a68b6ca/model_futures_order.go#L17) has `User int32` and [futures_order.py:37](https://github.com/gateio/gateapi-python/blob/22fdf34ae4af855863ccb01b95e8c996de4cbea3/gate_api/models/futures_order.py#L37) has `'user': 'int'`.
- The push sends a string: `"user":"12870774"` in the live capture at [gateio_test.go:3991](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/gateio/gateio_test.go#L3991).

#2158 ([`97c28a8`](https://github.com/thrasher-corp/gocryptotrader/commit/97c28a88881136d2404376327105df00a2732f89)) merged the two types and kept the websocket's `string`, which broke REST. This change moves the break to the websocket. Same httptest server and same live payload on both commits:

```
                            master ce7da89                                       this PR 82fb0e9
REST GetSingleFuturesOrder  cannot unmarshal number into .user of type string    ok  user=12870774
futures.orders push (live)  ok  AccountID="12870774"                             cannot unmarshal string into .user of type uint64
```

That accounts for all of CI's test failures from this PR: `TestProcessFuturesOrdersPushData`, `TestProcessFinishedFuturesOrdersPushData`, `TestFuturesLiveCapturedPayloads` and `TestFuturesDataHandler` fail in all 8 test jobs (for example [job 105193327309](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35218716049/job/105193327309)). All four pass at [`ce7da89`](https://github.com/thrasher-corp/gocryptotrader/commit/ce7da894e85098ddb88ad4a627575f494b98c7ef).

exchanges/okx/okx_types.go
```diff
@@ -865,7 +868,8 @@ type CancelMassReqParam struct {
 
 // AmendOrderRequestParams represents amend order requesting parameters
 type AmendOrderRequestParams struct {
-	InstrumentID    string  `json:"instId"`
+	InstrumentID    string
```
thrasher-: Heimdall: **Blocker** — the `json:"instId"` tag was dropped from `InstrumentID`, so REST amends now send `"InstrumentID"` instead of `instId`.

**Fix:** restore `` `json:"instId"` ``.

**Why:** both REST amend endpoints require `instId`: [single](https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-amend-order) and [batch](https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-amend-multiple-orders). REST is the path `ModifyOrder` takes whenever the authenticated websocket isn't connected ([okx_wrapper.go:1191-1195](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_wrapper.go#L1191-L1195)). Here is the body `ModifyOrder` sends through an httptest server, for the same call on both commits:

```
master ce7da89   POST trade/amend-order  {"instId":"BTC-USDT","ordId":"111","newSz":"2"}
this PR 82fb0e9  POST trade/amend-order  {"InstrumentID":"BTC-USDT","instIdCode":"3","ordId":"111","newSz":"2"}
```

The websocket amend ops ignore `instId` since [2026-04-07](https://www.okx.com/docs-v5/log_en/#2026-04-07), so the tag only matters on REST. There it breaks both `AmendOrder` and `AmendMultipleOrders`. CI can't see this: all 12 place, amend and cancel tests (REST, websocket and wrapper) skip without credentials. I couldn't confirm the rejection live for the same reason. If you've had a 200 from a REST amend on this branch, say so and I'll drop this.

exchanges/okx/okx_types.go
```diff
@@ -371,6 +371,7 @@ type InstrumentsFetchParams struct {
 type Instrument struct {
 	InstrumentType                  string        `json:"instType"`
 	InstrumentID                    currency.Pair `json:"instId"`
+	InstIdCode                      int64         `json:"instIdCode"`
```
thrasher-: Heimdall: **Blocker** — the `InstIdCode` / `instIdCode` / `getInstIdCode` names fail the lint job with 14 revive `var-naming` errors ([lint job](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35218716100/job/105193326948)). All 14 are on lines this PR adds: 8 struct fields in okx_types.go, plus 5 variables and the method in okx_wrapper.go.

**Fix:** rename to `InstrumentIDCode`, `instrumentIDCode` and `getInstrumentIDCode`, and make the type `uint64`. That clears revive and matches `InstrumentID` / `InstrumentType` on the same structs. The [coding guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/docs/CODING_GUIDELINES.md#L44-L45) ask for full field names and `uint64` for integers, and these codes are always positive. While you're there, drop "add by dazi" from the GoDoc at [okx_wrapper.go:3101](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_wrapper.go#L3101).

**Why:** a local `golangci-lint run ./exchanges/gateio/... ./exchanges/okx/...` reports exactly these 14 on top of what [`ce7da89`](https://github.com/thrasher-corp/gocryptotrader/commit/ce7da894e85098ddb88ad4a627575f494b98c7ef) reports, the same 14 CI shows.

exchanges/okx/okx_types.go
```diff
@@ -865,7 +868,8 @@ type CancelMassReqParam struct {
 
 // AmendOrderRequestParams represents amend order requesting parameters
 type AmendOrderRequestParams struct {
-	InstrumentID    string  `json:"instId"`
+	InstrumentID    string
+	InstIdCode      int64   `json:"instIdCode,string,omitempty"`
```
thrasher-: Heimdall: **Should fix** — the `instIdCode` tags disagree with each other. Amend sends the code as a string, and cancel sends `0` when the lookup misses.

**Fix:**

- Use `` `json:"instIdCode,omitempty"` `` here and on `CancelOrderRequestParam` ([okx_types.go:858](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L858)), as `PlaceOrderRequestParam` already does ([okx_types.go:772](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L772)).
- Drop the field, and its wrapper assignments, from `AlgoOrderParams` ([okx_types.go:1092](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L1092)), `AmendAlgoOrderParam` ([okx_types.go:1152](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L1152)) and `AlgoOrderCancelParams` ([okx_types.go:1247](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L1247)). Those types only go to the REST algo endpoints, and OKX doesn't document `instIdCode` on any REST order or algo endpoint.

**Why:** OKX's websocket ops take `instIdCode` as an `Integer` ([amend-order](https://www.okx.com/docs-v5/en/#order-book-trading-trade-ws-amend-order)). Here are frames this branch sent to a mock websocket server, from wrapper calls on BTC-USDT:

```
cache filled  order                {"instId":"BTC-USDT","instIdCode":3,...}
              amend-order          {"InstrumentID":"BTC-USDT","instIdCode":"3",...}
              cancel-order         {"instId":"BTC-USDT","instIdCode":3,...}
cache empty   cancel-order         {"instId":"BTC-USDT","instIdCode":0,...}
              batch-cancel-orders  [{"instId":"BTC-USDT","instIdCode":0,...}]
```

The same zero goes out over REST on `cancel-order`, `cancel-batch-orders` and `cancel-advance-algos`. Without credentials I couldn't check whether OKX coerces a quoted code, or what it does with a zero one. Matching the documented type, and omitting a code that isn't known, removes both questions.

exchanges/okx/okx_types.go
```diff
@@ -983,6 +987,7 @@ type OrderListRequestParams struct {
 	InstrumentType string    `json:"instType"` // SPOT , MARGIN, SWAP, FUTURES , OPTIONS
 	Underlying     string    `json:"uly"`
 	InstrumentID   string    `json:"instId"`
+	InstIdCode     int64     `json:"instIdCode,string,omitempty"`
```
thrasher-: Heimdall: **Minor** — `GetOrderList` never reads this field, so the lookup at [okx_wrapper.go:1451](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_wrapper.go#L1451) has no effect on what's sent.

**Fix:** drop the field and the lookup. If the aim was cancel-all over the websocket, the code has to go on the cancel params built at [okx_wrapper.go:1468-1486](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_wrapper.go#L1468-L1486). Those carry neither `InstrumentID` nor a code today.

**Why:** [okx.go:226-257](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx.go#L226-L257) builds the query from named fields only. With the cache populated, this branch still sends `GET trade/orders-pending` with `instId=BTC-USDT&instType=SPOT`. The missing `instId` on those cancel params is pre-existing, and I've listed it as optional in the review body.

exchanges/okx/okx_wrapper.go
```diff
@@ -3076,3 +3097,31 @@ func (e *Exchange) MessageID() string {
 	hex.Encode(buf[:], u[:])
 	return string(buf[:])
 }
+
+// getInstIdCode returns the instIdCode for a given asset type and instrument ID add by dazi
+func (e *Exchange) getInstIdCode(a asset.Item, instID string) int64 {
```
thrasher-: Heimdall: **Should fix** — when the lookup misses it silently returns 0, and nothing on the websocket path checks for it. A websocket order then goes out with no code, and the code is the only instrument identifier OKX now reads.

**Fix:**

- Resolve the code only on the websocket path, since REST doesn't need it, and have a miss return a sentinel error there instead of 0.
- Make `WSPlaceOrder`, `WSCancelOrder`, `WSAmendOrder` and their batch variants require a non-zero code. The six credential-gated tests in `ws_requests_test.go` will need a code as well.
- Add request-body tests for `SubmitOrder`, `ModifyOrder` and `CancelOrder`, over REST with httptest and over the websocket with a mock server. They would have caught [okx_types.go:871](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L871) and [okx_types.go:872](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L872).

I tried exactly that shape against this branch. It builds, `exchanges/okx` passes under `-race`, the new tests also pass under `-tags sonic_on`, and it adds no lint findings. Its body tests fail if the `instId` tag, the `,string` or the missing `omitempty` come back.

**Why:** the lookup itself is right. I served OKX's live instrument lists through httptest, and every pair `FetchTradablePairs` returns resolves to its correct code: 1,425 spot, 266 margin, 482 swap, 215 futures and 2,680 options. The engine fills the cache at bootstrap for enabled assets ([exchange.go:1824-1832](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/exchange.go#L1824-L1832)).

Library callers and supported-but-disabled assets hit the miss, though. So does anyone calling the exported websocket methods directly, because those methods never resolve the code themselves. OKX has ignored `instId` on these ops since [2026-03-26](https://www.okx.com/docs-v5/log_en/#2026-03-26) (place) and [2026-04-07](https://www.okx.com/docs-v5/log_en/#2026-04-07) (amend and cancel), so these frames from a mock websocket server carry nothing OKX reads:

```
wrapper, cache empty        order        {"instId":"BTC-USDT","tdMode":"","side":"buy",...}
wrapper, cache empty        amend-order  {"InstrumentID":"BTC-USDT","ordId":"111","newSz":"2"}
WSPlaceOrder, cache filled  order        {"instId":"BTC-USDT","tdMode":"cash","side":"buy",...}
```

Optional: the same six methods also refuse a request with only the code set, which is the shape OKX's websocket examples use, with `missing instrument ID`. The wrapper always sets both fields, so this affects direct callers only.

Nit on the same function: the scan is linear. It measured 34 µs per order over the 1,425 spot instruments and 63 µs over the 2,680 options, against 12 ns for a map keyed by `instId` and built in `getInstrumentsForAsset`. Spot and margin share codes (BTC-USDT is 3 in both), so one map covers both.

thrasher- (changes requested):
Heimdall: Both halves of this fix real problems:

- **Gate methods:** both changes match Gate's SDK. `GetPosition` is a `GET` ([api_futures.go:1753](https://github.com/gateio/gateapi-go/blob/44355712305a5ea00490fae347322dc99a68b6ca/api_futures.go#L1753)) and `SetDualMode` is a `POST` with `dual_mode` in the query ([api_futures.go:2351](https://github.com/gateio/gateapi-go/blob/44355712305a5ea00490fae347322dc99a68b6ca/api_futures.go#L2351)). An httptest capture confirms the branch now sends both correctly.
- **OKX `instIdCode`:** master's websocket `order`, `amend-order`, `cancel-order` and `batch-cancel-orders` frames carry only `instId`. OKX has ignored `instId` on those ops since [2026-03-26](https://www.okx.com/docs-v5/log_en/#2026-03-26) (place) and [2026-04-07](https://www.okx.com/docs-v5/log_en/#2026-04-07) (amend and cancel), so websocket trading needs this. `getInstIdCode` resolves the correct code for all 5,068 live instruments across spot, margin, swap, futures and options.

The evidence for each point below is in the inline comment on that line.

**To unblock:**
1. [gateio_types.go:1883](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/gateio/gateio_types.go#L1883): `User uint64` fails every `futures.orders` push, which is the four Gate test groups failing in all 8 CI test jobs. The field needs a type that accepts both the REST number and the websocket string. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4043163834))
2. [okx_types.go:871](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L871): restore the dropped `json:"instId"` tag. REST amends currently send no `instId`. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4043163849))
3. [okx_types.go:374](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L374): the 14 revive `var-naming` errors fail the lint job. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4043163858))

The other inline comments don't block the merge:
- [okx_types.go:872](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L872): the code is sent as a quoted string, and as `0` on a lookup miss. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4043163865))
- [okx_types.go:990](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_types.go#L990): the `OrderListRequestParams` field is never read. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4043163873))
- [okx_wrapper.go:3102](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_wrapper.go#L3102): a miss silently returns 0, the websocket validation checks the wrong field, and there are no tests. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4043163895))

**Optional, and outside this diff:** two issues already on master at [`ce7da89`](https://github.com/thrasher-corp/gocryptotrader/commit/ce7da894e85098ddb88ad4a627575f494b98c7ef), which this PR neither causes nor fixes. Both were reproduced through httptest:
- [okx_wrapper.go:1468-1486](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_wrapper.go#L1468-L1486): `CancelAllOrders` builds its cancel requests without `instId`, so it fails locally with `missing instrument ID` whenever there are orders to cancel.
- [okx_wrapper.go:1184-1190](https://github.com/thrasher-corp/gocryptotrader/blob/82fb0e94722e4f12c094f939d0ef3a11fb768598/exchanges/okx/okx_wrapper.go#L1184-L1190): `ModifyOrder` never sets `NewPrice` for regular orders, so a price change is dropped and only `newSz` goes out.

Happy for these to land here; otherwise we'll open a separate issue.

**Ambient CI, unrelated to this PR** (both in [job 105193327309](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35218716049/job/105193327309)):
- `TestGetSettlementHistory` hit a Gate client timeout on `options/settlements` and passes locally.
- kucoin `TestProcessOrderbook/last_updated_fallback` is an existing flake on master, caused by parallel subtests sharing one exchange name.

I have no OKX credentials, so everything here about OKX's server behaviour comes from its docs, not a live order. One question only a live order can answer: the REST place, cancel and amend bodies now carry `instIdCode`, which no REST endpoint documents. If you've placed and cancelled an order over REST on this branch, that settles it.

exchanges/gateio/gateio_types.go
```diff
@@ -1880,7 +1880,7 @@ type DeliveryOrderCreateParams struct {
 // FuturesOrder represents future order response
 type FuturesOrder struct {
 	ID                        int64         `json:"id"`
-	User                      string        `json:"user"`
+	User                      types.Number  `json:"user"`
```
thrasher-, line 1883: Heimdall: **Blocker** — `types.Number` is the right type, but ten `futures.orders` fixtures still carry Gate's `"110xxxxx"` documentation placeholder, which it can't parse.

**Fix:** replace `"user":"110xxxxx"` with a numeric string in those ten fixtures — [gateio_test.go:3951](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/gateio/gateio_test.go#L3951), [gateio_test.go:4084-4091](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/gateio/gateio_test.go#L4084-L4091) and [testdata/wsFutures.json:4](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/gateio/testdata/wsFutures.json#L4). The live captures in the same file already use `"user":"12870774"` ([gateio_test.go:3991](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/gateio/gateio_test.go#L3991)). Leave the `futures.positions` and `futures.liquidates` placeholders alone — those structs still type `user` as `string` and decode fine.

**Why:** three tests fail in all eight CI test jobs (for example [job 105538906460](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35325947905/job/105538906460)) and pass at the base commit. Same `go test ./exchanges/gateio/... -race -count 1` on both:

```
                                          this PR 4c1bff3                            base ce7da89
TestProcessFuturesOrdersPushData          invalid value for Number type: 110xxxxx    ok
TestProcessFinishedFuturesOrdersPushData  invalid value for Number type: 110xxxxx    ok
TestFuturesDataHandler                    invalid value for Number type: 110xxxxx    ok
```

`110xxxxx` can never arrive from Gate — it is Gate's own redaction in the `futures.orders` doc example — so correcting it is what the guidelines mean by "[maintain original test inputs unless they are incorrect](https://github.com/thrasher-corp/gocryptotrader/blob/c6e47c6e1e6d6f7353148379d587188ec47c0367/docs/CODING_GUIDELINES.md#L213)". With exactly those ten changed and nothing else, `exchanges/gateio` passes under `-race` and under `-tags sonic_on`.

Keep the type: `FuturesOrder` is shared by 12 REST methods and by `processFuturesOrdersPushData` ([gateio_websocket_futures.go:743](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/gateio/gateio_websocket_futures.go#L743)), and the two sides disagree. REST sends a bare number — [model_futures_order.go:17](https://github.com/gateio/gateapi-go/blob/44355712305a5ea00490fae347322dc99a68b6ca/model_futures_order.go#L17) declares `User int32` — while the push sends a quoted string. Decoding the same two payloads on each commit:

```
                            master c6e47c6                                             this PR 4c1bff3
{"user":12870774}    cannot unmarshal number into FuturesOrder.user of type string     ok  AccountID="12870774"
{"user":"12870774"}  ok  AccountID="12870774"                                          ok  AccountID="12870774"
```

Nothing in the suite decodes the REST shape into `FuturesOrder` today, which is why master's half of this is invisible. A small test over both shapes would pin it, and would have caught the fixtures too.

exchanges/okx/okx_wrapper.go
```diff
@@ -896,6 +896,7 @@ func (e *Exchange) SubmitOrder(ctx context.Context, s *order.Submit) (*order.Sub
 		return nil, currency.ErrCurrencyPairEmpty
 	}
 	pairString := pairFormat.Format(s.Pair)
+	instIdCode := e.getInstrumentIDCode(s.AssetType, pairString)
```
thrasher-: Heimdall: **Blocker** — the local variables kept the `instIdCode` spelling, so the lint job is still red with four revive `var-naming` errors.

**Fix:**

```suggestion
	instrumentIDCode := e.getInstrumentIDCode(s.AssetType, pairString)
```

Not a clean one-click: the same rename is needed at [okx_wrapper.go:1175](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_wrapper.go#L1175), [:1281](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_wrapper.go#L1281) and [:1333](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_wrapper.go#L1333), along with the four `InstrumentIDCode: instIdCode,` assignments that read them.

**Why:** the struct renames cleared 10 of the original 14; these four are what is left, and they are the entire content of the [lint job](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35325947613/job/105538902975). It reports them at lines 922, 1198, 1307 and 1359 rather than 899/1175/1281/1333 because CI lints the *merge* of this branch with master — I reproduced that exact set by merging into [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) and running `golangci-lint v2.13.2` on go1.27, the pair master's #2341 moved CI to. With the rename it reports 0 issues, which is what master at [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) reports on its own under the same linter and config.

exchanges/okx/okx_types.go
```diff
@@ -865,13 +869,14 @@ type CancelMassReqParam struct {
...
 type AmendOrderRequestParams struct {
-	InstrumentID    string  `json:"instId"`
-	CancelOnFail    bool    `json:"cxlOnFail,omitempty"`
-	OrderID         string  `json:"ordId,omitempty"`
-	ClientOrderID   string  `json:"clOrdId,omitempty"`
-	ClientRequestID string  `json:"reqId,omitempty"`
-	NewQuantity     float64 `json:"newSz,omitempty,string"`
-	NewPrice        float64 `json:"newPx,omitempty,string"`
+	InstrumentID     string  `json:"instId"`
+	InstrumentIDCode uint64  `json:"instIdCode,string,omitempty"`
```
thrasher-: Heimdall: **Blocker** — the `,string` makes the amend path send the code as a quoted string, while `order` and `cancel-order` send the same value unquoted.

**Fix:**

```suggestion
	InstrumentIDCode uint64  `json:"instIdCode,omitempty"`
```

**Why:** OKX types `instIdCode` as **Integer** on all six websocket order ops, and every example sends it bare — [WS / Amend order](https://www.okx.com/docs-v5/en/#order-book-trading-trade-ws-amend-order) has the row `> instIdCode | Integer | Yes | Instrument ID code` above a request example of `{"instIdCode": 123456, "ordId": "2510789768709120", "newSz": "2"}`. Here is what this branch actually puts on the wire for BTC-USDT (code 3), captured through an httptest server and a mock websocket server:

```
REST  POST trade/amend-order   {"instId":"BTC-USDT","instIdCode":"3","ordId":"111","newSz":"2"}
WS    amend-order              {"instId":"BTC-USDT","instIdCode":"3","ordId":"111","newSz":"2"}
WS    order                    {"instId":"BTC-USDT","instIdCode":3,"tdMode":"cash",...}
WS    cancel-order             {"instId":"BTC-USDT","instIdCode":3,"ordId":"111"}
```

The two shapes can't both be right, and amend is where it costs most. OKX's [2026-04-07 changelog](https://www.okx.com/docs-v5/log_en/#2026-04-07) deprecated `instId` on `amend-order`, `batch-amend-orders`, `cancel-order` and `batch-cancel-orders` — "To reduce latency in WS order operations… `instId` | String | - | Instrument ID | Deprecated, will be ignored" — so on those four channels the code is the only instrument identifier the server still reads. The place ops went first, on [2026-03-26](https://www.okx.com/docs-v5/log_en/#2026-03-26): "Delisted `instId` request parameter from the following channels. WS / Place order, WS / Place multiple orders… `instId` | String | - | Instrument ID | Delisted. Any value provided will be ignored." `instId` is no longer a row in any of the six request tables.

This also lines the field up with [`PlaceOrderRequestParam`](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_types.go#L773), which already has plain `instIdCode,omitempty`.

I have no OKX credentials, so I can't show you the rejection — if you've had a 200 from a websocket amend on this branch, say so and I'll drop this to a nit.

exchanges/okx/okx_types.go
```diff
@@ -852,9 +855,10 @@ func (r *ResponseResult) Error() error {
 
 // CancelOrderRequestParam represents order parameters to cancel an order
 type CancelOrderRequestParam struct {
-	InstrumentID  string `json:"instId"`
-	OrderID       string `json:"ordId"`
-	ClientOrderID string `json:"clOrdId,omitempty"`
+	InstrumentID     string `json:"instId"`
+	InstrumentIDCode uint64 `json:"instIdCode"`
```
thrasher-: Heimdall: **Should fix** — without `omitempty` this sends `"instIdCode":0` on every cancel where the lookup misses.

**Fix:**

```suggestion
	InstrumentIDCode uint64 `json:"instIdCode,omitempty"`
```

**Why:** measured on this branch with an empty instrument cache, through an httptest server and a mock websocket server:

```
REST  POST trade/cancel-order          {"instId":"BTC-USDT","instIdCode":0,"ordId":"111"}
REST  POST trade/cancel-batch-orders   [{"instId":"BTC-USDT","instIdCode":0,"ordId":"111"}]
WS    cancel-order                     {"instId":"BTC-USDT","instIdCode":0,"ordId":"111"}
WS    batch-cancel-orders              [{"instId":"BTC-USDT","instIdCode":0,"ordId":"111"}]
```

Zero is not an instrument, and `instIdCode` is not a documented request parameter on any REST order endpoint — [POST / Cancel order](https://www.okx.com/docs-v5/en/#order-book-trading-trade-post-cancel-order) lists only `instId`, `ordId` and `clOrdId`. With `omitempty` the REST bodies fall back cleanly to `instId` alone, and the three request structs finally agree.

Worth considering alongside it: even with `omitempty`, a populated cache still puts an undocumented `instIdCode` into every REST order body. Resolving the code only on the websocket path — where it's required — would keep REST bodies exactly as they are on master.

exchanges/okx/ws_requests.go
```diff
@@ -29,6 +29,10 @@ func (e *Exchange) WSPlaceOrder(ctx context.Context, arg *PlaceOrderRequestParam
 		return nil, err
 	}
 
+	if arg.InstrumentIDCode == 0 {
```
thrasher-, line 32: Heimdall: **Should fix** — the two place ops require a non-zero code; the four cancel and amend ops don't, and those are the channels where OKX has already stopped reading `instId`.

Keeping the check out of the shared `Validate()` was the right split — REST doesn't need a code, so it shouldn't be forced to carry one.

**Fix:** add the same guard to `WSCancelOrder`, `WSCancelMultipleOrders`, `WSAmendOrder` and `WSAmendMultipleOrders`. Put it **after** the existing order-ID checks, and in the amend ops after the size/price check too — anywhere earlier changes which error comes back first. I tried it before the order-ID checks and broke four assertions in `ws_requests_test.go`; between the order-ID and size/price checks it still breaks two.

**Why:** with an empty cache the wrapper still emits these, measured against a mock websocket server:

```
amend-order          {"instId":"BTC-USDT","ordId":"111","newSz":"2"}            <- no code at all
cancel-order         {"instId":"BTC-USDT","instIdCode":0,"ordId":"111"}
batch-cancel-orders  [{"instId":"BTC-USDT","instIdCode":0,"ordId":"111"}]
order                <blocked by errMissingInstrumentIDCode>
```

OKX's [2026-04-07 changelog](https://www.okx.com/docs-v5/log_en/#2026-04-07) marks `instId` on exactly those four channels "Deprecated, will be ignored", so the first three frames carry nothing the server acts on. Failing fast the way `WSPlaceOrder` does is better than a silent no-op amend.

In the engine this is a narrow path — `Bootstrap` fills the cache for every enabled asset through `UpdateOrderExecutionLimits`, so it bites library callers, supported-but-disabled assets, and instruments listed since the last refresh. That's why it's a should-fix rather than a blocker. With the guard added, all four return `errMissingInstrumentIDCode` instead, and `exchanges/okx` stays green under `-race` and `-tags sonic_on`.

**Related, same validation block:** all six methods still require a non-empty `InstrumentID`, so a caller who follows OKX's current docs and sends only a code is refused with `missing instrument ID`. The wrapper always sets both, so this only affects direct library callers — but it's worth deciding now whether `instId` should become optional when a code is present, rather than after someone hits it.

exchanges/okx/okx_types.go
```diff
@@ -83,6 +83,7 @@ var (
 	errIndexComponentNotFound               = errors.New("unable to fetch index components")
 	errLimitValueExceedsMaxOf100            = errors.New("limit value exceeds the maximum value 100")
 	errMissingInstrumentID                  = errors.New("missing instrument ID")
+	errMissingInstrumentIDCode              = errors.New("missing instrument ID Code")
```
thrasher-: Heimdall: **Should fix** — the new sentinel has no test, and the live-order tests build their requests without a code, so they'd fail before the order is sent.

**Fix:** add a `require.ErrorIs` for each op next to the existing validation assertions, for example in `TestWSPlaceOrder`:

```go
	_, err = e.WSPlaceOrder(t.Context(), &PlaceOrderRequestParam{
		InstrumentID: mainPair.String(), TradeMode: TradeModeCash, Side: "buy", OrderType: orderLimit, Amount: 1,
	})
	require.ErrorIs(t, err, errMissingInstrumentIDCode)
```

and the same in `TestWSPlaceMultipleOrders`, `TestWSCancelOrder`, `TestWSCancelMultipleOrders`, `TestWSAmendOrder` and `TestWSAmendMultipleOrders`. In the two amend tests, reuse the `NewPrice` the live request already sets rather than introducing a `NewQuantity` — setting one here leaks into the live amend below and silently changes the order size. Then give the live requests a real code: `e.getInstrumentsForAsset(ctx, asset.Spot)` followed by `e.getInstrumentIDCode(asset.Spot, mainPair.String())`.

**Why:** the guidelines ask for [direct `errors.Is` coverage when a sentinel is returned](https://github.com/thrasher-corp/gocryptotrader/blob/c6e47c6e1e6d6f7353148379d587188ec47c0367/docs/CODING_GUIDELINES.md#L208), and nothing exercises this path today. The live half is latent rather than active — `canManipulateRealOrders` is a `false` constant ([okx_test.go:46](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_test.go#L46)) so all six skip even with credentials — but [`out` in `TestWSPlaceOrder`](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/ws_requests_test.go#L26-L34) has no `InstrumentIDCode`, so the moment anyone flips that constant to exercise live orders, `WSPlaceOrder` returns `errMissingInstrumentIDCode` and the `require.NoError` under it fails. (`testexch.SetupWs` has already connected by then, so it's the order request that never goes out, not the connection.) Worth fixing in the same change that makes the code mandatory.

I added the six assertions locally and `exchanges/okx` stays green under `-race`, `-tags sonic_on` and `golangci-lint v2.13.2`. I stopped there — I haven't written the live-request half, since picking the right instrument and asset for it is your call.

While you're in there, `getInstrumentIDCode` itself has no direct test either: its swap, margin, options and unsupported-asset branches are all uncovered, and nothing asserts the outbound JSON tags. A table test over the five asset types plus a body assertion would close both.

thrasher-: Heimdall: **Minor** — the capital C reads oddly next to its neighbour `errMissingInstrumentID`, "missing instrument ID".

**Fix:**

```suggestion
	errMissingInstrumentIDCode              = errors.New("missing instrument ID code")
```

exchanges/okx/okx_wrapper.go
```diff
@@ -3076,3 +3085,31 @@ func (e *Exchange) MessageID() string {
 	hex.Encode(buf[:], u[:])
 	return string(buf[:])
 }
+
+// getInstrumentIDCode returns the instrumentIDCode for a given asset type and instrument ID add by dazi
```
thrasher-: Heimdall: **Minor** — the attribution is left over from drafting, and the GoDoc doesn't say what a miss returns.

**Fix:**

```suggestion
// getInstrumentIDCode returns the OKX instrument ID code for an asset type and instrument ID, or zero if it is not cached
```

**Why:** zero is what callers have to handle — it's what [`WSPlaceOrder`](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/ws_requests.go#L32) now rejects on — so it belongs in the doc.

exchanges/okx/okx_wrapper.go
```diff
@@ -3076,3 +3085,31 @@ func (e *Exchange) MessageID() string {
...
+func (e *Exchange) getInstrumentIDCode(a asset.Item, instID string) uint64 {
```
thrasher-: Heimdall: **Minor** — the lookup is a linear scan over the whole instrument list, on the order submit, amend and cancel paths.

**Fix, if you want it:** build a `map[string]uint64` keyed by `instId` alongside the slice in `getInstrumentsForAsset` ([okx_wrapper.go:2205](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_wrapper.go#L2205)) and read from that. Spot and margin can share one map — I checked all 266 live margin pairs against spot and the codes match in every case, BTC-USDT being 3 in both. I haven't put a suggestion block on it because where the map lives and how it's invalidated is your call.

**Why:** benchmarked against OKX's live instrument lists, three runs at `-benchtime 3s`:

```
BenchmarkLookupSpotLast-8    44346 ns/op   0 allocs/op   (1,411 spot instruments, last entry)
BenchmarkLookupSpotMiss-8    43577 ns/op   0 allocs/op   (full scan, no match)
BenchmarkMapLookup-8            20.03 ns/op              (map keyed by instId)
```

The absolute numbers move a lot with machine load — a second opinion measured ~105 µs and ~32 ns on the same code — so take it as three orders of magnitude rather than an exact figure. `CancelBatchOrders` pays it once per order ([okx_wrapper.go:1333](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_wrapper.go#L1333)), and OKX's stated reason for introducing `instIdCode` was "[to reduce latency in WS order operations](https://www.okx.com/docs-v5/log_en/#2026-04-07)". That said, a microbenchmark of the lookup doesn't prove a matching improvement in end-to-end order latency, so this stays a minor.

Related, and cheap: [`SubmitOrder`](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_wrapper.go#L899) resolves the code at the top, but the six algo-order branches below ([from okx_wrapper.go:989](https://github.com/thrasher-corp/gocryptotrader/blob/4c1bff3ff7a4c658ff941809645c6bd0c19d5573/exchanges/okx/okx_wrapper.go#L989)) no longer use it — correctly, since `AlgoOrderParams` doesn't carry it — so every trigger, TP/SL, chase, trailing-stop, TWAP and conditional order pays a full scan for a value that's discarded. Resolving it inside the one branch that needs it would fix that. (`asset.Spread` is already free: it hits the `default` and returns before the scan.)

exchanges/gateio/gateio_types.go
```diff
@@ -1891,7 +1891,7 @@ type FuturesOrder struct {
 	MakerFeeRate              types.Number  `json:"mkfr"`
 	TakerFeeRate              types.Number  `json:"tkfr"`
 	TimeInForce               string        `json:"tif"`
-	ReferenceUserID           int64         `json:"refu"`
+	ReferenceUserID           types.Number  `json:"refu"`
```
thrasher-: Heimdall: **Minor** — a question rather than a finding: was this one forced?

I couldn't find any payload that sends `refu` as a string. Gate's SDK declares it `Refu int32` ([model_futures_order.go:59](https://github.com/gateio/gateapi-go/blob/44355712305a5ea00490fae347322dc99a68b6ca/model_futures_order.go#L59)), every fixture in the package has it as a bare number, and master's `int64` decodes all of those. That's unlike `User` two fields up, where the REST and websocket shapes genuinely disagree and I could reproduce the break.

Keeping it symmetrical with `User` is a perfectly good answer — `types.Number` accepts everything `int64` did. I just didn't want it to look verified when it isn't.

thrasher- (changes requested):
Heimdall: Both halves of this are right, and they held up when I checked them independently: the Gate verbs match Gate's own SDK (`GetPosition` a [`GET`](https://github.com/gateio/gateapi-go/blob/44355712305a5ea00490fae347322dc99a68b6ca/api_futures.go#L1753), `SetDualMode` a [`POST` with `dual_mode` in the query](https://github.com/gateio/gateapi-go/blob/44355712305a5ea00490fae347322dc99a68b6ca/api_futures.go#L2367)), `types.Number` on `FuturesOrder.User` fixes a REST break that exists on master today, and `getInstrumentIDCode` resolved the correct code for all 2,379 live OKX instruments I fed it. The evidence for each is in the comment on that line.

Three round-one items came through only partly, and CI still has all three:

```
                          this PR 4c1bff3         control
exchanges/gateio  -race   FAIL 3 tests            ok        (base ce7da89)
golangci-lint v2.13.2     4 × revive var-naming   0 issues  (master c6e47c6)
```

**To unblock:**
1. Ten `futures.orders` fixtures still carry Gate's `"110xxxxx"` placeholder, which `types.Number` can't parse. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894137))
2. The four `instIdCode` local variables are the last of the 14 revive errors. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894139))
3. `,string` sends the amend code as `"3"` where OKX documents an Integer — and where `order` and `cancel-order` in this same PR send `3`. ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894146))

Three should-fixes and four minors are inline and none of them blocks: the missing `omitempty` on the cancel code ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894147)), the place ops requiring a code where cancel and amend don't ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894150)), the untested new sentinel ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894153)), and four small ones ([error string](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894154), [GoDoc](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894158), [linear scan](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894161), [`ReferenceUserID`](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4056894163)).

I applied the lot locally: `go test ./...` green across all 163 packages, gateio and okx green under `-race` and `-tags sonic_on`, `golangci-lint v2.13.2` clean, `TestAllExchangeWrappers/okx_wrapper_tests` passing.

**Still outside this diff** — both already on [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367), neither caused nor fixed here: `CancelAllOrders` builds its cancel requests without `instId`, so it fails locally with `missing instrument ID` whenever there is anything to cancel; and `ModifyOrder` never sets `NewPrice` for regular orders, so a price change is dropped. Happy for them to land here, or say the word and I'll open an issue with the tested fix.

exchanges/okx/okx_wrapper.go
```diff
@@ -1437,47 +1454,57 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
-	cancelAllOrdersRequestParams := make([]CancelOrderRequestParam, len(myOrders))
+	cancelAllOrdersRequestParams := make([]CancelOrderRequestParam, 0, len(myOrders))
 ordersLoop:
 	for x := range myOrders {
 		switch {
 		case orderCancellation.OrderID != "" || orderCancellation.ClientOrderID != "":
 			if myOrders[x].OrderID == orderCancellation.OrderID ||
 				myOrders[x].ClientOrderID == orderCancellation.ClientOrderID {
-				cancelAllOrdersRequestParams[x] = CancelOrderRequestParam{
+				cancelAllOrdersRequestParams = append(cancelAllOrdersRequestParams, CancelOrderRequestParam{
```
thrasher-, line 1464: Heimdall: **Blocker** — this branch cancels the wrong order outright when the orders carry no client order ID, and then stops looking.

**Fix** — at [okx_wrapper.go:1462-1463](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1462-L1463), only compare an identifier the caller actually supplied:

```go
			if (orderCancellation.OrderID != "" && myOrders[x].OrderID == orderCancellation.OrderID) ||
				(orderCancellation.ClientOrderID != "" && myOrders[x].ClientOrderID == orderCancellation.ClientOrderID) {
```

**Why:** with `ClientOrderID` unset on the request — the normal case — `myOrders[x].ClientOrderID == orderCancellation.ClientOrderID` is `"" == ""` for every order that has no `clOrdId`, so the first order in the list always matches. `break ordersLoop` then means the order you actually named is never reached. Two pending orders, neither with a `clOrdId`, asking for the second one:

```
requested      &order.Cancel{OrderID: "CANCEL-ME"}

on the wire    [{"instId":"BTC-USDT","ordId":"KEEP-ME"}]
returned       err=<nil>, Status map[KEEP-ME:CANCELLED]
```

It cancels the order you meant to keep, leaves the one you named open, and reports success.

Like the [side filter below](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791334), the predicate is inherited from master and has never run there — the same probe at [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) stops with `missing instrument ID`, because master omits the `instId` you've added on the line below. Adding it is the right fix; it just means both predicates in this switch now decide real cancellations.

exchanges/okx/okx_wrapper.go
```diff
@@ -1437,47 +1454,57 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
 					OrderID:       myOrders[x].OrderID,
 					ClientOrderID: myOrders[x].ClientOrderID,
-				}
+				})
 				break ordersLoop
 			}
 		case orderCancellation.Side == order.Buy || orderCancellation.Side == order.Sell:
 			if myOrders[x].Side == order.Buy || myOrders[x].Side == order.Sell {
-				cancelAllOrdersRequestParams[x] = CancelOrderRequestParam{
+				cancelAllOrdersRequestParams = append(cancelAllOrdersRequestParams, CancelOrderRequestParam{
```
thrasher-, line 1473: Heimdall: **Blocker** — the side filter never compares against the requested side, so cancelling "all buys" cancels the sells too.

**Fix** — one line above this hunk, at [okx_wrapper.go:1472](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1472):

```go
			if myOrders[x].Side == orderCancellation.Side {
```

**Why:** `myOrders[x].Side == order.Buy || myOrders[x].Side == order.Sell` is true for every real order, so this case behaves exactly like `default`. Measured against an httptest OKX returning one buy and one sell, asking only for buys:

```
requested      &order.Cancel{AssetType: asset.Spot, Pair: BTC-USDT, Side: order.Buy}

on the wire    [{"instId":"BTC-USDT","ordId":"BUY-1"},{"instId":"BTC-USDT","ordId":"SELL-1"}]
with the fix   [{"instId":"BTC-USDT","ordId":"BUY-1"}]
```

The condition is inherited from master, but it has never run there. The same test at [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) stops earlier:

```
    Error:  Received unexpected error:
            missing instrument ID
```

— because master builds these params without `instId`, which is exactly the [`CancelAllOrders` finding](https://github.com/thrasher-corp/gocryptotrader/pull/2392#pullrequestreview-5260523328) you've now fixed on [okx_wrapper.go:1474](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1474). Good fix — it's what turns this branch from dead code into a live order-cancellation path, and that's why the filter matters now.

`exchanges/okx` stays green with the fix applied (`go test ./exchanges/okx/... -count 1`), which also says nothing exercises this branch today. Worth a test in the same change.

exchanges/okx/okx_wrapper.go
```diff
@@ -1437,47 +1454,57 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
-			}
+			})
+		}
+	}
+	useWebsocket := e.Websocket.CanUseAuthenticatedWebsocketForWrapper()
+	if useWebsocket {
+		// Websocket cancels require the instrument ID code; the REST endpoint
+		// does not document instIdCode.
+		for x := range cancelAllOrdersRequestParams {
+			cancelAllOrdersRequestParams[x].InstrumentIDCode = e.getInstrumentIDCode(cancelAllOrdersRequestParams[x].InstrumentID)
```
thrasher-, line 1492: Heimdall: **Should fix** — resolving codes here gives the batch loop a new way to fail locally, and [okx_wrapper.go:1513](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1513) discards that failure whenever an earlier batch succeeded.

**Fix:** check for zeros here, where you already have the whole slice, rather than letting `WSCancelMultipleOrders` reject batch *n*:

```go
	if useWebsocket {
		for x := range cancelAllOrdersRequestParams {
			cancelAllOrdersRequestParams[x].InstrumentIDCode = e.getInstrumentIDCode(cancelAllOrdersRequestParams[x].InstrumentID)
			if cancelAllOrdersRequestParams[x].InstrumentIDCode == 0 {
				return cancelAllResponse, fmt.Errorf("%w: %s", errMissingInstrumentIDCode, cancelAllOrdersRequestParams[x].InstrumentID)
			}
		}
	}
```

**Why:** the loop returns an error only while `cancelAllResponse.Status` is still empty, so any batch after the first fails silently. 21 pending orders, second batch rejected:

```
returned err          <nil>
statuses reported     20 of 21
ORD-20 accounted for  false
```

The caller is told it succeeded and one order is still live. I triggered that with an exchange-side rejection because it is easy to stage, but your guard is a purely local source of exactly the same thing: one instrument listed since the last `UpdateOrderExecutionLimits` has no cached code, `WSCancelMultipleOrders` rejects the batch containing it, and the orders in that batch vanish from the result.

Failing up front also gives a far better message — the caller learns *which* instrument is missing, instead of getting a silently short status map.

exchanges/okx/okx_wrapper.go
```diff
@@ -3076,3 +3110,10 @@ func (e *Exchange) MessageID() string {
 	hex.Encode(buf[:], u[:])
 	return string(buf[:])
 }
+
+// getInstrumentIDCode returns the OKX instrument ID code for an instrument ID, or zero if it is not cached
+func (e *Exchange) getInstrumentIDCode(instID string) uint64 {
+	e.instrumentsInfoMapLock.Lock()
+	defer e.instrumentsInfoMapLock.Unlock()
+	return e.instrumentIDCodeMap[instID]
```
thrasher-, line 3118: Heimdall: **Should fix** — a miss is now fatal to websocket trading, and nothing in a running process can repair it.

**Fix:** cache on the `instruments` push. [okx_websocket.go:467](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_websocket.go#L467) already decodes into `WSInstrumentResponse`, whose `Data` is `[]Instrument` and so carries the codes:

```go
	case channelInstruments:
		var response WSInstrumentResponse
		if err := e.wsProcessPushData(ctx, respRaw, &response); err != nil {
			return err
		}
		e.cacheInstruments(response.Argument.InstrumentType, response.Data)
		return nil
```

**Why:** `getInstrumentsForAsset` is the only writer, and it runs from `Bootstrap` ([exchange.go:1826](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/exchange.go#L1826)) once at startup, plus `UpdateTradablePairs` which the engine only calls from the [RPC handler](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/engine/rpcserver.go#L2932). There is no periodic refresh. So an instrument listed after startup has no code for the life of the process, and every websocket place, cancel and amend on it returns `errMissingInstrumentIDCode` until someone restarts or calls `UpdateTradablePairs` by hand. The `instruments` channel is already delivering exactly the record that would fix it, and it is dropped.

Before this change a stale cache cost nothing, because the code was optional — that is what makes the refresh path worth having now rather than later. A REST fallback inside `getInstrumentIDCode` on a miss would do the same job if you'd rather not touch the websocket handler; it needs a `context.Context` and the asset, which all five call sites have.

exchanges/okx/okx_wrapper.go
```diff
@@ -2192,10 +2217,19 @@ func (e *Exchange) getInstrumentsForAsset(ctx context.Context, a asset.Item) ([]
...
+// cacheInstruments stores instruments by instrument type and indexes their
+// instrument ID codes by instrument ID for getInstrumentIDCode lookups
+func (e *Exchange) cacheInstruments(instType string, instruments []Instrument) {
 	e.instrumentsInfoMapLock.Lock()
+	defer e.instrumentsInfoMapLock.Unlock()
 	e.instrumentsInfoMap[instType] = instruments
-	e.instrumentsInfoMapLock.Unlock()
-	return instruments, nil
+	for x := range instruments {
+		e.instrumentIDCodeMap[instruments[x].InstrumentID.String()] = instruments[x].InstrumentIDCode
```
thrasher-, line 2231: Heimdall: **Should fix** — the new test seeds the map by hand, so this line, the one load-bearing line of the refactor, has no coverage.

**Why:** delete the loop entirely and the package is still green:

```
// cacheInstruments with the instrumentIDCodeMap loop removed
$ go test ./exchanges/okx/... -count 1
ok   github.com/thrasher-corp/gocryptotrader/exchanges/okx   8.343s
```

The cache is keyed on `Instrument.InstrumentID.String()` — a `currency.Pair` round-trip — while every read uses a different spelling: `pairFormat.Format(pair)` in [SubmitOrder](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L982), [CancelOrder](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1294) and [CancelBatchOrders](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1371), and a raw OKX `instId` off `OrderDetail.InstrumentID` in [CancelAllOrders](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1492). If those three ever disagree every code silently becomes 0 and all six websocket order ops start returning `errMissingInstrumentIDCode`.

They agree today. Feeding the live instrument lists through `cacheInstruments` and reading every entry back both ways:

```
             n      raw instId miss   pairFormat miss
SPOT       1411           0                 0
MARGIN      266           0                 0
SWAP        482           0                 0
FUTURES     225           0                 0
OPTION      734           0                 0
```

3,118 instruments, no misses, including `BTC-USD-251226-60000-C` and `BTC-USDT-SWAP`. A table test that pushes a handful of those shapes through `cacheInstruments` and reads them back through `getInstrumentIDCode` would pin it — that also covers the futures and options shapes, which the literal seed doesn't.

Two smaller things while you're here: [okx_wrapper_test.go:27](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper_test.go#L27) sets `instrumentsInfoMap`, which `getInstrumentIDCode` never reads; and the flat `map[string]uint64` is safe against cross-asset collisions — 266 instIds appear in both SPOT and MARGIN and the code is identical in every one (BTC-USDT = 3 in both) — so last-writer-wins costs nothing.

exchanges/okx/okx_wrapper.go
```diff
@@ -1353,6 +1365,11 @@ func (e *Exchange) CancelBatchOrders(ctx context.Context, o []order.Cancel) (*or
 	if len(cancelOrderParams) > 0 {
 		var canceledOrders []*OrderData
 		if e.Websocket.CanUseAuthenticatedWebsocketForWrapper() {
+			// Websocket cancels require the instrument ID code; the REST
+			// endpoint does not document instIdCode.
+			for x := range cancelOrderParams {
+				cancelOrderParams[x].InstrumentIDCode = e.getInstrumentIDCode(cancelOrderParams[x].InstrumentID)
```
thrasher-, line 1371: Heimdall: **Minor** — `CancelBatchOrders` is the only cancel path without an `asset.Spread` branch, and a spread ID can never resolve a code.

**Why:** `getInstrumentsForAsset` is the sole writer of the cache, and it is never reached for Spread — [FetchTradablePairs](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L307) and [UpdateOrderExecutionLimits](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L351) both route it to `GetPublicSpreads`, and calling it directly errors out:

```
SupportsAsset(Spread)                    true
spread instId "BTC-USDT-SWAP"            getInstrumentIDCode = 0
getInstrumentsForAsset(Spread)           invalid instrument type, empty instrument type
CancelBatchOrders(Spread)                builds a plain CancelOrderRequestParam — no Spread branch
```

[CancelOrder](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1269) and [CancelAllOrders](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1419) both special-case Spread and route to the `sprd` endpoints; this one doesn't, so over an authenticated websocket a spread order now comes back `errMissingInstrumentIDCode` instead of going out. It's reachable — `CancelBatchOrders` takes whatever asset the gRPC caller passes ([rpcserver.go:1420](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/engine/rpcserver.go#L1420)).

Minor because it was already broken: on master the frame went out to `batch-cancel-orders` carrying a spread ID, which isn't an instrument on that channel. Your guard changes the symptom rather than causing it. Routing Spread to `WSCancelSpreadOrder` here, the way its two siblings do, would close it properly — happy for that to be a separate issue if you'd rather not widen this PR.

exchanges/okx/okx_wrapper.go
```diff
@@ -976,6 +977,9 @@ func (e *Exchange) SubmitOrder(ctx context.Context, s *order.Submit) (*order.Sub
 			}
 		}
 		if e.Websocket.CanUseAuthenticatedWebsocketForWrapper() {
+			// Websocket order placement requires the instrument ID code; the
+			// REST endpoint does not document instIdCode.
```
thrasher-, line 981: Heimdall: **Nit** — this comment is repeated near-verbatim five times: here and at [okx_wrapper.go:1187](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1187), [:1293](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1293), [:1368](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1368) and [:1489](https://github.com/thrasher-corp/gocryptotrader/blob/e7567a7039d98dcbf479ba845a9f1086551747dc/exchanges/okx/okx_wrapper.go#L1489).

The *why* — that the code is resolved inside the websocket branch because no REST order endpoint documents `instIdCode` — is worth saying once. On `getInstrumentIDCode` itself it would cover all five sites, and the guidelines ask for [comments that explain why, not what](https://github.com/thrasher-corp/gocryptotrader/blob/c6e47c6e1e6d6f7353148379d587188ec47c0367/docs/CODING_GUIDELINES.md#L270). Keeping the code out of REST bodies was the right call and I've confirmed it holds — with the cache populated, REST still sends:

```
POST trade/order          {"instId":"BTC-USDT","tdMode":"cash","side":"buy","ordType":"limit","sz":"1","px":"20000"}
POST trade/cancel-order   {"instId":"BTC-USDT","ordId":"111"}
POST trade/amend-order    {"instId":"BTC-USDT","ordId":"111","newSz":"2","newPx":"21000"}
```

thrasher- (changes requested):
Heimdall: Everything from round two landed and I re-ran all of it: the ten Gate fixtures (three test groups failed at [`4c1bff3`](https://github.com/thrasher-corp/gocryptotrader/commit/4c1bff3ff7a4c658ff941809645c6bd0c19d5573) on `invalid value for Number type: 110xxxxx`, green now), the four `var-naming` renames, `,string` off the amend code, `omitempty` on cancel, and guards on all six websocket ops with matching `require.ErrorIs`. Both out-of-diff items were taken on too — `CancelAllOrders` now carries `instId`, `ModifyOrder` sends `newPx`. The map that replaced the linear scan resolves correctly for all 5,368 live instruments, and REST bodies stay free of `instIdCode`. Evidence is on the lines themselves.

**To unblock:**

1. **No CI has run on [`e7567a7`](https://github.com/thrasher-corp/gocryptotrader/commit/e7567a7039d98dcbf479ba845a9f1086551747dc) at all.** The PR conflicts with master, so GitHub has no merge ref to build: `mergeable_state: dirty`, `merge_commit_sha: null`, and 0 check runs where [`82fb0e9`](https://github.com/thrasher-corp/gocryptotrader/commit/82fb0e94722e4f12c094f939d0ef3a11fb768598), [`3add79d`](https://github.com/thrasher-corp/gocryptotrader/commit/3add79dc3ff0aedc2175f0df25d181e27d28b4be) and [`4c1bff3`](https://github.com/thrasher-corp/gocryptotrader/commit/4c1bff3ff7a4c658ff941809645c6bd0c19d5573) each have 14. The conflict is one import block in `exchanges/okx/okx_wrapper_test.go` — #2341 moved `github.com/gofrs/uuid` to Go 1.27's stdlib `uuid`, this PR added `stretchr/testify/assert`. Merged onto [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) locally it is clean: build, `golangci-lint v2.13.2` 0 issues, `go test ./...` green but for two ambient network failures (below), and okx and gateio green under `-race` and `-tags sonic_on`.
2. **`CancelAllOrders` selects the wrong orders in two of its three branches**, and adding `instId` is what makes them run — the same probes at [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) stop with `missing instrument ID`. Asked for one order ID it cancels a different order and reports success ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791325)); asked for buys it cancels the sells too ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791334)). Neither predicate is yours, but fixing the `instId` omission promotes both from dead code to live cancellations.

Three should-fixes and two smaller things inline, none of them gating: the batch loop [discards a failed batch](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791340) once an earlier one has succeeded, [nothing refreshes the code cache](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791342) in a running process, [`cacheInstruments` has no coverage](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791345), [`CancelBatchOrders` and `asset.Spread`](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791349), and [a comment repeated five times](https://github.com/thrasher-corp/gocryptotrader/pull/2392#discussion_r4059791351).

**Ambient, not this PR:** the 29 `golangci-lint` issues on the unmerged branch are byte-identical at the base [`ce7da89`](https://github.com/thrasher-corp/gocryptotrader/commit/ce7da894e85098ddb88ad4a627575f494b98c7ef) and vanish on merge (master's #2341 bumped the linter and cleared them). In the full run, `bitflyer/TestGetAddressInfoCA` is a live blockchain call that timed out after 66s, and a bybit failure passes 3/3 in isolation.

**Still your call, carried over from round two:** all six websocket ops require a non-empty `InstrumentID`, so a caller following OKX's current docs and sending only a code is refused with `missing instrument ID`.

I have no OKX credentials, so the `instIdCode` shapes here come from wire captures against local mocks and OKX's docs, not a live order. Happy to approve once those two are in.

PR #2397, opened 2026-09-21
Ticker: Migrate WS processing to exchanges

gloriousCode:
# PR Description

Moved this out of an existing PR to prevent poor reviewer fatigue. 
Tickers now processed within the `{{exchange}}_websocket.go` rather than in engine `websocketroutine_hanlder.go`. One of the last important things to get out of there that slows down that process.

### Only AI reads this section:
`ProcessBatch` visits every item and combines the errors, consistent with the `common.AppendError` guidance in PR #1480. Coinbase then fails the complete dispatch when any item fails, consistent with the review on this exact function in PR #2183 that one failure should fail the stack rather than produce repeated or partial output.

Ticker caching currently depends on the engine websocket receiver, so exchange websocket implementations do not update the ticker store when used outside that receiver.

Process ticker updates in each exchange handler before forwarding them to the data handler. Add `ticker.ProcessBatch` for handlers that produce multiple ticker values, and leave the engine receiver responsible for sync reporting.

Tests also cover batch processing and the receipt timestamps assigned when an exchange message does not contain an update time.

## Validation:
- Focused tests for the engine, ticker package and every changed exchange
- `make lint`
- `make misc_checks`


thrasher-:
@codex please review

exchanges/okx/okx_websocket.go
```diff
@@ -1317,13 +1321,13 @@ func (e *Exchange) wsProcessTickers(ctx context.Context, data []byte) error {
...
 				LastUpdated:  response.Data[i].TickerDataGenerationTime.Time(),
-			}
-			if err := e.Websocket.DataHandler.Send(ctx, tickData); err != nil {
-				return err
-			}
+			})
 		}
 	}
-	return nil
+	if err := ticker.ProcessBatch(tickerPrices); err != nil {
```
thrasher-, line 1327: Heimdall: **Blocker** — this relays an empty `[]ticker.Price` where master relayed nothing. kucoin guards the same case; okx doesn't.

**Fix:** the guard you already have at [kucoin_websocket.go:774](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket.go#L774) and [kucoin_websocket.go:954](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket.go#L954) — `if len(tickerPrices) == 0 { return nil }` — before the `ProcessBatch` here.

**Why:** `response.Data` yields no entries whenever it is `[]`, `null` or absent, and the send below is unconditional. Master's loop sent per asset, so it sent nothing:

```
okx wsProcessTickers, relay depth after each frame
                                  0678bbe   c6e47c6
  {"arg":{...},"data":[]}            1         0
  {"arg":{...}}  (no data key)       2         0
    dispatched []ticker.Price len=0
    dispatched []ticker.Price len=0
```

kucoin pairs its guard with [two untracked-pair assertions](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket_test.go#L1576); okx has neither the guard nor a test, and `testdata/wsTickers.json` carries no empty-`data` line — which is why CI is quiet about it. Worth a fixture line alongside the guard.

For what it's worth I checked the other `ProcessBatch` callers — [okx_websocket.go:752](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/okx/okx_websocket.go#L752), coinbase, both gateio and both poloniex sites all relay an empty batch on master too, so those are pre-existing rather than anything this PR introduced.

exchanges/kraken/kraken_websocket.go
```diff
@@ -388,7 +388,11 @@ func (e *Exchange) wsProcessTickers(ctx context.Context, dataRaw json.RawMessage
 		Open:         t.Open[0].Float64(),
 		AssetType:    asset.Spot,
 		Pair:         pair,
-	})
+	}
+	if err := ticker.ProcessTicker(tickPrice); err != nil {
```
thrasher-, line 392: Heimdall: **Blocker** — returning early on the store write means dispatcher back-pressure now drops valid tickers that master delivers. This shape is in every converted handler; kraken is just where I measured it.

**Fix:** report the store failure without letting it gate the relay:

```go
procErr := ticker.ProcessTicker(tickPrice)
return common.AppendError(procErr, e.Websocket.DataHandler.Send(ctx, tickPrice))
```

**Why:** `Service.update` publishes through the dispatch mux ([ticker.go:210](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/ticker/ticker.go#L210)), which returns `errDispatcherJobsAtLimit` once the hundred-job queue saturates ([dispatch.go:224](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/dispatch/dispatch.go#L224)). `subscriberCount` is global ([mux.go:61](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/dispatch/mux.go#L61)), so one subscription to any exchange's tickers puts every ticker on that queue. A thousand valid frames, one subscription to an unrelated exchange, producer and consumer on one goroutine so neither side can be starved:

```
                  c6e47c6 delivered   0678bbe delivered   jobs-at-limit both sides
GOMAXPROCS=1          1000/1000             55/1000                945
GOMAXPROCS=2          1000/1000           1000/1000                  0
GOMAXPROCS=4          1000/1000           1000/1000                  0
```

Identical pressure, identical error count; master delivers every frame regardless. The single-core row is deterministic over repeated runs. Two cores usually keep up, but not always — one run in six lost 316 of 1000 — and a second trigger reaches it at full GOMAXPROCS: a [GetExchangeTickerStream](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/engine/rpcserver.go#L2255) client that stops reading cost 333-502 of 1000 in five of eight runs.

With the `AppendError` shape above, GOMAXPROCS=1 returns to 1000/1000 with all 945 errors still reported, and the kraken package stays green under `-race`.

Master loses the same store write and the same mux publish under that pressure — what it does not lose is the relay. Here the relay goes too, and with it the syncer's ticker timestamp and [stats.Add](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/engine/sync_manager.go#L726).

exchanges/ticker/ticker.go
```diff
@@ -176,6 +176,15 @@ func ProcessTicker(p *Price) error {
 	return service.update(p)
 }
 
+// ProcessBatch processes a batch of tickers.
+func ProcessBatch(p []Price) error {
```
thrasher-, line 180: Heimdall: **Should fix** — `ProcessBatch` commits the valid entries and then every caller withholds the whole batch, so the store and the relay apply opposite policies to the same message.

**Fix:** pick one. Either stop at the first failure so nothing is committed, or let the callers dispatch what succeeded. Both are defensible; disagreeing is not.

**Why:** the helper deliberately continues past a failure — [ticker_test.go:472](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/ticker/ticker_test.go#L472) asserts it — while every caller returns before the send. Running the payload from your own new case at [okx_test.go:4745](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/okx/okx_test.go#L4745):

```
0678bbe   BTC-USDT IS in the store, bid=1 ask=2      DataHandler: 0 messages
c6e47c6   BTC-USDT not in the store                  DataHandler: 2 messages
```

So [okx_test.go:4747](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/okx/okx_test.go#L4747) asserts no partial batch is dispatched while the store receives exactly that. The relay half follows the review on this function in #2183; the store half contradicts it.

engine/websocketroutine_manager_test.go
```diff
@@ -172,6 +172,17 @@ func TestWebsocketRoutineManagerHandleData(t *testing.T) {
 		AssetType:    asset.Spot,
 	})
 	assert.NoError(t, err)
+	testPair := currency.NewPair(currency.NewCode("AAA"), currency.NewCode("BBB"))
+	err = exch.GetBase().CurrencyPairs.SetAssetEnabled(asset.Spot, false)
```
thrasher-: Heimdall: **Minor** — the disable isn't needed for the assertion below, and it isn't restored.

**Fix:** drop these two lines. The point being proven is that the engine no longer writes the store, which holds with spot enabled.

**Why:** removing them leaves the test passing:

```
go test ./engine/ -run '^TestWebsocketRoutineManagerHandleData$' -count 1
ok   github.com/thrasher-corp/gocryptotrader/engine   0.024s
```

Nothing downstream reads the asset here — this syncer was never started, so [PrintTickerSummary](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/engine/sync_manager.go#L706) returns on its first line. Spot then stays disabled for the remaining kline, futures, account, order and orderbook checks.

The assertion itself is a real regression test, for what it's worth — putting `ticker.ProcessTicker(d)` back at [websocketroutine_manager.go:244](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/engine/websocketroutine_manager.go#L244) fails it with `Expected error with "no ticker found" in chain but got nil`.

exchanges/kucoin/kucoin_websocket_test.go
```diff
@@ -1556,7 +1567,13 @@ func TestProcessTicker(t *testing.T) {
 		AssetType:    asset.Spot,
 		LastUpdated:  time.UnixMilli(1789624664495),
 	}
-	assert.Equal(t, exp, (<-ku.Websocket.DataHandler.C).Data, "processTicker should map the latest fill's size to LastSize rather than a volume")
+	got, ok := (<-ku.Websocket.DataHandler.C).Data.([]ticker.Price)
+	require.True(t, ok, "processTicker must send a ticker batch")
+	assert.Equal(t, []ticker.Price{exp}, got, "processTicker should map the latest fill's size to LastSize rather than a volume")
```
thrasher-, line 1574: Heimdall: **Minor** — both kucoin ticker tests pass with the store writes removed, so they cover the batch shape rather than the behaviour this PR adds.

**Fix:** assert the store — `ticker.GetTicker(ku.Name, pair, asset.Spot)` after the call here, and the same in [TestProcessMarketSnapshot](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket_test.go#L1626).

**Why:** deleting both `ticker.ProcessBatch` calls from [kucoin_websocket.go:777](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket.go#L777) and [kucoin_websocket.go:957](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket.go#L957), keeping the batch-shaped send:

```
go test ./exchanges/kucoin/ -run '^TestProcessTicker$|^TestProcessMarketSnapshot$' -count 1
ok   github.com/thrasher-corp/gocryptotrader/exchanges/kucoin   0.770s
```

Worth pairing with `ku.Name = t.Name()`, since these now write the package-global ticker store for the first time under the shared default name. Nothing collides today — I checked the okx and kraken equivalents — but the surface is new.

One caveat if you sweep the others: bitfinex can't be renamed. The crossed-book exemption at [ticker.go:146](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/ticker/ticker.go#L146) matches the exchange name exactly, and case-sensitively, so a renamed instance loses it for every margin funding ticker:

```
same crossed funding book, bid 5.5 / ask 2.2, asset marginfunding
  ExchangeName "Bitfinex"   -> nil
  ExchangeName t.Name()     -> bid greater than ask this is a crossed or locked market
  ExchangeName "bitfinex"   -> bid greater than ask this is a crossed or locked market
```

exchanges/kucoin/kucoin_websocket_test.go
```diff
@@ -141,7 +141,18 @@ func TestPushData(t *testing.T) {
...
+		for len(e.Websocket.DataHandler.C) > 0 {
+			response := <-e.Websocket.DataHandler.C
+			if prices, ok := response.Data.([]ticker.Price); ok {
+				messageCount += len(prices)
+				continue
+			}
+			messageCount++
+		}
+		return messageCount == 31
+	}, time.Second, time.Millisecond*10, "must receive 31 messages")
```
thrasher-, line 156: Heimdall: **Minor** — the counter sums entries, so an empty batch is invisible to the one assertion here broad enough to catch one.

**Fix:** count envelopes as well, or add `require.NotEmpty(t, prices, ...)` inside the type switch above.

**Why:** `messageCount += len(prices)` contributes zero for an empty batch, so the total still reaches 31. Deleting both `len(tickerPrices) == 0` guards from [kucoin_websocket.go](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket.go#L774) and changing nothing else:

```
go test ./exchanges/kucoin/ -run '^TestPushData$|^TestProcessTicker$|^TestProcessMarketSnapshot$' -count 1
  --- PASS: TestPushData
  --- FAIL: TestProcessTicker            wsHandleData should not send an empty ticker batch
  --- FAIL: TestProcessMarketSnapshot    processMarketSnapshot should not send an empty ticker batch
```

The two targeted assertions hold the line, so nothing is lost today. It matters because this is the only test here that drives the whole fixture — it is what would catch a site nobody thought to write a targeted test for, and [okx_websocket.go:1327](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/okx/okx_websocket.go#L1327) is exactly that site.

thrasher- (changes requested):
Heimdall: The move is sound and the coverage is complete — 23 websocket functions dispatched a ticker without writing the store on [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367), none do now. The engine assertion is a real regression test, and kucoin's empty-batch guards and untracked-pair assertions both bite.

Two blockers; the evidence is on the lines.

**To unblock:**

1. [okx_websocket.go:1327](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/okx/okx_websocket.go#L1327) needs the guard kucoin has — it still relays empty ticker batches.
2. Stop the store write gating the relay at [kraken_websocket.go:392](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kraken/kraken_websocket.go#L392) and the other converted handlers.

Four non-gating comments: [ticker.go:180](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/ticker/ticker.go#L180), [kucoin_websocket_test.go:155](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket_test.go#L155), [websocketroutine_manager_test.go:176](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/engine/websocketroutine_manager_test.go#L176), [kucoin_websocket_test.go:1572](https://github.com/thrasher-corp/gocryptotrader/blob/0678bbefdcdd04c91f74c8e05d76c06eee349258/exchanges/kucoin/kucoin_websocket_test.go#L1572).

The four [CI](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35568466433) failures are ambient — COINUT 520s and timeouts, a momentarily locked GateIO SUI_USDT book, and a pool-loan test expecting a 504 — all in code this PR doesn't touch, and all eighteen changed packages pass locally under `-race`.

Happy to approve once the two land.

exchanges/kraken/kraken_websocket.go
shazbert: disagree because deploying to process ticker validates it

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -543,7 +543,11 @@ func (e *Exchange) processFuturesTickerV2(ctx context.Context, respData []byte)
 		Bid:          resp.BestBidPrice.Float64(),
 		AskSize:      resp.BestAskSize.Float64(),
 		BidSize:      resp.BestBidSize.Float64(),
-	})
+	}
+	if err := ticker.ProcessTicker(tickPrice); err != nil {
+		return err
+	}
+	return e.Websocket.DataHandler.Send(ctx, tickPrice)
```
shazbert, line 550: optional: Any chance we can have a regression check to test all websocket handling that it has been loaded into memory and that it has been sent into the channel.

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -764,11 +769,15 @@ func (e *Exchange) processTicker(ctx context.Context, respData []byte, instrumen
...
-		}); err != nil {
-			return err
-		}
+		})
 	}
-	return nil
+	if len(tickerPrices) == 0 {
+		return nil
+	}
+	if err := ticker.ProcessBatch(tickerPrices); err != nil {
```
shazbert, line 777: Can we update coding guidelines that enforce this standard for websocket tickers?

exchanges/kucoin/kucoin_websocket.go
```diff
@@ -939,11 +949,15 @@ func (e *Exchange) processMarketSnapshot(ctx context.Context, respData []byte, t
...
 			LastUpdated:  response.Data.Datetime.Time(),
-		}); err != nil {
-			return err
-		}
+		})
 	}
-	return nil
+	if len(tickerPrices) == 0 {
+		return nil
+	}
```
shazbert, line 956: Can this be an error?

exchanges/ticker/ticker.go
```diff
@@ -176,6 +176,15 @@ func ProcessTicker(p *Price) error {
...
+	var errs error
+	for i := range p {
+		errs = common.AppendError(errs, ProcessTicker(&p[i]))
```
shazbert: I would have a seperate function call that has one lock and processes all under that lock. More efficient than staggering.

shazbert (changes requested):
This is a great improvement and you deserve the best of what life gives you. 🕊️

exchanges/kraken/kraken_websocket.go
gloriousCode: Yes shazbert is right in that we used to send all data through the handler and then validate it. Now we're validating before sending.
I think as well, that this ticker change is following our precedent for websocket order books. We only pass on order books that are valid. Yes we could argue for passing all WS updates through the handler, but that itself has caused issue of having too much unnecessary, unprocessed data flow through it - slowing WS handling. It is a bias, it can lead to less data sent, but this handling will highlight whether there are issues with tickers too - now we properly error instead of sending it to a processor to just log the error

I'll still investigate further before getting a robotReview again though

exchanges/ticker/ticker.go
gloriousCode: Lovely suggestion

exchanges/kucoin/kucoin_websocket_test.go
```diff
@@ -1609,50 +1631,66 @@ func TestProcessFuturesTickerV2(t *testing.T) {
 func TestProcessMarketSnapshot(t *testing.T) {
 	t.Parallel()
 	ku := testInstance(t)
+	ku.Name = t.Name()
```
thrasher-, line 1634: Heimdall: **Minor** — [`TestProcessFuturesTickerV2`](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/kucoin/kucoin_websocket_test.go#L1586), just above, needs the same rename this test and `TestProcessTicker` got. It now writes the global store under the shared name `Kucoin`, and its first case stores SOL_USDTM futures with `Last` 0. SOL_USDTM is an enabled futures pair that `TestUpdateTickers` reads back under the same name, in parallel, and asserts positive ([kucoin_test.go:2215](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/kucoin/kucoin_test.go#L2215)).

**Fix:** `ku.Name = t.Name()` after `testInstance` at [kucoin_websocket_test.go:1586](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/kucoin/kucoin_websocket_test.go#L1586).

**Why:** replaying its two cases around `TestUpdateTickers`' REST write and read:

```
                          REST write   after first case   after second case
shared name "Kucoin"        115.125           0                99.892
  -> FAIL: futures SOL_USDTM Tick Last should be positive
with ku.Name = t.Name()     115.151        115.151            115.151
```

The second case restores a positive price 94-805µs later, so the window is under a millisecond and I'd expect it to fire rarely, but it is the one real overlap. I listed every websocket test that now writes the store. The others either have no reader asserting on their entries, or, like huobi's, write values that still pass the reader's check.

docs/CODING_GUIDELINES.md
```diff
@@ -28,6 +28,11 @@ See [SECURITY.md](../SECURITY.md) for the project's security policy, supported v
 
 Refer to the [ADD_NEW_EXCHANGE.md](../docs/ADD_NEW_EXCHANGE.md) document for comprehensive steps on integrating a new exchange.
 
+Websocket ticker handlers must process tickers into the shared store before
+relaying them to `DataHandler`. Do not relay a ticker that fails processing.
+For batches, relay the successfully processed entries and report any failures;
+skip empty batches from feeds that can include untracked markets.
```
thrasher-, line 34: Heimdall: **Minor** — one condition at the nine `ProcessBatch` callers would make this rule unconditional: `if len(processed) == 0 { return err }` in place of `if err != nil && len(processed) == 0 { return err }`.

**Fix:** that condition at every caller; drop the separate guards at [okx_websocket.go:1328](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/okx/okx_websocket.go#L1328), [kucoin_websocket.go:774](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/kucoin/kucoin_websocket.go#L774) and [kucoin_websocket.go:955](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/kucoin/kucoin_websocket.go#L955); and this line becomes "never relay an empty batch".

**Why:** an empty input leaves `processed` empty and `err` nil, which the current condition relays. The six callers without a guard of their own still do:

```
                                     empty batches relayed
                                     dd86392   with len(processed) == 0
gateio spot, untracked pair             1          0
gateio futures, empty result            1          0
poloniex spot, empty data               1          0
poloniex futures, empty data            1          0
coinbase, product with no alias         1          0
okx spread, empty data                  1          0
```

Those are [gateio_websocket.go:284](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/gateio/gateio_websocket.go#L284), [gateio_websocket_futures.go:502](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/gateio/gateio_websocket_futures.go#L502), [poloniex_websocket.go:339](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/poloniex/poloniex_websocket.go#L339), [poloniex_futures_websocket.go:445](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/poloniex/poloniex_futures_websocket.go#L445), [coinbase_websocket.go:141](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/coinbase/coinbase_websocket.go#L141) and [okx_websocket.go:753](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/okx/okx_websocket.go#L753). All six relayed empty batches on [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) too, so this is optional. The one condition covers both the all-failed and the empty cases, spares each author judging which feeds "can include untracked markets", and nets nine fewer lines. With it applied, coinbase, gateio, kucoin, okx and poloniex all pass under `-race`, including the okx `[]`/`null` and kucoin untracked-pair assertions the three guards were added for.

exchanges/ticker/ticker_test.go
```diff
@@ -414,6 +416,180 @@ func TestProcessTicker(t *testing.T) { // non-appending function to tickers
...
+	t.Parallel()
+	for _, tc := range []struct {
+		name     string
+		price    *Price
+		wantErr  error
+		wantText string
+	}{
+		{name: "nil", wantErr: errNilTickerPrice},
+		{name: "missing exchange", price: &Price{Pair: currency.NewBTCUSD(), AssetType: asset.Spot}, wantErr: common.ErrExchangeNameNotSet},
+		{name: "missing pair", price: &Price{ExchangeName: "test", AssetType: asset.Spot}, wantText: errPairNotSet},
```
thrasher-, line 539: Heimdall: **Nit** — `errPairNotSet` and `errAssetTypeNotSet` are still string constants ([ticker_types.go:16-17](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/ticker/ticker_types.go#L16-L17)), which is why these two cases need `wantText`.

**Fix:** declare both beside `errNilTickerPrice` as sentinels, as the orderbook package already does ([orderbook_types.go:31-32](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/orderbook/orderbook_types.go#L31-L32)), and wrap them with `%w` at [ticker.go:149](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/ticker/ticker.go#L149) and [ticker.go:177](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/ticker/ticker.go#L177). Every case here can then use `wantErr`, the `wantText` field and its branch go, and [ticker_test.go:590](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/ticker/ticker_test.go#L590) can assert `common.ErrNilPointer` rather than any error.

**Why:** the [coding guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/docs/CODING_GUIDELINES.md#L135) prefer sentinels for validation failures that tests match. The error strings stay identical, nothing outside the ticker package uses either constant, and with the change the package passes under `-race` with lint clean.

thrasher- (approved):
Heimdall: Approving. The okx empty-batch guard, the relay-what-processed batch policy and all three test fixes landed, and each one fails its test when reverted. dd86392's one-lock `ProcessBatch` holds under `-race -count 5`, and its dispatch-failure subtest bites. At [`dd86392`](https://github.com/thrasher-corp/gocryptotrader/commit/dd86392fbf220f471e11f072482bb60672ab1247), all 25 ticker-typed relays across the exchanges process the ticker first. On [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367), 24 of those 25 did not.

Inline, none blocking: Minors on [kucoin_websocket_test.go:1634](https://github.com/thrasher-corp/gocryptotrader/pull/2397#discussion_r4089340623), the rename one sibling test missed, and on [CODING_GUIDELINES.md:34](https://github.com/thrasher-corp/gocryptotrader/pull/2397#discussion_r4089340627), plus a Nit on [ticker_test.go:539](https://github.com/thrasher-corp/gocryptotrader/pull/2397#discussion_r4089340631). I've conceded the policy in the [back-pressure thread](https://github.com/thrasher-corp/gocryptotrader/pull/2397#discussion_r4060272237).

The failing CI jobs are ambient. One is a Windows timing assertion in `exchanges/request` (11ms against a 20ms floor). The other is a Coinbase futures `UpdateOrderbook` 404, "no pricebook found". Neither is in code this PR touches. codecov/project follows from them, since its upload step only runs after a passing test step.

The description's Coinbase paragraph still says it "fails the complete dispatch when any item fails", which [`4c04a45`](https://github.com/thrasher-corp/gocryptotrader/commit/4c04a45aeaafc417947295201f69f194f1daa1dd) changed.

exchanges/kraken/kraken_websocket.go
thrasher-: Heimdall: Conceded. Validating before relaying is your call. Bybit's `wsProcessPublicTicker` already worked that way on master ([bybit_websocket.go:644](https://github.com/thrasher-corp/gocryptotrader/blob/c6e47c6e1e6d6f7353148379d587188ec47c0367/exchanges/bybit/bybit_websocket.go#L644)), and dd86392's dispatch-failure subtest ([ticker_test.go:491](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/ticker/ticker_test.go#L491)) now pins the same choice for batches.

For the record, the orderbook precedent does differ on the failure I measured. `update` stores the ticker and then returns the mux publish error ([ticker.go:221-228](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/ticker/ticker.go#L221-L228)). `Depth.Publish` logs that same error ([depth.go:58-62](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchanges/orderbook/depth.go#L58-L62)), and the buffer relays regardless ([buffer.go:81-82](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/exchange/websocket/buffer/buffer.go#L81-L82)). This run pushes one kraken ticker frame and one websocket orderbook snapshot per iteration, so both hit the same queue at the same moment:

```
GOMAXPROCS=1, 1000 iterations, one subscription to an unrelated exchange's tickers
            ticker relayed   in the store, not relayed   orderbook relayed
c6e47c6       1000/1000                -                   1000/1000
dd86392        37/1000                963                  1000/1000
```

If that's intended, one clause at [CODING_GUIDELINES.md:32](https://github.com/thrasher-corp/gocryptotrader/blob/dd86392fbf220f471e11f072482bb60672ab1247/docs/CODING_GUIDELINES.md#L32) would answer the question for the next reader: a dispatch failure after the store accepts a ticker withholds the relay too. Optional.

shazbert (approved):
Nothing more from me, there are two unresolved nits by me, you can resolve them if you don't want to do them, but they are not blocking. ch ACK

PR #2398, opened 2026-09-21
snore: rpcserver if err := func() usage

gloriousCode:
# PR Description

In an effort to cut down on poor reviewer fatigue, I've taken this out of an existing PR


thrasher-:
@codex please review

engine/rpcserver.go
```diff
@@ -2798,8 +2782,7 @@ func (s *RPCServer) GCTScriptAutoLoadToggle(_ context.Context, r *gctrpc.GCTScri
 		return &gctrpc.GenericResponse{Status: "success", Data: "script " + r.Script + " removed from autoload list"}, nil
 	}
 
-	err := s.gctScriptManager.Autoload(r.Script, false)
-	if err != nil {
+	if err := s.gctScriptManager.Autoload(r.Script, false); err != nil {
```
thrasher-, line 2785: Heimdall: **Minor** — this converts the `r.Status == false` half and leaves the `true` half eight lines above in the old form, so one function now carries both spellings of the same call.

**Fix:** take [engine/rpcserver.go:2777](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L2777) as well.

```go
	if r.Status {
		if err := s.gctScriptManager.Autoload(r.Script, true); err != nil {
			//nolint:nilerr // error is returned in the generic response
			return &gctrpc.GenericResponse{Status: "error", Data: err.Error()}, nil
		}
```

**Why:** at [`c6e47c6`](https://github.com/thrasher-corp/gocryptotrader/commit/c6e47c6e1e6d6f7353148379d587188ec47c0367) both halves were spelled identically, so this PR is where they diverge. That error's only uses are inside its own check, so narrowing it is safe, and the standalone `//nolint:nilerr` keeps working — applied locally together with the other two suggestions:

```
go build ./engine/...           ok
go vet ./engine/...             ok
gofumpt -l engine/rpcserver.go  (no output)
golangci-lint run ./engine/...  0 issues
go test ./engine/... -race      ok  36.457s
```

One cost worth naming: #2183 makes the same choice at this line, so taking this diverges from the parent and buys a small conflict on rebase. If you would rather keep this PR a byte-exact subset of #2183, say so and I will drop it.

engine/rpcserver.go
```diff
@@ -2687,8 +2673,7 @@ func (s *RPCServer) GCTScriptUpload(_ context.Context, r *gctrpc.GCTScriptUpload
 				failedFiles = append(failedFiles, files[x])
 			}
 		}
-		err = os.Remove(fPath)
-		if err != nil {
+		if err := os.Remove(fPath); err != nil {
```
thrasher-, line 2676: Heimdall: **Minor** — scoping `os.Remove` here leaves the function-level `err` holding a swallowed validation failure for the rest of the block.

[engine/rpcserver.go:2671](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L2671) writes a failed `Validate` into the function-level `err` and deliberately does not return it; it collects the file into `failedFiles`. Before this PR the very next statement, `err = os.Remove(fPath)`, overwrote that value. Now it does not, so from line 2678 onwards the function-level `err` carries the last validation failure.

**Nothing reads it, so there is no behaviour change** — the only later uses of that object are the writes at [:2680](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L2680) and [:2687](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L2687). It is a trap for the next edit in this block, though, and converting :2671 too removes it:

```go
		for x := range files {
			if err := s.gctScriptManager.Validate(files[x]); err != nil {
				failedFiles = append(failedFiles, files[x])
			}
		}
```

**Why:** this is the one site in the 44 where the old code's assignment was load-bearing as a *clear*, not just as a check. Verified dead with a go/types pass that resolves the exact `types.Object` of the shadowed variable and orders its reads and writes; the same pass reports a read hazard at all three sites in `SetExchangeAsset` when I inject a synthetic stale-`err` read, so the negative result is meaningful.

engine/rpcserver.go
```diff
@@ -2876,31 +2856,26 @@ func (s *RPCServer) SetAllExchangePairs(_ context.Context, r *gctrpc.SetExchange
 			if err != nil {
 				return nil, err
 			}
-			err = exchCfg.CurrencyPairs.StorePairs(assets[i], pairs, true)
-			if err != nil {
+			if err := exchCfg.CurrencyPairs.StorePairs(assets[i], pairs, true); err != nil {
```
thrasher-, line 2859: Heimdall: **Nit** — two `var` hoists in this PR's functions now exist only to feed the outer `err`, and can go with the rest.

[engine/rpcserver.go:2854](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L2854) here, and [engine/rpcserver.go:3421](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L3421) in `GetHistoricTrades`:

```go
-			var pairs currency.Pairs
-			pairs, err = base.CurrencyPairs.GetPairs(assets[i], false)
+			pairs, err := base.CurrencyPairs.GetPairs(assets[i], false)

-	var trades []trade.Data
 	...
-		trades, err = exch.GetHistoricTrades(stream.Context(), cp, a, iterateStartTime, iterateEndTime)
+		trades, err := exch.GetHistoricTrades(stream.Context(), cp, a, iterateStartTime, iterateEndTime)
```

**Why:** the split declare-then-assign form is only there so the assignment could reuse the function-level `err`, which is the same thing the rest of this PR is undoing. Neither `pairs` nor `trades` is read outside its loop body, so both collapse cleanly — four lines to two, and it drops the last two `err =` assignments in the two functions. Build, vet, gofumpt, `golangci-lint` (0 issues) and `go test ./engine/... -race` all clean with this applied.

thrasher- (approved):
Heimdall: Approving. Faithful extraction — all 44 conversions appear verbatim in #2183; the 31 held back are the `checkParams`/`IsAssetEnabled` lines it rewrites to functions not yet on master.

No behaviour change: a `go/types` liveness pass over all 44 sites found no stale-`err` read, expanding the scoped `if`s back gives an AST identical to the base, and of 477 `engine` symbols exactly one changed size — [`GetDataHistoryJobsBetween`](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L3887), 2655 → 2643 bytes, the compiler dropping the dead `error` slot.

Three comments, at `:2785`, `:2676` and `:2859`; evidence is on the lines and none gates.

Optional, out of scope: [engine/rpcserver.go:3899](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L3899) validates `start.Local()` but passes plain `start` on at [:3903](https://github.com/thrasher-corp/gocryptotrader/blob/21383f2904d5ad1dbe65400ff343f4fa22560454/engine/rpcserver.go#L3903) — `.Local()` is a no-op for `StartEndTimeCheck`, which only compares instants, and it is the only one of 14 call sites doing it. Happy to raise separately.

Both red jobs are ambient — [macOS](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35565357120/job/106225974811) a Kraken Cloudflare 502, [ubuntu sonic=false](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35565357120/job/106225974778) 33 × `context deadline exceeded` on `api.hbdm.com`; both sibling jobs running the same wrapper suite on this commit passed. `codecov/patch` is re-attribution: zero coverage-status changes against the base lines replaced.

Take the three or leave them — nothing here blocks.

shazbert (approved):
LGTM!

PR #698, opened 2021-06-22
gctcli: bump urfave cli depends version

shazbert:
# PR Description

Had a stale branch I found locally, opened if we want to merge. Seems like we were already ref: v2. 🤷 

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions/AppVeyor with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2399, opened 2026-09-21
gctcli: reject mixed positional and flag arguments

gloriousCode:
# PR Description
This is to cut down on poor reviewer fatigue. This came up in a Heimdall review comment about how my code was terrible because it tested against mixed argument types. This PR resolves it by blocking when you try to use two types of gctcli command params

Add recursive command validation that rejects mixed positional and flag forms before command execution. Global options, negative numeric values and arguments after `--` remain supported, and existing command hooks continue to run.

### Blocked:
- `cancelallorders --asset perpetualcontract Bitmex ETH-USD`

### Okay:
- `cancelallorders --exchange Bitmex --asset perpetualcontract --pair ETH-USD`
- `cancelallorders Bitmex perpetualcontract ETH-USD`

Validation:
- `go test ./cmd/gctcli -count=1`
- `make lint`
- `make misc_checks`


thrasher-:
@codex please review

cmd/gctcli/arguments.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	lineage := c.Lineage()
+	// Parent command flags also supply command parameters, but app flags only
+	// configure the connection and may accompany either argument style.
+	for i, context := range lineage {
+		if i+1 == len(lineage) || lineage[i+1].Command == nil {
+			break
+		}
+		if context.NumFlags() != 0 {
```
thrasher-: Heimdall: **Should fix** — options that have no positional slot can no longer be given alongside positional arguments, which leaves two commands reachable only in the all-flags form.

**Fix:** decide whether a flag with no positional slot should be exempt here, or give those options positional slots. Either way [`cmd/gctcli/README.md:35`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/README.md#L35) should stop offering "entirely as positional arguments" to commands that cannot provide it.

**Why:** the regression is the removal of working *mixed* invocations — the missing positional implementations predate this PR. Against `upstream/master`:

```text
                                                             master                    this PR
modifyorder --price 100 Binance spot BTC-USD oid123          reaches the RPC           cannot mix positional arguments…
modifyorder --amount 1 Binance spot BTC-USD oid123           reaches the RPC           cannot mix positional arguments…
withdrawalrequesthistory byexchange --asset spot Binance 10  reaches the RPC           cannot mix positional arguments…
gethistoriccandles --fillmissingdatawithtrades Binance BTC-USDT spot 86400 86400
                                                             reaches the RPC           cannot mix positional arguments…
gethistoriccandlesextended --sync | --db | --fill | --sync --force
    Binance BTC-USD spot 60 "2026-09-01 00:00:00" "2026-09-02 00:00:00"
                                                             reaches the command body  cannot mix positional arguments…
trade convertsavedtradestocandles --sync … (same operands)   reaches the command body  cannot mix positional arguments…
```

`modifyorder` reads `--price` and `--amount` at [`commands.go:2047`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L2047) with no `else { … c.Args() … }` branch, so with otherwise valid operands its all-positional form always returns `either --price or --amount should be present`. `withdrawalrequesthistory byexchange` does the same with `--asset` at [`commands.go:2904`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L2904) and then rejects the empty value. The other flag-only reads are [`commands.go:3925`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L3925), [`4084`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L4084), [`4089`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L4089), [`4094`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L4094), [`4101`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L4101) and [`trades.go:673`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/trades.go#L673), [`678`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/trades.go#L678). (`--force` is not in the table on its own: with valid operands and no `--sync` it already failed on master with `cannot forcefully overwrite without sync`.)

Trailing flags are not the regression — Go's `flag` package stops at the first operand, so `modifyorder Binance spot BTC-USD oid --price 100` already failed on master, with a different error. It is the leading form that worked. Nothing goes silently wrong and these all reach the RPC on this head:

```sh
gctcli modifyorder --exchange Binance --asset spot --pair BTC-USD --order_id oid123 --price 100
gctcli withdrawalrequesthistory byexchange --exchange Binance --limit 10 --asset spot
gctcli gethistoriccandles --exchange Binance --pair BTC-USDT --asset spot --rangesize 86400 --granularity 86400 --fillmissingdatawithtrades
```

so this is a compatibility decision rather than a correctness one — but worth making before merge.

cmd/gctcli/README.md
```diff
@@ -32,6 +32,33 @@ go build or go run .
...
+
+# Allowed: global option with named command flags
+gctcli --rpchost localhost:9052 getticker --exchange Binance --pair BTC-USDT --asset spot
+
+# Rejected: a named command flag mixed with positional command arguments
+gctcli --rpchost localhost:9052 getticker --exchange Binance BTC-USDT spot
+```
+
+Negative numeric positional values remain supported. Use `--` before positional
+values that begin with a dash and should be treated literally.
```
thrasher-: Heimdall: **Should fix** — this does not say where `--` has to go, and the placement a reader would pick is rejected.

**Fix:**

```suggestion
Negative numeric positional values remain supported. To pass a positional value
that begins with a dash and is not a number, put `--` immediately after the
command name; every value after it is taken literally. A `--` placed after a
positional value is not accepted as a separator.
```

**Why:** a positional value beginning with a dash that is not a number is now refused, on lines that contain no flag at all, and `--` is the only escape:

```text
$ gctcli removeportfolioaddress 1ABCdef BTC -label
master : reaches the RPC
head   : cannot mix positional arguments and command flags; use either form exclusively

$ gctcli withdrawcryptofunds Binance USDT 10 TJU9addr tag 1 -urgent- trx     same
$ gctcli getorder Binance spot BTC-USD -oid123                               same

$ gctcli getticker Binance BTC-USD -- -weird
cannot mix positional arguments and command flags; use either form exclusively
$ gctcli getticker -- Binance BTC-USD -weird
invalid asset supplied                        # -weird arrived as the asset
$ gctcli removeportfolioaddress -- 1ABCdef BTC -label
reaches the RPC
```

Numbers are unaffected — `gctcli getticker Binance BTC-USD -1` and a bare `-` still get through. 6b3315b28 turned a `--` found among the positional arguments into an error, so the separator only works as the first token; after a leading one a later `--` is passed through as a literal value. At the leaf too: `gctcli orderbook buy nominal -- Binance BTC-USD spot 1` reaches the RPC. Applied, markdownlint and `misc_checks` stay clean.

cmd/gctcli/arguments.go
```diff
@@ -0,0 +1,68 @@
...
+		}
+	}
+	for _, arg := range args {
+		if arg == "--" {
+			return errMixedArguments
+		}
+		if len(arg) > 1 && strings.HasPrefix(arg, "-") {
+			// Negative numeric parameters are positional values, not flags.
+			if _, err := strconv.ParseFloat(arg, 64); err != nil {
+				return errMixedArguments
```
thrasher-: Heimdall: **Should fix** — a rejected line does not say which argument caused it.

**Fix:** name the offending token. This needs `"fmt"` adding to the import block, so it is not a clean one-click apply, and it only covers this rejection path — the leading-flag rejection at line 43 and the `--` rejection at line 58 would still name nothing:

```suggestion
				return fmt.Errorf("%w: %q", errMixedArguments, arg)
```

**Why:** `gctcli removeportfolioaddress 1ABCdef BTC -label` prints `cannot mix positional arguments and command flags; use either form exclusively` on a line with no flag on it; wrapped it prints `…: "-label"`, which points at the value needing the `--` escape. Applied with the import, `go test ./cmd/gctcli -race -count=1` and `golangci-lint run ./cmd/gctcli/...` are clean — `errors.Is` still matches through the wrap, so the existing `require.ErrorIs` assertions are unaffected. It is the form [the guidelines ask for](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/docs/CODING_GUIDELINES.md#L131): "When returning additional context for a sentinel error, wrap the sentinel with `fmt.Errorf` and `%w` rather than replacing it with a new inline `errors.New(...)` value."

cmd/gctcli/arguments_test.go
```diff
@@ -0,0 +1,139 @@
...
+		{name: "equals flags", args: "getticker --exchange=Bitmex --asset=perpetualcontract --pair=ETH-USD"},
+		{name: "reported missing pair", args: "getticker --asset perpetualcontract Bitmex ETH-USD", mixed: true},
+		{name: "flagged exchange", args: "getticker --exchange Bitmex ETH-USD perpetualcontract", mixed: true},
+		{name: "flagged pair", args: "getticker --pair ETH-USD Bitmex perpetualcontract", mixed: true},
+		{name: "two flags", args: "getticker --asset perpetualcontract --pair ETH-USD Bitmex", mixed: true},
+		{name: "trailing flag", args: "getticker Bitmex ETH-USD --asset perpetualcontract", mixed: true},
+		{name: "trailing equals flag", args: "getticker Bitmex ETH-USD --asset=perpetualcontract", mixed: true},
+		{name: "trailing ignored flag", args: "getticker Bitmex ETH-USD perpetualcontract --asset spot", mixed: true},
+		{name: "unknown trailing flag", args: "getticker Bitmex --typo spot", mixed: true},
+		{name: "single dash flag", args: "getticker Bitmex -asset spot", mixed: true},
```
thrasher-: Heimdall: **Should fix** — the table no longer covers a parsed leading boolean or `--flag=value` mixed with positional operands.

**Fix:** restore the cases dropped in 6b3315b28, plus the leading equals form. The fixture also needs `Aliases: []string{"a"}` on the asset flag and a `&cli.BoolFlag{Name: "force"}` at [`arguments_test.go:59-63`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/arguments_test.go#L59-L63), so this is not a clean one-click apply:

```suggestion
		{name: "single dash flag", args: "getticker Bitmex -asset spot", mixed: true},
		{name: "leading equals flag", args: "getticker --asset=spot Bitmex ETH-USD", mixed: true},
		{name: "alias before positional", args: "getticker -a spot Bitmex", mixed: true},
		{name: "alias after positional", args: "getticker Bitmex -a spot", mixed: true},
		{name: "boolean flag", args: "getticker --force Bitmex", mixed: true},
		{name: "explicit false", args: "getticker --force=false Bitmex", mixed: true},
```

**Why:** a mutant that only rejects flags which consumed a *separate* value token passes the table as it stands. Replacing the `NumFlags()` check, for `i == 0`, with `len(lineage[1].Args().Tail()) - len(args) >= 2`:

```text
current table : ok   github.com/thrasher-corp/gocryptotrader/cmd/gctcli
restored      : --- FAIL: TestValidateCommandArguments/boolean_flag
                --- FAIL: TestValidateCommandArguments/explicit_false
                --- FAIL: TestValidateCommandArguments/leading_equals_flag
```

It lets `gctcli getticker --asset=spot Binance BTC-USD` and `gctcli gethistoriccandlesextended --sync=false Binance BTC-USD spot 60 …` through — both reach the RPC on master. That is the boolean/equals gap; the alias cases I would restore on the general principle rather than a mutant, since 124 flag declarations in this tree carry aliases and 49 are bools, and [the guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/docs/CODING_GUIDELINES.md#L215) ask to "Cross relevant input dimensions, such as aliases, wildcards, accepted capitalisation, … explicit, omitted or `null` values". With the fixture change all six pass under `-race`.

cmd/gctcli/main.go
```diff
@@ -245,6 +245,7 @@ func main() {
 		orderbookCommand,
 		getCurrencyTradeURLCommand,
 	}
+	registerArgumentValidation(app.Commands)
```
thrasher-: Heimdall: **Should fix** — nothing covers this line, so the feature can be unwired without a test noticing.

**Fix:** lift the app construction out of `main` into something a test can call, drive one mixed invocation through it and assert `errMixedArguments` specifically. Deleting this line should make that test fail.

**Why:** deleting it today leaves the suite green:

```text
$ sed -i '/registerArgumentValidation(app.Commands)/d' cmd/gctcli/main.go
$ go test ./cmd/gctcli -count=1
ok  	github.com/thrasher-corp/gocryptotrader/cmd/gctcli	0.011s
```

It is also the only uncovered added line in codecov's patch report.

thrasher- (changes requested):
Heimdall: Good, well-scoped fix — it closes the mixed-argument problem from #2183, and rejects two inputs that previously sent unintended RPC values: `getexchanges --enabled false` asked for `Enabled: true`, and `orderbook buy nominal --asset spot Binance BTC-USD 42` took the percentage as 0 ([`orderbook.go:110`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/orderbook.go#L110)). First tests in a package that had none.

Five things in the comments: a compatibility decision on options with no positional slot, the README's `--` placement, the error naming no token, four test cases dropped in 6b3315b28, and the `registerArgumentValidation` call in `main` having no test behind it.

Two that cannot anchor. The description's three `cancelallorders` examples behave identically on master and on this head — both flag forms die with `Incorrect Usage: flag provided but not defined: -asset`, since master's command declares only `--exchange` ([`commands.go:1929-1934`](https://github.com/thrasher-corp/gocryptotrader/blob/6b3315b2871ef68805a2666cda428dc637205b52/cmd/gctcli/commands.go#L1929-L1934)); they match #2183's richer definition. And #2183 still carries `arguments.go` with the pre-6b3315b28 `break` and the older fixture — `git merge-tree` conflicts in all three files in both directions.

CI is not from this diff: `cmd/gctcli` passes in both failing jobs, which fail on GateIO ticker tests (`SUI_USDT bid equals ask`), GateIO in `cmd/exchange_wrapper_standards`, and Huobi (`request limit`). `codecov/patch` passes; `codecov/project` fails against a base two commits behind, cause unestablished. Lint, `misc_checks`, markdownlint and `-race` are clean here.

**Before merge:** decide what happens to options with no positional slot, and make the README agree ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4062140744)).

Happy to approve once that's settled.

cmd/gctcli/arguments.go
```diff
@@ -0,0 +1,68 @@
...
+	for i, context := range lineage {
+		if i+1 == len(lineage) || lineage[i+1].Command == nil {
+			break
+		}
+		if context.NumFlags() != 0 {
+			return errMixedArguments
+		}
+	}
+
+	// The parser stops at the first positional argument. Inspect the remainder
```
shazbert: Why don't we just purge positional arguments and force flags that way we cut down on a lot of code and not have to worry about all these shenanigans

shazbert (changes requested):
One suggestion which will annoy you.

cmd/gctcli/arguments.go
gloriousCode: REEEEEEEEEE
I thought that may have been too heavy handed to just remove it all out of my sheer rage at the initial nit. It would cut down on a lot of code which is nice. I SHALL PONDER THIS

gloriousCode: I have added a new commit which just wrecks positional cli commands instead. I'll see if I can get heimdalls approval

Just know it was initially a 231 line change, now its 1.5k adds and 2.3k removals. all ur fault

shazbert: 😢 sorry bro

cmd/gctcli/websocket_management.go
```diff
@@ -64,12 +67,14 @@ var websocketManagerCommand = &cli.Command{
...
+					Name:     exchangeFlag,
+					Required: true,
+					Usage:    exchangeUsage,
 				},
 				&cli.StringFlag{
-					Name:  "proxy",
-					Usage: "proxy address to change to, if proxy string is not set, this will stop the utilisation of the prior set proxy.",
+					Name:     "proxy",
+					Required: true,
+					Usage:    "proxy address to change to, if proxy string is not set, this will stop the utilisation of the prior set proxy.",
```
thrasher-, line 77: Heimdall: **Should fix** — `--proxy` and `--url` are now required, but leaving them unset is how these commands clear the proxy and restore the default URL.

**Fix:** drop `Required` here, and on `--url` at [websocket_management.go:93](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/websocket_management.go#L93):

```suggestion
					Name:  "proxy",
					Usage: "proxy address to change to, if proxy string is not set, this will stop the utilisation of the prior set proxy.",
```

**Why:** both flags' usage text documents the empty value, and the server acts on it: `SetProxyAddress` removes the proxy for `""` ([manager.go:977](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchange/websocket/manager.go#L977)) and `SetWebsocketURL` falls back to the packaged default ([manager.go:876](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchange/websocket/manager.go#L876)). Master sends those requests; this head refuses them:

```text
                                         master                          this PR
websocket setproxy --exchange Binance    WebsocketSetProxy{proxy: ""}    Required flag "proxy" not set
websocket seturl --exchange Binance      WebsocketSetURL{url: ""}        Required flag "url" not set
```

`--proxy ''` still gets through, but nothing points to it; the help printed above the error still says an unset proxy clears it. With both `Required` lines dropped, setproxy and seturl send master's requests again.

cmd/gctcli/futures.go
```diff
@@ -218,79 +222,86 @@ var futuresCommands = &cli.Command{
...
+					Usage:    pairUsage,
 				},
 				&cli.StringFlag{
-					Name:    "margintype",
-					Aliases: []string{"margin", "mt", "m"},
-					Usage:   "the margin type, such as 'isolated', 'multi' or 'cross'",
+					Name:     "margintype",
+					Required: true,
+					Aliases:  []string{"margin", "mt", "m"},
+					Usage:    "the margin type, such as 'isolated', 'multi' or 'cross'",
```
thrasher-, line 298: Heimdall: **Should fix** — `--margintype` is now required on `setleverage`, `getleverage` and `changepositionmargin`, but an unset margin type is a valid request for all three.

**Fix:** drop `Required` here and at [futures.go:339](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/futures.go#L339) and [futures.go:376](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/futures.go#L376). Keep it on `setmargintype`, where an unset type can never succeed.

```suggestion
					Name:    "margintype",
					Aliases: []string{"margin", "mt", "m"},
					Usage:   "the margin type, such as 'isolated', 'multi' or 'cross'",
```

**Why:** master accepted an omitted margin type, since `margin.IsValidString("")` is true. The server turns `""` into `margin.Unset` without error ([margin.go:76-77](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/margin/margin.go#L76-L77)), and the wrappers take it: Binance ignores the margin type for leverage ([binance_wrapper.go:2750](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/binance/binance_wrapper.go#L2750), [binance_wrapper.go:2764](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/binance/binance_wrapper.go#L2764)) and refuses only `Multi` when changing position margin ([binance_wrapper.go:2257](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/binance/binance_wrapper.go#L2257)); Bybit ignores it ([bybit_wrapper.go:1731](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/bybit/bybit_wrapper.go#L1731)); Gate.io treats `Unset` as isolated ([gateio_wrapper.go:2466](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/gateio/gateio_wrapper.go#L2466)).

```text
futures … --exchange Binance --asset usdtmarginedfutures --pair BTC-USDT      master                          this PR
setleverage --leverage 10                                                     SetLeverage{margin_type: ""}    Required flag "margintype" not set
getleverage                                                                   GetLeverage{margin_type: ""}    Required flag "margintype" not set
changepositionmargin --originalallocatedmargin 10 --newallocatedmargin 20     ChangePositionMargin{…}         Required flag "margintype" not set
```

With the three lines dropped, all three send master's requests again.

`--originalallocatedmargin` on the same command went the other way. Master's positional fallback made omitting it an error (`strconv.ParseFloat: parsing "": invalid syntax`); now it goes out as `0`, which Binance and OKX refuse ([binance_wrapper.go:2254](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/binance/binance_wrapper.go#L2254), [okx_wrapper.go:2450](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/okx/okx_wrapper.go#L2450)) and KuCoin doesn't need. Leaving it optional is fine; mark it `Required` only if you want master's local error back.

cmd/gctcli/main.go
```diff
@@ -245,6 +262,7 @@ func main() {
 		orderbookCommand,
 		getCurrencyTradeURLCommand,
 	}
+	rejectPositionalArguments(app.Commands)
```
thrasher-, line 272: Heimdall: **Should fix** — nothing tests that `main` wires this up: delete this line, or its twin at [btcli/main.go:144](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/backtester/btcli/main.go#L144), and both suites stay green.

**Fix:** run the real `main` in a child process, so nothing leaks between runs and `main` needs no refactor:

```go
func TestMainRejectsPositionalArguments(t *testing.T) {
	if os.Getenv("GCTCLI_TEST_MAIN") == "1" {
		os.Args = []string{"gctcli", "--cert", os.DevNull, "getexchanges", "unexpected"}
		main()
		return
	}
	cmd := exec.CommandContext(t.Context(), os.Args[0], "-test.run=^TestMainRejectsPositionalArguments$") //nolint:gosec // re-runs this test binary to exercise main
	cmd.Env = append(os.Environ(), "GCTCLI_TEST_MAIN=1")
	out, err := cmd.CombinedOutput()
	require.Error(t, err, "main must exit with an error")
	assert.Contains(t, string(out), errPositionalArgument.Error(), "main should reject the positional argument")
}
```

btcli wants the same with `listalltasks unexpected`. `--cert os.DevNull` keeps an unwired build off the network: its TLS setup fails before anything dials.

**Why:** the new tests only call `rejectPositionalArguments` on hand-built commands. With this line removed, both packages still pass, and the build goes back to forwarding stray arguments:

```text
$ go test ./cmd/gctcli/ ./backtester/btcli/
ok  	github.com/thrasher-corp/gocryptotrader/cmd/gctcli
ok  	github.com/thrasher-corp/gocryptotrader/backtester/btcli

$ gctcli getticker --exchange Binance --pair BTC-USDT --asset spot extra
this head:     unexpected positional argument; use named flags: "extra"
line removed:  GetTicker sent, "extra" ignored
```

With both tests added, the suites pass under `-race -count 2 -shuffle on` and lint is clean; with the line removed, each fails with `"credentials: failed to append certificates" does not contain "unexpected positional argument; use named flags"`. An in-process `newApp()` looks simpler, but once a test has run one of the real command trees, as `TestGetDataHistoryJob` does, `newApp()` wraps urfave's shared `help` command too, and `gctcli help getticker` then fails in that process with `unexpected positional argument`. This is [my earlier point on `main`](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4062140807), carried to the new line.

cmd/gctcli/README.md
```diff
@@ -32,6 +32,23 @@ go build or go run .
 For a full list of commands, you can run `gctcli --help`. Alternatively, you can also
 visit our [GoCryptoTrader API reference](https://api.gocryptotrader.app/).
 
+Supply command parameters as named flags. Positional arguments are rejected for
+all commands and subcommands.
```
thrasher-, line 36: Heimdall: **Minor** — with the positional placeholders gone, `--help` no longer says which flags are required, and the other ways people ask for help or pass booleans now fail with misleading errors.

**Fix:** mark required flags through urfave's formatter hook, once in each CLI's `main`:

```go
flagString := cli.FlagStringer
cli.FlagStringer = func(f cli.Flag) string {
	if r, ok := f.(cli.RequiredFlag); ok && r.IsRequired() {
		return flagString(f) + " (required)"
	}
	return flagString(f)
}
```

Then say here that `gctcli <command> --help` lists a command's flags, and that booleans take `=`, as in `--enabled=false`.

**Why:**

```text
master    USAGE: gctcli getticker [command options] <exchange> <pair> <asset>
this PR   USAGE: gctcli getticker [command options]
             --exchange value  the exchange to get the ticker for
hooked       --exchange value  the exchange to get the ticker for (required)
```

- `gctcli getticker help`, via the `help` subcommand urfave adds to every command, now prints the `help` command's own page and exits 1 with `Required flags "exchange, pair, asset" not set`, because urfave checks required flags before it dispatches. That hits every command that gained a required flag, and btcli's four. `--help` and `gctcli help getticker` still work.
- Go's flag parser never lets a boolean take the next token, so `getexchanges --enabled false` leaves `false` behind. Master silently sent `enabled: true`; this PR refuses with `unexpected positional argument; use named flags: "false"`. That is safer, but the user did use the named flag, and `--sync`, `--force`, `--db`, `--status`, `--overwrite` and `--archived` still advertise `<true/false>` ([commands.go:3690](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L3690), [trades.go:32](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/trades.go#L32)).
- The same hint answers a mistyped subcommand: `gctcli futures bogus` gives `use named flags: "bogus"` where master said `No help topic for 'bogus'`.

cmd/gctcli/main_test.go
```diff
@@ -0,0 +1,96 @@
...
+		{name: "ticker pair", command: getTickerCommand, flagName: pairFlag, required: true},
+		{name: "ticker asset", command: getTickerCommand, flagName: assetFlag, required: true},
+		{name: "submit side", command: submitOrderCommand, flagName: sideFlag, required: true},
+		{name: "submit amount", command: submitOrderCommand, flagName: amountFlag, required: true},
+		{name: "submit margin optional", command: submitOrderCommand, flagName: "margintype"},
+		{name: "cancel pair optional", command: cancelOrderCommand, flagName: pairFlag},
+		{name: "open interest pair optional", command: futuresCommands.Subcommands[len(futuresCommands.Subcommands)-1], flagName: pairFlag},
+		{name: "margin rates currency", command: getMarginRatesHistoryCommand, flagName: "currency", required: true},
+		{name: "job id alternative", command: dataHistoryCommands.Subcommands[2], flagName: "id"},
+		{name: "job nickname alternative", command: dataHistoryCommands.Subcommands[2], flagName: "nickname"},
```
thrasher-: Heimdall: **Nit** — these indexes resolve to the right commands today, but they depend on declaration order and on no other test having run the tree first: urfave appends its `help` command to `Subcommands` on the first run.

**Fix:** look the commands up by name:

```suggestion
		{name: "open interest pair optional", command: futuresCommands.Command("getopeninterest"), flagName: pairFlag},
		{name: "margin rates currency", command: getMarginRatesHistoryCommand, flagName: "currency", required: true},
		{name: "job id alternative", command: dataHistoryCommands.Command("getajob"), flagName: "id"},
		{name: "job nickname alternative", command: dataHistoryCommands.Command("getajob"), flagName: "nickname"},
```

**Why:** adding a test that runs `futures --help` through `app.Run` makes `TestRequiredCommandFlags/open_interest_pair_optional` fail with `flag "pair" missing from help`, because `help` is then the last subcommand. With the lookups above, the package passes alongside that test under `-race -count 2 -shuffle on`.

While here: the closing checks at [main_test.go:45](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/main_test.go#L45), [48](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/main_test.go#L48), [56](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/main_test.go#L56) and [90](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/main_test.go#L90), and their btcli twins, gate nothing that follows, so [the guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/docs/CODING_GUIDELINES.md#assertion-usage) want `assert` with a "should" message. The child `Before`'s `errBefore` branch at [main_test.go:34-36](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/main_test.go#L34-L36) never fires, since no case passes `--value before`.

cmd/gctcli/data_history_test.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func TestGetDataHistoryJob(t *testing.T) {
+	tests := []struct {
+		name    string
+		args    []string
+		wantErr error
+	}{
+		{name: "by ID", args: []string{"--id", "job-id"}, wantErr: os.ErrNotExist},
+		{name: "by nickname", args: []string{"--nickname", "job-name"}, wantErr: os.ErrNotExist},
```
thrasher-, line 20: Heimdall: **Nit** — these two cases can't tell the selectors apart. They pass as soon as validation lets the call through to `setupClient`, which then fails on the unset `certPath` with `os.ErrNotExist`.

**Fix:** have the action build its request in a helper, say `dataHistoryJobRequest(c *cli.Context) (*gctrpc.GetDataHistoryJobDetailsRequest, error)`, and assert its `Id` and `Nickname` here.

**Why:** swapping the two fields in the request at [data_history.go:327-328](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/data_history.go#L327-L328) leaves the package green:

```text
-		Id:       id,
-		Nickname: nickname,
+		Id:       nickname,
+		Nickname: id,

$ go test ./cmd/gctcli/
ok  	github.com/thrasher-corp/gocryptotrader/cmd/gctcli
```

thrasher- (changes requested):
Heimdall: Going flags-only settles last round's question about options with no positional slot, and the rejection now names the offending token. The conversion mostly holds: I replayed 1,343 gctcli invocations against master and this head through a fake gRPC server that records each request. Apart from what's below, every flag-form call master sent goes out unchanged, or is refused only where the server would reject master's request anyway, such as an empty exchange. It also makes flag forms work that master declared but never read (`setloggerdetails --flags`, `script autoload --command/--script`), gives `datahistory getajob`, `deletejob`, `pausejob` and `unpausejob` the `--nickname` they advertised, and stops `getexchanges --enabled false` silently sending `true`.

**To unblock:**
1. [websocket_management.go:76](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640480): `--proxy` and `--url` are now required, but unset is how you clear the proxy and restore the default URL.
2. [futures.go:296](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640505): `--margintype` is now required on three commands where unset is valid.
3. [main.go:265](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640526): still nothing tests that `main` wires the rejection up.

Inline, not blocking: a Minor on [README.md:35](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640536) (help no longer marks required flags; `<command> help` and `--flag false` fail with misleading errors) and Nits on [main_test.go:75](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640547) and [data_history_test.go:17](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640563).

GitHub sends no patch for `commands.go`, so these can't anchor:
- **Minor**: `submitorder` now sends `--margintype` ([commands.go:1452](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L1452)). Master validated it and then dropped it, so `--asset margin --margintype isolated` sent `margin_type: ""`. The server applies it to the order ([rpcserver.go:1205-1206](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/engine/rpcserver.go#L1205-L1206)), and OKX turns it into the trade mode ([okx_wrapper.go:922](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/okx/okx_wrapper.go#L922)). It's a fix, but it changes what an order does, so it deserves a line in the description or its own PR.
- **Minor**: `withdrawalrequesthistory byexchange` and `bydate` lost their only `limit > math.MaxInt32` check, which sat on the deleted positional path ([master commands.go:2894](https://github.com/thrasher-corp/gocryptotrader/blob/c6e47c6e1e6d6f7353148379d587188ec47c0367/cmd/gctcli/commands.go#L2894), [2971](https://github.com/thrasher-corp/gocryptotrader/blob/c6e47c6e1e6d6f7353148379d587188ec47c0367/cmd/gctcli/commands.go#L2971)). `--limit 3000000000` now always goes out as `-1294967296` (and `-2147483649` as `2147483647`, on both builds). Bound both ends on the flag reads instead (`limit < math.MinInt32 || limit > math.MaxInt32`), at [commands.go:2702-2704](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L2702-L2704) and [2756-2758](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L2756-L2758).
- **Minor**: `withdrawfiatfunds --bankaccountid` is now required ([commands.go:2494-2498](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L2494-L2498)). On master, omitting it sent `""`, which the server matches to a configured bank account without an `id` (for an exchange's own accounts, one that lists the currency: [config.go:51-52](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/config/config.go#L51-L52)); IDs are optional. Requiring an explicit account for a withdrawal is defensible, but then the usage should say that `--bankaccountid ''` picks an ID-less one.
- **Nit**: dead code from the conversion. The `if c.NumFlags() == 0` help branches can't run where a flag is required ([commands.go:476](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L476) and most actions), and `if c.IsSet(x)` guards on required flags always pass. `var err error` at [commands.go:2692](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L2692) and [2751](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L2751) no longer does anything, nor does `marginType != "" &&` at [commands.go:1423](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L1423), since `margin.IsValidString("")` is true.
- **Nit**: some descriptions still read as argument lists; `withdrawalrequesthistory bydate` is "exchange start end limit" ([commands.go:2617](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L2617)), and its siblings are the same.

Same pattern, lower stakes: `orderbook … nominal` without `--percent` and `whalebomb --side SELL` without `--price` are now refused, though the server accepts 0 for both ([levels.go:344](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/orderbook/levels.go#L344), [calculator.go:30](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/exchanges/orderbook/calculator.go#L30)). Both are valid, if degenerate, queries. Likewise `script execute --path <dir>/timer.gct` without `--filename`: master sent it and the server joins the two ([rpcserver.go:2583](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/engine/rpcserver.go#L2583)), but `--path` is documented as a directory. Drop `Required` in these, or keep it deliberately.

The description's "Okay" example, `cancelallorders --exchange Bitmex --asset perpetualcontract --pair ETH-USD`, fails on this head with `flag provided but not defined: -asset`, because `cancelallorders` only has `--exchange`.

- `gethistoriccandles` declares `"fillmissingdatawithtrades, fill"` as one name, so `--fill` is undefined ([commands.go:3560](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L3560)).
- `getmarginrateshistory` gives `includeallrates` the aliases `v` and `verbose`, so `--verbose` after the command name sets `include_all_rates` ([commands.go:4022](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/commands.go#L4022)).
- The data history manager docs use `enablesubsystem --subsystemname=…`, but the flag is `--subsystem` ([datahistory_manager.tmpl:60](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/documentation/engine_templates/datahistory_manager.tmpl#L60)).

CI isn't from this diff. The Windows job fails `TestGetDefaultConfigurations/btse` on a BTSE `429 Too Many Requests`, and the sonic job fails Coinut's `TestSeedInstruments` on a request timeout; `cmd/gctcli` and `backtester/btcli` pass in both.

cmd/gctcli/futures.go
```diff
@@ -18,32 +17,33 @@ import (
 // futuresCommands contains all commands related to futures
 // position data, funding rates, collateral, pnl etc
 var futuresCommands = &cli.Command{
-	Name:      "futures",
-	Aliases:   []string{"f"},
-	Usage:     "contains all futures based rpc commands",
-	ArgsUsage: commandArgsUsage,
+	Name:    "futures",
```
thrasher-, line 20: Heimdall: **Nit** — dropping `ArgsUsage` from the parent commands also dropped the only hint in their help that a subcommand comes next.

**Fix:** keep a subcommand placeholder. Restore `commandArgsUsage` in [main.go:69-77](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/main.go#L69-L77) as `"<command>"` and set `ArgsUsage: commandArgsUsage` on the thirteen parents: `futures`, `pair`, `script`, `websocket`, `trade`, `datahistory`, `datahistory addjob`, `currencystate`, `technicalanalysis`, `withdrawalrequesthistory`, `orderbook`, `orderbook sell` and `orderbook buy`.

**Why:** `gctcli futures --help`, and every other parent the same way:

```text
master   USAGE: gctcli futures [command options] <command> <args>
this PR  USAGE: gctcli futures [command options]
fixed    USAGE: gctcli futures [command options] <command>
```

`ArgsUsage` only feeds the help template, so with the change applied all thirteen parents show `<command>` and `futures getleverage … extra` is still refused with `unexpected positional argument`. Lint and `-race` tests stay clean.

cmd/gctcli/data_history.go
```diff
@@ -201,6 +198,7 @@ var (
 			Name:  "id",
 			Usage: guidExample,
 		},
+		nicknameFlag,
```
thrasher-, line 201: Heimdall: **Nit** — this gives `deletejob`, `pausejob` and `unpausejob` a `--nickname` for the first time, but `setDataHistoryJobStatus` still checks the pair the old way, so the same two flags now behave differently from `getajob`.

**Fix:** move the new checks into a helper that both actions call, and drop the inline `errors.New("can only set 'id' OR 'nickname'")` from [data_history.go:593](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/data_history.go#L593):

```go
// dataHistoryJobIdentifier returns the job ID or nickname a command selects, rejecting both or neither
func dataHistoryJobIdentifier(c *cli.Context) (id, nickname string, err error) {
	if c.IsSet("id") && c.IsSet("nickname") {
		return "", "", errDataHistoryJobIdentifiersConflict
	}
	id, nickname = c.String("id"), c.String("nickname")
	if id == "" && nickname == "" {
		return "", "", errDataHistoryJobIdentifierRequired
	}
	return id, nickname, nil
}
```

**Why:**

```text
                                getajob                              deletejob / pausejob / unpausejob
--id ''                         a job ID or nickname must be set     sends SetDataHistoryJobStatus{id: "", nickname: ""}
--id <uuid> --nickname ''       set a job ID or nickname, not both   sends the request by ID
--id <uuid> --nickname <name>   set a job ID or nickname, not both   can only set 'id' OR 'nickname'
```

The engine rejects an empty pair itself ([rpcserver.go:4025-4027](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/engine/rpcserver.go#L4025-L4027)), so nothing goes wrong; this is consistency only. With the helper, the three commands answer as `getajob` does, every valid request goes out unchanged, and `TestGetDataHistoryJob` still passes.

thrasher- (commented):
Heimdall: Nothing has been pushed since my last review, and the three items that gate the merge still reproduce at [`9c1cf47`](https://github.com/thrasher-corp/gocryptotrader/commit/9c1cf4791e203111122bd2300b175bf8de5b9c29): [websocket_management.go:76](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640480), [futures.go:296](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640505) and [main.go:265](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4080640526).

A second pass found nothing else that changes what the CLI sends: replaying flag-form invocations of every command against master, the only requests that change are the intended fixes and what's already raised, and each documented example still reaches the RPC. Two new Nits inline: [futures.go:20](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4089327659) and [data_history.go:201](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4089327665).

- The README's Autocomplete section still links `contrib/`, but #698 moved the Bash/ZSH scripts to [`cmd/gctcli/autocomplete`](https://github.com/thrasher-corp/gocryptotrader/tree/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/autocomplete) and added a PowerShell one ([README.md:54](https://github.com/thrasher-corp/gocryptotrader/blob/9c1cf4791e203111122bd2300b175bf8de5b9c29/cmd/gctcli/README.md#L54)).

Those three items are still what I'd need before approving.

gloriousCode:
I'll have you know I forgot to press push

backtester/btcli/README.md
```diff
@@ -26,6 +26,17 @@ For a list of commands, you can run the following
 go run .
 ```
 
+Run `btcli <command> --help` to see its flags. Required flags are marked in
+the help output. Set boolean flags with `=`, for example `--archived=false`.
```
thrasher-: Heimdall: **Minor** — btcli has no `--archived` flag, so this example fails as written.

**Fix:** use one of btcli's own booleans. This applies as a pair with [the template suggestion](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4091175144); without it, the next docs regeneration deletes the paragraph:

```suggestion
the help output. Set boolean flags with `=`, for example `--donotstore=false`.
```

**Why:** `--archived` belongs to gctcli's `script` commands. btcli's only booleans are `--donotrunimmediately` and `--donotstore` ([`commands.go:20-29`](https://github.com/thrasher-corp/gocryptotrader/blob/40f7b0b1941648dfa2d5eab4dfcfd3cd9d9844f3/backtester/btcli/commands.go#L20-L29)), and every btcli command rejects the example:

```text
$ btcli executestrategyfromfile --path strategy.json --archived=false
Incorrect Usage: flag provided but not defined: -archived

$ btcli executestrategyfromfile --path strategy.json --donotstore=false
sends ExecuteStrategyFromFile{strategy_file_path: "strategy.json", do_not_store: false}
```

cmd/documentation/backtester_templates/backtester_btcli_readme.tmpl
```diff
@@ -12,5 +12,13 @@ For a list of commands, you can run the following
 go run .
 ```
 
+Supply command parameters as named flags. Positional arguments are rejected.
```
thrasher-: Heimdall: **Minor** — the paragraph 40f7b0b19 added to the btcli README isn't in this template, so the next docs regeneration deletes it.

**Fix:** add it here. This applies as a pair with [the README suggestion](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4091175117), which corrects the flag it names:

```suggestion
Run `btcli <command> --help` to see its flags. Required flags are marked in
the help output. Set boolean flags with `=`, for example `--donotstore=false`.

Supply command parameters as named flags. Positional arguments are rejected.
```

**Why:** `backtester/btcli/README.md` is generated from this file. Regenerating on this head removes the three new README lines; on master the same run only updates contributor data. [The guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/40f7b0b1941648dfa2d5eab4dfcfd3cd9d9844f3/docs/CODING_GUIDELINES.md?plain=1#L295-L297) ask to "Update the source templates under `cmd/documentation` before changing their generated Markdown". No CI job diffs regenerated docs, which is why this passed. With both suggestions applied, the generator reproduces the README byte for byte, a second run changes nothing, and markdownlint and `misc_checks` are clean.

cmd/gctcli/main_test.go
```diff
@@ -0,0 +1,157 @@
...
+
+func TestWithdrawalRequestByDate(t *testing.T) {
+	for _, limit := range []string{"2147483648", "-2147483649"} {
+		t.Run(limit, func(t *testing.T) {
+			app := &cli.App{Commands: []*cli.Command{withdrawalRequestCommand}}
+			err := app.Run([]string{"gctcli", "withdrawalrequesthistory", "bydate", "--exchange", "Binance", "--limit", limit})
+			assert.ErrorIs(t, err, errWithdrawalLimitOutOfRange, "out-of-range limit should be rejected before RPC setup")
+		})
+	}
+}
```
thrasher-: Heimdall: **Nit** — both tests check only values outside the range, so tightening either bound by one still passes.

**Fix:** also check that the two boundary values get through. They then fail at `setupClient` on the tests' empty cert path, the `os.ErrNotExist` that [`TestGetDataHistoryJob`](https://github.com/thrasher-corp/gocryptotrader/blob/40f7b0b1941648dfa2d5eab4dfcfd3cd9d9844f3/cmd/gctcli/data_history_test.go#L19-L20) already relies on:

```suggestion
func TestWithdrawalRequestByExchangeID(t *testing.T) {
	for _, tc := range []struct {
		limit   string
		wantErr error
	}{
		{limit: "2147483648", wantErr: errWithdrawalLimitOutOfRange},
		{limit: "-2147483649", wantErr: errWithdrawalLimitOutOfRange},
		{limit: "2147483647", wantErr: os.ErrNotExist},
		{limit: "-2147483648", wantErr: os.ErrNotExist},
	} {
		t.Run(tc.limit, func(t *testing.T) {
			app := &cli.App{Commands: []*cli.Command{withdrawalRequestCommand}}
			err := app.Run([]string{"gctcli", "withdrawalrequesthistory", "byexchange", "--exchange", "Binance", "--asset", "spot", "--limit", tc.limit})
			assert.ErrorIs(t, err, tc.wantErr, "limit should be rejected only outside the int32 range")
		})
	}
}

func TestWithdrawalRequestByDate(t *testing.T) {
	for _, tc := range []struct {
		limit   string
		wantErr error
	}{
		{limit: "2147483648", wantErr: errWithdrawalLimitOutOfRange},
		{limit: "-2147483649", wantErr: errWithdrawalLimitOutOfRange},
		{limit: "2147483647", wantErr: os.ErrNotExist},
		{limit: "-2147483648", wantErr: os.ErrNotExist},
	} {
		t.Run(tc.limit, func(t *testing.T) {
			app := &cli.App{Commands: []*cli.Command{withdrawalRequestCommand}}
			err := app.Run([]string{"gctcli", "withdrawalrequesthistory", "bydate", "--exchange", "Binance", "--limit", tc.limit})
			assert.ErrorIs(t, err, tc.wantErr, "limit should be rejected only outside the int32 range")
		})
	}
}
```

**Why:** changing `>` to `>=` or `<` to `<=` in either check, at [`commands.go:2706`](https://github.com/thrasher-corp/gocryptotrader/blob/40f7b0b1941648dfa2d5eab4dfcfd3cd9d9844f3/cmd/gctcli/commands.go#L2706) or [`2763`](https://github.com/thrasher-corp/gocryptotrader/blob/40f7b0b1941648dfa2d5eab4dfcfd3cd9d9844f3/cmd/gctcli/commands.go#L2763), leaves the package green. With the suggestion all four mutants fail, and the package passes `-race -count 2 -shuffle on` with lint clean.

thrasher- (approved):
Heimdall: The three items I was holding on are fixed, and replaying every command's flag forms against master turns up no unintended change to what the CLI sends. Approving; please take the paired docs fix at [README.md:30](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4091175117) and [its template](https://github.com/thrasher-corp/gocryptotrader/pull/2399#discussion_r4091175144) before merge, and consider retitling: the default squash title still says "reject mixed positional and flag arguments", but this now removes positional arguments from gctcli and btcli.

The red 386 job is Gate.io live data (`XRD_ETH` listed without an execution limit), not this diff; the same test passes locally on master and on the merge tree.

PR #2400, opened 2026-09-21
fix:gateio order & okx instIdCode

dazi005:
# PR Description

Fixes # (issue)

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


thrasher-:
@codex please review

exchanges/gateio/gateio_types.go
```diff
@@ -1885,7 +1885,7 @@ type DeliveryOrderCreateParams struct {
 // FuturesOrder represents future order response
 type FuturesOrder struct {
 	ID                        int64         `json:"id"`
-	User                      string        `json:"user"`
+	User                      types.Number  `json:"user"`
```
thrasher-: Heimdall: **Should fix** — right type, but nothing in the repo covers the shape that forced it.

`types.Number` is the only correct choice here, and the reason is worth a comment on the field: gate types this identifier differently on the two transports that share this struct. Its REST client declares `FuturesOrder.User` as `int32` — a bare JSON number ([model_futures_order.go](https://github.com/gateio/gateapi-go/blob/master/model_futures_order.go)) — while its websocket client declares the same field on the same channel as `string` ([response_future.go](https://github.com/gateio/gatews/blob/master/go/response_future.go)). `FuturesOrder` serves twelve REST methods in `gateio.go` plus the `futures.orders` push at [gateio_websocket_futures.go:699](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/gateio/gateio_websocket_futures.go#L699), so neither `string` nor `int64` can serve both. A one-line note here would stop the next person "tidying" it back to an integer.

**Why it needs a test:** every fixture the PR touched still quotes the value, so the tests only exercise the string branch — the shape that already worked. The bare-number REST shape, which is what `string` actually broke on, is untested. Reverting the fix wholesale on this branch (`User` back to `string`, `AccountID: resp.Result[x].User.String()` back to `resp.Result[x].User`) leaves the entire package green:

```
$ go test ./exchanges/gateio/ -count 1
ok  	github.com/thrasher-corp/gocryptotrader/exchanges/gateio	16.038s
```

`TestProcessFuturesOrdersPushData`, `TestProcessFinishedFuturesOrdersPushData` and `TestFuturesDataHandler` all pass against the broken type, so the `110xxxxx` → `110110110` fixture edits are making room for `types.Number` rather than guarding it.

**Fix:** a focused decode test on the struct covers both transports without touching the ws fixtures, which are correctly quoted — one `exp := FuturesOrder{...}` compared against `{"user":110110110}` and `{"user":"110110110"}`. That fails on `string` and passes here.

exchanges/gateio/gateio.go
```diff
@@ -2081,7 +2081,7 @@ func (e *Exchange) EnableOrDisableDualMode(ctx context.Context, settle currency.
 	params := url.Values{}
 	params.Set("dual_mode", strconv.FormatBool(dualMode))
 	var response *DualModeResponse
-	return response, e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, perpetualToggleDualModeEPL, http.MethodGet, futuresPath+settle.Item.Lower+"/dual_mode", params, nil, &response)
+	return response, e.SendAuthenticatedHTTPRequest(ctx, exchange.RestSpot, perpetualToggleDualModeEPL, http.MethodPost, futuresPath+settle.Item.Lower+"/dual_mode", params, nil, &response)
```
thrasher-, line 2084: Heimdall: **Should fix** — this endpoint now works, which turns its test into a real account mutation.

**Fix:** gate `TestEnableOrDisableDualMode` ([gateio_test.go:1613](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/gateio/gateio_test.go#L1613)) on `canManipulateRealOrders`, the way the other 54 mutating tests in that file already are. It currently passes only `SkipTestIfCredentialsUnset(t, e)`, so anyone running the suite with keys flips their own futures position mode.

**Why:** the verb change itself is right. Gate's generated client has `SetDualMode` as `http.MethodPost` on `/futures/{settle}/dual_mode` with `localVarQueryParams.Add("dual_mode", ...)` and an empty post body ([api_futures.go](https://github.com/gateio/gateapi-go/blob/master/api_futures.go)), which is exactly what `SendAuthenticatedHTTPRequest` emits for POST with `params` and nil `data` — `param` always goes to the query string, only `data` to the body. Gate's HTML docs page describes `dual_mode` as a body parameter; the generated client is the better source here and the page is wrong.

The `GetSinglePosition` flip on [gateio.go:2035](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/gateio/gateio.go#L2035) checks out the same way — gate has `GET /futures/{settle}/positions/{contract}`.

Both verified against gate's client and docs, not live: neither method has a non-test caller anywhere in the repo, both tests skip without credentials, and gate's auth check runs before method routing, so GET and POST both return `MISSING_REQUIRED_HEADER` unauthenticated. If you exercised either against a real key, that settles it better than I can.

exchanges/okx/okx_wrapper.go
```diff
@@ -1463,31 +1487,42 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
-	cancelAllOrdersRequestParams := make([]CancelOrderRequestParam, len(myOrders))
+	cancelAllOrdersRequestParams := make([]CancelOrderRequestParam, 0, len(myOrders))
 ordersLoop:
 	for x := range myOrders {
 		switch {
 		case orderCancellation.OrderID != "" || orderCancellation.ClientOrderID != "":
-			if myOrders[x].OrderID == orderCancellation.OrderID ||
-				myOrders[x].ClientOrderID == orderCancellation.ClientOrderID {
-				cancelAllOrdersRequestParams[x] = CancelOrderRequestParam{
+			if (orderCancellation.OrderID != "" && myOrders[x].OrderID == orderCancellation.OrderID) ||
```
thrasher-: Heimdall: **Should fix** — the rewrite fixes two of the three selection filters; the third is still a no-op, and the first can still pick the wrong order.

The `InstrumentID`, order-ID and side fixes here are real and well covered — I ran your test against [`158e9c1`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322) and all four subtests fail, and reverting just these two conditions while keeping `InstrumentID` makes `by order ID` cancel `KEEP-ME` and `by side` cancel both. Two gaps remain in the same selection logic.

**1. Supplying both IDs can cancel an unrequested order.** The condition is an `OR` followed by `break ordersLoop`, so whichever pending order matches *either* field first wins. Cancelling `CANCEL-ME` while also passing a client order ID that belongs to an earlier `KEEP-ME`:

```
requested OrderID=CANCEL-ME ClientOrderID=mine -> cancelled=[KEEP-ME]
```

Which order gets cancelled then depends on listing order rather than on anything the caller asked for. Matching `OrderID` first, or requiring both to match when both are set, would be unambiguous either way.

**2. `orderCancellation.Type` never reaches OKX.** `CancelAllOrders` passes `oType` into `GetOrderList`, which sends it as `orderType`:

```
requested Type=LIMIT -> query GCT sent: instId=BTC-USDT&instType=SPOT&orderType=limit
```

The parameter is `ordType`. OKX's own Python SDK uses `ordType` for `/api/v5/trade/orders-pending` ([Trade.py `get_order_list`](https://github.com/okxapi/python-okx/blob/master/okx/Trade.py)), and this package already spells it `ordType` in four other places — the two outliers are [okx.go:244](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx.go#L244) (`GetOrderList`) and [okx.go:289](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx.go#L289) (`Get7DayOrderHistory`), against 15 `json:"ordType"` tags and 2 `json:"orderType"` ones. OKX ignores unknown query parameters rather than rejecting them, so a caller asking to cancel only limit orders would get every open order cancelled. I could only capture the outgoing query — confirming what OKX does with the bad parameter needs credentials, so if you have seen it filter correctly, say so and I will drop this half.

Both are pre-existing causes, but this PR is what makes them observable: on the base commit the whole function failed with `missing instrument ID` before any of it mattered, and fixing two of the three filters in this block is a natural place to finish the third. Reachability keeps this off the blocker list — `RPCServer.CancelAllOrders` passes `nil` ([rpcserver.go:1439](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/engine/rpcserver.go#L1439)) and `OrderManager` cancels singly, so populated filters arrive only from `cmd/exchange_wrapper_issues` and library callers. Happy to split these into their own issue if you would rather keep this PR to `instIdCode`.

exchanges/okx/okx_wrapper.go
```diff
@@ -1352,6 +1359,21 @@ func (e *Exchange) CancelBatchOrders(ctx context.Context, o []order.Cancel) (*or
...
+			}
+			if e.Websocket.CanUseAuthenticatedWebsocketForWrapper() {
+				_, err = e.WSCancelSpreadOrder(ctx, ord.OrderID, ord.ClientOrderID)
+			} else {
+				_, err = e.CancelSpreadOrder(ctx, ord.OrderID, ord.ClientOrderID)
+			}
+			if err != nil {
+				return nil, err
+			}
+			resp.Status[ord.OrderID] = order.Cancelled.String()
```
thrasher-: Heimdall: **Should fix** — spread cancels given only a client order ID all collapse into one empty status key.

**Fix:** fall back to `ord.ClientOrderID` when `ord.OrderID` is empty, or key off the ID the response carries. The guard five lines up deliberately accepts either identifier, so an empty `ord.OrderID` is a supported input here rather than a caller error.

**Why:** two spread cancels by client order ID against an `httptest` server standing in for `sprd/cancel-order` — both requests go out, one meaningless entry comes back:

```
API calls made: 2
resp.Status = map[string]string{"":"CANCELLED"}
--- FAIL: "map[:CANCELLED]" should have 2 item(s), but has 1
```

Two more on this branch while you are here:

- A failure part-way through `return nil, err`s, discarding the statuses of spread orders already cancelled by earlier iterations. Each iteration is its own API call, so those cancels are real and the caller cannot learn which succeeded — unlike the single-call batch path below, where there is nothing to lose.
- The status is written unconditionally on success, where the batch path below checks `StatusCode != 0` first.

**Why it got through CI:** the assertion added at [okx_test.go:3876](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_test.go#L3876) passes on the base commit too. With neither ID set, [`158e9c1`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322) falls through to the order-type switch and returns the same sentinel, just with a different message:

```
b772665: order id or client order id is not set, order ID required for spread order cancel
158e9c1: order id or client order id is not set, order ID required for order of type UNKNOWN
```

So the only genuinely new path in this function — a spread cancel that succeeds — has no coverage, which is where the defect above lives.

**Separately, and the reason this is worth a close look:** your new branch checks `ord.ClientOrderID`, which is right. The ordinary-order guard five lines below at [okx_wrapper.go:1379](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_wrapper.go#L1379) checks `ClientID` — a different field, commented `// TODO: Shift to credentials` in `order.Cancel` — while the request it builds populates `ClientOrderID`. Both spellings therefore fail:

```
spot   batch cancel, ClientOrderID only -> order id or client order id is not set, order ID required for order of type UNKNOWN
spot   batch cancel, ClientID only      -> order id or client order id is not set
spread batch cancel, ClientOrderID only -> (reaches the API)
```

The first is rejected by the guard; the second passes the guard and is then rejected by `CancelMultipleOrders`, because `ClientOrderID` was never populated. Batch-cancelling an ordinary order by client order ID is impossible today. Pre-existing, but your branch is the correct model sitting directly above it, so it is a one-word change while you are in this function.

exchanges/okx/okx_wrapper.go
```diff
@@ -1463,31 +1487,42 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
 				ClientOrderID: myOrders[x].ClientOrderID,
+			})
+		}
+	}
+	useWebsocket := e.Websocket.CanUseAuthenticatedWebsocketForWrapper()
+	if useWebsocket {
+		for x := range cancelAllOrdersRequestParams {
+			cancelAllOrdersRequestParams[x].InstrumentIDCode = e.getInstrumentIDCode(cancelAllOrdersRequestParams[x].InstrumentID)
+			if cancelAllOrdersRequestParams[x].InstrumentIDCode == 0 {
+				return cancelAllResponse, fmt.Errorf("%w: %s", errMissingInstrumentIDCode, cancelAllOrdersRequestParams[x].InstrumentID)
```
thrasher-: Heimdall: **Minor** — aborting is the wrong failure mode for cancel-all, and the batch loop below already hides real failures.

**Fix:** when a code is missing, fall back to `CancelMultipleOrders` for the whole set rather than returning. Cancel-all is the call you least want to fail under stress, and the REST path needs no `instIdCode` at all — your own `TestCancelAllOrdersMatchesOnlyRequestedOrders` proves that, since it drives the REST path with no codes set and passes. As written, one unknown instrument blocks cancelling every other order.

**While you are in this loop**, the batching below it swallows failures. [okx_wrapper.go:1547](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_wrapper.go#L1547) returns an error only while the status map is still empty, then falls through to `return cancelAllResponse, nil`. With 40 pending orders and an API error on the second batch:

```
pending=40 batches=2 reported=20 err=<nil>
```

Twenty orders are still live and the caller is told nothing went wrong. Pre-existing, and invisible until now because the function never got as far as a second batch on the base commit — but it is the same "did my cancel-all actually work" question as the guard above, so worth settling together.

Related, and a smaller thing: `CancelAllOrders` takes whatever single page `GetOrderList` returns and never follows `before`/`after`, and it sets no `limit`, so on an account with more open orders than OKX's default page size the extras are silently left alone. I did not manage to pull OKX's documented default and maximum for that endpoint to give you the exact number, so treat the cap as unconfirmed — the missing pagination loop is plain in the code either way.

exchanges/okx/okx_wrapper.go
```diff
@@ -2216,10 +2249,33 @@ func (e *Exchange) getInstrumentsForAsset(ctx context.Context, a asset.Item) ([]
...
+// channel pushes deltas, not full snapshots.
+func (e *Exchange) cacheInstrumentIDCodes(instruments []Instrument) {
+	e.instrumentsInfoMapLock.Lock()
+	defer e.instrumentsInfoMapLock.Unlock()
+	e.cacheInstrumentIDCodesLocked(instruments)
+}
+
+func (e *Exchange) cacheInstrumentIDCodesLocked(instruments []Instrument) {
+	for x := range instruments {
+		e.instrumentIDCodeMap[instruments[x].InstrumentID.String()] = instruments[x].InstrumentIDCode
```
thrasher-, line 2341: Heimdall: **Minor** — this write can replace a good code with zero, and OKX documents the value that does it.

**Fix:** skip the zero — `if c := instruments[x].InstrumentIDCode; c != 0 { e.instrumentIDCodeMap[...] = c }`. A stale-but-valid code is recoverable; a zeroed one turns every subsequent websocket order for that instrument into `errMissingInstrumentIDCode`.

**Why:** OKX documents this field as nullable — *"Instrument ID code. For simple binary encoding, you must use `instIdCode` instead of `instId`. For the same `instId`, it's value may be different between production and demo trading. It is `null` when the value is not generated."* ([Get instruments](https://www.okx.com/docs-v5/en/#public-data-rest-api-get-instruments)). Seeding a good code, then decoding a push carrying `null` and feeding it to `cacheInstrumentIDCodes`:

```
seeded BTC-USDT = 3
decoded 1 instrument(s), InstrumentIDCode=0     <- null decodes to 0, list still decodes fine
after a null push, BTC-USDT = 0
```

Live public responses are clean today: across every live instrument on each type — spot 1410, margin 266, perpetual swap 486, futures 225, options 3086 — all 5473 resolved a non-zero code, so this is defensive hardening against a documented case rather than a defect I can show you in production. It is one line against a failure that surfaces as a rejected order rather than at the push.

**One correction to the GoDoc above:** the channel path is narrower than the comment implies. `instruments` is not in `defaultSubscriptions` ([okx_websocket.go:217](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_websocket.go#L217)), and configuring the raw channel name is not enough either — the subscription template emits `{"channel":"instruments"}` with no `instType`, which live OKX rejects with error 60018; adding `instType` by hand gets a successful `subscribe`. So "instruments listed after the startup fetch resolve codes in a running process" does not hold for any ordinary configuration. Worth either softening the comment or wiring the channel up properly, since as it stands the upsert is unreachable outside tests.

exchanges/okx/ws_requests.go
```diff
@@ -29,6 +29,10 @@ func (e *Exchange) WSPlaceOrder(ctx context.Context, arg *PlaceOrderRequestParam
 		return nil, err
 	}
 
+	if arg.InstrumentIDCode == 0 {
+		return nil, errMissingInstrumentIDCode
```
thrasher-: Heimdall: **Minor** — this error names neither the instrument nor the remedy, and an empty cache is reachable at startup.

**Fix:** wrap with the instrument, as `CancelAllOrders` already does at [okx_wrapper.go:1525](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_wrapper.go#L1525) — `fmt.Errorf("%w: %s", errMissingInstrumentIDCode, arg.InstrumentID)` — at all six guards. Cleaner still: have `getInstrumentIDCode` return `(uint64, error)`, since `SubmitOrder`, `ModifyOrder`, `CancelOrder` and `CancelBatchOrders` currently pass a zero straight through and leave these guards to do the talking, so the six messages are the only diagnostic a user gets.

**Why it matters:** "missing instrument ID code" sits one word away from the existing `errMissingInstrumentID` ("missing instrument ID") and tells an operator nothing about which instrument or what to do about it. The cache is normally warm — `Bootstrap` calls `UpdateOrderExecutionLimits` for every enabled asset unconditionally, so this works even with `AutoPairUpdates` off — but a failed startup leaves it empty while the exchange stays live. Pointing the endpoints at a 502 and running `exchange.Bootstrap`:

```
Bootstrap err = failed to update tradable pairs: Okx unsuccessful HTTP status code: 502 raw response: upstream unavailable
cached instrument ID codes after failed Bootstrap = 0
getInstrumentIDCode("BTC-USDT") = 0
WSPlaceOrder with cold cache = missing instrument ID code
```

That state survives, because `bot.ExchangeManager.Add(exch)` ([engine.go:886](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/engine/engine.go#L886)) runs *before* `exchange.Bootstrap(ctx, exch)` ([engine.go:911](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/engine/engine.go#L911)) and `SetupExchanges` only logs the failure ([engine.go:1036](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/engine/engine.go#L1036)). A transient instruments fetch failure therefore leaves a registered, authenticated exchange whose every websocket order, amend and cancel fails with this bare sentinel, while REST would have worked fine. Naming the instrument — and ideally saying the instruments cache is empty — turns that from a mystery into a restart.

exchanges/okx/okx_wrapper.go
```diff
@@ -1192,15 +1194,18 @@ func (e *Exchange) ModifyOrder(ctx context.Context, action *order.Modify) (*orde
...
+	instrumentID := pairFormat.Format(action.Pair)
 	switch action.Type {
 	case order.UnknownType, order.Market, order.Limit, order.OptimalLimit, order.MarketMakerProtection:
 		amendRequest := AmendOrderRequestParams{
-			InstrumentID:  pairFormat.Format(action.Pair),
+			InstrumentID:  instrumentID,
 			NewQuantity:   action.Amount,
 			OrderID:       action.OrderID,
 			ClientOrderID: action.ClientOrderID,
+			NewPrice:      action.Price,
```
thrasher-, line 1205: Heimdall: **Nit** — real fix, worth a guard.

**Why:** before this line an amend could never change a price — `NewPrice` was left zero, so a price-only amend was rejected locally and never reached OKX. Same call on each side, `Type: order.Limit`, `Price: 21000`, `Amount: 0`:

```
158e9c1:  price-only amend -> err=invalid new size or price information
b772665:  price-only amend -> err=error checking credentials: Okx REST or Websocket authentication support is not enabled
```

**Fix:** that comparison is the test, and it needs no credentials — a `ModifyOrder` with those three fields plus `AssetType: asset.Spot` and an `OrderID`, asserted with `require.NotErrorIs(t, err, errInvalidNewSizeOrPriceInformation)`, fails on the base commit and passes here. `TestModifyOrder` already reaches this branch unauthenticated at [okx_test.go:3951](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_test.go#L3951), so it is a two-line addition there.

exchanges/okx/okx_wrapper_test.go
```diff
@@ -17,10 +27,210 @@ func TestMessageID(t *testing.T) {
...
+	assert.EqualValues(t, 11111, fresh.getInstrumentIDCode("SOL-USDT"), "a pushed SOL-USDT should upsert its instrument ID code")
+	assert.Len(t, fresh.instrumentsInfoMap[instTypeSpot], 2, "an instruments push should not replace the cached instrument list")
+
+	assert.Zero(t, fresh.getInstrumentIDCode("DOGE-USDT"), "an uncached instrument ID should resolve a zero code")
+}
+
+func TestCancelAllOrdersMatchesOnlyRequestedOrders(t *testing.T) {
+	// Subtests share the mock server's pending order queue, so they run
+	// sequentially.
+	okx := new(Exchange)
```
thrasher-: Heimdall: **Nit** — two small things in this test, plus one note on what it proves.

`okx` shadows the package name; `TestMessageID` in this same file and nine tests in `okx_test.go` use `e := new(Exchange)`.

`btcUSDT` at [okx_wrapper_test.go:167](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_wrapper_test.go#L167) is the same value as the package-level `mainPair` ([okx_test.go:67](https://github.com/thrasher-corp/gocryptotrader/blob/b77266515a814c9f16aaa68d010e3bd8127264bd/exchanges/okx/okx_test.go#L67)), and since it only ever reaches `pairFormat.Format`, the delimiter is irrelevant there — `mainPair` will do. The three at lines 72-74 genuinely do need `NewPairWithDelimiter`, because `InstrumentID.String()` is the cache key.

On what the subtests guard: three of the four do real work. Reverting only the two match conditions while keeping `InstrumentID` makes `by order ID` cancel `KEEP-ME` instead of `CANCEL-ME` and `by side` cancel both `BUY-1` and `SELL-1`, and both subtests catch it. `by client order ID` stays green under that revert — with real `ordId`s on both pending orders the old condition cannot false-positive in that direction — so it is coverage rather than a guard. Fine as is; just worth knowing it is not the thing protecting the client-order-ID discriminator.

thrasher- (approved):
Heimdall: Approving — the `instIdCode` work is correct and lands something OKX now requires, and the `CancelAllOrders` rewrite repairs a function that failed outright.

Verified: all six deprecated websocket ops covered, with mass-cancel and the spread ops correctly left alone; cache keys resolve on all 5473 live instruments, zero misses; the 266 instIds shared by spot and margin carry identical codes; `CancelAllOrders` never set `InstrumentID` on [`158e9c1`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322), where your new test fails all four subtests; both gate.io verbs match gate's generated client. Lint clean, `-race` and `-count 2` green.

Nine inline comments carry the evidence — four should-fix, three minor, two nits. **None gate the merge.** The description is still the unfilled template, which hides four changes the title omits.

shazbert (changes requested):
@dazi005 This is looking great, if you can continue with the nits outlined above to harden everything up then I will approve. Thanks heaps!

exchanges/okx/okx_wrapper.go
```diff
@@ -1455,64 +1483,95 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
 				response, err = e.CancelMultipleOrders(ctx, remaining)
 			}
 		}
 		if err != nil {
-			if len(cancelAllResponse.Status) == 0 {
-				return cancelAllResponse, err
-			}
+			// Statuses already collected are returned alongside the error so a
+			// mid-loop failure does not discard them.
+			return cancelAllResponse, err
```
thrasher-: Heimdall: **Should fix** — a partial success in any batch now ends the cancel-all, so every later batch is never sent.

**Fix:** keep going and join the errors, recording whatever per-order results come back with them, and stop early only when the context is done, returning its error:

```go
	var errs error
	for range loop {
		if err := ctx.Err(); err != nil {
			if !errors.Is(errs, err) {
			}
			return cancelAllResponse, errs
		}
		// ...send the batch as now...

		// OKX answers a batch that only partly cancelled with an error, and the
		// websocket path still returns its per-order results, so record them and
		// keep cancelling the remaining batches.
		for y := range response {
			// ...record each status as now...
		}
	}
	return cancelAllResponse, errs
```

Returning the context error matters. Breaking without it would report success for batches never sent. Stopping on the context rather than on every error is deliberate. On REST, a partial success and a request-level failure (say, a read-only key's `50120`) reach this loop as the same kind of error, because `SendHTTPRequest` returns before decoding `data`. Repeating a permission failure once per batch costs far less than leaving orders live.

A `"code":"2"` reply to the first of two batches in the `TestCancelAllOrdersMatchesOnlyRequestedOrders` mock covers it.

**Why:** OKX answers a batch in which only some orders succeed with `"code":"2"`. Its [REST error codes](https://www.okx.com/docs-v5/en/#error-code-rest-api-public) list `2 | Bulk operation partially succeeded`, and the [websocket batch-cancel docs](https://www.okx.com/docs-v5/en/#order-book-trading-trade-ws-cancel-multiple-orders) show it for exactly this case: `sCode` `5XXXX` "order not exist" beside a success. That is the normal race for a cancel-all in a moving market, where an order fills between the list and the cancel. Both transports turn it into an error, and this `return` abandons the rest. 40 pending orders, first batch partially successful:

```
REST:       cancel requests=1  never sent=20  statuses=0  err=authenticated request failed, bulk operation partially succeeded
websocket:  cancel requests=1  never sent=20  statuses=0  err=bulk operation partially succeeded, *okx.OrderData[1]: status code: `51400` ...
```

The websocket row is the sharper one. `WSCancelMultipleOrders` returns the per-order results together with the error (`X1:sCode=51400 X2:sCode=0`), so the caller is told nothing was cancelled while 19 orders were. With the loop continuing, the same run sends all 40, reports 40 statuses, and still returns the error. REST can't report the failed batch's per-order results either way, since `SendHTTPRequest` returns before decoding `data` on a non-zero code ([okx.go:5939](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx.go#L5939)), but it can still send the remaining batches.

This follows from [the earlier thread on swallowed batch errors](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4068491800): surfacing the error was right, but the loop doesn't need to stop to do it.

exchanges/okx/okx_wrapper.go
```diff
@@ -1455,64 +1483,95 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
+		for x := range cancelAllOrdersRequestParams {
+			code := e.getInstrumentIDCode(cancelAllOrdersRequestParams[x].InstrumentID)
+			if code == 0 {
+				// OKX documents instIdCode for websocket operations only, so
+				// an uncached instrument falls back to REST, which identifies
+				// orders by instId alone.
+				useWebsocket = false
+				break
 			}
+			cancelAllOrdersRequestParams[x].InstrumentIDCode = code
```
thrasher-: Heimdall: **Minor** — the REST fallback can still send `instIdCode`, which [the GoDoc on `getInstrumentIDCode`](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L3191-L3193) says REST bodies must not carry.

**Fix:** resolve every code before assigning any, so a miss leaves the requests untouched. As a small helper, `CancelBatchOrders` can share it (see the comment on `SubmitOrder`):

```go
// applyWebsocketInstrumentIDCodes sets every request's instrument ID code and reports true
// only when the websocket can be used and every code is cached; otherwise it leaves the
// requests untouched so the caller falls back to REST.
func (e *Exchange) applyWebsocketInstrumentIDCodes(args []CancelOrderRequestParam) bool {
	if !e.Websocket.CanUseAuthenticatedWebsocketForWrapper() {
		return false
	}
	codes := make([]uint64, len(args))
	for x := range args {
		if codes[x] = e.getInstrumentIDCode(args[x].InstrumentID); codes[x] == 0 {
			return false
		}
	}
	for x := range args {
		args[x].InstrumentIDCode = codes[x]
	}
	return true
}
```

**Why:** this loop writes codes in order until the first uncached instrument, then the `break` leaves the earlier ones set, and the whole list goes to `CancelMultipleOrders`. With only BTC-USDT cached, an authenticated websocket, and a spot cancel-all with no pair:

```
listed BTC-USDT, ETH-USDT -> REST body [{"instId":"BTC-USDT","instIdCode":12345,"ordId":"A"},{"instId":"ETH-USDT","ordId":"B"}]
listed ETH-USDT, BTC-USDT -> REST body [{"instId":"ETH-USDT","ordId":"A"},{"instId":"BTC-USDT","ordId":"B"}]
```

So what REST receives depends on listing order. I couldn't test whether authenticated REST rejects the extra field without credentials. If OKX ignores it, this is only a contradiction with the GoDoc, and the GoDoc is what should change. The fallback itself is from [the earlier thread](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4068491800).

exchanges/okx/okx_wrapper.go
```diff
@@ -999,6 +1000,7 @@ func (e *Exchange) SubmitOrder(ctx context.Context, s *order.Submit) (*order.Sub
 			}
 		}
 		if e.Websocket.CanUseAuthenticatedWebsocketForWrapper() {
+			orderRequest.InstrumentIDCode = e.getInstrumentIDCode(pairString)
```
thrasher-: Heimdall: **Minor** — `CancelAllOrders` now falls back to REST when a code isn't cached, but this call and three others don't, so websocket orders on instruments listed after startup fail outright.

**Fix:** use the websocket only when the code is cached, here and at [okx_wrapper.go:1208](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1208) (`ModifyOrder`) and [okx_wrapper.go:1315](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1315) (`CancelOrder`):

```go
		if code, ok := e.websocketInstrumentIDCode(pairString); ok {
			orderRequest.InstrumentIDCode = code
			placeOrderResponse, err = e.WSPlaceOrder(ctx, orderRequest)
		} else {
			placeOrderResponse, err = e.PlaceOrder(ctx, orderRequest)
		}
```

where `websocketInstrumentIDCode` returns the cached code and `CanUseAuthenticatedWebsocketForWrapper() && code != 0`. `CancelBatchOrders` at [okx_wrapper.go:1404](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1404) takes the all-or-nothing helper from the comment on the `CancelAllOrders` fallback.

**Why:** the cache is refreshed at startup and on a manual pair update (`FetchTradablePairs`, `UpdateOrderExecutionLimits`). Otherwise only the `instruments` channel refreshes it, and that channel isn't subscribed by default. Configuring it gets OKX error `60018`, because the template sends `{"channel":"instruments"}` with no `instType`. Meanwhile OKX lists continuously. Live today:

- the futures list holds contracts listed on eight separate days since 11 September, three to six at a time;
- BTC options gained between 8 and 234 strikes a day over the past week;
- four swaps listing between 09:00 and 09:45 UTC today (`KSTR`, `MSTU`, `CYPH`, `GTLB`-USDT-SWAP) returned `"instIdCode": null` at 05:12 UTC and had codes by 07:23 UTC, so any process that fetched instruments before the codes appeared holds none for them.

For any of these, with an authenticated websocket:

```
SubmitOrder KSTR-USDT-SWAP, instruments fetched while its code was null -> missing instrument ID code: KSTR-USDT-SWAP   (nothing sent)

cold cache, same mock for all five:
  SubmitOrder, ModifyOrder, CancelOrder, CancelBatchOrders -> missing instrument ID code: BTC-USDT   REST requests=0
  CancelAllOrders                                          -> <nil>   REST requests=[/trade/orders-pending /trade/cancel-batch-orders]
```

REST needs no code, so each of those four calls succeeds against the same mock once it takes that path. The comment you added in `CancelAllOrders` states the rule, and it holds at every call site.

exchanges/gateio/gateio_test.go
```diff
@@ -517,6 +517,28 @@ func TestFuturesBaseVolume(t *testing.T) {
...
+		expected float64
+	}{
+		{"REST sends a bare number", `{"id":123,"user":110110110}`, 110110110},
+		{"websocket sends a quoted string", `{"id":123,"user":"110110110"}`, 110110110},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			var resp FuturesOrder
+			require.NoErrorf(t, json.Unmarshal([]byte(tc.payload), &resp), "Unmarshal must not error for %s", tc.payload)
+			assert.Equal(t, tc.expected, resp.User.Float64(), "User should decode from both shapes")
```
thrasher-: Heimdall: **Nit** — compare one expected struct rather than a single field, as `TestUsageUnmarshal` in `currency/forexprovider/openexchangerates` does:

```go
	exp := FuturesOrder{ID: 123, User: 110110110}
	...
			assert.Equal(t, exp, resp, "FuturesOrder should decode User from both shapes")
```

That also drops the `expected float64` column. The test already does its job: with the assertion made type-neutral, `User string` fails the REST subtest (`json: cannot unmarshal number into Go struct field FuturesOrder.user of type string`) and `User int64` fails the websocket one.

exchanges/okx/okx_wrapper_test.go
```diff
@@ -17,10 +28,334 @@ func TestMessageID(t *testing.T) {
...
+		encoded, err := json.Marshal(map[string]any{"code": "0", "msg": "", "data": data})
+		if err != nil {
+			t.Errorf("marshalling mock response should not error: %v", err)
+			return
+		}
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write(encoded)
+	}
+
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```
thrasher-: Heimdall: **Nit** — `httptest.NewTestServer` here and at [okx_wrapper_test.go:328](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper_test.go#L328), as the rest of the tests do (43 uses; master has a single `NewServer`, in `exchanges/request`):

```go
	srv := httptest.NewTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
	...
	require.NoError(t, e.SetHTTPClient(srv.Client()), "SetHTTPClient must not error")
```

It shuts itself down, so `t.Cleanup(srv.Close)` goes. Both tests pass that way under `-race -count 3`.

thrasher- (changes requested):
Heimdall: Requesting changes. Eight of last round's nine threads are fixed, most of them pinned by a test that fails when the fix is reverted. The zero-code guard already earns its place: four swaps listing today returned `"instIdCode": null` at 05:12 UTC. Two things block. A partial success in `CancelAllOrders` now stops every later batch, and the spread status-key fix landed in `CancelAllOrders` rather than `CancelBatchOrders`.

**To unblock:**
1. [Keep cancelling after a partially successful batch](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4080081079).
2. [The spread branch of `CancelBatchOrders`](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4068491795): the empty status key, a spread order cancelled before the rest of the batch validates, and the untested `ClientOrderID` guard.

Inline, not blocking: Minors on [the REST fallback's leftover codes](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4080081090) and [the four call sites without that fallback](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4080081095), and Nits on [the gateio decode test](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4080081099) and [`NewTestServer`](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4080081104).

- `CancelAllOrders`' spread branch passes `OrderID` as `sprdId` and ignores `Pair`, so cancelling one spread's orders sends `{}` to `sprd/mass-cancel` and cancels every spread order: [okx_wrapper.go:1455](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1455).
- `CancelAllOrders` filters with `Pair.Upper().String()` rather than the pair format, so `currency.NewBTCUSDT()` queries `instId=BTCUSDT`, which OKX rejects with `51001`: [okx_wrapper.go:1484](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1484).
- `CancelBatchOrders` returns `resp, nil` when the algo cancel fails but anything before it succeeded, so a failed algo cancellation reads as success: [okx_wrapper.go:1424-1432](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1424-L1432).

Happy to approve once those two are in.

exchanges/okx/okx_wrapper.go
thrasher-: Heimdall: **Still open.** This line is unchanged: two spread cancels by client order ID still make two API calls and report one entry, `map[string]string{"":"CANCELLED"}`. The `ClientOrderID` fallback went into the spread branch of `CancelAllOrders` instead ([okx_wrapper.go:1459](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1459)), where it has no effect: that branch passes `OrderID` as `sprdId` and never sends a client order ID.

There is one more problem on this branch, which I missed last time. The spread cancel runs inside the validation loop, so an invalid entry later in the batch is caught only after the spread order is gone. Batch `[spread OrderID S1, spot with no IDs]`:

```
b772665, cfeb17a: requests=[/sprd/cancel-order]  resp=<nil>  err=order id or client order id is not set, order ID required for order of type UNKNOWN
6619da5 (master): requests=[]                    resp=<nil>  err=order id or client order id is not set, order ID required for order of type UNKNOWN
```

The caller is told the batch failed validation, while `S1` has been cancelled.

**Fix:** collect spread cancels in the loop and run them after it. Key each by the `ordId` the cancel response returns, which OKX documents for both `sprd/cancel-order` and `sprd-cancel-order`, or else by the requested `OrderID`. Never key by the client order ID: this map also holds the exchange order IDs the ordinary path writes, so a spread `clOrdId` of `123` would overwrite a spot `ordId` of `123`. Check the response for nil first, because a `data:null` or `[null]` reply otherwise panics. A successful reply with no `ordId` to a cancel that sent only a client order ID is malformed, so return an error rather than invent a key. Finally, check the reply's `sCode` as the ordinary batch below does, rather than recording every reply as cancelled. If a spread cancel fails, return `resp` with the statuses already recorded. Do the same when the ordinary batch below fails, after recording what it returned, since the spread orders before it are already cancelled. Against the same mocks, both client-ID-only cancels come back under their own `ordId`, nothing is sent before the validation error, and a failing ordinary batch still reports the spread order it follows.

The guard change at [okx_wrapper.go:1379](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_wrapper.go#L1379) is right, but untested: reverting it to `ClientID` leaves the package green. Two lines in `TestCancelBatchOrders`, ahead of the credentials skip, catch it:

```go
	_, err = e.CancelBatchOrders(contextGenerate(), []order.Cancel{{AssetType: asset.Spot, Pair: mainPair, ClientOrderID: "my-order"}})
	assert.NotErrorIs(t, err, order.ErrOrderIDNotSet, "a client order ID alone should identify an order to cancel")
```

The spread assertion at [okx_test.go:3876-3881](https://github.com/thrasher-corp/gocryptotrader/blob/cfeb17a8cc249ac5dfdf53d43c0cb464f5430a03/exchanges/okx/okx_test.go#L3876-L3881) still passes on master too (`order ID required for order of type UNKNOWN`), so no committed test reaches this branch's success path, which is where both defects live.

exchanges/okx/okx_wrapper_test.go
```diff
@@ -17,10 +28,643 @@ func TestMessageID(t *testing.T) {
...
+		_, err := e.CancelAllOrders(t.Context(), &order.Cancel{
+			AssetType: asset.Spot,
+			Pair:      slashPair,
+		})
+		require.NoError(t, err, "CancelAllOrders must not error for a differently delimited pair")
+		mu.Lock()
+		query := pendingQuery
+		mu.Unlock()
+		assert.Contains(t, query, "instId=BTC-USDT", "the instrument filter should use the exchange's pair format")
+		assert.NotContains(t, query, "BTC%2FUSDT", "a raw slash-delimited pair should not reach OKX as instId")
```
thrasher-: Heimdall: **Blocker** — this line fails `miscellaneous checks`: the testify f-variant check at [misc_checks.sh:156](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/scripts/misc_checks.sh#L156) reads `%2F` as a format verb.

**Fix:** delete it. The assertion above already fails if the unformatted pair reaches OKX:

```suggestion
```

**Why:** from the [failed job](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35845796092/job/107131511698), and the same with `./scripts/misc_checks.sh` locally:

```
exchanges/okx/okx_wrapper_test.go:280:  assert.NotContains(t, query, "BTC%2FUSDT", "a raw slash-delimited pair should not reach OKX as instId")
FAIL: Replace func with the `…f` func variant (e.g. Equalf, Errorf)
```

With the line deleted every misc check passes, and the subtest still catches what it was written for. With `CancelAllOrders` sending the pair unformatted:

```
okx_wrapper_test.go:279: "instId=BTC%2FUSDT&instType=SPOT" does not contain "instId=BTC-USDT"
```

exchanges/okx/okx_wrapper.go
```diff
@@ -1375,10 +1395,25 @@ func (e *Exchange) CancelBatchOrders(ctx context.Context, o []order.Cancel) (*or
...
 		}
 	}
-	resp := &order.CancelBatchResponse{Status: make(map[string]string)}
+	for x := range cancelSpreadOrderParams {
+		ord := cancelSpreadOrderParams[x]
+		if e.Websocket.CanUseAuthenticatedWebsocketForWrapper() {
+			_, err = e.WSCancelSpreadOrder(ctx, ord.OrderID, ord.ClientOrderID)
+		} else {
+			_, err = e.CancelSpreadOrder(ctx, ord.OrderID, ord.ClientOrderID)
+		}
```
thrasher-: Heimdall: **Should fix** — with an authenticated websocket, every spread cancel here fails: `WSCancelSpreadOrder` sends on the private connection, and OKX accepts spread operations only on `/ws/v5/business`.

**Fix:** cancel spreads over REST, as the spread branch of `CancelAllOrders` already does:

```suggestion
		// OKX accepts spread operations only on its business websocket, and
		// WSCancelSpreadOrder sends on the private one, so spread cancels use REST.
		_, err = e.CancelSpreadOrder(ctx, ord.OrderID, ord.ClientOrderID)
```

The loop in [the spread-branch thread](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4068491795) includes this and keeps the reply, which it needs for the status key.

**Why:** OKX checks the URL before the login, so this shows without credentials. The same operation on each endpoint, live:

```
/ws/v5/private   sprd-cancel-order -> {"code":"60028","msg":"The current operation is not supported by this URL. Please use the correct WebSocket URL for the operation."}
/ws/v5/business  sprd-cancel-order -> {"code":"60011","msg":"Please log in"}
```

`SendAuthenticatedWebsocketRequest` always takes `privateConnection` ([ws_requests.go:283](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/ws_requests.go#L283)). With the websocket forced to authenticated, and that `60028` reply scripted on the private connection, a batch holding one spread order sends a single `sprd-cancel-order` there, nothing to the business connection or REST, and returns `status code: 60028`. With the change it goes to `sprd/cancel-order` over REST and reports the order cancelled. Spread websocket support is already disabled in `SetDefaults` ([okx_wrapper.go:70-71](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L70-L71)) until the business connection is wired up.

exchanges/okx/okx_wrapper.go
```diff
@@ -1425,14 +1458,28 @@ func (e *Exchange) CancelAllOrders(ctx context.Context, orderCancellation *order
...
+		success, err = e.CancelAllSpreadOrders(ctx, spreadID)
 		if err != nil {
 			return cancelAllResponse, err
 		}
-		cancelAllResponse.Status[orderCancellation.OrderID] = strconv.FormatBool(success)
+		statusKey := orderCancellation.OrderID
+		if statusKey == "" {
+			statusKey = orderCancellation.ClientOrderID
+		}
+		cancelAllResponse.Status[statusKey] = strconv.FormatBool(success)
```
thrasher-: Heimdall: **Nit** — this key names an order the mass cancel never targeted, or is empty.

**Fix:** key the result by the scope the request sent:

```suggestion
		cancelAllResponse.Status[spreadID] = strconv.FormatBool(success)
```

At [okx_wrapper_test.go:623](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper_test.go#L623), which currently pins the empty key, that becomes `assert.Equal(t, map[string]string{"BTC-USDT_BTC-USDT-SWAP": "true"}, resp.Status, "the result should be reported under the spread it cancelled")`.

**Why:** now that `sprdId` comes from `Pair`, `OrderID` and `ClientOrderID` scope nothing here:

```
{Spread, Pair}                                  -> sprdId BTC-USDT_BTC-USDT-SWAP     status {"": "true"}
{Spread, Pair, OrderID: "2341161427393388544"}  -> sprdId BTC-USDT_BTC-USDT-SWAP     status {"2341161427393388544": "true"}
{Spread, ClientOrderID: "mine"}                 -> {}, every spread order            status {"mine": "true"}
```

The second reads as one order cancelled, when every order on that spread was.

exchanges/okx/okx_wrapper_test.go
```diff
@@ -17,10 +28,643 @@ func TestMessageID(t *testing.T) {
...
+		mu.Lock()
+		last := massCancelBodies[len(massCancelBodies)-1]
+		mu.Unlock()
+		assert.Equal(t, "{}", last, "an unscoped cancel should omit sprdId so OKX cancels every spread order")
+	})
+}
+
+// TestCancelBatchOrdersReportsAlgoResults guards the per-item algo cancel
+// results: before the array decode, a batch response failed to decode and the
+// swallowed error left every algo reported as Cancelled.
```
thrasher-: Heimdall: **Nit** — before this fix no algo was reported as Cancelled; a reply for more than one algo order failed to decode instead.

**Fix:**

```suggestion
// TestCancelBatchOrdersReportsAlgoResults guards the per-item algo cancel
// results: before the array decode, a response for more than one algo order
// failed to decode, so the batch errored, or dropped the algo statuses without
// an error when ordinary orders in the same batch were cancelled.
```

**Why:** this test's mock against the base commit [`6619da5`](https://github.com/thrasher-corp/gocryptotrader/commit/6619da514cef80207dc09965205febef63f59874):

```
two algo orders:              resp=<nil>  err=cannot unmarshal response data directly (…) or as a single-item array (got 2 items)
a limit order and two algos:  resp=&{Status:map[O1:CANCELLED]}  err=<nil>
```

exchanges/okx/okx_wrapper_test.go
```diff
@@ -17,10 +28,643 @@ func TestMessageID(t *testing.T) {
...
+			http.NotFound(w, r)
+		}
+	}))
+
+	b := e.GetBase()
+	b.SkipAuthCheck = true
+	require.NoError(t, e.SetHTTPClient(srv.Client()), "SetHTTPClient must not error")
+	for k := range b.API.Endpoints.GetURLMap() {
+		require.NoErrorf(t, b.API.Endpoints.SetRunningURL(k, srv.URL+"/"), "Setup must point endpoint %s at the mock server", k)
+	}
```
thrasher-: Heimdall: **Nit** — this block, with the `new(Exchange)` and `testexch.Setup` above it, is repeated in all six mock tests in this file, and `writeOKXData` is defined twice. The [coding guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/docs/CODING_GUIDELINES.md#test-deduplication) make deduplication the default, and one helper covers all six:

```go
// newMockExchange returns an exchange whose REST endpoints all point at a test
// server running h.
func newMockExchange(t *testing.T, h http.Handler) *Exchange {
	t.Helper()
	e := new(Exchange)
	require.NoError(t, testexch.Setup(e), "Test instance Setup must not error")
	srv := httptest.NewTestServer(t, h)
	b := e.GetBase()
	b.SkipAuthCheck = true
	require.NoError(t, e.SetHTTPClient(srv.Client()), "SetHTTPClient must not error")
	for k := range b.API.Endpoints.GetURLMap() {
		require.NoErrorf(t, b.API.Endpoints.SetRunningURL(k, srv.URL+"/"), "Setup must point endpoint %s at the mock server", k)
	}
	return e
}
```

Each test then opens with `e := newMockExchange(t, http.HandlerFunc(...))`, and a package-level `writeOKXData(t, w, data)` replaces the two closures. That takes 38 lines out of the file, and the six tests pass unchanged under `-race -count 2`.

thrasher- (changes requested):
Heimdall: Requesting changes. Four of the six open threads are fixed and resolved, each confirmed by running it: the REST fallback carries no `instIdCode` in either listing order, all four call sites use the websocket only with a cached code, and the gateio test and test servers are as suggested. You also fixed all three items I had listed as out of scope, each pinned by a test that fails when the fix is reverted.

**To unblock:**
1. [okx_wrapper_test.go:280](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4088533256) fails `miscellaneous checks`; deleting the line is the fix.
2. [okx_wrapper.go:1400](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4088533270): with an authenticated websocket, OKX rejects every spread cancel in the new branch (`60028`, wrong URL).
3. [Record the websocket results of a partially successful batch](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4080081079) in `CancelAllOrders`.
4. [The spread branch of `CancelBatchOrders`](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4068491795): executed cancels dropped on failure, the client order ID key, and the untested guard.

Inline, not blocking: Nits on [the spread mass-cancel status key](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4088533283), [the algo test's comment](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4088533288) and [the repeated mock setup](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4088533294).

Outside this diff, optional here or as a separate issue. With an authenticated websocket, OKX rejects these with the same `60028`:
- spread `SubmitOrder`, which sends `sprd-order` on the private websocket: [okx_wrapper.go:955](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L955)
- spread `ModifyOrder`, which sends `sprd-amend-order`: [okx_wrapper.go:1179](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1179)
- spread `CancelOrder`, which sends `sprd-cancel-order`: [okx_wrapper.go:1293](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1293)

The other four failing jobs are ambient, and the same jobs [passed on `cfeb17a`](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35820163231): three hit Deribit's live `instrument is not open` in `TestAllExchangeWrappers`, and Windows hit a sqlite temp-file lock in `engine`'s `TestConvertTradesToCandles`.

Happy to approve once those four are in.

exchanges/okx/okx_wrapper.go
thrasher-: Heimdall: **Partly fixed, still open.** The loop now reaches every batch and joins the errors, and `TestCancelAllOrdersContinuesAfterFailedBatch` fails if the `continue` goes back to a `return`. That settles REST, which never returns per-order results with an error.

The websocket half remains. `WSCancelMultipleOrders` returns the decoded per-order results together with `errPartialSuccess` ([ws_requests.go:315-328](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/ws_requests.go#L315-L328)), and the `continue` at [okx_wrapper.go:1604](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1604) skips past them. 40 orders over the websocket, with the first batch answered `code:"2"`, 19 cancelled and 1 `51400`:

```
a0f68e2:   2 batches sent, 39 cancelled by OKX, 20 statuses, ORD-00 to ORD-18 unreported
fix below: 2 batches sent, 39 cancelled by OKX, 40 statuses: 39 Cancelled, ORD-19 with its sMsg
```

**Fix:** record the rows when the error is a partial success, and only then:

```go
		// A failed batch does not stop later batches; the errors are joined so
		// a cancel-all still reaches every remaining order.
		if !cancelResultsUsable(err) {
			continue
		}
		for y := range response {
			if response[y] == nil || response[y].OrderID == "" {
				continue
			}
			if response[y].StatusCode == 0 {
```

```go
// cancelResultsUsable reports whether per-order cancel results can be recorded
// alongside err. A websocket partial success returns fully decoded results with
// its error; any other error, a decode error included, can leave them half
// populated.
func cancelResultsUsable(err error) bool {
	return err == nil || errors.Is(err, errPartialSuccess)
}
```

The gate is what keeps this safe. The websocket decoder fills rows before it fails, so recording on any error turns a reply with a non-numeric `sCode` into `{"ORD-00":"CANCELLED"}` next to a decode error. The helper records nothing there, as now, and nothing for a `[{}]` or code-1 reply either. Like the success path today, it trusts each row's `sCode`, which OKX documents on every row. `common.AppendError` ignores a nil error, so the success path is unchanged.

Smaller, at [okx_wrapper.go:1581-1585](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1581-L1585): when the context dies mid-request, the batch error already carries it, so the joined error reads `… context canceled, authenticated request failed, context canceled`. Renaming the shadowed `err` to `ctxErr` and appending only `if !errors.Is(errs, ctxErr)` keeps one. `errors.Is(err, context.Canceled)` still holds whether the context dies before, during or between batches.

exchanges/okx/okx_wrapper.go
thrasher-: Heimdall: **Partly fixed, still open.** Spread cancels now wait for the whole batch to validate, and `TestCancelBatchOrdersSpreadGuards` fails when the cancel moves back inside the loop. A client-order-ID-only cancel also no longer reports under `""`. Three things remain.

**1. Executed cancels are dropped on failure.** [okx_wrapper.go:1406](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1406) and [okx_wrapper.go:1422](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1422) return `nil` after earlier spread cancels have gone through, while the algo branch at [okx_wrapper.go:1436](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1436) now returns `resp, err`:

```
[spread S1, spot O1], batch cancel fails:  sent sprd/cancel-order, trade/cancel-batch-orders  resp=<nil>
[spread S1, spread S2], S2 fails:          sent sprd/cancel-order twice                     resp=<nil>
fix below, both cases:                     resp=&{Status:map[S1:CANCELLED]}, same error
```

**2. The client order ID key collides**, as my last reply cautioned. The ordinary path keys this map by exchange `ordId`, so a spread `clOrdId` can land on one: `[spread ClientOrderID "123", spot OrderID "123"]`, both cancelled, gives `map[123:CANCELLED]`, one entry for two orders. OKX returns the spread order's `ordId` in the cancel reply, so key by that; with the fix below, the same batch gives `map[123:CANCELLED 2341161427393388544:CANCELLED]`.

**3. The guard at [okx_wrapper.go:1377](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1377) is still untested**, and the two lines I suggested for it last time were wrong. They run ahead of the credentials skip, so anyone running the suite with keys and `canManipulateRealOrders` off would send a real cancel. With keys set:

```
canManipulateRealOrders=false requests=1 sent=/trade/cancel-batch-orders [{"instId":"BTC-USDT","ordId":"","clOrdId":"my-order"}]
```

Disregard them; the mocked subtest below covers it safely.

**Fix** for 1 and 2. The loop cancels over REST only, for the reason in the new comment on [okx_wrapper.go:1400](https://github.com/thrasher-corp/gocryptotrader/blob/a0f68e2d7f084e1660f194ef0b12d41ab8082940/exchanges/okx/okx_wrapper.go#L1400):

```go
	// Cancels from here on execute, so a failure returns resp with the
	// statuses already recorded rather than discarding them.
	for x := range cancelSpreadOrderParams {
		ord := cancelSpreadOrderParams[x]
		// OKX accepts spread operations only on its business websocket, and
		// WSCancelSpreadOrder sends on the private one, so spread cancels use REST.
		var cancelled *SpreadOrderResponse
		cancelled, err = e.CancelSpreadOrder(ctx, ord.OrderID, ord.ClientOrderID)
		switch {
		case err != nil:
			return resp, err
		case cancelled == nil:
			return resp, fmt.Errorf("%w cancelling spread order ID %q client order ID %q", common.ErrNoResponse, ord.OrderID, ord.ClientOrderID)
		case cancelled.StatusCode != 0:
			return resp, getStatusError(cancelled.StatusCode, cancelled.StatusMessage)
		case cancelled.OrderID == "":
			return resp, fmt.Errorf("%w: no order ID cancelling spread order ID %q client order ID %q", common.ErrInvalidResponse, ord.OrderID, ord.ClientOrderID)
		}
		// Status keys are exchange order IDs, as on the ordinary path below, so
		// each cancel is keyed by the ordId OKX returns, even one sent by client
		// order ID.
		resp.Status[cancelled.OrderID] = order.Cancelled.String()
	}
```

REST decodes `data:null` and `[null]` to a nil reply with no error, and `[{}]` to one without an `ordId`. The switch turns each into an error rather than a panic or an empty key. In the ordinary batch below, record `canceledOrders` under the `cancelResultsUsable` gate from the `CancelAllOrders` thread, skipping nil and empty-`ordId` rows, then `if err != nil { return resp, err }`.

For 3, in `TestCancelBatchOrdersSpreadGuards`: have the `/sprd/cancel-order` mock return an `ordId` (expect the status under it), answer `/trade/cancel-batch-orders` too, and add:

```go
	t.Run("client order ID alone passes the ordinary order guard", func(t *testing.T) {
		resp, err := e.CancelBatchOrders(t.Context(), []order.Cancel{{
			AssetType:     asset.Spot,
			Pair:          mainPair,
			ClientOrderID: "spot-cl",
		}})
		require.NoError(t, err, "CancelBatchOrders must accept a client-ID-only spot cancel")
		assert.Equal(t, map[string]string{"2341161427393388545": order.Cancelled.String()}, resp.Status,
			"a client-ID-only spot cancel should report the status under the order ID OKX returns")
	})
```

with the mock answering `{"code":"0","msg":"","data":[{"sCode":"0","sMsg":"","ordId":"2341161427393388545","clOrdId":"spot-cl"}]}`. It fails with the guard reverted to `ClientID`, and sends nothing past the mock.

exchanges/okx/okx_wrapper_test.go
```diff
@@ -17,10 +58,621 @@ func TestMessageID(t *testing.T) {
...
+		resp, err := e.CancelBatchOrders(t.Context(), []order.Cancel{
+			{AssetType: asset.Spread, Pair: spreadPair, ClientOrderID: "spread-cl"},
+			{AssetType: asset.Spot, Pair: mainPair, OrderID: "spot-order"},
+		})
+		assert.Error(t, err, "the failed ordinary batch should be reported")
+		assert.Equal(t, 1, spreadCancelsSent(), "the spread cancel before the failed batch should have been sent")
+		assert.Equal(t, map[string]string{spreadOrdID: order.Cancelled.String()}, resp.Status,
+			"a failed ordinary batch should not discard the spread statuses already recorded")
+	})
+}
```
thrasher-: Heimdall: **Minor** — three new sentinel returns have no `ErrorIs` coverage, and deleting any of their guards leaves the suite green.

**Fix:**

```suggestion
}

// TestCancelBatchOrdersSpreadReplies guards spread cancel replies that confirm
// no order: each fails the batch, with a sentinel error when the reply names no
// order, and keeps the status of the spread order cancelled before it.
func TestCancelBatchOrdersSpreadReplies(t *testing.T) {
	t.Parallel()
	for _, tc := range []struct {
		name  string
		reply string
		err   error
	}{
		{"no data", `{"code":"0","msg":"","data":null}`, common.ErrNoResponse},
		{"no order ID", `{"code":"0","msg":"","data":[{"ordId":"","clOrdId":"S2","sCode":"0","sMsg":""}]}`, common.ErrInvalidResponse},
		{"cancel failed", `{"code":"1","msg":"All operations failed","data":[{"ordId":"","clOrdId":"S2","sCode":"51400","sMsg":"Order does not exist"}]}`, nil},
	} {
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			var mu sync.Mutex
			var cancels int
			e := newMockExchange(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
				if r.URL.Path != "/sprd/cancel-order" {
					t.Errorf("unexpected request path %s", r.URL.Path)
					http.NotFound(w, r)
					return
				}
				mu.Lock()
				cancels++
				first := cancels == 1
				mu.Unlock()
				w.Header().Set("Content-Type", "application/json")
				if first {
					_, _ = w.Write([]byte(`{"code":"0","msg":"","data":[{"ordId":"SPRD-1","clOrdId":"S1","sCode":"0","sMsg":""}]}`))
					return
				}
				_, _ = w.Write([]byte(tc.reply))
			}))
			resp, err := e.CancelBatchOrders(t.Context(), []order.Cancel{
				{AssetType: asset.Spread, Pair: spreadPair, ClientOrderID: "S1"},
				{AssetType: asset.Spread, Pair: spreadPair, ClientOrderID: "S2"},
			})
			if tc.err != nil {
				require.ErrorIs(t, err, tc.err, "CancelBatchOrders must return the sentinel for a reply that names no order")
			} else {
				require.Error(t, err, "CancelBatchOrders must return the failed cancel's error")
			}
			require.NotNil(t, resp, "CancelBatchOrders must return the statuses already recorded")
			assert.Equal(t, map[string]string{"SPRD-1": order.Cancelled.String()}, resp.Status,
				"the spread order cancelled before the failed reply should still be reported")
		})
	}
}

// TestCancelOrderAlgoEmptyReply guards the algo cancel reply that names no
// order, which would otherwise index an empty result.
func TestCancelOrderAlgoEmptyReply(t *testing.T) {
	t.Parallel()
	e := newMockExchange(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		if r.URL.Path != "/trade/cancel-advance-algos" {
			t.Errorf("unexpected request path %s", r.URL.Path)
			http.NotFound(w, r)
			return
		}
		writeOKXData(t, w, []any{})
	}))
	err := e.CancelOrder(t.Context(), &order.Cancel{AssetType: asset.Spot, Pair: mainPair, OrderID: "ALGO-1", Type: order.Trigger})
	assert.ErrorIs(t, err, common.ErrNoResponse, "an algo cancel reply without results should return ErrNoResponse")
}
```

**Why:** the [coding guidelines](https://github.com/thrasher-corp/gocryptotrader/blob/28c970e823e128031adbebfa857d55b458c1bf78/docs/CODING_GUIDELINES.md?plain=1#L206-L208) ask for direct `ErrorIs` coverage of every sentinel return, wrapped or not. [`okx_wrapper.go:1410`](https://github.com/thrasher-corp/gocryptotrader/blob/28c970e823e128031adbebfa857d55b458c1bf78/exchanges/okx/okx_wrapper.go#L1410) and [`okx_wrapper.go:1414`](https://github.com/thrasher-corp/gocryptotrader/blob/28c970e823e128031adbebfa857d55b458c1bf78/exchanges/okx/okx_wrapper.go#L1414) came from the switch I suggested last round without tests, so that gap is mine. [`okx_wrapper.go:1332`](https://github.com/thrasher-corp/gocryptotrader/blob/28c970e823e128031adbebfa857d55b458c1bf78/exchanges/okx/okx_wrapper.go#L1332) came with the algo array decode last round. All three have zero coverage today. Each of these passes the current suite and fails the added tests:

```
guard at :1410 deleted, data:null reply        nil pointer dereference
guard at :1414 deleted, reply with no ordId    {"": "CANCELLED", "SPRD-1": "CANCELLED"}, no error
guard at :1332 deleted, empty algo reply       index out of range [0] with length 0
:1408 back to return nil, err                  the spread order already cancelled goes unreported
```

The last row is last round's `[S1, S2]` case where `S2` fails ([`okx_wrapper.go:1408`](https://github.com/thrasher-corp/gocryptotrader/blob/28c970e823e128031adbebfa857d55b458c1bf78/exchanges/okx/okx_wrapper.go#L1408)); only the ordinary-batch variant is pinned today.

thrasher- (approved):
Heimdall: Approving. All four blockers from last round are fixed, each confirmed by running it at `28c970e`: spread batch cancels go to REST under an authenticated websocket, and a partially successful websocket batch now reports every per-order result, in `CancelAllOrders` (40 of 40 orders, against 20 before) and in `CancelBatchOrders`. The three nits are in too, so I've resolved all seven threads. [One small test comment](https://github.com/thrasher-corp/gocryptotrader/pull/2400#discussion_r4091724189) inline, not blocking. The one red job is Coinbase's live `404` in `TestAllExchangeWrappers`, which [failed the same way on #2397](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35946621984/job/107465886017).

exchanges/okx/ws_requests.go
```diff
@@ -23,12 +23,22 @@ var (
 	privateConnection = "private"
 )
 
+// missingInstrumentIDCode wraps errMissingInstrumentIDCode with the instrument
+// ID so the failure names the instrument lacking a cached code.
+func missingInstrumentIDCode(instID string) error {
+	return fmt.Errorf("%w: %s", errMissingInstrumentIDCode, instID)
+}
+
```
shazbert: we can just `fmt.Errorf("%w: %s", errMissingInstrumentIDCode, instID)` at the call sites without introducing a new func

shazbert (changes requested):
Thanks for the update, this is looking great, one pending test request from heimdall and one nit form me.

PR #2401, opened 2026-09-22
orderbook: add decimal execution calculation

shazbert:
Add a contract-aware Levels.CalculateExecution helper that walks book liquidity using decimal arithmetic and reports requested, executed and remaining amounts.

Expose base and quote totals, VWAP, marginal price and consumed level count. Preserve exact string-backed level values where available, validate malformed inputs and return partial execution details when liquidity is insufficient.

Cover successful, exact-precision, partial-liquidity and validation paths, and include an allocation benchmark.

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


thrasher-:
@codex please review

exchanges/orderbook/calculator.go
```diff
@@ -3,18 +3,40 @@ package orderbook
 import (
 	"errors"
 	"fmt"
+	stdmath "math"
 
+	"github.com/shopspring/decimal"
```
thrasher-: Heimdall: **Blocker** — direct `shopspring/decimal` import. CI lint is red, and under `udecimal_on` the new API cannot accept the repo's own decimal type.

**Fix:** import `github.com/thrasher-corp/gocryptotrader/types/decimal` here and in [calculator_test.go:8](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator_test.go#L8). Three symbols don't port straight across: the façade exports no `Min` (use `result.RemainingAmount.LessThan(levelAmount)`), `NewFromFloat` is `MustFromFloat` and **panics** on NaN/Inf — so the guard at [calculator.go:371](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator.go#L371) stops being belt-and-braces and becomes load-bearing — and `RequireFromString` is `MustFromString`.

**Why:** attributable to this PR, not ambient — `golangci-lint run ./exchanges/orderbook/...`:

```
branch 6bd4b5a:
  calculator.go:8:2: import 'github.com/shopspring/decimal' is not allowed from
    list 'prevent_decimal_package_use' (depguard)
  calculator_test.go:8:2: ... (depguard)
  2 issues:  * depguard: 2
base   158e9c1:
  0 issues.
```

It is not only lint. Under `udecimal_on` the façade's `Decimal` is a distinct struct rather than an alias, so the repo's own contract multiplier cannot be handed to this method:

```go
func Walk(l orderbook.Levels, p *futures.PositionSummary, amount decimal.Decimal) (orderbook.ExecutionCalculation, error) {
	return l.CalculateExecution(amount, p.ContractMultiplier)
}
```

```
go build ./...                     # ok
go build -tags udecimal_on ./...
  cannot use amount (variable of struct type ".../types/decimal".Decimal) as
    "github.com/shopspring/decimal".Decimal value in argument to l.CalculateExecution
  cannot use p.ContractMultiplier (variable of struct type ".../types/decimal".Decimal) as
    "github.com/shopspring/decimal".Decimal value in argument to l.CalculateExecution
```

`futures.PositionSummary.ContractMultiplier` ([futures_types.go:379](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/futures/futures_types.go#L379)) is exactly the value this parameter exists to take.

exchanges/orderbook/calculator_test.go
```diff
@@ -490,3 +491,121 @@ func TestFindNominalAmount(t *testing.T) {
...
+		StrAmount: "0.100000000000000001",
+		Price:     0.2,
+		StrPrice:  "0.200000000000000003",
+	}}
+	amount := decimal.RequireFromString("0.100000000000000001")
+
+	result, err := levels.CalculateExecution(amount, decimal.NewFromInt(1))
+	require.NoError(t, err, "CalculateExecution must use valid exact level values")
+	assert.True(t, result.QuoteAmount.Equal(decimal.RequireFromString("0.020000000000000000500000000000000003")),
+		"quote amount should retain exact level precision")
```
thrasher-, line 579: Heimdall: **Should fix** — this expected value is shopspring-specific and takes down the `udecimal=true` CI job once the import moves to the façade.

**Fix:** assert to a precision both backends produce, or derive the expectation from the backend.

**Why:** I did the port — both imports swapped, `MustFromFloat`/`MustFromString` in place, a local two-argument min helper — and ran it:

```
go test ./exchanges/orderbook/...                   PASS
go test -tags udecimal_on ./exchanges/orderbook/...
  --- FAIL: TestLevelsCalculateExecutionExactValues
  panic: decimal precision out of range:
    "0.020000000000000000500000000000000003" requires 36 fractional digits
```

It panics building the expected value, before `Equal` is reached — udecimal's parser rejects the literal outright rather than returning a different number. The product it actually computes is `0.0200000000000000005`. Division differs too (`5/3` is `1.6666666666666667` on shopspring, `1.6666666666666666666` on udecimal), so any exactness assertion on `AveragePrice` needs the same treatment.

exchanges/orderbook/calculator_test.go
```diff
@@ -490,3 +491,121 @@ func TestFindNominalAmount(t *testing.T) {
...
+	require.ErrorIs(t, err, ErrInsufficientLiquidity, "CalculateExecution must report insufficient liquidity")
+	assert.True(t, result.ExecutedAmount.Equal(decimal.NewFromInt(5)), "executed amount should describe the partial fill")
+	assert.True(t, result.RemainingAmount.Equal(decimal.NewFromInt(1)), "remaining amount should describe the shortfall")
+	assert.True(t, result.QuoteAmount.Equal(decimal.NewFromInt(530)), "quote amount should describe consumed liquidity")
+	assert.True(t, result.AveragePrice.Equal(decimal.NewFromInt(106)), "average price should describe consumed liquidity")
+	assert.True(t, result.MarginalPrice.Equal(decimal.NewFromInt(110)), "marginal price should describe the final consumed level")
+	assert.Equal(t, 2, result.LevelsUsed, "levels used should describe consumed liquidity")
+}
+
+func TestLevelsCalculateExecutionValidation(t *testing.T) {
```
thrasher-, line 645: Heimdall: **Should fix** — the validation matrix tests zero but never a negative or a non-finite, and the non-finite branch is the only uncovered statement in the new code.

**Fix:** add NaN, `+Inf` and negative rows for the request amount, the multiplier, the level amount and the level price.

**Why, non-finite:** `decimal.NewFromFloat(math.NaN())` panics with `Cannot create a Decimal from NaN`, so [calculator.go:374](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator.go#L374) is only safe because of the guard at [calculator.go:371](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator.go#L371) — and that guard is never executed:

```
levelDecimal   80.0%
profile:  calculator.go:372.3,373.1  1  0     <- return decimal.Zero, ErrOrderbookInvalid
```

Nothing upstream excludes it. `checkAlignment` rejects `Amount <= 0` and `Price == 0`, but `NaN <= 0` and `NaN == 0` are both false, so `checkAlignment` returns `<nil>` for a NaN amount *and* a NaN price — and that validation is opt-in anyway ([validate.go:35](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/validate.go#L35)). It matters more after the façade port, where `MustFromFloat` panics on the same input on both backends.

**Why, negatives:** swapping all four `!IsPositive()` guards to `IsZero()` leaves every test in the package passing, while the behaviour goes to pieces:

```
                        HEAD                                    guards weakened to IsZero()
negative orderAmount    execution amount must be greater...     err=<nil>  executed=-1
negative multiplier     contract multiplier must be greater...  err=<nil>  base=-1  quote=-100
negative level amount   orderbook data integrity compromised    remaining=3 for a requested 1
negative level price    orderbook data integrity compromised    err=<nil>  quote=-100
```

HEAD is right; nothing currently holds it there.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +308,68 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+	return result, fmt.Errorf("%w: requested amount %s, remaining amount %s, multiplier %s", ErrInsufficientLiquidity, orderAmount, result.RemainingAmount, multiplier)
+}
+
+func (e *ExecutionCalculation) setAveragePrice() {
+	if e.LevelsUsed == 1 {
+		e.AveragePrice = e.MarginalPrice
+		return
+	}
+	if e.BaseAmount.IsPositive() {
+		e.AveragePrice = e.QuoteAmount.Div(e.BaseAmount)
```
thrasher-: Heimdall: **Minor** — `AveragePrice` is exact when one level is consumed and rounded to 16 decimal places when two or more are, and neither the field nor the shortcut says so.

**Fix:** a line on the shortcut explaining it preserves exactness, and either `DivRound` with an explicit precision or a note on the field that `AveragePrice` is rounded. `shopspring`'s package-level `DivisionPrecision` is 16 and nothing in the repo changes it, so the value is whatever that global happens to be.

**Why:** the two paths disagree from 17 decimal places of price precision onward — same liquidity, same `QuoteAmount`, same `BaseAmount`, different answer:

```
dp   level price               1 level      2 levels
16   0.1000000000000003        exact        0.1000000000000003   exact
17   0.10000000000000003       exact        0.1
18   0.100000000000000003      exact        0.1
```

To be clear on the practical weight: I swept 100,000 uniform-price books with prices up to 16 decimal places and **none** diverged, and the long prices in this repo's exchange fixtures are trailing zeros rather than 17 significant places. So this is about the contract being stated, not about a number that is wrong today.

The test is the part I would change regardless. [calculator_test.go:530](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator_test.go#L530) reads as proof that `AveragePrice` retains exact level precision, but with one level `setAveragePrice` returns `MarginalPrice` and never divides — delete the `LevelsUsed == 1` shortcut and that assertion is the only failure in the package. It is a regression test for the shortcut, not for the average. A second case with the same price spread over two levels would pin what the message claims.

exchanges/orderbook/calculator.go
```diff
@@ -3,18 +3,40 @@ package orderbook
...
+	// ErrInsufficientLiquidity is returned when the levels cannot fill the requested amount.
+	ErrInsufficientLiquidity = errors.New("insufficient liquidity")
+	errPriceTargetInvalid    = errors.New("price target is invalid")
+	errCannotShiftPrice      = errors.New("cannot shift price")
 )
 
+// ExecutionCalculation contains the result of walking an orderbook side for a
+// requested order amount. Order amounts are expressed in the same units as the
+// level amounts; base amounts include the supplied contract multiplier.
+type ExecutionCalculation struct {
```
thrasher-, line 25: Heimdall: **Minor** — a fill that drains the entire side is indistinguishable from one that had room to spare, and `MarginalPrice` answers the boundary question differently from the walkers directly above it.

**Fix:** carry a `FullLiquidityUsed` alongside `LevelsUsed` — `DeploymentAction` ([calculator.go:201](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator.go#L201)) already has one for exactly this and `WhaleBomb`/`SimulateOrder` surface it as `fullLiquidityUsageWarning` — and document `MarginalPrice` on the field.

**Why:** an order that takes the whole side returns clean, with nothing to distinguish it:

```
Levels{{Price: 100, Amount: 2}, {Price: 110, Amount: 3}}

b.buy(530)              -> LevelPositionPrice 110  FullLiquidityUsed true
CalculateExecution(5,1) -> MarginalPrice 110  LevelsUsed 2  RemainingAmount 0  err nil
```

The naming difference shows at an exact level boundary. Consuming precisely level one:

```
b.buy(200)              -> LevelPositionPrice 110   the next level, where the book now sits
CalculateExecution(2,1) -> MarginalPrice      100   the level just consumed
```

Both readings are defensible — "marginal" usually means the price the next unit would pay, which is `110` — but two neighbouring functions answering the same question differently wants a sentence saying which one this is. Nothing currently pins it either: a mutation that reports the next level's price at an exact boundary passes all four new tests, because the success fixture stops partway through its second level and never lands on a boundary.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +308,68 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+	}
+
+	result := ExecutionCalculation{
+		RequestedAmount: orderAmount,
+		RemainingAmount: orderAmount,
+	}
+	for i := range l {
+		levelAmount, err := levelDecimal(l[i].Amount, l[i].StrAmount)
+		if err != nil || !levelAmount.IsPositive() {
+			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid amount %q", ErrOrderbookInvalid, i, l[i].StrAmount)
```
thrasher-: Heimdall: **Minor** — the message quotes `StrAmount`, which is empty whenever the failure came from the float, so every float-side failure renders identically.

**Fix:** report the float when the string is empty, and wrap the `decimal.NewFromString` failure instead of dropping it — `levelDecimal`'s error is swallowed by the `err != nil || !levelAmount.IsPositive()` test, so a parse failure loses its reason. Same at [calculator.go:335](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator.go#L335) for price.

**Why:** three different defects, one message:

```
Level{Amount: math.NaN(),  Price: 100}  ->  orderbook data integrity compromised: level 0 has invalid amount ""
Level{Amount: math.Inf(1), Price: 100}  ->  orderbook data integrity compromised: level 0 has invalid amount ""
Level{Amount: -1,          Price: 100}  ->  orderbook data integrity compromised: level 0 has invalid amount ""
```

exchanges/orderbook/calculator.go
```diff
@@ -3,18 +3,40 @@ package orderbook
...
 
 var (
-	errPriceTargetInvalid = errors.New("price target is invalid")
-	errCannotShiftPrice   = errors.New("cannot shift price")
+	// ErrInvalidExecutionAmount is returned when the requested execution amount is not positive.
+	ErrInvalidExecutionAmount = errors.New("execution amount must be greater than zero")
+	// ErrInvalidContractMultiplier is returned when the base amount multiplier is not positive.
+	ErrInvalidContractMultiplier = errors.New("contract multiplier must be greater than zero")
+	// ErrInsufficientLiquidity is returned when the levels cannot fill the requested amount.
+	ErrInsufficientLiquidity = errors.New("insufficient liquidity")
```
thrasher-: Heimdall: **Minor** — the package already carries this vocabulary, and it keeps exported errors in their own block.

**Fix:** move the three new exported errors into a `// Public errors` block — the package separates them in [orderbook_types.go:23](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/orderbook_types.go#L23) and [depth.go:19](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/depth.go#L19) — rather than into calculator.go's unexported one. Worth reusing what is there while you are at it: `errNotEnoughLiquidity` ([orderbook_types.go:39](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/orderbook_types.go#L39)) and `errNoLiquidity` ([levels.go:26](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/levels.go#L26)) already cover `ErrInsufficientLiquidity`, and `errAmountInvalid` ([orderbook_types.go:33](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/orderbook_types.go#L33)) covers `ErrInvalidExecutionAmount`.

**Why:** that makes three spellings of the same condition, and an empty side now answers with a fourth:

```
Levels{}.CalculateExecution(1, 1)     -> insufficient liquidity: requested amount 1, remaining amount 1, multiplier 1
(&Book{}).GetAveragePrice(true, 1)    -> not enough liquidity for  on exchange  to support a buy amount of 1
b.buy(1) / b.sell(1) on an empty side -> no liquidity
```

Exporting them is a separate call — it does give callers something to match on, which the unexported ones don't. If that is the intent, worth saying so, because it is the argument for keeping all three.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +308,68 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
 	}
 	return aggNominalAmount, remainingAmount
 }
+
+// CalculateExecution walks the levels required to execute orderAmount. The
```
thrasher-, line 317: Heimdall: **Minor** — the multiplier is linear-contract only; the doc comment does not say so.

**Fix:** state that level amounts convert to base by multiplication, i.e. linear / USDT-margined contracts, and that inverse and quanto are out of scope.

**Why:** the repo models the alternatives — `futures.Inverse` and `futures.Quanto` ([contract.go:50](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/futures/contract.go#L50)) and `asset.IsCoinMargined()` ([asset.go:232](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/asset/asset.go#L232)). A coin-margined contract's size is denominated in quote, so contracts → base is `size / price`, not `size × multiplier`. `QuoteAmount = used × multiplier × levelPrice` at [calculator.go:343](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator.go#L343) then returns a wrong number rather than an error, and the GoDoc's "such as a derivative contract" reads as though it covers them.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +308,68 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+		result.BaseAmount = result.BaseAmount.Add(baseAmount)
+		result.QuoteAmount = result.QuoteAmount.Add(baseAmount.Mul(levelPrice))
+		result.MarginalPrice = levelPrice
+		result.LevelsUsed++
+		if result.RemainingAmount.IsZero() {
+			result.setAveragePrice()
+			return result, nil
+		}
+	}
+	if result.BaseAmount.IsPositive() {
```
thrasher-: Heimdall: **Nit** — redundant. `setAveragePrice` already guards `BaseAmount.IsPositive()` internally, and the `LevelsUsed == 1` path cannot reach a non-positive base because `used` and `multiplier` are both positive by then.

**Fix:** call `result.setAveragePrice()` unconditionally.

**Why:** removing the check leaves the package green, empty levels included — `AveragePrice` stays `0` and no division happens.

exchanges/orderbook/calculator.go
```diff
@@ -3,18 +3,40 @@ package orderbook
...
+// level amounts; base amounts include the supplied contract multiplier.
+type ExecutionCalculation struct {
+	RequestedAmount decimal.Decimal
+	ExecutedAmount  decimal.Decimal
+	RemainingAmount decimal.Decimal
+	BaseAmount      decimal.Decimal
+	QuoteAmount     decimal.Decimal
+	AveragePrice    decimal.Decimal
+	MarginalPrice   decimal.Decimal
+	LevelsUsed      int
```
thrasher-: Heimdall: **Nit** — `LevelsUsed` is a count that cannot go negative; the coding guidelines default struct integers to `uint64` and discourage `int`, whose width varies by architecture.

The type comment already carries the order-unit/base-unit distinction, so this is genuinely optional: `AveragePrice` being the VWAP and `MarginalPrice` being the last consumed level's price are stated only in the test messages, and the package is split on whether that warrants a field comment — `Movement` ([orderbook_types.go:155](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/orderbook_types.go#L155)) documents its fields, `DeploymentAction` and `WhaleBombResult` in this file do not.

thrasher- (changes requested):
Heimdall: Changes requested — one blocker, the rest suggestions.

The arithmetic holds: across 200,000 randomised books `ExecutedAmount + RemainingAmount == RequestedAmount` never broke and `BaseAmount == ExecutedAmount × multiplier` was exact every time.

**To unblock:** [calculator.go:8](https://github.com/thrasher-corp/gocryptotrader/blob/6bd4b5a92b1b39f7da8ae1b96259135a9fdeaee2/exchanges/orderbook/calculator.go#L8) — the direct `shopspring/decimal` import fails `depguard`, so lint is red here and clean at the base. It also puts the new API out of reach of `types/decimal.Decimal` under `udecimal_on`.

Nine non-gating comments carry their evidence on the lines. Happy to approve once the import moves to the façade.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +311,89 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+		}
+		baseAmount := used.Mul(multiplier)
+		result.ExecutedAmount = result.ExecutedAmount.Add(used)
+		result.RemainingAmount = result.RemainingAmount.Sub(used)
+		result.BaseAmount = result.BaseAmount.Add(baseAmount)
+		result.QuoteAmount = result.QuoteAmount.Add(baseAmount.Mul(levelPrice))
+		result.MarginalPrice = levelPrice
+		result.LevelsUsed++
+		if result.RemainingAmount.IsZero() {
+			result.FullLiquidityUsed = i == len(l)-1
```
thrasher-: Heimdall: **Blocker** — this tests whether the fill *ended on* the last level, not whether the book was exhausted, so `FullLiquidityUsed` is true for any order that stops partway through the final level.

**Fix:** require the final level to have been fully consumed:

```go
result.FullLiquidityUsed = i == len(l)-1 && used.Equal(levelAmount)
```

**Why:** on `Levels{{Price: 100, Amount: 2}, {Price: 110, Amount: 3}, {Price: 120, Amount: 5}}` — ten units in total:

```
request  LevelsUsed  FullLiquidityUsed   units actually left
2        1           false               8
4        2           false               6
5        2           false               5
6        3           TRUE                4
7        3           TRUE                3
9        3           TRUE                1
10       3           true                0
```

Three of those are wrong, and the field comment at [calculator.go:39](https://github.com/thrasher-corp/gocryptotrader/blob/45080d015ede592b2909bca1d7275cb8d09646e2/exchanges/orderbook/calculator.go#L39) states the contract the code doesn't keep — "reports that every level on this side was consumed". The walker directly above gets it right: `(&Book{Bids: levels}).sell(6)` returns `FullLiquidityUsed=false`, `sell(10)` returns `true`.

With the one-line change above, all seven requests agree with `sell()` exactly, and every test in the package still passes — which is the other half of the problem. `TestLevelsCalculateExecution` exercises requests of 2, 4 and 10; none stops partway through the final level, so nothing pins this either way. A request of 6 against that same fixture is the case worth adding.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +311,89 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+// of one when level amounts are already in base units.
+func (l Levels) CalculateExecution(orderAmount, multiplier decimal.Decimal) (ExecutionCalculation, error) {
+	if !orderAmount.IsPositive() {
+		return ExecutionCalculation{}, fmt.Errorf("%w: %s", errAmountInvalid, orderAmount)
+	}
+	if !multiplier.IsPositive() {
+		return ExecutionCalculation{}, fmt.Errorf("%w: %s", ErrInvalidContractMultiplier, multiplier)
+	}
+	if len(l) == 0 {
+		return ExecutionCalculation{}, errNoLiquidity
```
thrasher-, line 332: Heimdall: **Minor** — moving to the package's existing sentinels means an external caller can no longer detect the partial-fill case, which is the feature's headline behaviour.

**Fix:** export the not-enough-liquidity sentinel, or keep a distinct exported one for this path.

**Why:** I raised the reuse-versus-export trade-off last round and flagged that exporting was the separate call; this is just what it costs, so if it is intended, say so and I will drop it. From a package that imports `orderbook`:

```
bad multiplier        errors.Is(ErrInvalidContractMultiplier) = true
bad level             errors.Is(ErrOrderbookInvalid)          = true
bad amount            no exported sentinel matches
no liquidity          no exported sentinel matches
not enough liquidity  no exported sentinel matches
```

`CalculateExecution` deliberately returns populated results alongside the not-enough-liquidity error, so a caller is expected to act on that case — resize the order, route elsewhere — but reaching it now means matching on the message text. Invalid input, an empty side and a genuine shortfall are all equally unmatchable, and those want different handling.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +311,89 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+		}
+		if !levelPrice.IsPositive() {
+			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid price %q", ErrOrderbookInvalid, i, levelInput(l[i].Price, l[i].StrPrice))
+		}
+
+		used := result.RemainingAmount
+		if levelAmount.LessThan(used) {
+			used = levelAmount
+		}
+		baseAmount := used.Mul(multiplier)
```
thrasher-: Heimdall: **Minor** — scaling each level individually lets the `udecimal` backend truncate every intermediate to zero and report a successful fill with no base, no quote and no price.

**Fix:** apply the multiplier once rather than per level — accumulate `used` and `used × price`, then scale both at the exit. I tried it: both backends then return `base=1e-19 quote=5.5e-17 avg=550` for the case below, and every test in the package still passes under both tags.

**Why:** ten levels of `1e-10` at ascending prices, requesting `1e-9` with a multiplier of `1e-10` — identical inputs, both backends:

```
                BaseAmount    QuoteAmount   AveragePrice   err
shopspring      1e-19         5.5e-17       550            nil
udecimal        0             0             0              nil
```

Both report `ExecutedAmount=1e-9`, `RemainingAmount=0`, `LevelsUsed=10`, `FullLiquidityUsed=true`. Each level's `used.Mul(multiplier)` is `1e-20`, one digit past udecimal's 19-place floor, so it truncates to zero before it is ever added — yet the aggregate `1e-19` is perfectly representable, and a single multiplication at the end produces it on both backends.

Honest about the weight: you need `used × multiplier` below `1e-19` to reach it, and realistic multipliers and amounts don't get close — at a multiplier of `1e-19` the two backends still agree. The part worth fixing is that it fails silently: `err` is nil and `FullLiquidityUsed` is true, so nothing tells a caller the numbers are empty.

exchanges/orderbook/calculator_test.go
```diff
@@ -490,3 +491,220 @@ func TestFindNominalAmount(t *testing.T) {
...
+			multiplier: decimal.NewFromInt(1),
+			expected:   errNoLiquidity,
+		},
+		{
+			name:       "zero level amount",
+			levels:     Levels{{Price: 1}},
+			amount:     decimal.NewFromInt(1),
+			multiplier: decimal.NewFromInt(1),
+			expected:   ErrOrderbookInvalid,
+			contains:   "0",
```
thrasher-: Heimdall: **Nit** — four of the ten `contains` values match the error template rather than the offending input, so that half of those rows holds whatever `levelInput` returns.

**Fix:** assert the rendered field — `amount "0"` rather than `0` — for the two zero rows and the two `invalid exact` rows.

**Why:** `level 0 has invalid amount …` already contains both `0` and `invalid`. Making `levelInput` return `""` unconditionally fails six rows and leaves those four green:

```
FAIL  negative / NaN / infinite level amount
FAIL  negative / NaN / infinite level price
  ok  zero level amount    zero level price
  ok  invalid exact amount  invalid exact price
```

To be fair to the rows: every one of them still runs `assert.ErrorIs`, and that part does its job — swapping the returned sentinel fails all four. It is only the `contains` half that is free. Same gap covers the new parser detail: dropping the appended `: %v` from both error paths leaves all fifteen rows passing.

thrasher- (changes requested):
Heimdall: Changes requested — one new blocker; all ten findings from the last round are fixed and resolved.

Verified rather than read: lint clean, `-tags udecimal_on` green via the two build-tagged expectation files, and `CalculateExecution`, `setAveragePrice`, `levelDecimal` and `levelInput` all at 100% coverage — including the non-finite guard that was the one uncovered statement before.

**To unblock:** [calculator.go:363](https://github.com/thrasher-corp/gocryptotrader/blob/45080d015ede592b2909bca1d7275cb8d09646e2/exchanges/orderbook/calculator.go#L363) — `FullLiquidityUsed` tests whether the fill ended *on* the last level rather than whether the book was drained, so it reports true with liquidity still available. One-line fix in the comment; no existing test pins it either way.

Three non-gating comments alongside it, evidence on the lines. Happy to approve once that boolean is right.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +311,92 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+}
+
+func (e *ExecutionCalculation) setAveragePrice() {
+	if e.LevelsUsed == 1 {
+		// Avoiding division preserves the exact price supplied by the level.
+		e.AveragePrice = e.MarginalPrice
+		return
+	}
+	if e.BaseAmount.IsPositive() {
+		e.AveragePrice = e.QuoteAmount.Div(e.BaseAmount)
```
thrasher-: Heimdall: **Minor** — `AveragePrice` is divided out of the multiplier-scaled totals, so when udecimal truncates either of them the average comes out silently wrong.

**Fix:** a VWAP doesn't depend on the multiplier, so take it from the unscaled accumulator. One helper also replaces the three lines duplicated at [calculator.go:363](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator.go#L363) and [calculator.go:370](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator.go#L370):

```go
// setTotals scales the aggregated amounts once so small per-level values keep
// their precision. The multiplier cancels out of a volume-weighted average, so
// AveragePrice is derived before scaling.
func (e *ExecutionCalculation) setTotals(unscaledQuoteAmount, multiplier decimal.Decimal) {
	e.BaseAmount = e.ExecutedAmount.Mul(multiplier)
	e.QuoteAmount = unscaledQuoteAmount.Mul(multiplier)
	if e.LevelsUsed == 1 {
		// Avoiding division preserves the exact price supplied by the level.
		e.AveragePrice = e.MarginalPrice
		return
	}
	e.AveragePrice = unscaledQuoteAmount.Div(e.ExecutedAmount)
}
```

`ExecutedAmount` is always positive at both exits, since every consumed level adds a positive `used`, so the `IsPositive` guard can go. A second case at the end of `TestLevelsCalculateExecutionAggregatesBeforeScaling` pins it. It fails under `udecimal_on` as the code stands and passes with the helper:

```go
	for i := range levels {
		levels[i].StrAmount = "0.00000000015"
	}
	result, err = levels.CalculateExecution(decimal.MustFromString("0.0000000015"), decimal.MustFromString("0.0000000001"))
	require.NoError(t, err, "CalculateExecution must not error when scaled totals exceed the backend precision")
	assert.True(t, result.AveragePrice.Equal(decimal.NewFromInt(550)), "average price should not depend on the contract multiplier")
```

**Why:** under udecimal, filling each book exactly. Every row returns `err=nil` and `FullLiquidityUsed=true`:

```
levels                                    multiplier     BaseAmount              QuoteAmount             AveragePrice   true VWAP
10 × 0.00000000015 at 100 … 1000          0.0000000001   0.0000000000000000001   0.0000000000000000825   825            550
10 × 0.00000000001 at 100 … 1000          0.0000000001   0                       0.0000000000000000055   0              550
2 × 0.00000001 at 0.00001001, 0.00001002  0.0001         0.000000000002          0.00000000000000002     0.00001        0.000010015
2 × 0.00000001 at 0.00000001, 0.00000002  0.0001         0.000000000002          0                       0              0.000000015
```

The first two rows truncate `BaseAmount`, the last two `QuoteAmount`. Neither truncation is avoidable at 19 places. Dividing one by the other is what turns it into a wrong average. Shopspring returns the true VWAP in every row, and so does udecimal with the helper.

On the weight: it takes dust to move the average materially. The last two rows are 0.00000001 contracts at a real multiplier. At ordinary sizes it stays in the trailing digits. Across 17,356 random books per multiplier, with prices from 1 to 100,000, nothing differs at a multiplier of 1 or above, and at 0.0001 the largest difference is 0.000000000000157. On shopspring the helper changes nothing: 0 of 86,767 books differ, because its `Mul` is exact and scaling both operands leaves its rounded quotient unchanged. It's worth the few lines because the bad case fails silently.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +311,92 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+	}
+}
+
+func levelDecimal(value float64, exact string) (decimal.Decimal, error) {
+	if exact != "" {
+		return decimal.NewFromString(exact)
+	}
+	if stdmath.IsNaN(value) || stdmath.IsInf(value, 0) {
+		return decimal.Zero, ErrOrderbookInvalid
```
thrasher-: Heimdall: **Nit** — a non-finite level reports the sentinel twice.

**Fix:** return a dedicated error here, e.g. `errNonFiniteValue = errors.New("non-finite value")` beside `errCannotShiftPrice`, with an `assert.ErrorIs` on `levelDecimal` for NaN and both infinities to cover it. The call sites at [calculator.go:339](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator.go#L339) and [calculator.go:346](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator.go#L346) already wrap `ErrOrderbookInvalid`.

**Why:** `levelDecimal` returns the bare `ErrOrderbookInvalid`, and the caller appends it with `%v` after wrapping the same sentinel with `%w`:

```
Level{Amount: NaN, Price: 1}   ->  orderbook data integrity compromised: level 0 has invalid amount "NaN": orderbook data integrity compromised
Level{Amount: 1, Price: +Inf}  ->  orderbook data integrity compromised: level 0 has invalid price "+Inf": orderbook data integrity compromised
```

With the change these read `… invalid amount "NaN": non-finite value`, every validation row still matches `ErrOrderbookInvalid`, and the package passes on both backends.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +311,92 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
 	}
 	return aggNominalAmount, remainingAmount
 }
+
+// CalculateExecution walks the levels required to execute orderAmount. The
```
thrasher-, line 317: Heimdall: **Nit** — the partial-result contract, this method's headline behaviour, is written down only on the sentinel, and three nearby descriptions no longer match the code.

**Fix:**

- Here: say that when a non-empty side runs out before `orderAmount` is filled, the method returns the partial execution together with `ErrNotEnoughLiquidity`. Go callers conventionally ignore the result when `err` is non-nil, so this needs saying where they will read it.
- [orderbook_types.go:30](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/orderbook_types.go#L30): `GetAveragePrice` ([calculator.go:293](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator.go#L293)) returns the same value with a zero result and no partial fill, so "could only be partially filled" doesn't describe what its callers get. Word it to cover both. Renaming `errNotEnoughLiquidity` outright would also drop the alias, the only `ErrX = errX` in the repository. Besides the declaration, that touches one production line and two test lines.
- [calculator.go:25](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator.go#L25) reads as though only `BaseAmount` carries the multiplier, but `QuoteAmount` does too. [calculator_test.go:509](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator_test.go#L509) expects `210` at a multiplier of 0.5, where the unscaled walk is `420`.
- [calculator_test.go:560](https://github.com/thrasher-corp/gocryptotrader/blob/61e27798fbec55125868a2a16e863e3fdc2c2806/exchanges/orderbook/calculator_test.go#L560): "single-level execution should report full liquidity use" describes the behaviour last round removed. It holds only because the request is the whole level. A partial single-level fill now correctly reports `false`.

thrasher- (approved):
Heimdall: Approving. All four findings from last round are fixed, verified by execution rather than read. `FullLiquidityUsed` now agrees with `sell()` at all 48 request sizes from 0.25 to 12 on both backends. Reverting either fix fails its new test; the aggregation one fails under `udecimal_on`, which CI runs.

Inline, none blocking: a Minor on [calculator.go:383](https://github.com/thrasher-corp/gocryptotrader/pull/2401#discussion_r4079322135), and Nits on [calculator.go:392](https://github.com/thrasher-corp/gocryptotrader/pull/2401#discussion_r4079322139) and [calculator.go:315](https://github.com/thrasher-corp/gocryptotrader/pull/2401#discussion_r4079322144).

The windows-latest failure, `TestEnableDisableRateLimit` in `exchanges/request`, is ambient: that test binary links neither `exchanges/orderbook` nor `types/decimal`.

thrasher- (approved):
Heimdall: Approving. All three findings from last round are fixed, verified by execution rather than read. The average now comes from the unscaled totals at both exits, and the four cases in [last round's table](https://github.com/thrasher-corp/gocryptotrader/pull/2401#discussion_r4079322135) return the true VWAP under udecimal. On shopspring the change is inert: 20,000 random executions are identical in every field. Reverting the average fails the new tests under `udecimal_on`, and reverting the non-finite error fails `TestLevelDecimalNonFinite` on both backends.

Nothing further from me.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +313,91 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+
+	result := ExecutionCalculation{
+		RequestedAmount: orderAmount,
+		RemainingAmount: orderAmount,
+	}
+	var unscaledQuoteAmount decimal.Decimal
+	for i := range l {
+		levelAmount, err := levelDecimal(l[i].Amount, l[i].StrAmount)
+		if err != nil {
+			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid amount %q: %v", ErrOrderbookInvalid, i, levelInput(l[i].Amount, l[i].StrAmount), err)
```
gloriousCode:
```suggestion
			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid amount %q: %w", ErrOrderbookInvalid, i, levelInput(l[i].Amount, l[i].StrAmount), err)
```

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +313,91 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+		}
+		if !levelAmount.IsPositive() {
+			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid amount %q", ErrOrderbookInvalid, i, levelInput(l[i].Amount, l[i].StrAmount))
+		}
+		levelPrice, err := levelDecimal(l[i].Price, l[i].StrPrice)
+		if err != nil {
+			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid price %q: %v", ErrOrderbookInvalid, i, levelInput(l[i].Price, l[i].StrPrice), err)
```
gloriousCode:
```suggestion
			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid price %q: %w", ErrOrderbookInvalid, i, levelInput(l[i].Price, l[i].StrPrice), err)
```

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +315,160 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+		}
+		if !levelPrice.IsPositive() {
+			return ExecutionCalculation{}, fmt.Errorf("%w: level %d has invalid price %q", ErrOrderbookInvalid, i, levelInput(l[i].Price, l[i].StrPrice))
+		}
+		if filled {
+			continue
+		}
```
thrasher-: Heimdall: **Minor** — checking every level means every call now walks the whole side, so a small order costs O(book depth) rather than O(levels consumed).

**Fix:** return as soon as the order is filled, as the previous revision did. This drops `filled` and this `continue`:

```go
		if result.RemainingAmount.IsZero() {
			result.FullLiquidityUsed = i == len(l)-1 && used.Equal(levelAmount)
			if err := result.setTotals(unscaledQuoteAmount, exactQuoteAmount, multiplier); err != nil {
				return ExecutionCalculation{}, err
			}
			return result, nil
		}
	}
	result.FullLiquidityUsed = true
```

Say so in the doc comment, e.g. "Levels beyond the fill are not read, so the cost follows the levels consumed rather than the depth of the book." Then swap the two "after fill" validation rows for a test that pins it:

```go
func TestLevelsCalculateExecutionStopsAtFill(t *testing.T) {
	t.Parallel()
	levels := Levels{{Price: 100, Amount: 1}, {Price: 0, Amount: 0}}
	result, err := levels.CalculateExecution(decimal.NewFromInt(1), decimal.NewFromInt(1))
	require.NoError(t, err, "CalculateExecution must not read levels beyond the fill")
	assert.Equal(t, uint64(1), result.LevelsUsed, "only the first level should be consumed")
}
```

It fails against this commit on both backends and passes with the change, and the package stays green and lint-clean on both.

If whole-side integrity is the aim, it belongs with the package's opt-in validation. With `VerifyOrderbook` on, snapshots go through [`Book.Validate`](https://github.com/thrasher-corp/gocryptotrader/blob/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc/exchanges/orderbook/validate.go#L31) and [incremental updates](https://github.com/thrasher-corp/gocryptotrader/blob/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc/exchanges/orderbook/incremental_updates.go#L115) through the same checks, once per change, and those also check ordering and duplicates, which this loop does not. They don't parse `StrAmount`/`StrPrice`, so stopping at the fill does give up rejecting a malformed exact string on a level the order never reaches. A parse check would fit there too.

**Why:** every level's amount and price are converted before the `filled` check. A 0.1 order that fills inside the first level:

```
levels  backend               this commit                stop at fill
   100  default            63 µs      420 allocs       1.2 µs   24 allocs
 1,000  default           647 µs    4,020 allocs       1.2 µs   24 allocs
 1,000  default, strings  329 µs    5,865 allocs       1.1 µs   23 allocs
 5,000  default         3,142 µs   20,020 allocs       1.2 µs   24 allocs
 1,000  udecimal          628 µs    6,171 allocs       5.8 µs   53 allocs
```

"strings" means `StrAmount`/`StrPrice` are set. Kraken is the one exchange that sets them ([kraken_websocket.go:532](https://github.com/thrasher-corp/gocryptotrader/blob/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc/exchanges/kraken/kraken_websocket.go#L532)), and its default book subscription is 1,000 levels ([kraken_websocket.go:91](https://github.com/thrasher-corp/gocryptotrader/blob/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc/exchanges/kraken/kraken_websocket.go#L91)).

It also applies the backend's input limits to levels the order never reaches. Under `udecimal_on`, a 0.1 order against `Levels{{Amount: 1, Price: 100}, {Amount: 1e-20, Price: 101}}` now fails with `level 1 has invalid amount "1e-20"`. The default build fills it, and so did the previous revision on both backends.

It's Minor only because nothing calls `CalculateExecution` yet. On a per-update path it would be the dominant cost.

exchanges/orderbook/calculator.go
```diff
@@ -286,3 +315,160 @@ func (l Levels) FindNominalAmount(amount float64) (aggNominalAmount, remainingAm
...
+				return ExecutionCalculation{}, err
+			}
+			priceRat, err := executionDecimalRat(levelPrice)
+			if err != nil {
+				return ExecutionCalculation{}, err
+			}
+			exactQuoteAmount.Add(exactQuoteAmount, new(big.Rat).Mul(usedRat, priceRat))
+		} else {
+			unscaledQuoteAmount = unscaledQuoteAmount.Add(used.Mul(levelPrice))
+		}
```
thrasher-: Heimdall: **Minor** — the `big.Rat` path is exact, but it makes a udecimal walk about four times slower, and eight of its lines can't be reached. The façade's own arithmetic gives identical results.

**Fix:** multiply each consumed amount by 10^19 before multiplying it by the price. udecimal's `Mul` truncates by dividing the product's coefficient by 10^(p+q−19), and after the scale-up that division only removes zeros, so every level's product is exact. Unscaling the totals then truncates once. Keep the scale behind the build tags you already have, so the default build's arithmetic is unchanged:

```go
//go:build udecimal_on

package orderbook

import "github.com/thrasher-corp/gocryptotrader/types/decimal"

// udecimal truncates each product to 19 fractional digits. Scaling an amount by
// 10^19 before multiplying it by a price keeps each level's product exact, so
// later truncation only drops digits the unscaled totals would drop anyway.
var (
	executionScale   = decimal.MustFromString("10000000000000000000")
	executionUnscale = decimal.MustFromString("0.0000000000000000001")
)

func scaleExecutionAmount(amount decimal.Decimal) decimal.Decimal {
	return amount.Mul(executionScale)
}

func unscaleExecutionAmount(amount decimal.Decimal) decimal.Decimal {
	return amount.Mul(executionUnscale)
}
```

In the shopspring file both helpers return `amount`, because its `Mul` is exact. This block becomes one line, with `scaledQuoteAmount` replacing both accumulators:

```go
		scaledQuoteAmount = scaledQuoteAmount.Add(scaleExecutionAmount(used).Mul(levelPrice))
```

and `setTotals` can no longer fail:

```go
func (e *ExecutionCalculation) setTotals(scaledQuoteAmount, multiplier decimal.Decimal) {
	e.BaseAmount = e.ExecutedAmount.Mul(multiplier)
	e.QuoteAmount = unscaleExecutionAmount(scaledQuoteAmount.Mul(multiplier))
	if e.LevelsUsed == 1 {
		// Avoiding division preserves the exact price supplied by the level.
		e.AveragePrice = e.MarginalPrice
		return
	}
	// The multiplier cancels out of VWAP. Scaling the divisor to match the quote
	// applies the backend's division precision to the price itself.
	e.AveragePrice = scaledQuoteAmount.Div(scaleExecutionAmount(e.ExecutedAmount))
}
```

`executionDecimalRat`, `truncatedExecutionDecimal`, `executionFractionalDigits` and the `math/big` and `strings` imports all go.

Two cases at the end of `TestLevelsCalculateExecutionRetainsSubPrecisionLevelValues` would catch the two ways the scaling itself can quietly go wrong. Averaging the truncated quote, or scaling by only 10^18, passes the current suite on both backends. Each fails one of these under `udecimal_on`, and both pass against this commit's code too:

```go
	levels = Levels{
		{StrAmount: "0.0000000000000000001", StrPrice: "0.1"},
		{StrAmount: "0.0000000000000000001", StrPrice: "0.2"},
	}
	result, err = levels.CalculateExecution(decimal.MustFromString("0.0000000000000000002"), decimal.NewFromInt(1))
	require.NoError(t, err, "CalculateExecution must average an aggregate beyond the backend precision")
	assert.Equal(t, "0.15", result.AveragePrice.String(), "average price should use the untruncated aggregate")

	levels = Levels{{StrAmount: "0.0000000000000000001", StrPrice: "0.0000000000000000001"}}
	result, err = levels.CalculateExecution(decimal.MustFromString("0.0000000000000000001"), decimal.MustFromString("10000000000000000000"))
	require.NoError(t, err, "CalculateExecution must apply a large multiplier before truncating the quote amount")
	assert.Equal(t, "0.0000000000000000001", result.QuoteAmount.String(), "quote amount should retain a product below the backend precision")
```

**Why:** against a copy of this commit's implementation, 38,707 random executions give identical results in every field and error on both backends. They include 5,395 partial fills, amounts and prices from dust to 25 integer digits with up to 19 decimal places, and float-backed levels. `TestLevelsCalculateExecutionRetainsSubPrecisionLevelValues` still guards it: dropping the scaling fails it under `udecimal_on`.

```
udecimal, 1,000-level book       big.Rat                    scaled
order fills in the first level   5.8 µs      53 allocs      0.7 µs      5 allocs
order takes all 1,000 levels     2.91 ms  34,667 allocs     0.76 ms  7,228 allocs
```

Both columns of the first row stop at the fill, as my other comment suggests, so it measures the cost of one level. On the default build the change is inert: same allocations and no significant time difference.

The unreachable lines are the error returns at [calculator.go:375](https://github.com/thrasher-corp/gocryptotrader/blob/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc/exchanges/orderbook/calculator.go#L375), 379, 396, 409, 413, 426, 430 and 442. A udecimal `String()` is always text that `big.Rat` parses, which [decimal_udecimal.go:221](https://github.com/thrasher-corp/gocryptotrader/blob/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc/types/decimal/decimal_udecimal.go#L221-L222) relies on too. `truncatedExecutionDecimal` always builds a valid 19-place decimal. Merging both backends' coverage leaves exactly those eight lines unhit, which matches Codecov's eight. With the change, `CalculateExecution` and `setTotals` are at 100% on both backends.

exchanges/orderbook/calculator_test.go
```diff
@@ -490,3 +491,320 @@ func TestFindNominalAmount(t *testing.T) {
...
+}
+
+func TestLevelsCalculateExecutionInsufficientLiquidity(t *testing.T) {
+	t.Parallel()
+	levels := Levels{{Price: 100, Amount: 2}, {Price: 110, Amount: 3}}
+
+	result, err := levels.CalculateExecution(decimal.NewFromInt(6), decimal.NewFromInt(1))
+	require.ErrorIs(t, err, ErrNotEnoughLiquidity, "CalculateExecution must report insufficient liquidity")
+	assert.True(t, result.ExecutedAmount.Equal(decimal.NewFromInt(5)), "executed amount should describe the partial fill")
+	assert.True(t, result.RemainingAmount.Equal(decimal.NewFromInt(1)), "remaining amount should describe the shortfall")
```
thrasher-, line 636: Heimdall: **Nit** — this test never checks `BaseAmount`, so reporting the requested base instead of the executed one passes the whole suite.

**Fix:** one line after this assertion:

```go
	assert.True(t, result.BaseAmount.Equal(decimal.NewFromInt(5)), "base amount should describe the partial fill")
```

**Why:** changing [calculator.go:405](https://github.com/thrasher-corp/gocryptotrader/blob/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc/exchanges/orderbook/calculator.go#L405) to `e.BaseAmount = e.RequestedAmount.Mul(multiplier)` leaves the package green on both backends. With the line above, this test fails on both, and it passes against this commit. `BaseAmount` is the one computed amount the partial-fill test leaves unchecked.

thrasher- (approved):
Heimdall: Approving. The exact quote accumulation in [`d44cb4e`](https://github.com/thrasher-corp/gocryptotrader/commit/d44cb4e4f715b174bdf5c81c6364ba3ee3a7a3cc) holds. Restoring the previous calculator fails the new sub-precision test under `udecimal_on`, and 19,743 random udecimal executions match an independent `big.Rat` reference exactly.

Inline, none blocking: two Minors and a Nit. On [calculator.go:362](https://github.com/thrasher-corp/gocryptotrader/pull/2401#discussion_r4088809554), checking every level makes a small order cost O(book depth). On [calculator.go:372](https://github.com/thrasher-corp/gocryptotrader/pull/2401#discussion_r4088809558), the façade's own arithmetic can replace `big.Rat` with identical results. On [calculator_test.go:615](https://github.com/thrasher-corp/gocryptotrader/pull/2401#discussion_r4088809559), the partial-fill test leaves `BaseAmount` unchecked.

The windows-latest failures are ambient. OKX's `TestGetOpenInterestAndVolumeStrike` got a live 51021 "Contract to be listed" and passes at this head now. The `exchange/websocket/buffer` runtime crash and the `exchanges/subscription` timeout come from test binaries that execute none of calculator.go.

thrasher- (approved):
Heimdall: Approving. All three findings from last round are fixed, verified by execution rather than read. A small order now costs the same at any depth: 1.1 µs at both 1,000 and 5,000 levels on the default backend, down from 597 µs and 3.0 ms. Across 58,338 random executions per backend the new arithmetic matches the previous commit in every field and error. Each new test fails when its fix is reverted, the scaling ones under `udecimal_on`.

The docker job's failure is ambient. Every failing test is `api.hbdm.com` timing out from that runner, the endpoint answers now, and the three ubuntu jobs that run the same wrapper tests passed.

Nothing further from me.

PR #2402, opened 2026-09-22
math: add signed decimal percentage comparison

shazbert:
Add SignedPercentageDifferenceDecimal while preserving the existing absolute percentage-difference API. Define zero-sum inputs as zero and retain the sign so callers can distinguish movement direction.

Add CompareSignedPercentageDifferenceDecimal with standard comparison semantics. For positive sums, compare cross-products directly to avoid decimal division and prevent rounding from changing decisions at target boundaries; retain the calculated fallback for zero and negative sums.

Cover positive, negative, equal, zero-sum and negative-sum behavior and add a benchmark for the comparison path. Tests support both configured decimal backends.

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


thrasher-:
@codex please review

common/math/math.go
```diff
@@ -58,6 +59,29 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
 	return x.Sub(y).Abs().Div(x.Add(y).Div(two)).Mul(oneHundred)
 }
 
+// SignedPercentageDifferenceDecimal returns the signed difference between two
+// decimal values as a percentage of their average. A positive result indicates
+// x is greater than y. A zero sum returns zero to avoid division by zero.
```
thrasher-: Heimdall: **Should fix** — "A positive result indicates x is greater than y" is false whenever `x+y < 0`, and the test table below asserts the counterexample.

Dividing by a negative sum inverts the sign, so the result tracks the sign of the denominator rather than the direction of movement:

```
x        y       x > y    SignedPercentageDifferenceDecimal    doc holds
-3       -1      false    100                                  no
-1       -3      true     -100                                 no
-100     -110    true     -9.52380952380952                    no
-0.5     -0.4    false    22.22222222222222                    no
 3        1      true     100                                  yes
 1        3      false    -100                                 yes
```

The first row is [math_test.go:97](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math_test.go#L97) — `{name: "negative sum", x: "-3", y: "-1", expected: "100"}` — and `-3` is not greater than `-1`. The test pins the behaviour the GoDoc forbids.

[math.go:77](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L77) inherits it: `CompareSignedPercentageDifferenceDecimal(-3, -1, 0)` returns `1`, i.e. "up by more than 0%", for a pair that moved down.

**Fix:** either

1. qualify the doc — the sign tracks `x-y` only when `x+y > 0`; or
2. divide by `sum.Abs()` so the sign always tracks direction, and change the `-3, -1` expectation to `-100`.

(2) is what the PR description asks for ("retain the sign so callers can distinguish movement direction"). (1) is fine if every caller is guaranteed a positive sum, but that is worth stating on the function rather than leaving to the reader.

common/math/math.go
```diff
@@ -58,6 +59,29 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
...
+func SignedPercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
+	sum := x.Add(y)
+	if sum.IsZero() {
+		return decimal.Zero
+	}
+	return x.Sub(y).Div(sum).Mul(twoHundred)
```
thrasher-: Heimdall: **Should fix** — dividing before multiplying discards two to three significant digits for nothing. `x.Sub(y).Mul(twoHundred).Div(sum)` is algebraically identical and spends the division's precision budget on the final magnitude instead of on a quotient that is then scaled up 200x.

Default backend, `x=1.471 y=1.469` — the pair this file already benchmarks at [math_test.go:142](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math_test.go#L142):

```
exact                              = 0.13605442176870748299
PercentageDifferenceDecimal        = 0.13605442176871       (14 dp)  err +2.52e-15
SignedPercentageDifferenceDecimal  = 0.1360544217687        (13 dp)  err -7.48e-15
mul-then-div                       = 0.1360544217687075     (16 dp)  err +1.70e-17
```

440x closer, and as it stands the new function is ~3x less accurate than the unsigned sibling immediately above it at [math.go:59](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L59) — same file, same inputs, one fewer digit. Under `-tags udecimal_on` the two siblings agree exactly (`0.13605442176870748`, err -2.99e-18) and the reorder still gains 32x (`0.1360544217687074829`, err -9.32e-20).

The cost is about 10% on the default backend (746 -> 818 ns/op, 448 -> 528 B/op, +2 allocs) and it is marginally faster under `udecimal_on` (65.0 -> 62.2 ns/op). Across 299,957 differential cases per backend, inputs up to 19 fractional places, the reorder was closer to the exact rational 278,816 times, identical 21,141 times and worse **0** times, with no panic and no overflow past `10^60`. Under `udecimal_on`: 289,625 / 10,332 / 0.

None of this is visible to the test: the 1e-12 tolerance at [math_test.go:86](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math_test.go#L86) is 260x the gap between the two backends for the table's own `110 / 100` row.

common/math/math.go
```diff
@@ -58,6 +59,29 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
...
+}
+
+// CompareSignedPercentageDifferenceDecimal compares the signed percentage
+// difference between x and y with target, returning standard Cmp semantics. It
+// avoids division when the sum is positive, preventing division rounding from
+// changing comparisons near a target boundary.
+func CompareSignedPercentageDifferenceDecimal(x, y, target decimal.Decimal) int {
+	sum := x.Add(y)
+	if !sum.IsPositive() {
```
thrasher-: Heimdall: **Should fix** — the exactness this function exists for silently lapses for negative sums, and most of it is recoverable. Dividing both sides by a negative sum only flips the inequality, so the cross-product works there too:

```go
sum := x.Add(y)
if sum.IsZero() {
	return decimal.Zero.Cmp(target)
}
c := x.Sub(y).Mul(twoHundred).Cmp(target.Mul(sum))
if sum.IsNegative() {
	return -c
}
return c
```

300,000 randomised comparisons against exact `math/big.Rat` arithmetic, inputs up to 19 fractional places, two thirds of the targets sitting on the rounded boundary:

```
                        default backend                 -tags udecimal_on
sum sign     cases   this PR   naive Div   negated    this PR   naive Div   negated
positive    224674         0      145227         0          4      145227         4
negative     75267     48760       48760         0      48760       48760         1
```

The positive-sum path is the point of the PR and it delivers — 0 wrong against 145,227 for the division it replaces. The negative branch is wrong exactly as often as the naive division, because it is the naive division.

On the default backend the negated form is exact in both branches. Under `udecimal_on` it inherits the same 19-place `Mul` truncation as the positive branch (see my comment on [math.go:82](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L82)), so it takes the negative branch from 48,760 to 1 rather than to 0 — and there is a constructible case where it is the worse of the two, `x=-0.0000000100000000001 y=-0.00000001 target=0.000000001`, where the current code happens to land on `-1` and the negated form on `0`. One against 48,760 still looks like the right trade, but it is not a clean sweep and worth knowing before you take it.

The fallback is also the slowest path: 1391-1846 ns/op against 500 ns/op for the cross-product, 760 B and 26 allocs against 376 B and 12.

The boundary-heavy target distribution makes these adversarial rates rather than typical ones — the comparison that matters is between the columns, not the absolute counts.

common/math/math.go
```diff
@@ -58,6 +59,29 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
...
+		return SignedPercentageDifferenceDecimal(x, y).Cmp(target)
+	}
+	return x.Sub(y).Mul(twoHundred).Cmp(target.Mul(sum))
```
thrasher-: Heimdall: **Minor** — "preventing division rounding from changing comparisons near a target boundary" holds on the default backend but not under `-tags udecimal_on`, which CI also builds.

udecimal's `Mul` truncates at 19 fractional digits ([decimal.go:472-493](https://github.com/quagmt/udecimal/blob/v1.10.1/decimal.go#L472), `defaultPrec = 19`; the façade says so too at [decimal_udecimal.go:76](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/types/decimal/decimal_udecimal.go#L76)). So `target.Mul(sum)` loses digits whenever `target` and `x+y` carry more than 19 fractional places between them, putting the rounding error straight back into the comparison the cross-product was written to make exact.

Eighteen-place amounts against a plain two-place percentage target are enough:

```
x = 1.000250000000002001   y = 0.999750000000002   target = 0.05

                  (x-y)*200          target*(x+y)              Compare
default      0.1000000000000002      0.10000000000000020005          -1    correct
udecimal_on  0.1000000000000002      0.1000000000000002               0    wrong
```

Exact rational arithmetic gives `-1`. Same inputs, same call, different answer per build tag.

The window is narrow — the comparison has to land within one ulp of the boundary. Over 224,674 randomised positive-sum comparisons with inputs up to 19 places it bit 4 times, and never when inputs were capped at 10 places. Nothing in the repo calls either function yet, so nothing is broken today. But the GoDoc states the guarantee unconditionally, so it is worth scoping it to the default backend or saying the exactness is best effort.

While you are in here: the comparator's equality is not the value function's equality, on either backend.

```
x = 20.000000000000000001   y = 19.999999999999999999   target = 0.00000000000000001

Compare(x, y, target)                     =  0     exact, correct
SignedPercentageDifferenceDecimal(x, y)   =  0     rounded to nothing
                          .Cmp(target)    = -1
```

The comparator is the one that is right. But a caller who reaches for both will get contradictory answers, which is worth a line on the GoDoc.

common/math/math_test.go
```diff
@@ -81,6 +81,73 @@ func BenchmarkDecimalPercentageDifference(b *testing.B) {
...
+	t.Parallel()
+	tests := []struct {
+		name     string
+		x        string
+		y        string
+		target   string
+		expected int
+	}{
+		{name: "above target", x: "3", y: "1", target: "99", expected: 1},
+		{name: "equal target", x: "3", y: "1", target: "100", expected: 0},
```
thrasher-, line 123: Heimdall: **Should fix** — none of the six rows can tell this implementation apart from the one-liner it was written to replace.

Replacing the entire body of `CompareSignedPercentageDifferenceDecimal` with

```go
return SignedPercentageDifferenceDecimal(x, y).Cmp(target)
```

leaves every row green, on both backends, and the rest of `common/math` with them:

```
--- PASS: TestCompareSignedPercentageDifferenceDecimal (0.00s)
    --- PASS: .../above_target          --- PASS: .../below_target
    --- PASS: .../equal_target          --- PASS: .../zero_sum
    --- PASS: .../negative_difference   --- PASS: .../negative_sum
ok  github.com/thrasher-corp/gocryptotrader/common/math   0.006s   (default)
ok  github.com/thrasher-corp/gocryptotrader/common/math   0.005s   (-tags udecimal_on)
```

Every row divides exactly — `3, 1` gives `2/4 = 0.5`, `-3, -1` gives `-2/-4 = 0.5` — so the division never rounds and the two implementations cannot disagree. This row is the one that ought to be the boundary case and isn't.

**Fix:** one row covers it, and discriminates on both backends:

```go
{name: "boundary above target", x: "1.471", y: "1.469", target: "0.13605442176870748", expected: 1},
```

```
                 this PR    naive Div    exact
default                1           -1        1
udecimal_on            1            0        1
```

common/math/math_test.go
```diff
@@ -81,6 +81,73 @@ func BenchmarkDecimalPercentageDifference(b *testing.B) {
 	}
 }
 
+func TestSignedPercentageDifferenceDecimal(t *testing.T) {
+	t.Parallel()
+	tolerance := decimal.MustFromString("0.000000000001")
```
thrasher-: Heimdall: **Nit** — two small things, neither blocking.

The 1e-12 tolerance is 260x the actual spread between the two backends for this table's own `110 / 100` row (`9.52380952380952` against `9.5238095238095238`, 3.8e-15), so it cannot see a precision change — including the 440x one available at [math.go:70](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L70). The two percentage tests directly above pin exact per-backend strings through [decimal_expectations_shopspring_test.go](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/decimal_expectations_shopspring_test.go) and [decimal_expectations_udecimal_test.go](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/decimal_expectations_udecimal_test.go), which would.

`assert.True` is used elsewhere in this file so I am not arguing the style, but on a bare boolean it prints neither number when it fails:

```
    math_test.go:106:
        	Error Trace:	common/math/math_test.go:106
        	Error:      	Should be true
        	Test:       	TestSignedPercentageDifferenceDecimal/positive_difference
        	Messages:   	SignedPercentageDifferenceDecimal should return the expected signed difference
```

`assert.Truef` carrying `result` and `expected` would at least say what went wrong.

Separately, [BenchmarkCompareSignedPercentageDifferenceDecimal](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math_test.go#L141) is missing the measured-result comment that both existing benchmarks in this file carry. On the default backend I get:

```go
// 2376517	       503.5 ns/op	     376 B/op	      12 allocs/op
```

thrasher- (changes requested):
Heimdall: The positive-sum cross-product is exact — 0 disagreements with exact `math/big.Rat` arithmetic over 224,674 boundary comparisons, against 145,227 for the division it replaces, at half the time and allocations. Evidence is on the lines.

**To unblock:**

1. [math.go:64](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L64) — the sign guarantee is false for every negative sum; the test's `-3, -1` row is the counterexample.
2. [math.go:70](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L70) — dividing before multiplying costs 2-3 digits; the reorder is 440x closer and never worse across 299,957 cases.
3. [math.go:79](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L79) — negative sums fall back to the naive division: 48,760 wrong of 75,267, against 0 for positive sums.
4. [math_test.go:122](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math_test.go#L122) — the table still passes with the cross-product deleted.

Not gating: a `udecimal_on` truncation caveat ([math.go:82](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math.go#L82)) and two nits ([math_test.go:86](https://github.com/thrasher-corp/gocryptotrader/blob/11558ed779b03e79085e58349bb94f5143bc9711/common/math/math_test.go#L86)). Pre-existing and out of scope: `PercentageDifferenceDecimal` panics on the default backend for equal positive inputs below ~1e-16 — happy to raise that separately. The one red check is ambient ([job](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35680010621/job/106594724006)).

Happy to approve once the four are in.

common/math/math.go
```diff
@@ -58,6 +59,32 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
 	return x.Sub(y).Abs().Div(x.Add(y).Div(two)).Mul(oneHundred)
 }
 
+// SignedPercentageDifferenceDecimal returns the difference between two decimal
+// values as a percentage of the absolute value of their average. The result's
+// sign indicates whether x is greater than y. A zero sum returns zero to avoid
```
thrasher-: Heimdall: **Nit** — two bits of wording that the new implementation has outgrown.

"The result's sign indicates whether x is greater than y" is now true for every non-zero result — `Abs()` fixed that, and I could not break it. But a zero result still cannot carry the distinction, and it is reachable with `x != y`:

```
x = 100.0000000000000000001   y = 100

default      S = 0      x > y
udecimal_on  S = 0      x > y
```

The quotient underflows the backend's division precision and the sign goes with it. `Compare(x, y, 0)` returns `1` here, correctly — so the comparator keeps the information the value function drops, which is worth a reader knowing. "A zero result means the difference is below the backend's precision, not that x equals y" would cover it.

Second, at [math.go:78-79](https://github.com/thrasher-corp/gocryptotrader/blob/997a51824776d3829a8412c4638b0904ae08af79/common/math/math.go#L78): "Its equality may differ" understates it — the two can land on opposite strict inequalities, not just disagree about equality.

```
x = 2   y = 1   target = 66.66666666666666668

default   S = 66.6666666666666667   S.Cmp(target) = +1   Compare = -1   exact = -1
```

The comparator is the correct one. "Its result may differ from comparing SignedPercentageDifferenceDecimal's rounded value with target" would be accurate without promising anything extra.

common/math/math.go
```diff
@@ -58,6 +59,32 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
...
+// avoids division to reduce rounding near a target boundary. Multiplication can
+// still truncate under the udecimal backend, so exactness is best effort there.
+// Its equality may differ from comparing SignedPercentageDifferenceDecimal's
+// rounded value with target.
+func CompareSignedPercentageDifferenceDecimal(x, y, target decimal.Decimal) int {
+	sum := x.Add(y)
+	if sum.IsZero() {
+		return decimal.Zero.Cmp(target)
+	}
+	return x.Sub(y).Mul(twoHundred).Cmp(target.Mul(sum.Abs()))
```
thrasher-: Heimdall: **Minor** — you took the "say it's best effort" option I offered, which is fair. But I have a number now, and I think it changes how this wants wording.

Targets drawn from exact `math/big.Rat` arithmetic rather than from `SignedPercentageDifferenceDecimal`'s own rounded output — 300,000 comparisons, operands `±10^12` with 0-19 fractional places, two thirds of targets sitting on the rounded boundary:

```
                      cases    Compare wrong     rate
default   positive   149655                0    0.000%
default   negative   150345                0    0.000%
udecimal  positive   149655             6657    4.448%
udecimal  negative   150345             6704    4.459%
```

Last round I put this at 1 in 225,031, which was an artefact of my harness — I had been deriving targets from `S`, which is already truncated, so the two agreed by construction. Against exact arithmetic it is about one boundary comparison in twenty-two.

It also does not need exotic inputs. Ten decimal places on each side is enough:

```
x = 100.0000000001   y = 100   target = 0.0000000001

               target*(x+y)                Compare
default        0.00000002000000000001           -1     correct
udecimal_on    0.00000002                        0     wrong
```

`target.precision + |x+y|.precision` is 20, one past udecimal's 19, and the last digit is the one that decides it. `x=0.3, y=0.2, target=40.0000000000000000001` does the same thing.

**Fix:** nothing in the arithmetic — I could not find a way to avoid it inside the façade's API. But "exactness is best effort there" reads like a precision footnote, and the failure mode is a *wrong comparison result*, not a slightly-off one. Something closer to "under the udecimal backend the comparison can be wrong when target and the sum together carry more than 19 fractional digits" would stop a caller assuming this is exact everywhere.

common/math/math_test.go
```diff
@@ -81,6 +81,78 @@ func BenchmarkDecimalPercentageDifference(b *testing.B) {
...
+		expected int
+	}{
+		{name: "above target", x: "3", y: "1", target: "99", expected: 1},
+		{name: "equal target", x: "3", y: "1", target: "100", expected: 0},
+		{name: "below target", x: "3", y: "1", target: "101", expected: -1},
+		{name: "negative difference", x: "1", y: "3", target: "-99", expected: -1},
+		{name: "zero sum", x: "0", y: "0", target: "1", expected: -1},
+		{name: "negative sum and difference", x: "-3", y: "-1", target: "-100", expected: 0},
+		{name: "negative sum and positive difference", x: "-1", y: "-3", target: "100", expected: 0},
+		{name: "boundary above target", x: "1.471", y: "1.469", target: "0.13605442176870748", expected: 1},
```
thrasher-, line 130: Heimdall: **Nit** — this row does the work on the default backend, but under `-tags udecimal_on` the comparator's reason to exist is still unprotected.

Mutating `common/math/math.go` and running just these two test functions:

```
mutation                                        default   udecimal_on
Compare -> SignedPct(x, y).Cmp(target)          FAIL      ok        <-- hole
Compare -> restore the !IsPositive fallback     FAIL      ok        <-- hole
S -> divide before multiply                     FAIL      FAIL
S -> drop Abs (restore the sign bug)            FAIL      FAIL
Compare -> drop Abs                             FAIL      FAIL
```

Three of the five are now caught on both backends, which is the important part. The two that aren't are both "revert the comparator to division" — they survive because the only rows that discriminate lean on shopspring's 16-digit `DivisionPrecision`, and udecimal's 19 places absorb it. Same reason [math_test.go:129-130](https://github.com/thrasher-corp/gocryptotrader/blob/997a51824776d3829a8412c4638b0904ae08af79/common/math/math_test.go#L129) pass vacuously there — `S` and the comparator agree under udecimal, so the rows named "rounded value differs from comparator" don't:

```
x = 20.000000000000000001  y = 19.999999999999999999  target = 0.00000000000000001
default      Compare=0   S=0                     S.Cmp(target)=-1   differs
udecimal_on  Compare=0   S=0.00000000000000001   S.Cmp(target)= 0   identical
```

**Fix:** two rows. One alone does not do it — the positive-sum row kills the first mutation, and the fallback mutation only reappears on a negative sum:

```go
{name: "boundary above target udecimal", x: "2", y: "1", target: "66.6666666666666666666", expected: 1},
{name: "negative sum boundary above target udecimal", x: "-1", y: "-2", target: "66.6666666666666666666", expected: 1},
```

Both pass against the implementation as written, on both backends, and with them in place:

```
mutation                                        default   udecimal_on
Compare -> SignedPct(x, y).Cmp(target)          FAIL      FAIL
Compare -> restore the !IsPositive fallback     FAIL      FAIL
```

Nineteen fractional digits on the target is what puts it past udecimal's division precision while staying inside its 19-place limit.

One more while you are here: both tables cover the zero-sum guard with `0, 0`, which is the uninteresting half of it. `5, -5` is the case a caller trips over — non-zero inputs that cancel, documented to report no difference at all. It still behaves (`S(5, -5) = 0`, `Compare(5, -5, 0) = 0` on both backends), it just is not pinned anywhere.

common/math/decimal_expectations_shopspring_test.go
```diff
@@ -5,6 +5,8 @@ package math
 const (
 	expectedDecimalPercentageDifferenceLarge = "196.03960396039604"
 	expectedDecimalPercentageDifferenceSmall = "0.13605442176871"
+	expectedSignedPercentageDifferenceLarge  = "9.5238095238095238"
```
thrasher-: Heimdall: **Nit** — the other six constants in this file are `expectedDecimal…`; these two are `expectedSigned…`. `expectedDecimalSignedPercentageDifference{Large,Small}` would keep the one prefix, at the cost of some length.

Worth a second look too: `Large`/`Small` now mean different things in the two families. `expectedDecimalPercentageDifferenceLarge` is `196.03960396039604`, from `1` against `100`; `expectedSignedPercentageDifferenceLarge` is `9.5238095238095238`, from `110` against `100`. Both files are consistent with themselves, so nothing is wrong — it just reads oddly that the "large" signed value is the smaller number.

thrasher- (approved):
Heimdall: All four unblock items fixed, and `sum.Abs()` is a better answer than the negated branch I proposed — it deletes the negative-sum special case rather than mirroring it.

Verified at [`997a518`](https://github.com/thrasher-corp/gocryptotrader/commit/997a51824776d3829a8412c4638b0904ae08af79):

```
                                    before             after
sign tracks x-y                     4 of 6 wrong       every non-zero result, both backends
Compare vs exact arithmetic         48760 / 75267      0 / 300000, both sum signs (default)
S(1.471, 1.469)                     13 dp, -7.5e-15    16 dp, +1.7e-17
implementation mutations caught     0 of 5             5 of 5 on default, 3 of 5 on udecimal
```

No panic or overflow through the new `Mul` path up to 300-digit magnitudes. CI green, lint clean on both build tags.

Four comments, none gating: a udecimal test-coverage gap, a measured rate behind the `udecimal_on` caveat you added, two bits of doc wording, and a constant-naming nit.

Still untouched and still out of scope: `PercentageDifferenceDecimal` two lines above has the same missing `Abs()` you just fixed here, returning `-100` for `(-3, -1)` from a function whose job is an absolute difference. It also panics on both backends for a cancelling pair like `(5, -5)`, and on the default backend alone for equal positive inputs below `5e-17`. Say the word and I'll raise it.

thrasher- (approved):
Heimdall: All four taken. Coverage is now complete on both backends — every implementation mutation I can construct fails the suite:

```
mutation                                    round 2         now
Compare -> naive division                   default only    both
Compare -> restore !IsPositive fallback     default only    both
S -> divide before multiply                 both            both
S -> drop Abs                               both            both
Compare -> drop Abs                         both            both
S -> drop zero-sum guard                    not tested      both
Compare -> drop zero-sum guard              not tested      both
S / Compare -> twoHundred to oneHundred     not tested      both
S -> swap operands                          not tested      both
```

Ten of ten, both build tags.

The reworded caveat is exactly right rather than merely directional. Over 1,053,154 boundary comparisons against exact `math/big.Rat` arithmetic under `udecimal_on`: **0 wrong** when target and `|x+y|` carry 19 or fewer fractional digits between them (659,105 cases), **3.63% wrong** above that (14,299 of 394,049). The default backend is 0 in both buckets. Writing a falsifiable number into a GoDoc and having it hold is not the usual outcome.

The rename preserved every constant byte-for-byte, `golangci-lint fmt --diff` and `gofmt -l` are clean, and `-race` and `-count 2` pass under both build tags.

Nothing outstanding from me. The `PercentageDifferenceDecimal` follow-up is still there whenever you want it.

common/math/math.go
```diff
@@ -58,6 +59,33 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
...
+	}
+	return x.Sub(y).Mul(twoHundred).Div(sum.Abs())
+}
+
+// CompareSignedPercentageDifferenceDecimal compares the signed percentage
+// difference between x and y with target, returning standard Cmp semantics. It
+// avoids division to reduce rounding near a target boundary. Under the udecimal
+// backend, the comparison can be incorrect when target and the sum of x and y
+// together carry more than 19 fractional digits. Its result may differ from
+// comparing SignedPercentageDifferenceDecimal's rounded value with target.
```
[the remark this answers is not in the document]
shazbert: I will do this for you AI gods. 🙏

common/math/math.go
```diff
@@ -58,6 +61,57 @@ func PercentageDifferenceDecimal(x, y decimal.Decimal) decimal.Decimal {
...
+// differ from comparing SignedPercentageDifferenceDecimal's rounded value with
+// target.
+func CompareSignedPercentageDifferenceDecimal(x, y, target decimal.Decimal) int {
+	sum := x.Add(y)
+	if sum.IsZero() {
+		return decimal.Zero.Cmp(target)
+	}
+	// Multiplication by 200 preserves scale; only target times sum can exceed
+	// the backend's fractional precision.
+	if decimal.MaxFractionalDigits == 0 ||
```
thrasher-: Heimdall: **Minor** — this guard is exact, but under `udecimal_on` it formats `target` and `sum` as strings before every comparison, when only an equal result can ever need them.

`BenchmarkCompareSignedPercentageDifferenceDecimal`, `-tags udecimal_on`, pinned to one CPU, eight interleaved runs:

```
                                  ns/op    B/op   allocs/op
3fb879f                            63.8       0           0
c82c31f                           153.9       8           2
reordered guard (below)            69.5       0           0
```

Single-character values such as `7` format without allocating. Anything longer pays. The default backend doesn't pay at all. `MaxFractionalDigits == 0` is a constant, the default build's comparator makes no call to `fractionalDigits`, `String` or `big.Rat`, and its timings are indistinguishable across all three versions (p > 0.15, identical allocations).

**Fix:** compare first, and consult the digit count only when the result is equal:

```go
	result := x.Sub(y).Mul(twoHundred).Cmp(target.Mul(sum.Abs()))
	// A truncated product differs from the exact product by less than one unit in the
	// backend's last fractional place, and the scaled difference is a whole number of
	// those units, so only an equal result can be inexact.
	if result != 0 || decimal.MaxFractionalDigits == 0 ||
		fractionalDigits(target)+fractionalDigits(sum) <= decimal.MaxFractionalDigits {
		return result
	}
```

**Why:** when udecimal truncates `target.Mul(sum.Abs())`, it cuts the coefficient to exactly 19 places ([decimal.go:472-517](https://github.com/quagmt/udecimal/blob/c9f302d60acf8142f608b419c5c95134f05969ae/decimal.go#L472-L517)). The truncated product is therefore a whole number of 10⁻¹⁹ units, and less than one unit from the exact product. `x.Sub(y).Mul(twoHundred)` never has more than 19 places, so it is a whole number of the same units. If it differs from the truncated product, it differs by at least a full unit, so it lies on the same side of the exact product and the sign is already right. Only equality can hide the discarded digits.

Checked against exact `math/big.Rat` arithmetic over 3,004,150 comparisons per backend: boundary targets at 0-19 places and ±1 in the last place, integer parts past u128, negative sums, and every case repeated with values padded to 19 internal places. This shape gets **0 wrong on both backends**, the same as yours. Every row in the table passes on both.

The cost moves to equal results, which now do the udecimal multiplication before falling back. On `PrecisionFallback` that is within noise (3.85 against 4.05 µs, ±5%). With operands past u128, where that multiplication goes through `big.Int`, it adds 26 allocations to the fallback's 97.

While here: neither new benchmark carries the result comment that the two above them do. `PrecisionFallback` only reaches the fallback under `udecimal_on` ([math_test.go:162](https://github.com/thrasher-corp/gocryptotrader/blob/c82c31fc8243174717667963825e5c2be6143b19/common/math/math_test.go#L162)), so that is the figure worth recording.

common/math/math_test.go
```diff
@@ -81,6 +81,95 @@ func BenchmarkDecimalPercentageDifference(b *testing.B) {
...
+		y        string
+		target   string
+		expected int
+	}{
+		{name: "above target", x: "3", y: "1", target: "99", expected: 1},
+		{name: "equal target", x: "3", y: "1", target: "100", expected: 0},
+		{name: "below target", x: "3", y: "1", target: "101", expected: -1},
+		{name: "negative difference", x: "1", y: "3", target: "-99", expected: -1},
+		{name: "zero sum", x: "0", y: "0", target: "1", expected: -1},
+		{name: "cancelling values", x: "5", y: "-5", target: "1", expected: -1},
```
thrasher-, line 129: Heimdall: **Minor** — the comparator's basic cases are only partly pinned, on both backends. The table never compares moving values against a zero target, never compares equal values, and only tries the zero-sum branch against a positive target. These mutations of [math.go:84-87](https://github.com/thrasher-corp/gocryptotrader/blob/c82c31fc8243174717667963825e5c2be6143b19/common/math/math.go#L84-L87) pass it on the default build as well as under `udecimal_on`:

```
mutation                                           default   udecimal_on
zero sum returns -1 always                         passes    passes
zero sum reports equality as +1                    passes    passes
zero sum returns -1 for any non-zero target        passes    passes
equal x and y return 0 before comparing            passes    passes
zero guard widened to cover a zero target          passes    passes
```

The zero-sum result depends only on the target's sign, and every zero-sum row uses target `1`. That gap predates this commit, and my "every mutation I can construct" last round missed it.

**Fix:** four rows. The first two go after `negative difference`, the other two after `cancelling values`:

```go
{name: "above zero target", x: "3", y: "1", target: "0", expected: 1},
{name: "matching values below target", x: "42.5", y: "42.5", target: "1", expected: -1},
```

```go
{name: "cancelling values above negative target", x: "5", y: "-5", target: "-1", expected: 1},
{name: "cancelling values equal zero target", x: "5", y: "-5", target: "0", expected: 0},
```

**Why:**

- **A zero target** is this comparator's most natural use: has x moved up at all?
- **`matching values`** mirrors the row the value-function table above already has.
- **The two cancelling rows** complete the zero-sum branch's three target sign classes.

All four pass on both backends, and with them in, every mutation above fails on both.

common/math/math_test.go
```diff
@@ -81,6 +81,95 @@ func BenchmarkDecimalPercentageDifference(b *testing.B) {
...
+		{name: "negative sum and positive difference", x: "-1", y: "-3", target: "100", expected: 0},
+		{name: "boundary above target", x: "1.471", y: "1.469", target: "0.13605442176870748", expected: 1},
+		{name: "boundary above target udecimal", x: "2", y: "1", target: "66.6666666666666666666", expected: 1},
+		{name: "negative sum boundary above target udecimal", x: "-1", y: "-2", target: "66.6666666666666666666", expected: 1},
+		{name: "rounded value differs from comparator", x: "20.000000000000000001", y: "19.999999999999999999", target: "0.00000000000000001", expected: 0},
+		{name: "negative sum rounded value differs from comparator", x: "-19.999999999999999999", y: "-20.000000000000000001", target: "0.00000000000000001", expected: 0},
+		{name: "cross-product precision below target", x: "1.0000000000000000001", y: "1", target: "0.00000000000000001", expected: -1},
+		{name: "negative difference above negative target", x: "1", y: "1.0000000000000000001", target: "-0.00000000000000001", expected: 1},
+		{name: "negative sum cross-product precision", x: "-1", y: "-1.0000000000000000001", target: "0.00000000000000001", expected: -1},
```
thrasher-, line 141: Heimdall: **Minor** — the table proves the fallback exists, but not the seam in front of it or the arithmetic inside it. Under `-tags udecimal_on`, these mutations of [math.go:88-113](https://github.com/thrasher-corp/gocryptotrader/blob/c82c31fc8243174717667963825e5c2be6143b19/common/math/math.go#L88-L113) pass it, each checked to compile:

```
mutation                                                        udecimal_on
guard removed, always the fast path                             caught
seam allows one digit more, or counts the difference's          passes
  digits instead of the sum's
fractionalDigits under-counts, caps at 18, or miscounts         passes
  negatives
fallback scales by 100 or 199, compares |x-y|, divides by       passes
  |x|+|y|, reports equality as +1, or rounds the target
  to 18 places
```

(The default build compiles all of this out.)

- **The seam** slips through because every row that reaches the fallback has 36 combined fractional digits, nowhere near 20.
- **The arithmetic** slips through because each fallback row keeps its sign under a wrong scale, sign or denominator, and none of them is an equality.

**Fix:** two rows at the end:

```go
{name: "negative 19-place target just beyond combined limit", x: "-0.8", y: "1.2", target: "-1000.0000000000000000001", expected: 1},
{name: "opposite signs cross-product precision equal target", x: "-0.0000000000000000001", y: "0.0000000000000000126", target: "-203.2", expected: 0},
```

**Why:**

- **The 19-place row** sits at exactly 20 combined digits, where udecimal truncates the deciding digit. `−1000.0000000000000000001 × 0.4` gives `−400`, not `−400.00000000000000000004`. It uses every place the backend has, with a negative target, and its difference (`−2`) has fewer digits than its sum (`0.4`). The seam, the digit count and the choice of operand all have to be right.
- **The last row** is an exact equality past the seam, with the inputs on opposite sides of zero: `200 × −127e-19 = −203.2 × 125e-19`. To return zero, the fallback has to get the scale, the sign of the difference, the denominator and equality all right.

Both pass on both backends. With them in, every mutation above fails, and the seven the table already catches stay caught. The same holds with the guard reordering on [math.go:90](https://github.com/thrasher-corp/gocryptotrader/blob/c82c31fc8243174717667963825e5c2be6143b19/common/math/math.go#L90), where inverting its new shortcut is caught too.

types/decimal/decimal_shopspring.go
```diff
@@ -7,6 +7,9 @@ import shopspring "github.com/shopspring/decimal" //nolint:depguard // Default b
 // Implementation identifies the selected decimal backend.
 const Implementation = "shopspring/decimal"
 
+// MaxFractionalDigits is zero because the default backend has no fixed fractional precision limit.
```
thrasher-, line 10: Heimdall: **Nit** — "no fixed fractional precision limit" holds for shopspring values and products, which is all this constant gates, but not for `Div`, which rounds to `DivisionPrecision`.

For an exported constant named `MaxFractionalDigits`, that ranks the backends the wrong way round. Default-backend division is *less* precise than udecimal's, down to dividing a representable value by one and getting zero:

```
                     MaxFractionalDigits   1/3                     0.0000000000000000001/1
shopspring/decimal                     0   0.3333333333333333      0
quagmt/udecimal                       19   0.3333333333333333333   0.0000000000000000001
```

**Fix:** scope it to what it guarantees:

```go
// MaxFractionalDigits is zero because default-backend values and products keep every
// fractional digit. Div still rounds to shopspring's DivisionPrecision.
```

thrasher- (approved):
Heimdall: Approving. The `big.Rat` fallback does what I told you the façade couldn't. Over 3,004,150 comparisons per backend against exact rational arithmetic, it gets 0 wrong on both. The same inputs at [`3fb879f`](https://github.com/thrasher-corp/gocryptotrader/commit/3fb879f0383bf9d7c45dc0aa89aa964378ba60a6) gave 5,402 wrong under `udecimal_on`, every one of them past 19 combined fractional digits.

Inline, none blocking:

- Minor, [math.go:90](https://github.com/thrasher-corp/gocryptotrader/pull/2402#discussion_r4079599232): the guard's cost on the udecimal fast path, with an exact reordering that removes it.
- Minor, [math_test.go:127](https://github.com/thrasher-corp/gocryptotrader/pull/2402#discussion_r4079599243): four rows for the comparator's basic cases, on both backends.
- Minor, [math_test.go:137](https://github.com/thrasher-corp/gocryptotrader/pull/2402#discussion_r4079599250): two rows for the seam and the fallback's arithmetic.
- Nit, [decimal_shopspring.go:10](https://github.com/thrasher-corp/gocryptotrader/pull/2402#discussion_r4079599257).

The PR description still says negative sums "retain the calculated fallback", which stopped being true in [`997a518`](https://github.com/thrasher-corp/gocryptotrader/commit/997a51824776d3829a8412c4638b0904ae08af79).

Take them or leave them. Nothing here blocks.

thrasher- (approved):
Heimdall: Approving [`c699457`](https://github.com/thrasher-corp/gocryptotrader/commit/c699457f897bd077cc21363904b7b16058dd0746). All four from the last round are in, and each holds when run:

- **The guard reorder** makes the udecimal fast path allocation-free again: 68 ns/op with no allocations, against about 155 ns/op and 2 allocations at [`c82c31f`](https://github.com/thrasher-corp/gocryptotrader/commit/c82c31fc8243174717667963825e5c2be6143b19). I compared it against exact `math/big.Rat` arithmetic over 6,000,000 comparisons, 30,042 of which reached the fallback, and none was wrong on either backend. Every non-zero fast result was already right, as the new comment says it must be.
- **The six new rows** catch every mutation named in the two test threads under `udecimal_on`, and they catch an inverted shortcut. The comparator-basics mutations fail on the default build as well.
- **The `MaxFractionalDigits` GoDoc** is now scoped to values and products.

The description still says negative sums "retain the calculated fallback", which stopped being true in [`997a518`](https://github.com/thrasher-corp/gocryptotrader/commit/997a51824776d3829a8412c4638b0904ae08af79).

The one red job is ambient. On [windows-latest](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35936260429/job/107433871875), LBank's live `TestGetTickers` timed out reading the response body. This diff touches no exchange code, and nothing calls the new functions yet.

Nothing outstanding from me.

gloriousCode (approved):
TySB

PR #2403, opened 2026-09-22
order limits: add contract-aware amount stepping

shazbert:
Add AmountStep to keep an exchange order increment and contract multiplier together, with spot markets represented by a multiplier of one.

Provide exact base-unit floor and ceiling operations, order-unit flooring, and base increment calculation. Validate non-positive amounts, increments and multipliers through exported sentinel errors.

Calculate the smallest common base increment for two markets using rational least-common-multiple arithmetic. Preserve scales beyond the decimal backend's default division precision so small contract increments remain exact.

Cover spot and derivative quantities, fractional and large multipliers, validation failures, symmetry, high-precision increments and both supported decimal backends, with a benchmark for the rational path.

## Type of change

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist


thrasher-:
@codex please review

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,144 @@
...
+// amount step.
+func (a AmountStep) BaseIncrement() (decimal.Decimal, error) {
+	increment, err := a.orderIncrement()
+	if err != nil {
+		return decimal.Zero, err
+	}
+	if !a.ContractMultiplier.IsPositive() {
+		return decimal.Zero, fmt.Errorf("%w: %s", ErrContractMultiplierNotPositive, a.ContractMultiplier)
+	}
+	return increment.Mul(a.ContractMultiplier), nil
```
thrasher-: Heimdall: **Should fix** — `BaseIncrement()` validates both operands but not the product, so it can return `(0, nil)`. `FloorBaseAmount`, `CeilBaseAmount` and `CommonBaseIncrement` then use that zero as a divisor and panic under `-tags udecimal_on`.

`Mul` truncates at udecimal's 19 fractional digits, so any `Increment × ContractMultiplier` below `1e-19` underflows to zero and slips past the guard on [amount_step.go:36](https://github.com/thrasher-corp/gocryptotrader/blob/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028/exchange/order/limits/amount_step.go#L36).

**Why:** with `step = AmountStep{Increment: 0.0000000001, ContractMultiplier: 0.0000000001}` and `unit = AmountStep{1, 1}`:

```
                                 default build             -tags udecimal_on
BaseIncrement()                  "0.00000000000000000001"  "0", err = <nil>
FloorBaseAmount(1)               "1"                       PANIC: can't divide by zero
CeilBaseAmount(1)                "1"                       PANIC: can't divide by zero
step.CommonBaseIncrement(unit)   "1"                       "0", err = <nil>
unit.CommonBaseIncrement(step)   "1"                       PANIC: can't divide by zero
```

The last two rows are the same pair in opposite order: `CommonBaseIncrement` returns a silently wrong zero one way and panics the other, while [amount_step_test.go:201-203](https://github.com/thrasher-corp/gocryptotrader/blob/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028/exchange/order/limits/amount_step_test.go#L201-L203) asserts symmetry.

It is not only the degenerate case. Sweeping 5000 random `(Increment, ContractMultiplier)` pairs per scale under `-tags udecimal_on`, then calling `FloorBaseAmount(1)` on each:

```
max fractional digits   BaseIncrement == 0   FloorBaseAmount panics
 8                          0 / 5000             0 / 3000
10                          0 / 5000             0 / 3000
12                         18 / 5000            11 / 3000
16                        651 / 5000           437 / 3000
19                       1370 / 5000           793 / 3000
```

The default build is 0 across every row.

**Fix:** validate the product rather than just the operands:

```go
baseIncrement := increment.Mul(a.ContractMultiplier)
if !baseIncrement.IsPositive() {
    return decimal.Zero, fmt.Errorf("%w: increment %s with multiplier %s underflows to %s",
        ErrAmountStepNotPositive, increment, a.ContractMultiplier, baseIncrement)
}
return baseIncrement, nil
```

With that applied all four call sites return the sentinel instead of panicking, symmetry is restored, the default build is byte-for-byte unchanged, and both backends stay green — `go test ./exchange/order/limits/ -race -count 1` with and without `-tags udecimal_on`, `golangci-lint run ./exchange/order/limits/...` clean under both.

For the regression test, the boundary is exactly 20 combined fractional digits — one row in the existing table covers it:

```
inc=0.000000000000000002  mul=1     -> "0.000000000000000002"   both backends   (18 dp, what the suite has today)
inc=0.000000000000000002  mul=0.1   -> "0.0000000000000000002"  both backends   (19 dp)
inc=0.000000000000000002  mul=0.01  -> "0.00000000000000000002" default
                                       "0"                      udecimal_on     (20 dp — first divergence)
```

The unvalidated product is not a udecimal quirk, either — the default build has its own route through the same `Mul`. `1e-1073741825` parses fine under shopspring and reports `IsPositive() == true`:

```
backend=shopspring/decimal
  parsed ok, IsPositive=true
  BaseIncrement PANIC: exponent -2147483650 overflows an int32!
```

(udecimal rejects that input at parse time, so it is default-build only.) The check above does *not* catch this one — the panic happens inside `Mul`, before there is a product to test — and the input is absurd enough that I would not chase it. I mention it only because it shows the root cause is "the product is never examined", not "udecimal truncates".

To be straight about severity: I could not find a live market that gets anywhere near this. It needs `Increment × ContractMultiplier < 1e-19`; the smallest `lotSz × ctVal` across OKX's 486 swaps and 225 futures is `1e-4` (`BTC-USDT-SWAP`, `lotSz=0.01 × ctVal=0.01`), and the smallest `LOT_SIZE` `stepSize` across Binance spot's 3705 symbols is `1e-6`. Nothing in the tree calls `AmountStep` yet either.

So this is hardening, not an outage waiting to happen. It is still worth the one line: it is a hole in the contract of a method whose entire job is validate-then-multiply, the failure mode is a panic out of a package that returns errors everywhere else, and it is cheaper to close now while the API has no callers than after it has some.

exchange/order/limits/amount_step_test.go
```diff
@@ -0,0 +1,215 @@
...
+			name:     "one increment divides the other",
+			first:    AmountStep{Increment: decimal.MustFromString("0.1"), ContractMultiplier: decimal.NewFromInt(1)},
+			second:   AmountStep{Increment: decimal.MustFromString("0.01"), ContractMultiplier: decimal.NewFromInt(1)},
+			expected: "0.1",
+		},
+		{
+			name:     "non-divisible decimal increments",
+			first:    AmountStep{Increment: decimal.MustFromString("0.2"), ContractMultiplier: decimal.NewFromInt(1)},
+			second:   AmountStep{Increment: decimal.MustFromString("0.03"), ContractMultiplier: decimal.NewFromInt(1)},
+			expected: "0.6",
```
thrasher-, line 222: Heimdall: **Should fix** — no case in this table distinguishes a least common multiple from a plain numerator product, so the `greatestCommonDivisor` call at [amount_step.go:113](https://github.com/thrasher-corp/gocryptotrader/blob/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028/exchange/order/limits/amount_step.go#L113) is never load-bearing.

Both non-divisible cases have coprime numerators once `big.Rat` reduces them — `0.2`/`0.03` is `1/5` and `3/100`, `2e-18`/`3e-18` is `1/(5·10¹⁷)` and `3/10¹⁸` — so `lcm(num₁, num₂)` already equals `num₁·num₂` and removing the GCD changes nothing.

**Why:** swapping `numeratorLCM` for `numeratorProduct` at [amount_step.go:116](https://github.com/thrasher-corp/gocryptotrader/blob/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028/exchange/order/limits/amount_step.go#L116) leaves the suite green:

```
M2: numerator LCM -> plain numerator product   : SURVIVED (tests still pass)
M4: denominator GCD -> denominator of first     : killed (3 failing tests)
M6: CeilBaseAmount LessThan -> LessThanOrEqual  : killed (2 failing tests)
M8: FloorBaseAmount drops the multiplier        : killed (6 failing tests)
```

**Fix:** add one case with non-coprime numerators. `0.6` and `0.9` reduce to `3/5` and `9/10`, so the answer is `lcm(3,9)/gcd(5,10)` = `9/5` = `1.8`:

```
LCM(0.6, 0.9) = "1.8"   as written
LCM(0.6, 0.9) = "5.4"   with the numerator GCD dropped
```

For what it is worth the arithmetic itself is right — 20,000 random increment pairs across 0–6 fractional digits agree exactly with an independent `big.Rat` reference on both backends, and every result is an exact common multiple of both inputs. This is purely about the test that proves it.

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,144 @@
...
+// FloorBaseAmount rounds a base-asset amount down to an executable increment.
+func (a AmountStep) FloorBaseAmount(amount decimal.Decimal) (decimal.Decimal, error) {
+	if !amount.IsPositive() {
+		return decimal.Zero, fmt.Errorf("%w: %s", ErrAmountNotPositive, amount)
+	}
+	increment, err := a.BaseIncrement()
+	if err != nil {
+		return decimal.Zero, err
+	}
+	return amount.Sub(amount.Mod(increment)), nil
```
thrasher-, line 65: Heimdall: **Minor** — `FloorBaseAmount` rejects `amount <= 0`, then returns `0` with a nil error for any `0 < amount < increment`.

`AmountStep{Increment: 1, ContractMultiplier: 1}.FloorBaseAmount(0.5)` returns `("0", <nil>)` on both backends; `FloorOrderAmount` does the same. A caller that floors and submits sends a zero-size order, and has to re-check the result to notice.

The existing [levels.go:99-101](https://github.com/thrasher-corp/gocryptotrader/blob/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028/exchange/order/limits/levels.go#L99-L101) behaves the same way, so this may well be deliberate — but that helper does no validation at all, whereas these methods take the trouble to reject a zero input with `ErrAmountNotPositive` and then hand back a zero output.

**Fix:** either say so in the two GoDoc comments, or return a sentinel when the floor collapses to zero so the distinction is in the type rather than in the caller.

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,144 @@
...
+	if first.Mod(second).IsZero() {
+		return first, nil
+	}
+	if second.Mod(first).IsZero() {
+		return second, nil
+	}
+
+	firstRat, ok := new(big.Rat).SetString(first.String())
+	if !ok {
+		return decimal.Zero, fmt.Errorf("cannot parse first base increment %q", first)
```
thrasher-: Heimdall: **Nit** — both `!ok` branches are unreachable, and unlike the rest of the package's failures they aren't sentinels.

`big.Rat.SetString` accepts everything either backend's `Decimal.String()` can emit. I checked the property that matters here — neither ever returns exponent notation — across `1e-19`, `1e20` and trailing-zero values on both builds:

```
in=0.60                        String()="0.6"                    scale=1
in=1.50                        String()="1.5"                    scale=1
in=0.0000000000000000001       String()="0.0000000000000000001"  scale=19
in=100000000000000000000       String()="100000000000000000000"  scale=0
```

Worth keeping as defence in depth — and it is the same property [`decimalScale`](https://github.com/thrasher-corp/gocryptotrader/blob/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028/exchange/order/limits/amount_step.go#L131) silently relies on, since an exponent form would make it return `0` and round the LCM to an integer. If either is ever expected to be matched on, they want to be sentinels like the other three.

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,144 @@
...
+		return decimal.Zero, err
+	}
+	if first.Mod(second).IsZero() {
+		return first, nil
+	}
+	if second.Mod(first).IsZero() {
+		return second, nil
+	}
+
+	firstRat, ok := new(big.Rat).SetString(first.String())
```
thrasher-: Heimdall: **Nit** — `first.String()` and `second.String()` are each evaluated twice, once here and again for the scale on [amount_step.go:117](https://github.com/thrasher-corp/gocryptotrader/blob/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028/exchange/order/limits/amount_step.go#L117). Since you benchmarked this path, the numbers:

```
                       ns/op   B/op   allocs/op
as written             3225    1096      70       (default build)
strings hoisted        2840    1064      64       -12%
as written             2245     568      44       (-tags udecimal_on)
strings hoisted        2060     560      42        -8%
```

`go test -bench BenchmarkAmountStep -benchtime 2s -count 3`, medians. Both backends stay green with the strings hoisted into locals.

thrasher- (changes requested):
Heimdall: Arithmetic verified correct — 20,000 random LCM pairs and all 648 floor/ceil combinations exact on both decimal backends. Evidence for each finding is on the line.

**To unblock:**
1. [amount_step.go:39](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068253426) — `BaseIncrement()` never validates the product, so it returns `(0, nil)` and three methods then panic.
2. [amount_step_test.go:156](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068253428) — no case distinguishes the LCM from a plain numerator product.

Non-gating: [amount_step.go:51](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068253431) · [amount_step.go:104](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068253436) · [amount_step.go:102](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068253440).

Both red CI jobs are ambient — live-network tests (OKX, Deribit, etherscan, BTSE) in packages this PR doesn't touch.

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,154 @@
...
+// amount step.
+func (a AmountStep) BaseIncrement() (decimal.Decimal, error) {
+	increment, err := a.orderIncrement()
+	if err != nil {
+		return decimal.Zero, err
+	}
+	if !a.ContractMultiplier.IsPositive() {
+		return decimal.Zero, fmt.Errorf("%w: %s", ErrContractMultiplierNotPositive, a.ContractMultiplier)
+	}
+	baseIncrement := increment.Mul(a.ContractMultiplier)
```
thrasher-, line 45: Heimdall: **Should fix** — the guard rejects a product that truncates to *zero*, but accepts one that truncates to a *wrong non-zero value*. The result is a floored quantity that is not executable on the market it describes, returned with a nil error.

**Why:** `Increment = 0.0000000003`, `ContractMultiplier = 0.0000000005`. The exact product is `1.5e-19`; udecimal holds 19 fractional digits, so it truncates to `1e-19` — positive, so it passes.

```
                            default build              -tags udecimal_on
exact product               0.00000000000000000015     0.00000000000000000015
BaseIncrement()             0.00000000000000000015     0.0000000000000000001   <- truncated
matches exact?              true                       false
FloorBaseAmount(1)          0.9999999999999999999      1
CeilBaseAmount(1)           1.00000000000000000005     1
CommonBaseIncrement(unit)   3                          1
multiple of exact increment true                       false (quotient 20000000000000000000/3)
```

Under `udecimal_on` floor and ceil both return `1`, and `1` is not a whole number of `1.5e-19` increments, so an order sized from it would be rejected by the venue. No panic, no error — and floor and ceil agreeing is exactly what a caller reads as "already on an increment".

Sampling 20,000 random increment/multiplier pairs against an exact rational oracle: **0 mismatches on the default build, 236 under `udecimal_on`**. Not a regression from this commit — the same 236 reproduce at [`7c6ad5db0`](https://github.com/thrasher-corp/gocryptotrader/commit/7c6ad5db0ed60fd3b8c967541616a3b79cf8d028).

**Fix:** check that the product is exact, not just positive. An unconditional `big.Rat` comparison catches every case above and leaves the legitimate ones alone, but costs 2889 → 6516 ns/op and 64 → 134 allocs on your benchmark. Gating it on the combined scale keeps the common path close to free:

```go
if incText, mulText := increment.String(), a.ContractMultiplier.String(); decimalScale(incText)+decimalScale(mulText) > 19 {
    // exact big.Rat comparison here
}
```

Identical results on both backends, 3448 ns/op and 76 allocs. The `19` wants to come from the backend rather than a literal, which the façade doesn't expose — so the alternative is to leave the arithmetic alone and document that `AmountStep` requires a representable product, making it the caller's problem knowingly rather than silently.

Reachability caveat as before: OKX's smallest `lotSz × ctVal` across 711 instruments is `1e-4`, and Binance spot's smallest `stepSize` across 3705 symbols is `1e-6`, so nothing live is close. Your call whether it is worth the cycles — but silently returning a non-executable quantity is a worse failure mode than the panic was.

exchange/order/limits/amount_step_test.go
```diff
@@ -0,0 +1,250 @@
...
+		_, err = step.CeilBaseAmount(decimal.NewFromInt(1))
+		assert.ErrorIs(t, err, ErrAmountStepNotPositive, "CeilBaseAmount should reject an underflowed increment")
+		_, err = step.CommonBaseIncrement(unit)
+		assert.ErrorIs(t, err, ErrAmountStepNotPositive, "CommonBaseIncrement should reject an underflowed first increment")
+		_, err = unit.CommonBaseIncrement(step)
+		assert.ErrorIs(t, err, ErrAmountStepNotPositive, "CommonBaseIncrement should reject an underflowed second increment")
+		return
+	}
+	require.NoError(t, err, "BaseIncrement must accept a representable positive product")
+	assert.Equal(t, expectedProduct, result, "BaseIncrement should return the representable product")
```
thrasher-: Heimdall: **Nit** — this compares two `decimal.Decimal` values with `assert.Equal`, which is `reflect.DeepEqual` on the backend's struct rather than numeric equality. It is the only assertion in the file not going through `.String()`.

A numerically neutral change to `BaseIncrement` breaks it on the default build:

```go
return baseIncrement.Mul(decimal.MustFromString("1.0")), nil
```

```
expected: coefficient=1,  exponent=-20
actual:   coefficient=10, exponent=-21
```

Same number, different representation, failed assertion. udecimal has the same hazard from the other direction — it treats `0.1` and `0.10` as equal where shopspring does not — so the two backends disagree about what this assertion even means.

Safe as written, since both sides come from the same `Mul`. But that is also why it is weak: `result` *is* `Increment.Mul(ContractMultiplier)`, so both sides are the same expression. `assert.Equal(t, expectedProduct.String(), result.String(), …)` would be consistent with the rest of the file and immune to both.

The underflow half of the test is doing real work — reverting the guard fails it under `-tags udecimal_on`, which is the configuration CI runs it in.

thrasher- (commented):
Heimdall: All five fixed and resolved, each verified by execution.

- The panics are gone — all five operations return the wrapped sentinel under `-tags udecimal_on`, both `CommonBaseIncrement` operand orders agree, and the default build is unchanged. Reverting the guard fails `TestAmountStepBaseIncrementUnderflow` under that tag, and the `udecimal=true` job passed.
- `{0.6, 0.9} → 1.8` closes the LCM gap: dropping the numerator GCD now fails with `expected: "1.8"` / `actual: "5.4"`.
- Allocations down 70 → 64 (default) and 44 → 42 (`udecimal_on`).
- No regression — 20,000 random LCM pairs exact against an independent oracle, all 648 floor/ceil combinations in contract, both backends. Lint, vet and `fmt --diff` clean under both tags.

Two new comments, evidence on the lines: [amount_step.go:40](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068819528) · [amount_step_test.go:46](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068819536).

The red Windows job is ambient: `TestGetLatestFundingRates` fails on `ICX_USDT pair not contained in available pairs` after Gate moved 444 pairs in and 95 out — upstream data drift, not this PR.

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,175 @@
...
+	if !baseIncrement.IsPositive() {
+		return decimal.Zero, fmt.Errorf("%w: increment %s with multiplier %s underflows to %s",
+			ErrAmountStepNotPositive, increment, a.ContractMultiplier, baseIncrement)
```
thrasher-: Heimdall: **Minor** — a zero product here can only mean udecimal's `Mul` truncated it, so it wants the same sentinel as the truncation check at [amount_step.go:61-63](https://github.com/thrasher-corp/gocryptotrader/blob/3d114e70d5fa96404a945d98b88527300eddc713/exchange/order/limits/amount_step.go#L61-L63), not `ErrAmountStepNotPositive`.

Both operands are already validated positive at [amount_step.go:35-41](https://github.com/thrasher-corp/gocryptotrader/blob/3d114e70d5fa96404a945d98b88527300eddc713/exchange/order/limits/amount_step.go#L35-L41), and the default build's `Mul` is exact, so this branch is unreachable there. Under `-tags udecimal_on` it splits one cause across two sentinels, depending on whether the truncated product happens to be zero:

```
Increment=0.000000000000000003  ContractMultiplier=0.05  ->  base increment cannot be represented exactly: … truncates to 0.0000000000000000001
Increment=0.000000000000000003  ContractMultiplier=0.01  ->  amount step must be greater than zero: … underflows to 0
```

The second blames the exchange data, since everywhere else `ErrAmountStepNotPositive` means a zero or negative `Increment`. A caller matching `ErrBaseIncrementNotRepresentable` to mean "this market needs the default build" misses it. Across 200,000 random increment/multiplier pairs, 42,641 truncations came back as `ErrAmountStepNotPositive` and 39,161 as `ErrBaseIncrementNotRepresentable`.

**Fix:** return `ErrBaseIncrementNotRepresentable` on this line, and swap the five `ErrAmountStepNotPositive` expectations in `TestAmountStepBaseIncrementUnderflow` at [amount_step_test.go:33-42](https://github.com/thrasher-corp/gocryptotrader/blob/3d114e70d5fa96404a945d98b88527300eddc713/exchange/order/limits/amount_step_test.go#L33-L42). With that, all 81,802 carry the one sentinel, the default build is unchanged, and `go test -race` and `golangci-lint run` are clean under both tags. Deleting lines 43-46 also works, because the exactness check then rejects the zero as "truncates to 0".

This sentinel is here because of my round-one suggestion, which predates `ErrBaseIncrementNotRepresentable`.

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,175 @@
...
+	}
+	if decimal.MaxFractionalDigits > 0 &&
```
thrasher-: Heimdall: **Nit** — under `-tags udecimal_on` this gate allocates two strings on every call, including the common case it exists to skip. A `Div` round-trip is an exact check on its own and allocates nothing.

udecimal's `Div` truncates just like its `Mul`, so a truncated product divides back to strictly less than the increment, and an exact one divides back to it unchanged. The `MaxFractionalDigits > 0` guard keeps it off the default build, where `Div` rounds instead:

```go
	baseIncrement := increment.Mul(a.ContractMultiplier)
	// udecimal's Div truncates like its Mul, so a truncated product, including one
	// that underflows to zero, cannot divide back to the increment.
	if decimal.MaxFractionalDigits > 0 && !baseIncrement.Div(a.ContractMultiplier).Equal(increment) {
		return decimal.Zero, fmt.Errorf("%w: increment %s with multiplier %s truncates to %s",
			ErrBaseIncrementNotRepresentable, increment, a.ContractMultiplier, baseIncrement)
	}
	return baseIncrement, nil
```

That replaces [amount_step.go:42-66](https://github.com/thrasher-corp/gocryptotrader/blob/3d114e70d5fa96404a945d98b88527300eddc713/exchange/order/limits/amount_step.go#L42-L66) and drops the three unreachable `!ok` branches. It also settles the sentinel comment on line 45, because a zero product comes out as "truncates to 0".

```
-tags udecimal_on, -cpu 1, n=8   93c3c93a6   as written             Div round-trip
BaseIncrement                    19.9 ns     118.3 ns, 2 allocs     59.3 ns, 0 allocs
FloorBaseAmount                  86.9 ns     192.9 ns, 2 allocs     112.2 ns, 0 allocs
CommonBaseIncrement              80.3 ns     265.1 ns, 3 allocs     160.3 ns, 0 allocs
BaseIncrement, 2e-10 × 5e-10     23.1 ns     2,360 ns, 36 allocs    59.9 ns, 0 allocs
```

The first three rows use steps shaped like OKX `BTC-USDT-SWAP` (`0.01 × 0.01`). The last row uses the inputs from `TestAmountStepBaseIncrementHighScaleExactProduct`, an exact product that takes the `big.Rat` path. The default build is unchanged either way, because the `MaxFractionalDigits > 0` short-circuit folds the check out. Against an exact `big.Rat` oracle, the `Div` form and the current check agreed on every one of 400,000 pairs, 50,795 of them beyond udecimal's u128 range. With it applied, both tags pass `-race` and lint, and removing the check fails both the truncation and underflow tests. This corrects my round-two claim [in that thread](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4068819528).

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,175 @@
...
+	// ErrContractMultiplierNotPositive is returned when a contract multiplier is invalid.
+	ErrContractMultiplierNotPositive = errors.New("contract multiplier must be greater than zero")
+	// ErrBaseIncrementNotRepresentable is returned when the decimal backend truncates a positive base increment.
+	ErrBaseIncrementNotRepresentable = errors.New("base increment cannot be represented exactly")
+	errCannotParseBaseIncrement      = errors.New("cannot parse base increment")
+)
+
+// AmountStep describes the executable order increment and the base amount
+// represented by each order unit. Spot markets conventionally use a contract
+// multiplier of one.
```
thrasher-: Heimdall: **Nit** — worth stating the unit `ContractMultiplier` must be in. GCT's own `futures.Contract.Multiplier` is the obvious thing to feed in here, and it isn't a base amount everywhere.

From GCT's `GetFuturesContractDetails`, live today:

```
                   SettlementType  Multiplier  actually
OKX BTC-USDT-SWAP  linear          0.01        ctVal in BTC: a base amount, correct
OKX BTC-USD-SWAP   linear          100         ctVal in USD: a quote amount, on an inverse swap
Bybit BTCUSDT      linear          0.01        leverageStep: quantity is already in BTC
```

Fed straight in, both of the last two return wrong base amounts with a nil error:

```
AmountStep{Increment: 0.1,   ContractMultiplier: 100}   BaseIncrement=10       FloorBaseAmount(0.5)=0
AmountStep{Increment: 0.001, ContractMultiplier: 0.01}  BaseIncrement=0.00001  FloorBaseAmount(0.0015)=0.0015
```

The first is 10 USD, so 0.5 BTC floors to nothing. The second is 100 times finer than Bybit's `qtyStep` of 0.001, so it returns a quantity Bybit would reject. Nothing calls `AmountStep` yet, so this only concerns the contract the doc sets for the first caller. Both wrapper bugs are listed in the review body.

```suggestion
// multiplier of one. The base-amount methods require ContractMultiplier to be
// a fixed base-asset amount per order unit, so a quote-denominated contract
// value, such as an inverse contract's, cannot be used directly.
// FloorOrderAmount needs order units only.
```

types/decimal/decimal_shopspring.go
```diff
@@ -7,6 +7,9 @@ import shopspring "github.com/shopspring/decimal" //nolint:depguard // Default b
 // Implementation identifies the selected decimal backend.
 const Implementation = "shopspring/decimal"
 
+// MaxFractionalDigits is zero because the default backend has no fixed fractional precision limit.
```
thrasher-: Heimdall: **Nit** — this is right about representation and about `Add`, `Sub` and `Mul`, but next to udecimal's "the fractional precision retained by udecimal arithmetic" it reads as covering division too. This backend's `Div` keeps *fewer* fractional digits than udecimal's: 16 (`shopspring.DivisionPrecision`) against 19.

```
                      default build          -tags udecimal_on
MaxFractionalDigits   0                      19
1 / 1048576           0.0000009536743164     0.0000009536743164062
1e-10 / 1e10          0                      0
```

`AmountStep` only multiplies, so it is unaffected. The risk is a later reader of an exported constant taking `0` to mean division is exact too.

```suggestion
// MaxFractionalDigits is zero because the default backend's Add, Sub and Mul
// never truncate. Its Div still rounds to shopspring.DivisionPrecision digits.
```

thrasher- (approved):
Heimdall: Approving. Both round-two findings are fixed. Across 200,000 random pairs under `-tags udecimal_on`, no inexact base increment is accepted, where [`93c3c93a6`](https://github.com/thrasher-corp/gocryptotrader/commit/93c3c93a6e19e006c4c81d19582ac25fa369192e) accepted 39,161. Every mutation that weakens the new check fails `TestAmountStepBaseIncrementTruncation` under that tag.

Inline, none blocking: a Minor on [amount_step.go:45](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385858), and Nits on [amount_step.go:47](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385864) · [amount_step.go:26](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385881) · [decimal_shopspring.go:10](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385900).

The red ubuntu job is ambient. Huobi's `api.hbdm.com` timed out from the runner, both touched packages passed in that job, and the failed Huobi tests pass on this branch today.

Outside this diff, optional here or as a separate issue, both found while checking what `futures.Contract.Multiplier` would feed `AmountStep`:
- OKX reports every inverse swap as linear, because swaps have an empty `baseCcy` — [okx_wrapper.go:2853-2856](https://github.com/thrasher-corp/gocryptotrader/blob/3d114e70d5fa96404a945d98b88527300eddc713/exchanges/okx/okx_wrapper.go#L2853-L2856)
- Bybit fills `Multiplier` from the leverage step — [bybit_wrapper.go:1995](https://github.com/thrasher-corp/gocryptotrader/blob/3d114e70d5fa96404a945d98b88527300eddc713/exchanges/bybit/bybit_wrapper.go#L1995)

exchange/order/limits/amount_step.go
thrasher-: Heimdall: Fixed. Across 200,000 random increment/multiplier pairs under `-tags udecimal_on`, none now comes back inexact and none exact is rejected, where [`93c3c93a6`](https://github.com/thrasher-corp/gocryptotrader/commit/93c3c93a6e19e006c4c81d19582ac25fa369192e) accepted 39,161 inexact. Floor, ceil and `CommonBaseIncrement` are exact on every accepted step, and the default build is exact at both commits.

One correction to what I wrote above: gating on the combined scale is close to free on the default build, but not under `udecimal_on`. There the two `String()` calls run on every call, taking `BaseIncrement` from 19.9 to 118.3 ns and from 0 to 2 allocations. An allocation-free alternative is on [amount_step.go:47](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385864).

thrasher- (approved):
Heimdall: Approving [`c5d8500c7`](https://github.com/thrasher-corp/gocryptotrader/commit/c5d8500c716385000b11fbd885ae310a6c5c085e). All four landed, and I've resolved each thread: [the sentinel](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385858), [the `Div` round-trip](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385864), [the multiplier's units](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385881) and [the division doc](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4079385900).

I ran the same 400,000 increment/multiplier pairs under `-tags udecimal_on` at both commits. The round-trip accepts exactly what the `big.Rat` check did: 237,107, all exact. The 57,608 underflows that [`3d114e70d`](https://github.com/thrasher-corp/gocryptotrader/commit/3d114e70d5fa96404a945d98b88527300eddc713) reported as `ErrAmountStepNotPositive` now carry `ErrBaseIncrementNotRepresentable`. `BaseIncrement` no longer allocates (116.9 → 59.4 ns), and deleting, disabling or weakening the check fails both regression tests.

Nothing new, and CI is green.

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,161 @@
...
+	return result, nil
+}
+
+func greatestCommonDivisor(first, second *big.Int) *big.Int {
+	var result big.Int
+	result.GCD(nil, nil, first, second)
+	return &result
+}
+
+func decimalScale(value string) int {
```
gloriousCode: rn to `fractionalDigits`

exchange/order/limits/amount_step.go
```diff
@@ -0,0 +1,161 @@
...
+	point := strings.IndexByte(value, '.')
+	if point == -1 {
+		return 0
+	}
+	return len(value) - point - 1
+}
+
+func (a AmountStep) orderIncrement() (decimal.Decimal, error) {
```
gloriousCode: Add basic function tests for each.

gloriousCode (changes requested):
# 🍑

exchange/order/limits/amount_step_test.go
```diff
@@ -0,0 +1,352 @@
...
+			assert.Equal(t, tests[i].expected, result.String(), "CommonBaseIncrement should return the expected increment")
+
+			reverse, err := tests[i].second.CommonBaseIncrement(tests[i].first)
+			require.NoError(t, err, "CommonBaseIncrement must be symmetric")
+			assert.True(t, result.Equal(reverse), "CommonBaseIncrement should be symmetric")
+		})
+	}
+}
+
+func BenchmarkAmountStepCommonBaseIncrement(b *testing.B) {
```
thrasher-: Heimdall: **Minor** — `errCannotParseBaseIncrement` has no direct `ErrorIs` coverage, yet the default build reaches both of its returns. Each `!ok` check in [`amount_step.go:119-126`](https://github.com/thrasher-corp/gocryptotrader/blob/0d7123b8e97b4b22eefeb5abaef086bf7bab694b/exchange/order/limits/amount_step.go#L119-L126) is all that stands between that input and a nil `*big.Rat`.

**Fix:** add this test. Each operand order reaches one of the two checks:

```suggestion
func TestAmountStepCommonBaseIncrementUnparseable(t *testing.T) {
	t.Parallel()
	if decimal.MaxFractionalDigits > 0 {
		t.Skip("big.Rat rejects more than 1e6 fractional digits, which udecimal cannot represent")
	}
	parseable := AmountStep{Increment: decimal.MustFromString("0.2"), ContractMultiplier: decimal.NewFromInt(1)}
	unparseable := AmountStep{Increment: decimal.MustFromString("3e-1000001"), ContractMultiplier: decimal.NewFromInt(1)}
	_, err := unparseable.CommonBaseIncrement(parseable)
	assert.ErrorIs(t, err, errCannotParseBaseIncrement, "CommonBaseIncrement should reject a first increment big.Rat cannot parse")
	_, err = parseable.CommonBaseIncrement(unparseable)
	assert.ErrorIs(t, err, errCannotParseBaseIncrement, "CommonBaseIncrement should reject a second increment big.Rat cannot parse")
}

func BenchmarkAmountStepCommonBaseIncrement(b *testing.B) {
```

**Why:** `big.Rat.SetString` refuses a decimal with more than 1,000,000 fractional digits ([`ratconv.go:181-183`](https://github.com/golang/go/blob/go1.27.0/src/math/big/ratconv.go#L181-L183)), and shopspring holds one without complaint. `0.2` and `3e-1000001` are not multiples of each other, so the `Mod` shortcut at [`amount_step.go:110-115`](https://github.com/thrasher-corp/gocryptotrader/blob/0d7123b8e97b4b22eefeb5abaef086bf7bab694b/exchange/order/limits/amount_step.go#L110-L115) is skipped and parsing is reached:

```
                               this head                        with the test
delete the first !ok check     suite green on both backends     FAIL: nil pointer dereference
delete the second !ok check    suite green on both backends     FAIL: nil pointer dereference
```

[`CODING_GUIDELINES.md:208`](https://github.com/thrasher-corp/gocryptotrader/blob/4354537e9817d8404c9c38c0f5bdfab2dfe173b1/docs/CODING_GUIDELINES.md?plain=1#L208) asks for direct coverage of every sentinel return. The test skips under `udecimal_on`, which caps values at 19 fractional digits, and runs in 0.47s under `-race` on the default build. Lint is clean on both tags, gofumpt and misc checks pass, and it applies cleanly alongside [the dedup suggestion on line 87](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4093575010).

exchange/order/limits/amount_step_test.go
```diff
@@ -0,0 +1,352 @@
...
+		assert.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "CeilBaseAmount should reject a truncated increment")
+		_, err = step.CommonBaseIncrement(unit)
+		assert.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "CommonBaseIncrement should reject a truncated first increment")
+		_, err = unit.CommonBaseIncrement(step)
+		assert.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "CommonBaseIncrement should reject a truncated second increment")
+		return
+	}
+	require.NoError(t, err, "BaseIncrement must accept an exactly representable product")
+	assert.Equal(t, "0.00000000000000000015", result.String(), "BaseIncrement should retain the exact product")
+}
```
thrasher-: Heimdall: **Nit** — these two tests repeat the same block of six assertions, and [`CODING_GUIDELINES.md:233`](https://github.com/thrasher-corp/gocryptotrader/blob/4354537e9817d8404c9c38c0f5bdfab2dfe173b1/docs/CODING_GUIDELINES.md?plain=1#L233) makes deduplication the default.

**Fix:** fold them into one table:

```suggestion
func TestAmountStepBaseIncrementInexactProduct(t *testing.T) {
	t.Parallel()
	unit := AmountStep{Increment: decimal.NewFromInt(1), ContractMultiplier: decimal.NewFromInt(1)}
	for _, tc := range []struct {
		name       string
		increment  string
		multiplier string
		exact      string
	}{
		{name: "underflow", increment: "0.0000000001", multiplier: "0.0000000001", exact: "0.00000000000000000001"},
		{name: "truncation", increment: "0.0000000003", multiplier: "0.0000000005", exact: "0.00000000000000000015"},
	} {
		t.Run(tc.name, func(t *testing.T) {
			t.Parallel()
			step := AmountStep{
				Increment:          decimal.MustFromString(tc.increment),
				ContractMultiplier: decimal.MustFromString(tc.multiplier),
			}
			result, err := step.BaseIncrement()
			if decimal.MaxFractionalDigits == 0 {
				require.NoError(t, err, "BaseIncrement must accept an exactly representable product")
				assert.Equal(t, tc.exact, result.String(), "BaseIncrement should retain the exact product")
				return
			}
			require.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "BaseIncrement must reject an inexact product")
			assert.True(t, result.IsZero(), "BaseIncrement should return zero for an inexact product")
			_, err = step.FloorBaseAmount(decimal.NewFromInt(1))
			assert.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "FloorBaseAmount should reject an inexact increment")
			_, err = step.CeilBaseAmount(decimal.NewFromInt(1))
			assert.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "CeilBaseAmount should reject an inexact increment")
			_, err = step.CommonBaseIncrement(unit)
			assert.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "CommonBaseIncrement should reject an inexact first increment")
			_, err = unit.CommonBaseIncrement(step)
			assert.ErrorIs(t, err, ErrBaseIncrementNotRepresentable, "CommonBaseIncrement should reject an inexact second increment")
		})
	}
}
```

**Why:** the inputs and checks are the same, in 15 fewer lines. I mutated `BaseIncrement` five ways: dropped the exactness check, weakened it to `IsPositive`, inverted its backend gate, doubled the returned value, and swapped the sentinel. Each row kills exactly what its original test killed, on both backends. The branch moves from `expectedProduct.IsZero()` to `decimal.MaxFractionalDigits == 0`, which picks the same leg on both backends for these inputs. The underflow row now checks the default build against a literal, where the original recomputed the same `Mul` the method runs. `-race` and lint are clean on both tags, and gofumpt and misc checks pass. Optional.

thrasher- (approved):
Heimdall: Approving [`0d7123b8e`](https://github.com/thrasher-corp/gocryptotrader/commit/0d7123b8e97b4b22eefeb5abaef086bf7bab694b). Both of gloriousCode's asks landed: the rename is complete, and the new helper tests kill all seven mutations I tried, two of which the previous suite missed. Two non-blocking suggestions are inline ([sentinel coverage](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4093574975), [test dedup](https://github.com/thrasher-corp/gocryptotrader/pull/2403#discussion_r4093575010)). The red arm job is a live Deribit failure in a package this PR doesn't touch, and those tests pass on master and on this branch today.

PR #2272, opened 2026-06-20
exchanges/lbank: add WebSocket support for ticker, trades and orderbook

PARMESHWARPANWAR:
# PR Description

Adds WebSocket support for LBank exchange using wss://www.lbkex.net/ws/V2/

This completes the final remaining task from #1925 (the previous PR #2256 addressed fmt.Sprintf cleanup, v2 endpoint updates, and generic error capture).

Fixes # (issue)

## Type of change

Please delete options that are not relevant and add an `x` in `[]` as item is complete.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] This change requires a documentation update

## How has this been tested

Please describe the tests that you ran to verify your changes. Provide instructions so we can reproduce. Please also list any relevant details for your test configuration and
also consider improving test coverage whilst working on a certain feature or package.

- [ ] go test ./... -race
- [ ] golangci-lint run
- [ ] Test X

## Checklist

- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation and regenerated documentation via the documentation tool
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally and on Github Actions with my changes
- [ ] Any dependent changes have been merged and published in downstream modules

PR #2404, opened 2026-09-22
lbank: accept boolean response envelope results and fix the silent failed-request path

### LINKED ISSUE #1925: LBank: Update endpoints to v2 and other improvements

Robin1987China:
# PR Description

LBank reports the `result` member of its response envelope as a JSON string
(`"true"`/`"false"`) on some endpoints and as a JSON boolean on others. The
exchange decodes `V2Response.Result` into a Go `string`, so any response
carrying a boolean fails to unmarshal. Because the envelope check was guarded by
`if err := json.Unmarshal(...); err == nil`, that failure silently skipped the
failed-request check and fell through to parsing the envelope itself as the
payload — a failed request could be reported as a successful one, and a genuine
failure never reached the caller as an error.

PR #2272 covers the same shape problem for `ErrCapture.Result`; this PR covers
the other half, `V2Response.Result`, fixes the silent-failure path and adds
tests. The two changes are disjoint: this PR does not touch `ErrCapture`.

Fixes #1925 (partial)

## Type of change

## What changed

- `V2Response.Result` is now `*types.Boolean` instead of `string`.
  `types.Boolean` accepts both the string and the boolean representation, which
  is the same approach the coding guidelines prescribe for fields that may
  arrive in either form.
- The pointer keeps an absent or `null` `result` distinct from an explicit
  `false`, so a missing member is not misread as a request failure.
- The envelope handling moved into `unwrapV2Response`, which returns the inner
  payload, the unchanged payload for non-envelope responses, or the exchange
  failure. `SendHTTPRequest` now just unmarshals what it returns. The error
  message and every other behaviour are unchanged; the new function exists so
  the envelope logic is unit testable without a live endpoint.
- Removed the unused `lbankAPIVersion1` constant; every REST path already builds
  on `lbankAPIVersion2` (22 references, 0 to the v1 constant).
- Documented the four public endpoint paths that are no longer listed for the v2
  API — or are listed as abandoned — so their status is not silently lost.
  Their routes are deliberately left alone: the documented replacements return
  different payloads and cannot be adopted without recorded fixtures.
- `ErrCapture` is intentionally untouched; see the note above.

## How has this been tested

- [ ] go test ./... -race
- [ ] golangci-lint run
- [x] go test ./exchanges/lbank/ -run TestUnwrapV2Response -count=1

`TestUnwrapV2Response` crosses both result representations against the success,
failure, absent, null, no-data and non-envelope cases (8 sub-tests, all pass).
It fails against the previous implementation: restoring `Result` to a `string`
together with the previous `err == nil` guard makes the boolean cases fail —
the boolean failure case returns no error at all ("an error is expected but got
nil") because the unmarshal failure skips the check, and the boolean success
case returns the whole envelope instead of `data`. Both were verified by
temporarily restoring the previous code and re-running before restoring the fix.

## Checklist


exchanges/lbank/lbank.go
```diff
@@ -512,16 +517,31 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path st
...
 	var v2Resp V2Response
-	if err := json.Unmarshal(tempResp, &v2Resp); err == nil {
-		if v2Resp.Result == "false" {
-			return fmt.Errorf("lbank: request failed: %s", v2Resp.Msg)
-		}
-		if v2Resp.Data != nil {
-			return json.Unmarshal(v2Resp.Data, result)
-		}
+	if err := json.Unmarshal(payload, &v2Resp); err != nil {
+		return payload, nil //nolint:nilerr // not an envelope; return it unchanged for the caller to unmarshal
```
thrasher-: Heimdall: **Minor** — the envelope guard is unchanged in kind. "The struct decoded cleanly" is still the test for "this is an envelope", so a shape surprise in any *other* part of the envelope reproduces the bug this PR fixes, one field over — and for one input class the branch is now worse than the base.

**Fix:** only fall through when the payload is not a JSON object, and surface the decode error otherwise:

```go
if err := json.Unmarshal(payload, &v2Resp); err != nil {
	if bytes.HasPrefix(bytes.TrimLeft(payload, " \t\r\n"), []byte("{")) {
		return nil, fmt.Errorf("lbank: decoding response envelope: %w", err)
	}
	return payload, nil
}
```

All nine public paths this package builds return an envelope object, so nothing currently relies on a JSON object passing through untouched. A narrower variant — decode `result` into a `json.RawMessage` and reject an unrecognised value — closes the `result` case alone and leaves the rest as it is.

**Why:** `unwrapV2Response` swallows *every* decode failure, not just the ones a non-envelope payload causes. Served through `httptest` and called via `SendHTTPRequest`, then again through `GetMarketDepths`, against [`158e9c12e`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322) and this branch:

```
body                                                base            branch
{"result":"false","msg":"x","error_code":10003}     request failed  request failed
{"result":"false","msg":"x","error_code":10003} request failed  err=nil, empty book
{"result":"false","msg":"x","error_code":"10003"}   err=nil         err=nil
{"result":"false","msg":123,"error_code":10003}     err=nil         err=nil
{"result":"success","data":[1,2,3]}                 [1,2,3]         err=nil, whole envelope
```

Row 2 is the regression: `"false"` is a valid JSON spelling of `false`, and [boolean.go:11-19](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/types/boolean.go#L11-L19) switches on the raw token bytes, so it no longer matches where `string` decoded the escape. Rows 3 and 4 are pre-existing and untouched; both are failed requests reported as successes, which is the silent path the PR description sets out to remove. Row 5 is a narrowing — eight accepted tokens where `string` accepted anything. Identical under `-tags sonic_on`.

I could not get LBank to emit any of those shapes: across the nine paths this package builds plus the `/v2/supplement/*` endpoints, every error envelope came back with a bare numeric `error_code` and an unescaped string `msg`. So this is hardening, not a live defect — if you would rather leave the fallback alone, say so and I will drop it.

exchanges/lbank/lbank.go
```diff
@@ -512,16 +517,31 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path st
...
+	}
+	if v2Resp.Result != nil && !v2Resp.Result.Bool() {
+		return nil, fmt.Errorf("lbank: request failed: %s", v2Resp.Msg)
```
thrasher-: Heimdall: **Minor** — this path now actually fires, and when it does it drops the only machine-readable part of the response.

**Fix:** carry `error_code`, and make the failure a sentinel so callers and tests can match the category rather than the wording:

```go
var errRequestFailed = errors.New("lbank: request failed")
...
return nil, fmt.Errorf("%w: %s (error_code %d)", errRequestFailed, v2Resp.Msg, v2Resp.ErrorCode)
```

**Why:** `V2Response.ErrorCode` is decoded and never read, while [lbank.go:487](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/lbank/lbank.go#L487) already turns exactly these codes into text for the authenticated path, off the map at [lbank_types.go:234](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/lbank/lbank_types.go#L234). Every failed envelope I pulled from the live API carried one:

```
/v2/trades.do?symbol=eth_btc   {"result":"false","msg":"currency pair nonsupport","error_code":10008,...}
/v2/user_info.do               {"result":"false","msg":"Illegal parameter","error_code":10003,...}
```

`msg` is free prose and the code is the stable identifier, so a caller wanting to branch on "currency pair nonsupport" has to string-match today. Note this loss is pre-existing for string-result failures — it is just newly reachable for the boolean form.

On the sentinel: [CODING_GUIDELINES.md](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/docs/CODING_GUIDELINES.md#error-handling) prefers `errors.Is`-matchable errors on parsing and validation paths — not a rule this breaks, since there is no sentinel here yet, but the package already has a block at [lbank.go:76-80](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/lbank/lbank.go#L76-L80) and both new tests currently pin this error by its text — exact at [lbank_test.go:562](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/lbank/lbank_test.go#L562), substring at [lbank_test.go:591](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/lbank/lbank_test.go#L591). Adding `error_code` to the message would break the first of those; a sentinel plus `ErrorIs` would not.

One consequence worth knowing either way: with `msg` absent the message degrades to `lbank: request failed: ` — measured on this branch against `{"result":false,"error_code":10007,"ts":1568796039610}`. I found no live LBank response with a false `result` and no `msg`, so that half is synthetic; the dropped code is not.

thrasher- (approved):
Heimdall: Approving. Both new tests are real regression tests — restore the base envelope logic and `TestUnwrapV2Response/boolean_result_{true,false}` plus `TestSendHTTPRequestEnvelopeFailure` fail; restore the fix and they pass. The premise holds live: `/v2/supplement/incrDepth.do` returns `{"result":false,...}` and `/v2/supplement/ticker/price.do` returns `{"result":true,...}`, bare JSON booleans both. Against [`158e9c12e`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322), `{"result":false,"msg":"instrument not found"}` went from `err=nil` with the envelope leaking into the caller's result to a proper error. All nine public paths this package builds still return the string form, so the built-in wrappers are unchanged — `SendHTTPRequest` is exported, though, so a caller aimed at `/v2/supplement/*` does see the difference. `lbankAPIVersion1` had zero references at the base. Lint, `-race -count 2`, `sonic_on`, `sonic_on udecimal` and `mock_test_off` all clean, and [`a02654aa8`](https://github.com/thrasher-corp/gocryptotrader/commit/a02654aa8ba494601f3c38e81008f9f6471ad0a9) cleared misc-checks.

Two comments on the lines — [the envelope guard](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4068447236) and [the error detail](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4068447239) — with the evidence on each. Neither is a condition on the merge.

Optional, out of scope and pre-existing — say the word and I will raise issues:

- [lbank.go:590](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/lbank/lbank.go#L590) — `SendAuthHTTPRequest` never prepends the configured host, so every authenticated wrapper dies before it leaves the process: `GetUserInfo` against a local server gives `Post "/v2/user_info.do": unsupported protocol scheme ""` with zero hits on the server, while the same call with an absolute URL succeeds. #2272 already fixes it — but it does mean the `ErrCapture.Result` half you point at is unreachable until that lands, which is likely why it has gone unnoticed.
- [lbank.go:404](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/lbank/lbank.go#L404) — `USD2RMBRate` always returns 0: `rate.Float64()` is a method call, so it is evaluated before the `SendHTTPRequest` that fills `rate`. Live the endpoint returns `6.6951`; the method returns `0`. Same shape at [kucoin_futures.go:618](https://github.com/thrasher-corp/gocryptotrader/blob/a02654aa8ba494601f3c38e81008f9f6471ad0a9/exchanges/kucoin/kucoin_futures.go#L618).
- #2272 also removes `lbankAPIVersion1` and rewrites the same const block, so expect a conflict there. The split you describe in the PR body is real — I checked its diff, it does move `ErrCapture.Result` to `types.Boolean`.

Take the two comments or leave them — nothing here blocks.

thrasher-:
@codex please review

exchanges/lbank/lbank.go
```diff
@@ -512,16 +521,36 @@ func (e *Exchange) SendHTTPRequest(ctx context.Context, ep exchange.URL, path st
...
+func unwrapV2Response(payload json.RawMessage) (json.RawMessage, error) {
 	var v2Resp V2Response
-	if err := json.Unmarshal(tempResp, &v2Resp); err == nil {
-		if v2Resp.Result == "false" {
-			return fmt.Errorf("lbank: request failed: %s", v2Resp.Msg)
-		}
-		if v2Resp.Data != nil {
-			return json.Unmarshal(v2Resp.Data, result)
+	if err := json.Unmarshal(payload, &v2Resp); err != nil {
+		if bytes.HasPrefix(bytes.TrimLeft(payload, " \t\r\n"), []byte("{")) {
```
thrasher-, line 544: Heimdall: **Verified, no action** — this is the line my [round-1 comment](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4068447236) asked for, and it earned its place: it closed two silent holes that the round-1 code had opened. Through `GetMarketDepths` over `httptest`, identical under `-tags sonic_on`:

```
body                                          158e9c12e        a02654aa8        6efd857b9
{"result":true,"data":{..}}                   nil, empty book  data             data
{"result":false,"msg":..,"error_code":10008}  nil, empty book  request failed   request failed (error_code 10008)
{"result":"\u0066alse","msg":..}              request failed   nil, empty book  decoding response envelope
{"result":"\u0074rue","data":{..}}            data             nil, empty book  decoding response envelope
{"result":"false","error_code":"10008"}       nil              nil              decoding response envelope
{"result":null,"data":{..}} / {"data":{..}}   data             data             data
```

Rows three and four are the regression it repairs. `types.Boolean` switches on the raw token, so a JSON-escaped `false` or `true` stops matching where `string` decoded the escape, and at `a02654aa8` that fell straight through to the caller with an empty book and no error. Every divergence from the base in that table now turns a right answer into a loud error rather than a silent one.

Worth knowing about the two new table rows that feed this branch a non-object, `non json payload is unchanged` and `truncated object is an error`: [`request.go:239`](https://github.com/thrasher-corp/gocryptotrader/blob/6efd857b9c7005d6f172137fe6173b5954f83490/exchanges/request/request.go#L239) decodes the body into the `json.RawMessage` before `unwrapV2Response` ever sees it, so only syntactically valid JSON reaches here. Both exercise the helper directly and cannot arrive through `SendHTTPRequest` — fine as unit coverage, just not endpoint-reachable.

exchanges/lbank/lbank_types.go
```diff
@@ -220,8 +220,12 @@ type ErrCapture struct {
 }
 
 // V2Response wraps all LBank v2 API responses
+//
+// result is reported as a JSON boolean by some endpoints and as a JSON string by
+// others, so it is decoded as a types.Boolean, which accepts both. A pointer is
+// used so that an absent or null result stays distinct from an explicit false.
 type V2Response struct {
-	Result    string          `json:"result"`
+	Result    *types.Boolean  `json:"result"`
```
thrasher-, line 231: Heimdall: **Verified, no action** — both halves of this line check out live, including the pointer, which I had read as belt-and-braces last round.

`GET /v2/supplement/ticker/price.do?symbol=btc_usdt` returns `"result":true` and `GET /v2/supplement/incrDepth.do` on an unsupported symbol returns `"result":false`, bare booleans both, while all nine public paths this package builds return the string form. And `GET /v2/supplement/incrDepth.do?symbol=btc_usdt&size=3` returns an object carrying **only** `data`, with no `result` member at all — so a non-pointer `types.Boolean` would read that success as a failure. The pointer is doing real work.

On the other three fields: I censused the 48 `/v2/*.do` paths named in the [docs](https://www.lbank.com/docs/index.html) plus `incrDepth.do` and `usdToCny.do`. `result` came back only as `"true"`, `"false"`, bare `true` or absent; `msg` only as a string or absent; `error_code` only as a bare number or absent. The documented examples agree — 19 x `"true"`, 7 x `true`, no third spelling. That is one response per route rather than a proof about every state, but it is the whole documented surface, and it is why I am not asking for `types.Number` on `error_code`.

One sharp edge of the pointer, for the record, and it is synthetic: a duplicated `result` whose last value is `null` — `{"result":"false","result":null,"data":{..}}` — reports the failure at the base and returns populated data with `err == nil` here, because `null` clears a pointer where it was a no-op for a `string`. I have no evidence LBank emits duplicate keys and would not act on it.

exchanges/lbank/lbank.go
```diff
@@ -38,12 +38,17 @@ type Exchange struct {
...
 	lbankFeeNotFound = 0.0
 	tradeBaseURL     = "https://www.lbank.com/trade/"
 
 	// Public endpoints
+	//
+	// lbankTicker, lbankTrades, lbankUSD2CNYRate and lbankWithdrawConfig are no
+	// longer listed for the v2 API, or are listed there as abandoned. They are
+	// left on their current paths because the documented replacements return
+	// different payloads and cannot be adopted without recorded fixtures.
+	// See https://github.com/thrasher-corp/gocryptotrader/issues/1925
```
thrasher-, line 51: Heimdall: **Verified, no action** — the four statuses hold against the [live docs](https://www.lbank.com/docs/index.html) as fetched today. `withdrawConfigs.do` sits under *Abandoned Endpoints*, "Not recommended for use, it will be taken down later", with `GET /v2/assetConfigs.do` named as the replacement. `ticker.do` survives only in the changelog, where the replacement is `GET /v2/ticker/24hr.do`. `trades.do` and `usdToCny.do` have no entry at all — only `/v2/supplement/trades.do` exists. The endpoints you left uncommented — `depth.do`, `accuracy.do`, `currencyPairs.do` — are all still documented, so the list is exactly right.

`lbankAPIVersion1` had exactly one reference in the repository at the base commit: its own declaration. Dropping it is safe.

exchanges/lbank/lbank_test.go
```diff
@@ -502,3 +504,124 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+			payload: `{"result":false,"msg":"Invalid parameter","error_code":10003}`,
+			err:     "lbank: request failed: Invalid parameter (error_code 10003)",
+			isErr:   errRequestFailed,
+		},
+		{
+			name:    "failed envelope without error code",
+			payload: `{"result":false,"msg":"instrument not found"}`,
+			err:     "lbank: request failed: instrument not found (error_code 0)",
+			isErr:   errRequestFailed,
+		},
```
thrasher-, line 550: Heimdall: **Should fix** — no row crosses a false `result` with a present `data`, so the one invariant this PR exists to enforce, that a failure beats a payload, is not pinned by anything.

**Fix:**

```suggestion
		},
		{
			name:    "failed envelope carrying data is still a failure",
			payload: `{"result":false,"msg":"Invalid parameter","error_code":10003,"data":[1,2,3]}`,
			err:     "lbank: request failed: Invalid parameter (error_code 10003)",
			isErr:   errRequestFailed,
		},
```

**Why:** the table is otherwise strong — restore the base envelope logic and six of its sub-tests fail, along with `TestSendHTTPRequestEnvelopeFailure`. But of fifteen mutations of `unwrapV2Response` I ran, five survive the whole package suite, and this is the one that matters: swap the two branches so that [returning `data`](https://github.com/thrasher-corp/gocryptotrader/blob/6efd857b9c7005d6f172137fe6173b5954f83490/exchanges/lbank/lbank.go#L550) comes before [the failure check](https://github.com/thrasher-corp/gocryptotrader/blob/6efd857b9c7005d6f172137fe6173b5954f83490/exchanges/lbank/lbank.go#L547) and everything still passes — `-race`, `-tags sonic_on`, the wrapper standards, all of it. That mutation hands `data` back with `err == nil` for a failed request, which is the shape this PR exists to remove. With the row above in place it fails immediately, on `failed_envelope_carrying_data_is_still_a_failure`. `"data":null` is no protection either: a JSON `null` is four bytes, not a nil `json.RawMessage`, so it takes the same branch.

The other four survivors I looked at and would leave alone — requiring a non-empty `msg`, dropping `\r` from the trim set, treating `"data":null` as absent, `%w` to `%v` on the decode error — are unobservable at the boundary or need inputs LBank does not emit. This one is not: [CODING_GUIDELINES.md](https://github.com/thrasher-corp/gocryptotrader/blob/6efd857b9c7005d6f172137fe6173b5954f83490/docs/CODING_GUIDELINES.md#L215) asks to "cross relevant input dimensions ... explicit, omitted or `null` values ... where they can affect runtime behaviour", and the table already crosses the two `result` spellings against true, false, absent and null. Falsity against a present `data` is the one pair left open. Not a live defect — every failed envelope I pulled across the 50 paths I probed carried no `data`.

thrasher- (approved):
Heimdall: Re-checked at [`6efd857b9`](https://github.com/thrasher-corp/gocryptotrader/commit/6efd857b9c7005d6f172137fe6173b5954f83490). Both comments are applied verbatim, and the [envelope guard](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4068447236) mattered more than I had it scored. The evidence is on the lines: [the guard](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4071483704), [the type](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4071483710) and [the const comment](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4071483715). One [Should fix](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4071483721), a missing test row — I would like it, but I am not holding the merge for it.

CI is green on all 16 checks; locally `-race`, `-tags sonic_on`, `gofumpt`, `misc_checks` and `golangci-lint 2.13.2` over the merge ref are all clean, `TestAllExchangeWrappers/lbank` runs the same 85 sub-tests to the same outcome here and on master, `unwrapV2Response` is at 100% coverage and `SendHTTPRequest` goes 78.9% to 87.5%, master merges in conflict-free, and `cmd/documentation` regenerates the same two pre-existing drift files on both.

Housekeeping: the description still says "the error message and every other behaviour are unchanged" and "8 sub-tests", where the message now carries `(error_code %d)` and there are thirteen; and [`46f6cb6e7`](https://github.com/thrasher-corp/gocryptotrader/commit/46f6cb6e78df91137e690bea621ec6c4bb8a2d47)'s subject describes the `nolint:nilerr` that `6efd857b9` took back out, which a subject-built squash message would carry.

Still unfiled from last round: the missing host prefix in `SendAuthHTTPRequest` ([`lbank.go:599`](https://github.com/thrasher-corp/gocryptotrader/blob/6efd857b9c7005d6f172137fe6173b5954f83490/exchanges/lbank/lbank.go#L599)) and `USD2RMBRate` returning 0 ([`lbank.go:408`](https://github.com/thrasher-corp/gocryptotrader/blob/6efd857b9c7005d6f172137fe6173b5954f83490/exchanges/lbank/lbank.go#L408)). Say the word and I will raise them.

exchanges/lbank/lbank.go
```diff
@@ -72,6 +77,10 @@ var (
 	errPEMBlockIsNil           = errors.New("pem block is nil")
 	errUnableToParsePrivateKey = errors.New("unable to parse private key")
 	errPrivateKeyNotLoaded     = errors.New("private key not loaded")
+	// errRequestFailed is returned when an LBank v2 envelope reports a failed
+	// request. Callers and tests match this category with errors.Is instead of
+	// matching the exchange supplied message text.
+	errRequestFailed = errors.New("lbank: request failed")
```
shazbert: We can wrap the exchange name to the error instead if needed, and other errors across this PR.

```suggestion
	errRequestFailed = errors.New("request failed")
```

exchanges/lbank/lbank_types.go
shazbert, line 231: A comment directly here about the pointer usage for the field would be better so that we don't do something silly at a later date.

shazbert (changes requested):
PR looks great! Some minor things. Then it looks good to go, and there are other review request here, can you let us know what you want to do and what you reject. @Robin1987China 🚀

Robin1987China:
Thanks for the review, @shazbert. Point by point:

**Accepted**

1. `exchanges/lbank/lbank.go:83` — the exchange name is out of the error string. `errRequestFailed` is now `errors.New("request failed")`, the decode error in `unwrapV2Response` no longer carries the prefix either, and `SendHTTPRequest` wraps what the helper returns with `e.Name` (`fmt.Errorf("%s: %w", e.Name, err)`), which is where the rest of the repository adds it (`exchanges/credentials.go:116`, `exchanges/request/request.go:255`). Both were the only prefixed strings this PR introduced — `grep -rn 'lbank: ' exchanges/lbank/` is now empty apart from pre-existing code untouched by this PR.
2. `exchanges/lbank/lbank_types.go:228` — the pointer rationale now sits directly on `Result` instead of on the type, stating the concrete failure mode: several successful endpoints omit `result` and send only `data`, and a non-pointer `types.Boolean` would read those successes as failures.

**Accepted, previously implemented**

3. The `Should fix` from the last round (the missing `false` result + present `data` row) is in `ae6a5b3` — the row is verbatim Heimdall's suggestion.
4. Both round-1 comments (the envelope guard, and carrying `error_code` behind a sentinel) were implemented in `6efd857b` and are unchanged.

**Declined for this PR**

5. `SendAuthHTTPRequest` not prepending the host, and `USD2RMBRate` always returning 0 — both are pre-existing at the base and unrelated to the boolean envelope work, so they stay out of this PR. The host one is already fixed by #2272, so it stays out of this PR. For `USD2RMBRate`: Heimdall offered to file it — if that offer stands, please go ahead; otherwise say so and I will file it separately.
6. PR description and the `46f6cb6e7` subject are stale, as Heimdall noted; I will refresh the description and sort the squash subject separately rather than widening this branch.

**Commits**

- `ae6a5b3` — the test row above.
- `fb70504` — the two changes in this reply. Both are now pushed, so the branch head is `fb70504` (`6efd857b` → `fb70504`).

One thing to confirm: `e.Name` is `Lbank`, so the caller-visible string is now `Lbank: request failed: instrument not found (error_code 10076)` rather than lower-case `lbank: request failed: …`. That matches how `b.Name` is used elsewhere, but say the word if you want it lower-case.

Thanks to you both for the careful passes — the `error_code` and envelope-guard rounds both caught real problems.

exchanges/lbank/lbank_test.go
```diff
@@ -502,3 +504,132 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+			err:     "decoding response envelope:",
+		},
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			payload, err := unwrapV2Response([]byte(tc.payload))
+			if tc.err != "" {
+				assert.ErrorContains(t, err, tc.err, "unwrapV2Response should report the expected error")
+				assert.NotContains(t, err.Error(), "lbank:", "unwrapV2Response should return a bare error; SendHTTPRequest adds the exchange name")
```
thrasher-, line 598: Heimdall: **Minor** — the bare-error guard on line 598 panics on the regression this table exists to catch, and it only checks for the old lower-case `lbank:` prefix.

**Fix:**

```suggestion
				require.ErrorContains(t, err, tc.err, "unwrapV2Response must report the expected error")
				assert.Truef(t, strings.HasPrefix(err.Error(), tc.err), "unwrapV2Response should return a bare error starting %q; SendHTTPRequest adds the exchange name", tc.err)
```

**Why:** [lbank_test.go:598](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/exchanges/lbank/lbank_test.go#L598) calls `err.Error()` after a non-fatal `assert`, so a nil error dereferences. And the name is [`Lbank`](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/exchanges/lbank/lbank_wrapper.go#L32), so putting it back into the helper as the exchange spells it gets through both new tests, while callers see `Lbank: Lbank: request failed: …`. Mutating [lbank.go](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/exchanges/lbank/lbank.go#L541-L556) and running `TestUnwrapV2Response` and `TestSendHTTPRequestEnvelopeFailure`:

| mutation | this head | with the fix |
|---|---|---|
| failed envelope returns `payload, nil` | 1 row FAIL, then `nil pointer dereference` at :598 | 4 rows + `TestSendHTTPRequestEnvelopeFailure` FAIL |
| `data` returned before the `result` check | 1 row FAIL, then `nil pointer dereference` at :598 | 1 row FAIL |
| `errors.New("Lbank: request failed")` | ok | 4 rows FAIL |
| `"Lbank: decoding response envelope: %w"` | ok | 3 rows FAIL |

The first row is the silent path this PR removes. At [`ae6a5b3`](https://github.com/thrasher-corp/gocryptotrader/commit/ae6a5b3ffb2cb8d5ac03a5ad38633622bfb8221f), before this line existed, it failed cleanly on all five. A panic in a subtest takes the whole package binary down, so the rest of that run goes unreported. The fix passes unmutated under `-race -count 2` on both the default build and `-tags sonic_on`, and lints clean.

exchanges/lbank/lbank_types.go
```diff
@@ -220,8 +220,15 @@ type ErrCapture struct {
...
 // V2Response wraps all LBank v2 API responses
+//
+// result is reported as a JSON boolean by some endpoints and as a JSON string by
+// others, so it is decoded as a types.Boolean, which accepts both.
 type V2Response struct {
-	Result    string          `json:"result"`
+	// Result is a pointer so that an absent or null result stays distinct from an
+	// explicit false: several successful endpoints omit result altogether and
+	// send only data, and a non-pointer types.Boolean would read those successes
+	// as failures.
```
thrasher-, line 230: Heimdall: **Nit** — "several successful endpoints" is one, as far as I can find. Naming it keeps the claim checkable.

```suggestion
	// Result is a pointer so that an absent or null result stays distinct from an
	// explicit false: some successful endpoints, such as /v2/supplement/incrDepth.do,
	// omit result altogether and send only data, and a non-pointer types.Boolean
	// would read those successes as failures.
```

**Why:** of 16 public endpoints that answer a plain GET today — the nine paths this package builds plus `ticker/24hr.do`, `supplement/trades.do`, `assetConfigs.do`, `supplement/ticker/price.do`, `supplement/ticker/bookTicker.do`, `supplement/incrDepth.do` and `etfTicker/24hr.do` — only `/v2/supplement/incrDepth.do?symbol=btc_usdt&size=3` succeeds without `result`, answering with a `data` member and nothing else. The rationale is right either way. If you know of others, "several" stands and I will drop this.

exchanges/lbank/lbank_test.go
```diff
@@ -502,3 +504,132 @@ func TestGetCurrencyTradeURL(t *testing.T) {
...
+	sm.HandleFunc("/", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"result":false,"msg":"instrument not found","error_code":10076}`))
+	})
+	server := httptest.NewServer(sm)
+	defer server.Close()
+
+	ex := new(Exchange)
+	require.NoError(t, testexch.Setup(ex), "Setup must not error")
+	require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL), "SetRunningURL must not error")
```
thrasher-, line 628: Heimdall: **Nit** — this is the only test function left on `httptest.NewServer`. [#2341](https://github.com/thrasher-corp/gocryptotrader/pull/2341) moved the rest of the codebase to `httptest.NewTestServer`, which takes `t`, cleans up after itself and serves over an in-memory network.

```suggestion
	server := httptest.NewTestServer(t, sm)

	ex := new(Exchange)
	require.NoError(t, testexch.Setup(ex), "Setup must not error")
	require.NoError(t, ex.SetHTTPClient(server.Client()), "SetHTTPClient must not error")
	require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL), "SetRunningURL must not error")
```

**Why:** there are 45 `NewTestServer` calls against this one, and the other `NewServer` left is in [request_test.go's `TestMain`](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/exchanges/request/request_test.go#L206), which has no `t` to pass. The `SetHTTPClient` line is not optional, as in [bitfinex_test.go:690](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/exchanges/bitfinex/bitfinex_test.go#L690): `server.URL` stays empty until `server.Client()` starts the in-memory server, so without it `SetRunningURL` fails with `empty url`. With the change the test passes under `-race -count 3` on both the default build and `-tags sonic_on`, and still fails when the failed envelope is returned as a success.

exchanges/lbank/lbank_types.go
```diff
@@ -220,8 +220,15 @@ type ErrCapture struct {
...
+	Result    *types.Boolean  `json:"result"`
 	Msg       string          `json:"msg"`
```
thrasher-, line 232: Heimdall: **Nit** — `Msg` is the API's short alias, and with the struct already changing shape here, this is the moment to spell it out.

**Fix:** rename the field to `Message`, keeping `json:"msg"`, along with its only read at [lbank.go:550](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/exchanges/lbank/lbank.go#L550).

**Why:** [CODING_GUIDELINES.md](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/docs/CODING_GUIDELINES.md#L44) asks for full and descriptive field names and to "avoid short API-provided aliases unless compatibility requires it". Compatibility doesn't hold it back here, because the `Result` change already breaks anyone reading this exported type. The rename builds, passes and stays `gofumpt`-clean with the padding as it is.

thrasher- (approved):
Heimdall: Approving. Round 2's missing row landed in [`ae6a5b3`](https://github.com/thrasher-corp/gocryptotrader/commit/ae6a5b3ffb2cb8d5ac03a5ad38633622bfb8221f) and does its job: return `data` before the `result` check and `failed_envelope_carrying_data_is_still_a_failure` fails, where at [`6efd857b9`](https://github.com/thrasher-corp/gocryptotrader/commit/6efd857b9c7005d6f172137fe6173b5954f83490) the whole suite passed. The name now sits in the wrapping layer: live, `GetTrades` on the unlisted `eth_btc` goes from `lbank: request failed: currency pair nonsupport` at the base to `Lbank: request failed: currency pair nonsupport (error_code 10008)`, with one prefix through the wrappers.

Inline, none blocking: a Minor on [lbank_test.go:597](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4088303559), where the new bare-error guard panics instead of failing and misses the name as the exchange spells it, and Nits on [lbank_types.go:227](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4088303572), [lbank_test.go:623](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4088303582) and [lbank_types.go:232](https://github.com/thrasher-corp/gocryptotrader/pull/2404#discussion_r4088303594).

On `USD2RMBRate`: no issue needed. Your #2406 already fixes the lbank site alongside the kucoin one (its description still says the lbank one is being fixed separately), and the two branches merge cleanly.

The four red checks are ambient. Each GoCryptoTrader job timed out on a different `chainflyer.bitflyer.jp` test in `exchanges/bitflyer` ([arm64](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35850378912/job/107146377325), [386](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35850378912/job/107146377655), [sonic](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35850378912/job/107146377531)), and those tests pass at this head now; lbank passed in all three. codecov/project is comparing four uploads against the base's seven, since [tests.yml:137](https://github.com/thrasher-corp/gocryptotrader/blob/fb70504f65b495a4bc92a691b68b5ce077faefa2/.github/workflows/tests.yml#L137-L140) skips the upload when a job fails. A re-run should clear them.

shazbert (approved):
Thank changes ACK, there are some other review things that should harden everything up here. Thanks @Robin1987China

PR #2405, opened 2026-09-22
lbank: page through order history instead of skipping pages

Robin1987China:
# PR Description

`lbank.Exchange.GetOrderHistory` walks the exchange's `orders_info_history.do`
endpoint by incrementing a page counter, but that counter is incremented inside
the per-order loop rather than once per page:

```go
b := int64(1)
tempResp, err := e.QueryOrderHistory(ctx, fPair.String(), strconv.FormatInt(b, 10), "200")
if err != nil {
	return finalResp, err
}
for len(tempResp.Orders) != 0 {
	tempResp, err = e.QueryOrderHistory(ctx, fPair.String(), strconv.FormatInt(b, 10), "200")
	if err != nil {
		return finalResp, err
	}
	for x := range tempResp.Orders {
		...
		b++ // advances once per order, not once per page
	}
}

### CODE REVIEW

**Comment by thrasher-** [comment]
@codex please review

File: exchanges/lbank/lbank_wrapper.go [hunk 1]
```diff
@@ -637,18 +637,16 @@ func (e *Exchange) GetOrderHistory(ctx context.Context, getOrdersRequest *order.
...
 				fPair.String(), strconv.FormatInt(b, 10), "200")
 			if err != nil {
 				return finalResp, err
 			}
+			if len(tempResp.Orders) == 0 {
+				break
+			}
+
 			for x := range tempResp.Orders {
 				resp.Exchange = e.Name
```
thrasher-: Heimdall: **Should fix** — `resp` lives outside both loops, so every order after the first inherits the previous order's `Cost`, `CostAsset` and `LastUpdated`, and `OrderID` is never set at all.

**Fix:** build it per order here in [lbank_wrapper.go:651](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L651) and drop `var resp order.Detail` at [lbank_wrapper.go:623](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L623):

```go
for x := range tempResp.Orders {
    resp := order.Detail{Exchange: e.Name, OrderID: tempResp.Orders[x].OrderID}
```

**Why:** [`InferCostsAndTimes`](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/order/orders.go#L863) fills `Cost`, `CostAsset` and `LastUpdated` only while they are still zero, and nothing resets the struct between iterations. Two orders on one page — 10 @ 1 filled, then 20 @ 2 filled:

```
order[0]: OrderID="" Price=10 ExecAmount=1 Cost=10 Date=2025-09-22 LastUpdated=2025-09-22
order[1]: OrderID="" Price=20 ExecAmount=2 Cost=10 Date=2025-09-23 LastUpdated=2025-09-22
                                                   ^^^^^^^ 20*2=40    ^^^^^^^^^^ order[0]'s
```

`CostAsset` bleeds across pairs too, since `resp` also survives the `tempCurr` loop.

Pre-existing — byte-identical output at [`158e9c1`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322) — so not a regression. Raising it because it is the loop you are rewriting, and now that later pages actually get fetched, many more records carry it. With the two lines above:

```
order[0]: OrderID="A" ... Cost=10 LastUpdated=2025-09-22
order[1]: OrderID="B" ... Cost=40 LastUpdated=2025-09-23
```

Package still green under `-race -count 1`, `golangci-lint` 0 issues.

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +380,68 @@ func TestGetOrderHistory(t *testing.T) {
...
+
+	got, err := ex.GetOrderHistory(t.Context(), &order.MultiOrderRequest{
+		Pairs:     currency.Pairs{testPair},
+		Side:      order.AnySide,
+		AssetType: asset.Spot,
+		Type:      order.AnyType,
+	})
+	require.NoError(t, err, "GetOrderHistory must not error")
+	assert.Equal(t, []string{"1", "2", "3"}, transport.pages, "GetOrderHistory should request each page once, in order")
+	assert.Len(t, got, 3, "GetOrderHistory should return the orders held on every page")
```
thrasher-: Heimdall: **Should fix** — the test never looks at which orders came back, so it stays green against a mapping that returns the wrong ones.

**Fix:** assert the decoded orders rather than the length — one `exp := []order.Detail{...}` compared with `assert.Equal`. The fixture already mints three distinct `order_id`s at [lbank_test.go:441](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_test.go#L441); nothing reads them back.

**Why:** mutation testing. Replacing every `tempResp.Orders[x]` with `tempResp.Orders[0]` in the mapping loop — so each order on a page becomes a copy of that page's first order — leaves the test passing:

```
--- PASS: TestGetOrderHistoryPagination (0.05s)
ok  	github.com/thrasher-corp/gocryptotrader/exchanges/lbank	0.082s
```

`assert.Equal` on the page sequence at [lbank_test.go:407](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_test.go#L407) is doing all the work; the `Len` check below it only proves three things came back, not that they are the three the exchange sent. Pinning content would also catch the field bleed flagged on [lbank_wrapper.go:651](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L651).

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +380,68 @@ func TestGetOrderHistory(t *testing.T) {
...
+}
+
+// orderHistoryTransport serves a three page order history for the mocked
+// exchange and records the current_page values LBank was asked for.
+type orderHistoryTransport struct {
```
thrasher-: Heimdall: **Should fix** — the hand-rolled `RoundTripper` is the only reason this test is green. `GetOrderHistory` cannot reach a real server at all, and a custom transport is precisely the thing that hides that.

**Fix:** use the idiom the rest of the repo uses — `httptest.NewTestServer(t, handler)`, `SetHTTPClient(server.Client())` and `SetRunningURL(exchange.RestSpot.String(), server.URL)`. That is 40 call sites across 18 exchange test files; this is the only exchange test defining its own transport.

**Why:** [`SendAuthHTTPRequest`](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank.go#L593) assigns the caller's relative path straight to `request.Item.Path` without the configured base URL, where the unauthenticated [`SendHTTPRequest`](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank.go#L500) uses `endpoint + path`. Same exchange, same client, same endpoint override:

```
AUTHENTICATED   QueryOrderHistory: serverCalls=0  err=Post "/v2/orders_info_history.do":
                                                      unsupported protocol scheme ""
UNAUTHENTICATED GetCurrencyPairs:  serverCalls=1
```

`http.Client` hands the request to a custom transport whatever the URL, so the mock answers a call the real client never manages to make. The host bug is pre-existing and not yours to fix here — it is in the review body — but as written the test cannot tell a working request path from a broken one, which is most of what a wrapper test is for.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -637,18 +637,16 @@ func (e *Exchange) GetOrderHistory(ctx context.Context, getOrdersRequest *order.
...
-		}
-		for len(tempResp.Orders) != 0 {
-			tempResp, err = e.QueryOrderHistory(ctx,
+		for b := int64(1); ; b++ {
+			tempResp, err := e.QueryOrderHistory(ctx,
 				fPair.String(), strconv.FormatInt(b, 10), "200")
 			if err != nil {
 				return finalResp, err
 			}
+			if len(tempResp.Orders) == 0 {
```
thrasher-, line 648: Heimdall: **Nit** — on a successful response the only exit is an empty page: no page cap, and LBank registers no rate limiter.

**Fix:** optional. A cap on `b` that returns an explicit error beats spinning, e.g. bail past some sane page count. LBank also returns `total` on this endpoint, which [`OrderHistoryResponse`](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_types.go#L108) does not decode — that would bound it exactly.

**Why:** [`request.New`](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L94) is called with no `request.WithLimiter`, unlike peers such as coinbase, so nothing throttles the loop client-side.

Two things I want to be straight about. I could not establish what LBank returns for a page past the end — the endpoint is authenticated and I have no credentials, and its own Python, Java, Node and Go SDKs all just forward the page number without crawling. And the base had exactly the same exposure, so this is not a regression; it only becomes reachable now that pages advance one at a time.

Worth noting `len(orders) < 200` is not the easy answer: it would break your current fixture, whose first page holds two orders, and it still cannot stop a server that repeats a full page. If you have seen this endpoint return an empty page past the end, say so and drop this one.

thrasher- (approved):
Heimdall: Approving. The diagnosis holds and the fix is right. Reverting just the wrapper to [`158e9c1`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322) while keeping your test reproduces the bug exactly — pages `["1","1","3"]` and 2 of 3 orders — so `TestGetOrderHistoryPagination` is a real regression test rather than a restatement of the new code. Build, vet, `golangci-lint` (0 issues), `golangci-lint fmt` and the full package under `-race -count 2` are all clean, and all 16 CI checks pass.

Four comments inline, evidence on the lines: [lbank_wrapper.go:651](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L651) order-to-order field bleed, [lbank_test.go:408](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_test.go#L408) the test does not pin content, [lbank_test.go:413](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_test.go#L413) the mock hides a dead request path, [lbank_wrapper.go:646](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L646) unbounded loop. None of them is a condition on the merge.

Separately, four pre-existing defects I reproduced while verifying this, all outside your diff. Happy to raise one issue covering them if you would rather not widen the PR:

- [lbank.go:593](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank.go#L593) — `SendAuthHTTPRequest` never prefixes the configured host, so all 12 authenticated LBank endpoints fail with `unsupported protocol scheme ""` before any I/O. That includes the one this PR fixes.
- [lbank_wrapper.go:745](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L745) — `getAllOpenOrderID` iterates the *previous* page's length over the current page: a pair whose second page is shorter panics with `index out of range`. It also refetches page 1, and [:742](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L742) returns from the whole function at the first pair that runs out, so later enabled pairs are never queried. Same paging bug family as this PR.
- [lbank_wrapper.go:576](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L576) — `GetActiveOrders` and `GetOrderInfo` index `Orders[0]` on a response `QueryOrder` explicitly allows to come back empty.
- [lbank_wrapper.go:657](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L657) — the side mapping has no unknown branch, so any `type` that is not exactly `buy` becomes SELL, an absent one included. [:313](https://github.com/thrasher-corp/gocryptotrader/blob/4279368a11ca8efcc7b63edd8c77d8d5b8590600/exchanges/lbank/lbank_wrapper.go#L313) in the same file already does the tolerant thing for trades.

Take the four inline or leave them — nothing here blocks.

Robin1987China:
Thanks for the review — and for the Heimdall pass, it caught real things.

**Per-order `resp` (lbank_wrapper.go:651)** — fixed. `order.Detail` is now built inside the per-order loop with `OrderID` set, so `Cost`, `CostAsset` and `LastUpdated` no longer bleed between records.

**Content assertion (lbank_test.go:408)** — the test now compares the full decoded orders against expected values built from the fixture instead of asserting the count, so the `tempResp.Orders[0]` mutation no longer passes.

**Hand-rolled transport (lbank_test.go:413)** — the standalone mock transport is gone. The test now follows the repo idiom — `httptest.NewTestServer`, `SetRunningURL(exchange.RestSpot.String(), server.URL)`, and `SetHTTPClient` with the server's own client — so requests go through a real `http.Client` against a real server. The one deviation is a small adapter (`relativeURLTransport`) on that client's transport, and it exists only because of the pre-existing `SendAuthHTTPRequest` bug from your review body: the request path is never prefixed with the configured endpoint, so without the adapter the request never leaves the client at all. It rewrites scheme/host-less URLs onto the test server and passes absolute URLs straight through, so once #2272 lands (its diff prefixes the endpoint in `SendAuthHTTPRequest`) the adapter is a no-op and can be deleted with it. As you said, that bug pre-dates this PR, so I've left it alone here.

**Empty-page exit (Nit)** — added `lbankOrderHistoryMaxPages` (1000) with an explicit error, so a server that keeps returning full pages can't spin the loop. I didn't decode `total` for an exact bound; happy to switch to that if you'd rather.

**One more fix since the last push** — the fixture timestamp was an untyped constant that overflowed `int` on `386`, which was failing that job. It's typed `int64` now and the job is green.

Ready for another look when you have a moment.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -637,20 +636,24 @@ func (e *Exchange) GetOrderHistory(ctx context.Context, getOrdersRequest *order.
...
+				break
+			}
+
 			for x := range tempResp.Orders {
-				resp.Exchange = e.Name
+				resp := order.Detail{
```
thrasher-, line 653: Heimdall: **Should fix** — every market, post-only, IOC and FOK **buy** in this history is mapped to SELL.

**Fix:** replace [`:662-666`](https://github.com/thrasher-corp/gocryptotrader/blob/d8aaebe4e842d09735d446d13c040629816bf967/exchanges/lbank/lbank_wrapper.go#L662-L666) with a match on the documented values:

```go
switch strings.ToLower(tempResp.Orders[x].Type) {
case "buy", "buy_market", "buy_maker", "buy_ioc", "buy_fok":
	resp.Side = order.Buy
case "sell", "sell_market", "sell_maker", "sell_ioc", "sell_fok":
	resp.Side = order.Sell
default:
	resp.Side = order.UnknownSide
}
```

**Why:** LBank documents `type` for this exact endpoint as `buy, sell, buy_market, sell_market, buy_maker, sell_maker, buy_ioc, sell_ioc, buy_fok, sell_fok` ([Query history order](https://www.lbank.info/docs/stylesheets/)), and only the bare `buy` survives `strings.EqualFold(..., order.Buy.String())`. ccxt holds a captured response from the same endpoint family carrying `"type":"buy_market"` and splits it on `_` for this reason — [`lbank.ts:1865`](https://github.com/ccxt/ccxt/blob/21f7cba7171b26898e5aac843c6886308f66d981/ts/src/lbank.ts#L1865). Driving this loop with all ten:

```
"buy"         -> BUY      "sell"        -> SELL
"buy_market"  -> SELL     "sell_market" -> SELL
"buy_maker"   -> SELL     "sell_maker"  -> SELL
"buy_ioc"     -> SELL     "sell_ioc"    -> SELL
"buy_fok"     -> SELL     "sell_fok"    -> SELL
```

and a `Side: order.Buy` request comes back with 1 of the 5 buys. With the switch above it is 10 of 10 and 5 of 5. `UnknownSide` is what [`FilterOrdersBySide`](https://github.com/thrasher-corp/gocryptotrader/blob/d8aaebe4e842d09735d446d13c040629816bf967/exchanges/order/orders.go#L894-L907) deliberately keeps rather than drops, so the default branch does not lose rows. Package green under `-race`, `golangci-lint` 0 issues, `golangci-lint fmt` clean.

Pre-existing and byte-identical on master, so not a regression. Your fixture pins `"type":"buy"` on every row, so the test cannot see it either way.

#2409 is converting the two sibling sites to `order.StringToOrderSide`. That helper will not serve here — it rejects all eight compound values (`"buy_market" -> UNKNOWN, "BUY_MARKET" order side is invalid`). Doing all three together in #2409 may be tidier than doing this one here; happy either way.

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +383,191 @@ func TestGetOrderHistory(t *testing.T) {
...
+}
+
+// orderHistoryTimestamp is the 2025-09-22T00:00:00Z created_time every fixture
+// order carries. It is typed int64 so the millisecond value stays representable
+// where the fixtures format it on 32-bit platforms.
+const orderHistoryTimestamp int64 = 1758499200000
+
+// orderHistoryEntry returns the JSON of a single LBank order history entry
+func orderHistoryEntry(o orderHistoryOrder) string {
+	return fmt.Sprintf(`{"order_id":%q,"symbol":"btc_usdt","type":"buy","price":%v,"amount":%v,"deal_amount":%v,"avg_price":%v,"status":0,"created_time":%d}`,
```
thrasher-, line 534: Heimdall: **Minor** — entirely optional, and not a condition on this PR: the fixture sends `created_time`, but LBank sends `create_time`. Under the real field name every order this function returns has a zero `Date`.

**Fix:** if you want it, send the documented name here and correct the tag on [`OrderResponse.CreateTime`](https://github.com/thrasher-corp/gocryptotrader/blob/d8aaebe4e842d09735d446d13c040629816bf967/exchanges/lbank/lbank_types.go#L86) to match — the two go together, since the fixture change on its own turns `TestGetOrderHistoryPagination` red. Otherwise leave it and I will carry it into a follow-up issue with the rest of the LBank mapping work.

**Why:** LBank's reference for this exact endpoint ([Query history order](https://www.lbank.info/docs/stylesheets/)) gives the field as `create_time` in both the example and the Returns table, and ccxt reads the same name off this endpoint family from a captured response — [`lbank.ts:1858`](https://github.com/ccxt/ccxt/blob/21f7cba7171b26898e5aac843c6886308f66d981/ts/src/lbank.ts#L1858):

```js
const timestamp = this.safeInteger2 (order, 'time', 'create_time');
```

Same payload, one field name changed:

```
create_time    -> Date=0001-01-01T00:00:00Z  LastUpdated=0001-01-01T00:00:00Z
created_time   -> Date=2017-01-18T05:09:58Z  LastUpdated=2017-01-18T05:09:58Z
```

A zero `Date` is not cosmetic: `InferCostsAndTimes` copies it into `LastUpdated`, and any `StartTime`/`EndTime` on the request then drops every order through `FilterOrdersByTimeRange`.

The tag is pre-existing — `json:"created_time"` has been there since [`2dc813b5f`](https://github.com/thrasher-corp/gocryptotrader/commit/2dc813b5f307d47840166e958a6d6a8d52b55ea5) in 2019 — and master behaves identically, so it is not yours, and none of it is reachable anyway until #2272 fixes the host prefix. I am mentioning it on your line only because this fixture is the first thing that has ever exercised the mapping. Applied as described, the package is green under `-race`, `golangci-lint` reports 0 issues and `golangci-lint fmt` is clean.

exchanges/lbank/lbank.go
```diff
@@ -66,12 +66,19 @@ const (
 	lbankWithdraw                = "withdraw.do"
 	lbankRevokeWithdraw          = "withdrawCancel.do"
 	lbankTimestamp               = "timestamp.do"
+
+	// lbankOrderHistoryMaxPages bounds how far order history is crawled. The
+	// endpoint is only bounded by an empty page, and LBank registers no rate
+	// limiter, so a server that keeps returning rows for a page past the end
+	// would spin this loop forever.
+	lbankOrderHistoryMaxPages = 1000
```
thrasher-, line 74: Heimdall: **Minor** — 1000 is not the bound this crawl actually has, in either direction.

**Fix:** none needed for correctness; the cap does stop a repeating server, which is what I asked for last round. But it is worth a sentence in the comment saying what really limits the crawl, or widening [`CurrentPage`/`PageLength`](https://github.com/thrasher-corp/gocryptotrader/blob/d8aaebe4e842d09735d446d13c040629816bf967/exchanges/lbank/lbank_types.go#L108-L121) to `uint64` so the number means something.

**Why:** if the server echoes the page it was asked for, `CurrentPage uint8` gives out first:

```
calls=256  orders=255  errors.Is(err, errOrderHistoryPageLimit)=false
err = json: cannot unmarshal number 256 into Go struct field
      OrderHistoryResponse.current_page of type uint8, authenticated request failed
```

Same under `-tags sonic_on`. With both structs at `uint64`: `calls=1000`, `errors.Is(...)=true`. `TestGetOrderHistoryPageLimit` reaches 1000 only because its mock pins `current_page` to `1`.

Going the other way, LBank says this endpoint holds "only the records in recent two days" ([Query history order](https://www.lbank.info/docs/stylesheets/)), so 1000 pages of 200 is far past anything it will serve. And its published example returns the pagination metadata quoted — `"current_page": "1"`, `"page_length": "20"`, `"total": "1"` — which the current `uint8` fields cannot decode at all:

```
json: cannot unmarshal string into Go struct field OrderHistoryResponse.page_length of type uint8
```

I want to be clear that last one is a documented-shape concern, not something I have observed live: ccxt's captured samples for the neighbouring `supplement` endpoint show these as bare numbers, and I have no credentials to settle which the legacy endpoint sends today. Pre-existing either way — both structs are unchanged from master.

Also worth knowing: the cap is what the bound test costs. It makes 1000 signed round trips, about 6s on its own and 10s under `-race` on my machine.

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +383,191 @@ func TestGetOrderHistory(t *testing.T) {
...
+	OrderID    string
+	Price      float64
+	Amount     float64
+	DealAmount float64
+}
+
+// orderHistoryFixture is the three page order history the mocked server serves.
+// The fourth page, and every page after it, is empty.
+var orderHistoryFixture = [][]orderHistoryOrder{
+	{{"1", 10, 2, 1}, {"2", 20, 3, 2}},
```
thrasher-, line 523: Heimdall: **Minor** — the fixture varies three fields and holds seven constant, so several mappings next to the ones you fixed are still unpinned.

**Fix:** vary `avg_price` independently of `price`, and give the rows different sides, statuses and timestamps. One row with `"type":"sell"`, one `status:2`, and a distinct `created_time` per order would cover it.

**Why:** mutation testing on this commit. Each of these leaves both new tests passing:

```
resp.AverageExecutedPrice = Orders[x].AvgPrice  ->  Orders[x].Price     PASS
lbankOrderHistoryMaxPages = 1000                ->  3                   PASS
if b > lbankOrderHistoryMaxPages                ->  b >=                PASS
```

The first slips through because every fixture row sets `avg_price` equal to `price`. The last two slip through because neither test pins the request count or the boundary — `TestGetOrderHistoryPageLimit` only asserts that the sentinel eventually comes back. Asserting the number of requests in that test would close both.

This is a real step up from the last round — `Cost`, `OrderID` and the per-page sequence are all properly pinned now, and I confirmed that by reverting each fix. This is about the rows the fixture does not distinguish.

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +383,191 @@ func TestGetOrderHistory(t *testing.T) {
...
+// fields over from the previous one, cannot pass.
+type orderHistoryOrder struct {
+	OrderID    string
+	Price      float64
+	Amount     float64
+	DealAmount float64
+}
+
+// orderHistoryFixture is the three page order history the mocked server serves.
+// The fourth page, and every page after it, is empty.
```
thrasher-, line 521: Heimdall: **Nit** — this comment does not describe the fixture below it.

**Fix:**

```suggestion
// orderHistoryFixture is the two page order history the mocked server serves.
// The third page, and every page after it, is empty.
```

**Why:** the fixture holds two pages, of two orders and one. The handler falls through to the empty body from page 3 onward, which is what the `["1", "2", "3"]` assertion expects.

thrasher- (commented):
Heimdall: Both fixes from last round hold up under mutation — restoring the shared `resp` fails `TestGetOrderHistoryPagination` on `Cost` and `OrderID`, and `Orders[x]` -> `Orders[0]` fails it too. You also caught the 32-bit overflow in [`d8aaebe4`](https://github.com/thrasher-corp/gocryptotrader/commit/d8aaebe4e842d09735d446d13c040629816bf967) before I could post it.

Nothing here blocks. Everything below is pre-existing, byte-identical on master, and unreachable anyway until #2272 fixes the host prefix.

Checking what the fixture asserts turned up LBank's own reference for this endpoint, which is not on their current docs site: [Query history order](https://www.lbank.info/docs/stylesheets/). Two places the fixture and the real payload differ, each hiding a live defect — `created_time` where LBank sends `create_time`, which zeroes `Date` on every order, and `"type":"buy"` on every row, which hides 4 of 5 buy types mapping to SELL. Both inline, both applied and tested. Take them or leave them; I am just as happy to fold them into one issue covering the LBank order-mapping family, alongside #2409.

Your note on `relativeURLTransport` is accurate, and I checked it rather than took it on trust: applying #2272's `Path: endpoint + path` does turn the adapter into a no-op with the test still passing, so it deletes cleanly with that PR.

Three minors inline. CI is green bar the Windows job, which is [huobi network timeouts](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35719179871/job/106717786551) — `lbank` passes in that same job. Locally `golangci-lint` v2.13.2/go1.27 is 0 issues, `fmt` clean, `TestAllExchangeWrappers/lbank` green here and on master.

PR #2406, opened 2026-09-22
kucoin: return the max withdraw margin the endpoint sent, not a zero value

Robin1987China:
# PR Description

`kucoin.Exchange.GetMaxWithdrawMargin` always returns `0`:

```go
func (e *Exchange) GetMaxWithdrawMargin(ctx context.Context, symbol string) (float64, error) {
	if symbol == "" {
		return 0, currency.ErrSymbolStringEmpty
	}
	var resp types.Number
	return resp.Float64(), e.SendAuthHTTPRequest(ctx, exchange.RestFutures, maxWithdrawMarginEPL, http.MethodGet, "/v1/margin/maxWithdrawMargin?symbol="+symbol, nil, &resp)
}
```

`resp.Float64()` is a **method call**, and Go evaluates the operands of a return
statement left to right (the spec's *Order of evaluation* rule covers function
calls, method calls, receive operations and binary logical operations). So
`resp.Float64()` is guaranteed to run **before** `SendAuthHTTPRequest` fills
`resp` through `&resp`, and it reads the zero value. The call succeeds, no error
is returned, and the caller silently receives `0` instead of the real margin.

The filling happens through a pointer, so nothing in the return values can pick
up the written value: the value that is returned was read before the write.

This change hoists the request into `err` and reads the margin afterwards:

```go
	var resp types.Number
	err := e.SendAuthHTTPRequest(ctx, exchange.RestFutures, maxWithdrawMarginEPL, http.MethodGet, "/v1/margin/maxWithdrawMargin?symbol="+symbol, nil, &resp)
	return resp.Float64(), err
```

Error handling is unchanged: on failure the zero value is still returned
alongside the error, exactly as before. Only the success path changes.

## Test

`TestGetMaxWithdrawMarginMocked` serves the endpoint from an `httptest` server
wired in with `SetHTTPClient` / `SetRunningURL`, so no live request is made. It
asserts the request path and `symbol`, and that the returned margin equals the
value the endpoint sent rather than `0`.

Reverting the fix makes the new test fail with `expected: 1.5, actual: 0`, so it
genuinely pins the behaviour.

## Scope

An audit for the same shape (`return <ident>.<Method>(), <call that fills
<ident> via a pointer>`) across `exchanges/` found exactly two sites: this one
and `lbank.USD2RMBRate`. Only the method-call form is affected; the far more
common `return resp, e.Send...(&resp)` idiom returns a plain variable, not a
call, and is unaffected. The lbank site is being fixed separately.


thrasher-:
@codex please review

exchanges/lbank/lbank_test.go
```diff
@@ -198,6 +201,29 @@ func TestUSD2RMBRate(t *testing.T) {
...
 
+// TestUSD2RMBRateMocked pins the rate the caller receives. The live endpoint returns a bare
+// JSON number; USD2RMBRate must return that value rather than the zero value it read before the
+// filling call ran.
+func TestUSD2RMBRateMocked(t *testing.T) {
+	t.Parallel()
+
+	server := httptest.NewTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		assert.Equal(t, "/v"+lbankAPIVersion2+"/"+lbankUSD2CNYRate, r.URL.Path, "the request path should be the USD to CNY rate endpoint")
+		_, err := fmt.Fprint(w, `6.6951`)
```
thrasher-, line 212: Heimdall: **Should fix** — this mock serves a response shape the live endpoint does not send, and the comment states it as fact.

**Fix:**

```suggestion
// TestUSD2RMBRateMocked pins the rate the caller receives: USD2RMBRate must return the value the
// endpoint sent rather than the zero value it read before the filling call ran.
func TestUSD2RMBRateMocked(t *testing.T) {
	t.Parallel()

	server := httptest.NewTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		assert.Equal(t, "/v"+lbankAPIVersion2+"/"+lbankUSD2CNYRate, r.URL.Path, "the request path should be the USD to CNY rate endpoint")
		// Trimmed from GET /v2/usdToCny.do
		_, err := fmt.Fprint(w, `{"result":"true","data":"6.6951","error_code":0,"ts":1790073372701}`)
```

**Why:** the live endpoint returns the v2 envelope, not a bare number:

```
$ curl -s https://api.lbkex.com/v2/usdToCny.do
{"result":"true","data":"6.6951","error_code":0,"ts":1790073372701}
```

That matters beyond the comment being wrong. [`SendHTTPRequest`](https://github.com/thrasher-corp/gocryptotrader/blob/a53f7f8b8275b1f6367093cab89e8a2338297b48/exchanges/lbank/lbank.go#L516-L525) parses the envelope first and, when `Data != nil`, unmarshals `Data` into the result; only if the envelope fails to parse does it fall through to `json.Unmarshal(tempResp, result)`. A bare `6.6951` fails the envelope parse, so the test exercises the fallback branch while production for this endpoint takes the envelope branch — the one path the test never covers.

I applied the suggestion and re-ran it both ways: it still passes with the fix and still fails without it (`expected: 6.6951, actual: 0`), so it keeps all of its teeth while testing the path that actually runs. Same value, so no other assertion changes.

This also matters for your own #2404, which rewrites this exact envelope handling into `unwrapV2Response`. I merged master + #2404 + this PR: no conflict, `go build ./...` clean and the full `go test ./exchanges/lbank/... -race` green, and the suggestion above passes there and still fails there without the fix — so it is robust to whichever lands first. Under #2404 a bare number takes the new "not an envelope; return it unchanged" branch, so this test would still never touch the envelope path that #2404 exists to fix.

For contrast, the kucoin mock in this PR is already faithful — `{"code":"200000","data":"1.5"}` matches [KuCoin's documented response](https://www.kucoin.com/docs-new/rest/futures-trading/positions/get-max-withdraw-margin) (`{"code":"200000","data":"21.1135719252"}`, quoted-string `data`) and goes through the real `&Response{Data: result}` unwrap.

thrasher- (commented):
Heimdall: Well-argued and correctly diagnosed — it cites the real spec rule, and both tests have
teeth: reverting just the two production lines fails them at `actual: 0`.

I checked the scope claim rather than take it. A type-aware, deliberately over-approximating pass
over the whole module — every package, every test file, every build config — leaves 64 hand-written
candidates, of which exactly **two** are the bug. The two you fixed. No third site, by two
independent methods.

One correction for the description: it is not only the *method*-call form. Measured on go1.27, an
ordinary function call, `len()`, and `var v, err = n.Float64(), fill(&n)` break identically — any
call in an earlier operand is ordered before the filling call. And `return resp, e.Send(&resp)` is
safe only because of how gc happens to order non-call operands;
[the spec](https://go.dev/ref/spec#Order_of_evaluation) leaves that unspecified. Only
`return &resp, …` is safe by construction.

Two smaller things. One should-fix inline ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2406#discussion_r4071537683)): the lbank mock serves a
shape the live endpoint does not send. And in the description, "on failure the zero value is still
returned alongside the error, exactly as before" no longer holds —
`{"code":"400100","msg":"nope","data":"7.25"}` now returns `7.25` with the error. Harmless and
idiomatic, but the sentence overstates.

**CI is red and neither failure is yours.** `engine` is the only failing package — kucoin and lbank
both passed in that same run — and it fails on a live BTSE `429`. Master's windows job failed at the
same test and the same line with `/coinut` and a timeout, so a re-run should clear it.
`codecov/project` is -0.01% against a base it admits is behind head; `codecov/patch` passes and both
your files gain coverage. Locally on the merge ref everything is green: lint, `-race`, `-shuffle`,
wrapper standards, misc_checks.

Nice catch — the spec citation saved me a lot of time.

PR #2407, opened 2026-09-22
okx: allow GetLiquidationOrders to query SWAP, FUTURES and OPTION

Robin1987China:
### What this fixes

`GetLiquidationOrders` is unusable for every instrument type other than MARGIN.

The instrument-type switch requires MARGIN in both of its cases, so any other
`instType` falls through to `default` and returns `errEitherInstIDOrCcyIsRequired`:

    switch {
    case arg.InstrumentType == instTypeMargin && arg.InstrumentID != "":
        params.Set("instId", arg.InstrumentID)
    case arg.InstrumentType == instTypeMargin && arg.Currency.String() != "":
        params.Set("ccy", arg.Currency.String())
    default:
        return nil, errEitherInstIDOrCcyIsRequired
    }

The `uly` and `alias` handling below the switch is written for SWAP/FUTURES/OPTION
and is therefore dead code — it can never be reached. `instId` and `ccy` only apply
to MARGIN, so requiring one of them for MARGIN is correct, but the requirement must
not apply to the other instrument types. A caller asking for SWAP liquidation
orders cannot get them at all, and the error they receive (`either parameter
instId or ccy is required`) points at a parameter that is irrelevant to SWAP.

There is a second, independent bug: the response is decoded into a single
`*LiquidationOrder`, but OKX always answers this endpoint with an array, so even a
successful call fails to decode.

### The change

- The `instId`/`ccy` requirement is scoped to MARGIN. MARGIN behaviour, including
  its error path, is unchanged and is now locked down by a test.
- SWAP/FUTURES/OPTION are filtered by `instFamily` (preferred) or `uly`, plus
  `state`; FUTURES additionally honours `alias`. These are the documented filters
  for those types and were already being assembled — they are now reachable.
- The instrument-type guard now rejects any unsupported `instType` up front, not
  just the empty string.
- The response is decoded into `[]LiquidationOrder`, matching the array OKX
  actually returns.
- The order-limit guard (`arg.Limit > 0 && arg.Limit < 100`) is unchanged from
  master.

### Testing

New offline test `TestGetLiquidationOrdersInstrumentTypes` (11 subtests) drives
every supported instrument type against an `httptest` server and asserts the
outgoing request parameters and the decoded length, so no credentials or network
access are needed. The stub serves the two-element array shape captured from a
live OPTION response.

The pre-existing tests for this endpoint are credential gated
(`sharedtestvalues.SkipTestIfCredentialsUnset`), so these paths were previously
untested in CI.

Without this change, SWAP, FUTURES and OPTION calls return `either parameter
instId or ccy is required`; with this change they pass. (The new test file does
not compile against the base because it exercises the new `InstrumentFamily`
field, so the base behaviour was confirmed with a minimal probe rather than by
re-running the new subtests.)

Verified locally: `go build ./...`, `go vet`, `gofmt -l` (empty),
`go test -race -count 1` on the affected package, `scripts/misc_checks.sh`, and
golangci-lint (0 issues).

Fixes #

## Type of change

Breaking only in that the exported signature changes from `*LiquidationOrder` to
`[]LiquidationOrder`.

## How has this been tested

- [x] `go build ./...`
- [x] `go test ./exchanges/okx/... -race -count 1` — package suites pass; `TestBusinessWSCandleSubscriptions`
      fails on this machine with a websocket connection reset, and fails identically on the base commit
- [x] `golangci-lint run ./exchanges/okx/...` — 0 issues

`TestGetLiquidationOrdersInstrumentTypes` covers MARGIN, SWAP, FUTURES and OPTION against a stub of the
real response shape, asserts the decoded length and the full encoded query per case. Removing the `state`
forwarding, the `instFamily` preference, the `instType` guard or the multi-element stub each turns the
corresponding case red.

## Checklist


thrasher-:
@codex please review

exchanges/okx/okx.go
```diff
@@ -5090,9 +5090,12 @@ func (e *Exchange) GetLiquidationOrders(ctx context.Context, arg *LiquidationOrd
 		params.Set("mgnMode", arg.MarginMode)
 	}
 	switch {
-	case arg.InstrumentType == instTypeMargin && arg.InstrumentID != "":
+	case arg.InstrumentType != instTypeMargin:
```
thrasher-, line 5095: Heimdall: **Blocker** — SWAP, FUTURES and OPTION still do not work after this change. All three of the new test cases fail against the live endpoint.

**Why:** I ran the three new cases through the real `GetLiquidationOrders` at `2bf391a`, with no stub and no credentials:

```
"swap is filtered by underlying"
    -> Okx unsuccessful HTTP status code: 400 raw response: {"code":"50014","data":[],"msg":"Parameter state can not be empty."}
"futures is filtered by underlying and alias"
    -> error code: `51021`; message: "This crypto isn't listed yet."
"option is filtered by underlying"
    -> cannot unmarshal response data directly (error: json: cannot unmarshal array into Go value of type okx.LiquidationOrder) or as a single-item array (got 2 items)
```

Three independent causes, none of them addressed here.

**1. `state` is never sent, and it is required for SWAP and FUTURES.** The [archived parameter table](https://web.archive.org/web/20220530164307id_/https://www.okx.com/docs-v5/en/) describes it as "State `unfilled` `filled`, `unfilled` by default, Required for FUTURES/SWAP", and the live endpoint enforces it — `instType=SWAP&uly=BTC-USDT` returns `50014 Parameter state can not be empty.`. [`LiquidationOrderRequestParams.State`](https://github.com/thrasher-corp/gocryptotrader/blob/2bf391a2ee057148fcb6c4b1da17b8fe1b7ffa78/exchanges/okx/okx_types.go#L510) already exists; nothing reads it.

**2. `uly` did not work for FUTURES in any combination I tried — only `instFamily` did,** and there is no `InstrumentFamily` field on the request struct:

```
instType=FUTURES&uly=BTC-USD&alias=quarter&state=filled        -> {"code":"51021","msg":"This crypto isn't listed yet."}
instType=FUTURES&instFamily=BTC-USD&alias=quarter&state=filled -> {"code":"0"}  1 element, 2 details
```

Every alias I tried through `uly` behaves the same way — `this_month`, `next_month`, `quarter`, `next_quarter` and `third_quarter` all return 51021, for `BTC-USD`, `BTC-USDT`, `ETH-USD` and `ETH-USDT` alike. The test's own value compounds it: it passes `mainPair` (`BTC-USDT`), and `GET /api/v5/public/instruments?instType=FUTURES&uly=BTC-USDT` lists zero instruments — OKX has no BTC-USDT futures. `this_week` is also no longer a live alias.

The reason this drifted unnoticed is that OKX delisted the endpoint: the [changelog](https://www.okx.com/docs-v5/log_en/) records "Delisted endpoints from the document: Get liquidation orders (GET /api/v5/public/liquidation-orders)" on 2023-03-27, and on 2023-04-03 that REST retrieval would stop being supported by the end of April 2023, in favour of the WebSocket channel. It still answers, but it is absent from the [current docs](https://www.okx.com/docs-v5/en/) entirely, so there is nothing left to validate `uly` or `alias` against — hence the archived table above.

**3. The return type cannot hold the response.** `*LiquidationOrder` is a single object, but `data` is an array, and OKX sends both multi-element and empty arrays. Replaying captured live payloads through this code:

```
MARGIN ccy=BTC (1 elem)                -> err=<nil>
OPTION uly=BTC-USD (2 elems)           -> err=cannot unmarshal ... or as a single-item array (got 2 items)
SWAP state=unfilled, no results (0)    -> err=cannot unmarshal ... or as a single-item array (got 0 items)
```

The element count is data-dependent, so this is a decode that fails whenever OKX returns anything other than exactly one element — but `uly=BTC-USD` returned 2 on ten consecutive samples, and `uly=ETH-USD` 2 as well, so in practice OPTION is simply broken. The empty case is not exotic either: it is the ordinary "no liquidations" response, and the archived table gives `unfilled` as the *default* state, so `instType=SWAP&uly=BTC-USDT&state=unfilled` returns `{"code":"0","data":[]}` and this code turns that into an error.

**Fix:** three changes, on top of the switch fix you already have. This is what I tested — it keeps your change intact:

```diff
-func (e *Exchange) GetLiquidationOrders(ctx context.Context, arg *LiquidationOrderRequestParams) (*LiquidationOrder, error) {
+func (e *Exchange) GetLiquidationOrders(ctx context.Context, arg *LiquidationOrderRequestParams) ([]LiquidationOrder, error) {
@@
-	if arg.InstrumentType != instTypeMargin && arg.Underlying != "" {
-		params.Set("uly", arg.Underlying)
+	if arg.InstrumentType != instTypeMargin {
+		if arg.InstrumentFamily != "" {
+			params.Set("instFamily", arg.InstrumentFamily)
+		} else if arg.Underlying != "" {
+			params.Set("uly", arg.Underlying)
+		}
+		if arg.State != "" {
+			params.Set("state", arg.State)
+		}
 	}
@@
-	var resp *LiquidationOrder
+	var resp []LiquidationOrder
```

plus `InstrumentFamily string` on `LiquidationOrderRequestParams`, next to `Underlying`. Live results with exactly that patch:

```
MARGIN ccy                               len=1 err=<nil>
SWAP uly+state                           len=1 err=<nil>
SWAP uly+state=unfilled (empty result)   len=0 err=<nil>
FUTURES instFamily+alias+state           len=1 err=<nil>
OPTION uly                               len=2 err=<nil>
```

It is not a one-click apply, because the new test needs updating alongside it — with a slice return the current stub payload is rejected outright, which is the subject of the other comment. Worth calling out that the return-type change is source breaking for anyone using this as a library, including on the MARGIN path that works today; there are no in-repo callers, but that is not the same as no users.

Two things worth considering but not worth holding the PR for. First, [`GetDeliveryHistory`](https://github.com/thrasher-corp/gocryptotrader/blob/2bf391a2ee057148fcb6c4b1da17b8fe1b7ffa78/exchanges/okx/okx.go#L4934) already has the pattern for rejecting a missing family locally, with an existing sentinel — `errInstrumentFamilyOrUnderlyingRequired` — that this function could reuse rather than letting OKX answer `50015`. Second, since the archived table gives `unfilled` as the default `state` and the server now rejects its absence outright, defaulting an empty `State` to `unfilled` rather than omitting it would make your existing SWAP test case work as written. Both variants are green here; I left them out above to keep the diff proportionate, and which one you want is your call.

If you would rather keep this PR smaller still, the alternative is to land the switch fix on its own, drop the three success cases from the test, and narrow the title and commit subject — because as it stands the test asserts an outcome the API does not produce, and that subject is what ends up in the merged history.

exchanges/okx/okx_test.go
```diff
@@ -605,6 +608,125 @@ func TestGetLiquidationOrders(t *testing.T) {
...
+			var (
+				mutex sync.Mutex
+				query url.Values
+			)
+			server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				mutex.Lock()
+				query = r.URL.Query()
+				mutex.Unlock()
+				w.Header().Set("Content-Type", "application/json")
+				_, _ = w.Write([]byte(`{"code":"0","msg":"","data":{"instType":"SWAP","instId":"BTC-USDT-SWAP","totalLoss":"0"}}`))
```
thrasher-: Heimdall: **Should fix** — this stub is the one response shape OKX never returns, which is why the test passes while all three new instrument types fail in practice.

On this endpoint `data` is an array — every live response I captured, across all four instrument types, was one. Here the stub makes it a JSON object. I replayed captured live payloads and this stub through the code at `2bf391a`:

```
MARGIN ccy=BTC (1 elem)                -> err=<nil>
OPTION uly=BTC-USD (2 elems)           -> err=cannot unmarshal ... or as a single-item array (got 2 items)
SWAP state=unfilled, no results (0)    -> err=cannot unmarshal ... or as a single-item array (got 0 items)
PR test stub (data as an OBJECT)       -> err=<nil>
```

Note that a single-element *array* would pass too — [`SendHTTPRequest`](https://github.com/thrasher-corp/gocryptotrader/blob/2bf391a2ee057148fcb6c4b1da17b8fe1b7ffa78/exchanges/okx/okx.go#L5952-L5968) unwraps `data:[{…}]` into a non-slice result — so only a multi-element array exercises the real shape.

**Fix:** make the stub a two-element array taken from a real OPTION response, and assert the decoded length:

```go
_, _ = w.Write([]byte(`{"code":"0","msg":"","data":[{"instType":"OPTION","uly":"BTC-USD","totalLoss":"-8.8193","details":[]},{"instType":"OPTION","uly":"BTC-USD","totalLoss":"0","details":[{"bkLoss":"-36.13","ts":"1790064000278"}]}]}`))
```

That change on its own makes the test fail against the current code, which is the point — it needs the return-type change from the other comment to go green.

Separately, the assertion loop only checks that the keys in `expected` are present, so it cannot catch anything extra. Every derivative row also leaves `InstrumentID` and `Currency` empty, which means the new MARGIN-only boundary — the actual subject of this PR — is never tested: populate them on a SWAP row and assert `instId` and `ccy` are *absent*. Asserting the whole encoded query per case, rather than a subset of keys, would have caught the missing `state` before this reached review.

exchanges/okx/okx.go
```diff
@@ -5090,9 +5090,12 @@ func (e *Exchange) GetLiquidationOrders(ctx context.Context, arg *LiquidationOrd
...
+		// instId and ccy only apply to MARGIN, SWAP, FUTURES and OPTION orders
+		// are filtered by uly and alias further below.
+	case arg.InstrumentID != "":
 		params.Set("instId", arg.InstrumentID)
-	case arg.InstrumentType == instTypeMargin && arg.Currency.String() != "":
+	case arg.Currency.String() != "":
 		params.Set("ccy", arg.Currency.String())
 	default:
```
thrasher-, line 5102: Heimdall: **Minor** — nothing rejects an unsupported instrument type any more, so bad input now costs a network round-trip.

Master's rejection here was incidental rather than deliberate validation — it fell out of the MARGIN-only branch — but it did mean `GetLiquidationOrders(ctx, &LiquidationOrderRequestParams{InstrumentType: "SPOT"})` returned `either parameter instId or ccy is required` with zero HTTP requests. At `2bf391a` it reaches OKX and comes back as `{"code":"51000","data":[],"msg":"Parameter instType error"}` — a worse message, one wasted request and one rate-limit slot. The endpoint accepts only `MARGIN`, `SWAP`, `FUTURES` and `OPTION`.

**Fix:** reject anything else up front, alongside the existing empty check and in the style already used at [`okx.go:2097`](https://github.com/thrasher-corp/gocryptotrader/blob/2bf391a2ee057148fcb6c4b1da17b8fe1b7ffa78/exchanges/okx/okx.go#L2097):

```go
	switch arg.InstrumentType {
	case instTypeMargin, instTypeSwap, instTypeFutures, instTypeOption:
	default:
		return nil, fmt.Errorf("%w, received %q", errInvalidInstrumentType, arg.InstrumentType)
	}
```

That subsumes the current empty-string check, since `""` falls to `default`. Raising it separately from the response-shape blocker so the two can be resolved on their own.

thrasher- (changes requested):
Heimdall: The diagnosis is right, and it is an easy bug to walk past — `instTypeMargin` appears in *both* arms of the switch, so every non-MARGIN caller falls to `default` and the `uly`/`alias` handling below it has been unreachable since it was written. I confirmed that by execution on master. MARGIN comes out byte-identical across a 48-case sweep, and the new test needs no credentials so it genuinely runs in CI.

The blocker is that the change does not deliver what the title promises. The three new test cases, run against the live endpoint through the real `GetLiquidationOrders` at 2bf391a with no stub:

```
"swap is filtered by underlying"
    -> Okx unsuccessful HTTP status code: 400 raw response: {"code":"50014","data":[],"msg":"Parameter state can not be empty."}
"futures is filtered by underlying and alias"
    -> error code: `51021`; message: "This crypto isn't listed yet."
"option is filtered by underlying"
    -> cannot unmarshal response data directly (error: json: cannot unmarshal array into Go value of type okx.LiquidationOrder) or as a single-item array (got 2 items)
```

Three separate causes, detailed inline, all of which go away with about fifteen lines on top of what you already have.

Neither red CI check is yours: the arm64 job hit a live BTSE `429 Too Many Requests` in `engine`, which passes here on master, and `codecov/project` is currently red on several other open PRs while `codecov/patch` passes. Locally, lint, `misc_checks.sh`, `gofmt -l` and `go test ./exchanges/okx/... -race -count 1` are green on both this branch and master, and `GetLiquidationOrders` coverage goes 66.7% → 87.5%.

Two corrections to the description: the pre-existing `TestGetLiquidationOrders` is not credential gated — it has no gate and passes with no credentials — and `uly`/`alias` are not *currently* the documented filters, since OKX withdrew that table in 2023.

**To unblock:**
1. Forward `state`, add and forward `instFamily`, and return `[]LiquidationOrder` — or narrow the title, commit subject and test ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2407#discussion_r4071311090))
2. Make the stub return `data` as a multi-element array so the test can see the real response shape ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2407#discussion_r4071311106))

Robin1987China:
Pushed `b995a47` — thanks for the detail, the three live failures were the useful part.

- `state` is now forwarded, and `InstrumentFamily` was added and takes precedence over `uly` (`instFamily` is the filter FUTURES actually honours).
- `GetLiquidationOrders` returns `[]LiquidationOrder`, so multi-element and empty `data` arrays both decode. Source-breaking for library users, as you noted; there are no in-repo callers.
- Unsupported `instType` is rejected up front with `errInvalidInstrumentType`.
- The stub is now the two-element OPTION array you gave, the decoded length is asserted, and each case asserts the full encoded query rather than a subset of keys.

One thing to flag: the diff also drops the pre-existing `Limit < 100` guard, so `limit` is forwarded as given. It is the same silently-dropped-filter class and a case pins it — happy to split it out if you would prefer to keep this PR to the three points above.

exchanges/okx/okx.go
```diff
@@ -5090,15 +5092,27 @@ func (e *Exchange) GetLiquidationOrders(ctx context.Context, arg *LiquidationOrd
...
 		return nil, errEitherInstIDOrCcyIsRequired
 	}
-	if arg.InstrumentType != instTypeMargin && arg.Underlying != "" {
-		params.Set("uly", arg.Underlying)
+	if arg.InstrumentType != instTypeMargin {
+		// instFamily is the only filter FUTURES honours, uly is ignored there.
```
thrasher-, line 5106: Heimdall: **Nit** — OKX doesn't ignore `uly` for FUTURES, it rejects it. Live, with `instType=FUTURES&alias=quarter&state=filled`:

```
uly=BTC-USD         -> 51021 This crypto isn't listed yet.
uly=NOPE-USD        -> 51014 Index doesn't exist.
instFamily=BTC-USD  -> 0
```

The code is right; only the comment misdescribes why.

**Fix:** `// OKX rejects uly for FUTURES, so instFamily takes precedence.`

exchanges/okx/okx_test.go
```diff
@@ -605,6 +608,187 @@ func TestGetLiquidationOrders(t *testing.T) {
...
+			},
+			expected: url.Values{
+				"instType":   {instTypeFutures},
+				"instFamily": {"BTC-USD"},
+				"alias":      {"quarter"},
+				"state":      {"filled"},
+			},
+		},
+		{
+			name: "futures prefers the instrument family over the underlying",
```
thrasher-, line 656: Heimdall: **Nit** — this FUTURES case has no `alias`, and OKX refuses that exact query:

```
instType=FUTURES&instFamily=BTC-USD&state=filled -> 400 {"code":"51000","msg":"Parameter alias error"}
```

**Fix:** add `Alias: "quarter"` to the request and `"alias": {"quarter"}` to the expected query. The precedence check is unchanged and the fixture then models a request OKX accepts.

thrasher- (approved):
Heimdall: Approving. All three round-one threads are fixed and verified live at [`583e5a7`](https://github.com/thrasher-corp/gocryptotrader/commit/583e5a70d368bd56abb30577a871330837fbb96b): SWAP, FUTURES and OPTION now return data through the real `GetLiquidationOrders`, including OPTION's two-element and SWAP's empty responses, `SPOT` is rejected without a request, and reverting the `okx.go` logic fails 6 of 11 subtests.

Inline, none blocking: Nits on the [FUTURES `uly` comment](https://github.com/thrasher-corp/gocryptotrader/pull/2407#discussion_r4077362510) and a [fixture missing `alias`](https://github.com/thrasher-corp/gocryptotrader/pull/2407#discussion_r4077362516).

- The websocket `liquidation-orders` handler decodes the whole push envelope into `*LiquidationOrder`, so it always emits an empty order ([okx_websocket.go:532](https://github.com/thrasher-corp/gocryptotrader/blob/583e5a70d368bd56abb30577a871330837fbb96b/exchanges/okx/okx_websocket.go#L532)).
- The description still says the existing test is credential gated; it isn't.

The red 386 job is ambient: live BTSE/Kraken ticker failures and websocket connection resets, including OKX `TestBusinessWSCandleSubscriptions`.

PR #1005, opened 2022-08-05
exchanges: Add OKX support

samuael:
# PR Description

This pull request Included a code implementation for the Okx exchange; with REST and Websocket support. All the Websocket and Wrapper functions are tested. My code follows the style guidelines of other exchanges of GCT.
Fixed some errors with the help of golangci-lint. Endpoint methods that have no support from the GCT wrapper are also implemented for future use.

Fixes # (issue)

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality not to work as expected)
- [x] This change requires a documentation update

## How has this been tested

For testing the exchange with your own API Credentials you can use your own API key and passphrase fields in the unit test okx_test.go file, or directly add credential information to the config file you will be using to the corresponding exchange.

- [ ] go test ./... -race
- [x] golangci-lint run
- [x] Test X

## Checklist

- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [x] I have commented on my code, particularly in hard-to-understand areas
- [x] My changes generate no new warnings
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] Any dependent changes have been merged and published in downstream modules

PR #2408, opened 2026-09-22
okx: forward the order ID filter in getTransactionDetails

Robin1987China:
### What this fixes

`getTransactionDetails` never sends `TransactionDetailRequestParams.OrderID`, so
a caller asking for the fills of one specific order silently receives every fill
in the requested window instead. The response is unmarshalled without error, so
the caller has no way to notice.

It likewise never sends `SubType`, so the transaction-type filter is dropped the
same way.

All other parameters of the request struct are forwarded, and both are declared
with the JSON names the endpoint expects (`ordId`, `subType`):

    type TransactionDetailRequestParams struct {
        ...
        OrderID   string    `json:"ordId"`
        OrderType string    `json:"orderType"`
        SubType   string    `json:"subType,omitempty"`
        ...
    }

The parameter block also set `instId` twice — the second block is where these
filters belong, which suggests a copy/paste slip:

    if arg.Limit > 0 {
        params.Set("limit", strconv.FormatInt(arg.Limit, 10))
    }
    if arg.InstrumentID != "" {          // duplicate
        params.Set("instId", arg.InstrumentID)
    }

### The change

The duplicated `instId` block is replaced by forwarding the order ID as `ordId`
and the transaction type as `subType`, fixing the dropped filters and the
duplication in one place. Both are optional: when empty, neither is sent, so
existing callers are unaffected.

`OrderType` remains unforwarded — it is not a request parameter of the fills
endpoints.

### Testing

New offline test `TestGetTransactionDetailsOrderID` asserts the outgoing query
contains `ordId` and `subType` (alongside `instType` and `limit`) for the
requested order, and that the request is a `GET` to `/trade/fills`. It uses an
`httptest` server with `SkipAuthCheck`, so it needs no credentials or network
access.

Without this change the `ordId` and `subType` assertions receive an empty string;
with this change they pass. (Reverting the `subType` forwarding fails with
`expected: "2" / actual: ""`.)

The pre-existing tests for this path are credential gated, so these filters had
no coverage before this change.

Verified locally: `go build ./...`, `go vet`, `gofmt -l` (empty),
`go test -race -count 1` on the affected package, `scripts/misc_checks.sh`, and
golangci-lint (0 issues).

Fixes #

## Type of change

## How has this been tested

- [x] `go build ./...`
- [x] `go test ./exchanges/okx/... -race -count 1` — package suites pass; `TestBusinessWSCandleSubscriptions`
      fails on this machine with a websocket connection reset, and fails identically on the base commit
- [x] `golangci-lint run ./exchanges/okx/...` — 0 issues

The test pins `subType` the way it already pins `ordId`, and asserts the request is a `GET` to
`/trade/fills`. Reverting the `subType` forwarding fails the assertion (`expected: "2" / actual: ""`), and
pointing the 3-day route at `trade/fills-history` fails the path assertion.

## Checklist


thrasher-:
@codex please review

exchanges/okx/okx.go
```diff
@@ -346,8 +346,8 @@ func (e *Exchange) getTransactionDetails(ctx context.Context, arg *TransactionDe
 	if arg.Limit > 0 {
 		params.Set("limit", strconv.FormatInt(arg.Limit, 10))
 	}
-	if arg.InstrumentID != "" {
-		params.Set("instId", arg.InstrumentID)
+	if arg.OrderID != "" {
+		params.Set("ordId", arg.OrderID)
 	}
```
thrasher-, line 351: Heimdall: **Blocker** — `subType` is a documented request parameter of both fills endpoints, so [`TransactionDetailRequestParams.SubType`](https://github.com/thrasher-corp/gocryptotrader/blob/233577c55a8c8156b5124ec4867f231759c38f84/exchanges/okx/okx_types.go#L1051) is a second silently-dropped filter, identical in kind to the `ordId` one this PR fixes.

**Fix:** forward it alongside the order ID.

```suggestion
	if arg.OrderID != "" {
		params.Set("ordId", arg.OrderID)
	}
	if arg.SubType != "" {
		params.Set("subType", arg.SubType)
	}
```

**Why:** the scope note is right that `subType` is a response field, but it is a request filter as well. OKX lists it in the Request Parameters table for [`GET /api/v5/trade/fills`](https://www.okx.com/docs-v5/en/#order-book-trading-trade-get-transaction-details-last-3-days) and [`GET /api/v5/trade/fills-history`](https://www.okx.com/docs-v5/en/#order-book-trading-trade-get-transaction-details-last-3-months) alike — optional, "Transaction type: 1 Buy, 2 Sell, 3 Open long, …". Probing this head on both routes:

```
trade/fills          ordId="312269865356374016" subType="" orderType=""
trade/fills-history  ordId="312269865356374016" subType="" orderType=""
```

So setting `SubType` today returns every fill in the window, unfiltered and without an error — the failure this PR opens with. `orderType` is a different story: neither table lists it, so leaving that field alone is right.

Applies as a pair with [the test suggestion](https://github.com/thrasher-corp/gocryptotrader/pull/2408#discussion_r4071199185). I ran the pair in a worktree: `go test ./exchanges/okx/... -race`, `golangci-lint` v2.13.2 and `scripts/misc_checks.sh` all pass, and deleting this block again fails the new `subType` assertion with `expected: "2" / actual: ""`.

exchanges/okx/okx_test.go
```diff
@@ -1110,6 +1113,45 @@ func TestTransactionHistory(t *testing.T) {
...
+		mutex.Unlock()
+		w.Header().Set("Content-Type", "application/json")
+		_, _ = w.Write([]byte(`{"code":"0","msg":"","data":[]}`))
+	}))
+	defer server.Close()
+
+	ex := new(Exchange)
+	require.NoError(t, testexch.Setup(ex), "Setup must not error")
+	ex.SkipAuthCheck = true
+	require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL+"/"), "SetRunningURL must not error")
```
thrasher-: Heimdall: **Should fix** — the test never asserts the method or the path, so it does not prove that the *3-day* endpoint was called. I pointed [`GetTransactionDetailsLast3Days`](https://github.com/thrasher-corp/gocryptotrader/blob/233577c55a8c8156b5124ec4867f231759c38f84/exchanges/okx/okx.go#L314-L316) at `trade/fills-history` at this head and it still passed.

**Fix:** two points, but they land on the same lines, so they are one suggestion.

```suggestion
	server := httptest.NewTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		assert.Equal(t, http.MethodGet, r.Method, "the fills request should be a GET")
		assert.Equal(t, "/trade/fills", r.URL.Path, "GetTransactionDetailsLast3Days should call the 3-day fills endpoint")
		mutex.Lock()
		query = r.URL.Query()
		mutex.Unlock()
		w.Header().Set("Content-Type", "application/json")
		_, _ = w.Write([]byte(`{"code":"0","msg":"","data":[]}`))
	}))

	ex := new(Exchange)
	require.NoError(t, testexch.Setup(ex), "Setup must not error")
	ex.SkipAuthCheck = true
	require.NoError(t, ex.SetHTTPClient(server.Client()), "SetHTTPClient must not error")
	require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL+"/"), "SetRunningURL must not error")
```

**Why:** the path assertion is the substantive half — with it, that same swap fails with `expected: "/trade/fills" / actual: "/trade/fills-history"`.

The other half is [`httptest.NewTestServer`](https://pkg.go.dev/net/http/httptest#NewTestServer): it serves over an in-memory network and registers its own `Close`, so `defer server.Close()` goes away and the test cannot reach the network at all. The repo has 45 `NewTestServer` call sites and two `httptest.NewServer` — this one and [`exchanges/request/request_test.go:206`](https://github.com/thrasher-corp/gocryptotrader/blob/2689feebbac7bdfed1af23ec86eb7519bec27974/exchanges/request/request_test.go#L206), which is inside a `TestMain` with no `*testing.T` to hand; [`kraken_test.go:199`](https://github.com/thrasher-corp/gocryptotrader/blob/2689feebbac7bdfed1af23ec86eb7519bec27974/exchanges/kraken/kraken_test.go#L199-L223) is the closest precedent for what this test does. The handler is reachable only through `server.Client()`, hence the added `SetHTTPClient`, and that call has to come before `server.URL` is read because `NewTestServer` does not start the server until `Client()` is first called. Verified: removing `SetHTTPClient` fails the call, and with the fake network in place removing `SetRunningURL` would otherwise still pass — the path assertion catches that too (`expected: "/trade/fills" / actual: "/api/v5/trade/fills"`).

exchanges/okx/okx_test.go
```diff
@@ -1110,6 +1113,45 @@ func TestTransactionHistory(t *testing.T) {
...
+		OrderID:        orderID,
+		Limit:          3,
+	})
+	require.NoError(t, err, "GetTransactionDetailsLast3Days must not error")
+	mutex.Lock()
+	defer mutex.Unlock()
+	require.NotNil(t, query, "a request must have been sent")
+	assert.Equal(t, orderID, query.Get("ordId"), "GetTransactionDetailsLast3Days should forward the order ID filter as ordId")
+	assert.Equal(t, instTypeSpot, query.Get("instType"), "GetTransactionDetailsLast3Days should forward the instrument type")
+	assert.Equal(t, "3", query.Get("limit"), "GetTransactionDetailsLast3Days should forward the limit")
```
thrasher-, line 1156: Heimdall: **Blocker** (pairs with [the `subType` fix](https://github.com/thrasher-corp/gocryptotrader/pull/2408#discussion_r4071198776)) — pin `subType` the way the test already pins `ordId`, so the second dropped filter cannot come back unnoticed.

**Fix:**

```suggestion
	_, err := ex.GetTransactionDetailsLast3Days(t.Context(), &TransactionDetailRequestParams{
		InstrumentType: instTypeSpot,
		OrderID:        orderID,
		SubType:        "2",
		Limit:          3,
	})
	require.NoError(t, err, "GetTransactionDetailsLast3Days must not error")
	mutex.Lock()
	defer mutex.Unlock()
	require.NotNil(t, query, "a request must have been sent")
	assert.Equal(t, orderID, query.Get("ordId"), "GetTransactionDetailsLast3Days should forward the order ID filter as ordId")
	assert.Equal(t, "2", query.Get("subType"), "GetTransactionDetailsLast3Days should forward the transaction type filter as subType")
	assert.Equal(t, instTypeSpot, query.Get("instType"), "GetTransactionDetailsLast3Days should forward the instrument type")
	assert.Equal(t, "3", query.Get("limit"), "GetTransactionDetailsLast3Days should forward the limit")
```

**Why:** `2` is OKX's documented `subType` for a sell fill. Applied on its own this fails (`expected: "2" / actual: ""`); applied together with the `okx.go` suggestion it passes under `-race -count 2`.

thrasher- (changes requested):
Heimdall: Good catch — the duplicate `instId` block traces to the original OKX import (#1005), and dropping `ordId` quietly turned "the fills of this one order" into "every fill in the window". Reverting [`okx.go:349-351`](https://github.com/thrasher-corp/gocryptotrader/blob/233577c55a8c8156b5124ec4867f231759c38f84/exchanges/okx/okx.go#L349-L351) fails your test exactly as you describe, and `ordId` now reaches the wire on both routes.

Two things inline: `subType` **is** a request parameter of both fills endpoints, so `SubType` is a second silently-dropped filter of the same class (you're right about `orderType`); and the new test can't tell `trade/fills` from `trade/fills-history`.

Lint, `-race` tests and misc checks are green at this head, as is CI; master has only moved in `exchanges/gateio` since your base. Out of scope, happy to file with tested patches: `uly` is retired in favour of `instFamily` package-wide, `instType` is optional for the 3-day route, `TransactionDetail` lacks `feeRate` and `tradeQuoteCcy`, and `getTransactionDetails` panics on a nil `arg`.

**To unblock:**
1. Forward `subType` alongside `ordId` ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2408#discussion_r4071198776)) and pin it in the new test ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2408#discussion_r4071199185)).

Robin1987China:
Pushed `3ec0742`. `subType` is now forwarded alongside `ordId` in `getTransactionDetails`, and the test pins it and asserts the request is a GET to `/trade/fills`, so the 3-day route can no longer be swapped for the history one unnoticed.

exchanges/okx/okx_test.go
```diff
@@ -1110,6 +1113,49 @@ func TestTransactionHistory(t *testing.T) {
 	require.NotNil(t, result)
 }
 
+// TestGetTransactionDetailsOrderID covers forwarding the order ID filter to the
+// fills endpoint. Before the fix the order ID was dropped, so a caller asking for
+// the fills of a single order silently received every fill in the requested
+// window.
+func TestGetTransactionDetailsOrderID(t *testing.T) {
```
thrasher-, line 1120: Heimdall: **Nit** — the test now pins `subType` as well as `ordId`, so the name and doc comment only tell half of it.

**Fix:** rename it to something like `TestGetTransactionDetailsFilters`, and mention `subType` in the comment. The PR title could say "order ID and subType filters" too.

thrasher- (approved):
Heimdall: Approving. The fix is real: with only `okx.go` reverted to the base, `TestGetTransactionDetailsOrderID` fails on both `ordId` (`actual: ""`) and `subType` (`actual: ""`); with the change it passes under `-race -count 2`. Both names match OKX's documented parameters for `/trade/fills` and `/trade/fills-history`, and the removed `instId` block was a duplicate `Set`, so no request changes.

Inline, not blocking: a Nit on the [test name](https://github.com/thrasher-corp/gocryptotrader/pull/2408#discussion_r4077400767).

- `instType` is required locally for both routes, but OKX only requires it for the 3-month one: [okx.go:329](https://github.com/thrasher-corp/gocryptotrader/blob/3ec0742abdb60b7c09998517b4b7dccb42ccec19/exchanges/okx/okx.go#L329). `{OrderID: …}` alone is rejected before any request is sent.
- `instFamily` is documented on both routes but never sent. `uly`, which OKX no longer documents, is still sent: [okx.go:337](https://github.com/thrasher-corp/gocryptotrader/blob/3ec0742abdb60b7c09998517b4b7dccb42ccec19/exchanges/okx/okx.go#L337).
- `OrderType` is not a parameter of either endpoint and is never read: [okx_types.go:1050](https://github.com/thrasher-corp/gocryptotrader/blob/3ec0742abdb60b7c09998517b4b7dccb42ccec19/exchanges/okx/okx_types.go#L1050).

The failing CI job is ambient: Bitflyer's live `TestGetAddressInfoCA`/`TestGetTransactionByHashCA` timed out ([job 106744507936](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35727461024/job/106744507936)), and nothing in OKX failed. Happy for any of the above to land here; otherwise we'll open a separate issue.

PR #2409, opened 2026-09-22
lbank: skip orders that left the book and reject order types that cannot be mapped

Robin1987China:
### What this changes

Two related problems in the LBank wrapper, both reachable in normal operation.

**1. `GetActiveOrders` and `GetOrderInfo` indexed `Orders[0]` without checking the slice**

`QueryOrder` leaves `Orders` nil when the exchange reports no orders
(`OrderHistoryResponse.Orders` is a `json.RawMessage` that is only unmarshalled
when `len(raw) > 2`), so both wrappers did `tempResp.Orders[0]` and could panic:

```
panic: runtime error: index out of range [0] with length 0
```

This is reachable: `GetActiveOrders` lists the open orders first and queries them
one at a time, so any order that fills or cancels in between is queried after it
has left the book; `GetOrderInfo` has the same gap for a named order.

- `GetActiveOrders` now skips an order the exchange no longer reports (`continue`).
- `GetOrderInfo` now returns `order.ErrOrderNotFound`, both when the exchange
  returns no order for the ID and when the ID is not in the open-order listing at
  all (previously it returned a zero-value `order.Detail` with a nil error).

**2. The order type was mapped with a two-way `if`, so anything that was not `buy` became `sell`**

```go
if strings.EqualFold(tempResp.Orders[0].Type, order.Buy.String()) {
    resp.Side = order.Buy
} else {
    resp.Side = order.Sell
}
```

LBank echoes the order type in compound form (e.g. `buy_maker`, `sell_market`), so
only the token before the `_` names the side. An empty, misspelled or newly added
value was silently reported as a sell. This is replaced with the repo's own
`order.StringToOrderSide` (`exchanges/order/orders.go:1015`) applied to the
leading token, in `GetOrderInfo`, `GetActiveOrders` and `GetOrderHistory`, and the
`ErrSideIsInvalid` error is now returned to the caller.

### Scope

Only `exchanges/lbank/lbank_wrapper.go` and its test change. `GetOrderInfo` still
does not set `resp.OrderID`; that is a separate contract question left for its own
discussion.

### Tests

Eight offline tests, no network and no credentials, using the repo's usual fixture
idiom (`httptest.NewTestServer` + `SetHTTPClient(server.Client())` +
`SetRunningURL(exchange.RestSpot.String(), server.URL)`):

| Test | Setup | Asserts |
|---|---|---|
| `TestGetActiveOrdersEmptyResponse` | order query answers `"orders":[]` | no error, no panic, nothing returned |
| `TestGetActiveOrdersCompoundOrderType` | `"type":"buy_maker"` | side mapped to buy |
| `TestGetActiveOrdersUnknownSide` | `"type":"hold"` | `order.ErrSideIsInvalid` |
| `TestGetOrderInfoEmptyResponse` | order query answers `"orders":[]` | `order.ErrOrderNotFound` |
| `TestGetOrderInfoUnknownSide` | `"type":"hold"` | `order.ErrSideIsInvalid` |
| `TestGetOrderInfoCompoundOrderType` | `"type":"sell_market"` | side mapped to sell |
| `TestGetOrderInfoOrderIDNotFound` | order ID absent from the open listing | `order.ErrOrderNotFound` |
| `TestGetOrderHistoryCompoundOrderType` | history page with a plain order then `sell_market` | both kept; sides buy then sell |

Each of the three side-mapping sites has its own regression test. Reverting the
`strings.Cut` at the history site turns `TestGetOrderHistoryCompoundOrderType`
red.

### A note on the test transport

These are authenticated wrappers, and `SendAuthHTTPRequest` assigns the caller's
path straight to `request.Item.Path` without the configured endpoint URL, so a
relative request never reaches the wire (`Post "/v2/orders_info.do": unsupported
protocol scheme ""`). That is pre-existing and is being fixed separately, not
here. To keep the tests on a real `http.Client` and a real server in the meantime,
they install a small adapter (`orderGuardTransport`) that resolves a host-less
request URL against the test server and passes anything that already carries a
host straight through. It is not a mock: the request really is sent and the
response really comes from the test server.

### Verification

```
go build ./...                                             # ok
go test ./exchanges/lbank/ -race -count 1                  # ok
go vet ./exchanges/lbank/                                  # ok
gofmt -l exchanges/lbank/                                  # empty
bash scripts/misc_checks.sh                                # All miscellaneous checks passed!
golangci-lint run ./exchanges/lbank/...                    # 0 issues
```

---

Fixes #

## Type of change

## How has this been tested

- [x] `go build ./...`
- [x] `go test ./exchanges/lbank/... -race -count 1`
- [x] `golangci-lint run ./exchanges/lbank/...` — 0 issues

Each of the three mapping sites has its own regression test, including `TestGetOrderHistoryCompoundOrderType`
for the history path. Reverting the history-site `strings.Cut` turns its own test red.

## Checklist

```

---

## 还原确认
- 变异测试：#2409 删 674 Cut → 红 → `git restore` 复绿；#2408 scratch worktree 撤 subType → 红 → worktree 删除。
- scratch worktree `/tmp/k3-base2407`、`/tmp/k3-mut2408`、`/tmp/k3-basecheck` 均已 `git worktree remove`；探针文件随 worktree 一并删除；`/tmp/k3-*.log` 已清理。
- `gct-v5` / `gct-v2` / `gct-audit` `git status` 均为空（干净）。未 push、未发布、未改任何 PR。

### CODE REVIEW

**Comment by thrasher-** [comment]
@codex please review

File: exchanges/lbank/lbank_wrapper.go [hunk 1]
```diff
@@ -567,16 +571,20 @@ func (e *Exchange) GetActiveOrders(ctx context.Context, getOrdersRequest *order.
...
 			resp.Pair, err = currency.NewPairFromString(key)
 			if err != nil {
 				return nil, err
 			}
 
-			if strings.EqualFold(tempResp.Orders[0].Type, order.Buy.String()) {
-				resp.Side = order.Buy
-			} else {
-				resp.Side = order.Sell
+			resp.Side, err = order.StringToOrderSide(tempResp.Orders[0].Type)
```
thrasher-: Heimdall: **Blocker** — `order.StringToOrderSide` cannot map eight of LBank's ten documented order types, and when it fails here the whole listing goes with it.

**Fix:**

```suggestion
			side, _, _ := strings.Cut(tempResp.Orders[0].Type, "_")
			resp.Side, err = order.StringToOrderSide(side)
```

**Why:** LBank's order `type` is one of `buy, sell, buy_market, sell_market, buy_maker, sell_maker, buy_ioc, sell_ioc, buy_fok, sell_fok` ([place an order](https://www.lbank.com/en-US/docs/index.html#place-an-order)), and `orders_info.do` echoes the compound form back — ccxt's captured sample for that exact path is `"type":"sell_market"`. [`StringToOrderSide`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/order/orders.go#L1015-L1034) accepts only BUY/SELL/BID/ASK/LONG/SHORT/ANY, so all eight compound values return `ErrSideIsInvalid`.

`*_maker` is post-only — a limit order that **rests on the book**, which is exactly what this loop enumerates. [`CreateOrder`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank.go#L195-L197) only ever sends `buy`/`sell`, so these arrive from orders placed on the same account through LBank's own UI or app. Measured on your branch against master:

```
PR      type=buy_maker   orders=0   err="BUY_MAKER" order side is invalid
master  type=buy_maker   orders=1   side=SELL       err=<nil>
```

and a mixed book loses the good order too — `[1:buy, 2:buy_maker]` returns `(1 order, error)` on the PR against 2 orders on master. Callers discard the slice when `err != nil`, so the effect is total: [`engine/order_manager.go:704-710`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/engine/order_manager.go#L704-L710) logs and `continue`s the asset loop — lbank is spot-only, so the entire active-order sync stops every cycle for as long as that order rests — and [`engine/rpcserver.go:933-935`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/engine/rpcserver.go#L933-L935) hands the error straight to the RPC caller. master mislabels one order's side; this drops the listing.

`strings.Cut` keeps your rejection intent — `""` and `"hold"` still fail — so `TestGetActiveOrdersUnknownSide` and `TestGetOrderInfoUnknownSide` stay green. The discarded suffix is the order *type*, which neither function assigns today, so nothing is lost. I applied it at both sites: build ok, all four new tests pass, `gofmt` clean, `golangci-lint v2.13.2` 0 issues.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -479,16 +479,20 @@ func (e *Exchange) GetOrderInfo(ctx context.Context, orderID string, _ currency.
...
```
thrasher-: Heimdall: **Blocker** — same mapping problem as in `GetActiveOrders`; here it turns a normal post-only or market order into an error instead of returning it.

**Fix:**

```suggestion
			side, _, _ := strings.Cut(tempResp.Orders[0].Type, "_")
			resp.Side, err = order.StringToOrderSide(side)
```

**Why:** see the detail on the `GetActiveOrders` copy. Measured here: `type=buy_maker` → `side=UNKNOWN err="BUY_MAKER" order side is invalid`; with the `Cut` it is `side=BUY err=<nil>`.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -479,16 +479,20 @@ func (e *Exchange) GetOrderInfo(ctx context.Context, orderID string, _ currency.
 			if err != nil {
 				return nil, err
 			}
+			if len(tempResp.Orders) == 0 {
+				// The order left the book between the open order listing and this
```
thrasher-: Heimdall: **Should fix** — the guards are right, but the reason both comments give for the empty slice isn't the mechanism.

**Fix:** reword this one and [its twin in `GetActiveOrders`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L575-L576) to say what is actually known — the exchange returned no order for this ID — rather than asserting why.

**Why:** `orders_info.do` is a *query* endpoint, not an open-order endpoint: it keeps answering for an order after it has left the book. ccxt's captured sample for that exact path carries `"status":2` with `deal_amount` equal to `amount`, i.e. a completely filled order still being returned, and this package's own [`GetStatus`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L856-L876) maps `2` to `order.Filled` and `-1` to `order.Cancelled` — so an order that filled or cancelled between the listing and the query is precisely the case that *does* still come back.

The realistic triggers for an empty slice are an ID the exchange no longer recognises, or [`QueryOrder`'s `len(…) > 2` check](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank.go#L256-L262) leaving `Orders` nil. The repo asks comments to explain *why*, and a confident wrong *why* is worse than none — it's what the next reader will reason from.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -479,16 +479,20 @@ func (e *Exchange) GetOrderInfo(ctx context.Context, orderID string, _ currency.
...
+				// query, so the exchange has nothing left to report for it.
+				return nil, fmt.Errorf("%w %v", order.ErrOrderNotFound, orderID)
```
thrasher-, line 484: Heimdall: **Should fix** — introducing `ErrOrderNotFound` here leaves the function reporting "not found" two different ways.

**Why:** if `orderID` never matches an entry from `getAllOpenOrderID`, both loops simply fall through to [`return &resp, nil`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L513) and the caller gets a zero-valued `*order.Detail` with a **nil error**. Measured on your branch:

```
GetOrderInfo("does-not-exist") -> side="UNKNOWN" status="UNKNOWN" pair="" amount=0 exchange=""  err=<nil>
```

Before this PR both paths were wrong in their own way, so they were at least consistent. Now one is right and the other silently returns an empty order — and your own test name, "should report an order the exchange no longer holds", sets the expectation the second path breaks. The sentinel is already imported and in scope.

I've deliberately not written this as a suggestion: it changes the contract for callers that today get `(&order.Detail{}, nil)`, and it wants a test of its own, so how to land it is your call.

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +382,155 @@ func TestGetOrderHistory(t *testing.T) {
...
+	ex.API.AuthenticatedSupport = true
+	ex.SkipAuthCheck = true
+	ex.SetCredentials(&accounts.Credentials{Key: "mock-key", Secret: "mock-secret"})
+	var err error
+	ex.privateKey, err = rsa.GenerateKey(rand.Reader, 2048)
+	require.NoError(t, err, "the RSA test key must generate")
+	require.NoError(t, ex.CurrencyPairs.StorePairs(asset.Spot, currency.Pairs{testPair}, false), "StorePairs must not error for available pairs")
+	require.NoError(t, ex.CurrencyPairs.StorePairs(asset.Spot, currency.Pairs{testPair}, true), "StorePairs must not error for enabled pairs")
+	require.NoError(t, ex.SetHTTPClient(orderGuardClient(server)), "SetHTTPClient must not error")
+	require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpot.String(), server.URL), "SetRunningURL must not error")
```
thrasher-: Heimdall: **Minor** — this line does nothing.

**Why:** `SendAuthHTTPRequest` never consults the configured endpoint — it assigns the relative path straight to `request.Item.Path` — and [`orderGuardTransport`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_test.go#L516-L532) then resolves that against `serverURL` itself. I removed the line and re-ran: the tests still pass and still reach the server. Worth dropping, or worth a word saying why it is kept, since as it stands it suggests the endpoint wiring is what routes these requests.

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +382,155 @@ func TestGetOrderHistory(t *testing.T) {
...
+			if r.Form.Get("current_page") == "1" {
+				body = orderGuardOpenOrder("1")
+			} else {
+				body = orderGuardNoOrders
+			}
+		case strings.HasSuffix(r.URL.Path, lbankQueryOrder):
+			body = queryResponse
+		default:
+			assert.Failf(t, "unexpected endpoint", "the wrapper should only request open orders and order queries, got %s", r.URL.Path)
+			return
```
thrasher-, line 549: Heimdall: **Minor** — an unexpected endpoint fails twice over.

**Fix:**

```suggestion
		default:
			assert.Failf(t, "unexpected endpoint", "the wrapper should only request open orders and order queries, got %s", r.URL.Path)
			w.WriteHeader(http.StatusNotFound)
			return
```

**Why:** `assert.Failf` records the failure and returns, but the handler then writes nothing, so the client gets a 200 with an empty body and the wrapper fails again on a JSON decode error. Whoever adds the next endpoint to this package sees two unrelated failures and has to work out which one is the real one.

exchanges/lbank/lbank_wrapper.go
thrasher-, line 1: Heimdall: **Should fix** — `GetOrderHistory` keeps the mapping this PR is removing, and the reason given for leaving it doesn't hold.

**Why:** [line 667](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L667-L671) is still `if strings.EqualFold(Type, order.Buy.String()) { Buy } else { Sell }`. History is where the compound types are *most* common — ccxt's captured `orders_info.do` sample is `"type":"sell_market"`, and a market order only ever appears in history, never in an open-order listing. After this PR the same field is handled two different ways in one file.

On the conflict with `#2405`: I measured it. `#2405`'s wrapper hunks are at 620, 637-656 and 679-685, so they never meet line 667. I merged `#2405` into this branch **with 667 also fixed** and `lbank_wrapper.go` auto-merged — the only conflict was `lbank_test.go`, and that conflicts either way, because both PRs insert their helpers at the same point in the file. `#2405` doesn't touch the mapping, so as things stand it survives both PRs.

Same two lines as the other two sites:

```go
				side, _, _ := strings.Cut(tempResp.Orders[x].Type, "_")
				resp.Side, err = order.StringToOrderSide(side)
				if err != nil {
					return nil, err
				}
```

Not a suggestion block because it can't anchor to an unchanged line. If you'd rather keep this PR to the two functions in the title, say so and I'll carry it into the follow-up instead.

thrasher-, line 1: Heimdall: **Pre-existing, not yours** — four things I measured in this call chain while checking the PR. None of them block it. Happy to take the lot to a follow-up issue with the tested fixes; say the word and I'll open it rather than grow this PR.

**1. `getAllOpenOrderID` has the very panic you are fixing.** [`tempData`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L742) holds the *previous* page's length while `tempResp` is the current page, so [`for c := range tempData`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L756) indexes past the end as soon as a page comes back short. A 200-order page followed by a 50-order one:

```
panic: runtime error: index out of range [50] with length 50
```

The same loop also [`return`s instead of `break`ing](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L752-L754) when a pair runs out of pages, so it abandons every remaining pair, and it fetches page 1 twice. With two enabled spot pairs:

```
ids  = map[btc_usdt:[1]]                       # eth_usdt never queried
calls = [btc_usdt#1 btc_usdt#1 btc_usdt#2]
```

One `for b := int64(1); ; b++` loop ranging over `tempResp.Orders` fixes all three — measured afterwards: 250 ids, both pairs, one fetch per page, your four tests still green.

**2. Neither function sets `OrderID`, `AssetType` or `Type`.** Measured on the happy path: `OrderID="" AssetType="" Type="UNKNOWN"`. The order manager matches store entries by `OrderID`, so every lbank active order collapses onto one empty-ID record.

**3. The side filter compares against the raw string.** [Line 612](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L612-L615) tests `getOrdersRequest.Side.String()` against `Orders[0].Type`, so even with the mapping fixed, `Side: order.Buy` returns nothing for a `buy_maker` order — measured. It's redundant with [`getOrdersRequest.Filter`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_wrapper.go#L619) two lines later, which already filters on `resp.Side`; deleting it makes that case return the order.

**4. The order structs look stale against the vendor.** ccxt reads `data` where [`QueryOrderResponse`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_types.go#L96-L99) reads `orders`, and `create_time` where [`OrderResponse`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_types.go#L86) reads `created_time`, and its captured samples for this endpoint carry a bare `"result":true` against [`json:"result,string"`](https://github.com/thrasher-corp/gocryptotrader/blob/73e6c2f40437a669e258fe7c3e6d238eb571b0b8/exchanges/lbank/lbank_types.go#L219). I fed the wrapper ccxt's verbatim payload and got `json: cannot unmarshal bool into Go struct field QueryOrderResponse.result of type bool`. I can't settle this one without API credentials, so I'm flagging it as a question rather than a finding — but if it is right, it's a bigger problem than anything above.

thrasher- (changes requested):
Heimdall: The two guards are the right fix and the evidence behind them is the good kind — I reproduced both panics on master using your own fixture, and all three of your mutations turn the new tests red, so they have real teeth. The asymmetric treatment (`continue` for a listing, `ErrOrderNotFound` for a named order) is the right call, and the claim behind the transport adapter holds: `SendAuthHTTPRequest` assigns the caller's path straight to `request.Item.Path`, so a relative request never leaves the client.

One blocker. `order.StringToOrderSide` rejects eight of LBank's ten documented order `type` values, and `*_maker` is post-only — a limit order that rests on the book, so it is exactly what `GetActiveOrders` enumerates. Measured on your branch, one such order resting:

```
PR      type=buy_maker   orders=0   err="BUY_MAKER" order side is invalid
master  type=buy_maker   orders=1   side=SELL       err=<nil>
```

Callers drop the slice on error, so the whole lbank active-order sync stops every cycle while that order rests. The mechanism and a tested one-line fix are in the comments; everything else — two should-fixes, two minors, and a set of pre-existing defects in the same call chain that I'd rather take to a follow-up issue than load onto this PR — is inline too.

CI: the two red jobs are in `exchanges/gateio` and `exchanges/subscription`, neither of which this PR touches, and the windows one fails on `#2405` too. Locally at head, and again with every suggestion applied: `go test ./exchanges/lbank/ -race` green (the live-network `TestGetHistoricCandlesExtended` timed out once and passed on both re-runs), `golangci-lint v2.13.2` 0 issues, `scripts/misc_checks.sh` all pass, `TestAllExchangeWrappers/lbank` ok.

**To unblock:**
1. Map the side from the token before the `_`, at both call sites ([comment](https://github.com/thrasher-corp/gocryptotrader/pull/2409#discussion_r4071244847), [comment](https://github.com/thrasher-corp/gocryptotrader/pull/2409#discussion_r4071245023))

Robin1987China:
Pushed `9eb6c8f`. Both call sites now map the side from the token before the `_`, so a resting `buy_maker` order no longer aborts the whole active-order listing and `sell_market` no longer fails `GetOrderInfo`. The two should-fixes and the two minors are in as well.

I also took the third `GetOrderHistory` site in this PR rather than moving it to a follow-up, for consistency within the file — but it has no dedicated test yet, so say the word and I will add one, or drop it back to the follow-up.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -567,16 +574,22 @@ func (e *Exchange) GetActiveOrders(ctx context.Context, getOrdersRequest *order.
...
 				return nil, err
 			}
 
-			if strings.EqualFold(tempResp.Orders[0].Type, order.Buy.String()) {
-				resp.Side = order.Buy
-			} else {
-				resp.Side = order.Sell
+			// orders_info.do echoes the order type back in its compound form, e.g.
+			// buy_maker, so only the leading token names the side.
+			side, _, _ := strings.Cut(tempResp.Orders[0].Type, "_")
```
thrasher-, line 589: Heimdall: **Should fix** — the side filter further down this function, at [lbank_wrapper.go:617](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/lbank/lbank_wrapper.go#L617), still compares the requested side against the raw `tempResp.Orders[0].Type`. So a side-filtered request silently drops every compound-typed order that this line now maps correctly.

**Fix:** drop the manual side check and append unconditionally. `getOrdersRequest.Filter` already filters by side at [orders.go:1246](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/order/orders.go#L1246), using the `resp.Side` mapped here:

```go
for y := range getOrdersRequest.Pairs {
	if getOrdersRequest.Pairs[y].String() != key {
		continue
	}
	finalResp = append(finalResp, resp)
}
```

**Why:** `GetActiveOrders` against the PR's own `setupOrderGuard` fixture, one open order, before and after that change:

```
type        requested side   HEAD            with fix
buy         BUY              1 order         1 order
buy_maker   BUY              0 orders, nil   1 order
sell_maker  SELL             0 orders, nil   1 order
buy_maker   SELL             0 orders        0 orders
```

The PR's tests all pass with the fix. A case in `TestGetActiveOrdersCompoundOrderType` requesting `Side: order.Buy` would cover it.

exchanges/lbank/lbank_test.go
```diff
@@ -378,6 +382,235 @@ func TestGetOrderHistory(t *testing.T) {
...
+
+	_, err := ex.GetOrderInfo(t.Context(), "1", testPair, asset.Spot)
+	assert.ErrorIs(t, err, order.ErrSideIsInvalid, "GetOrderInfo should reject an order type it cannot map")
+}
+
+// TestGetOrderInfoCompoundOrderType ensures a compound order type LBank
+// documents, such as sell_market, is mapped to its side instead of erroring.
+func TestGetOrderInfoCompoundOrderType(t *testing.T) {
+	t.Parallel()
+	ex := setupOrderGuard(t, orderGuardHandler(t, orderGuardSingleOrder("sell_market")))
```
thrasher-, line 458: Heimdall: **Should fix** — this test and `TestGetOrderHistoryCompoundOrderType` both pass on the base commit [`158e9c1`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322). The old rule mapped everything that wasn't `buy` to Sell, so it also maps `sell_market` to Sell. As written, neither test would catch a return to the original bug.

**Fix:** use a buy-side compound type here and at the history fixture ([lbank_test.go:512](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/lbank/lbank_test.go#L512)), e.g. `buy_market`, and expect `order.Buy`.

**Why:** the PR's `lbank_test.go` run against the base wrapper:

```
TestGetActiveOrdersEmptyResponse       panic: index out of range
TestGetActiveOrdersCompoundOrderType   FAIL
TestGetActiveOrdersUnknownSide         FAIL
TestGetOrderInfoEmptyResponse          panic: index out of range
TestGetOrderInfoUnknownSide            FAIL
TestGetOrderInfoOrderIDNotFound        FAIL
TestGetOrderInfoCompoundOrderType      ok    ← passes on base
TestGetOrderHistoryCompoundOrderType   ok    ← passes on base
```

With both fixtures switched to `buy_market`, both tests fail on base and pass on HEAD.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -656,10 +669,12 @@ func (e *Exchange) GetOrderHistory(ctx context.Context, getOrdersRequest *order.
 					return nil, err
 				}
 
-				if strings.EqualFold(tempResp.Orders[x].Type, order.Buy.String()) {
-					resp.Side = order.Buy
-				} else {
-					resp.Side = order.Sell
+				// orders_info.do echoes the order type back in its compound form, e.g.
```
thrasher-, line 672: Heimdall: **Nit** — this path reads from `QueryOrderHistory`, which calls `orders_info_history.do` ([lbank.go:286](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/lbank/lbank.go#L286)), not `orders_info.do`. The comment should name that endpoint.

exchanges/lbank/lbank_wrapper.go
```diff
@@ -479,16 +479,22 @@ func (e *Exchange) GetOrderInfo(ctx context.Context, orderID string, _ currency.
...
 				return nil, err
 			}
 
-			if strings.EqualFold(tempResp.Orders[0].Type, order.Buy.String()) {
-				resp.Side = order.Buy
-			} else {
-				resp.Side = order.Sell
+			// orders_info.do echoes the order type back in its compound form, e.g.
+			// buy_maker, so only the leading token names the side.
+			side, _, _ := strings.Cut(tempResp.Orders[0].Type, "_")
```
thrasher-, line 494: Heimdall: **Nit** — the same `Cut` then `StringToOrderSide` code appears three times (here, [lbank_wrapper.go:589](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/lbank/lbank_wrapper.go#L589) and [lbank_wrapper.go:674](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/lbank/lbank_wrapper.go#L674)). A small helper would keep the three in step. It could also accept only `buy` and `sell`: `StringToOrderSide` takes more than LBank sends, so these all map without an error today:

```
any_maker -> ANY    bid -> BID    long -> LONG    buy_ -> BUY
```

thrasher- (approved):
Heimdall: Approving. Both panics are real and fixed: the empty-response tests panic with index out of range on [`158e9c1`](https://github.com/thrasher-corp/gocryptotrader/commit/158e9c12e51d2a28e1976a6e6fdea0b58d4fc322) and pass here, and the compound-type parsing matches how LBank echoes order types.

Inline, none blocking: a Should fix on [the `GetActiveOrders` side filter](https://github.com/thrasher-corp/gocryptotrader/pull/2409#discussion_r4077406499), a Should fix on [two tests that pass on base](https://github.com/thrasher-corp/gocryptotrader/pull/2409#discussion_r4077406508), and Nits on [a comment's endpoint](https://github.com/thrasher-corp/gocryptotrader/pull/2409#discussion_r4077406512) and [the repeated side parsing](https://github.com/thrasher-corp/gocryptotrader/pull/2409#discussion_r4077406516).

The description's closing "还原确认" section looks like leftover working notes; worth trimming.

- `GetOrderHistory` advances the page per order, not per page: a two-order first page requests pages `[1, 1, 3]` ([lbank_wrapper.go:697](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/lbank/lbank_wrapper.go#L697)).
- `GetOrderInfo` ignores its pair and searches only open orders, so a filled or cancelled order is never found ([lbank_wrapper.go:466](https://github.com/thrasher-corp/gocryptotrader/blob/44cb7b39a42662f05071bae212f818c41d20ace0/exchanges/lbank/lbank_wrapper.go#L466)).

Take them or leave them. Nothing here blocks.

PR #2410, opened 2026-09-22
fix(binanceus): forward orderId in fiat deposit and withdrawal history

Robin1987China:
## Summary

`Exchange.FiatWithdrawalHistory` and `Exchange.FiatDepositHistory` both take a `*FiatWithdrawalRequestParams`, whose `OrderID` field is part of the public request struct but was never copied into the request query string. Callers therefore had no way to filter fiat history down to a single transfer by its `orderId`.

Binance.US documents `orderId` as an optional filter parameter on both endpoints. From the official API docs (`binance-us/binance-us-api-docs`, `wapi-api.md`), the parameter tables read:

- `GET /sapi/v1/fiatpayment/query/withdraw/history` — `orderId | STRING | NO`
- `GET /sapi/v1/fiatpayment/query/deposit/history` — `orderId | STRING | NO`

So the field was accepted, documented upstream, and silently dropped.

## Fix

`exchanges/binanceus/binanceus.go`: in both request builders, set `orderId` when `arg.OrderID` is non-empty, matching the parameter style already used in this file:

```go
if arg.OrderID != "" {
	params.Set("orderId", arg.OrderID)
}
```

No behaviour change when `OrderID` is empty (the parameter is simply omitted, preserving the previous request shape).

## Tests

`exchanges/binanceus/binanceus_test.go`: new offline table test `TestFiatHistoryForwardsOrderID` covering both endpoints against a stubbed `httptest` server (`SetHTTPClient` + `SetRunningURL`). It asserts:

1. the request path is the documented endpoint;
2. `orderId=<value>` is present in the query when `OrderID` is supplied;
3. `orderId` is absent when `OrderID` is empty.

No live/real exchange endpoint is contacted.

## Verification

- `go build ./...` — pass
- `go test ./exchanges/binanceus/ -run TestFiatHistoryForwardsOrderID -v` — pass (4/4 subtests)
- `go test ./exchanges/binanceus/ -run TestFiatHistoryForwardsOrderID -race -count 2` — pass
- `go vet ./exchanges/binanceus/` — pass
- `gofmt -l exchanges/binanceus/` — clean
- `bash scripts/misc_checks.sh` — all miscellaneous checks passed
- `golangci-lint run ./exchanges/binanceus/...` (v2.13.2) — 0 issues

Reverse verification: reverting only the two `params.Set("orderId", ...)` blocks makes both "forwards order ID" subtests fail with `expected: "1337", actual: ""`, confirming the new test actually exercises the fix rather than passing vacuously.

## Not verified

- No live authenticated request was made against Binance.US; the remote filtering behaviour relies on the documented API contract.
- The fiat history endpoints otherwise remain covered only by credential-gated live tests for the non-`orderId` path.


thrasher-:
@codex please review

exchanges/binanceus/binanceus_test.go
```diff
@@ -1157,6 +1159,76 @@ func TestFiatDepositHistory(t *testing.T) {
...
+			ex := new(Exchange)
+			require.NoError(t, testexch.Setup(ex), "Setup must not error")
+			ex.SkipAuthCheck = true
+			ex.SetCredentials(&accounts.Credentials{
+				Key:    "test-key",
+				Secret: "test-secret",
+			})
+			server := httptest.NewTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				assert.Equal(t, tc.path, r.URL.Path, "the fiat history request should target the documented endpoint")
+				assert.Equal(t, tc.wantOrderID, r.URL.Query().Get("orderId"), "the orderId query parameter should be forwarded only when set")
```
thrasher-, line 1218: Heimdall: **Minor** — the two "omits order ID" cases can't fail: `Query().Get("orderId")` returns `""` whether the key is absent or sent empty as `orderId=`.

**Fix:** assert presence as well as value:

```go
assert.Equal(t, tc.wantOrderID != "", r.URL.Query().Has("orderId"), "orderId should only be sent when set")
```

**Why:** dropping the `if arg.OrderID != ""` guard from both functions, so every request without an order ID carries `orderId=`, leaves all four subtests passing:

```
                                    this test          + Has() line
HEAD                                4/4 pass           4/4 pass
guard dropped (sends orderId=)      4/4 pass           omits cases FAIL
fix reverted                        forwards FAIL      forwards FAIL
```

Sending it empty isn't harmless: Binance.US rejects empty parameters, e.g. `GET /api/v3/ticker/price?symbol=` → `{"code":-1105,"msg":"Parameter 'symbol' was empty."}`. I couldn't confirm the signed `/sapi` fiat endpoints do the same without credentials.

exchanges/binanceus/binanceus_test.go
```diff
@@ -1157,6 +1159,76 @@ func TestFiatDepositHistory(t *testing.T) {
...
+			t.Parallel()
+			ex := new(Exchange)
+			require.NoError(t, testexch.Setup(ex), "Setup must not error")
+			ex.SkipAuthCheck = true
+			ex.SetCredentials(&accounts.Credentials{
+				Key:    "test-key",
+				Secret: "test-secret",
+			})
+			server := httptest.NewTestServer(t, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				assert.Equal(t, tc.path, r.URL.Path, "the fiat history request should target the documented endpoint")
```
thrasher-, line 1217: Heimdall: **Nit** — several regressions still pass this test: the path is checked against the same constants the code sends, the handler's assertions only run if a request arrives, and `1337` survives being parsed as a number.

**Fix:** assert the documented paths as literals, count requests, and add a string ID shaped like the docs' examples alongside the existing inputs. A method-expression table also collapses the four closures (two identical pairs, each borrowing the parent test's `t` for its context), and it folds in the presence check from the other comment:

```go
func TestFiatHistoryForwardsOrderID(t *testing.T) {
	t.Parallel()
	for _, tc := range []struct {
		name    string
		path    string
		history func(*Exchange, context.Context, *FiatWithdrawalRequestParams) (FiatAssetsHistory, error)
	}{
		{name: "withdrawal", path: "/sapi/v1/fiatpayment/query/withdraw/history", history: (*Exchange).FiatWithdrawalHistory},
		{name: "deposit", path: "/sapi/v1/fiatpayment/query/deposit/history", history: (*Exchange).FiatDepositHistory},
	} {
		for _, orderID := range []string{"1337", "6c2ff984890145fdac2b7160299062f0", ""} {
			t.Run(tc.name+" orderId="+orderID, func(t *testing.T) {
				t.Parallel()
				ex := new(Exchange)
				ex.SkipAuthCheck = true
				ex.SetCredentials(&accounts.Credentials{Key: "test-key", Secret: "test-secret"})
				var requests atomic.Int64
					requests.Add(1)
					assert.Equal(t, tc.path, r.URL.Path, "request should target the documented endpoint")
					assert.Equal(t, orderID != "", r.URL.Query().Has("orderId"), "orderId should only be sent when set")
					assert.Equal(t, orderID, r.URL.Query().Get("orderId"), "orderId should match the supplied value")
					_, err := w.Write([]byte(`{"assetLogRecordList":[]}`))
					assert.NoError(t, err, "writing the response should not error")
				}))
				require.NoError(t, ex.API.Endpoints.SetRunningURL(exchange.RestSpotSupplementary.String(), server.URL), "SetRunningURL must not error")
				_, err := tc.history(ex, t.Context(), &FiatWithdrawalRequestParams{FiatCurrency: "USD", OrderID: orderID})
				require.NoError(t, err, "fiat history request must not error")
				assert.Equal(t, int64(1), requests.Load(), "exactly one request should reach the server")
			})
		}
	}
}
```

(`context` and `sync/atomic` join the imports.)

**Why:** each mutation applied to both functions in `binanceus.go`:

```
                                            this test      above
HEAD                                        4/4 pass       6/6 pass
fix reverted                                2 FAIL         4 FAIL
guard dropped (sends orderId=)              4/4 pass       2 FAIL
constants point at paths that 404 live      4/4 pass       6 FAIL
constants swapped between the methods       4 FAIL         6 FAIL
returns before sending                      4/4 pass       6 FAIL
orderId parsed as a number                  4/4 pass       2 FAIL
```

Lint-clean, and passes `-race -count 3 -cpu 1,2,8 -shuffle on`.

thrasher- (approved):
Heimdall: Approving. Both endpoints document `orderId` (STRING, optional), the signed request now carries `orderId=1337` and leaves it out when unset, and reverting the two hunks fails both "forwards" subtests.

Inline, none blocking: a Minor on [binanceus_test.go:1218](https://github.com/thrasher-corp/gocryptotrader/pull/2410#discussion_r4077737907) (the omit cases can't fail) and a Nit on [binanceus_test.go:1217](https://github.com/thrasher-corp/gocryptotrader/pull/2410#discussion_r4077737915) (regressions the test still lets through, with a drop-in replacement).

- Both functions drop a past `EndTime`, sending only future ones ([binanceus.go:1420](https://github.com/thrasher-corp/gocryptotrader/blob/57b97292dde5896332a32513f2412ab8fd30ba18/exchanges/binanceus/binanceus.go#L1420), [binanceus.go:1552](https://github.com/thrasher-corp/gocryptotrader/blob/57b97292dde5896332a32513f2412ab8fd30ba18/exchanges/binanceus/binanceus.go#L1552)).
- The same never-sent parameter recurs: `GetTrades` drops `orderId` ([binanceus.go:1035](https://github.com/thrasher-corp/gocryptotrader/blob/57b97292dde5896332a32513f2412ab8fd30ba18/exchanges/binanceus/binanceus.go#L1035)); `CancelOCOOrder` sends `GET` without `symbol`, so it queries rather than cancels ([binanceus.go:1189](https://github.com/thrasher-corp/gocryptotrader/blob/57b97292dde5896332a32513f2412ab8fd30ba18/exchanges/binanceus/binanceus.go#L1189)); `WithdrawFiat` posts to a path that 404s and never sends `paymentAccount` or `amount` ([binanceus.go:95](https://github.com/thrasher-corp/gocryptotrader/blob/57b97292dde5896332a32513f2412ab8fd30ba18/exchanges/binanceus/binanceus.go#L95), [binanceus.go:1450](https://github.com/thrasher-corp/gocryptotrader/blob/57b97292dde5896332a32513f2412ab8fd30ba18/exchanges/binanceus/binanceus.go#L1450)).

CI: the three red jobs in [run 35713241082](https://github.com/thrasher-corp/gocryptotrader/actions/runs/35713241082) fail only in kucoin and deribit live wrapper tests, which pass in later unrelated runs; binanceus passes in every job.

